PR #236, opened 2021-06-02
Simulator work to enable random pool/lockup/pot creation

ValarDragon:
This PR adds operations for creating pools, pots, and lockups.

This also adds some simulation infrastructure for convenience, to save us some lines that plague current SDK operations.


ValarDragon:
Update: This PR is just going to have the pool creation operation, and the WIP infra for getting the entire state machine simulator setup.

This is to better enable @mconcat to work in parallel on adding more simulator ops

PR #251, opened 2021-06-06
Fix start time param to be nullable 

mconcat:
There was also an error on parameter setting under `setInitialPoolParams`


x/gamm/types/marshal_test.go
```diff
@@ -43,7 +43,7 @@ func TestPoolMarshalYAML(t *testing.T) {
     swap_fee: "0.025000000000000000"
     exit_fee: "0.025000000000000000"
     smooth_weight_change_params: null
-    start_time: 0001-01-01T00:00:00Z
+    start_time: null
```
ValarDragon, line 46: Lets make one of these not null, and lets make a TODO for figuring out how to better generalize for more optionality on parameters in the future

PR #257, opened 2021-06-08
test denom map active pot ids

### LINKED ISSUE #204: Add tests for the denom -> active pot ID map
Currently in #200 this was only indirectly tested. We should add direct tests for this map being correct

antstalepresh:
closes #204

### DIFF
--- a/x/incentives/keeper/gauge_test.go
+++ b/x/incentives/keeper/gauge_test.go
@@ -497,6 +497,14 @@ func (suite *KeeperTestSuite) TestPerpetualActiveGaugesByDenom() {
 	gaugeIds = suite.app.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.ctx, "lptoken")
 	suite.Require().Len(gaugeIds, 1)
 
+	// check gauge ids by other denom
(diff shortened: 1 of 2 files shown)


x/incentives/keeper/gauge_test.go
```diff
@@ -421,3 +429,87 @@ func (suite *KeeperTestSuite) TestNoLockNonPerpetualGaugeDistribution() {
...
+	// finish distribution
+	err = suite.app.IncentivesKeeper.FinishDistribution(suite.ctx, *gauge)
+	suite.Require().NoError(err)
+
+	// check gauge ids by denom
+	gaugeIds = suite.app.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.ctx, "lptoken")
+	suite.Require().Len(gaugeIds, 0)
+}
+
+func (suite *KeeperTestSuite) TestPerpetualActiveGaugesByDenom() {
```
ValarDragon, line 473: Can we combine the perp and non-perp tests into one. (Eventually we should make it table driven, but I don't think thats that important for now)

ValarDragon: I'll make this a TODO. atm an extra test prelaunch is fine haha

PR #260, opened 2021-06-09
Incentives missing endpoints

antstalepresh:


ValarDragon:
LGTM on brief pass, did not do an in-depth pass since its for the CLI (if there was a bug in getting response from a query, its very fixable post-launch)

PR #261, opened 2021-06-09
Missing endpoints for pool-incentives 

antstalepresh:


ValarDragon:
Post-launch, we should make tests with more setup, so we can test the responses as well. Not super important prelaunch though

PR #262, opened 2021-06-10
Fix setting initial pool start time bug

ValarDragon:


ValarDragon:
This should be re-investigated post launch

PR #266, opened 2021-06-10
Modify codedov.yml 

mconcat:


ValarDragon:
Don't really know what 
```


parsers:
  gcov:
    branch_detection:
      conditional: yes
      loop: yes
      method: no
      macro: no

comment:
  layout: "reach,diff,flags,files,footer"
  behavior: default
  require_changes: no
```

means, and punting on looking into them. Post-launch lets add some brief docs / links to docs here explaining what they are

PR #272, opened 2021-06-11
Add manual testing for x/gamm

mconcat:


x/lockup/keeper/behaviour.go
```diff
@@ -0,0 +1,309 @@
...
+	"time"
+
+	"github.com/stretchr/testify/suite"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/x/lockup/types"
+)
+
+type BehaviourAccount struct {
```
ValarDragon, line 14: Can you add some brief docs for what is a behavior account

x/lockup/keeper/behaviour.go
```diff
@@ -0,0 +1,309 @@
+package keeper
+
```
ValarDragon, line 2: lets add a todo for moving this to either an internaal folder or simulation folder, in the future. (Fine to be here for launch)

x/lockup/keeper/behaviour_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (suite *KeeperTestSuite) TestBehaviours() {
+	suite.SetupTest()
+
+	denoms := make([]string, 10)
+	for i := range denoms {
+		denoms[i] = fmt.Sprintf("token%d", i)
+	}
+
+	accounts := make([]*keeper.BehaviourAccount, 20)
```
ValarDragon, line 23:
```suggestion
    // Make 20 accounts
    	accounts := make([]*keeper.BehaviourAccount, 20) 
    // For each account, choose a random subset of denominations / coins to include in it.
    // And set them to have that balance. (Choice of coins is currently 50% probability for each denom, and random amount of coins)
```

ValarDragon:
As discussed on weekly call, this was really helpful for getting us some launch simulations, but likely needs significant reworks before we can get in a state that will help simulate main now.

Closing for now!

PR #292, opened 2021-06-13
Limit Gauges per Denom

### LINKED ISSUE #289: limit number of gauges per denom
Currently the distribution code is slower than we expected, and runs in time `O(sum_{denoms} (#gauges for that denom) * (#lockups for that denom)`. We expected to instead have been `O(sum_{denoms} (#gauges for that denom) + (#lockups for that denom)`. This is something thats not hard to fix post-launch, its just that fixing it doesn't fit into our launch timeline.

In light of this, we've decided that the best way to proceed for launch is to limit the number of gauges for each denom.

We want to limit the creatable gauges to just be ones that have a DistrCondition based on `Duration` (so for now just rejecting ones with a StartTime based DistrCondition. We can work out deleting all that code post-launch)

Then we also want to limit the creatable durations for a denom to just be one of a small number of durations. Lets for now just make it the same set as used in pool-incentives (but the parameters being managed separately). We should also ensure that only one gauge for a `(denom, duration)` pair can be created.

antstalepresh:
closes #289


x/incentives/keeper/gauge.go
```diff
@@ -103,6 +103,19 @@ func (k Keeper) SetGaugeWithRefKey(ctx sdk.Context, gauge *types.Gauge) error {
 
 // CreateGauge create a gauge and send coins to the gauge
 func (k Keeper) CreateGauge(ctx sdk.Context, isPerpetual bool, owner sdk.AccAddress, coins sdk.Coins, distrTo lockuptypes.QueryCondition, startTime time.Time, numEpochsPaidOver uint64) (uint64, error) {
+	durations := k.GetLockableDurations(ctx)
+	if distrTo.LockQueryType == lockuptypes.ByDuration {
+		durationOk := false
+		for _, duration := range durations {
+			if duration == distrTo.Duration {
+				durationOk = true
```
ValarDragon, line 111:
```suggestion
				durationOk = true
				break
```

x/incentives/simulation/genesis.go
```diff
@@ -14,23 +15,29 @@ import (
...
 		func(r *rand.Rand) { distrEpochIdentifier = GenParamsDistrEpochIdentifier(r) },
 	)
 
 	incentivesGenesis := types.GenesisState{
-		Params: types.Params {
+		Params: types.Params{
 			DistrEpochIdentifier: distrEpochIdentifier,
 		},
 		// Gauges: gauges,
+		LockableDurations: []time.Duration{
```
ValarDragon, line 35: TODO: Double check these make sense with epoch times and the defaults in pool-incentives

x/incentives/types/msgs.go
```diff
@@ -50,6 +50,10 @@ func (m MsgCreateGauge) ValidateBasic() error {
 		return errors.New("distribution period should be 1 epoch for perpetual gauge")
 	}
 
+	if lockuptypes.LockQueryType_name[int32(m.DistributeTo.LockQueryType)] != "ByDuration" {
+		return errors.New("only duration condition query is allowed for postlaunch")
```
ValarDragon, line 54:
```suggestion
		return errors.New("only duration query condition is allowed. Start time distr conditions is an obsolete codepath slated for deletion")
```

ValarDragon (approved):
Will update genesis params in a second PR

PR #293, opened 2021-06-13
Make developer vesting tokens exist in a module account

ValarDragon:
What remains in this PR is updating all the tests to seed the developer vesting account with sufficient funds for the relevant test.

### DIFF
--- a/x/mint/keeper/hooks_test.go
+++ b/x/mint/keeper/hooks_test.go
@@ -35,12 +35,16 @@ func TestEndOfEpochMintedCoinDistribution(t *testing.T) {
 	}
 	app.MintKeeper.SetParams(ctx, mintParams)
 
+	// setup developer rewards account
+	app.MintKeeper.CreateDeveloperVestingModuleAccount(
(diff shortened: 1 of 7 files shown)


x/mint/keeper/hooks_test.go
```diff
@@ -39,6 +39,8 @@ func TestEndOfEpochMintedCoinDistribution(t *testing.T) {
 	lastHalvenPeriod := app.MintKeeper.GetLastHalvenEpochNum(ctx)
 	// correct rewards
 	for ; height < lastHalvenPeriod+app.MintKeeper.GetParams(ctx).ReductionPeriodInEpochs; height++ {
+		app.MintKeeper.CreateDeveloperVestingModuleAccount(
+			ctx, sdk.NewCoin("stake", sdk.NewInt(500000)))
```
ValarDragon, line 43: It would be better to only create this vesting account once, with the correct balance.

Then on lines 56/57 we would just test that the balance decreased by the correct amount in that iteration.

antstalepresh: done

ValarDragon (commented):
LGTM

PR #294, opened 2021-06-13
Remove pool startTime from GAMM

### LINKED ISSUE #290: Remove pool start time from GAMM
We want to go with another approach for fixing the problem that this was created for. This code currently has a bug anyways (#262)

antstalepresh:
closes #290 

Am I missing removing few other start time options? (Not sure to keep this or remove)
* PoolFileStartTime
* smoothWeightChangeParamsInputs.StartTime


ValarDragon (approved):
LGTM. Re the questions, we want to keep LBP start time

ValarDragon:
EDIT: I'm not sure what PoolFileStartTime is, maybe that should be removed as well. Thats for the CLI though, so maybe we can do that post-launch?

ValarDragon:
Per discussion with Sunny, we are going ahead and deleting this for now.

We do still think its a good feature, but its going to take UI integration to make any sense, which isn't going to happen prelaunch. So we should go ahead and delete it to remove our bug surface prelaunch.

PR #296, opened 2021-06-13
Only perpetual gauges to be target of pool-incentives

### LINKED ISSUE #230: Make sure poolincentives can only add OSMO incentives to a perpetual pot

antstalepresh:
closes #230


sunnya97:
@antstalepresh Can you add a test for this?

ValarDragon (commented):
This is good, but we should also ensure that its impossible to make an update proposal that adds something to a non-perp pot

ValarDragon:
Closed in favor #298

PR #304, opened 2021-06-14
Removed duplicate vesting tx command

bsrinivas8687:


ValarDragon (approved):
Great catch, thanks for fixing this

PR #317, opened 2021-06-16
Speedup distribution

### LINKED ISSUE #310: Optimize x/incentives distribution function
The current distribution logic iterates over O(g*l). 

Proposal:

We maintain a mapping from gauges to distributable locks. When a lock is created/unlocked, it iterates over the existing gauges and updates the map accordingly. When a gauge is created/expired, it iterates over the existing locks and updates the map accordingly.

We maintain a mapping from locks to reward entry, which is a pair of `(gaugeID, startEpoch, endEpoch)`, which means "the lock has ownership on rewards by gaugeID that has been distributed over [startEpoch, endEpoch)". 

At each distribution, it snapshots the total locked coins and total reward coins. The pair is stored under the mapping from `(gaugeID, epochNum)`.

Users can claim rewards that they have ownership of, by submitting lock IDs that they own. The state machine will iterate over the reward entries and send the tokens using the distribution snapshot. 

Currently, locks are removed from the state once they are fully unlocked, but for this, we need to maintain the lock ownership info permanently.

ValarDragon:
Close #310, potentially removes the need for #289

Converts distribution speed from `O(sum_{denom} #lockups_for_denom * #num_gauges_for_denom)` to instead be `O-tilde(#lockups_for_denom + #num_gauges_for_denom)`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,4 +29,4 @@ replace google.golang.org/grpc => google.golang.org/grpc v1.33.2
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.2-alpha.regen.4
 
-replace github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.42.5-0.20210613202848-b45039095bc3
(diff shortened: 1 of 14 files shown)


ValarDragon:
Punting to post launch, too many bugs andsubtle

PR #320, opened 2021-06-16
 improvements to incentives try 3

ValarDragon:
This takes some of the improvements of #317 and puts them into the old algo


ValarDragon:
Not so conservative lol

mconcat:
If I understood correctly, the change made in this PR uses an accumulation store to calculate the total locked amount and use it to calculate the reward per lockup, instead of iterate over them and sum up. I think this is a simple enough change to merge prelaunch.

ValarDragon:
I thought so as well, except we found out the problem is it fails simulation bank invariants

PR #326, opened 2021-06-17
GetRewardsEst grpc query constraints on end epoch

mconcat:
Took arbitrary number `100`, please suggest better number who knows better about this

### DIFF
--- a/x/incentives/keeper/grpc_query.go
+++ b/x/incentives/keeper/grpc_query.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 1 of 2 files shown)


sunnya97:
Would it be safe to make this 365, so we can get annualized APR?

x/incentives/keeper/grpc_query.go
```diff
@@ -112,6 +113,10 @@ func (k Keeper) UpcomingGauges(goCtx context.Context, req *types.UpcomingGaugesR
 // RewardsEst returns rewards estimation at a future specific time
 func (k Keeper) RewardsEst(goCtx context.Context, req *types.RewardsEstRequest) (*types.RewardsEstResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	diff := req.EndEpoch - k.GetEpochInfo(ctx).CurrentEpoch
+	if diff > 100 {
```
sunnya97, line 117:
```suggestion
	if diff > 365 {
```

PR #329, opened 2021-06-17
Test for get rewards est with pool incentives

ValarDragon:


ValarDragon:
I'm undoing the swapping of order here, so that this just updates tests, and is not state breaking.

PR #335, opened 2021-06-22
fixed parse

okwme:
fixes the `keys parse` bug: https://github.com/cosmos/cosmos-sdk/pull/9299


ValarDragon:
Thanks for fixing this billy!

PR #336, opened 2021-06-22
Resolve lock tokens gas growth

antstalepresh:


ValarDragon:
PR looks good to me! I think we should add docs specifying the key format expected for `AddLocksByRef`

PR #339, opened 2021-06-23
add genesis export for lockup module

antstalepresh:


x/lockup/genesis.go
```diff
@@ -9,9 +9,22 @@ import (
 // InitGenesis initializes the capability module's state from a provided genesis
 // state.
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
+	k.SetLastLockID(ctx, genState.LastLockId)
+	for _, lock := range genState.Locks {
+		if err := k.ResetLock(ctx, lock); err != nil {
```
ValarDragon, line 14: Lets add a comment for what is ResetLock here

app/app.go
```diff
@@ -297,6 +297,21 @@ func NewOsmosisApp(
 	)
 	app.UpgradeKeeper = upgradekeeper.NewKeeper(skipUpgradeHeights, keys[upgradetypes.StoreKey], appCodec, homePath)
 
+	// this configures a no-op upgrade handler for the "upgrade-lockup-module-store-management" upgrade
+	app.UpgradeKeeper.SetUpgradeHandler("upgrade-lockup-module-store-management", func(ctx sdk.Context, plan upgradetypes.Plan) {
+		locks, err := app.LockupKeeper.GetPeriodLocks(ctx)
```
ValarDragon, line 302: This approach looks right to me, just want to double check via a test that there isn't some edge case we're missing

antstalepresh: Added test

x/lockup/genesis.go
antstalepresh: Added comment

ValarDragon:
Tested yesterday, and LGTM.

PR #340, opened 2021-06-23
add in state upgrade for lockup

antstalepresh:


ValarDragon:
This looks good to me!

ValarDragon:
Actually I want to make sure this works completely before we do it, lets write a test for it

PR #347, opened 2021-06-24
Distribute not unlocking only

antstalepresh:


ValarDragon:
This isn't quite right, we want to have people who are unlocking still get rewards, just only rewards according to the amount of time they are still locked.

So if I had a lock for 15 days, and I unlock right now, I should:
* Get the 14 day reward for 1 day  (until my remaining unlocking time goes below 14 days)
* Get the 7 day rewards for 8 days (until my remaining unlocking time goes below 7 days)
* Get the 1 day reward for 14 days (until my remaining unlocking time goes below 1 day)

antstalepresh:
> This isn't quite right, we want to have people who are unlocking still get rewards, just only rewards according to the amount of time they are still locked.
> 
> So if I had a lock for 15 days, and I unlock right now, I should:
> 
> * Get the 14 day reward for 1 day  (until my remaining unlocking time goes below 14 days)
> * Get the 7 day rewards for 8 days (until my remaining unlocking time goes below 7 days)
> * Get the 1 day reward for 14 days (until my remaining unlocking time goes below 1 day)

Then we already have an existing function for that.
```
// GetLocksPastTimeDenom Returns the locks whose unlock time is beyond timestamp
func (k Keeper) GetLocksPastTimeDenom(ctx sdk.Context, denom string, timestamp time.Time) []types.PeriodLock {
	// returns both unlocking started and not started assuming it started unlocking current time
	unlockings := k.getLocksFromIterator(ctx, k.LockIteratorAfterTimeDenom(ctx, true, denom, timestamp))
	duration := time.Duration(0)
	if timestamp.After(ctx.BlockTime()) {
		duration = timestamp.Sub(ctx.BlockTime())
	}
	notUnlockings := k.getLocksFromIterator(ctx, k.LockIteratorLongerThanDurationDenom(ctx, false, denom, duration))
}
```
It queries by duration for not unlockings, and queries by endtime for unlocking started.

ValarDragon:
Sorry for the late response here. `GetLocksPastTimeDenom` should work!

We do need the accumulation store to be consistent with the actual total (This is because the accumulation store is critical for GetRewardsEst, which is how the front-end calculates APR's). (And we also need GetRewardsEst to be calculated fast)

I think we could make a second accumulation store for unbondings, using the subset-accumulation methods for getting this efficiently. (The key here would be the timestamp unbonding ends at) Will come back and write a more thought out API for that in a few min.

We probably can't get this into the upgrade tmrw unfortunately

ValarDragon:
So the way the second accumulation store would work is that it puts in entries only when they are unbonding. It then places them according to the unix time at which they stop unbonding / get auto-withdrawn. 

Then to get the rewards for 7 day lockups, you would get the total from the currently bonded folks as we do now. You would then add in, the amount from people who are unbonding and have an unbond end time that is after "current block time + 7 days".

Whenever someone's unbonding lock is auto-deleted, it would delete that timestamp entry from the accumulation store as well.

This is code that I believe would still get used when we add F1 here, so I think this is good to implement

antstalepresh:
> This is code that I believe would still get used when we add F1 here, so I think this is good to implement

Ok got it

antstalepresh:
One thing we need to care about is that we disabled time basis querying for speed optimization - only did duration basis.

ValarDragon:
Closing this in favor of the approach in #359, which is matching https://github.com/osmosis-labs/osmosis/pull/347#issuecomment-869378094

PR #348, opened 2021-06-24
automatic withdraw after unlock time pass

antstalepresh:


x/lockup/keeper/lock_test.go
```diff
@@ -312,3 +312,50 @@ func (suite *KeeperTestSuite) TestLockTokensAlot() {
 	}
 	// panic(1)
 }
+
+func (suite *KeeperTestSuite) TestEndblockerWithdrawAllMaturedLockups() {
```
ValarDragon, line 316: Can we have a test that shows if that locks with the following unbonding times are unbonded in sequence:

* 5 second lockup
* 1 second lockup
* 3 second lockup

and then test the following two situations:

After 1 second, only the 1 second is freed, 3 seconds the 3 second is freed, 5 seconds the 5 second is freed

After 5 seconds, all three are freed

x/lockup/keeper/lock.go
```diff
@@ -40,6 +40,10 @@ func (k Keeper) beginUnlockFromIterator(ctx sdk.Context, iterator db.Iterator) (
 	return locks, coins, nil
 }
 
+func (k Keeper) WithdrawAllMaturedLocks(ctx sdk.Context) {
```
ValarDragon, line 43:
```suggestion
// WithdrawAllMaturedLocks withdraws every lock thats in the process of unlocking, and has finished unlocking by
// the current block time. 
func (k Keeper) WithdrawAllMaturedLocks(ctx sdk.Context) {
```

x/lockup/spec/03_messages.md
```diff
@@ -23,38 +23,39 @@ type MsgLockTokens struct {
...
 
 Once time is over, users can withdraw unlocked coins from lockup `ModuleAccount`.
 
 ```go
-type MsgUnlockPeriodLock struct {
-  Owner  sdk.AccAddress
-  LockID uint64
+type MsgBeginUnlocking struct {
+	Owner string
+	ID    uint64
```
ValarDragon, line 50: I think this should still be called `LockID` right?

ValarDragon (approved):
This PR LGTM, I double checked the involved unlocking logic underneath. (Pleasantly surprised that this could be implemented with almost no changes to it!)

Lets merge once we get the test update in. This requires no store migrations

x/lockup/spec/03_messages.md
ValarDragon: I guess this would be proto breaking, so not worth doing

x/lockup/keeper/lock_test.go
ValarDragon: I'm working on adding this rn

ValarDragon: Added in #355

PR #363, opened 2021-06-30
Min commission change validation change ignore

antstalepresh:


ValarDragon:
Closed in favor of the must change commission rate approach. 

The scripts were super helpful though! (Minor bug, need to change `osmosisd tx lockup unlock-by-id` to `begin-unlock-by-id`)

PR #369, opened 2021-07-05
Export import genesis for gamm

antstalepresh:


x/gamm/genesis.go
```diff
@@ -0,0 +1,60 @@
...
+
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/osmosis-labs/osmosis/x/gamm/keeper"
+	"github.com/osmosis-labs/osmosis/x/gamm/types"
+)
+
+// InitGenesis initializes the capability module's state from a provided genesis
+// state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState, unpacker codectypes.AnyUnpacker) {
```
ValarDragon, line 14: Why do pass in an unpacker, instead of having a types.UnpackPools method, ala https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/genesis.go#L16

ValarDragon: I think doing a second iteration here would be worth it just for ease of readability in the future

ValarDragon (approved):
LGTM, sans small point around API for init genesis

x/gamm/genesis_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}, {
+		Weight: sdk.NewInt(100),
+		Token:  sdk.NewCoin("bar", sdk.NewInt(10000)),
+	}}, "")
+	require.NoError(t, err)
+
+	genesis := am.ExportGenesis(ctx, appCodec)
+	assert.NotPanics(t, func() {
+		am.InitGenesis(ctx, appCodec, genesis)
+	})
```
ValarDragon, line 132: Can we also have a consistency test here to ensure that the decoded pool is the same as the originally encoded one?

x/gamm/genesis.go
antstalepresh: I tried GetCachedValue solution but it didn't work - just spent a lot of time already and finally staying here

x/gamm/genesis_test.go
antstalepresh: We can, just to confirm, you can check v1.0.4 to see more detailed test, https://github.com/osmosis-labs/osmosis/pull/371/files

ValarDragon: oh perfect, thanks

PR #375, opened 2021-07-09
Dev/v2 rc1

ValarDragon:
Merge v2 fixes into main. We will then make a rc2 branch, so that all the updates accumulated into main also get included.


ValarDragon:
These commits were reviewed individually over the week.

PR #381, opened 2021-07-14
auto staking some percentage of lp rewards

### LINKED ISSUE #379: Implement auto-staking of some percentage of LP pool-incentives
Due to high OSMO LP incentives, we should implement a feature where {X} % of LP incentives get automatically staked. 

The structure of a first pass at this should be as follows in my opinion:
* Make a governance parameter for what is the percentage of LP pool-incentives that get automatically staked. (Lets just make a default of 50%)
* Make the bonding tx specify an optional validator address field, where all (if any) of your OSMO pool-incentives rewards will get staked to.
* Update pool incentives to have X% of the rewards directly put into your account, and 1-X% get delegated to your validator of choice. (This will require some changes to the Distribute API to make this possible)
  - Unfortunately this is going to cause a lot of F1 overhead. We will need to benchmark how it is, and consider implementing some batched F1 logic. This is something I worked out the math of before, will rederive if benchmarks show its important. 
* If the validator address is unspecified (e.g. all existing bonds), have those funds put those OSMO into an unbonding day lockup.

Then in a future pass, we can:
* update the optional validator address field to instead be a weighted list of validators
* allow editing the validator address for an existing bond

antstalepresh:
closes #379


proto/osmosis/incentives/autostaking.proto
```diff
@@ -0,0 +1,19 @@
...
+
+option go_package = "github.com/osmosis-labs/osmosis/x/incentives/types";
+
+message AutoStaking {
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.stringer)         = true;
+
+  string address = 1 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+  string autostaking_validator = 2;
+  string autostaking_rate = 3 [
```
ValarDragon, line 14: Can we implement this without an autostaking rate for the first version, and just use the governance param?

I think this is going to make many types of batch optimizations that we will likely need to later do for speed harder, and we should consider this once we know where stand in speed.

x/incentives/keeper/msg_server.go
```diff
@@ -64,3 +64,13 @@ func (server msgServer) AddToGauge(goCtx context.Context, msg *types.MsgAddToGau
 
 	return &types.MsgAddToGaugeResponse{}, nil
 }
+
+func (server msgServer) SetAutoStaking(goCtx context.Context, msg *types.MsgSetAutoStaking) (*types.MsgSetAutoStakingResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	err := server.keeper.SetAutostaking(ctx, &types.AutoStaking{
+		Address:              msg.Address,
```
ValarDragon, line 71: Great idea on just having an address just set their validator preference once. 

What I was thinking of was more complicated, this is definitely better!

x/incentives/keeper/autostaking.go
```diff
@@ -0,0 +1,110 @@
...
+	if autostaking != nil { // lock tokens forcefully - TODO: if lock tokens on every epoch, lots of locks will appear
+		valAddr, err := sdk.ValAddressFromBech32(autostaking.AutostakingValidator)
+		if err != nil {
+			return err
+		}
+
+		validator, found := k.sk.GetValidator(ctx, valAddr)
+		if found {
+			// NOTE: source funds are always unbonded
+			_, err = k.sk.Delegate(ctx, owner, autoDelegationAmt, stakingtypes.Unbonded, validator, true)
```
ValarDragon, line 96: Can there be a variable describing what the argument represents for the `stakingtypes.Unbonded` and `true` here?

e.g. `{arg_name} := stakingtypes.Unbonded`, `arg_name := true`. I currently have no idea what they are right now

x/incentives/keeper/gauge.go
```diff
@@ -306,6 +306,15 @@ func (k Keeper) Distribute(ctx sdk.Context, gauge types.Gauge) (sdk.Coins, error
 		if err := k.bk.SendCoinsFromModuleToAccount(ctx, types.ModuleName, owner, distrCoins); err != nil {
 			return nil, err
 		}
+		lockOwner, err := sdk.AccAddressFromBech32(lock.Owner)
```
ValarDragon, line 309: Above lines still need to be updated for auto-staking right? (Multiplying by `1 - auto-staking` decimal above, and `auto-staking` fraction below

proto/osmosis/incentives/autostaking.proto
antstalepresh: Disabled

x/incentives/keeper/msg_server.go
antstalepresh: +1

x/incentives/keeper/gauge.go
antstalepresh: The process is coins are sent to user fully by 100% and after that specific percentage is auto-staked.
The reason is because, not only OSMO is rewarded but other tokens are also rewarded as well. And it's convenient to reward everything and just delegate percentage of OSMO income for auto-staking.

x/incentives/keeper/autostaking.go
antstalepresh: Added comments

x/lockup/keeper/lock.go
```diff
@@ -281,6 +286,43 @@ func (k Keeper) UnlockPeriodLockByID(ctx sdk.Context, LockID uint64) (*types.Per
...
+	}
+	lock.Coins = lock.Coins.Add(coins...)
+
+	// store lock object into the store
+	store := ctx.KVStore(k.storeKey)
+	bz, err := proto.Marshal(lock)
+	if err != nil {
+		return nil, err
+	}
+	store.Set(lockStoreKey(lock.ID), bz)
```
ValarDragon, line 309: Shouldn't there be an existing method for saving the lock to the store?

ValarDragon: If theres no existing method, there should at least be a way to reduce the code duplication here

x/lockup/keeper/lock.go
```diff
@@ -281,6 +286,43 @@ func (k Keeper) UnlockPeriodLockByID(ctx sdk.Context, LockID uint64) (*types.Per
...
+	// modifications to accumulation store
+	for _, coin := range lock.Coins {
+		// remove previous store for the lock ID and add again with updated value
+		k.accumulationStore(ctx, coin.Denom).Remove(accumulationKey(lock.Duration, lock.ID))
+		k.accumulationStore(ctx, coin.Denom).Set(accumulationKey(lock.Duration, lock.ID), coin.Amount)
+	}
+
+	if k.hooks == nil {
+		return lock, nil
+	}
```
ValarDragon, line 320: Whats the intuition for why we have this check? Is there some performance concern?

antstalepresh: There is an issue with app initialization script for lockup keeper. Incentives module keep lockup keeper but that keeper is not initialized with hooks because there are circular dependencies in initialization. And for temporary fix, just added the code. And lockup hooks are not used now fyi.

x/lockup/keeper/lock.go
antstalepresh: Probably, will reduce duplications

antstalepresh: Reduced duplication on this commit https://github.com/osmosis-labs/osmosis/pull/381/commits/c39118acb6688bae2e99189c3b988d2aa6f86b28

x/incentives/keeper/autostaking.go
```diff
@@ -0,0 +1,123 @@
...
+			_, err := k.lk.AddTokensToLock(ctx, owner, locks[0].ID, sdk.Coins{sdk.NewCoin(bondDenom, autoDelegationAmt)})
+			if err != nil {
+				return err
+			}
+		} else { // create a new lock if this lock is not available
+			_, err := k.lk.LockTokens(ctx, owner, sdk.Coins{sdk.NewCoin(bondDenom, autoDelegationAmt)}, time.Hour*24*7*2)
+			if err != nil {
+				return err
+			}
+		}
```
ValarDragon, line 118: Can we make `k.lk.AddTokensToLock` do all of this logic? (Namely creating the lock if it doesn't already exist)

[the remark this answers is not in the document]
antstalepresh: Hmm, it will be confusing if a user need to provide lockID when creating a new one. Also the underlying functionality is different, creating a new lock is doing different stuff with Adding tokens to existing lock.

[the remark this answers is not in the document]
ValarDragon: Oh I see, you're totally right! What if it was `k.lk.AddTokensToLockedDurationDenom` that did all this logic?

ValarDragon:
I don't see any of the relevant ValidatorSetPreferences logic in this PR, should we close the PR?

PR #383, opened 2021-07-18
Update README.md

faddat:
I figure these belong approximately in that spot.  We can try state sync when 2.0 comes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 # Osmosis
-
 _(Note: This repository is under active development. Architecture and implementation may change without documentation)_
 
 Osmosis is a fair-launched, customizable automated market maker for interchain assets that allows the creation and management of non-custodial, self-balancing, interchain token index similar to one of Balancer.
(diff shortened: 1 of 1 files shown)


faddat:
Adding genesis now, kinda neglected that.

faddat:
I could also adapt this to use cosmos-omnibus?

ValarDragon (commented):
Instructions look great to me!

I think we should put this into `networks`, and have this repo point to the instructions there. What are other folks thoughts on this?

thecryptodrive:
@ValarDragon I think that is a great idea, I'm actually putting together a longer form instruction set with more verbosity and additional dependencies installs, will share soon.

antstalepresh:
> @ValarDragon I think that is a great idea, I'm actually putting together a longer form instruction set with more verbosity and additional dependencies installs, will share soon.

faddat:
@ValarDragon I assure you that the pattern of putting this info in `networks` super confuses newcomers.

Hence I propose that this repo's readme be like:

Product definition

(Osmosis is........)


Quick start


Other stuff

faddat (commented):
changed to v1.0.2 ; had some issues with .3 the other day.

xla:
This PR if merged would have saved me a good 20mins on running a full node locally.

faddat:
@xla I'll update e'r and we'll get e'r merged.

ValarDragon (approved):

PR #386, opened 2021-07-22
Pool creation fee

antstalepresh:


x/pool-incentives/keeper/keeper_test.go
```diff
@@ -52,6 +52,7 @@ func (suite *KeeperTestSuite) preparePoolWithPoolParams(poolParams gammtypes.Poo
 			suite.ctx,
 			acc,
 			sdk.NewCoins(
+				sdk.NewCoin("uosmo", sdk.NewInt(10000000000)),
```
ValarDragon, line 55: This is just a drive-by change right? Didn't really affect anything relative to this PR AFAICT. (If so, thats fine)

x/gamm/types/params.go
```diff
@@ -0,0 +1,70 @@
...
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v.Validate() != nil {
+		return fmt.Errorf("invalid pool creation fee: %+v", i)
+	}
+
+	if v.IsZero() {
+		return fmt.Errorf("zero pool creation fee is not allowed: %+v", i)
+	}
```
ValarDragon, line 63: Can we allow a `0` validation fee?

x/gamm/types/params.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Implements params.ParamSet
+func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyPoolCreationFee, &p.PoolCreationFee, validatePoolCreationFee),
+	}
+}
+
+func validatePoolCreationFee(i interface{}) error {
```
ValarDragon, line 51: Inspired the deposit bug, we should also add checking that the new balance is a balance that actually exists on chain. We can do that later though.

ValarDragon (approved):
LGTM! Only question is that can we easily allow 0 fee. If not, its fine

x/gamm/types/params.go
ValarDragon: (Like would we run into an error somewhere with coins.Valid() or sdk.NewCoins

antstalepresh: Will allow zero fee

x/gamm/types/params.go
antstalepresh: This is basic validation - and it's not using the kvstore - for additional check can set the validation function on SetParams probably.

x/gamm/types/params.go
antstalepresh: allowed empty fee

ValarDragon (approved):
Great job on this! LGTM

PR #388, opened 2021-07-25
Accumulation store optimization for x/lockup

mconcat:
Migration on lockup accumulation store to use amount per duration, not amount per lockup


mconcat:
Closing in favor of #389

PR #389, opened 2021-07-26
Add per-duration model for lockup accstore

mconcat:


x/lockup/keeper/lock.go
```diff
@@ -309,6 +309,13 @@ func (k Keeper) ClearAllLockRefKeys(ctx sdk.Context) {
 	k.clearLockRefKeysByPrefix(ctx, types.KeyPrefixUnlocking)
 }
 
+func (k Keeper) ClearAllAccumulationStores(ctx sdk.Context) {
+  denoms := k.getAccumulationDenoms(ctx)
```
ValarDragon, line 313: Can we combine the two methods, and just delete everything under the subspace `prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixLockAccumulation)` ?

ValarDragon:
This was merged via #479

Issue #150, opened 2021-05-03
Make epochs/GenesisState validation ensure identifiers are unique


ValarDragon:
the x/epochs module has many epoch definitions provided at genesis. We should do some validation to ensure that multiple definitions do not conflict, namely by checking that their identifiers are distinct. 
See https://github.com/osmosis-labs/osmosis/pull/108/commits/915c302212df967694ae5cff7692b86b8ce72a9d#diff-dcd993b301b1c9f9604e802eef8ee672aa5f8357130d260df09b849e9a99d5ffR40 

We should here enforce that the identifiers provided in genesis are all unique. 

Not really pre-launch blocking since we can manually verify this, but great if anyone gets a chance to do it

mattverse:
Issue solved with commit https://github.com/osmosis-labs/osmosis/commit/8840df7e51f30e019af4905eafca895c14ec4852.

PR #396, opened 2021-08-03
Add validation for unique epoch identifier

mattverse:
This issue is related to issue https://github.com/osmosis-labs/osmosis/issues/150.

This PR adds validation in genesis so that the identifier provided in genesis of the epoch module are all unique.
This ensures that multiple definitions do not conflict, namely checking that their identifiers are distinct.


ValarDragon (approved):
LGTM!

Issue #192, opened 2021-05-19
Make validateEpochIdentifier a function defined by x/epochs


ValarDragon:
Its currently duplicated in every module who calls it (e.g. https://github.com/osmosis-labs/osmosis/pull/189/files#diff-ab83f6f773690e81db9bb8b1adb169983c955d5782f61ba2aaff49eba48a303aR131)

ValarDragon:
So we should move that validate function into x/epochs, and search for all other usages of that function in the codebase.

mattverse:
Issue solved in PR https://github.com/osmosis-labs/osmosis/pull/398

PR #398, opened 2021-08-04
Make validateEpochIdentifier a function defined by x/epochs 

mattverse:
This PR is related to issue https://github.com/osmosis-labs/osmosis/issues/192.

This PR adds validateEpochIdentifier, a funciton for validating the epoch identifier so that it has valid epoch identifier, in the epoch module. Doing so would be necessary in the sense of scalability, as the function itself would be duplicated in every module which needs the process of validating the epoch identifier.

### DIFF
--- a/x/epochs/types/identifier.go
+++ b/x/epochs/types/identifier.go
@@ -1,0 +1,22 @@
+package types
+
+import (
+	"fmt"
+)
+
+func ValidateEpochIdentifierInterface(i interface{}) error {
+	v, ok := i.(string)
+	if !ok {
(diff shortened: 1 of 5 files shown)


ValarDragon:
LGTM! Turns out epochs are also used in incentives, can you remove this func in favor of the exported one as well: https://github.com/osmosis-labs/osmosis/blob/d3e00da98551cc77db54ea4de0f974703f298de2/x/incentives/types/params.go#L49

mattverse:
Np! Done!@!

PR #401, opened 2021-08-06
Derive balances from state export

antstalepresh:


ValarDragon:
Awesome, thanks for working on this! This is very useful.

Could a `total balance` field be added, and can `unclaimed staking reward` be added to the liquid balance? (Currently unclaimed staking reward doesn't show up at all)

ValarDragon:
Actually, it probably suffices to just have the total field for now, then we can go ahead and get this merged 💪

antstalepresh:
> Actually, it probably suffices to just have the total field for now, then we can go ahead and get this merged 💪

Added total field

ValarDragon (approved):

Issue #149, opened 2021-05-03
Fix Epoch Time Drift


sunnya97:
Reference:  https://github.com/osmosis-labs/osmosis/pull/108#discussion_r624009821

ValarDragon:
The solution were going with is:

In the first block after a timer ends, we run both the AfterEpochEnd and BeforeEpochStart logics, and these both run at the exact right start time, instead of the current block time, and in the same blocks `BeginBlock`.

This breaks the assumption that `AfterEpochEnd` and `BeforeEpochStart` happen on different blocks that currently exists. This was never a long term assumption to be making. We will need to see if we will have to fix things in incentives or minting because of this though.

PR #403, opened 2021-08-06
Fix Epoch Time Drift

mattverse:
This PR is related to issue https://github.com/osmosis-labs/osmosis/issues/149.

This PR fixes the current problem of time drift in the epoch module by modifying the logic in abci.go. With the modified logic, `CurrentEpochStartTime` is no longer set as the current time, but follows the actual epoch duration. 

Additional changes have been made to genesis.go, as the logic of initializing the `CurrentEpochStartTime` should be happening in abci.go to distinguish the initial start of the epoch.


x/epochs/abci.go
```diff
@@ -29,11 +29,13 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 			epochStart = true
 			epochInfo.CurrentEpoch = epochInfo.CurrentEpoch + 1
 		}
-
 		if epochStart {
 			epochInfo.CurrentEpochEnded = false
-			epochInfo.CurrentEpochStartTime = ctx.BlockTime()
-
+			if epochInfo.CurrentEpochStartTime.Equal(time.Time{}) { //genesis state, initialize
```
mattverse, line 34: If `CurrentEpochStartTime` initialization is done is genesis.go, there would be no way to figure out whether `CurrentEpoch` is the first, initializing epoch or not. 
Knowing whether `currentEpoch` is 0 or not within BeginBlocker is crucial in the sense that there is an exception for Epoch counting logic in the beginning. When an epoch is starting initially, `currentEpoch` is 0, while it gets incremented right away, creating an exceptional case to think in the Epoch logic. That is, at currentEpoch 0,it would create an error when trying to process with the same logic as any other currentEpoch, being needed to be processed exceptionally.

x/epochs/abci.go
```diff
@@ -29,11 +29,13 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
...
+				epochInfo.CurrentEpochStartTime = epochInfo.StartTime
+			} else {
+				epochInfo.CurrentEpochStartTime = epochInfo.CurrentEpochStartTime.Add(epochInfo.Duration)
```
mattverse, line 37: * Main Logic to fix epoch time drift

Currently, the time drift is taking place because epoch start time is set to the very time when the logic is being executed. 
To fix this problem, epoch start time should be resulted by adding duration to the previous epoch start time

x/epochs/abci_test.go
```diff
@@ -26,15 +26,15 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
 		fn                       func()
 	}{
 		{
-			expCurrentEpochStartTime: now,
+			expCurrentEpochStartTime: time.Time{},
```
mattverse, line 29: genesis.go is fixed. At currentEpoch 0, epoch start time should be in its initial, empty state.

x/epochs/genesis.go
```diff
@@ -18,11 +18,6 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 			epoch.StartTime = ctx.BlockTime()
 		}
 
-		// when epoch counting started and current_epoch_start_time is not set, set it
-		if epoch.EpochCountingStarted && epoch.CurrentEpochStartTime.Equal(time.Time{}) {
-			epoch.CurrentEpochStartTime = ctx.BlockTime()
```
mattverse, line 23: logic of initializing epoch start time is moved to abci.go

x/epochs/genesis_test.go
```diff
@@ -96,7 +96,7 @@ func TestEpochsInitGenesis(t *testing.T) {
 	require.Equal(t, epochInfo.StartTime.UTC().String(), now.UTC().String())
 	require.Equal(t, epochInfo.Duration, time.Hour*24)
 	require.Equal(t, epochInfo.CurrentEpoch, int64(0))
-	require.Equal(t, epochInfo.CurrentEpochStartTime.UTC().String(), ctx.BlockTime().UTC().String())
+	require.Equal(t, epochInfo.CurrentEpochStartTime.UTC().String(), time.Time{}.String())
```
mattverse, line 99: At genesis, epoch start time should be at an empty state as the initial value is given in BeginBlocker

x/epochs/abci_test.go
```diff
@@ -135,7 +135,7 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
 				{
 					Identifier:            "monthly",
 					StartTime:             time.Time{},
-					Duration:              time.Hour * 24,
+					Duration:              time.Hour * 24 * 31,
```
mattverse, line 138: Test code for monthly epoch. Duration should be time.Hour * 24 * 31

x/epochs/abci.go
ValarDragon, line 34:
```suggestion
			if epochInfo.CurrentEpochStartTime.Equal(time.Time{}) { // genesis state, initialize
```

x/epochs/abci.go
ValarDragon: Ok awesome. 

An idea that just occured to me is that maybe we should log that were starting a new epoch as well. (Using ctx.Logger().Info if I remember correctly)

ValarDragon: Bump on adding a ctx.logger here

ValarDragon: It only seems to be done in the uninitialized case atm?

ValarDragon:
I'm surprised theres no changes need to lockups or incentives needed here. @antstalepresh can you confirm? I thought there was some issues in the current lockup / incentive code if we had an epoch end and start on the same block

antstalepresh:
> I'm surprised theres no changes need to lockups or incentives needed here. @antstalepresh can you confirm? I thought there was some issues in the current lockup / incentive code if we had an epoch end and start on the same block

Since lockups and incentives are using hooks, I believe there won't be any issues doing it on the same block. We will need @sunnya97 's input here. At first, I think I did it with the same block but he mentioned that we will need to separate those.

x/epochs/abci.go
```diff
@@ -13,11 +13,12 @@ import (
 // BeginBlocker of epochs module
 func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
-
 	// Note: epoch length must be more than 2 blocks, as the implementation requires epoch_endblock and epoch_startblock are separate
 	// epoch_startblock(n+1) = epoch_endblock(n) + 1
```
ValarDragon, line 17: This comment should be deleted

x/epochs/abci.go
```diff
@@ -29,13 +30,16 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 			epochStart = true
 			epochInfo.CurrentEpoch = epochInfo.CurrentEpoch + 1
 		}
-
 		if epochStart {
```
ValarDragon, line 33: This looks to me like its never getting ran after the first epoch? `epochStart` only ever get set to true if `epochInfo.EpochCountingStarted() == false`. But when thats true (for active epochs), this logic should never be getting ran?

x/epochs/abci.go
```diff
@@ -65,6 +69,8 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
 		}
```
ValarDragon, line 69: tabbing issue?

x/epochs/abci.go
```diff
@@ -65,6 +69,8 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
...
 
 		k.AfterEpochEnd(ctx, epochInfo.Identifier, epochInfo.CurrentEpoch)
+		k.BeforeEpochStart(ctx, epochInfo.Identifier, epochInfo.CurrentEpoch)
```
ValarDragon, line 72: I'm confused why we have this logic in the end blocker at all. I thought we're going with only having begin blocker logic, no end blocker logic. https://github.com/osmosis-labs/osmosis/issues/149#issuecomment-889626135

(Or is there a reason to move these things to the end blocker)

x/epochs/abci.go
ValarDragon: Oh wait no, theres a tabbing / spaces issue on line 28/29. (Can golint be ran)

Ok this is still being based on `CurrentEpochEnded` -- which I don't think is right. We wanted the end of the prior epoch to get detected within begin block, and then have the prior epoch ended here, and the next one began here -- all in one block.

ValarDragon:
I think in general we should make the begin block pseudocode:
```
Iterate over every epoch {
  if epoch.not_initialized() {
   Try epoch initialization logic to see if its time to start
  }
  
  if cur_block_time >= epoch.end_time() {
    epoch_exact_end_time = epoch.start_time() + epoch.duration()
    epoch.End_Epoch_at_time(epoch_exact_end_time)
    epoch_exact_start_time = epoch_exact_end_time + time.nanosecond()
    epoch.Start_Epoch_at_time(epoch_exact_start_time)
  }
}
```

And then in EndBlock, we do nothing.

proto/osmosis/epochs/genesis.proto
```diff
@@ -27,7 +27,6 @@ message EpochInfo {
     (gogoproto.moretags) = "yaml:\"current_epoch_start_time\""
   ];
   bool epoch_counting_started = 6;
-  bool current_epoch_ended = 7;
```
ValarDragon, line 30: Can we keep this field, just to not break the proto file, and rename the variable to `deprecated_current_epoch_ended`

ValarDragon: Later we should figure out how to improve that, but I think we can punt on that state migration logic.

x/epochs/abci.go
```diff
@@ -13,67 +13,43 @@ import (
...
-				epochInfo.EpochCountingStarted = true
-				epochInfo.CurrentEpoch = 0
-			}
-		} else if epochInfo.CurrentEpochEnded { // epoch ended in last block
-			epochStart = true
-			epochInfo.CurrentEpoch = epochInfo.CurrentEpoch + 1
-		}
+		logger := k.Logger(ctx)
+
+		shouldInitialEpochStart := !epochInfo.EpochCountingStarted && !epochInfo.StartTime.After(ctx.BlockTime())
```
ValarDragon, line 19:
```suggestion
		// Has it not started, and is the block time > initial epoch start time
		shouldInitialEpochStart := !epochInfo.EpochCountingStarted && ctx.BlockTime().After(epochInfo.StartTime)
```

x/epochs/abci.go
```diff
@@ -13,67 +13,43 @@ import (
...
 
-		if epochStart {
-			epochInfo.CurrentEpochEnded = false
-			epochInfo.CurrentEpochStartTime = ctx.BlockTime()
+		epochEndTime := epochInfo.CurrentEpochStartTime.Add(epochInfo.Duration)
+		shouldEpochStart := ctx.BlockTime().After(epochEndTime) && !epochInfo.StartTime.After(ctx.BlockTime())
```
ValarDragon, line 22:
```suggestion
		shouldEpochStart := ctx.BlockTime().After(epochEndTime) && !shouldInitialEpochStart
```

x/epochs/abci.go
```diff
@@ -13,67 +13,43 @@ import (
...
 					sdk.NewAttribute(types.AttributeEpochStartTime, fmt.Sprintf("%d", epochInfo.CurrentEpochStartTime.Unix())),
 				),
 			)
+			ctx.EventManager().EmitEvent(
+				sdk.NewEvent(
+					types.EventTypeEpochEnd,
+					sdk.NewAttribute(types.AttributeEpochNumber, fmt.Sprintf("%d", epochInfo.CurrentEpoch)),
+				),
+			)
+			k.AfterEpochEnd(ctx, epochInfo.Identifier, epochInfo.CurrentEpoch)
```
ValarDragon, line 49: This should only be called in the `else` case, which is after line 33

proto/osmosis/epochs/genesis.proto
ValarDragon: Ah wait, theres a thing in protobuf for this. We can delete this field, but we need to add a line for `reserved 7`. https://developers.google.com/protocol-buffers/docs/proto3#reserved

x/epochs/abci.go
mattverse: I think it should stay as the original way because the "after" method of the time is not current time inclusive, while as it should be starting the epoch when the BeginBlocker is running at the exact same time as the epoch start time.

x/epochs/abci.go
mattverse: But yes we do need !shouldInitialEpochStart at this point as well

x/epochs/abci.go
ValarDragon: Ah thats fair, lets still add the comment though

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
 	tests := []struct {
 		expCurrentEpochStartTime time.Time
 		expCurrentEpoch          int64
-		expCurrentEpochEnded     bool
+		expEpochStartTime        time.Time
 		fn                       func()
 	}{
 		{
 			expCurrentEpochStartTime: now,
```
ValarDragon, line 29:
```suggestion
			// Only advance 2 seconds, do not increment epoch
			expCurrentEpochStartTime: now,
```

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
...
-			expCurrentEpochStartTime: now.Add(time.Second),
+			expCurrentEpochStartTime: now,
 			expCurrentEpoch:          1,
-			expCurrentEpochEnded:     false,
+			expEpochStartTime:        now,
 			fn: func() {
 				ctx = ctx.WithBlockHeight(2).WithBlockTime(now.Add(time.Second))
 				epochs.BeginBlocker(ctx, app.EpochsKeeper)
-				epochs.EndBlocker(ctx, app.EpochsKeeper)
 				epochInfo = app.EpochsKeeper.GetEpochInfo(ctx, "monthly")
```
ValarDragon, line 45: Can we delete this, this seems identical to the above test case?

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
...
 		{
-			expCurrentEpochStartTime: now.Add(time.Second),
+			expCurrentEpochStartTime: now,
 			expCurrentEpoch:          1,
-			expCurrentEpochEnded:     false,
+			expEpochStartTime:        now,
 			fn: func() {
 				ctx = ctx.WithBlockHeight(2).WithBlockTime(now.Add(time.Second))
 				epochs.BeginBlocker(ctx, app.EpochsKeeper)
-				epochs.EndBlocker(ctx, app.EpochsKeeper)
```
ValarDragon, line 53: Can we undo deleting the end blocker here, want to have the test ensure it doesn't change anything

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
...
+			expEpochStartTime:        now,
 			fn: func() {
 				ctx = ctx.WithBlockHeight(2).WithBlockTime(now.Add(time.Second))
 				epochs.BeginBlocker(ctx, app.EpochsKeeper)
 				epochInfo = app.EpochsKeeper.GetEpochInfo(ctx, "monthly")
 			},
 		},
 		{
-			expCurrentEpochStartTime: now.Add(time.Second),
+			expCurrentEpochStartTime: now,
```
ValarDragon, line 39:
```suggestion
			// Only advance 2 seconds, do not increment epoch. (Runs EndBlocker as well)
			expCurrentEpochStartTime: now,
```

x/epochs/abci_test.go
ValarDragon: Ah wait, we deleted the EndBlocker. So we should just delete this test entirely.

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
 	tests := []struct {
 		expCurrentEpochStartTime time.Time
 		expCurrentEpoch          int64
-		expCurrentEpochEnded     bool
+		expEpochStartTime        time.Time
```
ValarDragon, line 25:
```suggestion
		expInitialEpochStartTime        time.Time
```

ValarDragon: Can you do this find/replace, was quite confusing lol

x/epochs/abci_test.go
```diff
@@ -22,95 +22,78 @@ func TestEpochInfoChangesBeginEndBlockersAndInitGenesis(t *testing.T) {
...
 				ctx = ctx.WithBlockHeight(3).WithBlockTime(now.Add(time.Hour * 24 * 31))
 				epochs.BeginBlocker(ctx, app.EpochsKeeper)
 				epochInfo = app.EpochsKeeper.GetEpochInfo(ctx, "monthly")
 			},
 		},
 		{
-			expCurrentEpochStartTime: now.Add(time.Second),
-			expCurrentEpoch:          1,
-			expCurrentEpochEnded:     true,
+			expCurrentEpochStartTime: now.Add(time.Hour * 24 * 31),
```
ValarDragon, line 61:
```suggestion
			// Test that incrementing _exactly_ 1 month increments the epoch count.
			expCurrentEpochStartTime: now.Add(time.Hour * 24 * 31),
```

ValarDragon (approved):
LGTM other than my comments on the tests

x/epochs/abci_test.go
mattverse: Ah the function name might be confusing cause it still includes test endblocker. Fixed!

PR #409, opened 2021-08-12
Add tokens to lock keeper function

antstalepresh:


x/lockup/keeper/lock.go
```diff
@@ -281,6 +281,40 @@ func (k Keeper) UnlockPeriodLockByID(ctx sdk.Context, LockID uint64) (*types.Per
 	return lock, err
 }
 
+// AddTokensToLock locks more tokens into a lockup
+func (k Keeper) AddTokensToLock(ctx sdk.Context, owner sdk.AccAddress, lockID uint64, coins sdk.Coins) (*types.PeriodLock, error) {
```
ValarDragon, line 285: Can we separate this out into two functions:

```
AddTokensToLockID(ctx sdk.Context, owner sdk.AccAddress, lockID uint64, coins sdk.Coins)
``` 
and 
```
AddTokensToLock(ctx, lock *types.PeriodLock, coins sdk.Coins)
```

I think both will come in handy

x/lockup/keeper/lock.go
```diff
@@ -281,6 +281,40 @@ func (k Keeper) UnlockPeriodLockByID(ctx sdk.Context, LockID uint64) (*types.Per
 	return lock, err
 }
 
+// AddTokensToLock locks more tokens into a lockup
```
ValarDragon, line 284:
```suggestion
// AddTokensToLock locks more tokens into a lockup
// This also saves the lock to the store.
```

ValarDragon (approved):
LGTM, just some minor comments

x/lockup/keeper/lock.go
antstalepresh: Did split the function into two pieces

PR #411, opened 2021-08-12
fix parse

hydrogen18:
This applies the same patch that was previously applied to fix the `keys parse` command


ValarDragon:
Thanks!

PR #415, opened 2021-08-18
Add distribution event

antstalepresh:


ValarDragon (approved):

robert-zaremba:
Why not using proto events? Much better DevUX. You create a proto type for event and push it, instead of creating dynamic attributes.

sunnya97:
@robert-zaremba are proto events ready to use? Do you have an example code of how to use them?

robert-zaremba:
> are proto events ready to use?

Yes, although the type is set to the proto type (so it's long). 

> Do you have an example code of how to use them?

+ https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/authz/v1beta1/event.proto
+ https://github.com/cosmos/cosmos-sdk/blob/master/x/authz/keeper/keeper.go#L144

sunnya97:
@ValarDragon wdyt of switching to these ^

ValarDragon:
These are great! I had just thought that they were blocked on sdk v0.43.0

PR #332, opened 2021-06-18
Add x/incentives cli docs

mconcat:

PR #416, opened 2021-08-18
No More Hunting (seeds edition)

faddat:
- Add x/incentives cli docs (#332)
- Fix CLI bug with start-time being blank
- Fork in BeginBlock code (#395)
- Add Validator Commission upgrade logic
- Go mod tidy
- Add script to be used for testing the upgrade
- Change SDK version to v0.42.9
- Change upgrade height to 712,000 and upgrade to cosmos SDK v0.42.9
- Update testing script
- Add more instructions to testing script
- Add required beginblocker for SDK v0.42.9
- No more pass around document.  Validators add peers and seeds here.  It is then visible, versioned, and requires no thought from new users.  Genesis is next.
- Ended a pointless, harmful, centralization prone tradition.


faddat:
Basically:

* brings main current with v3.1.0
* Adds default seeds and peers.  We should do this.  Current state of affairs wastes hours of people's lives when moving towards most common setup of osmosis.  
* Next PR in the easy series embeds the genesis.json.  Starting osmosis mainnet will be done as:

```
osmosisd start
```

cmd/osmosisd/cmd/init.go
```diff
@@ -80,7 +80,10 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 			serverCtx := server.GetServerContextFromCmd(cmd)
 			config := serverCtx.Config
 
+			config.P2P.Seeds = "085f62d67bbf9c501e8ac84d4533440a1eef6c45@95.217.196.54:26656"
+			config.P2P.PersistentPeers = "2e3e3b7703a598024a2fb287587095bc4d14fe52@95.217.196.54:2000,f5be19f84deb843c18e9b612b7987138ba13ac02@5.9.106.185:2000,f9c49739f0641a0a673e7a1e8edc38054fefc840@144.76.183.180:2000,40aafcd9b6959d58dd1c567d9daf2a82a23311cf@162.55.132.230:2000,3fea02d121cb24503d5fbc53216a527257a9ab55@143.198.145.208:26656,e7916387e05acd53d1b8c0f842c13def365c7bb6@176.9.64.212:26656,785bc83577e3980545bac051de8f57a9fd82695f@194.233.164.146:26656,421e7ffd6c239cc51620e2ec2330d392a7baeeb3@207.246.109.0:26656,778fdedf6effe996f039f22901a3360bc838b52e@161.97.187.189:36656,2115945f074ddb038de5d835e287fa03e32f0628@95.217.43.85:26656,a79b9904461a98cdb98a5c15d7d75b02a55b6ab9@162.216.4.214:36656,1639d885f47c96c5cc4be6e4443ba53cdc74fc71@23.111.129.54:36656"
```
Sdpierret, line 84: peers and seed can change over time.
I think it's better to keep this blank and use external documentation.

faddat: Sir, that's the point.  They will change over time.  

They will change here, and be versioned, and there will be some sense of accountability and we won't have to pass around a semi-secret, hard to locate document.

faddat: please consider the harm done by asking 100+ people to routinely configure seeds and peers.  

This also allows us to begin working on formally tracking sync speed:

https://github.com/osmosis-labs/osmosis/pull/418

....and shouldn't we live in a world, where each validator has at least one commit on this repository?

Sdpierret: it's not something who need to be versioned.
it's a production context.

faddat: 2 things:

https://twitter.com/gadikian/status/1427959398857543687

Screen Shot 2021-08-18 at 6 51 55 PM

^ thought we had a team just shy of 100?

ValarDragon:
Thanks for adding this!

cmd/osmosisd/cmd/init.go
```diff
@@ -80,7 +80,10 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
 			config.P2P.MaxNumOutboundPeers = 40
+			config.P2P.MaxNumInboundPeers = 150
```
ValarDragon, line 86: +1

go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/osmosis-labs/osmosis
 
-go 1.15
+go 1.17
```
ValarDragon, line 3: Why change this?

cmd/osmosisd/cmd/init.go
ValarDragon, line 84: +1 to default seeds list. I think we should not be adding default persistent peers though

ValarDragon: The idea should be to make it easy to get access to nodes in the network, which is what the default seeds should be accomplishing.

Persistent peers makes you keep on trying to get a connection with them. They will get their inbound peers saturated, and then the node will get confused with logs saying that they keep getting refusals to connect. (And it will overly hammer bandwidth of these listed persistent peers)

go.mod
faddat: I changed it today because of the new release, and my desire to use `embed`.

cmd/osmosisd/cmd/init.go
faddat: hmmmmmm good point!

We will need more people to run seeds in that case.  

I'll ask in the validator channel.

faddat:
THe seed I put there is running the new tenderseed, which happens to osmo by default:

https://github.com/notional-labs/tenderseed which is of course a fork of a fork

go.mod
ValarDragon: What is `embed`? Also `go.1.17` is imo too recent, I definitely think we should be staying on 15/16 for awhile, unless quite important.

go.mod
```diff
@@ -25,8 +25,91 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require (
```
ValarDragon, line 28: Why did a second require section get added to the go.mod?

cmd/osmosisd/cmd/init.go
```diff
@@ -80,13 +80,17 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 			serverCtx := server.GetServerContextFromCmd(cmd)
 			config := serverCtx.Config
 
+			//Override settings in config.toml
+			config.P2P.Seeds = "085f62d67bbf9c501e8ac84d4533440a1eef6c45@95.217.196.54:26656"
```
ValarDragon, line 84: After talking to folks on Interchain Gmbh & Informal teams, I'm a bit worried about adding a default list of seeds for the following two reasons:

1) If you connect to a seed node, that itself doesn't consider itself to be a seed node, then you will remain connected to that node as a peer (not disconnect). So suppose we have N seeds listed, and `k` were malicious. As long as one of the seeds isn't malicious, I will get peers honestly on the network. However, in the current Tendermint, if those seeds are maliciously misconfigured to not disconnect from you, then they will fill up all your node connections. Then you won't sample any of the other peers you got from the honest seeds, and will just be connected to those `k` malicious seeds, thereby ruining your security.

2) Having a default seed list is great, as long as Tendermint itself is making random sampling guarantees. E.g. if I list 100 seeds, it should do one of:
 -  Get peers from each seed, build a large list, then randomly sample from the large list. 
 - Randomly sample which seeds to query, then build a sufficiently sized peer list. Then randomly sample peers of the local address book to go ping.

ValarDragon: I view (1) as a bug that needs to be fixed before I feel comfortable with this. 

I view (2) as something that I need to look into to double check, and if whats implemented is one of those scenarios, then lets go with this!

ValarDragon: Granted, there is an argument to be made that anything done with many validators putting things into this repo is much better than the currently mutable-by-anyone google sheets & hackmd's get passed around.

Thoughts @sunnya97 @UnityChaos ?

ValarDragon: Actually, I think (1) is fine by ensuring that we only seed nodes who we know the team for. This is similar to what bitcoin does: https://en.bitcoin.it/wiki/Satoshi_Client_Node_Discovery (Only 6 core devs curate lists) Thanks @sunnya97 for bringing this up.

We should still get that bug fixed tho.

(2) is something we should ensure, but less concerning.

ValarDragon (commented):
LGTM except for go mod situation. 

For the seed nodes, lets add some comments explaining who are the default seed nodes. Perhaps this can be done by defining a list of seed strings, each with a comment after it saying whose it is. And then combining this list to get the TM format

go.mod
faddat: go 1.17 does this.  First I'd seen it, too.

cmd/osmosisd/cmd/init.go
faddat: Yeah, so that is exactly what I am thinking:

we should know the operators.

faddat:
@ValarDragon go mod situation is just normal go 1.17, afaik.

https://golang.org/doc/go1.17#go-command


```
If a module specifies go 1.17 or higher, the module graph includes only the immediate dependencies of other go 1.17 modules, not their full transitive dependencies. (See Module graph pruning for more detail.)

For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.) 
```

In general, my thinking is that we should aggressively keep latest versions for go, etc.

wdyt?

faddat:
* go 1.16
* No persistent peers
* 150 inbound peers
* Seeds as a slice with a comment saying who is what by each seed

cmd/osmosisd/cmd/init.go
```diff
@@ -80,13 +82,31 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+			seeds := []string {
+			"085f62d67bbf9c501e8ac84d4533440a1eef6c45@95.217.196.54:26656"} // Notional
+		 
+			
+			
+
+			
+
+			//Override default settings in config.toml
+			config.LogLevel = "error"
```
ValarDragon, line 98: I think the default should be `info`?

faddat: oop, thank you!  I had been testing something.

ValarDragon (approved):
This PR is good to merge imo, except for the default log level being changed to error. 

Unless theres a reason you think this is a better default, in which case I'd love to discuss! I have a preference for discussing in a second PR tho, so we can unblock this one / get it merged in. 

I currently feel like the info level logging of consensus, where you see blocks fly by, is quite helpful.

cmd/osmosisd/cmd/init.go
UnityChaos: Yeah, I think well known validator operators making PRs to add their seed nodes would be fine. :+1:

ValarDragon:
I'll fix the go mod issue, just made a branch to fix it last night, and hadn't PR'd it to main yet.

faddat:
ooookay.  I may reopen this anew.  "works on my laptop not good enough."

ValarDragon:
Accidentally squash merged, fixing the commit history in #421

ValarDragon:
Fixed! Ok now the code changes should be trackable

PR #417, opened 2021-08-18
Create build.yml

faddat:
builds it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         include:
           - targetos: windows
             arch: amd64
-    name: dig ${{ matrix.arch }} for ${{ matrix.targetos }}
+    name: osmosis ${{ matrix.arch }} for ${{ matrix.targetos }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2.1.3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+        targetos: [darwin, linux]
+        include:
+          - targetos: windows
+            arch: amd64
+    name: dig ${{ matrix.arch }} for ${{ matrix.targetos }}
```
ValarDragon, line 21: Why is it called dig? I thought dig is a DNS utility

ValarDragon:
Thanks for adding this, this is great! Just one question on the name of `dig`, other than that LGTM

.github/workflows/build.yml
faddat: oops!

That is taken from another chain I am working on called dig, and yes, its binary gets confused with the dns tool often. 

Thanks!

ValarDragon: ah lol, makes sense now! Name collisions are the worst

ValarDragon (approved):

PR #418, opened 2021-08-18
Blocks per Second

faddat:
we will use this to figure out our blocks per second.

Then we will drill down, down, down, till we have some idea of what helps us improve.


faddat:
This needs some more work, I'll play with er some.

faddat:
On second thought, I may just try and code this in go, into osmosis.  

eg:

whatblockami at (wait ten seconds) whatblockamiat divide by ten; log bps & peer count

faddat:
closing in favor of just getting bps into osmosis prime.

PR #422, opened 2021-08-19
Gamm TWAP

mattverse:
This PR introduces the concept of TWAP(time weighted average price) to Gamm module storing cumulative data on-chain.

Added main logic for storing twap data, relevant test codes and query.  Additional test cases and CLI WIP, to be added ASAP.

Nice changes or additional functions for twap to have(not necessary for merge but would be nice) are 
- Use transient KV Store for more code efficiency. 
- Sort poolTwap as of Pool
- Delete poolTwap History that has passed a certain amount of time


proto/osmosis/gamm/v1beta1/pool.proto
```diff
@@ -150,3 +150,26 @@ message Pool {
...
+  ];
+}
+
+message PoolTwapHistory {
+  google.protobuf.Timestamp timeStamp = 1 [
+    (gogoproto.stdtime) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"time_stamp\""
+  ];
+  uint64 poolId = 2;
```
ValarDragon, line 173: Why is poolID in the struct? Should the PoolTwapHistory instead be indexed by `poolID` in state?

proto/osmosis/gamm/v1beta1/pool.proto
```diff
@@ -150,3 +150,26 @@ message Pool {
...
+  repeated TwapPair twapPairs = 3;
```
ValarDragon, line 174: How are TwapPairs going to be sequenced, or will it be that everything that wants to look at a given token-pair will iterate over all pairs to find it?

ValarDragon: Also looking at this makes me think we should lower the max number of assets per pool from 8 to 5. 8 assets -> 28 pairs, 5 assets -> 10 pairs.

proto/osmosis/gamm/v1beta1/pool.proto
```diff
@@ -150,3 +150,26 @@ message Pool {
...
+    (gogoproto.nullable) = false
+  ];
+  string spotPrice = 4 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message PoolTwapHistory {
+  google.protobuf.Timestamp timeStamp = 1 [
```
ValarDragon, line 168:
```suggestion
  google.protobuf.Timestamp last_updated_at = 1 [
```

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -172,3 +178,23 @@ message QueryTotalLiquidityResponse {
     (gogoproto.nullable) = false
   ];
 }
+
+//=============================== PoolTwap
+message QueryPoolTwapRequest {
+  uint64 poolId = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string tokenInDenom = 2 [ (gogoproto.moretags) = "yaml:\"token_in_denom\"" ];
+  string tokenOutDenom = 3
+      [ (gogoproto.moretags) = "yaml:\"token_out_denom\"" ];
```
ValarDragon, line 187: Should these be optional, OR should there be a query to get all twap responses?

x/gamm/client/cli/cli_test.go
```diff
@@ -866,6 +866,7 @@ func (s *IntegrationTestSuite) TestGetCmdPool() {
 
 				resp := types.QueryPoolResponse{}
 				s.Require().NoError(clientCtx.JSONMarshaler.UnmarshalJSON(out.Bytes(), &resp), out.String())
+				s.Require().Equal(resp.Pool, "a")
```
ValarDragon, line 869: debug code?

x/gamm/keeper/pool_service.go
```diff
@@ -232,6 +236,11 @@ func (k Keeper) JoinSwapExternAmountIn(
 	k.hooks.AfterJoinPool(ctx, sender, pool.GetId(), addedCoins, shareOutAmount)
 	k.RecordTotalLiquidityIncrease(ctx, addedCoins)
 
+	err = k.UpdatePoolTwap(ctx, poolId, tokenIn.Denom)
```
ValarDragon, line 239: Can we make a helper function in keeper.go called `GetPoolAndPrepareForSwap`, and in that function run `k.GetPool()`, and `k.UpdatePoolTwap()`?

That will save us some code duplication within all these methods

x/gamm/keeper/swap.go
```diff
@@ -60,6 +60,11 @@ func (k Keeper) SwapExactAmountIn(
 		return sdk.Int{}, sdk.Dec{}, err
 	}
 
+	err = k.UpdatePoolTwap(ctx, poolId, tokenIn.Denom, tokenOut.Denom)
```
ValarDragon, line 63: Lets add this into `k.updatePoolForSwap` instead

x/gamm/keeper/pool_service.go
ValarDragon: Err, it should be called `GetPoolAndPrepareForJoin` I think

x/gamm/keeper/pool_twap.go
```diff
@@ -0,0 +1,234 @@
...
+	store := ctx.KVStore(k.storeKey)
+	poolTwapKey := k.GetPoolTwapKey(ctx, poolId, queryTime)
+
+	// if twap have not existed before, create new pool twap
+	if len(poolTwapKey) == 0 {
+		poolTwap, err := k.newPoolTwapHistory(ctx, poolId)
+		if err != nil {
+			return types.PoolTwapHistory{}, err
+		}
+		k.SetPoolTwap(ctx, poolTwap)
```
ValarDragon, line 22: Feels odd for `GetPoolTwapHistory` to get to mutate state here

x/gamm/keeper/pool_twap.go
```diff
@@ -0,0 +1,234 @@
...
+	k.SetPoolTwap(ctx, poolTwap)
+	return nil
+}
+
+func (k Keeper) GetPoolTwapKey(ctx sdk.Context, poolId uint64, queryTime time.Time) []byte {
+	store := ctx.KVStore(k.storeKey)
+	timestamp := queryTime.Unix()
+
+	iteratorStart := types.GetKeyPoolTwaps(poolId, 0)
+	iteratorEnd := types.GetKeyPoolTwaps(poolId, timestamp)
```
ValarDragon, line 142: This iterator length is bounded, right?

x/gamm/keeper/pool_twap.go
```diff
@@ -0,0 +1,234 @@
...
+		TwapPairs: recentPoolTwap.TwapPairs,
+	}
+
+	k.SetPoolTwap(ctx, poolTwap)
+	return nil
+}
+
+func (k Keeper) GetPoolTwapKey(ctx sdk.Context, poolId uint64, queryTime time.Time) []byte {
+	store := ctx.KVStore(k.storeKey)
+	timestamp := queryTime.Unix()
```
ValarDragon, line 139: (Note to Self): TODO, double check this is what we normally do for converting timestamps to store keys across the SDK

proto/osmosis/gamm/v1beta1/pool.proto
mattverse: How are TwapPairs going to be sequenced, or will it be that everything that wants to look at a given token-pair will iterate over all pairs to find it?- Yes basically it would hold info on all the possible twap pairs, and if changes are necessary for a specific pair, it would iterate over all the pairs. Do you think I could use a better algorithm / structure?

proto/osmosis/gamm/v1beta1/query.proto
mattverse: hmmm...What do you refer to by all twap responses?

x/gamm/client/cli/cli_test.go
mattverse: oops my bad! Thanks for catching it out

proto/osmosis/gamm/v1beta1/pool.proto
ValarDragon: That seems fine to me for now, it really depends on how many assets we expect to ever get per pool. If we only want like 8 assets max, then its not actually an issue to iterate over everything. If we want like 20 assets, then we need to rethink it (but we can also punt rethinking it to later)

cc @UnityChaos

proto/osmosis/gamm/v1beta1/query.proto
ValarDragon: Like I want to get all the TWAP prices for a pool

ValarDragon: TODO: Write in a second PR later

x/gamm/keeper/pool_twap.go
mattverse: TBD in different PR

proto/osmosis/gamm/v1beta1/pool.proto
UnityChaos: hmm I think it's fine for now, non-binary pools aren't getting much use yet, especially up to the 8 limit. We can revisit / optimize if that does become a concern tho.

Also for vaults I don't think we would be calculating TWAP on the vault itself, so no concern there.

PR #424, opened 2021-08-22
Notional Port Standard

faddat:
In developing our relaying system, we found it best to run all of the networks on a single large machine.  To do this we made a bash profile, then systemd units, that configured the chains with environment variables.  

But wouldn't it be nice if there were some kind of standard so we don't get crazy conflicts?

I thought so, and here it is. ﻿


faddat:
I need to take version specific stuff out of dockerfile.

faddat:
I believe that the docker test will pass if this is merged. 

https://github.com/notional-labs/notional/issues/7

faddat:
You can see the port concept at:  

https://github.com/notional-labs/notional

ValarDragon:
I don't think we should be changing default ports. Using the same default ports across all the cosmos binaries is a big UX win.

If relayers want to run all the nodes on the same machine, that feels like a complexity they should be handling.

faddat:
As a developer, I have many cosmos chain binaries on my laptop.  

It would be nice if they wouldn't conflict with one another.

faddat:
But, no worries.  Want the dockerfile & auto-docker-build, or nah?

ValarDragon:
Dockerfile and auto-build would be great! Tysm

ValarDragon:
I'm also fully down to find a way to make ports easier to config across many different nodes. Though maybe that makes most sense as a second script that takes in all the config directories, and assigns accordingly?

PR #425, opened 2021-08-22
bech32 forwarding

sunnya97:


jackzampolin:
This is cool, but requires chain state and a bunch of changes. What about just changing the packet handling logic in `transfer` to handle a format in `packet.Receiver` like this:

https://github.com/strangelove-ventures/ibc-go/pull/1/files

sunnya97:
@jackzampolin that still requires the creation of an IBC transfer msg though, right?

sunnya97:
@jackzampolin That looks like that's for auto-forwarding of packets?

This PR is doing something a bit different. This is to auto convert bank/send messages into ICS20 transfers according to bech32 prefixes

jackzampolin:
@sunnya97 actually no. This is in the packet handling logic. So when user sends `MsgTransfer` on Terra bound for osmosis via the hub receiver field would look like:

```
cosmos1vzxkv3lxccnttr9rs0002s93sgw72h7ghukuhs|transfer/channel-141:osmo1vzxkv3lxccnttr9rs0002s93sgw72h7gl89vpz
```

Hub transfer module handles packet, pulls the timeout address off the front, and instead of depositing the coins in that account, it will create an IBC packet outgoing to `transfer/channel-141` (bound for osmosis) with the coins instead. When the packet is headed to osmosis, hub rewrites the receiver field as follows:

```
osmo1vzxkv3lxccnttr9rs0002s93sgw72h7gl89vpz
```

Then it becomes a normal IBC transfer.

x/bech32ibc/types/types.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import (
+	fmt "fmt"
+	"strings"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
```
sunnya97, line 9: Add comment saying this was copied from bech32 library
(find link to code)

x/bech32ics20/keeper/msg_server.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	nativePrefix, err := k.hrpToChannelMapper.GetNativeHrp(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if prefix == nativePrefix {
+
```
sunnya97, line 53: Have this just use the bankKeeper's function (i.e. ` k.bk.Send`)

x/bech32ics20/keeper/msg_server.go
```diff
@@ -0,0 +1,149 @@
...
+	)
+
+	_, err = k.ics20TransferMsgServer.Transfer(sdk.WrapSDKContext(ctx), ibcTransferMsg)
+
+	return &types.MsgSendResponse{}, err
+}
+
+func (k msgServer) MultiSend(goCtx context.Context, msg *types.MsgMultiSend) (*types.MsgMultiSendResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
```
sunnya97, line 118: Just use k.bk.MultiSeng

x/bech32ics20/keeper/msg_server.go
```diff
@@ -0,0 +1,149 @@
...
+	if msg.Amount.Len() == 0 {
+		return nil, sdkerrors.Wrap(types.ErrNoInputs, "invalid send amount")
+	}
+	if msg.Amount.Len() > 1 {
+		return nil, sdkerrors.Wrap(ibctransfertypes.ErrInvalidAmount, "cannot send multiple denoms via IBC")
+	}
+
+	ibcTransferMsg := ibctransfertypes.NewMsgTransfer(
+		ibctransfertypes.DefaultGenesisState().GetPortId(),
```
sunnya97, line 103: We need to read the actual port id being used by the transfer module

x/bech32ibc/genesis.go
```diff
@@ -0,0 +1,26 @@
...
+
+// InitGenesis initializes the capability module's state from a provided genesis
+// state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
+    // this line is used by starport scaffolding # genesis/module/init
+
+    // this line is used by starport scaffolding # ibc/genesis/init
+}
+
+// ExportGenesis returns the capability module's exported genesis.
```
sunnya97, line 17: TODO

x/bech32ics20/keeper/msg_server.go
antstalepresh: Is it secure to use other module's MsgServer on a module? there's no function called k.bk.Send but only msgServer.Send is available.

x/bech32ibc/types/types.go
antstalepresh: Done

x/bech32ics20/keeper/msg_server.go
antstalepresh: MultiSend is only available on msgServer - is it secure or normal to use other module's msgServer?

x/bech32ics20/keeper/msg_server.go
antstalepresh: Modified to use transferkeeper.

x/bech32ics20/keeper/msg_server.go
```diff
@@ -0,0 +1,149 @@
...
+
+// NewMsgServerImpl returns an implementation of the bank MsgServer interface
+// for the provided Keeper.
+func NewMsgServerImpl(keeper Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+var _ types.MsgServer = msgServer{}
+
+func (k msgServer) Send(goCtx context.Context, msg *types.MsgSend) (*types.MsgSendResponse, error) {
```
ValarDragon, line 30: We should also make all this logic into a method thats a drop-in replacement for bankkeeper.SendCoins

x/bech32ics20/keeper/msg_server.go
ValarDragon: Yeah, its safe to use the other modules MsgServer, though we should probably be using the other modules keeper directly, since thats what most of the SDK does.

ValarDragon:
This repo is being worked on in github.com/osmosis-labs/bech32-ibc

PR #428, opened 2021-08-22
Osmosis is not gaia

faddat:
Gaia -> Osmosis﻿


ValarDragon (approved):
Thanks for catching and fixing these!

PR #432, opened 2021-08-23
superfluid module 

antstalepresh:


proto/osmosis/superfluid/gov.proto
```diff
@@ -0,0 +1,18 @@
...
+// SetSuperfluidAssetsProposal is a gov Content type to update the superfluid assets
+message SetSuperfluidAssetsProposal {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_getters) = false;
+  option (gogoproto.goproto_stringer) = false;
+
+  string title = 1;
+  string description = 2;
+  repeated SuperfluidAsset assets = 3 [ (gogoproto.nullable) = false ];
+}
```
ValarDragon, line 18: This looks good to me, a bit unfortunate that theres nothing it can just "inherit" to be a base governance proposal

proto/osmosis/superfluid/superfluid.proto
```diff
@@ -0,0 +1,26 @@
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  bool enabled = 3;
+  string risk_adjust_ratio = 4 [
+    (gogoproto.moretags) = "yaml:\"risk_adjust_ratio\"",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
```
ValarDragon, line 26: We need two different structs here imo.

We need an enum describing superfluid assets, then a `SuperfluidAssetType` which specifies `denom, enum`.

I think superfluid_staked_amount, enabled, and risk_adjust_ratio shouldn't be in this struct.

`enabled` should be by virtue of being in the enabled superfluid asset list.

Risk adjust ratio will get computed by the state machine somehow. I think we will need to be storing historical records of the risk adjust ratio there, which is why I think it shouldn't go in this struct.
(Also it should be called `risk_adjusted_osmo_equivalent` I think)

antstalepresh: Modified a little based on your comments here. https://github.com/osmosis-labs/osmosis/pull/432/commits/18ef459a61da4795f22b834a8c5a50ce6cb2a8b0

antstalepresh: If we are good with current protobuf, I will start implementation of types and keepers.

proto/osmosis/superfluid/superfluid.proto
```diff
@@ -0,0 +1,37 @@
...
+// SuperfluidAsset stores the pair of superfluid asset type and denom pair
+message SuperfluidAsset {
+  SuperfluidAssetType asset_type = 1;
+  string denom = 2;
+}
+
+// SuperfluidAssetInfo stores the information of superfluid asset - real time changes
+message SuperfluidAssetInfo {
+  uint64 superfluid_asset_id = 1;
+  string superfluid_staked_amount = 2 [
```
ValarDragon, line 27:
```suggestion
  string total_staked_amount = 2 [ 
```

proto/osmosis/superfluid/gov.proto
```diff
@@ -0,0 +1,40 @@
...
+
+// EnableSuperfluidAssetsProposal is a gov Content type to enable the superfluid assets by id
+message EnableSuperfluidAssetsProposal {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_getters) = false;
+  option (gogoproto.goproto_stringer) = false;
+
+  string title = 1;
+  string description = 2;
+  repeated uint64 superfluid_asset_ids = 3;
```
ValarDragon, line 28:
```suggestion
  repeated SuperfluidAsset superfluid_assets = 3;
```

daniel-farina:
@ValarDragon This ios ready for review.

PR #434, opened 2021-08-23
Sync test

faddat:
This will check to ensure that osmosis still syncs after changes are made to it. 

It will also allow us to begin to get a grip on speed.


faddat:
This isn't as useful as I'd hoped, because we have upgraded the binary.  

Would be better to download state and then sync to tip but Github Actions won't let that work.

PR #435, opened 2021-08-23
Tendermint 0.34.12

faddat:


ValarDragon:
Wow, I'm surprised this doesn't cause issues with the SDK having a different tendermint version. But you added tests that show that this builds correctly, so should work!

ValarDragon (approved):

faddat:
Seems to also fix state sync.

ValarDragon:
Actually I'm really confused why this isn't using a `replace` directive. From what I can gather, the SDK is still going to be using the older tendermint version (https://stackoverflow.com/questions/57952397/how-to-resolve-conflicting-go-module-dependencies-when-a-top-level-module-and-on ) 

Can this be changed to add a replace line, I don't want to risk having complex bugs with the osmosis repo and SDK repo's TM versions being out of sync.

faddat:
Sure, we can do that.

faddat:
all set.

ValarDragon (approved):
Thanks!

PR #345, opened 2021-06-24
Accumulation store gas optimization and refactoring

mconcat:

Issue #207, opened 2021-05-29
Prefix Sum Tree: rename node and child structs


ValarDragon:
Currently I think the prefix sum tree types are a bit confusing. `node` is actually a pointer to a node in the tree, and `child` is the actual node.

I suggest we rename `node` to `nodePtr`, and `child` to `node`. 

What do you think @mconcat

mconcat:
Absolutely agreed

jblbit:
@ValarDragon @mconcat I was trying to find some starter tickets to work but, but looks like this one was already taken care in https://github.com/osmosis-labs/osmosis/commit/4bb6f342b76a5b539d2d4ab55dbcd534cfc1107f ?

Shall we close it to avoid other people spending more time on this?

mconcat:
Sounds good!

ValarDragon:
Oh I didn't realize this was done lol

ValarDragon:
Wait no, this isn't done as far as I can tell?

mariaxu1019:
I see a to do for improving the diagram in the spec, is this the only improvement that could be made? Is there anything else that could be helpful for improving the spec?

PR #437, opened 2021-08-24
Convert accumulation store logic to be using protobuf, and rename struct names

ValarDragon:
This PR was written by @mconcat in #345. I split it up to enable ease of review.

This PR renames the struct names, per #207 , and introduces protobuf fields.

Because the upgrade logic of the next upgrade will delete the existing lockup store, it should be safe to merge this without migration code. We just need to take care to make a `deleteOldStore` function. (However, Joon has written migration code that we can merge in as a follow-on PR!)


ValarDragon:
Merging this in, since its from Joon's ready for review PR, and I've reviewed it

PR #439, opened 2021-08-24
Script to generate LP incentive payouts from export (prop 12)

UnityChaos:


ValarDragon:
> So I assume I shouldn't commit the 50mb export, but do we have more specific instructions somewhere so others can replicate the export and check the result we compute?

We can put it in the networks repo for folks to check against

> Also, is this section where I should be doing the payments?

I think it is, and that the proposed way is the best way

UnityChaos:
So, I think the code is correct, but should I write an upgrade test for this?
like:
https://github.com/osmosis-labs/osmosis/blob/4cc28157ef4aa199eb3af35911d16cea1a356495/x/lockup/keeper/upgrade_test.go#L19
 and if so, where should it live?

Or should we just test this when we test the overall release? (ie mint some OSMO for the community pool, run the upgrade and see that the community pool amount is reduced correctly and the accounts got paid out correctly)

ValarDragon:
I think it'd be good to write a test for this. I think it could for now live as an `upgrade_test.go` file in the app folder, and later we refactor into a folder for each upgrade.

app/upgrade_test.go
```diff
@@ -0,0 +1,114 @@
...
+					coin := sdk.NewInt64Coin("uosmo", amount)
+
+					accBal := suite.app.BankKeeper.GetBalance(suite.ctx, addr, "uosmo")
+					suite.Require().Equal(coin, accBal)
+
+					total += amount
+				}
+
+				//check that the total paid out was as expected
+				suite.Require().Equal(total, int64(367926557424))
```
ValarDragon, line 86: Can we derive the expected total based on daily issuance, and hardcoded ratios?

ValarDragon (approved):
LGTM!

app/upgrade_test.go
UnityChaos: Hmm... yes, tho now that I'm doing that math I'm realizing the number there is off by a few hundred uosmo. (actual amount should be: 367926558129)
The correct calculation is:

(300,000,000,000 * 95000 * 45) / (100  * 365 * 95500)

(300mil osmo per year, 45% to LPs, 95000/95500 of that going to pools with 500/95500 going to community pool)

The constant I currently have there is the sum of the integer uosmo payment amounts output from the script. But it doesn't exactly match the theoretically correct number, because that would require including the sub uosmo amounts each account should be getting in the sum, and rounding to an integer after summing...

So I could change this to use the formula, but then the test will fail because there's... 705 uosmo, that aren't being paid out.

I suppose I could have the script output floats, and then the upgrade code would round them down to ints for the actual payments, but could sum up the floats for checking the total?

ValarDragon: Hrmm, can we have the python code round up, and then just check here that the amount overpaid is less than 1 osmo?

PR #441, opened 2021-08-24
Make the upgrade logic delete the old accumulation stores

ValarDragon:
This makes the `v4` upgrade logic delete the old accumulation store (so they then get created correctly during the upgrade), and deletes the min commission update logic that got included in the v3 fork.


mattverse (commented):
LGTM!

ValarDragon:
Hrmm, not sure why your approval didn't count as a green check, will investigate

PR #442, opened 2021-08-24
Make our test for understanding token locking gas usage give more detail

ValarDragon:
Increasing the number of locks we use to get stats for here is blocked on CacheKV store things, which I'm hoping to get tomorrow


mattverse (commented):
Looks good to merge!

PR #444, opened 2021-08-24
Statesync

faddat:
Adds a script and a test for state sync.  

This is currently failing, but in the future will give us a known-good state sync configuration.  

Request help with a fix, I am a bit puzzled.

### DIFF
--- a/.github/workflows/statesync.yml
+++ b/.github/workflows/statesync.yml
@@ -1,6 +1,6 @@
 # This is a basic workflow that is manually triggered
 
-name: compile osmosis
+name: State Sync
 
 # Controls when the action will run. Workflow runs when manually triggered using the UI
 # or API.
--- a/.github/workflows/statesync.yml
+++ b/.github/workflows/statesync.yml
@@ -1,0 +1,31 @@
+# This is a basic workflow that is manually triggered
+
+name: compile osmosis
+
+# Controls when the action will run. Workflow runs when manually triggered using the UI
+# or API.
+on: [push, pull_request, workflow_dispatch]
+
+# This workflow makes x86_64 binaries for mac, windows, and linux.
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: osmosis state sync
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2.1.3
+        with: 
+          go-version: '^1.17'
+
+      - name: Compile osmosis
+        run: ./...
+              
+      - name: init
+        run: |
+          osmosisd init test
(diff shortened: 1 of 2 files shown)


scripts/statesync.sh
```diff
@@ -0,0 +1,37 @@
...
+set -e
+
+export GOPATH=~/go
+export PATH=$PATH:~/go/bin
+go install ./...
+
+
+# MAKE HOME FOLDER AND GET GENESIS
+osmosisd init test 
+wget -O ~/.osmosisd/config/genesis.json https://cloudflare-ipfs.com/ipfs/QmXRvBT3hgoXwwPqbK6a2sXUuArGM8wPyo1ybskyyUwUxs
```
ValarDragon, line 13: Is this trying to state sync against mainnet?

faddat: Yep

faddat:
This now runs in 4 minutes and 17 seconds.

If we do a v3.10.1, And put tendermint 34.12 In it, that will allow people to smoothly state sync using this script.

Just like the peers in the init file, this will need more participation from the validator community to ensure that it is nice and decentralized but for now I can run a bootstrap node that's no problem.

scripts/statesync.sh
faddat: The way that I'm looking at this is we can use this as an indicator of changes that would cause differences in the app hash, and of changes that might break state sync itself.

faddat: It will currently fail in the master branch but we can tag it with that one setting that causes it to continue on error.

ValarDragon:
Awesome, glad that this is landing!

I don't super understand how we envision this test getting ran, should we have it be like a daily build against a branch?

`main` collects breaking changes for the next versions state machine

ValarDragon:
I'm happy to make the CI part `workflow_dispatch` only, get a daily build going there, and then add this script in.

I'll do the Tendermint change you need in the `v3.2.0` branch, which has accumulated a bunch of the minor improvemetns from `main` and epoch speedups.

.github/workflows/statesync.yml
```diff
@@ -0,0 +1,19 @@
+# This is a basic workflow that is manually triggered
+
+name: State Sync
+
+# Controls when the action will run. Workflow runs when manually triggered using the UI
+# or API.
+on: [push, pull_request, workflow_dispatch]
```
ValarDragon, line 7:
```suggestion
on: [workflow_dispatch]
```

ValarDragon (changes requested):
Once we change this. I think its good to merge

PR #445, opened 2021-08-24
Improve snapshot settings

faddat:
These new settings would make our defaults more friendly to state sync.


faddat:
It's best to close this because otherwise, it gets very confused. Now that we have many nodes with the 1500 setting it is best to leave it at 1500.

ValarDragon:
Sorry I didn't reply here, my bad. 

I think making it more infrequent was intentional to help spare the hard disks, and in anticipation of us reducing the block time in the future.

faddat:
Oh no worries. End result we should leave it at 1500 otherwise well it acts kind of drunk otherwise.

PR #446, opened 2021-08-25
WIP spec for F1 Fee architecture

ValarDragon:
WIP spec for the F1 fee distribution logic.

All the details described are AFAIK correct / quite efficient, I just think much better descriptions of them are required


ValarDragon:
Going to go ahead and merge this, and then correct it with the turnpike PR

PR #451, opened 2021-08-28
Switch to SDK branch that lowers epoch time

ValarDragon:
This switches us to the SDK branch that fixes the N^2 issues within the CacheKV store (https://github.com/osmosis-labs/cosmos-sdk/pull/24 )

It also adds some scaffolding so that we can test with LevelDB backends, which at some point ended up being necessary due to a memDB bug and makes benchmarks more accurate due to avoiding some weird oddities with the memDB.

(I'm not sure if the memDB bug still gets triggered if we make some insanely high loads, but it definitely seemed like a bug on google's sides, as I traced it back when it applied to a perfectly normal set operation that caused google's BTree to infinitely hang)

Before:
```
BenchmarkDistributionLogicMedium-16            4         270370131 ns/op        96074306 B/op    1334220 allocs/op
```
After:
```
BenchmarkDistributionLogicMedium-16          285           3644674 ns/op         1192226 B/op      20120 allocs/op
```

Speedup:
```
>>> old = 270370131
>>> new = 3644674
>>> 100 * (old - new) / old
98.6519686969416
```
~99% speedup of the benchmark!

We should backport this to v3.2.0, and check if this empirically improves the epoch time performance moreso than prior WIP v3.2.0 branch.

Also interestingly, now bech32 encoding / decoding work takes 10% of the compute time after this change.

### DIFF
--- a/x/lockup/keeper/lock_test.go
+++ b/x/lockup/keeper/lock_test.go
@@ -242,10 +242,12 @@ func (suite *KeeperTestSuite) TestLocksLongerThanDurationDenom() {
 }
 
 func (suite *KeeperTestSuite) TestLockTokensAlot() {
+	suite.SetupTest()
+
 	addr1 := sdk.AccAddress([]byte("addr1---------------"))
 	coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
 	startAveragingAt := 1000
-	totalNumLocks := 2000
+	totalNumLocks := 5000
 	for i := 1; i < startAveragingAt; i++ {
 		suite.LockTokens(addr1, coins, time.Second)
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -198,3 +198,4 @@ $RECYCLE.BIN/
 /artifacts/
 /.vscode/
 /scripts/local/
+/x/incentives/keeper/osmosis_testing/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,15 +34,21 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 # Changelog
 
-## [v2.0.0](https://github.com/osmosis/osmosis-labs/releases/tag/v2.0.0) - 2021-06-28
+## [Upcoming]
 
-* Update the cosmos-sdk version we modify to v0.42.6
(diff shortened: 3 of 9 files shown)


x/lockup/keeper/lock_test.go
```diff
@@ -242,10 +242,12 @@ func (suite *KeeperTestSuite) TestLocksLongerThanDurationDenom() {
 }
 
 func (suite *KeeperTestSuite) TestLockTokensAlot() {
+	suite.SetupTest()
+
 	addr1 := sdk.AccAddress([]byte("addr1---------------"))
 	coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
 	startAveragingAt := 1000
-	totalNumLocks := 2000
+	totalNumLocks := 5000
```
ValarDragon, line 250: This now works just fine with 20000, that'd take 7 seconds on my laptop. Just kept it lower to not take more time in CI.

ValarDragon:
Hrmm, on mainnet this still has execution time at 2 minutes total

ValarDragon:
Ok, benchmark accuracy improved! It still was that the bulk of the time before was in the CacheKV effects, but the prior benchmark wasn't well capturing the other things.

Now the time that is split around: 30% is in bech32 encoding / decoding, 10-15% in emitting events, 8% in validating denoms, and a lot in many random places in bank. I'm going to work on knocking out a few of these simple fixes at least, that I'm hoping should be a quick additional 20-30% speedup.

ValarDragon:
The above numbers have been improved (with only non-breaking changes) to be:
```
BenchmarkDistributionLogicLarge-16             1        88318071639 ns/op       34189730880 B/op        601589333 allocs/op
```

The next step non-state breaking big steps are basically:
1) Remove events from the Bank module that are triggered here - Should save massive amounts of computation time, as we will be allocating less data and garbage collecting less data.
2) Lower the number of calls to SubtractCoins by making a single SendCoinsFromModuleToManyAccounts
3) Lower bech32 overhead

Theres a lot of state breaking options to do afterwards, and changes to the logic code to combine processing of all gauges for denom at once. 

I think the main immediate thing worth doing is (1), and saving the rest for later.

PR #457, opened 2021-09-03
Remove Twap History

mattverse:
This PR Adds removing old twap history that would be saved in the state machine.

The following are set as Params of the gamm module for twap history
- The number of twap pool it tends go over and delete per iteration
- Twap history Deletion Interval(Per how many blocks would it be activating deletion iteration)
- How long would twap history be kept


faddat:
<3 paramaterization

-- question-- in the case of an archive node, the "deletion" does what exactly?

🙏 

(I assume that it simply remains, but, guess I am checking assumptions here)

ValarDragon:
This should only be deleting things from state. Other applications on-chain / integrators are expected to store the start time of TWAPs if they want TWAPs longer than the last day, or last 24 hrs. 

Not sure why theres merge conflicts though, maybe this needs to be rebased on top of `twap` branch?

app/app.go
```diff
@@ -323,7 +323,9 @@ func NewOsmosisApp(
 			app.DistrKeeper.SetParams(ctx, distrParams)
 
 			// configure upgrade for gamm module's pool creation fee param add
-			app.GAMMKeeper.SetParams(ctx, gammtypes.NewParams(sdk.Coins{sdk.NewInt64Coin("uosmo", 1)})) // 1 uOSMO
+			gammParams := app.GAMMKeeper.GetParams(ctx)
```
ValarDragon, line 326: Doesn't this need to set some new params as well?

proto/osmosis/gamm/v1beta1/genesis.proto
```diff
@@ -3,25 +3,35 @@ package osmosis.gamm;
...
-        (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
-        (gogoproto.moretags) = "yaml:\"pool_creation_fee\"",
-        (gogoproto.nullable) = false
-    ];
+  repeated cosmos.base.v1beta1.Coin pool_creation_fee = 1 [
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (gogoproto.moretags) = "yaml:\"pool_creation_fee\"",
+    (gogoproto.nullable) = false
+  ];
+  uint64 num_twap_history_per_deletion = 2;
```
ValarDragon, line 17: Can a comment be added for what this is?

proto/osmosis/gamm/v1beta1/genesis.proto
```diff
@@ -3,25 +3,35 @@ package osmosis.gamm;
...
+  // interval is in the unit of block height
+  uint64 twap_history_deletion_interval = 3;
```
ValarDragon, line 19: Oh interesting, I guess deletions are happening in begin/end block?

app/app.go
ValarDragon: Ah this isn't going in v4 so its not needed. We should make an issue for describing whats going to be needed in the `v5` upgrade handler then

app/app.go
```diff
@@ -487,6 +489,7 @@ func NewOsmosisApp(
 		crisistypes.ModuleName, govtypes.ModuleName, stakingtypes.ModuleName, claimtypes.ModuleName,
 		// Note: epochs' endblock should be "real" end of epochs, we keep epochs endblock at the end
 		epochstypes.ModuleName,
+		gammtypes.ModuleName,
```
ValarDragon, line 492: Can this be put before the epochs end blocker

proto/osmosis/gamm/v1beta1/genesis.proto
```diff
@@ -3,25 +3,35 @@ package osmosis.gamm;
...
 message GenesisState {
-    repeated google.protobuf.Any pools = 1
-        [ (cosmos_proto.accepts_interface) = "PoolI" ];
-    uint64 next_pool_number = 2;
-    Params params = 3 [ (gogoproto.nullable) = false ];
+  repeated google.protobuf.Any pools = 1
+      [ (cosmos_proto.accepts_interface) = "PoolI" ];
+  uint64 next_pool_number = 2;
+  Params params = 3 [ (gogoproto.nullable) = false ];
+  uint64 next_twap_history_delete_pool_id = 4;
```
ValarDragon, line 36: I'm confused by this as well

ValarDragon:
Can you update the GAMM spec you wrote earlier to briefly describe how the TWAP's are stored in state and then cleared?

x/gamm/keeper/grpc_query_test.go
```diff
@@ -257,3 +257,17 @@ func (suite *KeeperTestSuite) TestQuerySpotPrice() {
...
+
+	poolId := suite.preparePool()
+
+	_, err := queryClient.PoolTwap(gocontext.Background(), &types.QueryPoolTwapRequest{
+		PoolId:        poolId,
+		TokenInDenom:  "",
+		TokenOutDenom: "bar",
+		Duration:      10,
+	})
+	suite.Require().Error(err)
```
ValarDragon, line 272: TODO: We should add a test that doesn't error here as well

app/app.go
mattverse: So the whole set params part isnt needed?

`	// configure upgrade for gamm module's pool creation fee param add
			gammParams := app.GAMMKeeper.GetParams(ctx)
			gammParams.PoolCreationFee = sdk.Coins{sdk.NewInt64Coin("uosmo", 1)} // 1 uOSMO
			app.GAMMKeeper.SetParams(ctx, gammParams)`

proto/osmosis/gamm/v1beta1/genesis.proto
mattverse: yep the deletion happens on the end block!

proto/osmosis/gamm/v1beta1/genesis.proto
mattverse: So basically, every time delete twap history is activated, it only iterates over the designated number of pools. The field `next_twap_history_delete_pool_id` is needed to keep track the id of the pool that deletion should be starting off with

mattverse:
Due to rebase of this branch, I'm going to close this PR for now and open a new one.

PR #460, opened 2021-09-05
Starportify, begin a conversation on a monorepo

faddat:
Hey, 

This is probably a somewhat unusual pull request and certainly it's not attached to any issue.

I have found that Starport generates extremely clean dart libraries that are relatively easy to use. 

Furthermore, the flutter app that it scaffolds, well, it paves the way for a multiplatform mobile app. 

The "networks" repo, shoud be a folder underneath "osmosis".  It is an essentisal part of osmosis, the network cannot actually be used without it.  

Furthermore, the current front end repository should be a folder underneath "osmosis" as well.  

The thing is, all of these things are "osmosis", and lacking one of them, you don't get "osmosis".  

Last-- why this here flutter app?

It is nascent and barely does anything....

well, yes.  It is.

And by placing it in the main repository I believe that we can call attention to the fact that it exists and more easily attract developers who're excited by the idea of a mobile osmosis that has every feature of the web version.


ValarDragon:
I'm pretty amenable to moving networks into this repository, thoughts @sunnya97 ? I think your right, keeping it separate doesn't seem to help any actual user.

I don't think I agree with a monorepo in general. I think the front-end code should be in its own repository. The front-end repo should be managed and maintained independently of the chains repository. Regarding flutter, I'm not sure about this, I don't know anything about it. 

We should talk separately about for Osmosis mobile access, and how its going to look. I think once Keplr mobile is live, it will first be accessible via browser? cc @dogemos @Thunnini ?

faddat:
Hey, just seeing this.  I'll get networks on in here :).

Regarding flutter, currently what is there is a blank app for ios/android, and more importantly some really great dart libs that are generated from proto/modules.

ValarDragon:
Hey! Now that networks is merged in, should we close this PR?

PR #464, opened 2021-09-08
Update init.go to increase max outbound peers for validator nodes

bfritton:
Change to max outbound peers for validator node as requested by Valardragon and proposed by FONDSMATIVE and dff | mp20


ValarDragon (approved):
LGTM! Just going to leave open for a day or so, to check if any other folks would like to opine. I think this is long term too high, but probably useful short term.

My reasoning is that with high numbers of peers being saturated on the peerlist, increasing the number of outbound connections you try makes sense. And when the p2p network reforms each epoch, you retry connecting to more peers. 

(Granted Tendermint p2p is complex, and I may be misunderstanding this!)

faddat:
I'd say "do not merge".

I recently made a PR on Tendermint itself and the key to my logic was changing the ratio:

I went from previous default (10 outbound and 40 inbound)

to 

new default (20 outbound and 160 inbound)

the goal is to have peers gettin all promiscuous.  In the p2p layer, we want them connecting, lots.

ValarDragon:
Should we be updating the defaults here to those settings Jacob?

faddat:
Yes, I should have made a PR to his branch.

I'll do that now.

faddat:
40 and 320 actually seem more appropriate for us.  

The network has always seemed kind of choked.

faddat:
@bfritton I'm sorry I slowed this down!


Have been super busy lately.  Surely, we needed to change these settings.

❤️

PR #469, opened 2021-09-11
only distribute once per user per gauge, use batched distribute function

### LINKED ISSUE #468: Integration for SDK upgrade for SendManyCoins
We should do a very simple adaptation for https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/gauge.go#L275 to make it only do one big multi-send, instead of many different single sends.

Furthermore, we should batch together all the sends to the same address.

Basically what we need to do is maintain a map in that function, and two slices
The map from `lock owner addr -> ID`. Lets call this `lock_owner_to_id_map`.

Then we need two slices for `id_to_decoded_addr`, and `id_to_distr_coins`.

Then we replace the call here: https://github.com/osmosis-labs/osmosis/blob/bcfadbf62906fb5ca44d0b7c7cd12b558c6d43a4/x/incentives/keeper/gauge.go#L303-L320 to instead be
```go
if ownerAddr in lock_owner_to_id_map {
  distr_id := lock_owner_to_id_map[ownerAddr]
  oldDistrCoins := id_to_distr_coins[distr_id] 
  id_to_distr_coins[distr_id] = distrCoins.Add(oldDistrCoins)
}
else {
    latest_id += 1
    lock_owner_to_id_map[ownerAddr] = latest_id
    decodedOwnerAddr, err := sdk.AccAddressFromBech32(ownerAddr)
    id_to_decoded_addr[latest_id] = decodedOwnerAddr
    id_to_distr_coins[latest_id] = distrCoins
}
```

We need this ID because the order of state changes matters due to gas pricing, and some unfortunate properties of IAVL. And iterating over maps is non-deterministic. And we need to iterate in the right order. 

Then after iterating over all the locks, we do a single mega-send of:
```go
SendCoinsFromModuleAccountToManyAccounts(ctx, sending_module_account, id_to_decoded_addr, id_to_distr_coins)
```

And then we do a single loop for the liquidity events.

We should ensure that we had existing tests for the old code with multiple people receiving rewards from a gauge, and that those same tests work the same way after this change.

ValarDragon:
Closes #468 , using Matt's new work in the SDK there!

This basically needs a test to check compat between debugDistribute and the new distribution function

### DIFF
--- a/x/incentives/keeper/gauge.go
+++ b/x/incentives/keeper/gauge.go
@@ -272,7 +272,7 @@ func (k Keeper) FilteredLocksDistributionEst(ctx sdk.Context, gauge types.Gauge,
 }
 
 // Distribute coins from gauge according to its conditions
-func (k Keeper) Distribute(ctx sdk.Context, gauge types.Gauge) (sdk.Coins, error) {
+func (k Keeper) debugDistribute(ctx sdk.Context, gauge types.Gauge) (sdk.Coins, error) {
 	totalDistrCoins := sdk.NewCoins()
 	locks := k.GetLocksToDistribution(ctx, gauge.DistributeTo)
 	lockSum := lockuptypes.SumLocksByDenom(locks, gauge.DistributeTo.Denom)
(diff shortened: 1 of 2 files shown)


ValarDragon:
Reviewed per slack discussion

PR #471, opened 2021-09-11
Upgrade to tendermint 0.34.13 for the postgresql love.

faddat:
This Tendermint revision brings us the Postgresql endpoint, which in turn brings us graphql.


ValarDragon:
Will merge this in once v4.0.0 goes out! I want to make this on a v4.1.0 branch

faddat:
cool -- after the upgrade I will be working on a 0.44.0 + IBC 1.2.0 change-- it should not take too much.

faddat:
I cannot recommend merging this, even if I updated it.

#492

PR #475, opened 2021-09-13
Gauge test improvements

ValarDragon:
This PR adds more scaffolding to writing tests for gauges. It also adds simpler tests for testing the distsribute function of a gauge, and removes some duplicate code between perpetual and non-perpetual gauge tests.

As I was reading this code, I noticed we have a lot more test cleanup to do, and better testing to be adding here. I'll start writing it into issues, but wanted to add this quick PR

### DIFF
--- a/x/incentives/keeper/gauge_test.go
+++ b/x/incentives/keeper/gauge_test.go
@@ -8,22 +8,25 @@ import (
 	lockuptypes "github.com/osmosis-labs/osmosis/x/lockup/types"
 )
 
+var defaultLPDenom string = "lptoken"
+var defaultLPTokens sdk.Coins = sdk.Coins{sdk.NewInt64Coin(defaultLPDenom, 10)}
+var defaultLiquidTokens sdk.Coins = sdk.Coins{sdk.NewInt64Coin("foocoin", 10)}
+var defaultLockDuration time.Duration = time.Second
+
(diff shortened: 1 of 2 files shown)


x/incentives/keeper/gauge_test.go
```diff
@@ -8,22 +8,60 @@ import (
...
+	}{
+		{
+			users:           []userLocks{oneLockupUser, twoLockupUser},
+			gauges:          []perpGaugeDesc{defaultGauge},
+			expectedRewards: []sdk.Coins{oneKRewardCoins, twoKRewardCoins},
+		},
+	}
+	for _, tc := range tests {
+		suite.SetupTest()
+		gauges := suite.SetupGauges(tc.gauges)
```
mattverse, line 38: I was debugging this part and noticed that interestingly, three locks are existing despite we only created two locks above. Two were as expected, and the one extra one had the same defaultLPDenom and had the duration of `net/http.http2firstSettingsTimeout (2000000000)` has this lock been always existing?

mattverse (commented):
Comments in the code!

x/incentives/keeper/gauge_test.go
ValarDragon: Wait we created three locks tho, the one lockup user created one lock, and the two lockup user creates two locks

mattverse:
Yep! Think everythings lgtm!

PR #478, opened 2021-09-14
Unity/one entry per duration

UnityChaos:
Building from #389

Not sure if this is correct yet, needs a more through read through to make sure accumulation store is always properly updated.

### DIFF
--- a/x/lockup/keeper/lock.go
+++ b/x/lockup/keeper/lock.go
@@ -368,13 +368,6 @@ func (k Keeper) ClearAccumulationStores(ctx sdk.Context) {
 	k.clearKeysByPrefix(ctx, types.KeyPrefixLockAccumulation)
 }
 
-func (k Keeper) ClearAllAccumulationStores(ctx sdk.Context) {
-  denoms := k.getAccumulationDenoms(ctx)
-  for _, denom := range denoms {
-    k.accumulationStore(ctx, denom).Clear()
-  }
-}
-
 // ResetLock reset lock to lock's previous state on InitGenesis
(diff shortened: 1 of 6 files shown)


x/lockup/keeper/lock.go
```diff
@@ -300,10 +300,8 @@ func (k Keeper) addTokensToLock(ctx sdk.Context, lock *types.PeriodLock, coins s
 	}
 
 	// modifications to accumulation store
-	for _, coin := range lock.Coins {
-		// remove previous store for the lock ID and add again with updated value
-		k.accumulationStore(ctx, coin.Denom).Remove(accumulationKey(lock.Duration, lock.ID))
-		k.accumulationStore(ctx, coin.Denom).Set(accumulationKey(lock.Duration, lock.ID), coin.Amount)
+	for _, coin := range coins {
+		k.accumulationStore(ctx, coin.Denom).Increase(accumulationKey(lock.Duration), coin.Amount)
```
ValarDragon, line 304: Oh wow, looks like a bug before lol

UnityChaos: I think it was fine on current code. Just something that didn't exist when the PR was made. It was just replacing the whole lock accumulation obj and adding each coin in the new lock. Whereas now it's only increasing the accumulator by what is being added to the existing lock.

It would be a bug I think tho if I had left for _, coin := range lock.Coins and just changed the accumulationStore call, since then it would be adding the whole lock (duplicating the existing part)

ValarDragon: ah your right

PR #480, opened 2021-09-15
Switch accumulation store to ResetAllLocks, only remove after unbond complete

ValarDragon:
This also switches to a version of the SDK fork, that fixes CacheKVStore speed under unfortunate access patterns.

### DIFF
--- a/x/lockup/keeper/lock.go
+++ b/x/lockup/keeper/lock.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/store/prefix"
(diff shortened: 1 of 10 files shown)


x/lockup/keeper/lock.go
```diff
@@ -367,7 +368,68 @@ func (k Keeper) ClearAccumulationStores(ctx sdk.Context) {
 }
 
 // ResetLock reset lock to lock's previous state on InitGenesis
```
mattverse, line 370: `// ResetAllLocks reset all locks to previous states on InitGenesis`

x/lockup/keeper/lock.go
```diff
@@ -367,7 +368,68 @@ func (k Keeper) ClearAccumulationStores(ctx sdk.Context) {
...
+			amt := curDurationMap[d]
+			k.accumulationStore(ctx, denom).Increase(accumulationKey(d), amt)
+		}
+	}
+
+	return nil
+}
+
+// ResetLock reset lock to lock's previous state on InitGenesis
+func (k Keeper) resetLockNoAccumulationStore(ctx sdk.Context, lock types.PeriodLock) error {
```
mattverse, line 432: Is there a specific reason this function is called resetLockNoAccumulationStore? Curious why is changed from ResetLock -> resetLockNoAccumulationStore

x/lockup/keeper/lock.go
```diff
@@ -367,7 +368,68 @@ func (k Keeper) ClearAccumulationStores(ctx sdk.Context) {
...
+		for _, d := range durations {
+			amt := curDurationMap[d]
+			k.accumulationStore(ctx, denom).Increase(accumulationKey(d), amt)
+		}
+	}
+
+	return nil
+}
+
+// ResetLock reset lock to lock's previous state on InitGenesis
```
mattverse, line 431: `// resetLockNoAccumulationStore reset lock to lock's previous state on InitGenesis
`

mattverse (commented):
Questions in comment!
Logic looks LGTM!

x/lockup/keeper/lock.go
ValarDragon: I removed the accumulation store lines from here, to instead manage that separately in ResetAllLocks with faster batched logic. Maybe this should renamed to `ResetLockRefs`

mattverse: I see! I think `ResetLockRefs` would be a more intuitive function name long-term wise!

PR #482, opened 2021-09-17
fixed typo bug in osmosisdnode/Dockerfile

DiscreteLogarithm:
This bug prevents the localnet-start target of Makefile from creating necessary files needed for launch of a local testnet in the proper directory. The reason is that the `build` directory is mounted on `/osmosisd`, but the output directory is set to `.` (current directory) which in the docker file is specified as `WORKDIR /omsosisd`; therefore by the time testnet command finishes and docker exits the output files will be lost.


ValarDragon:
Lol, tysm for catching & fixing this

DiscreteLogarithm:
My pleasure.

PR #483, opened 2021-09-18
Small refactor to simplify distribution logic, and easily extend it

ValarDragon:


mattverse:
Nothing Much going on! Looks LGTM!

PR #484, opened 2021-09-22
Create archivenode.bash

faddat:
make an archive node

### DIFF
--- a/scripts/archivenode.bash
+++ b/scripts/archivenode.bash
@@ -3,9 +3,9 @@ git checkout v1.0.1
 make install
 osmosisd init archive
 wget -O ~/.osmosisd/config/genesis.json https://github.com/osmosis-labs/networks/raw/main/osmosis-1/genesis.json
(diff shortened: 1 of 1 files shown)


scripts/archivenode.bash
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+git checkout v1.0.1
```
ValarDragon, line 2:
```suggestion
git checkout v3.2.0
```

No reason to ever be using anything older, since its all state compatible!

ValarDragon (approved):

ValarDragon:
We should evolve this in the future to setup the cosmovisor folder fully, but this looks great to me! Thanks for making it

scripts/archivenode.bash
faddat: are you certain of this?

I've found it to explode unless following this pattern

faddat:
any time.  Also I now see what you mean about the docker.  I will "fix" it.

(remove arm build)

ValarDragon:
oh huh, not sure why I didn't merge this earlier

PR #485, opened 2021-09-22
Update docker.yml

faddat:
Temporary fix:

stop making arm64 docker images

Permanent fix: 

Our own build server with proper caching


ValarDragon:
I thought the problem is that it gets a 403 error whenever the person making the PR is PR'ing from their own fork

faddat:
seemed to take 20min, then 403 out :)

faddat:
We should fix this in hardware, not software, as that will allow us much more CI fun, without usual ci pain.

PR #486, opened 2021-09-22
Add Linter for github actions

mattverse:
This PR adds github action for adding default linters for the PRs.


ValarDragon:
I don't think this is working, since if it was there would be a million lint errors on the codebase. (Or alternatively it only lints the diff, but the doesn't look like that to me)

ValarDragon (approved):

PR #488, opened 2021-09-28
Add tokens to existing lock (MsgServer impl)

antstalepresh:


ValarDragon:
I suggest that for now we remove the new message, and just change the existing lock function to use AddTokensToLock internally. We can explore more once there are usecases that would need it, whether we should be supporting multiple locks per user-duration pair.

x/lockup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/osmosis-labs/osmosis/x/lockup/keeper"
+	"github.com/osmosis-labs/osmosis/x/lockup/types"
+)
+
+func (suite *KeeperTestSuite) TestMsgLockTokens() {
```
ValarDragon, line 11:
```suggestion
// TODO: Make table driven
func (suite *KeeperTestSuite) TestMsgLockTokens() {
```

ValarDragon (approved):

PR #489, opened 2021-09-28
fix: go fmt ./...

faddat:
This has the goal of making the `go fmt` github actions test pass.


ValarDragon:
Ah thanks, you race conditioned @mattverse to it haha

ValarDragon (approved):
Thanks for this 🙏

mattverse:
Oh wow. I was actually checking why the tests were passing and it was this PR. Thanks!

PR #494, opened 2021-10-02
Networksnnote

### LINKED ISSUE #493: Move networks into main repository
to simplify deployment of nodes and prime the pump for embedding genesis in the executable or some other means of ensuring that it takes a single command `osmosisd start` to start an osmosis node, we should move the networks repo into here.

faddat:
Closes #493  per discussion in #460


ValarDragon (approved):
Do you think we should only serve a single compressed file for the gentxs, or do you think we should keep all the distinct files?

Otherwise, this lgtm. I'll give it a day or two for anyone to chime in. (cc @sunnya97 @UnityChaos @Thunnini )

I think if were putting it into the main repo, we should perhaps rename the folder from networks

networks/.gitattributes
```diff
@@ -0,0 +1,4 @@
+cosmoshub-3/cosmoshub3_for_zero_height.json filter=lfs diff=lfs merge=lfs -text
+cosmoshub-3/snapshot.json filter=lfs diff=lfs merge=lfs -text
+osmosis-1/pregenesis.json filter=lfs diff=lfs merge=lfs -text
+osmosis-1/genesis.json filter=lfs diff=lfs merge=lfs -text
```
ValarDragon, line 4: Do you know if .gitattributes work if placed in subdirectories like this?

If not, we'll find out post-merge lol

ValarDragon:
Pinging folks again @sunnya97 @UnityChaos @Thunnini to chime in, otherwise lets merge it!

sunnya97:
Is there a way to test the .gitattributes thing?  Otherwise, we can just move the lfs stuff up to the root directory?

ValarDragon:
I think we should just merge it, and see if it works / play around with it after. Nothing is actively depending on the new links for a bit, will take a while to redirect things.

PR #496, opened 2021-10-03
chore: reuse docker containers when generating protobuf files

robert-zaremba:
* fixing the image version to the latest released (`@latest` is not good)
* optimizing the `make proto-gen` job by reusing proto containers instead of removing them all the time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,21 +156,26 @@ docs:
         echo "\033[91mSwagger docs are out of sync!!!\033[0m";\
         exit 1;\
     else \
-    	echo "\033[92mSwagger docs are in sync\033[0m";\
+        echo "\033[92mSwagger docs are in sync\033[0m";\
     fi
 	@echo
 	@echo "=========== Generate Complete ============"
 	@echo
 
+protoVer=v0.2
+protoImageName=tendermintdev/sdk-proto-gen:$(protoVer)
+containerProtoGen=osmosis-proto-gen-$(protoVer)
+containerProtoFmt=osmosis-proto-fmt-$(protoVer)
+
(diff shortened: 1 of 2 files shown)


ValarDragon (approved):
Oh amazing, this may have been the problem with my local docker install as well

ValarDragon:
Oh hrmm, doesn't work on my local docker install. I get
```
make proto-gen
Generating Protobuf files
/third_party/proto: does not exist
make: *** [proto-gen] Error 1
```

I'll check on my separate build server though, since make proto-gen didn't work on my laptop before either

robert-zaremba:
```
/third_party/proto: does not exist
```
sounds like a dependency problem with the cosmos-sdk. It's strange, because in SDK we are using the same docker container (v0.2) for 0.42 builds.
could you remove your docker images and containers and recheck?

ValarDragon:
@mattverse pointed out that the problem was that I have a local vendor directory. Not sure why that causes issues, but it does 😕 .

After deleting my vendor folder, both this branch and main worked for me!

ValarDragon (approved):
Thanks for fixing this!

robert-zaremba:
Cool, happy that it works and thanks for merging.

> the problem was that I have a local vendor directory

Ah, most likely you had proto files in that directory. The proto compilation scrip is importing all proto files and tries to compile them: https://github.com/osmosis-labs/osmosis/blob/main/scripts/protocgen.sh#L12

ValarDragon:
Ahh thanks, looks like we just need to add the line for ignoring the vendor directory there. I'll submit a PR here and on the SDK for that!

PR #497, opened 2021-10-03
chore: use latest go in builds and CI

robert-zaremba:
In the Cosmos SDK we already updated to Go v1.17. Let's update here as well.


ValarDragon:
Can you run go mod tidy?

robert-zaremba:
updated

ValarDragon (approved):
Thanks for the fix!

PR #498, opened 2021-10-03
chore: bump gogo/protobuf

robert-zaremba:
regen-network/protobuf was updated at the beginning of the year (9 months ago). Llet's keep this dependency synced in Osmosis as well.


go.mod
```diff
@@ -17,19 +18,17 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
-	github.com/tendermint/tendermint v0.34.12
+	github.com/tendermint/tendermint v0.34.13
```
ValarDragon, line 21: Can we hold off on this? It was reported to be causing some issues when used on mainnet.

go.mod
```diff
@@ -17,19 +18,17 @@ require (
...
+	google.golang.org/grpc v1.40.0
+	google.golang.org/protobuf v1.27.0
 	gopkg.in/yaml.v2 v2.4.0
 )
 
 replace google.golang.org/grpc => google.golang.org/grpc v1.33.2
 
-replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.2-alpha.regen.4
-
-replace github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.12
```
ValarDragon, line 32: Why was this replace removed?

robert-zaremba: you don't need to replace tendermint - it's already fixed in the "require" section.

ValarDragon: We should still be doing this just in case dependencies use other versions, no?

go.mod
robert-zaremba: I thought tendermint 0.34.12 has problems. `..13` is needed for IAVL update, which has good performance improvements.

robert-zaremba: you are right, Bez confirmed.

go.mod
robert-zaremba: Yes - it replaces all versions. It looks like a hack. You are depending on your own version of SDK - maybe it's better to fix the version you want there? Anyway - I will rollback that change and in the future you can consider removing it.

ValarDragon (approved):
Thank you for this!

PR #501, opened 2021-10-05
Hotfix - Fix event attribute in incentives module

mattverse:
This PR fixes the bug in the current event emitted in the incentive module.
Converting address to string without using the String() method ignores the process of going through bech32 encoding, consequently emitting events with the wrong address attribute, making it impossible to query based on the incentive events.


ValarDragon:
This is not how it should have been fixed. Should have used something else on the distrinfo struct

PR #505, opened 2021-10-05
Remove linear # of bech32 encodes introduced

ValarDragon:
Removes computational overhead introduced in #501, for fixing printed format of bech32 addresses


mattverse (approved):
Looks like a way more intuitive way to have the bech32addr while reducing computation! Nice!

PR #491, opened 2021-10-02
Fix lint - errcheck

mattverse:
This PR fixes all lint errors - mainly err check throughout the whole code.

PR #506, opened 2021-10-06
Make `make lint` work

ValarDragon:
Makes make lint match whats in CI.

Also fixes two errors that got added in parallel to #491 being merged


mattverse (approved):
LGTM! Couldn't think of fixing the make file when I was doing link-fixes! Looks great and good to go!

PR #541, opened 2021-10-18
Gamm Spec Out V3 - Modulize File Structure

mattverse:
## Description

This PR is the 3rd version of Gamm spec out, including the main generalization and spec out of gamm module by seperating pool-models/balancer package with types package. This PR includes the following changes in detail:

1. Proto Changes
- divides proto packages into two different directories: `proto/osmosis/gamm/v1beta`, `proto/osmosis/gamm/pool-models/balancer` where gamm/v1beta directory includes protobufs that could be used for generalized usage(such as `PoolAsset`), and gamm/pool-models/balancer directory only includes protobufs limited to the scope of a balancer pool.
- `proto/osmosis/gamm/pool-models/balancer` also includes tx.proto, which includes Msgs that uses protos defined in the package.

2. gamm module structure changes 
- gamm module has a new directory, pool-models at top level. pool-models would include packages for each specific pool models.
- each specific pool packages defined in pool-models would be importing gamm's types package, but should not vice-versa(due to import cycles in go lang)

3. pool-models requirements
A package defined under pool-models directory(e.g balancer package in this PR) should meet the following requirements.
- implement `poolI` for the specific pool
- have marshal.go file where marshaling would happen in accordance to the pool structure it has.

4. requirements for defining msgs in a package under pool-models.
In the cases of where tx or msgs that has dependency on the created pool model package, the following requirements has to be specified.
- create tx.proto under the correct proto directory 
- create msgs.go under the specified package, implement the necessary methods(e.g. Route(), Type(), ValidateBasic()).
- create codec.go under the specified package for the model, register legacy animo codec and interface each using `RegisterLegacyAminoCodec`, `RegisterInterfaces` respectively.
- register newly created codec in module.go in gamm module.

5. PoolParams Cli
Even though all pools would have their own poolParams proto, query for PoolParams still exists in generalized form(in the form of Any). Thereby, `grpc_query.go` should be implementing poolParams type of the specific module using switch cases.

______

For contributor use:


proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package osmosis.gamm.v1beta1;
+package osmosis.gamm.poolmodels;
```
ValarDragon, line 2: should this still be behind a v1beta1?

proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
```diff
@@ -89,7 +69,7 @@ message SmoothWeightChangeParams {
 
 // BalancerPoolParams defined the parameters that will be managed by the pool
 // governance in the future. This params are not managed by the chain
-// governance. Instead they will be managed by the token holders of the pool.
+// governanace. Instead they will be managed by the token holders of the pool.
```
ValarDragon, line 72: I think this introduces a typo 😅 
```suggestion
// governance. Instead they will be managed by the token holders of the pool.
```

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -90,9 +92,7 @@ message QueryNumPoolsResponse {
 message QueryPoolParamsRequest {
   uint64 poolId = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
 }
-message QueryPoolParamsResponse {
-  oneof params { BalancerPoolParams balancerPoolParams = 1; }
-}
+message QueryPoolParamsResponse { google.protobuf.Any params = 1; }
```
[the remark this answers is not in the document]
ValarDragon: I'm a bit worried this will end up causing lots of complexities unless every params contains a `type` field. Perhaps we can add that to every pools query response?

x/gamm/genesis.go
```diff
@@ -44,11 +42,7 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	}
 	poolAnys := []*codectypes.Any{}
 	for _, poolI := range pools {
-		pool, ok := poolI.(*types.BalancerPool)
-		if !ok {
-			panic(fmt.Errorf("pool (%d) is not basic pool", poolI.GetId()))
-		}
-		any, err := codectypes.NewAnyWithValue(pool)
```
ValarDragon, line 51: Just want to double check, theres no reason we should have this restriction right?

mattverse:
@ValarDragon Can you also review the `poolI` interface fields? Generalizing the gamm structure, I don't think we need all the methods currently in `poolI` and want to remove some of the unnecessary methods from poolI. Whats your opinion on this?

proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
mattverse: Are you referring to the entire file itself? If my understanding is right upon your question, pool-models is declared as a whole new package in gamm, thats where this is coming from.

proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
mattverse: My bad, fixed.

proto/osmosis/gamm/v1beta1/query.proto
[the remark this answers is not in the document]
mattverse: Are you referring to `type` field of a string to indicate which type it is?
ex)
```
message QueryPoolParamsResponse {
      google.protobuf.Any params = 1; 
      string poolType = 2;
}
```

x/gamm/keeper/pool.go
```diff
@@ -94,7 +95,7 @@ func (k Keeper) newBalancerPool(ctx sdk.Context, balancerPoolParams types.Balanc
 		return nil, sdkerrors.Wrapf(types.ErrPoolAlreadyExist, "pool %d already exist", poolId)
 	}
 
-	err = k.SetPool(ctx, pool)
+	err = k.SetPool(ctx, &pool)
```
mconcat, line 98: Is there a reason behind changing to pointer implementation of Pool interface?

mconcat (approved):
Looks good, just one question!

x/gamm/keeper/pool.go
mattverse: If not it omits an error saying that the methods of poolI has pointer receiver

proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
ValarDragon: I see, tbh I don't really understand the v1beta1 stuff. Lets make an issue for this question, and then ask regen folks post-holidays

proto/osmosis/gamm/v1beta1/pool.proto
```diff
@@ -0,0 +1,30 @@
...
+  // Coins we are talking about,
+  // the denomination must be unique amongst all PoolAssets for this pool.
+  cosmos.base.v1beta1.Coin token = 1
+      [ (gogoproto.moretags) = "yaml:\"token\"", (gogoproto.nullable) = false ];
+  // Weight that is not normalized. This weight must be less than 2^50
+  string weight = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"weight\"",
+    (gogoproto.nullable) = false
+  ];
```
ValarDragon, line 20: In a second issue/PR I think we should consider just removing `PoolAsset` entirely, and then having weights be fully managed in balancer params.

proto/osmosis/gamm/v1beta1/query.proto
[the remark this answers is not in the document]
ValarDragon: We discussed in call, `any` already has an `@type` field in the response that makes this easy to parse.

x/gamm/client/cli/cli_test.go
```diff
@@ -870,6 +870,7 @@ func (s *IntegrationTestSuite) TestGetCmdPool() {
 
 				resp := types.QueryPoolResponse{}
 				s.Require().NoError(clientCtx.JSONCodec.UnmarshalJSON(out.Bytes(), &resp), out.String())
+				// s.Require().NoError(clientCtx.JSONMarshaler.UnmarshalJSON(out.Bytes(), &resp), out.String())
```
ValarDragon, line 873: err should this just be deleted, or do we have a merge conflict?

x/gamm/keeper/pool.go
ValarDragon: That makes sense to me, several pool methods are mutative, so they need a pointer receiver.  (And the SetPool interface needs what matches the interface)

x/gamm/keeper/pool_service_test.go
```diff
@@ -60,8 +69,16 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
 			keeper := suite.app.GAMMKeeper
 			prevFeePool := suite.app.DistrKeeper.GetFeePoolCommunityCoins(suite.ctx)
 			prevAcc1Bal := suite.app.BankKeeper.GetAllBalances(suite.ctx, acc1)
-			poolId, err := keeper.CreateBalancerPool(suite.ctx, acc1, defaultBalancerPoolParams,
-				defaultPoolAssets, defaultFutureGovernor)
+			poolId, err := keeper.CreateBalancerPool(suite.ctx, acc1, balancer.BalancerPoolParams{
+				SwapFee: sdk.NewDecWithPrec(1, 2),
+				ExitFee: sdk.NewDecWithPrec(1, 2),
+			}, []types.PoolAsset{{
```
ValarDragon, line 75: Lets make these params be `defaultBalancerPoolParams`, and `defaultPoolAssets`?

x/gamm/types/pool_asset.go
```diff
@@ -64,75 +75,28 @@ func SortPoolAssetsByDenom(assets []PoolAsset) {
...
+			return err
 		}
-		curWeightSum := asset.Weight.Add(other[i].Weight)
-		weightSum[i] = PoolAsset{Token: asset.Token, Weight: curWeightSum}
-	}
-	return weightSum
-}
 
-// assumes 0 < d < 1
-func poolAssetsMulDec(base []PoolAsset, d sdk.Dec) []PoolAsset {
```
ValarDragon, line 126: Do we just no longer need these pool asset arithmetic methods?

ValarDragon (approved):
LGTM. 

I think we just need to make followup issues for:
1) Separating out general pool parameters into their own struct. (Swap fee, Exit fee)
2) Removing PoolAsset as a struct, and moving all weight handling to be internal to balancer
3) the .proto `v1beta1` question.

Then I think we're good to merge this!

proto/osmosis/gamm/v1beta1/pool.proto
mattverse: Agreed with this! 
Issued metioned here https://github.com/osmosis-labs/osmosis/issues/681

proto/osmosis/gamm/pool-models/balancer/balancerPool.proto
mattverse: haha. I'm supposing its just a legacy in structuring the proto files from the cosmos-sdk and that it does not have any additional functionallities. I didn't have any issues when I wasn't using v1beta1 as a package file structure, they should be good to be removed as well

x/gamm/client/cli/cli_test.go
mattverse: Deleted as of now, thanks for catching this!

mattverse:
Just created three issues following this PR
- https://github.com/osmosis-labs/osmosis/issues/681
- https://github.com/osmosis-labs/osmosis/issues/682
- https://github.com/osmosis-labs/osmosis/issues/683

x/gamm/types/pool_asset.go
mattverse: this was moved to `balancer_pool.go` bc the concept of pool weight moved to balancer package

mattverse:
@ValarDragon Could you also review poolI in `pool.go`? 
Now that we're generalizing gamm module, I don't think we need all the method that was in poolI to still be there.
If we want to get this PR merged and mention this in a separate issue, I think that would also be fine!

ValarDragon:
Agreed that a lot of these methods should be changed.

I think it'd be easiest to do that when separating out the parameters. (So we can remove all the weight references, etc.)

I think we also may want secondary interfaces for different types of pools. (CFMM, batched-CFMM, etc.) but I think it's perhaps too early for doing that rn / unclear if we want to push that to cosmwasm.

PR #551, opened 2021-10-24
Synthetic lockup

antstalepresh:
Closes: #XXX

## Description

______

For contributor use:


x/lockup/keeper/synthetic_lock_test.go
```diff
@@ -0,0 +1,311 @@
+package keeper_test
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/osmosis-labs/osmosis/x/lockup/types"
+)
+
+func (suite *KeeperTestSuite) TestSyntheticLockupCreateGetDeleteAccumulation() {
```
ValarDragon, line 10: Can we make all the tests in this file have much less duplicate test setup code & preferrably be table driven. 

I don't think we should have this much duplicate test setup code going forward, it becomes too hard to extend and too hard to understand

x/lockup/keeper/synthetic_lock_test.go
```diff
@@ -0,0 +1,311 @@
...
+	locks = suite.app.LockupKeeper.GetLocksPastTimeDenom(suite.ctx, "stake", suite.ctx.BlockTime())
+	suite.Require().Len(locks, 1)
+	locks = suite.app.LockupKeeper.GetLocksLongerThanDurationDenom(suite.ctx, "stake", time.Second)
+	suite.Require().Len(locks, 1)
+	amount = suite.app.LockupKeeper.GetLockedDenom(suite.ctx, "stake", time.Second)
+	suite.Require().Equal(amount.String(), "10")
+
+	// check queries for synthetic denom after creating synthetic lockup
+	locks = suite.app.LockupKeeper.GetAccountLockedPastTimeDenom(suite.ctx, addr1, "stakestakedtovalidator1", suite.ctx.BlockTime())
+	suite.Require().Len(locks, 1)
```
ValarDragon, line 121: Lets make an `expectedLocks` variable, and then check that the returned locks is equal to what we expect? (Rather than just checking that they are of the same length?

x/lockup/keeper/synthetic_lock_test.go
```diff
@@ -0,0 +1,311 @@
...
+	suite.Require().Nil(synthLock)
+}
+
+func (suite *KeeperTestSuite) TestSyntheticLockupDeleteAllMaturedSyntheticLocks() {
+	suite.SetupTest()
+
+	// lock coins
+	addr1 := sdk.AccAddress([]byte("addr1---------------"))
+	coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
+	suite.LockTokens(addr1, coins, time.Second)
```
ValarDragon, line 206: Lets refactor this / use general logic. (I feel like there should already be some somewhere in the modules test code?)

antstalepresh: What you mean by general logic here? LockTokens function is a utility function created to reduce the steps of creating a new lockup along with token mint in case token is not available on the account.

x/lockup/keeper/synthetic_lock_test.go
antstalepresh: There were lots of test modifications on superfluid minimum viable version and I would like to implement this comment on that PR

x/lockup/keeper/synthetic_lock_test.go
antstalepresh: Sure, can add expectedLocks. There were lots of test modifications on superfluid minimum viable version and I would like to implement this comment on that PR to avoid conflicts

antstalepresh: Fixed as part of this commit
https://github.com/osmosis-labs/osmosis/pull/587/commits/6f793762da20c3904f639f5c5419e846e2af1db9

ValarDragon:
Closing because its all in #587, with several updates there

Issue #553, opened 2021-10-27
Implement clawback of  ALL Osmo and Ion at end of claim period (Prop 32)


sunnya97:
This was requested by governance in the successful passing of Proposal 32
https://www.mintscan.io/osmosis/proposals/32

Should this happen automatically on-chain or during an upgrade?

ValarDragon:
Closing in favor of #592

PR #560, opened 2021-10-29
Implement clawback of airdrop from inactive genesis addresses

sunnya97:
Closes: #553

## Description

Implements the automatic clawback of all OSMO and ION from genesis airdrop recipients at the end of the claim period (6 months after genesis).

For more details, see Osmosis proposal 32:  https://www.mintscan.io/osmosis/proposals/32

______

For contributor use:


cmd/osmosisd/cmd/genaccounts.go
```diff
@@ -183,6 +183,116 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
...
+					if err != nil {
+						return err
+					}
+
+					airdropAddrs[osmoAddr] = true
+				}
+			}
+
+			for addr := range ionAmts {
+				ionAddr, err := CosmosToOsmoAddress(addr)
```
mattverse, line 274: What's the reason we're not checking if IONBalance is not zero like above?

mattverse: I'm supposing its probably because ION snapshots only include addresses that got the airdrop?

mattverse (commented):
Other than the questions asked above, looks lgtm!

cmd/osmosisd/cmd/genaccounts.go
sunnya97: Yep.  The ION snapshot file already did that filtering, but the Osmosis one doesn't it has all the addresses from the cosmoshub-3 export

x/claim/keeper/claim_test.go
```diff
@@ -344,3 +344,62 @@ func (suite *KeeperTestSuite) TestClaimOfDecayed() {
...
+		coins := suite.app.BankKeeper.GetAllBalances(suite.ctx, addr)
+		if tc.expectClawback {
+			suite.Require().True(coins.IsEqual(sdk.NewCoins()),
+				"balance incorrect. test: %s", tc.name)
+		} else {
+			suite.Require().True(coins.IsEqual(sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 100), sdk.NewInt64Coin("uion", 100),
+			)), "balance incorrect. test: %s", tc.name)
+		}
+	}
```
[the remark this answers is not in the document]
sunnya97: I need all the accounts initialized _before_ EndAirdrop is called

cmd/osmosisd/cmd/testnetify/cmd.go
```diff
@@ -62,5 +62,9 @@ func writeGenesis(genesis app.GenesisState, path string) error {
 	}
 	defer file.Close()
 	encoder := json.NewEncoder(file)
-	return encoder.Encode(genesis)
+	err = encoder.Encode(genesis)
+	if err != nil {
+		return err
+	}
```
antstalepresh, line 68: Why is it changed from a shorter implementation to a longer one?

x/claim/keeper/claim.go
```diff
@@ -43,6 +43,30 @@ func (k Keeper) EndAirdrop(ctx sdk.Context) error {
...
+		if err != nil {
+			return err
+		}
+		seq, err := k.accountKeeper.GetSequence(ctx, addr)
+		if err != nil {
+			return err
+		}
+		if seq == 0 {
+			osmoBal := k.bankKeeper.GetBalance(ctx, addr, "uosmo")
+			ionBal := k.bankKeeper.GetBalance(ctx, addr, "uion")
```
antstalepresh, line 60: How does it work if someone who owns this address didn't claim airdrop but have sent some of his other account balance to this account?

antstalepresh:
`claim` module isn't doing this stuff already?

x/claim/keeper/claim.go
sunnya97: Heh 😅 

I supposed they'd have all their coins clawed back, including the non-genesis ones that were sent there.  I think that's fine.  Seems like a very weird case that's not worth dealing with.

cmd/osmosisd/cmd/testnetify/cmd.go
sunnya97: I don't know...  I think my linter was demanding this for some reason.  Let me try putting it back to how it was

cmd/osmosisd/cmd/testnetify/cmd.go
```diff
@@ -62,5 +62,9 @@ func writeGenesis(genesis app.GenesisState, path string) error {
...
+	return nil
```
sunnya97, line 69:
```suggestion
	return encoder.Encode(genesis)
```

sunnya97:
@antstalepresh currently the claim module claws back the un-claimed coins from the `claims` module account (which was 80% of everyone's airdrop amount).  But Osmosis governance voted in Prop 32 to also clawback the 20% of OSMO that was in people's accounts as well the IONs for inactive addresses

x/claim/keeper/claim_test.go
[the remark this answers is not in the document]
mattverse: Ah makes sense

sunnya97:
@ValarDragon this should be added to this upgrade, in case we don't get another upgrade in by Dec 18

mattverse:
Yep, I marked it as a v5 upgrade checklist, would assign myself to go over the edge case @antstalepresh  mentioned within today, tomorrow and get it merged as well!

ValarDragon:
Sunny and I discussed today. Its probably not worth resolving that edge case, the governance proposal's "letter of the law" interpretation seems to support pruning the entire account. The "spirit of the proposal" is incredibly unclear, this edge case wasn't spelled out. But theres been no concern from the community about it.

Would rather we focus on other things, then resolving it.

mattverse (approved):
Cool! In that case, I think this PR is good to go, ready to merge!

x/claim/keeper/claim_test.go
```diff
@@ -344,3 +344,62 @@ func (suite *KeeperTestSuite) TestClaimOfDecayed() {
...
+			expectClawback: false,
+		},
+		{
+			name:           "airdrop address inactive",
+			address:        "osmo10008uvk6fj3ja05u092ya5sx6fn355wayx20rq",
+			sequence:       0,
+			expectClawback: false,
+		},
+		{
+			name:           "non airdrop address inactive",
```
ValarDragon, line 370:
```suggestion
			name:           "non airdrop address active",
```

This needs an actual test for an inactive account. @mattverse can you add that here?

x/claim/keeper/claim_test.go
```diff
@@ -344,3 +344,62 @@ func (suite *KeeperTestSuite) TestClaimOfDecayed() {
...
+			name:           "airdrop address active",
+			address:        "osmo1000644etcp4k4awz77kvy3pryplv8ky8py962d",
+			sequence:       1,
+			expectClawback: false,
+		},
+		{
+			name:           "airdrop address inactive",
+			address:        "osmo10008uvk6fj3ja05u092ya5sx6fn355wayx20rq",
+			sequence:       0,
+			expectClawback: false,
```
ValarDragon, line 367: This should be true?

ValarDragon (changes requested):
I corrected the test cases. @sunnya97 can you investigate failures. On scan appears to be clawing back from non-airdrop inactive addresses.

Also changes like this require updates to the spec. Please update the concepts.

sunnya97:
Fixed

x/claim/keeper/claim.go
```diff
@@ -51,17 +53,21 @@ func (k Keeper) ClawbackAirdrop(ctx sdk.Context) error {
...
-		if err != nil {
-			return err
-		}
-		if seq == 0 {
-			osmoBal := k.bankKeeper.GetBalance(ctx, addr, "uosmo")
-			ionBal := k.bankKeeper.GetBalance(ctx, addr, "uion")
-			err = k.distrKeeper.FundCommunityPool(ctx, sdk.NewCoins(osmoBal, ionBal), addr)
+
+		acc := k.accountKeeper.GetAccount(ctx, addr)
+		if acc != nil {
```
ValarDragon, line 61: I'm trying to make sure I understand the prior bug. Why would it clawback from all accounts? (Or were the addresses just wrong)

Also is there any expected scenario where acc would be nil, or this is just in case something unexpected occurs?

sunnya97: No, it was doing the opposite.  It wasn't clawing back from any accounts.
See original testcases:
https://github.com/osmosis-labs/osmosis/blob/cf7ba91f7bd74e6b9e7d3559f5e0e0ff9da87b8b/x/claim/keeper/claim_test.go

It's cause the function was exiting early, b/c as soon as it saw an account that wasn't in the account store it would error.

ValarDragon: Ahh that makes sense

ValarDragon (changes requested):
Airdrop addrs is a map that we iterate over. I'm concerned over determinism issues here. Lets update to not use a map, but instead a list of addresses. This `bool` value in the map is not important.

ValarDragon:
On further analysis, this non-determinism has pretty high risk of chain halts. The clawback deletes state entries. IAVL is deletion order dependent.

My understanding is that the delete orders do get sorted by key prior to hitting IAVL in commit, but I'd definitely have to double check, and am quite hesistant to rely on that.

x/claim/keeper/claim.go
```diff
@@ -40,6 +40,55 @@ func (k Keeper) EndAirdrop(ctx sdk.Context) error {
...
+		// This is viewed as an edge case, that the text of Osmosis proposal 32 indicates should
+		// be done via sending these excess funds to the community pool.
+		//
+		// Proposal text to go off of: https://www.mintscan.io/osmosis/proposals/32
+		// ***Reminder***
+		// 'Unclaimed' tokens are defined as being in wallets which have a Sequence Number = 0,
+		// which means the address has NOT performed a single action during the 6 month airdrop claim window.
+		// ******CLAWBACK PROPOSED FRAMEWORK******
+		// TLDR -- Send ALL unclaimed ION & OSMO back to the community pool
+		// and prune those inactive wallets from current state.
```
sunnya97, line 82: We don't prune accounts right now.  I don't want to deal with the potential side effects

sunnya97:
Swapped map for an array

ValarDragon (approved):

ValarDragon:
LGTM, A second person should just verify the airdrop accounts (I have not), and then its good to merge.

cmd/osmosisd/cmd/genaccounts.go
```diff
@@ -183,6 +183,116 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
...
+func GetAirdropAccountsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get-airdrop-accounts [input-snapshot-file] [input-ions-file]",
+		Short: "Get list of all accounts that are being airdropped to at genesis",
+		Long: `Get list of all accounts that are being airdropped to at genesis
+Both OSMO and ION recipients
+
+Example:
+	osmosisd import-genesis-accounts-from-snapshot ../snapshot.json ../ions.json ./address.txt
+`,
```
ValarDragon, line 245:
```suggestion
		Use:   "get-airdrop-accounts [input-snapshot-file] [input-ions-file] [output-file]",
		Short: "Get list of all accounts that are being airdropped to at genesis",
		Long: `Get list of all accounts that are being airdropped to at genesis
Both OSMO and ION recipients. If erroring, ensure to `git lfs pull`

Example:
	osmosisd import-genesis-accounts-from-snapshot networks/cosmoshub-3/snapshot.json networks/osmosis-1/ions.json output_address.json
`,
```

ValarDragon:
Checked the file, its good

Issue #546, opened 2021-10-22
Cannot set incentive gauge weight to 0


UnityChaos:
The validation checks for gauge weights in an `update-pool-incentives` proposal are required to be positive. This makes setting a gauge to 0 weight impossible. Should be changed to positive or zero.

https://github.com/osmosis-labs/osmosis/blob/59db94c0b49cfcda93a71fe665ea684be1e48390/x/pool-incentives/types/record.go#L3-L5

(Unless there's some way to propose deleting a gauge rather than setting it to zero I'm unaware of)

PR #563, opened 2021-10-29
Pool Incentives: Allow Zero Weight

jeebster:
Closes: #546 

## Description

Validation check for pool-incentive gauge weight will only throw for negative values
______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


ValarDragon (approved):
LGTM! Just checked in https://github.com/osmosis-labs/osmosis/blob/ea6bda76c5158b75a630671e821524c074b19f31/x/pool-incentives/keeper/distr.go#L154, the Update function already correctly handles zero weights, so we should be set.

ValarDragon:
The lint issue seems unrelated, something broke on main 😕

Issue #535, opened 2021-10-16
Make an active gauges per pool query


ValarDragon:
This should be done to remove the need for PR's like this on the front-end: https://github.com/osmosis-labs/osmosis-frontend/pull/62/commits/974a72f1016937f1e1c874a00c9df12f43cf11b1 (so removing the need for this entire field: https://github.com/osmosis-labs/osmosis-frontend/blob/master/src/config.ts#L70)

mattverse:
Closed with PR https://github.com/osmosis-labs/osmosis/pull/572

Issue #507, opened 2021-10-07
Remove JSON marshalling of gaugeIDs


ValarDragon:
We use JSON marshalling for gaugeIDs in several locations. We should remove this, and replace it with a more efficient serialization format. This will need a state migration.

I just noticed it being used here: https://github.com/osmosis-labs/osmosis/blob/d3e00da98551cc77db54ea4de0f974703f298de2/x/incentives/keeper/store.go#L80 and in a couple iterators. We should change all of these to no longer be using json.

mattverse:
I'm closing this issue noticing that it isn't getting mistakenly getting JSON marshalled to the state machine, but only a process of unmarshalling.

PR #572, opened 2021-11-03
Create Query for Active Gauge Per Denom & Externally Incentivized Gauges

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/535

## Description

This PR mainly focuses on adding two queries for gauges. 
- Query for active gauges for a specified denom
- Query for externally incentivized gauges.

Two queries are defined in two different modules, incentives and pool-incentives, as the state related to pool-gauges are specified within the pool-incentives module, whereas Gauge states are specified within the incentive module.

 

______

For contributor use:


x/incentives/keeper/gauge.go
```diff
@@ -446,13 +446,8 @@ func (k Keeper) GetGaugeByID(ctx sdk.Context, gaugeID uint64) (*types.Gauge, err
...
 
 // GetGaugeFromIDs returns gauges from gauge ids reference
-func (k Keeper) GetGaugeFromIDs(ctx sdk.Context, refValue []byte) ([]types.Gauge, error) {
+func (k Keeper) GetGaugeFromIDs(ctx sdk.Context, gaugeIDs []uint64) ([]types.Gauge, error) {
 	gauges := []types.Gauge{}
-	gaugeIDs := []uint64{}
-	err := json.Unmarshal(refValue, &gaugeIDs)
-	if err != nil {
-		return gauges, err
-	}
```
mattverse, line 455: Unrelated to this PR, but noticed that this method in PR takes in refValue(array of bytes) as parameter. 
I'm guessing this was originally written this way because pagination sends in paginated results as array of bytes, but IMO I think it's weird that a method takes in a parameter not json Unmarshalled, and doing it within the method.

mattverse: If folks think this change should not happen here, or need further discussion upon this change, I could also address this in a seperate PR

x/incentives/keeper/grpc_query.go
```diff
@@ -46,7 +47,12 @@ func (k Keeper) Gauges(goCtx context.Context, req *types.GaugesRequest) (*types.
 	valStore := prefix.NewStore(store, types.KeyPrefixGauges)
 
 	pageRes, err := query.FilteredPaginate(valStore, req.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
-		newGauges, err := k.GetGaugeFromIDs(ctx, value)
+		gaugeIDs := []uint64{}
+		err := json.Unmarshal(value, &gaugeIDs)
+		if err != nil {
+			panic(err)
+		}
```
ValarDragon, line 54: And this is an example of a place we'd use the `getGaugeFromIDJsonBytes`

x/incentives/keeper/grpc_query.go
```diff
@@ -86,6 +97,36 @@ func (k Keeper) ActiveGauges(goCtx context.Context, req *types.ActiveGaugesReque
...
+		for _, id := range guageIDs {
+			gauge, err := k.GetGaugeByID(ctx, id)
+			if err != nil {
+				panic(err)
+			}
+			if gauge.IsPerpetual {
+				gauges = append(gauges, *gauge)
+			} else if gauge.NumEpochsPaidOver < gauge.FilledEpochs {
+				gauges = append(gauges, *gauge)
+			}
```
ValarDragon, line 118: Is there no direct keeper method to determine if the gauge is active?

x/pool-incentives/client/cli/cli_test.go
```diff
@@ -203,3 +203,37 @@ func (s *IntegrationTestSuite) TestGetCmdIncentivizedPools() {
...
+	testCases := []struct {
+		name      string
+		expectErr bool
+		respType  proto.Message
+	}{
+		{
+			"query external incentivized pools",
+			false, &types.QueryExternalIncentiveGaugesResponse{},
+		},
+	}
```
ValarDragon, line 219: Ditto for CLI command test looks incomplete

x/pool-incentives/keeper/grpc_query.go
```diff
@@ -88,3 +90,27 @@ func (k Keeper) IncentivizedPools(ctx context.Context, _ *types.QueryIncentivize
 		IncentivizedPools: incentivizedPools,
 	}, nil
 }
+
+func (k Keeper) ExternalIncentiveGauges(ctx context.Context, req *types.QueryExternalIncentiveGaugesRequest) (*types.QueryExternalIncentiveGaugesResponse, error) {
```
ValarDragon, line 94: Can you add godoc comments for this function, and some comments inside the function to explain it?

x/pool-incentives/keeper/grpc_query.go
```diff
@@ -88,3 +90,27 @@ func (k Keeper) IncentivizedPools(ctx context.Context, _ *types.QueryIncentivize
...
+		poolGaugeIds[sdk.BigEndianToUint64(iterator.Value())] = true
+	}
+
+	allGauges := k.GetAllGauges(sdkCtx)
+	gauges := []incentivetypes.Gauge{}
+	for _, gauge := range allGauges {
+		if _, ok := poolGaugeIds[gauge.Id]; !ok {
+			gauges = append(gauges, gauge)
+		}
+	}
```
ValarDragon, line 113: This function as written will be iterating over all gauges, which is ~ 1500 state entries. I have a suspition that this may cause issues, but lets test.

Seems like the way to fix this would be in the future to index ActiveGauges separately, where Active was taken to me Active in the current sense, and having non-zero rewards. Is there a way to do this right now? I'm not sure. If not, we can merge the PR without this fix, and then flag this as something to fix in a separate issue.

x/incentives/client/cli/query.go
```diff
@@ -234,6 +235,47 @@ $ %s query incentives active-gauges
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+			queryClient := types.NewQueryClient(clientCtx)
+			pageReq, err := client.ReadPageRequest(cmd.Flags())
+			if err != nil {
+				return err
+			}
```
ValarDragon, line 262: Just flagging this, we should make a helper method somewhere in our repo that we can use to do all of these in one line, to reduce our boilerplate. We should probably make a utils folder in our codebase somewhere, and make an issue in the SDK.

Gotta eliminate this boilerplate from the SDK :/

(Doesn't block this PR at all)

x/incentives/keeper/gauge.go
mattverse: yep, I think #507 is good to be closed! 
Actually let me make a different PR to migrate the method into grpc_query.go and fix other methods with this as well!

x/incentives/client/cli/cli_test.go
```diff
@@ -206,6 +206,40 @@ func (s *IntegrationTestSuite) TestGetCmdActiveGauges() {
...
+	testCases := []struct {
+		name      string
+		expectErr bool
+		respType  proto.Message
+	}{
+		{
+			"query active gauges per denom",
+			false, &types.ActiveGaugesPerDenomResponse{},
+		},
+	}
```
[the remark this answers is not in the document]
mattverse: It's not WIP, I've noticed that the other cli_test.go in any other modules are doing the same thing: only testing if it's getting the right type of response from the query.
Planning to get all these fix at once for all the cli_test.go in one PR if that would be ok!

x/pool-incentives/client/cli/cli_test.go
mattverse: Same for this comment, I've noticed that the other cli_test.go in any other modules are doing the same thing: only testing if it's getting the right type of response from the query.
Planning to get all these fix at once for all the cli_test.go in one PR if that would be ok!

x/incentives/client/cli/query.go
mattverse: Yep, totally agreed, should we put this issue as a backlog?

x/pool-incentives/keeper/grpc_query.go
Thunnini: @ValarDragon There seems to be no other way now. I think we try it and fix it if there is a problem in future.

x/incentives/keeper/grpc_query.go
mattverse: Nope there isn't

mattverse: Which, I think would be better to make instead of making this part of the code dangle in `grpc_query.go`. Let me methodize it and move it to keeper

mattverse: There actually is, let me replace this to the method

x/pool-incentives/keeper/grpc_query.go
mattverse: Done in https://github.com/osmosis-labs/osmosis/pull/572/commits/cbabd871a117231d1e3db4a0257d46d555c098de.

x/incentives/keeper/grpc_query.go
mattverse: Done in https://github.com/osmosis-labs/osmosis/pull/572/commits/cbabd871a117231d1e3db4a0257d46d555c098de.

x/incentives/keeper/grpc_query.go
mattverse: Done in https://github.com/osmosis-labs/osmosis/pull/572/commits/cbabd871a117231d1e3db4a0257d46d555c098de.

x/incentives/client/cli/query.go
ValarDragon: Yes please. Really, we should make an issue for it on the SDK. If you have bandwidth to do so, would be very appreciated.

x/incentives/client/cli/cli_test.go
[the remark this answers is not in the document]
ValarDragon: Ok sounds good. My view of this is atm we don't have CLI tests then, so should write an issue detailing the lack of them here with help wanted. (I dont think adding CLI tests is very high priority at the moment)

ValarDragon (approved):
LGTM! Lets be sure to make follow-up issues for the relevant items.

PR #584, opened 2021-11-15
Token Factory module

sunnya97:
The token factory module allows any account to start minting coins.  Denoms are namespaced by the coin creator.

## Description

______

For contributor use:


app/app.go
```diff
@@ -284,6 +290,7 @@ func NewOsmosisApp(
 		gammtypes.StoreKey, lockuptypes.StoreKey, claimtypes.StoreKey, incentivestypes.StoreKey,
 		epochstypes.StoreKey, poolincentivestypes.StoreKey, authzkeeper.StoreKey, txfeestypes.StoreKey,
 		bech32ibctypes.StoreKey,
+		tokenfactorytypes.StoreKey,
```
ValarDragon, line 293: We need to make a github issue to write the migration logic to initialize the store key
(Issue creation blocks merge, solving it does not)

sunnya97: #689

x/tokenfactory/keeper/keeper.go
```diff
@@ -0,0 +1,69 @@
...
+	store := ctx.KVStore(k.storeKey)
+	return prefix.NewStore(store, types.GetCreatorsPrefix())
+}
+
+// func (k Keeper) GetAdminPrefixStore(ctx sdk.Context, admin string) sdk.KVStore {
+// 	store := ctx.KVStore(k.storeKey)
+// 	return prefix.NewStore(store, types.GetAdminPrefix(admin))
+// }
+
+// CreateModuleAccount creates a module account with minting and burning capabilities
```
ValarDragon, line 65:
```suggestion
// CreateModuleAccount creates a module account with minting and burning capabilities
// This account isn't intended to store any coins, 
// it purely mints and burns them on behalf of the admin of respective denoms, 
// and sends to the relevant address.
```

x/tokenfactory/keeper/msg_server.go
```diff
@@ -0,0 +1,159 @@
...
+
+// NewMsgServerImpl returns an implementation of the MsgServer interface
+// for the provided Keeper.
+func NewMsgServerImpl(keeper Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+var _ types.MsgServer = msgServer{}
+
+func (server msgServer) CreateDenom(goCtx context.Context, msg *types.MsgCreateDenom) (*types.MsgCreateDenomResponse, error) {
```
ValarDragon, line 23: (Doesn't block merge)

Should we add a governance tunable osmo cost for minting denoms?

x/tokenfactory/keeper/msg_server.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	authorityMetadata, err := server.Keeper.GetAuthorityMetadata(ctx, msg.Amount.GetDenom())
+	if err != nil {
+		return nil, err
+	}
+
+	if msg.Sender != authorityMetadata.GetAdmin() {
+		return nil, types.ErrUnauthorized
+	}
```
ValarDragon, line 111: Can we refactor this into a `server.EnsureAuthorized(ctx, msg.Sender, msg.Amount.Denom())` method?

x/tokenfactory/module.go
```diff
@@ -0,0 +1,201 @@
...
+// RandomizedParams creates randomized txfees param changes for the simulator.
+func (am AppModule) RandomizedParams(r *rand.Rand) []simtypes.ParamChange {
+	return nil
+}
+
+// RegisterStoreDecoder registers a decoder for supply module's types
+func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
+}
+
+// WeightedOperations returns the all the gov module operations with their respective weights.
```
ValarDragon, line 198:
```suggestion
// WeightedOperations returns simulator module operations with their respective weights.
```

x/tokenfactory/types/denoms.go
```diff
@@ -0,0 +1,43 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+const (
+	ModuleDenomPrefix = "factory"
+)
+
+func GetTokenDenom(creator, nonce string) (string, error) {
+	denom := strings.Join([]string{"factory", creator, nonce}, "/")
```
ValarDragon, line 15:
```suggestion
	denom := strings.Join([]string{ModuleDenomPrefix, creator, nonce}, "/")
```

x/tokenfactory/types/denoms.go
```diff
@@ -0,0 +1,43 @@
...
+		return "", "", sdkerrors.Wrapf(ErrInvalidDenom, "denom prefix is incorrect. Is: %s.  Should be: %s", strParts[0], ModuleDenomPrefix)
+	}
+
+	creator = strParts[1]
+	_, err = sdk.AccAddressFromBech32(creator)
+	if err != nil {
+		return "", "", sdkerrors.Wrapf(ErrInvalidDenom, "Invalid creator address (%s)", err)
+	}
+
+	nonce = strings.Join(strParts[2:], "/")
```
ValarDragon, line 40: Why are we making this end in a "/"?

x/tokenfactory/types/denoms.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func DeconstructDenom(denom string) (creator string, nonce string, err error) {
+	err = sdk.ValidateDenom(denom)
+	if err != nil {
+		return "", "", err
+	}
+
+	strParts := strings.Split(denom, "/")
+	if len(strParts) < 3 {
```
ValarDragon, line 26: Why are we allowing more slashes? Is it to allow tokens to be able to be sub-indexed?

x/tokenfactory/types/keys.go
```diff
@@ -0,0 +1,45 @@
...
+	CreatorPrefixKey          = "creator"
+	AdminPrefixKey            = "admin"
+)
+
+func GetDenomPrefixStore(denom string) []byte {
+	return []byte(strings.Join([]string{DenomsPrefixKey, denom, ""}, "|"))
+}
+
+func GetCreatorPrefix(creator string) []byte {
+	return []byte(strings.Join([]string{CreatorPrefixKey, creator, ""}, "|"))
```
ValarDragon, line 36: Lets refactor "|" into a "key seperator" constnat here. (This is different than the state separator everywhere else in the SDK AFAIR)

ValarDragon (commented):
LGTM on first pass! Will do a second pass soon, then should be good to merge with a few follow-on issues.

proto/osmosis/tokenfactory/v1beta1/authorityMetadata.proto
```diff
@@ -0,0 +1,13 @@
...
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/osmosis-labs/osmosis/x/tokenfactory/types";
+
+message DenomAuthorityMetadata {
+  option (gogoproto.equal) = true;
+
+  string Admin = 1 [ (gogoproto.moretags) = "yaml:\"admin\"" ];
```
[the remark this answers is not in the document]
sunnya97: Wdym?

proto/osmosis/tokenfactory/v1beta1/genesis.proto
```diff
@@ -0,0 +1,27 @@
...
+message GenesisState {
+  option (gogoproto.equal) = true;
+
+  repeated GenesisDenom factory_denoms = 1 [
+    (gogoproto.moretags) = "yaml:\"factory_denoms\"",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message GenesisDenom {
```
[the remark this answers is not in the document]
sunnya97: Wdym?  This is just a struct that pairs a denom with its authority metadata, so that it can be included in a genesis file.  I don't think this merged struct is used anywhere else.

proto/osmosis/tokenfactory/v1beta1/authorityMetadata.proto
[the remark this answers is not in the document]
ValarDragon: Empty for no admin, or osmo address

x/tokenfactory/types/denoms.go
sunnya97: Yeah

sunnya97: Or like imagine I want to issue a centralized bridged version of an asset on another chain whose denom name has a slash in it, then maybe I want my token name to be:
`factory/sunnyaddr/ibc/asdfasdf`

x/tokenfactory/keeper/msg_server.go
sunnya97: I thought that we should add it for spam resistance, but then realized that we already have a non-spam resistant way of creating denoms... the ibc-transfer module.

Although arguably, that should also require a fee perhaps.

proto/osmosis/tokenfactory/v1beta1/genesis.proto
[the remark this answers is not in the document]
ValarDragon: Ah yeah, was looking at this at the beginning of review.

In state, it stores authority_metadata at a key for the denom. I was thinking we would be storing the entire struct as the value for simplicity, rn we just store Authority_metadata. (Which is more efficient, just means we have to deal with a separate genesis type)

I can't wait for the ORM module, kinda ridiculous we have to basically maintain our own indexing by every entry. (Ideally we'd just have a single struct, and just specify what we want to index on, ORM does this)

This is good

x/tokenfactory/types/denoms.go
sunnya97: It wouldn't end with a "/"

This is for if there were more than 2 slashes, when we did the split, we'd turn `factory/sunnyaddr/atomderivative/sikka` into ["factory", "sunnyaddr", "atomderivative", "sikka"]

To get the nonce we have to join [2:] with a "/" as the delimiter to get back the correct nonce which should be "atomderivative/sikka"

x/tokenfactory/keeper/msg_server.go
sunnya97: Should be for a specific action too, right?

sunnya97:
I feel we should move this authority metadata system to the bank module itself.  @ValarDragon wdyt?

ValarDragon:
I think mint/burn permissions for a token do generally need to be rethought much more, and get into bank natively. I'm not fully sure how to square different approaches to this. This approach feels too restrictive for many general tokens (e.g. osmo, where we want many different modules to get mint perms -- Superfluid is in this set). 

I was suggesting this approach as a non-state breaking approach to add more protections https://github.com/cosmos/cosmos-sdk/issues/10386 . I think something cool would be to get this module to emit a "FactoryMintRestrictions" and "FactoryBurnRestrictions" functions, that we then compose into that approach.

(We should also use that approach to restrict factory module perms)

The status of that is its in a PR thats imo good to merge pending spec update. (https://github.com/cosmos/cosmos-sdk/pull/10771) I think we should go ahead and just use that even if the SDK doesn't want to. (It solves having to worry about gamm module minting non-LP shares for instance, and is imo critical for superfluid)

faddat:
I really like how this approach doesn't require the use of a contract per token, which is a broken pattern.

x/tokenfactory/keeper/msg_server.go
ValarDragon: Yeah, I'd really prefer to add it in. (I think we should add it in there once we figure out a sensible way for there to be a fee!) 

IBC has a notable annoyance overhead, this will be trivial spam.

x/tokenfactory/keeper/msg_server.go
ValarDragon: I think we should add that method for now, and when we do de-activations or more complex structures, we make a fourth param for "permission type"

x/tokenfactory/types/denoms.go
ValarDragon: ah gotcha., makes sense! Can you add a comment to this affect?

x/tokenfactory/types/denoms.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const (
+	ModuleDenomPrefix = "factory"
+)
+
+func GetTokenDenom(creator, nonce string) (string, error) {
+	denom := strings.Join([]string{ModuleDenomPrefix, creator, nonce}, "/")
+	return denom, sdk.ValidateDenom(denom)
+}
```
ValarDragon, line 17: I think we should have a safety assert somewhere that `creator` has no slashes in it. Otherwise there is a malleability attack between creator and nonce.

sunnya97:
Rebased this to main, as well as
1. Removed ForceTransfers
2. Only allowed Mint and Burns to/from the admin address

sunnya97:
Closing because has been replaced by other PRs

PR #587, opened 2021-11-16
Superfluid minimum viable version implementation

antstalepresh:
Closes: #XXX

## Description

______

For contributor use:


proto/osmosis/lockup/lock.proto
```diff
@@ -55,3 +55,14 @@ message QueryCondition {
...
+
+// SyntheticLock is a single unit of synthetic lockup
+message SyntheticLock {
+  uint64 lock_id = 1;
+  string suffix = 2;
+  google.protobuf.Timestamp end_time = 3 [
+    (gogoproto.stdtime) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"end_time\""
+  ];
```
ValarDragon, line 67: Shouldn't this be duration?

proto/osmosis/superfluid/genesis.proto
```diff
@@ -7,9 +7,22 @@ import "osmosis/superfluid/params.proto";
 
 option go_package = "github.com/osmosis-labs/osmosis/x/superfluid/types";
 
+message EpochOsmoEquivalentTWAP {
+  int64 epoch = 1;
```
ValarDragon, line 11:
```suggestion
  int64 epoch_number = 1;
```

proto/osmosis/superfluid/superfluid.proto
```diff
@@ -25,17 +25,10 @@ message SuperfluidAsset {
...
-// SuperfluidAssetInfo stores the information of superfluid asset - real time changes
-message SuperfluidAssetInfo {
+// SuperfluidIntermediaryAccount takes the role of intermediary between LP token and OSMO tokens for superfluid staking
+message SuperfluidIntermediaryAccount {
   string denom = 1;
-  string total_staked_amount = 2 [ 
-    (gogoproto.moretags) = "yaml:\"total_staked_amount\"",
-    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
-    (gogoproto.nullable) = false
-  ];
```
ValarDragon, line 35: Should we keep this, or is it put in state somewhere? Seems useful to know, though I guess we could push that to an off-chain db

x/lockup/keeper/lock.go
```diff
@@ -73,7 +76,36 @@ func (k Keeper) deleteLockRefs(ctx sdk.Context, lockRefPrefix []byte, lock types
 	return nil
 }
 
+func (k Keeper) addSyntheticLockRefs(ctx sdk.Context, lockRefPrefix []byte, lock types.PeriodLock) error {
```
ValarDragon, line 79: Not really an actionable comment, but I'm generally a bit worried about the complexity growing of all our separate lock indices. Is there a way we can simplify / re-use a lot of the logic?

x/lockup/keeper/lock.go
```diff
@@ -313,6 +345,22 @@ func (k Keeper) addTokensToLock(ctx sdk.Context, lock *types.PeriodLock, coins s
 	return nil
 }
 
+func (k Keeper) removeTokensFromLock(ctx sdk.Context, lock *types.PeriodLock, coins sdk.Coins) error {
```
ValarDragon, line 348: Lets add a comment here, and some descriptions of where its ok to be called from?

x/lockup/keeper/lock.go
```diff
@@ -435,6 +509,88 @@ func (k Keeper) ResetAllLocks(ctx sdk.Context, locks []types.PeriodLock) error {
 	return nil
 }
 
+func (k Keeper) ResetAllSyntheticLocks(ctx sdk.Context, syntheticLocks []types.SyntheticLock) error {
```
ValarDragon, line 512: Can we re-use any of this logic with normal Locks?

x/lockup/keeper/synthetic_lock_test.go
```diff
@@ -0,0 +1,311 @@
...
+	suite.Require().Len(locks, 0)
+	locks = suite.app.LockupKeeper.GetAccountLockedDurationNotUnlockingOnly(suite.ctx, addr1, "stakestakedtovalidator1", time.Second)
+	suite.Require().Len(locks, 0)
+	locks = suite.app.LockupKeeper.GetAccountLockedLongerDurationDenom(suite.ctx, addr1, "stakestakedtovalidator1", time.Second)
+	suite.Require().Len(locks, 0)
+	locks = suite.app.LockupKeeper.GetLocksPastTimeDenom(suite.ctx, "stakestakedtovalidator1", suite.ctx.BlockTime())
+	suite.Require().Len(locks, 0)
+	locks = suite.app.LockupKeeper.GetLocksLongerThanDurationDenom(suite.ctx, "stakestakedtovalidator1", time.Second)
+	suite.Require().Len(locks, 0)
+	amount = suite.app.LockupKeeper.GetLockedDenom(suite.ctx, "stakestakedtovalidator1", time.Second)
```
ValarDragon, line 57: Can we mitigate a lot of this code repetition in the tests via making other functions, with a struct for all the inputs, and a struct for all the expected answers?

ValarDragon (commented):
some initial comments, need to transcribe more of comments from slack! Apologies for delay in review, v5 upgrade delayed it 😅 

Nice work on the PR structure, will review more tmrw.

x/superfluid/keeper/distribution.go
```diff
@@ -0,0 +1,30 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (k Keeper) MoveIntermediaryDelegationRewardToGauges(ctx sdk.Context) {
```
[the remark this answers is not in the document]
antstalepresh: This looks better idea - I will make the function name changes along with tests

proto/osmosis/lockup/lock.proto
antstalepresh: Synthetic lockup is managed by external module and synthetic lockup's duration is different from native lockup.
Here end time is when synthetic lock up is deleted automatically from the store. And this end time is set on superfluid module as current time + 3 weeks unbonding time - users could try to do superfluid staking for his 100 year lockup.

proto/osmosis/superfluid/superfluid.proto
antstalepresh: Total staked amount can be queried by queries - I didn't find any use case for total staked amount to be stored on chain and removed for simplificaiton.

x/lockup/keeper/lock.go
antstalepresh: I was thinking a lot but we need different entries list for synthetic lockups to prevent conflict with native lockups.

x/lockup/keeper/lock.go
antstalepresh: Okay

x/lockup/keeper/lock.go
antstalepresh: Synthetic locks are using different entries to avoid conflicts with native lockups and I think we will need to keep this initialization logic separately from native lockups.

x/lockup/keeper/synthetic_lock_test.go
antstalepresh: That way, code can be reduced but it won't be easy to read the code while reviewing tests.

proto/osmosis/lockup/lock.proto
ValarDragon: I think this needs more comments here to explain it then.

x/lockup/abci.go
```diff
@@ -20,5 +20,6 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) []abci.ValidatorUpdate {
 		return []abci.ValidatorUpdate{}
 	}
 	k.WithdrawAllMaturedLocks(ctx)
+	k.DeleteAllMaturedSyntheticLocks(ctx)
```
ValarDragon, line 23: Does it make sense to have `DeleteAllMaturedSyntheticLocks` just be called within `WithdrawAllMaturedLocks`?

x/lockup/keeper/lock.go
ValarDragon: I think we can re-use all of the accumulation store logic at least. Doesn't need to block this PR though, we can make this into a separate issue and do it later.

x/lockup/keeper/synthetic_lock.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		synthLocks = append(synthLocks, synthLock)
+	}
+	return synthLocks
+}
+
+// CreateSyntheticLockup create synthetic lockup with lock id and suffix
+func (k Keeper) CreateSyntheticLockup(ctx sdk.Context, lockID uint64, suffix string, unlockDuration time.Duration) error {
+	// Note: synthetic lockup is doing everything same as lockup except coin movement
+	// There is no relationship between unbonding and bonding synthetic lockup, it's managed separately
```
ValarDragon, line 86: Can you explain this more somewhere?

x/lockup/keeper/synthetic_lock_test.go
ValarDragon: err, I don't think this style of tests is that easy to read, review, or give the guarantees we want. It ends up being very hard to adjust in the future / require rewriting. (Doesn't block this PR merge though)

proto/osmosis/lockup/lock.proto
ValarDragon: If its supposed to be the case that end time is optional, can there be a boolean for "active"

proto/osmosis/lockup/lock.proto
```diff
@@ -55,3 +55,14 @@ message QueryCondition {
     (gogoproto.moretags) = "yaml:\"timestamp\""
   ];
 }
+
+// SyntheticLock is a single unit of synthetic lockup
+message SyntheticLock {
+  uint64 lock_id = 1;
```
ValarDragon, line 61: Can this be 'underlying_lock_id' perhaps?

proto/osmosis/lockup/lock.proto
```diff
@@ -55,3 +55,14 @@ message QueryCondition {
...
+  string suffix = 2;
```
ValarDragon, line 62: Is this the name of the synthetic lock denom?

proto/osmosis/lockup/lock.proto
antstalepresh: For active one, end_time is set to time.Time{} - this kind of mechanism is working on native lockups as well. Yeah here end time is only set for unbonding synthetic lockups. For delegated lockups, it's bonding synthetic lockups.
I will add a bit of comments for end_time.

x/lockup/keeper/lock.go
antstalepresh: Ok

x/lockup/abci.go
antstalepresh: I was thinking of splitting the native locks and synthetic locks in the same level so that native locks does not care about synthetic locks.

x/lockup/keeper/synthetic_lock.go
antstalepresh: On spec document, would be adding few more details for better understanding

proto/osmosis/lockup/lock.proto
antstalepresh: Okay, will update to underlying_lock_id

proto/osmosis/lockup/lock.proto
antstalepresh: Yes, it's what comes after native lockup's denom.

proto/osmosis/lockup/lock.proto
antstalepresh: Updated

proto/osmosis/lockup/lock.proto
antstalepresh: Added comment for suffix field on proto definition

proto/osmosis/lockup/lock.proto
ValarDragon: Hrmm, seems ok for now then, but we should consider changing to be different structs prior to it going live. Lets file this into an issue pre-merge.

ValarDragon:
Per call: Advocating for just merging this PR. Lets write a list of the breaking changes to existing modules at least, so we can know what to double check to ensure nothing existing is broken

x/lockup/abci.go
```diff
@@ -19,6 +19,14 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) []abci.ValidatorUpdate {
...
+
+	// delete synthetic locks matured before lockup deletion
+	k.DeleteAllMaturedSyntheticLocks(ctx)
+
+	// withdraw and delete locks
 	k.WithdrawAllMaturedLocks(ctx)
 	return []abci.ValidatorUpdate{}
 }
+
+// TODO: add invariant that no native lockup existent synthetic lockup exists by calling GetAllSyntheticLockups
```
antstalepresh, line 31: Rather than panic for this case, preventing deletion of native lockups with synthetic lockup would be better and let the tokens on lockup freeze.

PR #597, opened 2021-11-19
Update README.md versions

faddat:
## Description

Made the Readme file copy-pastable again.

______

For contributor use:

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This is what you'd use to get a node up and running, fast. It assumes that it is
 
 **Install go**
 ```bash
-wget -q -O - https://git.io/vQhTU | bash -s -- --version 1.17.2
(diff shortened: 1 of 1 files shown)


sunnya97 (dismissed):

ValarDragon:
This should be v3.2.0, as its not using statesync or quicksync. Therefore you need v3.2.0 to sync from genesis.

We should make the default script download from quicksync, then we can use v4.x as well

faddat:
ok.  Very interesting.  Now I need to figure out what quicksync will do.

ValarDragon:
Going to close this for now. Though Chainlayer made a command/ public data so we can actually fully automate their setup. (The only non-automatic part before was getting the filename. We can now get it with `curl https://quicksync.io/osmosis.json|jq -r '.[] |select(.network=="default")|select (.mirror=="Netherlands")|.filename'`)

Issue #591, opened 2021-11-18
Backport v4.2.0 fixes to main


daniel-farina:
Description: Backport v4.2.0 fixes to main

PR #599, opened 2021-11-19
Backport v4.2.0 updates to main

ValarDragon:
Closes: #591 

## Description

Backports the v4.2.0 fixes to main. I simplified the NewMempoolMaxGasPerTxDecorator decorator substantially, since the fees have been relegated to our own separate ante-handler.

The rest of the changes are essentially direct cherry-picks.

______

For contributor use:

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -2,6 +2,8 @@ package app
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
 	"github.com/cosmos/cosmos-sdk/x/auth/signing"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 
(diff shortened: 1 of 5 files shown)


ValarDragon:
Note: This PR should be rebased and merged, not squashed.

mattverse (commented):
One question I have is about NewMempoolMaxGasPerTxDecorator, wouldn't it be better to user params / gov to decide maximum gas wanted instead of hardcoding it? 
Do we have future plans on changing this or would this be a fixed way that we would manage max gas in the future as well?

ValarDragon:
This is in CheckTx, not consensus. (So it only impacts whether a tx gets into the mempool) So every node can just change this locally.

I think we should expose this via a config, but we can do that in a state-compatible update later. (As that decorator doesn't affect consensus logic at all!)

mattverse (approved):
Cool! Otherwise looks LGTM!

PR #610, opened 2021-11-21
Update app.go to correctly do module upgrade logic, add authz, IBC v2

ValarDragon:
Update app.go to correctly do SDK module upgrade logic. This follows after what was done in gaia. https://github.com/cosmos/gaia/compare/v5.0.8...v6.0.0-rc3

The line for adding the IBC channel is added, to make the SDK v0.42 -> IBC v1.2.3 update safe.

This also adds authz into the `v5` upgrade.
This also adds the IBC v2 upgrade logic.

______

For contributor use:


app/app.go
```diff
@@ -340,18 +356,49 @@ func NewOsmosisApp(
...
+
+			// Set all modules "old versions" to 1.
+			// Then the run migrations logic will handle running their upgrade logics
+			fromVM := make(map[string]uint64)
+			for moduleName := range app.mm.Modules {
+				fromVM[moduleName] = 1
+			}
+			// override versions for _new_ modules as to not skip InitGenesis
+			fromVM[authz.ModuleName] = 0
+			fromVM[txfees.ModuleName] = 0
```
ValarDragon, line 383: I think these need to be deleted from the map from looking at the code for RunMigrations: https://github.com/cosmos/cosmos-sdk/blob/994219a4d2f467a0bd628d60d8670c1102edbe88/types/module/module.go#L396-L444 . I'm not sure how gaia's upgrade / default genesis setting works. Maybe all of their modules need no state initialization?

As it stands, InitGenesis won't be ran.

Also, they both get ran with their DefaultGenesis -- we probably need to change that.

I think we have to do `RunMigrations`, then override their params as we want.

ValarDragon: Fixed in latest commit

sunnya97 (approved):

PR #612, opened 2021-11-21
misc test improvements

ValarDragon:
This PR contains the following misc test improvements:
* x/gamm: Deduplicate the FundAccounts coins constant
* x/gamm: Fix staticcheck being angry about unnecessary fmt.Sprintf's
* x/lockup: Fix noise from FundAccount in gas_test
* osmotestutils: Add GetFeeString function
* Replace complex fee cli line with simple osmotestutils call

## Description

This osmotestutils should accrue more improvements to testing here. The latency of trying to get improvements into the SDK, then back onto our own repo is far too high.

______

For contributor use:

### DIFF
--- a/x/gamm/client/cli/cli_test.go
+++ b/x/gamm/client/cli/cli_test.go
@@ -16,6 +16,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	banktestutil "github.com/cosmos/cosmos-sdk/x/bank/client/testutil"
 	"github.com/osmosis-labs/osmosis/app"
+	"github.com/osmosis-labs/osmosis/osmotestutils"
 	"github.com/osmosis-labs/osmosis/x/gamm/client/cli"
 	gammtestutil "github.com/osmosis-labs/osmosis/x/gamm/client/testutil"
 	"github.com/osmosis-labs/osmosis/x/gamm/types"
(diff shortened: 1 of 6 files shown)


x/gamm/client/cli/cli_test.go
```diff
@@ -78,7 +79,7 @@ func (s *IntegrationTestSuite) TestNewCreatePoolCmd() {
 		newAddr,
 		sdk.NewCoins(sdk.NewInt64Coin(s.cfg.BondDenom, 200000000), sdk.NewInt64Coin("node0token", 20000)), fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
 		fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
```
mattverse, line 81: Do we also want to deduplicate broadcast mode?

x/lockup/keeper/gas_test.go
```diff
@@ -13,8 +14,13 @@ var (
...
+	err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, addr, coins)
+	suite.Require().NoError(err)
+	// start measuring gas
 	alreadySpent := suite.ctx.GasMeter().GasConsumed()
-	suite.LockTokens(addr, coins, dur)
+	_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, addr, coins, dur)
+	suite.Require().NoError(err)
 	newSpent := suite.ctx.GasMeter().GasConsumed()
 	spentNow := newSpent - alreadySpent
 	return spentNow
```
mattverse, line 26: Good catch on this! 
One further thing we can do in the future is set a range and simply test if the avgGas and maxGas is within the range, or something similar to remove hard coding gas, considering that there could be further instances where we would have to manually change the hard-coded gas whenever there is a logic change within the method.

mattverse (approved):
Looks LGTM!

x/gamm/client/cli/cli_test.go
ValarDragon: thats a good idea, though I wonder if theres a better way we can just get more of these common flags to all be handled, to better mitigate the boilerplate. (E.g. Just appending all the common flags to the args array)

We should probs open an issue for tracking how to improve this

x/lockup/keeper/gas_test.go
ValarDragon: hrmm, thats a good idea. (Basically lower the sensitivity to small noise that we care about)

Depends on our general attitude towards gas, where does tracking small changes matter. I tend to agree with you, at this stage, having small aberrations is totally fine, as long as we can bound the max, and that max and average don't get too far apart.

PR #618, opened 2021-11-23
Fix v44 auth upgrade bug

ValarDragon:
Workaround for https://github.com/cosmos/cosmos-sdk/issues/10591 , https://github.com/provenance-io/provenance/blob/407c89a7d73854515894161e1526f9623a94c368/app/upgrades.go#L86-L122

Just do auth upgrade last.

______

For contributor use:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -378,7 +378,14 @@ func NewOsmosisApp(
 			for moduleName := range app.mm.Modules {
 				fromVM[moduleName] = 1
 			}
-			// override versions for authz module as to not skip InitGenesis
+			// EXCEPT Auth needs to run _after_ staking (https://github.com/cosmos/cosmos-sdk/issues/10591),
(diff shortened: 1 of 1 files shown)


robert-zaremba:
This look as a dirty hack. I'm thinking about changing the map value type to a structure with priority.

robert-zaremba:
Let's continue the discussion in the SDK issue

ValarDragon:
Seeing as this is being tested on vega upgrade as well, merging

robert-zaremba:
sure, was only saying that it's feels hacky, but solves the problem. In 0.44 we don't have other solution which is not breaking :/ except the patch for x/auth.

So your one is the best so far. I have created a draft PR for 0.45 to change the structure.

ValarDragon:
ah 100% agreed, its absolutely a gross workaround lol

PR #629, opened 2021-11-29
Semver for the v3.x series of Osmosis

faddat:
This adds a /v3 import path to osmo and its modules so that devs can use it more easily.


ValarDragon:
Thanks for adding this 🙏

ValarDragon:
Looks like mint modules cli_test needs to be updated. Once tests pass, this is good to merge imo

faddat:
was a little more than that.  We hit a conflict on sdk.error

ValarDragon (approved):
Amazing, tysm for adding this in!

PR #643, opened 2021-12-08
Prune go.sum

faddat:
## Description

To explore preformance issues more deeply, we need to prune go.sum as much as possible.  This represents a good first start, but is in no way complete.  Since when do cosmos projects need four different versions of btcd, for example?

Sure, likely they need **one**.... but I reckon that we've got **four** may be causing trouble. 

______

For contributor use:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -122,9 +122,10 @@ replace (
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211119062100-2f81f59f3aa1
 	github.com/cosmos/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
 	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
-	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db => github.com/osmosis-labs/tm-db v0.6.5-0.20210911033928-ba9154613417
+	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
 )
--- a/go.mod
+++ b/go.mod
@@ -3,10 +3,10 @@ module github.com/osmosis-labs/osmosis
 go 1.17
 
 require (
-	github.com/cosmos/cosmos-sdk v0.44.3
+	github.com/cosmos/cosmos-sdk v0.44.5
 	github.com/cosmos/go-bip39 v1.0.0
-	github.com/cosmos/iavl v0.17.2
-	github.com/cosmos/ibc-go/v2 v2.0.0
+	github.com/cosmos/iavl v0.17.3
+	github.com/cosmos/ibc-go/v2 v2.0.1
 	github.com/gogo/protobuf v1.3.3
 	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/mux v1.8.0
@@ -120,6 +120,9 @@ require (
 
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211119062100-2f81f59f3aa1
+	github.com/cosmos/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
+	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
+	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db => github.com/osmosis-labs/tm-db v0.6.5-0.20210911033928-ba9154613417
--- a/go.mod
+++ b/go.mod
@@ -121,11 +121,9 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211119062100-2f81f59f3aa1
 	github.com/cosmos/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
-	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db => github.com/osmosis-labs/tm-db v0.6.5-0.20210911033928-ba9154613417
-	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
 )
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -175,13 +175,14 @@ github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7
 github.com/coreos/go-systemd/v22 v22.3.2/go.mod h1:Y58oyj3AT4RCenI/lSvhwexgC+NSVTIJ3seZv2GcEnc=
 github.com/coreos/pkg v0.0.0-20160727233714-3ac0863d7acf/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
 github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
-github.com/cosmos/go-bip39 v0.0.0-20180819234021-555e2067c45d/go.mod h1:tSxLoYXyBmiFeKpvmq4dzayMdCjCnu8uqmCysIGBT2Y=
 github.com/cosmos/go-bip39 v1.0.0 h1:pcomnQdrdH22njcAatO0yWojsUnCO3y2tNoV1cb6hHY=
 github.com/cosmos/go-bip39 v1.0.0/go.mod h1:RNJv0H/pOIVgxw6KS7QeX2a0Uo0aKUlfhZ4xuwvCdJw=
-github.com/cosmos/iavl v0.17.2 h1:BT2u7DUvLLB+RYz9RItn/8n7Bt5xe5rj8QRTkk/PQU0=
-github.com/cosmos/iavl v0.17.2/go.mod h1:prJoErZFABYZGDHka1R6Oay4z9PrNeFFiMKHDAMOi4w=
```
faddat, line 182: this

go.sum
```diff
@@ -832,8 +833,6 @@ github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9dec
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
 github.com/zondax/hid v0.9.0 h1:eiT3P6vNxAEVxXMw66eZUAAnU2zD33JBkfG/EnfAKl8=
 github.com/zondax/hid v0.9.0/go.mod h1:l5wttcP0jwtdLjqjMMWFVEE7d1zO0jvSPA9OPZxWpEM=
-go.etcd.io/bbolt v1.3.2/go.mod h1:IbVyRI1SCnLcuJnV2u8VeU0CEYM7e686BmAb1XKL+uU=
```
faddat, line 835: this

go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/osmosis-labs/osmosis
 go 1.17
 
 require (
-	github.com/cosmos/cosmos-sdk v0.44.3
+	github.com/cosmos/cosmos-sdk v0.44.5
```
ValarDragon, line 6: Lets not bump this, we're really importing v0.44.3 with our osmosis commits

go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/osmosis-labs/osmosis
...
 	github.com/cosmos/go-bip39 v1.0.0
-	github.com/cosmos/iavl v0.17.2
-	github.com/cosmos/ibc-go/v2 v2.0.0
+	github.com/cosmos/iavl v0.17.3
+	github.com/cosmos/ibc-go/v2 v2.0.1
```
ValarDragon, line 9: Not really a fan of bumping minor versions unless we have a reason to do so, or have looked carefully into changes at hand. Its more things to think about for non-determinism / surface area for incompatibility between versions.

In the ibc-go case, seems fine to me. (Looked into it due to this commit)

In IAVL, not sure. Would prefer to not be first to test it, sounds liable to cause slowdowns for parallel queries. (And i think theres a better fix for the bug at hand)

go.mod
```diff
@@ -120,8 +120,12 @@ require (
 
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211119062100-2f81f59f3aa1
+	github.com/cosmos/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
```
ValarDragon, line 123: I don't really grok why we had a bip39 issue, especailly if its past 1.0. The bip39 library authors should be guaranteeing API compatibility.

go.mod
```diff
@@ -120,8 +120,12 @@ require (
...
+	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db => github.com/osmosis-labs/tm-db v0.6.5-0.20210911033928-ba9154613417
+	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
```
ValarDragon, line 129: The bbolt stuff is odd, but it all only differs in patch version. I don;t want to add maintainence burden on us to check every time we update a dependency if it for some reason depends on a new feature in next minor bump, and we get a subtle error due to this replace.

go.mod
```diff
@@ -120,8 +120,12 @@ require (
 
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211119062100-2f81f59f3aa1
+	github.com/cosmos/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
+	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
```
ValarDragon, line 126: ditto for my comment re: bbolt. Do we have to upgrade this every patch version of TM/SDK/TM-db?

I'm more happy to do this if we decide to give more rocksdb support, but I want to be clear on why were getting multiple versions, and if we can do this replace in (say) our SDK fork, or a tm-db fork.

ValarDragon (changes requested):
Hey, thanks for looking into this in such depth!

Aside from the IBC-go bump to v2.0.1, I don't agree with the other changes here. I feel like this replace based approach adds untenable amounts of overheads / safety issues for updating any constituent library. I'm also unclear on where the problem stems from.

I think we do need replaces for SDK, Tendermint, and maybe SDK direct dependencies long term. This is due to our direct forking of an SDK version, and propensity to choose our own tendermint version, and need for compat. 

But the rest of the dependencies should be chosen by tendermint / tm-db alone imo. I'd be much more happy to do just a tm-db version replace, and then check tendermint & tm-db in tandem.

ValarDragon:
Also not sure I agree with the performance improvement aspect of this. (Not at all clear to me how this impacts performance, aside from potentially binary size)

I do agree with the dependency mitigation / security surface area reduction aspects of consideration with the imports / replaces though!

go.mod
faddat: 🙏

faddat:
Re: performance-- the claim isn't so much that this will improve performance, but instead that it'll allow for investigation into performance issues.

go.mod
faddat: It seems to me that they've resolved this issue in tm-db v0.6.6

go.sum
faddat: I think this is also resolved by tm-db 0.6.6

faddat:
@ValarDragon thanks so much for your review and responsiveness throughought my discovery process on this.  I summarized my findings in this thread of tweets:

https://twitter.com/gadikian/status/1469512833809678342

Conclusion:  We can actually close this PR and should work to get this:

https://github.com/cosmos/ibc-go/pull/611

merged asap, and we're all good.  I'll see if I can get our tm-db updated w/the changes from tendermint/tm-db so all of this is simply eliminated as a concern.

ValarDragon:
Sweet! Thanks for the work on investigating all of the dependencies

PR #649, opened 2021-12-10
Update errors.go: odd testing thing

faddat:
## Description

We've got sdk.Error 1100 assigned to both claim module errors and epoch module errors. I don't know why the build system doens't throw an error for this when building the main branch, but it doesn't.  Hence, I've made error 69420 the error for epochs, because when you need a number, 69420 has got you covered. 

______

For contributor use:

### DIFF
--- a/x/epochs/types/errors.go
+++ b/x/epochs/types/errors.go
@@ -8,5 +8,5 @@ import (
 
 // x/epochs module sentinel errors
 var (
-	ErrSample = sdkerrors.Register(ModuleName, 1100, "sample error")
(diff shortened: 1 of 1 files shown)


ValarDragon:
This is really odd, since the module names are distinct. Can you help me reproduce the error on main?

faddat:
@ValarDragon no, I can't. 

Isn't that wierd!?!?

ValarDragon:
Should we close this until we can reproduce? Would rather not change error codes, which is state machine & client breaking if we don't have to. (And I can't figure out an issue here, since the errors should be namespaced)

PR #650, opened 2021-12-10
Always use latest golangci-lint

faddat:
## Description

This will always use the latest golangci-lint, so we don't need to update that. 

______

For contributor use:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v2
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.39
+          version: latest
           args: --timeout 10m
           github-token: ${{ secrets.github_token }}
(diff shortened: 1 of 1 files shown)


ValarDragon:
Is this a good idea? Wouldn't it be worse if golangci had a breaking change, and then our ci broke, breaking workflows for a few days?

faddat:
heh, all I can say on this topic is that in my experience, it doesn't :).

faddat:
but that doesn't mean that it won't!

:upside_down_face:

faddat:
https://github.com/cosmos/gaia/runs/4489341951?check_suite_focus=true

ValarDragon:
I'd really prefer to have CI versions pinned for determinism, especially long term. Do we anticipate a lot changing between golangci versions?

faddat:
I anticipate linter deprecation, resulting in errors

faddat:
Screen Shot 2021-12-12 at 6 46 55 PM


They're warnings :).  Let's keep as is.

Issue #502, opened 2021-10-05
x/GAMM: Create a function to refund all LP shares for a pool


ValarDragon:
We should create a function in `x/gamm` that takes in a pool ID, and then dissolves that pool, and refunds the LP shares back into the constituent assets, and distributes to all holders. This would eventually become something callable in upgrades or governance, to dissolve pools with less than $X liquidity. (See https://gov.osmosis.zone/proposal/discussion/2047-proposal-raise-the-pool-creation-fee-and-refund-pools-with-less-than-1000-tvl/ for discussion from Osmosis community on why this is quite desirable!)

This would be done by iterating over all accounts looking for folks who hold that pools LP shares. If you hold that pools shares, then it burns those shares, and then gives you the constituent tokens back.

It then destroys the pool. 

If there are any LP shares IBC'd out of Osmosis, I think those should for now just be 'stuck', as a v1 of this feature. We can make a way to redeem them back subsequently.

daniel-farina:
Will be closed by #652

PR #652, opened 2021-12-10
Pool Cleanup

mconcat:
Closes: #502

## Description

______

For contributor use:


x/gamm/keeper/pool.go
```diff
@@ -79,6 +79,66 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+		return err
+	}
+
+	// sanity check
+	if !pool.GetTotalShares().IsZero() {
+		panic("pool total share should be zero after cleanup")
+	}
+
+	pool.SetDestruction(ctx.BlockTime())
+	k.SetPool(ctx, pool)
```
sunnya97, line 137:
```suggestion
	err = pool.SetDestruction(ctx.BlockTime())
	if err != nil {
		return err
	}
	err = k.SetPool(ctx, pool)
	if err != nil {
		return err
	}
```

ValarDragon:
Is the idea of setting a destruction time that I can make a pool get deleted in the current block (e.g. at upgrade), and schedule a deletion in the future?

Shouldn't we make it so that the pool deletion also deletes the pool from state, or is it a goal / low cost to keep the literal pool in state as long as its marked inactive for integrators?

x/gamm/keeper/pool.go
```diff
@@ -79,6 +79,72 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+	})
+
+	if err != nil {
+		return err
+	}
+
+	// sanity check
+	if !pool.GetTotalShares().IsZero() {
+		panic("pool total share should be zero after cleanup")
+	}
```
mattverse, line 134: Can zero share be assured everytime(division factor etc.)?
My only concern is that it might not: considering the fact that I remember that somewhere in the code what we did was iterate through coins and send a part of it every iteration, and at the last iteration, we would send the all the remainings.

mattverse:
Similar but different to @ValarDragon 's comment, I'm curious when this `CleanupBalancerPool` is designed to be ran

mattverse:
+ A friendly reminder to fix test!

x/gamm/keeper/pool.go
mconcat: Great point, I originally thought the total shares are only added/subtracted, not divided; I will check if there are any case of division.

mconcat:
I intended pool destruction in order to make it able to resurrected in the future, but I think we don't need to store the actual time here. However, I agree that it does not really make sense to just pause the pool in this use case, as we completely destruct the pool and refund the whole assets. I will change it to state removal.

> Is the idea of setting a destruction time that I can make a pool get deleted in the current block (e.g. at upgrade), and schedule a deletion in the future?
> 
> Shouldn't we make it so that the pool deletion also deletes the pool from state, or is it a goal / low cost to keep the literal pool in state as long as its marked inactive for integrators?

mconcat:
R4R

x/gamm/keeper/pool.go
mattverse: Is this fixed? / checked?

ValarDragon:
Notes from call earlier today/yesterday
            * The flow is:
                * DeletePool ->
                    * Force Unlock every LP share (`ForceUnlock`)
                    * Iterate through bank, and migrate every LP share to constituent assets
                    * Delete Pool data (via `ClearPool`)

x/gamm/keeper/pool.go
mconcat: Because shares are just a sdk.Coin, it could be IBC'd out or sent to module logics, and stay under the module account. I am pretty sure if we count all the address held shares and module held shares, it would sum up to GetTotalShares(). Let me add the logic for module accounts to(or, is it already iterated by GetAllBalances?)

x/gamm/keeper/pool.go
```diff
@@ -79,6 +79,80 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+	if err != nil {
+		return err
+	}
+
+	totalShares := pool.GetTotalShares().Amount
+	shareDenom := pool.GetTotalShares().Denom
+	poolAssets := pool.GetAllPoolAssets()
+
+	// first iterate through the share holders and burn them
+	k.bankKeeper.IterateAllBalances(ctx, func(addr sdk.AccAddress, coin sdk.Coin) (stop bool) {
```
ValarDragon, line 99: Lets file an issue to revisit performance of this. (Either we change to do one iteration for all pools, or we wait for https://github.com/cosmos/cosmos-sdk/pull/9611 to get merged into a release)

x/gamm/keeper/pool.go
```diff
@@ -79,6 +79,80 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+		if coin.Denom != shareDenom || coin.Amount.IsZero() {
+			return
+		}
+
+		shareAmount := coin.Amount
```
ValarDragon, line 104: lets alias coin to `userPoolShare` here. I think it makes a lot of subsequent lines easier to read

x/gamm/keeper/pool.go
```diff
@@ -79,6 +79,80 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+		}
+
+		err = k.bankKeeper.BurnCoins(ctx, types.ModuleName, sdk.Coins{coin})
+		if err != nil {
+			return true
+		}
+
+		// Refund assets
+		for _, asset := range poolAssets {
+			assetAmount := asset.Token.Amount.Mul(shareAmount).Quo(totalShares)
```
ValarDragon, line 121:
```suggestion
			// lpShareEquivalentTokens = (amount in pool) * (your shares) / (total shares)
			lpShareEquivalentTokens := asset.Token.Amount.Mul(shareAmount).Quo(totalShares)
```

x/gamm/keeper/pool.go
ValarDragon: Wait aren't we going to be iterating over IBC's module accounts like this? If so, we will be making IBC'd tokens never redeemable. Do we need to care about this, cc @sunnya97 ?

x/gamm/keeper/pool.go
ValarDragon: Can you investigate if module accounts are iterated over? I think we should definitely exclude certain module accounts, e.g. IBC

ValarDragon:
> I intended pool destruction in order to make it able to resurrected in the future, but I think we don't need to store the actual time here. However, I agree that it does not really make sense to just pause the pool in this use case, as we completely destruct the pool and refund the whole assets. I will change it to state removal.

This still needs to be done right? (I'm still seeing the destruction time logic)

x/lockup/keeper/lock.go
```diff
@@ -219,6 +219,12 @@ func (k Keeper) GetLocksPastTimeDenom(ctx sdk.Context, denom string, timestamp t
 	return combineLocks(notUnlockings, unlockings)
 }
 
+func (k Keeper) GetLocksDenom(ctx sdk.Context, denom string) []types.PeriodLock {
+	unlockings := k.getLocksFromIterator(ctx, k.LockIteratorDenom(ctx, true, denom))
+	notUnlockings := k.getLocksFromIterator(ctx, k.LockIteratorDenom(ctx, false, denom))
+	return combineLocks(notUnlockings, unlockings)
+}
```
ValarDragon, line 226: pretty surprised we don't already have a function for this lying around, pretty sure we use it in distribution

x/lockup/keeper/lock.go
```diff
@@ -560,3 +566,43 @@ func (k Keeper) Unlock(ctx sdk.Context, lock types.PeriodLock) error {
...
+		return err
+	}
+
+	// remove from accumulation store
+	for _, coin := range lock.Coins {
+		k.accumulationStore(ctx, coin.Denom).Decrease(accumulationKey(lock.Duration), coin.Amount)
+	}
+
+	k.hooks.OnTokenUnlocked(ctx, owner, lock.ID, lock.Coins, lock.Duration, lock.EndTime)
+	return nil
```
ValarDragon, line 606: Can we reuse logic from the existing methods for a lock going from NotUnlocking -> Unlocking, and Unlocking -> finish unlocking?

x/lockup/keeper/lock.go
```diff
@@ -560,3 +566,43 @@ func (k Keeper) Unlock(ctx sdk.Context, lock types.PeriodLock) error {
 	k.hooks.OnTokenUnlocked(ctx, owner, lock.ID, lock.Coins, lock.Duration, lock.EndTime)
 	return nil
 }
+
+// ForceUnlock ignores unlock duration and immediately unlock and refund.
+// CONTRACT: should be used only at the chain upgrade script
```
ValarDragon, line 571:
```suggestion
// CONTRACT: should be used only at the chain upgrade script
// TODO: Revisit for Superfluid Staking
```

x/gamm/keeper/pool.go
mconcat: In the original issue I think the IBC'd LP tokens are never redeemable for now(also there is no use case of IBCing out LP tokens yet). I could apply the first and the third suggestion now.

x/lockup/keeper/lock.go
mconcat: Will do so!

mconcat:
Few modification needed:
* Remove pool destruction time, instead delete from state
* Optimize refund logic performance
* Check module account are iterated over

ValarDragon:
We don't need to optimize performance for this PR, can just write an issue. (Performance gets ~basically solved if we just wait lol)

x/gamm/keeper/pool.go
mconcat: The iteration do include the module accounts. I will exclude the module account held LP tokens.

mconcat: Okay, annoyingly, there is no way to distinguish module address and normal address by only using the address itself. We could read all the module address and compare to them each time iterating over the list of address, but at this stage it feels little adhoc. Let me think about some alternatives a bit.

x/lockup/keeper/lock.go
mconcat: Problem is that the existing method checks whether the time has been passed specified on the lock duration, and ForceUnlock need to ignore them.

x/lockup/keeper/lock.go
mattverse: Hmmm... True... Did we / can we use `GetLocksLongerThanDurationDenom`?

mconcat: That worked!

x/lockup/keeper/lock.go
mconcat: Fixed by separating duration validation and actual unlocking logic

mconcat:
* Pools are now deleted when cleaned up
* Reuse existing unlock methods
* Module accounts are excluded
* Cleanup process is batched now(single iteration over the account space)

x/gamm/keeper/pool.go
```diff
@@ -80,6 +80,108 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+	if !store.Has(poolKey) {
+		return fmt.Errorf("pool with ID %d does not exist", poolId)
+	}
+
+	store.Delete(poolKey)
+	return nil
+}
+
+// CleanupBalancerPool destructs a pool and refund all the assets according to
+// the shares held by the accounts. CleanupBalancerPool should be called not during
```
ValarDragon, line 95:
```suggestion
// the shares held by the accounts. CleanupBalancerPool should not be called during
```

x/gamm/keeper/pool.go
```diff
@@ -80,6 +80,108 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+// the chain execution time, as it iterates the entire account balances.
```
ValarDragon, line 96:
```suggestion
// the chain execution time, as it iterates the entire account balances.
// TODO: once SDK v0.46.0, use https://github.com/cosmos/cosmos-sdk/pull/9611
```

x/gamm/keeper/pool.go
```diff
@@ -80,6 +80,108 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+//
+// All locks on this pool share must be unlocked in prior. Execute LockupKeeper.ForceUnlock
+// on remaning locks before calling this function.
```
ValarDragon, line 99:
```suggestion
// All locks on this pool share must be unlocked prior to execution. Use LockupKeeper.ForceUnlock
// on remaining locks before calling this function.
```

x/gamm/keeper/pool.go
```diff
@@ -80,6 +80,108 @@ func (k Keeper) SetPool(ctx sdk.Context, pool types.PoolI) error {
...
+		}
+		shareDenom := pool.GetTotalShares().Denom
+		pools[shareDenom] = pool
+		totalShares[shareDenom] = pool.GetTotalShares().Amount
+	}
+
+	moduleAccounts := make(map[string]string)
+	for _, module := range excludedModules {
+		moduleAccounts[string(authtypes.NewModuleAddress(module))] = module
+	}
```
ValarDragon, line 116: Nice!

ValarDragon (approved):
Awesome, looks good to merge to me! Thanks for fixing everything and addressing comments

PR #654, opened 2021-12-11
Added go mod download github.com/avast/retry-go

lightiv:
This was required before I was able to build the new binary.

Closes: #XXX

## Description

______

For contributor use:


lightiv:
Look like this is only need if go version is not v1.17.1

lightiv:
Probable should just have users check go version before starting the instructions.  Thanks.

ValarDragon:
We really want people to be using `go 1.17`. I'm actually shocked that the compiler doesn't fail to compile on go 1.16. Several of the dependencies depend on `go 1.17`, and I don't really know if/how they depend on go 1.17 features.

I'm pretty hesistant to merge this.

lightiv:
I agree, you should just make step 1 be to ensure Go v1.17.x is installed.

PR #655, opened 2021-12-13
Add default params to pool-incentives genesis

mattverse:
Due to added validation on genesis on validating whether distr_info in pool-incentives module is non-negative, the default genesis file currently omits error when starting from the initial state.

This PR adds default genesis for distr_info so that it does not omit error when starting chain from the initial state.

## Description

______

For contributor use:


ValarDragon:
Can you fix the test?

mattverse:
Fixed test: now changed default state to having nil instead of empty struct

ValarDragon (approved):
LGTM

Issue #656, opened 2021-12-13
[Epic] Testnet Automation


UnityChaos:
Tracking issue for automating testnet creation, upgrade, testing, etc.

The big things that are next step improvements seem to be:

- [x] Run pprof to see where memory is going during initialization
- [ ] Speedup InitGenesis/ExportGenesis time
- [x] Tendermint 2 peer required bug
- [ ] Testnets in CI

Future sprint (needs more details)
- [ ] Automate creation of export based testnets from Github Action
- [x] #732
- [ ] #729
- [ ] Github actions (using Terraform) for fully automated testnet creation and test running

daniel-farina:
@UnityChaos @czarcas7ic 

I started a very simple example project with instructions on how to create a node in DigitalOcean with Terraform and then executing two of the scripts you created during the upgrade just as an example.  This is very basic and WIP but it should give you a good place to get started. 

- The server gets created (deleted if it already exist with the same name)
- Single validator testnet is created
- the genesis state is configured
- The other scripts I did not get to but this should give you a good idea of how things work.


This runs locally with Terraform, not via Terraform cloud. We will eventually have a scripts that configures Terraform cloud as well but I recommend getting familiar running Terraform locally first. 

Link to infrastructure repo: https://github.com/osmosis-labs/infrastructure

I think this infrastructure repo could grown into a collection of modules and different scenarios that includes front-end and backend. We can use it as we experiment with the tool.

faddat:
I am not sure about the infra repo, because of:

#663 

and

https://github.com/osmosis-labs/infrastructure/issues/1

and of course, this isn't to say "don't do it" but maybe to say "let's do testnets as close as possible to how we do mainnets".

czarcas7ic:
@UnityChaos @daniel-farina Would love to set up a call sometime to figure out the best way forward with this, specifically hashing out clear expectations for some of these bullets.

ValarDragon:
This issue conjoins multiple things:
1) Starting up testnets more easily
2) Configurations for setting up testnets
3) Getting input & output from the testnet automatically
4) Getting testnet in CI completion steps

Everything depends on 1, and 1 is high priority. Adam's been doing a lot to improve it

I'm not sure what the plan/vision for 2 relative to 3 is

3 is important and a precursor for 4. Likely not a this-sprint priority though, as it is devops skill/time- constrained

czarcas7ic:
[...]

PR #657, opened 2021-12-13
Testnet Scripts [WIP - DNM]

UnityChaos:
Ref #656

Partially usable testnet automation scripts, used in testing v4 -> v5 upgrade.

Clean/start.sh and Mainnet/export.sh are usable, but need cleanup. The rest are WIP and should only be used as reference for building better scripts.

### DIFF
--- a/scripts/testnets/clean/start.sh
+++ b/scripts/testnets/clean/start.sh
@@ -1,0 +1,75 @@
+##### Spins up and starts a single validator Osmosis testnet
+
+#### Parameters
+export CHAIN_ID="osmosis-clean-testnet-X"
+export VERSION="v4.2.0"
+
+#### Initial node setup
+sudo apt-get update
+sudo apt-get upgrade -y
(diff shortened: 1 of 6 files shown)


UnityChaos:
Small issue discovered with the `clean/start.sh` script. Fails on new branches like `v5.x` and `v5.0.0`, but can be fixed by running `prepare-genesis mainnet` before running gentx. Not sure what the actual source of the problem is tho, since `prepare-genesis mainnet` does a lot of genesis setup, which is not necessarily what we want for a generic testnet setup. (Will investigate more later and fix separately)

UnityChaos:
Okay, fixed the #667 issue by moving up the `jq` line which sets up `distr_info`. In future that can be removed entirely once the `InitGenesis` fix for that is merged into the current release. (I believe it's already currently in main, but didn't make it into v5.x)

ValarDragon:
Closing this PR, because to my understanding the important bits have been extracted?

PR #663, opened 2021-12-15
Whisky: Tweaks and hardware to eliminate the epoch lag

faddat:
# Whisky would change the requirements for running osmosis and should be carefully considered.  

## Description

This would change osmosis by:

* making nodes basically require 64gb RAM (mine stay steadyish at 40gb, so having 64gb would be best)
* solving nearly all relayer issues
* Nearly requiring that validators run their nodes on bare metal, preferably on-prem for cost reasons

## Here is a build that won't have an epoch lag:

$1250
https://newegg.io/05ba1f6

You could do it with 64GB of RAM, too, and then it would cost about $1000. 

But here are some facts:

* AWS keeps exploding
* Hetzner is the only place where you can get reliable metal but their disks are much slower from a write perspective (at **only** 180k iops vs the ss980pro above at 410k iops in fio 4kb rw tests)
* Osmo doesn't need to have an epoch lag **at all**
* We processed the upgrade in about 8 minutes. On a system that should have exactly the same performance characteristics as the one in the shopping list above, except that ours uses a higher core count cpu, has a gpu for cyber, and runs 15 validators.

To validators, I'd say:

This is a one-time cost, though there's an ongoing time cost to maintaining your own machines, and you do need to think about your network some.  In order to help, here is some info on how we're doing things:

https://whimsical.com/ibc-relaying-Y3L552gPH9vzThRGjqpgjW

I reckon that validators strongly concerned with user experience and the health of the network may wish to begin tracking our aggregate performance as validators by looking for validators who are significantly faster or slower than average.  

8 mins vs 2 hours

All I want to say here, is that the epoch time wait, is a choice-- or at least, now it is. 

______

For contributor use:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -119,7 +119,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211209072213-711e78b4f6b4
+	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.43.0-rc3.0.20211215174815-40427335a890
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.14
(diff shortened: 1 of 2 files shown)


faddat:
I no longer think that it is a good idea for the vast majority of nodes to be drunk.

Issue #671, opened 2021-12-17
Simplify our app.go


ValarDragon:
As evidenced in the v5 upgrade, our app.go is ridiculously complex. The bug got introduced here https://github.com/osmosis-labs/osmosis/commit/dedcfb862313a57470f9b8b1be411a684a108747#diff-0f1d2976054440336a576d4[…]13012bce0e3c425819b7L342-R360 , but honestly the root cause imo is that this entire file is needlessly complex, where the order of every single line matters. It must be made more modular.

- [x] Move upgrade migration logic into its own folder, `/upgrades/v3`, `/upgrades/v4`, etc.
- [x] Make initParmsKeeper a general function somewhere, don't manually list out each module name. (Maybe this gets upstreamed to the SDK)
- [ ] #713

To goalpost, app.go is currently almost 1000 lines long. Lets first try to get this to under 500 lines.

ValarDragon:
I think we should make an `AppKeepers` struct, which has references to every keeper. Then OsmosisApp maintains a reference to the AppKeepers.

Then we make `AppKeepers.InitializeKeepers` and `AppKeepers.SetupHooks` as two cleanly divided methods.

faddat:
@ValarDragon this would surely reduce a lot of boilerplate in app.go

daniel-farina:
ref #697

ValarDragon:
Closing the tracking issue as the most important part is done, and #713 captures the remaining small component (that is still wanted / a PR would be very appreciated, but not at all a priority)

voluntatis-summ:
> I think we should make an `AppKeepers` struct, which has references to every keeper. Then OsmosisApp maintains a reference to the AppKeepers.
> 
> Then we make `AppKeepers.InitializeKeepers` and `AppKeepers.SetupHooks` as two cleanly divided methods.

@ValarDragon A side effect... other packages will have to access the App keepers like this `app.Keepers.StakingKeeper`?
applies to mainly test files that I see...  inside "/x" directory

ValarDragon:
Yeah, I don't think the keepers struct is worth the tradeoff. (You could make the main app 'inherit' all of those, but its still not great)

I actually tried writing the AppKeepers struct, before then switching to the solution of everything be a pointer

PR #538, opened 2021-10-17
chore: Cosmos SDK 0.44.* series upgrade for Osmosis

faddat:
Closes: #419 

## Description

This is ready for review.  One thing that'll need review especially is the decimal math stuff in the faddat/basecoin cosmos sdk.  I am working to get that up to 0.44.2 and when that's done, the sdk stuff will need review also. 

Thanks!

Tracking checklist:

- [x] Update my existing cosmos-sdk code to v0.44.2
- [x] Merge the  v0.44.2 cosmos sdk into osmosis-sdk
- [x] Get a proper Cosmos-sdk [pushed to osmosis-labs/cosmos-sdk](https://github.com/osmosis-labs/cosmos-sdk/tree/faddat/0442-osmo) instead of faddat/basecoin
- [x] ~~Cherry pick~~ (merge?) individual commits on [Osmo-v2-upgrade](https://github.com/osmosis-labs/cosmos-sdk) into 0.44.2 branch <- maybe-complete. 
- [x] create a 045-osmo branch and bring it current with sdk mainline-- this branch will become mainline-osmo once 045 is cut 
- [x] Update to 0.44.3
- [x] Make tests work again
- [ ] Use this as a jumping off point for putting #467 into action
- [ ] IBC-go/v2 readiness

NB:  This has also added & enabled authz and feegrant.  It should also ease some of @sunnya97's work on ethereum address formats. 

### Big big Goal
Osmo is IBC yo.

oh, and I hate, hate hate Marshaler. 

______

For contributor use:

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

PR #680, opened 2021-12-20
Move fork and upgrade logic into sub-directory structure

ValarDragon:
Cref: #671 

## Description

Pull request for trying to re-arranging around the fork and migration code. This moves them, to their own sub-directory.

I mainly did this to play around with some of the code structure here. The main bottleneck was import cycles due to app.go, and having to pass around keepers instead of the app.

I suggest we keep this structure, and then make a second "AppKeepers" struct, which actually stores all of the keepers. Then OsmosisApp stores a reference to an AppKeepers struct that we pass to all of these migration logics.

I think these sub-folders for each upgrade will make it easier to track old upgrade code, and keep a reasonable location to have it be commented out, without cluttering an important file like app.go.

Thoughts?

______

For contributor use:


UnityChaos (approved):
Full support on the idea, utAck on the implementation. Looks good to me, but we should probably try to test it somehow?
I think the idea of separating out each upgrade and then eventually commenting them out once the code is no longer logically reachable makes sense.

Also interesting that the v6 upgrade has essentially no contact with this refactor, because it does no actual state migrations.

ValarDragon:
Good point, this should have tests before we go through it. I'll write a test for v3 and v5 upgrades. (I'll do this by checking the parameters set during the upgrade)

ValarDragon:
Committed a test that updates the v4 upgrade keeper. (It reinstates an old test with mild adjustments, that got deleted in #538 ) This shows that the keeper reference setup is generally working here.

The thing thats not tested is adding new stores during an upgrade, but thats really essentially the same as what exists right now

ValarDragon:
Proceeding with merge, since Unity approved, and test added.

PR #696, opened 2021-12-28
WIP: AppKeepers struct concept

ValarDragon:
Closes: #XXX

## Description

Makes a separate AppKeepers struct, to avoid issues like the v6 upgrade from occuring. We store all Keepers as pointers, and separate out hook setting from them.

Moreover, by having the keepers be a separate struct, we can have upgrades take in AppKeepers directly without circular dependencies.

The main reason for a second struct in a different module is the upgrades module. Otherwise we could keep everything in the app.go, with this new method structure & local variable separation.

Thoughts on this approach / if its worth continuing as a separate module / struct, versus just separating initialization methods in app.go?

This is going to break many tests / require aliasing, which I'm now thinking may not be worth the more convenient upgrade logic.

______

For contributor use:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -57,7 +57,6 @@ import (
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/cosmos/cosmos-sdk/x/params"
 	paramsclient "github.com/cosmos/cosmos-sdk/x/params/client"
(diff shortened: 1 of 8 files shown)


ValarDragon:
I'd rather replace this with the approach in #697

PR #697, opened 2021-12-29
Make all keepers pointers, use multiple initialization methods

ValarDragon:
## Description

This is an alternate design to #696 , I actually like this better. It doesn't involve sweeping breaking changes across the app tests. I think we get the benefits of split up methods, and pointers unless necessary to not use them, without sacrificing usability elsewhere. I think we can get the abstraction we want in a keepers module still, by making a method to go from app -> AppKeepers struct, but not force its use throughout. (This is also really low priority imo)

We are refactoring initializing keepers into five steps, using pointer refs to keepers the whole way through:
* Initializing keepers with special SDK tie-ins", e.g. UpgradeKeeper, ParamsKeeper, CrisisKeeper, CapabilityKeeper.
* Setup upgrade keeper store loaders for new modules
* Initialize all other keepers
* Setup all the hooks
* Setup upgrade migrations

Then we have all other prior logic for setting module managers. The module managers should likely have code deduplicated in a future PR.

______

For contributor use:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -309,8 +309,9 @@ func NewOsmosisApp(
 	// we prefer to be more strict in what arguments the modules expect.
 	var skipGenesisInvariants = cast.ToBool(appOpts.Get(crisis.FlagSkipGenesisInvariants))
 
-	// NOTE: Any module instantiated in the module manager that is later modified
-	// must be passed by reference here.
+	// NOTE: All module / keeper changes should happen prior to this module.NewManager line being called.
+	// However in the event any changes do need to happen after this call, ensure that that keeper
+	// is only passed in its keeper form (not de-ref'd anywhere)
 	app.mm = module.NewManager(
 		genutil.NewAppModule(
 			app.AccountKeeper, app.StakingKeeper, app.BaseApp.DeliverTx,
--- a/app/app.go
+++ b/app/app.go
@@ -298,9 +298,10 @@ func NewOsmosisApp(
 	}
 
 	app.InitSpecialKeepers(skipUpgradeHeights, homePath, invCheckPeriod)
+	app.setupUpgradeStoreLoaders()
 	app.InitNormalKeepers()
 	app.SetupHooks()
-	app.setupUpgrades()
+	app.setupUpgradeHandlers()
 
 	/****  Module Options ****/
 
@@ -626,22 +627,7 @@ func (app *OsmosisApp) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
 }
 
-// RegisterTendermintService implements the Application.RegisterTendermintService method.
-func (app *OsmosisApp) setupUpgrades() {
-	// this configures a no-op upgrade handler for the v4 upgrade,
-	// which improves the lockup module's store management.
-	app.UpgradeKeeper.SetUpgradeHandler(
-		v4.UpgradeName, v4.CreateUpgradeHandler(
-			app.mm, app.configurator,
-			*app.BankKeeper, app.DistrKeeper, app.GAMMKeeper))
-
-	app.UpgradeKeeper.SetUpgradeHandler(
-		v5.UpgradeName,
-		v5.CreateUpgradeHandler(
-			app.mm, app.configurator,
-			&app.IBCKeeper.ConnectionKeeper, app.TxFeesKeeper,
-			app.GAMMKeeper, app.StakingKeeper))
-
+func (app *OsmosisApp) setupUpgradeStoreLoaders() {
 	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
 	if err != nil {
 		panic(fmt.Sprintf("failed to read upgrade info from disk %s", err))
@@ -655,6 +641,22 @@ func (app *OsmosisApp) setupUpgrades() {
 		// configure store loader that checks if version == upgradeHeight and applies store upgrades
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
+}
+
+func (app *OsmosisApp) setupUpgradeHandlers() {
+	// this configures a no-op upgrade handler for the v4 upgrade,
+	// which improves the lockup module's store management.
+	app.UpgradeKeeper.SetUpgradeHandler(
+		v4.UpgradeName, v4.CreateUpgradeHandler(
+			app.mm, app.configurator,
+			*app.BankKeeper, app.DistrKeeper, app.GAMMKeeper))
+
+	app.UpgradeKeeper.SetUpgradeHandler(
+		v5.UpgradeName,
+		v5.CreateUpgradeHandler(
+			app.mm, app.configurator,
+			&app.IBCKeeper.ConnectionKeeper, app.TxFeesKeeper,
+			app.GAMMKeeper, app.StakingKeeper))
 }
 
 // RegisterSwaggerAPI registers swagger route with API Server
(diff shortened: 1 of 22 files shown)


app/keepers.go
```diff
@@ -0,0 +1,326 @@
...
+		appCodec, keys[gammtypes.StoreKey],
+		app.GetSubspace(gammtypes.ModuleName),
+		app.AccountKeeper, app.BankKeeper, app.DistrKeeper)
+	app.GAMMKeeper = &gammKeeper
+
+	app.LockupKeeper = lockupkeeper.NewKeeper(
+		appCodec, keys[lockuptypes.StoreKey],
+		// TODO: Visit why this needs to be deref'd
+		*app.AccountKeeper,
+		app.BankKeeper)
```
ValarDragon, line 215: Need to double check why Account Keeper has to be deref'd for LockupKepeer. Its odd

app/keepers.go
```diff
@@ -0,0 +1,326 @@
...
+		appCodec,
+		keys[txfeestypes.StoreKey],
+		app.GAMMKeeper,
+	)
+	app.TxFeesKeeper = &txFeesKeeper
+
+	// register the proposal types
+	// TODO: This appears to be missing tx fees proposal type
+	govRouter := govtypes.NewRouter()
+	govRouter.AddRoute(govtypes.RouterKey, govtypes.ProposalHandler).
```
ValarDragon, line 256: Gov router is missing the proposal type for TxFees, this needs an issue to get added in next release.

app/keepers.go
```diff
@@ -0,0 +1,326 @@
...
+		*app.AccountKeeper,
+		app.BankKeeper)
+
+	app.EpochsKeeper = epochskeeper.NewKeeper(appCodec, keys[epochstypes.StoreKey])
+
+	app.IncentivesKeeper = incentiveskeeper.NewKeeper(
+		appCodec, keys[incentivestypes.StoreKey],
+		app.GetSubspace(incentivestypes.ModuleName),
+		*app.AccountKeeper,
+		app.BankKeeper, app.LockupKeeper, app.EpochsKeeper)
```
ValarDragon, line 223: Need to double check why Account Keeper has to be deref'd for IncentivesKeeper. Its odd, we may not have made the expected_keeper interface right. (For a new issue)

app/keepers.go
ValarDragon: New issue made

app/keepers.go
```diff
@@ -0,0 +1,326 @@
...
+	app.AccountKeeper = &accountKeeper
+	bankKeeper := bankkeeper.NewBaseKeeper(
+		appCodec,
+		keys[banktypes.StoreKey],
+		app.AccountKeeper,
+		app.GetSubspace(banktypes.ModuleName),
+		app.BlockedAddrs(),
+	)
+	app.BankKeeper = &bankKeeper
+
```
mattverse, line 126: Where were all these happening previously?

app/keepers.go
mattverse: Ah, I know why! It's because we're not using Keeper interface from expected_keeper as a parameter `NewKeeper` is taking in. Instead what NewKeeper is taking in as a parameter is authKeeper brought directly from cosmos-sdk. Issue needed for these!

app/keepers.go
mattverse: Same as above

app/keepers.go
ValarDragon: Ah awesome, thanks for finding that!

mattverse (approved):
LGTM! 
I like how this change affects app.go in the way that it became so much easier to understand and modulized as well!

app/upgrades/v4/prop12.go
```diff
@@ -10,7 +10,7 @@ import (
 	distrkeeper "github.com/cosmos/cosmos-sdk/x/distribution/keeper"
 )
 
-func Prop12(ctx sdk.Context, bank *bankkeeper.Keeper, distr *distrkeeper.Keeper) {
+func Prop12(ctx sdk.Context, bank bankkeeper.Keeper, distr *distrkeeper.Keeper) {
```
UnityChaos, line 13: Why not distrkeeper too?

x/gamm/genesis_test.go
```diff
@@ -113,7 +113,7 @@ func TestMarshalUnmarshalGenesis(t *testing.T) {
 
 	encodingConfig := osmoapp.MakeEncodingConfig()
 	appCodec := encodingConfig.Marshaler
-	am := gamm.NewAppModule(appCodec, app.GAMMKeeper, app.AccountKeeper, app.BankKeeper)
+	am := gamm.NewAppModule(appCodec, *app.GAMMKeeper, app.AccountKeeper, app.BankKeeper)
```
UnityChaos, line 116: Why only deref GammKeeper? Is this just some AppModule pattern? the main keeper is derefed, but account/bank aren't? (I see the pattern repeated with other NewAppModule calls)

app/app.go
```diff
@@ -498,30 +316,30 @@ func NewOsmosisApp(
...
-		distr.NewAppModule(appCodec, app.DistrKeeper, app.AccountKeeper, app.BankKeeper, app.StakingKeeper),
-		staking.NewAppModule(appCodec, app.StakingKeeper, app.AccountKeeper, app.BankKeeper),
-		upgrade.NewAppModule(app.UpgradeKeeper),
-		evidence.NewAppModule(app.EvidenceKeeper),
-		authzmodule.NewAppModule(appCodec, app.AuthzKeeper, app.AccountKeeper, app.BankKeeper, app.interfaceRegistry),
+		crisis.NewAppModule(app.CrisisKeeper, skipGenesisInvariants),
+		gov.NewAppModule(appCodec, *app.GovKeeper, app.AccountKeeper, app.BankKeeper),
+		mint.NewAppModule(appCodec, *app.MintKeeper, app.AccountKeeper),
+		slashing.NewAppModule(appCodec, *app.SlashingKeeper, app.AccountKeeper, app.BankKeeper, *app.StakingKeeper),
+		distr.NewAppModule(appCodec, *app.DistrKeeper, app.AccountKeeper, app.BankKeeper, *app.StakingKeeper),
```
UnityChaos, line 327: Not following why some of these are pointers and some aren't.
Something to do with:
```
	// NOTE: Any module instantiated in the module manager that is later modified
	// must be passed by reference here.
```
?
or patterns we're getting form upstream / SDK stuff?

app/app.go
```diff
@@ -612,7 +430,7 @@ func NewOsmosisApp(
 	app.SetAnteHandler(
 		NewAnteHandler(
 			app.AccountKeeper, app.BankKeeper,
-			app.TxFeesKeeper, app.GAMMKeeper,
+			*app.TxFeesKeeper, app.GAMMKeeper,
```
UnityChaos, line 433: ?

UnityChaos (approved):
Overall LGTM. Just pointed out some areas where it didn't make sense to me why some keepers were pointers and some were not. (I assume there are reasons)

x/gamm/genesis_test.go
ValarDragon: Yeah, NewAppModule depends on a concrete struct for the keeper that module defines, but interfaces for everything else. We can do pointers for interfaces

app/upgrades/v4/prop12.go
ValarDragon: Because we alter the distr modules fields in lines 35-37

app/app.go
ValarDragon: Yeah, these are all patterns from upstream, likely based on whether it depends on structs or interfaces.

Except for the first parameter, ideally everything else would be an interface dependency

app/app.go
ValarDragon: This is because our NewAnteHandler struct depends on the concrete struct (theres no interface for it to match)

Let me fix this to be pointer, good catch!

app/app.go
ValarDragon: Added comments to explain this better right there

app/upgrades/v4/prop12.go
UnityChaos: hmm... okay? but we do that through method calls, so it should be the same as bank right? Like with both modules we're using method calls that cause mutations, but we're not doing any direct modifications by setting fields directly?

It's been a while since I've thought seriously about golang call by reference / call by value differences, so I'm probably just not getting it. (I'm spoiled by Haskell / FP where everything is immutable lol)

PR #698, opened 2021-12-29
State Export Testnet Creation Automation Script

czarcas7ic:
I created a python script that automates the process of creating a testnet based off of a state export. This script does not rely on any magic numbers so should stay pretty stable and require little maintenance. I have tested this script on v6 and it worked without any issues. I also included a README to make the use of the script even easier. Would love a second pair of eyes on it to ensure it works across various setups.


ValarDragon:
This is so exciting, thanks for getting this in!

ValarDragon (approved):

ValarDragon:
You mentioned in DM theres some more testing you want to do, lmk once good to merge. (I think you should have perms actually?)

czarcas7ic:
@ValarDragon thank you! All the testing I wanted to do has been complete, so it's good to merge. With that being said, looking at #656 there is still a long ways to go, but this for sure will help with quick testnet setups in the interim. Also I do not have perms to merge.

PR #702, opened 2021-12-30
create global swap default swap and exit fees

sunnya97:
Closes: #XXX

## Description

______

For contributor use:


ValarDragon:
Hrmm, not sure I agree that we should restrict things to one swap fee. Its a pretty clear trade-off for tail-end things / things with less fungibility, you want higher swap fees. We also want more flexibility to change things to lower swap fees imo.

I'm fine with restricting to some presets / not giving the full range of motion, but I don't think restricting to one is a good move

I'm fine with just removing exit fee though tbh

sunnya97:
Yeah, governance should be able to change swap fees on pools.  This just sets the default for the permissionless creation of new pools

faddat:
I think this makes sense.  Can users override this default?

ValarDragon:
Going to close this for now. Lets revisit a design that fits the desired needs.

Issue #723, opened 2022-01-06
x/txfees: Make an ante-handler filter for recognizing Gamm arb txs, and charging a flat fee for them


ValarDragon:
To help combat arb bot fee spam, we should add a mempool filter detecting if a tx is an arb transaction. If so, require a different minimum fee than every other tx.

Theres three steps for this:
* Make a gamm function for detecting if this multiMsg is an on-chain arb cycle
* Make the anteHandler method use this for getting minimum gas price
* Extend the app.toml config to expose this (instead of hardcoding something)

We can add new parameters to the app.toml config, e.g. see this PR: https://github.com/cosmos/cosmos-sdk/pull/10561/files

I don't yet know if its possible to add these from the Osmosis repo, or if we have to do it in the SDK fork.

PR #741, opened 2022-01-09
Arbitrage tx mempool filter

ValarDragon:
Closes: #723 

## Description

This PR adds a mempool filter for arbitrage txs. When its hit, it charges a separate gas price, as specified in the full nodes application config.  This defaults to `0 uosmo` if not set in the app.toml, but the app.toml defaults to `.005uosmo/gas`.

The detection algorithm for arbitrage txs does have some false positives, and likely many false negatives. The false positives are described in the code comments, and I think are pretty miniscule. (Only thing I can see hitting it is a service that batches many users swap / LP positions together. Such a batching service seems like it can maintain connections to folks with lower fees though)

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Features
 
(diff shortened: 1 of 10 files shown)


UnityChaos (approved):
tested in production (as is known to be best practice)
:shipit:

PR #750, opened 2022-01-11
F1 Distribution

mattverse:
## Description

This PR implements F1 distribtuion mainly in the `incentives` module and the `lockup` module. The implementation of f1 distribution also fixes the current bug in distribution logic where unbonding locks would still receive rewards.

This PR mainly introduces three new state entries in `gauge.proto` in the process of implementing f1 distribution. 

1. CurrentReward
- CurrentReward has a unique key for each denom + lock duration combination. If current epoch is bigger than the last epoch the reward was processed, it increments the period entry in `CurrentReward`, reset CurrentReward and move what was in currentReward to HistoricalReward.

2. HistoricalReward
- HistoricalReward's main usage is to keep track of an accumulated reward ratio per denom of gauge, represented as `CumulativeRewardRatio`. This is used to calculate how much user is eligible to get distributed for a specific gauge reward(ex. rewardRatio * amtofTokens = amtToBeDistributed). Each unique denom + lock duration + period has its own key, each being stored with a different key. 

3. PeriodLockReward
- PeriodLockReward stores lockReward to be claimed for every individual lock. A map of string->uint64 is also included inside the state structure for PeriodLockReward, keeping track of what period each denom+lock duration combination it is in. PeriodLockReward references HistoricalReward to calculate reward for each lock.

Main implementation of the f1 distribution is in `distribute.go` and `rewards.go`(where getters and setters are implemented for the new state entries).

Additional points to be reviewed:
- Do we need all the HistoricalRewards in the state machine? Do we want to keep only the most recent _n_ number of historicalRewards
- Do we want all the event happening in tx? Are we currently avoiding emitting events?
______

For contributor use:


app/app.go
```diff
@@ -451,18 +451,18 @@ func NewOsmosisApp(
...
 		incentivestypes.NewMultiIncentiveHooks(
 		// insert incentive hooks receivers here
 		),
 	)
 
+	app.LockupKeeper = *lockupKeeper.SetHooks(
+		lockuptypes.NewMultiLockupHooks(
+			app.IncentivesKeeper.Hooks(),
+		),
+	)
```
ValarDragon, line 464: Any reason for moving this?

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
...
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"duration\""
+  ];
+}
+
+message PeriodLockReward {
+  option (gogoproto.stable_marshaler) = true;
+  uint64 lock_id = 1;
+  map<string, uint64> period = 2;
```
ValarDragon, line 80: Need to be really careful what this deserializes to. If it deserializes and serializes to a hashmap, pretty unclear that its going to be determinsitically ordered.

Would vastly prefer not using a map here. Why is there even a string identifier for period?

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
     (gogoproto.stdduration) = true,
     (gogoproto.moretags) = "yaml:\"lockable_durations\""
   ];
+}
+
+message HistoricalReward {
```
ValarDragon, line 46: Lets add some comments to each of these proto structs.

This one seems clear -- Its the historical accumulator values, for every time someone starts something that is earning rewards.

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
...
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+  uint64 period = 2;
+  uint64 last_eligible_epoch = 3;
+}
+
+message CurrentReward {
+  uint64 period = 1;
+  int64 last_processed_epoch = 2;
```
ValarDragon, line 58: Its not obvious to me on read here, why is there a separate period and last_processed_epoch.

Why don't we assume that theres just one period per epoch?

ValarDragon: (So concretely, only making there be a period variable, and commenting that at the moment its expected to be once per epoch, but in the future this could change)

Or is that what the code already does, and this last_processed_epoch is for debugging or unbonding?

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
...
+  // Reward per share for each denom
+  repeated cosmos.base.v1beta1.DecCoin cumulative_reward_ratio = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+  uint64 period = 2;
+  uint64 last_eligible_epoch = 3;
+}
+
+message CurrentReward {
```
ValarDragon, line 56: add comment saying this is the Latest accumulator value for rewards

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
...
+  ];
+  string denom = 5;
+  google.protobuf.Duration lock_duration = 6 [
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"duration\""
+  ];
+}
+
+message PeriodLockReward {
```
ValarDragon, line 77: I don't understand just via reading the proto files what this struct is for.

Is it the return value when getting rewards for a lock?

proto/osmosis/incentives/genesis.proto
```diff
@@ -18,4 +18,13 @@ message GenesisState {
     (gogoproto.stdduration) = true,
     (gogoproto.moretags) = "yaml:\"lockable_durations\""
   ];
+  repeated GenesisReward genesis_reward = 4 [ (gogoproto.nullable) = false ];
+  repeated PeriodLockReward period_lock_reward = 5
+      [ (gogoproto.nullable) = false ];
 }
+
+message GenesisReward {
```
ValarDragon, line 26: Oh interesting, this exports to genesis undistributed rewards. The cosmos SDK's staking auto distributes the rewards at export genesis.

What do you think makes the overall code simpler?

app/app.go
mattverse: LockupKeeper is now taking `IncentivesKeeper`, in which previously it didn't. 
I thought this should be done 'after' IncentivesKeeper has been declared, thus moved it below LockupKeeper.
Does the order of initialization not matter in this case?

ValarDragon: Ah this isn't built off main. I changed how this is all done in the keepers.go file there, so we no longer have to think about order here

mattverse: Cool, would have this back to the original state once this PR is ready and  merging with main

proto/osmosis/incentives/gauge.proto
mattverse: Good point! PeriodLockReward currently uses denom + lockDuration as. 
The move i'm planning to make is simply removing this map as a whole and instead use denom + lockDuration + period as key.

ValarDragon: I'm generally not clear on what this does atm, maybe it'll make sense to me after reviewing more of the logic

UnityChaos (commented):
So my reading of this is that we're copying over the pattern from staking/distribution, but with epochs instead of blocks. Is this intended to be the final result? or is this an intermediate solution, and we intend to eventually do block level updates of CurrentReward for each gauge?

I suppose a middle ground is always possible where we keep it epoch based, but progressively shorten the epoch length?

Also I'm assuming that the reason to make this initially (or always) epoch based is performance, as epoch=block would be highly costly, even if we aren't iterating all locks and are only iterating the gauges.

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,45 @@ message LockableDurationsInfo {
...
+  uint64 period = 2;
+  uint64 last_eligible_epoch = 3;
+}
+
+message CurrentReward {
+  uint64 period = 1;
+  int64 last_processed_epoch = 2;
+
+  // total coin locked per denom + lockDuration
+  cosmos.base.v1beta1.Coin coin = 3 [
```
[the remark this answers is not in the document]
mattverse: Correct! Changed!

proto/osmosis/incentives/gauge.proto
mattverse: Removed the whole period concept

proto/osmosis/incentives/gauge.proto
mattverse: Comments added!

proto/osmosis/incentives/gauge.proto
mattverse: Correct, we store how much rewards each individual locks are eligible for.

proto/osmosis/incentives/genesis.proto
mattverse: Do you mean adding distribution logic at export genesis?

x/incentives/README.md
```diff
@@ -72,3 +72,22 @@ MsgAddToGauge:
...
+
+
+## Estimating and Claiming Rewards
+
+The accumulated rewards can be estimated and claimed.
+
+### Examples
+
+```bash
+# estimate rewards for validator with a list of lock IDs
```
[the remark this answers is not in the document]
mattverse: Ah thanks for catching this!

mattverse:
So there has been a major change, I've deleted the whole `period` state in proto struct, it turned out that we do not need to track periods separately bc locks' distribution only happen based on epoch. With the proto changes, logics for f1 distribution also became way simpler!

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,57 @@ message LockableDurationsInfo {
     (gogoproto.stdduration) = true,
     (gogoproto.moretags) = "yaml:\"lockable_durations\""
   ];
+}
+
+// HistoricalReward stores the accumulator value for rewards
```
ValarDragon, line 46: Can we add a comment for how we're indexing this in state? Presumably its by lockup duration + lockup asset? (Don't care about the details of the key format, just the fields we index it on)

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,57 @@ message LockableDurationsInfo {
...
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coin"
+  ];
+  // total reward that is to be distributed
+  repeated cosmos.base.v1beta1.Coin rewards = 3 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+  // denom of share
+  string denom = 4;
```
ValarDragon, line 70:
```suggestion
  string lockup_denom = 4;
```

proto/osmosis/incentives/gauge.proto
```diff
@@ -41,4 +41,57 @@ message LockableDurationsInfo {
...
+  google.protobuf.Duration lock_duration = 5 [
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"duration\""
+  ];
+}
+
+// LastEligibleEpochByDurationAndDenom is a struct to store the latest epoch for each denom + lock duration combination
+message LastEligibleEpochByDurationAndDenom {
```
ValarDragon, line 79: I don't apriori understand the purpose of this struct. (Why isn't the latest epoch for every denom, lock duration combo the latest)

proto/osmosis/incentives/genesis.proto
ValarDragon: Yeah. (I'm also fine not doing it, just a question of overall simplicity)

proto/osmosis/incentives/gauge.proto
ValarDragon: or is this instead for when a user is claiming rewards, what the 'start' epoch is?

If so, is this indexed by user-address, rather than lock ID?

proto/osmosis/incentives/gauge.proto
ValarDragon: if its per lock, why is it `repeated LastEligibleEpochByDurationAndDenom last_eligible_epochs = 3;` instead of just a single entry for `last_claimed_epoch`

x/lockup/keeper/lock.go
```diff
@@ -718,6 +718,8 @@ func (k Keeper) unlock(ctx sdk.Context, lock types.PeriodLock) error {
 		return err
 	}
 
+	k.hooks.OnTokenUnlocked(ctx, owner, lock.ID, lock.Coins, lock.Duration, lock.EndTime)
```
ValarDragon, line 721: Is there a reason behind moving the hook? (I don't really have an opinion, just curious)

Also perhaps worth us coming back and renaming the hook to be `AfterTokenUnlocked`

mattverse: So there's logic added to hooks that uses lock reference to update periodLockReward before we delete the lock!

mattverse: Oer renaming the hooks, do you think it's a better decision to do it in a separate PR or straight up ahead on this one?

proto/osmosis/incentives/gauge.proto
mattverse: Ah, this can be rather confusing. So even though a period lock would have a single lock duration that could be locked, why would there be arrays of denom + lock duration?
Lock duration here does not mean the duration of the lock itself, we're keeping an array of lock duration for keeping track of at which epoch for each lock duration was this lock eligible for reward.

proto/osmosis/incentives/gauge.proto
mattverse: Same answer as in the comment below

p0mvn:
Hi team! What is the status of this?

ValarDragon:
Going to close for now due to how stale this is. Lets revisit as we do more F1 work.

But realistically, I think what we should be doing is making a generalized accumulator system, and then switch out parts of the stack to use that as a back-end.

Issue #716, opened 2022-01-05
Make lockup and incentives keepers take in AuthKeeper interface


ValarDragon:
The LockupKeeper and IncentiveKeeper currently take in AuthKeeper as a struct, this should change to be an interface.

_Originally posted by @mattverse in https://github.com/osmosis-labs/osmosis/pull/697#discussion_r779024500_

PR #755, opened 2022-01-12
fix: Change usage of account keeper in lockup module, incentives module

mattverse:
Closes: [#716](https://github.com/osmosis-labs/osmosis/issues/716)

## Description

Currently keeper in the lockup module and the incentives module are using authkeeper as a struct, importing directly from the cosmos-sdk instead of using them with the usage of `expected_keeper.go`.

This PR suggests change in the lockup module so that it uses accountKeeper interface via expected_keeper. 

Additionally, it was found that accountKeeper was imported in incentives module without any usage. Thus, this PR also includes removing taking in accountKeeper as a parameter for the incentives module keeper, resulting in changes in `app.go` as well.

______

For contributor use:

### DIFF
--- a/app/keepers.go
+++ b/app/keepers.go
@@ -220,7 +220,6 @@ func (app *OsmosisApp) InitNormalKeepers() {
 	app.IncentivesKeeper = incentiveskeeper.NewKeeper(
 		appCodec, keys[incentivestypes.StoreKey],
 		app.GetSubspace(incentivestypes.ModuleName),
-		*app.AccountKeeper,
 		app.BankKeeper, app.LockupKeeper, app.EpochsKeeper)
 
 	mintKeeper := mintkeeper.NewKeeper(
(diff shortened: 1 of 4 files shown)


app/keepers.go
```diff
@@ -220,7 +220,6 @@ func (app *OsmosisApp) InitNormalKeepers() {
 	app.IncentivesKeeper = incentiveskeeper.NewKeeper(
 		appCodec, keys[incentivestypes.StoreKey],
 		app.GetSubspace(incentivestypes.ModuleName),
-		*app.AccountKeeper,
```
ValarDragon, line 223: Can you remove the deref in the `app.LockupKeeper` init right above this?

ValarDragon (approved):
Awesome, thanks! LGTM. Feel free to merge once the deref is removed in the lockup keeper initialization (keepers.go line 214, 215)

app/keepers.go
mattverse: Fixed!

mattverse:
Deref removed in lockup keeper initialization, now squash and merging this PR!

PR #763, opened 2022-01-17
add rocksdb option to makefile

mircea-c:
## Description

Add build flags for rocksdb to the makefile


ValarDragon (approved):
Thank you!!

PR #769, opened 2022-01-19
Add wasmd module

ethanfrey:
This is to add a v1 of wasmd integration.

- [x] Builds, app compiles, and doesn't break tests
- [x] Properly read in more system config info from commands (more args to NewApp)
- [x] Update default genesis to use permissioned wasmd
- [x] Add minimal integration test - cannot upload, can create gov proposal to upload code
- [x] Add tx and query commands to cli

In another PR we look to add the x/upgrade logic to set the [global permissions](https://docs.cosmwasm.com/tutorials/governance/#available-configurations) on upgrade not just genesis


ethanfrey (commented):
Good point

ethanfrey:
I just want to write a few full stack tests on the integration, but the "production" code is done and can be reviewed

app/app.go
```diff
@@ -311,6 +349,7 @@ type OsmosisApp struct {
 	TxFeesKeeper         *txfeeskeeper.Keeper
 	SuperfluidKeeper     superfluidkeeper.Keeper
 	GovKeeper            *govkeeper.Keeper
+	WasmKeeper           wasm.Keeper
```
ValarDragon, line 352: Would it be alright to make this a pointer, for consistency / easing potential hooks? (superfluid not being pointer receiver is a bug right now)

ValarDragon: We do this to reduce the error chance when passing in keepers to hooks or other module initializations. (We messed up once via missing a subtle change, and this is what made us have the emergency upgrade from v5 to v6)

Happy to do this in a second PR though

app/genesis.go
```diff
@@ -13,8 +15,25 @@ import (
...
-	return ModuleBasics.DefaultGenesis(encCfg.Marshaler)
+	gen := ModuleBasics.DefaultGenesis(encCfg.Marshaler)
+
+	// here we override wasm config to make it permissioned by default
+	wasmGen := wasm.GenesisState{
+		Params: wasmtypes.Params{
+			CodeUploadAccess:             wasmtypes.AllowNobody,
+			InstantiateDefaultPermission: wasmtypes.AccessTypeEverybody,
+			MaxWasmCodeSize:              DefaultMaxWasmCodeSize,
+		},
```
ValarDragon, line 35: Oh cool! So when we do the upgrade to cosmwasm, do we need to override these in InitGenesis as well?

Heres what we had to do for `x/txfees` in our v5 upgrade: https://github.com/osmosis-labs/osmosis/blob/main/app/upgrades/v5/upgrades.go#L60-L66

app/keepers.go
```diff
@@ -104,7 +110,13 @@ func (app *OsmosisApp) InitSpecialKeepers(
 	app.UpgradeKeeper = &upgradeKeeper
 }
 
-func (app *OsmosisApp) InitNormalKeepers() {
+// Note: I put x/wasm here as I need to write it up to these other ones
+func (app *OsmosisApp) InitNormalKeepers(
+	homePath string,
+	appOpts servertypes.AppOptions,
+	enabledProposals []wasm.ProposalType,
```
ValarDragon, line 117:
```suggestion
	wasmEnabledProposals []wasm.ProposalType,
```

app/keepers.go
```diff
@@ -270,6 +315,11 @@ func (app *OsmosisApp) InitNormalKeepers() {
 		AddRoute(bech32ibctypes.RouterKey, bech32ibc.NewBech32IBCProposalHandler(*app.Bech32IBCKeeper)).
 		AddRoute(txfeestypes.RouterKey, txfees.NewUpdateFeeTokenProposalHandler(*app.TxFeesKeeper))
 
+	// The gov proposal types can be individually enabled
+	if len(enabledProposals) != 0 {
+		govRouter.AddRoute(wasm.RouterKey, wasm.NewWasmProposalHandler(app.WasmKeeper, enabledProposals))
+	}
```
ValarDragon, line 321:
```suggestion
	if len(wasmEnabledProposals) != 0 {
		govRouter.AddRoute(wasm.RouterKey, wasm.NewWasmProposalHandler(app.WasmKeeper, wasmEnabledProposals))
	}
```

go.mod
```diff
@@ -111,14 +114,20 @@ require (
...
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/ini.v1 v1.66.2 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	nhooyr.io/websocket v1.8.6 // indirect
 )
 
+require (
+	github.com/dgraph-io/badger/v2 v2.2007.3 // indirect
+	github.com/prometheus/client_golang v1.12.0
+)
```
ValarDragon, line 129: Don't really have an opinion on this, but seems odd that we have another require section.

app/wasmtest/helpers_test.go
```diff
@@ -0,0 +1,45 @@
...
+func FundAccount(t *testing.T, ctx sdk.Context, osmosis *app.OsmosisApp, acct sdk.AccAddress) {
+	err := simapp.FundAccount(osmosis.BankKeeper, ctx, acct, sdk.NewCoins(
+		sdk.NewCoin("uosmo", sdk.NewInt(10000000000)),
+	))
+	require.NoError(t, err)
+}
+
+// we need to make this deterministic (same every test run), as content might affect gas costs
+func keyPubAddr() (crypto.PrivKey, crypto.PubKey, sdk.AccAddress) {
+	key := ed25519.GenPrivKey()
```
ValarDragon, line 32: So do we need to switch to GenPrivKeyFromSecret? (https://github.com/tendermint/tendermint/blob/master/crypto/ed25519/ed25519.go#L144 )

Also a bit surprised that this is an ed25519 address, versus a secp one.

ValarDragon (approved):
Looks good to me!

Feel free to merge, we can adjust the minor point around the pointer receiver for wasm keeper in a second PR if you'd like.

app/app.go
ethanfrey: Happy to change this

app/genesis.go
ethanfrey: Yeah, we will need a migration step, that will set the params explicitly.
I wanted to do that on another PR. I could copy your migration setup, but would be happier to add a few lines to an existing one (if you could set up the framework for v6 upgrade)

app/keepers.go
ethanfrey: sure

go.mod
ethanfrey: Hmmm... I think this came from `go mod tidy`. I didn't play with this file myself. just using the go tooling

ethanfrey: I can merge them together gladly

app/wasmtest/helpers_test.go
ethanfrey: Yeah, copied some code I found elsewhere (in wasmd).

The original one in wasmd had a deterministic seed (generated from a counter). I tried making this random, but can use this. 

Can you link me to where you generate random addresses in your codebase? I can just import that instead

ethanfrey:
Happy to make the adjustments here.

My only question is how to improve the "random key/address" generation, or align it better to the osmosis standard.

app/wasmtest/helpers_test.go
ethanfrey: Or just update this code in a separate PR yourself. As long as it returns valid addresses, it is fine

ethanfrey:
Okay, I am off to bed but made the changes and will let the CI verify.

If it passes, can you merge?

If you start a migration PR (it would be v6 or v7?) let me know and I will add the wasm-specific migration so we can test that.

marbar3778:
bye bye my hard drives 😉

ValarDragon:
Awesome thanks! I'll get the initial migration handlers up in a PR!

ValarDragon:
@marbar3778 haha, hopefully cosmwasm state sync will save us in the future

app/app.go
```diff
@@ -842,6 +890,7 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(poolincentivestypes.ModuleName)
 	paramsKeeper.Subspace(superfluidtypes.ModuleName)
 	paramsKeeper.Subspace(gammtypes.ModuleName)
+	paramsKeeper.Subspace(wasm.ModuleName)
```
alpe, line 893: @ethanfrey 
The pinned contracts are not loaded within the `if loadLatest` [block](https://github.com/CosmWasm/wasmd/blob/master/app/app.go#L676-L679)

ethanfrey: Very good catch. Thank you

Issue #724, opened 2022-01-06
x/txfees: Make an ante-handler filter for recognizing High gas txs, and having a min gas price for them


ValarDragon:
Similar to #723 , we should make a mempool filter for high gas fee txs to have non-zero fee requirements. (Configurable by every full node operator)

E.g. if gas fee > $THRESHOLD, require minimum-gas-price to be at least `x`. 

Would require two params in the app.toml config `threshold` and `x`. Though maybe its fine to just hardcode the threshold?

I think its probably worthwhile for governance to consider requiring this in the state machine. (Though that could also be premature with cosmwasm, unclear how cosmwasm contracts will perform gas-wise)

ValarDragon:
I suggest we hard code first $THRESHOLD=1 million gas, and get a config param for `MINIMUM_GAS_PRICE`.

We can probably implement this first with `MINIMUM_GAS_PRICE = .01uosmo`

hleb-albau:
took it

PR #777, opened 2022-01-22
Add separate min gas price option for txes with high gas wanted

hleb-albau:
Closes: #724

## Description

- Refactor osmosis mempool filter
- Add separate min gas price option for txes with high gas wanted

## Open questions to reviewer:

I think we should not move `HighGasTxThreshold` parameter into consensus, but leave it just as a node option. If so, it's better to add it to `app.toml`

## Other

Do not waste time for choosing words, if something is wrong just said it straightforward, i am OK with that.

______

For contributor use:

### DIFF
--- a/cmd/osmosisd/cmd/root.go
+++ b/cmd/osmosisd/cmd/root.go
@@ -110,9 +110,17 @@ func initAppConfig() (string, interface{}) {
 ###############################################################################
 
 [osmosis-mempool]
+# This is the max allowed gas any tx. 
+# This is only for local mempool purposes, and thus	is only ran on check tx.
+max-gas-wanted-per-tx = "25000000"
+
 # This is the minimum gas fee any arbitrage tx should have, denominated in uosmo per gas
 # Default value of ".005" then means that a tx with 1 million gas costs (.005 uosmo/gas) * 1_000_000 gas = .005 osmo
 arbitrage-min-gas-fee = ".005"
+
+# This is the minimum gas fee any tx with high gas demand should have, denominated in uosmo per gas
+# Default value of ".0025" then means that a tx with 1 million gas costs (.0025 uosmo/gas) * 1_000_000 gas = .0025 osmo
+min-gas-price-for-high-gas-tx = ".0025"
 `
 
 	return OsmosisAppTemplate, OsmosisAppCfg
--- a/x/txfees/types/options.go
+++ b/x/txfees/types/options.go
@@ -1,0 +1,80 @@
+package types
+
+import (
+	"fmt"
+	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cast"
+)
+
+// If Options are not set in a config somewhere,
+// use defaults to preserve functionality with old node software
+
+// TODO: Bump after next minor version. (in 6.2+)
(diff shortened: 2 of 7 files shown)


x/txfees/types/options.go
```diff
@@ -0,0 +1,80 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cast"
+)
+
+// If Options are not set in a config somewhere,
+// use defaults to preserve functionality with old node software
+
+// TODO: Bump after next minor version. (in 6.2+)
+var DefaultMinGasPriceForArbitrageTx = sdk.ZeroDec()
+var DefaultMinGasPriceForHighGasTx = sdk.MustNewDecFromStr("0.01")
```
ValarDragon, line 15:
```suggestion
var DefaultMinGasPriceForHighGasTx = sdk.ZeroDec()
```

ValarDragon (approved):
Woah awesome job on this! 

This is extremely well done imo, thanks for refactoring everything into Mempool options, and cleaning up the config parsing

x/txfees/types/options.go
ValarDragon: Do you think we should make the default for new people on the software to be 0, or .01?

ValarDragon:
I think your right re: Not moving high gas thresholds as something done in consensus as well, and just keeping it in app.go. (Which this PR does)

x/txfees/types/options.go
hleb-albau: We can start as with arbitrage txes - with default zero, but active community validators can set it to `.01` to prevent spam waves.

ValarDragon: Ok cool, and everyone who regenerates the app.toml will get the new setting

cmd/osmosisd/cmd/root.go
```diff
@@ -110,9 +110,17 @@ func initAppConfig() (string, interface{}) {
...
+# This is only for local mempool purposes, and thus	is only ran on check tx.
+max-gas-wanted-per-tx = "25000000"
+
 # This is the minimum gas fee any arbitrage tx should have, denominated in uosmo per gas
 # Default value of ".005" then means that a tx with 1 million gas costs (.005 uosmo/gas) * 1_000_000 gas = .005 osmo
 arbitrage-min-gas-fee = ".005"
+
+# This is the minimum gas fee any tx with high gas demand should have, denominated in uosmo per gas
+# Default value of ".01" then means that a tx with 1 million gas costs (.01 uosmo/gas) * 1_000_000 gas = .01 osmo
+min-gas-price-for-high-gas-tx = ".01"
```
ValarDragon, line 123:
```suggestion
# Default value of ".005" then means that a tx with 1 million gas costs (.005 uosmo/gas) * 1_000_000 gas = .005 osmo
min-gas-price-for-high-gas-tx = ".005"
```

ValarDragon: to be consistent with arb gas

hleb-albau: ah, copy-paste :)

hleb-albau: will fix a test quickly

ValarDragon: tysm!

Also no worries, didn't specify anything about concrete number to default to (or think about it too much before)

hleb-albau: test fixed, commits squashed

PR #779, opened 2022-01-22
create rocksdb guide

clemensgg:
re-open against correct branch


ValarDragon:
Can instructions be linked to for state syncing in here as well?

faddat (approved):
awesome!

@ValarDragon I do not think they can be.

ValarDragon (approved):
Ah I see, @czarcas7ic pointed out to me that callum's patch only works for goleveldb atm

Issue #788, opened 2022-01-27
Semantic Version Tracking issue


ValarDragon:
We need to fix the semantic versioning of all the branches in the repo, folks are currently unable to import the codebase for most versions. (And can't import multiple versions at once)

- [ ] v3.x
- [ ] v4.x
- [ ] v5.x
- [ ] v6.x
- [ ] main (I believe main should be `v7` for now?)

cref #621

PR #793, opened 2022-01-27
osmosis v6 semantic versioning

faddat:
Works Towards: #788 

## Description

Semver for osmosis v6.x

______

For contributor use:


mattverse:
Thanks for this! LGTM! Should be good to merge to 6.x branch

go.mod
```diff
@@ -1,9 +1,9 @@
-module github.com/osmosis-labs/osmosis
+module github.com/osmosis-labs/osmosis/v6
 
 go 1.17
 
 require (
-	github.com/cosmos/cosmos-sdk v0.44.5
+	github.com/cosmos/cosmos-sdk v0.45.0
```
mattverse, line 6:
```suggestion
	github.com/cosmos/cosmos-sdk v0.44.5
```

go.mod
```diff
@@ -114,17 +115,25 @@ require (
...
 	github.com/cosmos/ibc-go/v2 => github.com/osmosis-labs/ibc-go/v2 v2.0.2-osmo
+	// Use cosmos-compatible protobufs
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	// fix builds with rocksdb
 	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
+	// use osmosis-patched tendermint
 	github.com/tendermint/tendermint => github.com/osmosis-labs/tendermint v0.34.12-0.20220109173307-59a781894ea7
+	// Use grpc compatible with cosmos-flavored protobufs
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
 )
```
mattverse, line 139: This is wonderful!

mattverse (changes requested):
After changing cosmos-sdk version to the original state, this should be good to merge! Thanks for this!

mattverse (approved):
LGTM! Good to merge to v6.x branch!

ValarDragon (approved):
Thank you so much!!! This is really big for people to be able to work with Osmosis

marbar3778:
could you rename your long-lived branches to release-v6.x as im unable to import v6.x

ValarDragon:
Hrmm, not sure I undesrtand? Is it that release-v6.x looks better in imports (makes sense / seems reasonable) or that some tooling is failing with `v6.x` and would not fail with `release-v6.x`?

marbar3778:
mainly I can't import v6.x into my go mod file. It thinks im trying to import semantic version but .x breaks it

ValarDragon:
ohh interesting, ok will switch

marbar3778:
oh I forgot to edit, but release-6.x is better than v6.x since go errors on this as well

ValarDragon:
I'm generally confused, I thought you can't import branches in a go.mod. Only releases and commits. When you do `go get repo@branch` it just puts the link to that latest commit in the go.mod. 

https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies

> Using a branch name such as go get foo@master (foo@default with mercurial) is one way to obtain the latest commit regardless of whether or not it has a semver tag.

But I haven't tried much here / you probably understand this better than I do! `go get github.com/osmosis-labs/osmosis/v6@v6.x` is basically the command thats failing rn?

marbar3778:
oh nvm the v6@v6 fixed it. I didn't think the /v6 was needed but makes sense. Now on the sdk im running into the same issue

PR #797, opened 2022-01-27
Delete duplicate usage of setGauge at Incentives module InitGenesis

mattverse:
## Description

Currently, the `SetGaugeWithRefKey` method which is getting called in initGenesis for the incentive module has duplicate code where setGauge would be done twice(at the beginning of the method, at the end of the method). 

This PR removes the duplicate instance of setting gauges.

______

For contributor use:

### DIFF
--- a/x/incentives/keeper/gauge.go
+++ b/x/incentives/keeper/gauge.go
@@ -85,12 +85,7 @@ func (k Keeper) SetGaugeWithRefKey(ctx sdk.Context, gauge *types.Gauge) error {
 			return err
 		}
 	}
-	store := ctx.KVStore(k.storeKey)
-	bz, err := proto.Marshal(gauge)
-	if err != nil {
-		return err
-	}
-	store.Set(gaugeStoreKey(gauge.Id), bz)
+
 	return nil
 }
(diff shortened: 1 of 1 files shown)


x/incentives/keeper/gauge.go
```diff
@@ -85,12 +85,7 @@ func (k Keeper) SetGaugeWithRefKey(ctx sdk.Context, gauge *types.Gauge) error {
 			return err
 		}
 	}
-	store := ctx.KVStore(k.storeKey)
-	bz, err := proto.Marshal(gauge)
-	if err != nil {
-		return err
-	}
-	store.Set(gaugeStoreKey(gauge.Id), bz)
```
mattverse, line 93: Duplicate happens here
https://github.com/osmosis-labs/osmosis/pull/797/files#diff-d57570c2b30c7fe37108bd7847d1419ad3f0240d7dd28d37e0aecff0730236c8R62-R65

ValarDragon (approved):
LGTM! Thanks for the fix. I double checked that those respective addGauges were not writing anything to gauge struct

PR #804, opened 2022-01-29
Semantic versioning for v7

faddat:
Works on: #788 

## Description

Semantic versioning for osmosis v7

______

For contributor use:


go.mod
```diff
@@ -1,9 +1,9 @@
-module github.com/osmosis-labs/osmosis
+module github.com/osmosis-labs/osmosis/v7
 
 go 1.17
 
 require (
-	github.com/cosmos/cosmos-sdk v0.44.5
+	github.com/cosmos/cosmos-sdk v0.45.0
```
mattverse, line 6: We should probably remain on v0.44.5 for now!

mattverse (changes requested):
Should be good to merge once go.mod sdk version is reverted to original state!

ValarDragon:
I am generally lost as to what is standard for golang projects. Is it to make main with breaking changes with one version higher, or is it with the same last major release.

Also how do golang projects do backports with these version import differences :/

mattverse:
From my knowledge its make breaking changes with one version higher.

mattverse:
> Also how do golang projects do backports with these version import differences :/
@ValarDragon wdym? Can you give an example ?

faddat:
You know, there's an issue in this, too.  I'm going to jump off of main a second time, to avoid it.

faddat:
@ValarDragon if you still want to use mergify, we need to add it to this repo.  I am going to close this one because I think it might add commits of Sunny's not meant for `main`. 

Also seems like some of the wasm tests are trying to import /x/ibc

I think I got halfway through an upgrade of the tests w/ Juno and it should be the same kind of thing so I'll try and fix them.

faddat:
@mattverse this is complete and I've merged the latest main branch now, too.

mattverse (approved):
LGTM! Thanks for this @faddat 

In the future, we don't want main to have sem-ver of the next major release, but sem-ver for the latest release, since that would make sense for all the backporting that would be happening. Since we have a major upgrade coming up soon, this PR, in particular, makes sense + super helpful!

x/lockup/keeper/legacy.go
```diff
@@ -0,0 +1,220 @@
+package keeper
```
ValarDragon, line 1: Lets re-delete this file

ValarDragon (approved):
Thanks for the in-depth review @mattverse / checking what other projects do as the standard here.

Did a quick skim, and only saw the v7 change + the re-added legacy.go file.

Lets also file an issue to add the semver situation to the contributing.md

ValarDragon:
Going to go ahead and merge before this is broken by the next PR

PR #840, opened 2022-02-11
Move lock.go functions into iterator.go, lock_refs.go and store.go

ValarDragon:
This makes the code more structured, and is a precursor to a subsequent commit
of trying to de-duplicate a lot of this logic to reduce API & bug surface

I'm making this a separate PR to make it easy to review method re-shuffles / refactor work, independently from changes into the logic.


ValarDragon:
We don't need to do anything to get these added to appsim right? It happens automatically from the SimulationManager in app.go?

PR #846, opened 2022-02-11
Upgrade iavl and sdk with fast storage

r0mvn:
Closes: #XXX

**Background**
This release introduces an upgrade of iavl and the SDK with fast storage implementation that represents a copy of the latest version of the tree with performance improvements. The fast storage nodes are indexed by key on disk as opposed to by hash with the original tree implementation. It allows us to preserve data locality for the latest state, significantly improving iterations and queries. This implementation introduces a negligible overhead for writes.

**Downgrade-re-upgrade protection**
We introduced a downgrade and re-upgrade protection where we guard for potential downgrades of iavl and the subsequent enablement of the fast storage again. This is done so by storing the metadata about the current version of the storage and the latest live state stored.

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [v6.3.0]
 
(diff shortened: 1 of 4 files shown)


ValarDragon (approved):
Awesome!

hleb-albau:
WOW

PR #859, opened 2022-02-13
min duration from days to hours, changed weekly to day

czarcas7ic:
Fixes two queries

`osmosisd query lockup total-locked-of-denom gamm/pool/1`
Wont work since it tries to use `1d` as default min duration. Changed default to 86400s.

`osmosisd query epochs current-epoch -h`
Shows weekly as a potential input but its week not weekly. While I was at it I just changed default to day since it is most used.

______

For contributor use:

### DIFF
--- a/x/lockup/client/cli/flags.go
+++ b/x/lockup/client/cli/flags.go
@@ -14,13 +14,13 @@ const (
 func FlagSetLockTokens() *flag.FlagSet {
 	fs := flag.NewFlagSet("", flag.ContinueOnError)
 
-	fs.String(FlagDuration, "86400s", "The duration token to be locked. e.g. 86400s, 604800s, 1209600s")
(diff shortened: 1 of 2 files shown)


x/lockup/client/cli/flags.go
```diff
@@ -14,13 +14,13 @@ const (
 func FlagSetLockTokens() *flag.FlagSet {
 	fs := flag.NewFlagSet("", flag.ContinueOnError)
 
-	fs.String(FlagDuration, "86400s", "The duration token to be locked. e.g. 1h, 1m, 1s, 0.1s")
+	fs.String(FlagDuration, "86400s", "The duration token to be locked. e.g. 86400s, 604800s, 1209600s")
```
ValarDragon, line 17: Can we put these in hours? I believe thats the largest default unit supported by golang

ValarDragon (approved):
TY, LGTM! Can we change the units to hours, since its easier to understand than seconds?

Issue #761, opened 2022-01-17
Fix package in `proto/osmosis/gamm`


mattverse:
Currently the gamm module proto package has room for the two following fixes before we get it into a new release. 
1. fix package for genesis proto to osmosis.gamm.v1beta1 
 https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/v1beta1/genesis.proto#L2 

2. change package structure so that pool-models goes under gamm/v1beta1.

Both changes would be state-breaking.

ValarDragon:
I'm actually not sure that the genesis.proto one is state or genesis breaking. It is breaking the genesis proto struct, however we use genesis json's which atm don't contain the type fields. 

The package structure for pool-params is def state breaking, but since the pool models isn't on mainnet, no need for migration.

PR #866, opened 2022-02-14
Fix proto package

mconcat:
Closes: #761 

Because of the restriction of protobuf package structure, the gamm proto files(not the pool model files) should be also be under a separate folder, in this PR, `types`. The protobuf does not allow a package with subfolders in it.

## Description

______

For contributor use:


mattverse (commented):
Oh interesting is it the `error:inconsistent package import paths` or something different that forces us to put all the other files under a sub-directory?

marbar3778 (commented):
thank you for doing this!!

you can also go directly to v1 if your api has stabilised

proto/osmosis/gamm/v1beta1/types/pool.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package osmosis.gamm.v1beta1;
+package osmosis.gamm.v1beta1.types;
```
ValarDragon, line 2: Can we undo this? Worried about this breaking things

proto/osmosis/gamm/v1beta1/types/query.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
-package osmosis.gamm.v1beta1;
+package osmosis.gamm.v1beta1.types;
```
ValarDragon, line 2: Can we undo this? Worried about this breaking things

proto/osmosis/gamm/v1beta1/types/pool.proto
marbar3778: agree here, having one package for gamm.v1beta1 is better than nesting it in different directories. You have to think that the longer the package name the added few characters add a few bytes to storage.

ValarDragon:
I suspect the types is going to cause @mattverse 's regression test to fail when this is updated to main. (And why we should change the proto package to not include types)

If the regression test doesn't fail, then it doesn't matter!

ValarDragon:
This was really helpful! We learned (the _hard way_) that these proto packages are state breaking. So we changed in #887 to make these all match what was there before identically

PR #868, opened 2022-02-15
test: Refactored test to follow table driven approach

stackman27:
Refactored `msg_server_test.go` as per @ValarDragon request. More details [here](https://hackmd.io/W2PIo4SNREGMjN9Zyo2rbQ?both)

This PR allows devs to create test for `LockTokens` easily with table driven approach. I have only added "normal working lock test" in this refactoring, however once I receive Approach ACK, i plan on adding 

- [x] locking more coins than are in the address test

PS: very new to this codebase, so apologies for any style issues and looking for literally any feedback to improve

### DIFF
--- a/x/lockup/keeper/msg_server_test.go
+++ b/x/lockup/keeper/msg_server_test.go
@@ -9,55 +9,92 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/x/lockup/types"
 )
 
-// TODO: Make table driven
 func (suite *KeeperTestSuite) TestMsgLockTokens() {
-	suite.SetupTest()
+	type param struct {
+		coinsToLock         sdk.Coins
+		lockOwner           sdk.AccAddress
+		duration            time.Duration
+		coinsInOwnerAddress sdk.Coins
+	}
 
-	// lock coins
-	addr1 := sdk.AccAddress([]byte("addr1---------------"))
-	coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
+	tests := []struct {
+		name       string
+		param      param
+		expectPass bool
+	}{
+		{
+			name: "creation of lock via lockTokens",
+			param: param{
+				coinsToLock:         sdk.Coins{sdk.NewInt64Coin("stake", 10)},       // setup wallet
+				lockOwner:           sdk.AccAddress([]byte("addr1---------------")), // setup wallet
+				duration:            time.Second,
+				coinsInOwnerAddress: sdk.Coins{sdk.NewInt64Coin("stake", 10)},
+			},
+			expectPass: true,
+		},
+		{
+			name: "locking more coins than are in the address",
+			param: param{
+				coinsToLock:         sdk.Coins{sdk.NewInt64Coin("stake", 20)},       // setup wallet
+				lockOwner:           sdk.AccAddress([]byte("addr1---------------")), // setup wallet
+				duration:            time.Second,
+				coinsInOwnerAddress: sdk.Coins{sdk.NewInt64Coin("stake", 10)},
+			},
+			expectPass: false,
+		},
+	}
 
-	err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, addr1, coins)
-	suite.Require().NoError(err)
-	_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, addr1, coins, time.Second)
-	suite.Require().NoError(err)
+	for _, test := range tests {
+		suite.SetupTest()
 
-	// creation of lock via LockTokens
-	msgServer := keeper.NewMsgServerImpl(suite.app.LockupKeeper)
-	_, err = msgServer.LockTokens(sdk.WrapSDKContext(suite.ctx), types.NewMsgLockTokens(addr1, time.Second, coins))
+		err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, test.param.lockOwner, test.param.coinsInOwnerAddress)
+		suite.Require().NoError(err)
 
-	// check locks
-	locks, err := suite.app.LockupKeeper.GetPeriodLocks(suite.ctx)
-	suite.Require().NoError(err)
-	suite.Require().Len(locks, 1)
-	suite.Require().Equal(locks[0].Coins, coins)
-	// check accumulation store is correctly updated
-	accum := suite.app.LockupKeeper.GetPeriodLocksAccumulation(suite.ctx, types.QueryCondition{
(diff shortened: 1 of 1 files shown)


mattverse (commented):
Concept ACK. 
Per styling issues, `go fmt`ing code would shall suffice!

x/lockup/keeper/msg_server_test.go
```diff
@@ -9,55 +9,77 @@ import (
...
+	}
+
+	for _, test := range tests {
+		suite.SetupTest()
+
+		if test.expectPass {
+			err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, test.param.lockOwner, test.param.coinsToLock)
+			suite.Require().NoError(err)
+
+			_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, test.param.lockOwner, test.param.coinsToLock, test.param.duration)
```
stackman27, line 42: @ValarDragon @mattverse 
Why doesn't `suite.app.LockupKeeper.LockTokens(...)` throw an error when my coinsToLock > coinsInAddress? 

I was expecting an error in this line, but seems like I need to go all the way to line 53 `suite.Require().Equal(locks[0].Coins, test.param.coinsToLock)` to return the error 

Is there a better way to handle this?

mattverse: With what parameters are you trying to create a test? Are you trying to create a test case for trying to lock tokens with more than the account has?

stackman27: Yup

```	{
			name: "locking more coins than are in the address",
			param: param{
				coinsToLock:         sdk.Coins{sdk.NewInt64Coin("stake", 20)},       // setup wallet
				lockOwner:           sdk.AccAddress([]byte("addr1---------------")), // setup wallet
				duration:            time.Second,
				coinsInOwnerAddress: sdk.Coins{sdk.NewInt64Coin("stake", 10)},
			},
			expectPass: false,
		},
```

mattverse: I see, I think it's probably because you're funding the account at line 39 with `test.params.CoinsToLock`, not with `test.params.coinsInOwnerAddress`

stackman27: aah i see

stackman27: Also @mattverse the minimum duration to LockTokens is a day right and anything <=0 should throw an error? 

Can you please direct me to the file where errors are handled for `lockTokens`? I can't seem to figure out if it's done locally or not

mattverse: Actually there isn't a minimum duration for LockTokens, they are completely different from incentvies( as of which has the minimum duration of a day to be eligible for incentives) 

Per different test cases for lockTokens, you might want to take a look at https://github.com/osmosis-labs/osmosis/blob/main/x/lockup/keeper/lock_test.go

stackman27: I see that makes sense. In that case, this PR is ready for full review!

stackman27:
@mattverse @ValarDragon removed duration = 0 and ready for review!

mattverse:
planning to get this reviewed post v7 upgrade

x/lockup/keeper/msg_server_test.go
```diff
@@ -9,55 +9,97 @@ import (
...
+
+	for _, test := range tests {
+		suite.SetupTest()
+
+		if test.expectPass {
+			// Setup the Account
+			err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, test.param.lockOwner, test.param.coinsInOwnerAddress)
+			suite.Require().NoError(err)
+
+			_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, test.param.lockOwner, test.param.coinsToLock, test.param.duration)
```
ValarDragon, line 55: This structure isn't quite right.

It should do the funding of the account & the lock tokens off the params. Then it should do the checks based on the `if expPass` logic. 

E.g. imagine making this scalable, for adding more test cases with coinsToLock having `100k uosmo`, or coinsToLock having multiple entries `100uosmo,10uatom`

stackman27: In that case should I use `addr1` or `lockOwner` inside `if expPass` logic to test `LockTokens`?

My guess is to use `addr1` because it stays consistent across tests but then why would we even need `lockOwner` in params?

ValarDragon: Use `lockowner`, thats why its a param!

x/lockup/keeper/msg_server_test.go
```diff
@@ -47,14 +47,16 @@ func (suite *KeeperTestSuite) TestMsgLockTokens() {
...
+		// lock coins
+		addr1 := sdk.AccAddress([]byte("addr1---------------"))
+		coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
 
-			_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, test.param.lockOwner, test.param.coinsToLock, test.param.duration)
-			suite.NoError(err, "test %v", test.name)
+		err := simapp.FundAccount(suite.app.BankKeeper, suite.ctx, addr1, coins)
+		suite.Require().NoError(err)
+		_, err = suite.app.LockupKeeper.LockTokens(suite.ctx, addr1, coins, time.Second)
+		suite.Require().NoError(err)
```
stackman27, line 58: @ValarDragon updated as suggested. 

If the approach is correct, I was thinking about moving this functionality to `x/lockup/keeper/keeper_test.go` very similar to [this](https://github.com/osmosis-labs/osmosis/blob/f2a60d546887aca5890640a11c37518e6e6a7c81/x/gamm/keeper/keeper_test.go#L80) `swap_test` setup function feature. lmk what you think

x/lockup/keeper/msg_server_test.go
mattverse: In that case, I think we can still place `LockTokens` outside the `if expectPass / else` logic and only check if requires error or not within the if/else.

stackman27: Fixed!

mattverse (approved):
LGTM. 
I think the use cases of `expectPass` can vary or change in the future depending on additional test cases we would add in the future, but I think as of the current test cases, structure looks good

mattverse:
@stackman27 Once you resolve conflicts with main, we should be good to merge.

stackman27:
> 

Fixed!

Issue #867, opened 2022-02-15
x/lockup: Combine lockups for same duration and jitter amount


ValarDragon:
We want to move to a model where a user has one lockup for a given (denom, duration) pair, can do partial unlocks against that, and superfluid stake sub-components of that. This simplifies logic massively for all front-ends and data integrators.

Issues / constituents that we would like to resolve for this:
- Prior to the v5 upgrade, all locks for the same (owner, denom, duration) triplet created a new lock. (They now add to the existing lockup)
- Until the v4 upgrade, all locks from the front-end would add a jitter to the duration. This is due to a gas bug that existed on-chain. We should combine all the locks from this jitter.
- Reset all accumulation stores (its incorrect for locks created pre the v4 upgrade)

## Achieving this

We want to combine all lockups for the same (owner, denom, duration) triplet, and for durations that are off by a 1hr jitter. So for every (owner, denom) pair, we want to combine all the lockups between 24hrs - 25hrs into a single 24 hr lockup, all the lockups between 7d, 7d1hr into a single 7d lockup, and all the lockups between 14d, 14d1hr into a single 14d lockup

mconcat:
If we will be only using single lock per (user, denom, duration), should we add a reverse reference from that triple to the lock ID?

ValarDragon:
That should already exist via the lockRefs I believe

PR #882, opened 2022-02-16
Lock merge migration script

mconcat:
Closes: #867 

- [X] Lockup jitter merge
- [ ] Accumulation store reset

## Description

______

For contributor use:


x/lockup/keeper/migration.go
```diff
@@ -0,0 +1,85 @@
...
+	HourDuration, _    = time.ParseDuration("1h")
+	DayDuration, _     = time.ParseDuration("24h")
+	WeekDuration, _    = time.ParseDuration("168h")
+	TwoWeekDuration, _ = time.ParseDuration("336h")
+	BaselineDurations  = []time.Duration{DayDuration, WeekDuration, TwoWeekDuration}
+)
+
+func normalizeDuration(baselineDurations []time.Duration, allowedDiff time.Duration, duration time.Duration) time.Duration {
+	for _, base := range baselineDurations {
+		if base <= duration && duration < base+allowedDiff {
```
ValarDragon, line 21:
```suggestion
// baselineDurations is expected to be sorted by the caller 
func normalizeDuration(baselineDurations []time.Duration, allowedDiff time.Duration, duration time.Duration) time.Duration {
	for _, base := range baselineDurations {
	// if base > duration, continue to next base size.
	// if base <= duration, then we are in a duration that is greater than or equal to base size.
	// If its within base + allowed diff, we set it to base.
		if base <= duration && duration < base+allowedDiff {
```

x/lockup/keeper/migration.go
```diff
@@ -0,0 +1,85 @@
...
+		normals := make(map[string]uint64)
+		for _, lock := range k.GetAccountPeriodLocks(ctx, addr) {
+			// ignore multilocks
+			if len(lock.Coins) > 1 {
+				continue
+			}
+			// ignore unlocking locks; they will be removed from the state anyway
+			if lock.IsUnlocking() {
+				continue
+			}
```
ValarDragon, line 47: Nice!

x/lockup/keeper/migration.go
```diff
@@ -0,0 +1,85 @@
...
+			coin := lock.Coins[0]
+			normalizedDuration := normalizeDuration(baselineDurations, durationDiff, lock.Duration)
+			key := addr.String() + "/" + coin.Denom + "/" + strconv.FormatInt(int64(normalizedDuration), 10)
```
ValarDragon, line 50: What is this key? Is there a standard function for it?

mconcat: key is just a triplet for (address, denom, duration). We can do
```suggestion
			key := strings.Join([]string{addr.String(), coin.Denom, strconv.FormatInt(int64(normalizedDuration), 10)}, "/")
```

ValarDragon: Oh its a key just for all the normalized locks, we don't need to parse it ever / it doesn't affect state.

ValarDragon: SGTM! Latter SGTM

mconcat:
Having a weird build error that is not reproducible on local

x/lockup/keeper/migration.go
```diff
@@ -0,0 +1,88 @@
...
+	// if base <= duration, then we are in a duration that is greater than or equal to base size.
+	// If its within base + allowed diff, we set it to base.
+		if base <= duration && duration < base+allowedDiff {
+			return base
+		}
+	}
+	return duration
+}
+
+// between duration windows of baselineDurations.map((duration) => (duration, duration+durationDiff)).
```
ValarDragon, line 31:
```suggestion
// we combine all locks between duration windows of baselineDurations.map((duration) => (duration, duration+durationDiff)) into a single new lock.
// This means every lock ID will get updated.
```

ValarDragon (approved):
Nice job on this! it looks solid to me!

PR #887, opened 2022-02-17
Modify balancer pool pb.go

mattverse:
Closes: #XXX

## Description

Patch for the proto issue we are having upon gamm generalization changes. 
This should not be a permanent solution, further discussion is needed if this is the best solution or not to the current status. 

Additional checks also need to be done as to whether `SmoothWeighChangeParams` and `BalancerPoolParams` break or not is not clear.
```
func init() {
	proto.RegisterType((*SmoothWeightChangeParams)(nil), "osmosis.gamm.poolmodels.SmoothWeightChangeParams")
	proto.RegisterType((*BalancerPoolParams)(nil), "osmosis.gamm.poolmodels.BalancerPoolParams")
	proto.RegisterType((*BalancerPool)(nil), "osmosis.gamm.v1beta1.Pool")
}
```

______

For contributor use:


x/gamm/pool-models/balancer/balancerPool.pb.go
```diff
@@ -233,7 +233,7 @@ var xxx_messageInfo_BalancerPool proto.InternalMessageInfo
 func init() {
 	proto.RegisterType((*SmoothWeightChangeParams)(nil), "osmosis.gamm.poolmodels.SmoothWeightChangeParams")
 	proto.RegisterType((*BalancerPoolParams)(nil), "osmosis.gamm.poolmodels.BalancerPoolParams")
-	proto.RegisterType((*BalancerPool)(nil), "osmosis.gamm.poolmodels.BalancerPool")
+	proto.RegisterType((*BalancerPool)(nil), "osmosis.gamm.v1beta1.Pool")
```
ValarDragon, line 236: Lets get the proto package renamed to osmosis.gamm.v1beta1 for everything.

That should 100% remove the problem for SmoothWeightChangeParams. Then this last bit for BalancerPoolParams & BalancerPool vs PoolParams & Pool is this second intense mystery, that we could punt by file renaming. (And asking Regen / cosmos SDK team when we have more time)

mattverse:
Renamed the structs from BalancerPool -> Pool as well as BalancerPoolParams -> PoolParams, whilst keeping the new folder struct. 

This also felt like a less hacky way for me, as BalancerPoolParams and BalancerPool was already in `pool-models/balancer` anyways, calling them in keeper as of `balancer.Pool`, `balancer.PoolParams`, which wouldn't cause much of a confusion.

ValarDragon:
Agreed, I think these struct names are the long term correct ones, and we'll just in the future have to migrate to swap the proto paths.

Issue #479, opened 2021-09-14
Remove all (non-community pool) module accounts from derive-account-balances


ValarDragon:
We should remove all module accounts from the produced airdrop balances. (https://github.com/osmosis-labs/osmosis/blob/main/cmd/osmosisd/cmd/balances_from_state_export.go)

Alot of the balances are from module accounts. (Unvested developer tokens, all bonded osmo, etc.). You can get the module accounts from a genesis file hackily as `cat state_export.json | grep  -A 9 -B 1 /cosmos.auth.v1beta1.ModuleAccount > module_account_list.txt`

We should parse all of those in golang, and remove them (except the community-pool) from the output derived balances.

AlpinYukseloglu:
Is there an easy way to check if an account is a module account, or should we create a list using the hacky method you mentioned and hardcode it in as a "blacklist"?

It feels as though the cleaner approach would be to have some sort of check while the lists are being generated so that we don't have to manually update the module account list if it changes (especially if superfluid staking accounts are considered modules in which case some change to the airdrop list generation process will be necessary).

ValarDragon:
Its unfortunately pretty inconvenient atm.

Basically you have to call [GetAccount](https://github.com/osmosis-labs/cosmos-sdk/blob/v0.45.0x-osmo-v7/x/auth/keeper/keeper.go#L29). Then you have to see if you can type-cast the account into a [authtypes.ModuleAccount](https://github.com/osmosis-labs/cosmos-sdk/blob/v0.45.0x-osmo-v7/x/auth/types/account.go#L184). (You should attempt type casting into `*authtypes.ModuleAccount`. It needs both

AlpinYukseloglu:
> Its unfortunately pretty inconvenient atm.
> 
> Basically you have to call [GetAccount](https://github.com/osmosis-labs/cosmos-sdk/blob/v0.45.0x-osmo-v7/x/auth/keeper/keeper.go#L29). Then you have to see if you can type-cast the account into a [authtypes.ModuleAccount](https://github.com/osmosis-labs/cosmos-sdk/blob/v0.45.0x-osmo-v7/x/auth/types/account.go#L184). (You should attempt type casting into `*authtypes.ModuleAccount`. Its best to attempt type casting into both)

[...]

PR #889, opened 2022-02-17
Remove all (non-community pool) module accounts from derive-account-balances

AlpinYukseloglu:
Closes: #479

## Description
Added logic at the end of the airdrop snapshot generation/export function that checks each account address in the snapshot to see if it's a module account and, if it is, removes it from the snapshot map before the JSON file is generated.

______

For contributor use:

### DIFF
--- a/cmd/osmosisd/cmd/balances_from_state_export.go
+++ b/cmd/osmosisd/cmd/balances_from_state_export.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/server"
@@ -12,6 +13,7 @@ import (
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/osmosis-labs/osmosis/app"
 	appparams "github.com/osmosis-labs/osmosis/app/params"
 	"github.com/osmosis-labs/osmosis/osmotestutils"
 	claimtypes "github.com/osmosis-labs/osmosis/x/claim/types"
@@ -19,8 +21,8 @@ import (
 	lockuptypes "github.com/osmosis-labs/osmosis/x/lockup/types"
 	"github.com/spf13/cobra"
 	tmjson "github.com/tendermint/tendermint/libs/json"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	tmtypes "github.com/tendermint/tendermint/types"
-	//import keeper: keeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 )
 
 const FlagSelectPoolIds = "breakdown-by-pool-ids"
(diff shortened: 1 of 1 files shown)


cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -363,7 +346,6 @@ Example:
 				//ASSUMING THAT "addr" IS THE KEY
 				delete(snapshotAccs, addr)
```
AlpinYukseloglu, line 347: @ValarDragon @mattverse Is this a correct/valid way to remove an account from the snapshotAccs map? Since we don't have a proper testing suite for this file I want to make sure I'm not using a function that breaks something

cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -316,37 +318,18 @@ Example:
...
 
-				acc := keeper.accountKeeper.GetAccount(ctx, account.Address)
+				accAddr, err := sdk.AccAddressFromBech32(account.Address)
+				if err != nil {
+					return err
+				}
+				acc := app.AccountKeeper.GetAccount(ctx, accAddr)
 
-				err := *authtypes.ModuleAccount(acc)
+				err = *authtypes.ModuleAccount(acc)
```
AlpinYukseloglu, line 332: @ValarDragon @mattverse This line isn't compiling and gives the error: "cannot convert acc (variable of type "github.com/cosmos/cosmos-sdk/x/auth/types".AccountI) to "github.com/cosmos/cosmos-sdk/x/auth/types".ModuleAccount"

The original approach [we had discussed in the issue](https://github.com/osmosis-labs/osmosis/issues/479#issuecomment-1038720585) was to attempt to typecast each account as a ModuleAccount and check if it works/throws an error. Does this compiler error here mean this approach isn't valid, or is there a way around this?

mattverse (commented):
Answered and reviewed the comments left, try type casting interface and see if that actually differentiates and sorts out the existing module accounts!

cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -315,6 +315,32 @@ Example:
 				snapshotAccs[addr] = account
 			}
 
+			queryClient := authtypes.NewQueryClient(clientCtx)
```
AlpinYukseloglu, line 318: @mattverse Is this functionally the same as `app := app.Setup(false)` from my previous commit? Am I using it correctly?

cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -315,6 +315,32 @@ Example:
...
+
+				moduleAccount, ok := acc.(*authtypes.ModuleAccount)
+				if !ok {
+					continue
+				}
+
+				moduleAccount, ok = acc.(authtypes.ModuleAccount)
+				if !ok {
+					continue
+				}
```
AlpinYukseloglu, line 337: @mattverse These checks are giving me two errors:
1. `moduleAccount is declared but not used`
2. `acc (variable of type *"github.com/cosmos/cosmos-sdk/x/auth/types".QueryAccountResponse) is not an interface`

Also, how is "ok" different from "err"? Should I be deleting from the snapshotAccs map (i.e. if the cast to `*authtypes.ModuleAccount` is successful) in these if statements instead of continuing?

cmd/osmosisd/cmd/balances_from_state_export.go
mattverse: lgtm!

cmd/osmosisd/cmd/balances_from_state_export.go
mattverse: we conventionally use ok as a variable to see if type casting is legit or not. 
if !ok, it means that type casting the interface wouldnt work. 
```suggestion
		res, err := queryClient.Account(cmd.Context(), &authtypes.QueryAccountRequest{
					Address: addr,
				})
				if err != nil {
					return err
				}
				var newAccount authtypes.AccountI
				err = clientCtx.InterfaceRegistry.UnpackAny(res.Account, &newAccount)
				if err != nil {
					panic(err)
				}
				
				moduleAccount, ok := newAccount.(*authtypes.ModuleAccount)
				if !ok {
					continue
				}
```

So basically there are two things to get fixed here. 
1. queryClient.Account is not returning an account right away, it returns account wrapped inside a proto struct called `QueryAccountResponse`. Therefore we can't call acc right away, it should be inside the response. 
2. queryClient.Account returns any, so we also have to unpack any, and then use the unpacked any to do whatsoever logic that needs to be done.

mattverse (commented):
Per yesterday's conversation, decided on preferably using app keepers directly instead of queryClinets.

cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -316,37 +318,18 @@ Example:
...
-					}
-
-					//Else, remove account from list - how do we remove an item from a map in go?
-					//ASSUMING THAT "addr" IS THE KEY
-					delete(snapshotAccs, addr)
-				}
-				//Make sure the length etc. is updated so the resulting json file generated next in the func doesn't error/generate garbage
-			*/
+			app := app.Setup(false)
+			ctx := app.BaseApp.NewContext(false, tmproto.Header{Height: 1, Time: time.Now().UTC()})
```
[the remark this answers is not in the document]
ValarDragon: Ohh, I'm just realizing I misunderstood this. uhh none of this ways are right 😅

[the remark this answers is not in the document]
ValarDragon: client context would read off of the chains state, and app here reads of a null state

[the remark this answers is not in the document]
ValarDragon: yeah, so what has to happen here, is we have to read the account from `accounts` returned by `authtypes.SanitizeGenesisAccounts(accounts)`. 

I'd recommend making a map from addr -> GenesisAccount. Then you should be able to do 
```
acc := accountsMap[addr]
if mAcc, err := acc.AccountI.(ModuleAccountI); err != nil {
 // do thing
}
```

mattverse:
@AlpinYukseloglu  Can you request for review again once this is ready for the second round of review?(couldn't know if this is ready or not)

PR #894, opened 2022-02-18
x/superfluid: Delegate bug fixes and refactors

ValarDragon:
## Description

I'm adding more docs / code changes as I go through the delegate flow and what happens during it.

This PR thus far:
- cleans up some logic in delegate / elsewhere in superfluid
- Fixes a bug where you could superfluid delegate non-allowed assets and get invalid states
- Highlights some problematic parts of the code to fix pre-v7.

Furthermore: We need to change the code to allow:
* 1 superfluid delegation per (addr, denom) pair
* many superfluid unbondings per (addr, denom) pair

Need to understand what happens on delegate & AddTokensToLock, if the validator is:
* jailed
* unbonding
* tombstoned

I suspect nothing can cause a chain halt, worst case the tx reverts? If so, then we can just cacheCtx wrap it

______

For contributor use:


x/superfluid/keeper/stake.go
```diff
@@ -231,53 +281,31 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
 	if err != nil {
 		return err
 	}
-	k.ak.SetAccount(ctx, authtypes.NewBaseAccount(mAddr, nil, 0, 0))
+	// TODO: @Dev added this hasAccount gating, think through if theres an edge case that makes it not right
+	if !k.ak.HasAccount(ctx, mAddr) {
+		// TODO: Why is this a base account, not a module account?
```
mattverse, line 286: Aren't intermediary accounts supposed to be base account not module account?

mattverse (commented):
Comments left on some of the coments!

x/superfluid/keeper/stake.go
ValarDragon: oh is there a reason for base vs module?

mattverse, line 286: I fully agree we should change to this approach!

ValarDragon: Yeah BaseAccounts are a standard account.

The questions I think of for ModuleAccounts vs BaseAccount is:
* (Primary Q) Is there a private key that can control it
* should it have different permissions than a normal account (mint/burn of an asset)
* Should people just send coins to it

ValarDragon: It is ok to have many module accounts, we just have to do some API hacks IIRC, since the base SDK doesn't give us the best methods to do it

ValarDragon:
Going to merge this, as soon we will have far more substantial changes to synth locks & regular locks

PR #902, opened 2022-02-19
Get distribution working for synthetic locks

ValarDragon:
## Description

imo we need to rework the model of synthetic lockups implemented. We want them to be something where:
* A constraint on the underlying lock, where you can't release underlying locks tokens until done. (Eventually move to a new struct)
* Only another module can release this constraint
* A lockup for a denom that doesn't actually exist on-chain
* Ability to give rewards to this synthetic denom

At some point we need to start renaming fields:
* (coins -> 'synthetic coins', or 'underlying coin restriction', or no field for the entire lock)
* suffix -> synthetic denom (and construct it in a better way)

This PR gets distribution to something I think works (untested),
simplifies the API guarantees around AddTokensToLock (managed by superfluid)

It leaves:
* UnlockTokens as also cuasing a superfluid undelegation. (I really think it should be the other way around / via multi-message, but leaving as is for now)
* Slashing as slashing the underlying real lock, and then updating synthetic locks from within lockup. (Really not sure what the right thing to be doing here, figured getting slashing core problems >> better API management)
______

For contributor use:


x/incentives/keeper/distribute.go
```diff
@@ -218,67 +222,63 @@ func (k Keeper) doDistributionSends(ctx sdk.Context, distrs *distributionInfo) e
...
+		synthLocks = append(synthLocks, *synthLock)
+	}
+
+	lockSum := lockuptypes.SumSyntheticLocksByDenom(synthLocks, denom)
+
+	if lockSum.IsZero() {
+		return nil, nil
+	}
+
+	remainCoins := gauge.Coins.Sub(gauge.DistributedCoins)
```
ValarDragon, line 248: Note to reviewer: This function from here on is ~entirely copy-paste of the normal distribute code. I tried making a common function via interface, ran into an issue so stopped looking into it. But its probably not that insightful to review in detail.

Didn't spend too much time working through the interface issue, since this changes anyway with the complete F1 code.

x/superfluid/keeper/gov/gov.go
```diff
@@ -1,13 +1,14 @@
-package keeper
+package gov
```
ValarDragon, line 1: moved here, to reduce the number of files we need to keep re-reviewing.

x/incentives/keeper/distribute.go
```diff
@@ -218,67 +222,63 @@ func (k Keeper) doDistributionSends(ctx sdk.Context, distrs *distributionInfo) e
...
+	for _, lock := range locks {
+		synthLock, err := k.lk.GetSyntheticLockup(ctx, lock.ID, suffix)
+		if err != nil {
+			k.Logger(ctx).Error(err.Error())
+			continue
+		}
+		synthLocks = append(synthLocks, *synthLock)
+	}
+
+	lockSum := lockuptypes.SumSyntheticLocksByDenom(synthLocks, denom)
```
mattverse, line 242: Aren't the locks that are passed to the parameter already synthetic locks?

mattverse (commented):
I'm assuming the main reason for the bug on distribution & gauges was that lockSum doesn't work the same for synth Lockups, thus we need to use `distributeSyntheticInternal` and get individual synth locks through iteration. 

Questions left in comments!

x/incentives/keeper/distribute.go
ValarDragon: Nope, they're the underlying period locks very confusingly

PR #911, opened 2022-02-20
Add superfluid flow test

mattverse:
Closes: #XXX

## Description

______

For contributor use:


mattverse:
Note to reviewers: this test is mostly integrations of tests for different components

x/superfluid/keeper/superfluid_test.go
```diff
@@ -0,0 +1,182 @@
...
+				Identifier:            "day",
+				StartTime:             suite.ctx.BlockTime(),
+				Duration:              time.Hour * 24,
+				CurrentEpoch:          0,
+				CurrentEpochStartTime: time.Time{},
+				EpochCountingStarted:  false,
+			}
+			suite.app.EpochsKeeper.SetEpochInfo(suite.ctx, epochInfo)
+
+			epochs.BeginBlocker(suite.ctx, *suite.app.EpochsKeeper)
```
ValarDragon, line 56: Would be ideal if we can switch to App.BeginBlock() / app.EndBlock(), so we can understand whats the pipeline of all the moving components

ValarDragon: (Somewhat worried that the order of calls in here will be different than what we actually use)

mattverse: good catch!

mattverse:
Updated test so that it uses app.BeginBlock, app.EndBlock, app.Commit

mattverse:
Closing in favor of https://github.com/osmosis-labs/osmosis/pull/1070

PR #917, opened 2022-02-20
Make superfluid correctly block a second delegation from a lockID, when unbonding

ValarDragon:
## Description

Before we could get into invalid states by superfluid delegating to a different validator while unbonding.

This PR also deletes the second UnbondingDuration param in superfluid, to ensure we always use the underlying staking's unbonding duration.

This PR should be reviewed commit by commit

TODO: At the end of superfluid things, we need to write the message flow we want, this means add token to lock is the only way to up an existing superfluid delegation, rather than SuperfluidDelegate. This should be relatively easy to fix if we want to.
______

For contributor use:


UnityChaos (approved):
Read through each commit, utACK / LGTM

:+1:

PR #921, opened 2022-02-21
Superfluid - Corner case fixes + spec improvements

UnityChaos:
Closes: #XXX

## Description

______

For contributor use:


x/superfluid/keeper/stake.go
```diff
@@ -48,13 +48,14 @@ func (k Keeper) RefreshIntermediaryDelegationAmounts(ctx sdk.Context) {
 			continue
 		}
 
+		currentAmount := sdk.NewInt(0)
 		delegation, found := k.sk.GetDelegation(ctx, mAddr, valAddress)
 		if !found {
-			k.Logger(ctx).Error(fmt.Sprintf("delegation not found for %s with %s", mAddr.String(), acc.ValAddr))
-			continue
```
ValarDragon, line 54: Why are we continuing if there is no delegation from the mAddr?

ValarDragon: ah thats current delegation, good catch

ValarDragon: I guess we do have this issue when someone fully unbonds from a validator, this will error / be at 0 each epoch.

ValarDragon (approved):

ValarDragon:
Can you fix the conflicts, then go ahead and merge?

x/superfluid/keeper/stake.go
```diff
@@ -48,13 +48,14 @@ func (k Keeper) RefreshIntermediaryDelegationAmounts(ctx sdk.Context) {
...
-		}
+			k.Logger(ctx).Error(fmt.Sprintf("Existing delegation not found for %s with %s during superfluid refresh", mAddr.String(), acc.ValAddr))
```
ValarDragon, line 54:
```suggestion
			// continue if current delegation is 0, in case its really a dust delegation 
			// that becomes worth something after refresh.
			k.Logger(ctx).Error(fmt.Sprintf("Existing delegation not found for %s with %s during superfluid refresh", mAddr.String(), acc.ValAddr))
```

PR #926, opened 2022-02-21
AddTokensToLock test 

mconcat:
Closes: #XXX

## Description

______

For contributor use:

### DIFF
--- a/x/lockup/keeper/synthetic_lock.go
+++ b/x/lockup/keeper/synthetic_lock.go
@@ -117,20 +117,19 @@ func (k Keeper) CreateSyntheticLockup(ctx sdk.Context, lockID uint64, synthDenom
 		return err
 	}
 
+	var endTime time.Time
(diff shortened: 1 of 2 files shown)


x/lockup/keeper/synthetic_lock.go
```diff
@@ -117,20 +117,19 @@ func (k Keeper) CreateSyntheticLockup(ctx sdk.Context, lockID uint64, synthDenom
 		return err
 	}
 
+	var endTime time.Time
```
ValarDragon, line 120:
```suggestion
	endTime := time.Time{}
```

ValarDragon:
Nice! This is great that it tests that the accumulation store for AddTokensToLock works, and every different method of querying the synthetic locks!

We also need tests for superfluid side, for doing AddTokensToLock on a lock with a superfluid delegation. We need to check that it increases the corresponding delegation / maintains invariants accordingly. (Though perhaps best done by making a function for all the invariants expected)

PR #930, opened 2022-02-21
Add SuperfluidUnbondLock test and cli

mattverse:
## Description

This PR adds go tests and cli for `SuperfluidUnbondLock`.
Testing for SuperfluidUnbondLock covers the following areas: 
1. test SuperfluidUnbondLock does not work before undelegation
2. test SuperfluidUnbondLock makes underlying lock start unlocking
3. test that synthetic lockup being finished does not mean underlying lock is finished
4. test after SuperfluidUnbondLock + lockup time, the underlying lock is finished

______

For contributor use:

### DIFF
--- a/x/superfluid/keeper/stake_test.go
+++ b/x/superfluid/keeper/stake_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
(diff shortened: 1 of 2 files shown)


x/superfluid/keeper/stake_test.go
```diff
@@ -448,6 +449,84 @@ func (suite *KeeperTestSuite) TestSuperfluidUndelegate() {
...
+		suite.ctx = suite.ctx.WithBlockTime(unbondLockStartTime)
+		err = suite.app.SuperfluidKeeper.SuperfluidUnbondLock(suite.ctx, lock.ID, lock.GetOwner())
+		suite.Require().NoError(err)
+
+		// check that SuperfluidUnbondLock makes underlying lock start unlocking
+		// we run WithdrawAllMaturedLocks to ensure that lock isn't getting finished immediately
+		suite.app.LockupKeeper.WithdrawAllMaturedLocks(suite.ctx)
+		updatedLock, err := suite.app.LockupKeeper.GetLockByID(suite.ctx, lock.ID)
+		suite.Require().NoError(err)
+		suite.Require().False(lock.IsUnlocking())
```
ValarDragon, line 507:
```suggestion
```

That line is a no-op, since its referencing the prior copy

ValarDragon (approved):
Nice!

x/superfluid/keeper/stake_test.go
```diff
@@ -448,6 +449,84 @@ func (suite *KeeperTestSuite) TestSuperfluidUndelegate() {
...
+		suite.Require().NoError(err)
+		suite.Require().True(updatedLock.IsUnlocking())
+
+		// test after SuperfluidUnbondLock + lockup unbonding duration, lock is finished and does not exist
+		suite.ctx = suite.ctx.WithBlockTime(unbondLockStartTime.Add(unbondingDuration))
+		suite.app.LockupKeeper.WithdrawAllMaturedLocks(suite.ctx)
+		updatedLock, err = suite.app.LockupKeeper.GetLockByID(suite.ctx, lock.ID)
+		fmt.Println(updatedLock)
+		suite.Require().Error(err)
+
```
ValarDragon, line 526: We need to check that the users underlying balance is increased after the withdraw. (exactly as expected)

There was previously weird bugs where the locks underlying denom was changed on the synthetic lockup removal.

ValarDragon:
Merging now to unblock @czarcas7ic . Need to come back and update the test to check coins balance.

On further thought, I'm not as worried about the multi-message test. We can file an issue for it if you want to deal with it later.

We do have to do the check coins balance pre v7 though.

PR #931, opened 2022-02-21
removed superfluid change, uses 60s day epoch rather than min epoch

czarcas7ic:
This needs to be changed now that we are no longer using `unbonding_duration`

Also switched to using day epoch. Ran into issues when using min on recent superfluid changes.

______

For contributor use:


czarcas7ic:
@antstalepresh I see you made a PR for this, did you have issues as well when using the min epoch counter? If it was only me I can scrap this

gabriellahayden:
Are you still having issues using the min epoch counter?

ValarDragon (approved):

PR #934, opened 2022-02-21
Wasmd updates

ethanfrey:
## Description

Re-reviewing the wasmd changes, and adding a few little pieces.

______

For contributor use:

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -3,14 +3,16 @@ package app
 import (
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
+	ante "github.com/cosmos/cosmos-sdk/x/auth/ante"
 	"github.com/cosmos/cosmos-sdk/x/auth/signing"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 
-	ante "github.com/cosmos/cosmos-sdk/x/auth/ante"
-
 	channelkeeper "github.com/cosmos/ibc-go/v2/modules/core/04-channel/keeper"
 	ibcante "github.com/cosmos/ibc-go/v2/modules/core/ante"
+
+	wasm "github.com/CosmWasm/wasmd/x/wasm"
+	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
+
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/osmosis-labs/osmosis/v7
 go 1.17
 
 require (
-	github.com/CosmWasm/wasmd v0.22.0
+	github.com/CosmWasm/wasmd v0.23.0
```
ValarDragon, line 6: We replace the wasmd import due to this commit for our IAVL changes: https://github.com/osmosis-labs/wasmd/commit/fced0452df2f789a871da41e4d97e62998f51a7c

(Really silly API change)

Should we do this for v0.23.0 instead of v0.22.0?

ValarDragon (approved):

ValarDragon:
Please feel free to merge once you've looked at the v0.23.0 vs v0.22.0, lmk if we should make a new branch/tag cherrypicking the commit off of v0.23.0 for the IAVL patch

go.mod
ethanfrey: Good to know. I didn't dig into the replace blocks, but was surprised go.sum didn't update.

There is no different in x/wasm between 0.22 and 0.23, it is in app, which you don't import. One of those bugfix version bumps.

In general nice to track the latest, but not urgent.

ValarDragon: Thanks for the info! Going ahead and merging this then

app/ante.go
```diff
@@ -29,6 +32,7 @@ func NewAnteHandler(
 	mempoolFeeDecorator := txfeeskeeper.NewMempoolFeeDecorator(*txFeesKeeper, mempoolFeeOptions)
 	return sdk.ChainAnteDecorators(
 		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
+		wasmkeeper.NewLimitSimulationGasDecorator(wasmConfig.SimulationGasLimit),
```
alpe, line 35: @ethanfrey the [`CountTXDecorator`](https://github.com/CosmWasm/wasmd/blob/master/app/ante.go#L49) is missing in the list. The position data is passed to the contracts via [env](https://github.com/CosmWasm/wasmvm/blob/main/types/env.go#L13) and should be provided to be fully compatible.

PR #936, opened 2022-02-21
mint developer rewards supply offset

sunnya97:
Closes: #XXX

## Description

______

For contributor use:

### DIFF
--- a/x/mint/keeper/keeper.go
+++ b/x/mint/keeper/keeper.go
@@ -176,6 +176,11 @@ func (k Keeper) DistributeMintedCoin(ctx sdk.Context, mintedCoin sdk.Coin) error
 	if err != nil {
 		return err
 	}
+
+	// Take the current balance of the developer rewards pool and remove it from the supply offset
+	developerAccountBalance := k.bankKeeper.GetBalance(ctx, k.accountKeeper.GetModuleAddress(types.DeveloperVestingModuleAcctName), mintedCoin.Denom)
+	k.bankKeeper.AddSupplyOffset(ctx, mintedCoin.Denom, developerAccountBalance.Amount)
+
 	if len(params.WeightedDeveloperRewardsReceivers) == 0 {
 		// fund community pool when rewards address is empty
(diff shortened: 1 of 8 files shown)


x/mint/keeper/keeper.go
```diff
@@ -176,6 +185,11 @@ func (k Keeper) DistributeMintedCoin(ctx sdk.Context, mintedCoin sdk.Coin) error
 	if err != nil {
 		return err
 	}
+
+	// Take the current balance of the developer rewards pool and remove it from the supply offset
+	developerAccountBalance := k.bankKeeper.GetBalance(ctx, k.accountKeeper.GetModuleAddress(types.DeveloperVestingModuleAcctName), mintedCoin.Denom)
+	k.bankKeeper.AddSupplyOffset(ctx, mintedCoin.Denom, developerAccountBalance.Amount)
```
ValarDragon, line 191: Why doesn't this make the supply offset go infinite (adding the entire dev vesting balance each epoch)? It should just be adding the `devRewardCoin.Amount` to the supply offset.

ValarDragon: oh your adding the whole current balance, and then subtracting the new balance after distribution?

Seems overly complex versus devRewardCoin adding. Is there a rounding error for why we do this?

sunnya97: Seemed like the easiest way to make sure it's always in sync with the balance of the Developer Rewards Account

x/mint/keeper/keeper.go
```diff
@@ -176,6 +185,11 @@ func (k Keeper) DistributeMintedCoin(ctx sdk.Context, mintedCoin sdk.Coin) error
 	if err != nil {
 		return err
 	}
+
+	// Take the current balance of the developer rewards pool and remove it from the supply offset
```
ValarDragon, line 189:
```suggestion
	// Take the current balance of the developer rewards pool and remove it from the supply offset
	// We add back the new supply at the end, in order to avoid any rounding discrepancies.
```

ValarDragon (approved):

PR #937, opened 2022-02-21
Work on rewards pipeline changes

ValarDragon:
Due to safety / correctness / complexity issues, we are adapting the rewards distribution pipeline for superfluid staking rewards to be correct / much simpler.

______

For contributor use:


UnityChaos (approved):
utACK / LGTM.

Only question I have is if we want to try and move the `UpdateOsmoEquivalentMultipliers` / `RefreshIntermediaryDelegationAmounts` back into the actual epoch hook, so that the values minted are based on "now" rather than minting rewards based on 1 epoch ago prices.

ValarDragon:
Discussed IRL, seemed like we agree on for now doing what were currently doing. (Rewarding delegators for security already provided, rather security about to be provided)

Unsure if we agree on the same long term (I think the current is correct)

PR #962, opened 2022-02-23
Change balancer package to use balancer codec

mattverse:
## Description

the balancer package within the gamm module currently uses codec for gamm, should be changed in the way that it uses its own codec

______

For contributor use:


ValarDragon (approved):
Not sure I get this, but we should try on a new testnet

Issue #447, opened 2021-08-25
x/lockup, restructure and reduce the lockrefs


ValarDragon:
Currently x/lockups has many ways of iterating over locks / stores of references to a lock, and consequently repeated state. A lot of this is redundant / not needed imo, and should now be optimized out. (I think this will help alot with removing surplus code)

Here is the current list of references we store:

1. `{KeyPrefixLockTimestamp}{LockEndTime}`
2. `{KeyPrefixLockDuration}{Duration}`
3. `{KeyPrefixAccountLockTimestamp}{Owner}{LockEndTime}`
4. `{KeyPrefixAccountLockDuration}{Owner}{Duration}`
5. `{KeyPrefixDenomLockTimestamp}{Denom}{LockEndTime}`
6. `{KeyPrefixDenomLockDuration}{Denom}{Duration}`
7. `{KeyPrefixAccountDenomLockTimestamp}{Owner}{Denom}{LockEndTime}`
8. `{KeyPrefixAccountDenomLockDuration}{Owner}{Denom}{Duration}`

This means there are 8 separate sub-trees here. I don't think that subtrees `1-4` have any use case. All their use cases are taken over by `5-8`. I see no reason why we wouldn't just iterate over all denominations when we want all the denominations, using iterators.

I think we can also better combine the remaining lockrefs. I think we should store it as:
5. `{KeyPrefixLocksByDenom}/{Denom}/Bonded/{Duration}`
6. `{KeyPrefixLocksByDenom}/{Denom}/Unbonding/{EndTime}`
7. `{KeyPrefixLocksByAccount}/{Owner}/{Denom}/Bonded/{Duration}`
8. `{KeyPrefixLocksByAccount}/{Owner}/{Denom}/Unbonding/{EndTime}`

This enables better iterability over the lock references, and I think is generally clearer. Thoughts?

antstalepresh:
Currently those do not have any use-cases yet - if we won't have use-cases for the top 4 in the future as well, we can remove those keys and add if needed in the future.

mattverse:
One question I have is if we fix the key prefixes and refKeys, how would it affect existing stores?

ValarDragon:
yes, changes to refkey structure require migration code

ValarDragon:
I think we can close this issue tbh, and just deal with it all as part of refactoring for the ORM module once its ready

PR #964, opened 2022-02-24
Make creation of a bonded lock not create unused time keys

ValarDragon:
## Description

Cref #447 

Basically for a long time, we've known that we've been adding extra keys to bonding locks that are useless. This removes the creation of those useless ref keys, helping reduce the write load of our migration.

We are still deleting all refkeys, even if they don't exist. This is because we shouldn't keep any ref keys in state that aren't valid.

______

For contributor use:

### DIFF
--- a/x/lockup/keeper/lock_refs.go
+++ b/x/lockup/keeper/lock_refs.go
@@ -25,9 +25,7 @@ func (k Keeper) deleteLockRefs(ctx sdk.Context, lockRefPrefix []byte, lock types
 		return err
 	}
 	for _, refKey := range refKeys {
-		if err := k.deleteLockRefByKey(ctx, combineKeys(lockRefPrefix, refKey), lock.ID); err != nil {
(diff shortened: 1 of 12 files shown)


x/lockup/keeper/lock_refs.go
```diff
@@ -6,7 +6,10 @@ import (
 )
 
 func (k Keeper) addLockRefs(ctx sdk.Context, lock types.PeriodLock) error {
-	refKeys, err := lockRefKeys(lock)
+	refKeys, err := durationLockRefKeys(lock)
+	if lock.IsUnlocking() {
+		refKeys, err = lockRefKeys(lock)
```
mattverse, line 11: imo, I would prefer using something like `unlockingRefKeys` instead of `lockReftKeys` but its a minor thing, totally up to you!

mattverse (approved):
LGTM!

PR #978, opened 2022-02-25
v7 Upgrade Guide

czarcas7ic:
Guide for upgrading from v6 to v7.

I would not merge until the v7.0.0 tag is made, since this guide references this tag for people to prep their cosmovisor with.

______

For contributor use:

### DIFF
--- a/networks/osmosis-1/upgrades/v7/guide.md
+++ b/networks/osmosis-1/upgrades/v7/guide.md
@@ -1,0 +1,105 @@
+# v6 to v7 Upgrade Guide
+
+
+## Memory Requirements
+
+For this upgrade, nodes will need a total of 64GB of memory. This must consist of a **minimum** of 32GB of RAM, while the remaining 32GB can be swap. For best results, use 64GB of physical memory.
+
+
+Short version swap setup instructions:
+```sh
+sudo swapoff -a
+sudo fallocate -l 32G /swapfile
+sudo chmod 600 /swapfile
+sudo mkswap /swapfile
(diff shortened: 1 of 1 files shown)


networks/osmosis-1/upgrades/v7/guide.md
```diff
@@ -0,0 +1,105 @@
...
+To install Cosmovisor:
+
+```sh
+cd $HOME
+git clone https://github.com/cosmos/cosmos-sdk
+cd cosmos-sdk
+git checkout v0.44.0
+make cosmovisor
+cp cosmovisor/cosmovisor $GOPATH/bin/cosmovisor
+cd $HOME
```
ValarDragon, line 44: I think we can now do `go install github.com/cosmos/cosmos-sdk/cosmovisor/cmd/cosmovisor@v1.0.0` right?

czarcas7ic: You are correct, just tested it out and it worked (got an error log when starting cosmovisor `failed to read error="lstat /root/.osmosisd/cosmovisor/current/upgrade-info.json: no such file or directory"`, but it still works. I think this because I didn't fill out the "info" portion of the software upgrade prop on testnet?)

ValarDragon: Yeah I think thats why you get that error message as well.

We should test with something in the field, I believe theres some JSON structure we have to adhere to.

ValarDragon (approved):
Going to merge, we can do updates to it on main

PR #986, opened 2022-02-25
lock migration fix

czarcas7ic:
As discovered by @hleb-albau , the lockup migration does not work with this log entry. We decided to simply comment it out for now.

______

For contributor use:

### DIFF
--- a/x/lockup/keeper/migration.go
+++ b/x/lockup/keeper/migration.go
@@ -126,8 +126,8 @@ func MergeLockupsForSimilarDurations(
 
 			// don't call hooks, tokens are just moved from a lock to another
 		}
-		numLocksNormalized := len(locksToNormalize)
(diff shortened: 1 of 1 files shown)


ValarDragon:
Thank you @hleb-albau and @czarcas7ic for the fix!

Super embarassing that this untested log line is what causes to have to bump numbers.

I'm extremely sorry all, this is my fault for not testing this logging patch I added without migration testing. Albeit, its crazy that the logger panics for this, the logger is chiefly able to deal with this by logging an error. (its the one with the job of communicating such errors!!)

Thankfully this doesn't affect state, and happens well before IAVL commit happens, so even if people run into it, its not an issue. (It'd crash, they'd update the binary, and it it'd work just as before)

ValarDragon (approved):

PR #1007, opened 2022-03-01
clean go.mod

faddat:
## Description

Updates go.mod so that:

* tm-db is at v0.6.7
* we no longer need a replace statement for gorocksdb
* tendermint is at v0.34.16

______

For contributor use:


ValarDragon (approved):
Thanks for this! Closing #1007 in favor of this!

Copying my comment from there. This seems good to me looking at the TM changelog https://github.com/tendermint/tendermint/blob/v0.34.16/CHANGELOG.md#v0.34.16

But would appreciate someone more informed on the TM changes to opine. (In particular that https://github.com/tendermint/tendermint/pull/7885 isn't problematic with our app version being initially set wrong) Will leave this open for another day in case someone has thoughts on this.

The tm-db change LGTM

go.mod
```diff
@@ -126,18 +127,16 @@ require (
...
 	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.45.1-0.20220221115656-75545ea8cb2d
-
+	// Use Osmosis fast iavl
 	github.com/cosmos/iavl => github.com/osmosis-labs/iavl v0.17.3-osmo-v3
 	// Use osmosis fork of ibc-go
 	github.com/cosmos/ibc-go/v2 => github.com/osmosis-labs/ibc-go/v2 v2.0.2-osmo
 	// use cosmos-compatible protobufs
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	// enable building wihh rocksdb
-	github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
```
marbar3778, line 140: you may still need this, have you tried running without it?

marbar3778:
> (In particular that [tendermint/tendermint#7885](https://github.com/tendermint/tendermint/pull/7885) isn't problematic with our app version being initially set wrong)

this is a partial fix to the app version issue, the other fix is in the sdk. We are still waiting on a test for it.

go.mod
faddat: Let me check, thanks

faddat:
@marbar3778 I've tested this with both the -tags and the ldflag and it works for compiling with rocksdb.

PR #1046, opened 2022-03-05
Apptesting: Change Interface to Struct

mattverse:
## Description

Extension for pull https://github.com/osmosis-labs/osmosis/pull/1008.
We should compare this with the interface approach and compare on usability
______

For contributor use:

### DIFF
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -19,70 +19,70 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
-// TODO: Consider an embedded struct here rather than an interface
-type SuiteI interface {
-	GetSuite() *suite.Suite
-	GetCtx() sdk.Context
-	SetCtx(sdk.Context)
-	GetApp() *app.OsmosisApp
+type KeeperTestHelper struct {
+	Suite suite.Suite
+
+	App *app.OsmosisApp
+	Ctx sdk.Context
 }
 
-func SetupValidator(suite SuiteI, bondStatus stakingtypes.BondStatus) sdk.ValAddress {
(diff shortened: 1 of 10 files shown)


mattverse:
@ValarDragon this is just a draft as of now, can you review and see if this was similar to the approach you were thinking of?

x/superfluid/keeper/keeper_test.go
```diff
@@ -27,6 +27,7 @@ import (
 
 type KeeperTestSuite struct {
 	suite.Suite
+	helper apptesting.KeeperTestHelper
```
ValarDragon, line 30: Can we remove the embedded `suite.Suite` and just make the `apptesting.KeeperTestHelper` embedded here?

app/apptesting/test_suite.go
```diff
@@ -19,70 +19,70 @@ import (
...
 )
 
-// TODO: Consider an embedded struct here rather than an interface
-type SuiteI interface {
-	GetSuite() *suite.Suite
-	GetCtx() sdk.Context
-	SetCtx(sdk.Context)
-	GetApp() *app.OsmosisApp
+type KeeperTestHelper struct {
+	Suite suite.Suite
```
ValarDragon, line 23: Can we make the suite.Suite embedded here?

x/superfluid/keeper/keeper_test.go
ValarDragon: (or try how that makes things)

mattverse: gotcha, I like that approach as well!

ValarDragon (approved):
LGTM! I like the struct approach much more than the initial interface idea

PR #1098, opened 2022-03-15
Rrefresh intermediary accounts test

antstalepresh:
Closes: #XXX

## Description

______

For contributor use:


x/superfluid/keeper/stake_test.go
```diff
@@ -663,334 +659,253 @@ func (suite *KeeperTestSuite) TestSuperfluidUnbondLock() {
...
-// 				// check unbonded amount is removed after refresh operation
-// 				refreshed := suite.App.BankKeeper.GetBalance(suite.Ctx, expAcc.GetAccAddress(), sdk.DefaultBondDenom)
-// 				suite.Require().True(refreshed.IsZero())
-// 			}
-// 		})
-// 	}
-// }
+			for index, intAccIndex := range tc.checkAccIndexes {
+				expAcc := intermediaryAccs[intAccIndex]
+				valAddr, err := sdk.ValAddressFromBech32(expAcc.ValAddr)
```
mattverse, line 830: Can we group two for loops together for code sanity? Something like this:
```
	for index, intAccIndex := range tc.checkAccIndexes {
				expAcc := intermediaryAccs[intAccIndex]
				multiplier, ok := multiplierByDenom[expAcc.Denom]
				if !ok {
					multiplier = originMultiplier
				}

				targetDelegation := intermediaryDels[index].Mul(multiplier).Quo(originMultiplier)
				lpTokenAmount := sdk.NewInt(1000000)
				decAmt := multiplier.Mul(lpTokenAmount.ToDec())
				asset := suite.App.SuperfluidKeeper.GetSuperfluidAsset(suite.Ctx, expAcc.Denom)
				targetAmount := suite.App.SuperfluidKeeper.GetRiskAdjustedOsmoValue(suite.Ctx, asset, decAmt.RoundInt())

				valAddr, err := sdk.ValAddressFromBech32(expAcc.ValAddr)
				suite.Require().NoError(err)

				delegation, found := suite.App.StakingKeeper.GetDelegation(suite.Ctx, expAcc.GetAccAddress(), valAddr)
				if targetAmount.IsPositive() {
					suite.Require().True(found)
					suite.Require().Equal(delegation.Shares, targetDelegation)
				} else {
					suite.Require().False(found)
				}
			}
```

mattverse (approved):
Minor reviews for code sanity, otherwise looks good! 
The maths to fix this test would have been painful haha

x/superfluid/keeper/stake_test.go
antstalepresh: Merged two loops into one - thanks for finding this

x/superfluid/keeper/stake_test.go
```diff
@@ -663,334 +659,244 @@ func (suite *KeeperTestSuite) TestSuperfluidUnbondLock() {
...
+			}
 
-// 			panic("e")
-// 		})
-// 	}
-// }
+			// refresh intermediary account delegations
+			suite.NotPanics(func() {
+				suite.App.SuperfluidKeeper.RefreshIntermediaryDelegationAmounts(suite.Ctx)
+			})
```
ValarDragon, line 891: How would this panic? Isn't the sensitive logic all wrapped in an `ApplyIfNoErr` function?

x/superfluid/keeper/stake_test.go
```diff
@@ -663,334 +659,244 @@ func (suite *KeeperTestSuite) TestSuperfluidUnbondLock() {
...
+func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
+	testCases := []struct {
+		name                string
+		validatorStats      []stakingtypes.BondStatus
+		delegatorNumber     int
+		superDelegations    []superfluidDelegation
+		roundOneMultipliers []osmoEquivalentMultiplier
+		roundTwoMultipliers []osmoEquivalentMultiplier
+		checkAccIndexes     []int64
+	}{
```
ValarDragon, line 671: I don't understand what this test is doing. It seems like it applies two refreshes for osmo equivalent TWAPs, and then at the end checks that every account index provided has 0 osmo in it?

How is this checking that the refresh operation applied as expected?

x/superfluid/keeper/stake_test.go
antstalepresh: This won't panic - previously there was a case for panics - I will remove it

x/superfluid/keeper/stake_test.go
antstalepresh: It checks if intermediary account does not have balance after refresh operation.
And it is checking if the operation is applied by checking the delegation amount change.

```
targetDelegation := intermediaryDels[index].Mul(multiplier).Quo(originMultiplier)
```

Here `targetDelegation` is expected delegation after refresh.

alexanderbez (approved):

x/superfluid/keeper/stake_test.go
ValarDragon: Ohh I see, I very much misunderstood this on first read. Thanks for updating the name

PR #1107, opened 2022-03-16
Update wasmd to v0.24.0

ValarDragon:
Update wasmd to v0.24.0

cc @nikever , this will require updating the wasmvm version for main

______

For contributor use:


ValarDragon:
Going to go ahead and merge to give people working on Osmosis a better development UX if their on an M1 mac.

This updates the wasmvm version from `v1.0.0-beta5` to `v1.0.0-beta7`. It appears to me that the Dockerfile is already on beta7, which means it would be building a binary with an incompatible wasmvm version from mainnet. And therefore would get a state error when the first cosmwasm contract gets on mainnet. But also means that no changes are needed to get it working for this!

ethanfrey:
Not needed, but you can also test builds on Darwin arm64 if you want to try...
Just made that PR https://github.com/osmosis-labs/osmosis/pull/1109

Issue #1100, opened 2022-03-15
CI: Make PRs not run tests twice on each commit


ValarDragon:
Right now CI runs tests twice for every commit on a PR. One for (push) and one for (pull request). We actually often get into a situation where the push test is done, but we're still waiting on the pull request CI test.

Is it possible to de-duplicate this? Perhaps only having tests on commit? (Do we lose anything by doing that?)

nikever:
As you said we run the `test` workflow on any push to any branch and any pull_request event: 

```yaml

# Event that triggers the workflow
on:
  pull_request:           # any pull_request event:
  push:                   # any push to any branch 
    branches:
      - '**'	
```

Possible solutions to avoid duplicated runs are:

1) Don't run  the workflow on `pull_request` event:

```yaml
on:
  push:
    branches:
      - '**'	
```

This is super simple to do but doing so we lose the codecov report comment on PRs as in  https://github.com/osmosis-labs/osmosis/pull/1098#issuecomment-1067954796. 
I think that the report would still will be accessible inspecting the workflow run.

2) A more complicated solution to avoid running both on `push` and on `pull_request` would be to add a check before running the job using https://github.com/fkirc/skip-duplicate-actions. 
It would complicate a little bit the workflow and might need some tweaking to find the right configuration but it's doable. More info: https://github.com/fkirc/skip-duplicate-actions#example-1-skip-entire-jobs

----

If we don't care much about the codecov report on the comment I would go with `1` which is super simple to do. Otherwise I'll start to play around with https://github.com/fkirc/skip-duplicate-actions.

What do you think?

ValarDragon:
I think 2 would be nice. I don't really use codecov that often, but it seems like a good thing to do / incorporate more.

PR #1114, opened 2022-03-18
Add skip-duplicate-actions action

nikever:
Closes: [#1100](https://github.com/osmosis-labs/osmosis/issues/1100)

## Description

This PR improves the `test` workflow by skipping duplicate actions.
This is done using [skip-duplicate-actions](https://github.com/fkirc/skip-duplicate-actions).

The improvements made:

### 1. Avoid running the `test` workflow twice on `pull_request` and `push` events

> ⚠️ **IMPORTANT**
> Test will **ALWAYS** run on `pull_requests` events in any case. 

This means that:

If Alice is working on branch `A` and commits some changes to the code, then the `test` workflow will be triggered. 
   If Alice waits for the `test` workflow to be over (the expected behaviour) and then opens a PR to merge the branch to main, the same `test` workflow will run again. 
   This is good because you are sure that any new PR code is always tested, and it is necessary to run a `pull_request` workflow to let codecov post the report in the comment.

The "downside" of this:

If Alice is working on branch `A` and commits some changes to the code, then the `test` workflow will be triggered. 
If Alice doesn't wait for the `test` workflow to be over and opens the PR immediately, **two duplicated jobs** will run (one for the `push` event and one for the `pull_request` event).
  This is the only case (which I am aware of) in which there will be duplicated runs of the `test` workflow.

The good part:

- Any following commits to the PR will only RUN once. The workflow triggered by the `push` event will be skipped.
- When the PR gets merged, the `test` workflow is skipped again. 

## 2. Run the workflow only when go files are edited

This is a nice optimization to avoid running tests when it's not needed.
If Alice commits don't change any: `*.go` `*.mod` and  `*.sum`, then the `test` workflow will be skipped.

This is controlled by https://github.com/nikever/osmosis/blob/fix/remove-duplicate-ci-trigger/.github/workflows/test.yml#L26

Am I missing any relevant scenario where we would like to run tests?

In `pull_requests` tests will always run. 

## 3. Cancel outdated runs

This is another optimization, which is slightly more "aggressive" than the other. 

Let me know if you want to keep it:

This is controlled by https://github.com/nikever/osmosis/blob/fix/remove-duplicate-ci-trigger/.github/workflows/test.yml#L23

If Alice is working on branch `A` and commits some changes to the code, then the `test` workflow will be triggered. 
While the workflow is running, if Alice commits another change to the same branch, the previous `test` workflow **will be cancelled** in favour of the new `test` workflow. 

## Before merging 

I have run multiple tests in my local repository to test different scenarios, but CIs are always tricky, and it would be great if someone from the dev team could confirm that the workflow behaves as expected.

______

For contributor use:


nikever:
And (as before) with PRs coming from remote forks such as this one now the `test` workflow is not triggered twice since there are no `push` events.

I will update the `CHANGELOG.md` so we can check the behaviour.

nikever:
Even if I have modified only the `CHANGELOG.md` and not any `*.go` files the workflow still runs.
This might be unnecessary but I think it's **good** practise for always avoid merging *possibly* untested code.

p0mvn:
In 1) you mentioned a case where it is possible to have 2 duplicate jobs on open PRs.

I see that we have `concurrent_skipping: 'same_content'`. Does this not help mitigate that duplication?

I'm assuming the above setting is overwritten by  `do_not_skip: '["pull_request",...]` ?

ValarDragon (approved):
Awesome! Thanks for adding this and the extremely detailed description

nikever:
> I see that we have `concurrent_skipping: 'same_content'`. Does this not help mitigate that duplication?
> 
> I'm assuming the above setting is overwritten by  `do_not_skip: '["pull_request",...]` ?

It doesn't help unfortunately. You can skip jobs that gets triggered simultaneously.
In the case of scenario 1, the events happen sequentually (First you push, then you open the PR).

The solution would be to cancel the first `push` triggered workflow when a corresponding `pull_request` triggered workflow runs.

However this is not super simple to do but I don't think this would be necessary since:

1. Ideally, you should wait for the `push` triggered workflow to complete, before opening a PR

2. Both workflow runs only PR opens and it takes some time to review the PR.

ValarDragon:
I think the duplicate push run when opening PR isn't worth worrying about

The big value add was improving the situation for every subsequent push, as typically we merge some merge conflict before wanting to hit approve, and thats the flow breaking latency, which this PR solves!

p0mvn (approved):

PR #1125, opened 2022-03-22
chore: app package housekeeping

alexanderbez:
## Description

Some general linting and housekeeping to the `app` package. Changes are meant to be non-contentious and ultimately make the code and style consistent and most importantly legible.

Feel free to close/deny if deemed too contentious :)

______

For contributor use:


app/app.go
```diff
@@ -41,127 +23,110 @@ import (
...
+	"github.com/rakyll/statik/fs"
+	"github.com/spf13/cast"
+	abci "github.com/tendermint/tendermint/abci/types"
+	tmjson "github.com/tendermint/tendermint/libs/json"
+	"github.com/tendermint/tendermint/libs/log"
+	tmos "github.com/tendermint/tendermint/libs/os"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
 
-	// Osmosis application prarmeters
```
ValarDragon, line 71: Do you think we should revert these comments / the prior ordering of imports here?

app/upgrades/v3/forks.go
```diff
@@ -14,32 +16,31 @@ func RunForkLogic(ctx sdk.Context, gov *govkeeper.Keeper, staking *stakingkeeper
 	FixMinCommisionRate(ctx, staking)
 }
 
-// Fixes an error where minimum deposit was set to "500 osmo"
-// This denom does not exist, which makes it impossible for a proposal to go to a vote
+// Fixes an error where minimum deposit was set to "500 osmo". This denom does
+// not exist, which makes it impossible for a proposal to go to a vote.
```
ValarDragon, line 20: Is there a linter your running for these, that we should add to the linter in CI? (Noticed a few period adds, and words being moved to a prior line. I imagine they were automated)

app/app.go
alexanderbez: You mean keep them?

Import blocks aren't meant to be annotated. If you think devs would really find them useful, I can revert :)

app/upgrades/v3/forks.go
alexanderbez: I wrap comments at 80 chars which you can have enabled in your editor. There is also a linter for this but Im sure it would be super noisy if we enabled it (i.e. a lot of triggers on the current codebase).

ValarDragon: Gotcha, that makes sense! Sounds great to me, just didn't realize you were manually doing it haha

app/app.go
ValarDragon: @faddat was the one who added them, presumably he's found it helpful when explaining app.go's to people.

We could put the same comments in the AppKeepers struct though, I don't think theres anything particularly important about this info being in the imports

ValarDragon (approved):
LGTM! I'm fine with removing the import annotations, and moving the equivalent annotations elsewhere in structs directly

alexanderbez:
@ValarDragon can you merge this whenever you get a chance?

PR #1127, opened 2022-03-22
A much smaller set of newly enabled linters

faddat:
## Description

This PR enables many more linters, but not nearly as many as the one that it is replacing.

The reason it hits so many files is this:

https://github.com/mvdan/gofumpt

From here, we can re-enable additional linters in .golangci.yml as needed.

______

For contributor use:


.golangci.yml
```diff
@@ -4,23 +4,98 @@ run:
 #   timeout: 5m
 
 linters:
+  # Disable all linters.
+  # Default: false
```
alexanderbez, line 8: nit: I would remove any commented out sections unless we intend to re-enable them in the future.

```suggestion
```

app/upgrades/v5/whitelist_feetokens.go
```diff
@@ -13,6 +13,7 @@ import (
 // Every asset with a liquid osmo pairing pool on Osmosis, as of 12/01/21
 // Notably, Tick is not on this list because the osmo pool has $76 of liquidity.
 // Cheq'd and KRT are also not on this, due to neither having osmo pairings.
+//nolint:gosec // nolint because these are strings of whitelisted ibc denoms
```
alexanderbez, line 16:
```suggestion
// We nolint because these are strings of whitelisted ibc denoms.
//nolint:gosec
```

alexanderbez (approved):
Skimmed through a bunch of the changes and it LGTM. I left one minor comment on the config file.

Also, this might clobber a bit with the removed annotations in the import blocks in my PR #1125, but I can resolve conflicts if need be.

ValarDragon (approved):
LGTM as well! Thanks for pairing this down from the prior PR. RIP I made merge conflicts with merging #1129 , I'll fix the conflicts here

ValarDragon:
Actually, if its ok with you all, I'd really like to disable godot. (We can keep the changes)

`godot` just feels like an unnecessary headache. Also we need to update `make lint` to have parity with this

ValarDragon:
Latest commit:
* Updates `make lint` to use the config file
* removes godot (indefinitely if noone has other thoughts)
* removes gofumpt (only until we get make lint auto-format integration sorted & vscode / makefile installation worked out)

alexanderbez:
Sounds good to me @ValarDragon

faddat:
this all sounds good to me.   Just updating changelog now.

PR #1133, opened 2022-03-23
refactor: lint CI target

alexanderbez:
## Description

- Introduce `tools` methodology  to managing Go-based tool/deps (Tendermint and Umee do this). This allows a developer to avoid having to keep their deps up to date.
- Use tools-based tools in make lint target

______

For contributor use:


Makefile
```diff
@@ -231,8 +231,8 @@ benchmark:
 ###############################################################################
 
 lint:
-	golangci-lint run
-	find . -name '*.go' -type f -not -path "./vendor*" -not -path "*.git*" | xargs gofmt -d -s
```
ValarDragon, line 235: Should we keep the find for auto `gofmt`'ing?

alexanderbez: Shouldn't "lint" be idempotent and "stateless"? Like shouldn't the `format` target do formatting?

ValarDragon: oh, I actually didn't realize `make format` existed. I think your totally right!

Issue #1029, opened 2022-03-03
[Epic] Token Factory Module


ValarDragon:
The token factory module is a module that will enable new native tokens to be minted on Osmosis, with well prefixed names. This is important for cosmwasm applications on osmosis to thrive.

- [x] #584 
  - Get initial PR merged 
- [x] Disable force transfer code
- [x] #689 
- [x] Add friction (base osmo fee) to token factory token creation #1435
- [x] Code spec
- [ ] Test usage in cosmwasm contract / provide example contract making use of this

sunnya97:
We should add checks that prevent force transferring or burning from module accounts.

sunnya97:
Let's go ahead and reduce the scope such that minting and burning can only happen in the triggering account.

p0mvn:
I'm cleaning up the epics and seeing some checkboxes without issues. @sunnya97 @ValarDragon could you update the epic to cross out the completed items and create issues for items that still need work. Thank you!

mattverse:
Closing this issue to replace with issues with more details and specifics

PR #1135, opened 2022-03-23
Custom CosmWasm Integration

ethanfrey:
## Description

This follows up on the earlier CosmWasm integration work and adds custom bindings for Osmosis AMM functionality.

## Contract Bindings

The other half of the work lives in https://github.com/confio/osmosis-bindings (which we are happy to transfer to the osmosis org when this is merged). If you import `osmo-bindings` in your contract, it will get an Osmosis-specific feature flag (so it will be rejected on upload by other chains), and gets access to some custom functionality. In particular, [messages](https://github.com/confio/osmosis-bindings/blob/main/packages/bindings/src/msg.rs#L11-L28) to Swap or Mint native tokens. As well as [various queries](https://github.com/confio/osmosis-bindings/blob/main/packages/bindings/src/query.rs#L9-L33) on the AMM. We are currently missing queries for TWAP as that is not yet implemented, but we will add such bindings in a future PR when that feature is stable in the main branch.

These APIs are stable and the contracts will look the same even if the keepers and Go data types change (I just had to do one such fix right before merge), as this PR serves as a translation layer from contracts to the AMM. The Rust bindings comes with a [sample contract](https://github.com/confio/osmosis-bindings/tree/main/contracts/reflect), which we upload and use in a lot of CI tests in this PR to ensure the whole flow (contract -> Osmosis -> contract) is working correctly.

## Go adapters

I will comment key points in the PR, but we update OsmosisApp to register custom Wasm query and message handlers. All code is stored in `app/wasm`, which exposes high level interfaces that are imported by the OsmosisApp.

* `message_plugin.go` - Code to handle the `OsmosisMsg` types returned and make native calls
* `queries.go` and `query_plugin.go` - Code to handle the `OsmosisQuery` type and return proper data
*  `wasm.go` the high level binding that can be imported by `OsmosisApp` to register all the handlers
* `bindings/` - Go types to match the Rust types exported by https://github.com/confio/osmosis-bindings (same JSON representation)
* `testdata/` - Compiled wasm test contract and scripts to update
* `test/` - CI tests to ensure basic contract functionality (from initial permissioned integration) as well as all the custom callback handlers.

______

For contributor use:


app/keepers.go
```diff
@@ -367,7 +369,10 @@ func (app *OsmosisApp) InitNormalKeepers(
 
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
-	supportedFeatures := "iterator,staking,stargate"
+	supportedFeatures := "iterator,staking,stargate,osmosis"
```
ethanfrey, line 372: We add the extra feature here, so contract including [osmo-bindings](https://github.com/confio/osmosis-bindings/blob/main/packages/bindings/src/lib.rs#L13-L14) are supported here.

app/keepers.go
```diff
@@ -367,7 +369,10 @@ func (app *OsmosisApp) InitNormalKeepers(
...
+
+	wasmOpts = append(owasm.RegisterCustomPlugins(app.GAMMKeeper, app.BankKeeper), wasmOpts...)
```
ethanfrey, line 374: And here we register the custom plugins to handle those messages

app/wasm/message_plugin.go
```diff
@@ -0,0 +1,201 @@
...
+
+type CustomMessenger struct {
+	wrapped    wasmkeeper.Messenger
+	bank       *bankkeeper.BaseKeeper
+	gammKeeper *gammkeeper.Keeper
+}
+
+var _ wasmkeeper.Messenger = (*CustomMessenger)(nil)
+
+func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) ([]sdk.Event, [][]byte, error) {
```
ethanfrey, line 37: This is the entry point to handle messages. 
If you add another variant to `OsmosisMsg`, you can add the handler here.

app/wasm/message_plugin.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			return nil, sdkerrors.Wrap(err, "gamm perform swap exact amount in")
+		}
+		return &wasmbindings.SwapAmount{Out: &estimatedAmount}, nil
+	} else if swap.Amount.ExactOut != nil {
+		routes := []gammtypes.SwapAmountOutRoute{{
+			PoolId:       swap.First.PoolId,
+			TokenInDenom: swap.First.DenomIn,
+		}}
+		output := swap.First.DenomOut
```
ethanfrey, line 134: Note that we simplify the Rust API such that it always flows from first to last (even if we set ExactOut), and then convert to the different Go API here. This shows how we can easily adapt to new APIs as long as the same basic functionality is exposed.

app/wasm/testdata/version.txt
```diff
@@ -0,0 +1 @@
+v0.4.0
```
ValarDragon, line 1: Is this the contract version? A bit confused what this is haha

app/wasm/message_plugin.go
```diff
@@ -0,0 +1,201 @@
...
+			return nil, sdkerrors.Wrap(err, "gamm perform swap exact amount out")
+		}
+		return &wasmbindings.SwapAmount{In: &estimatedAmount}, nil
+	} else {
+		return nil, wasmvmtypes.UnsupportedRequest{Kind: "must support either Swap.ExactIn or Swap.ExactOut"}
+	}
+}
+
+// GetFullDenom is a function, not method, so the message_plugin can use it
+func GetFullDenom(contract string, subDenom string) (string, error) {
```
ethanfrey, line 161: This is where we build the custom denom if the contract mints a coin.
I assume this is enough, but if you need metadata registration or other such, please let us know.

app/wasm/query_plugin.go
```diff
@@ -0,0 +1,105 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+
+	bindings "github.com/osmosis-labs/osmosis/v7/app/wasm/bindings"
+)
+
+func CustomQuerier(osmoKeeper *QueryPlugin) func(ctx sdk.Context, request json.RawMessage) ([]byte, error) {
```
ethanfrey, line 14: This is the entry point to handle queries. 
If you add another variant to `OsmosisQuery`, you can add the dispatch code here. (and the actual logic to `QueryPlugin`). 
We need this structure as the API takes a function for querier, not an interface.

app/wasm/test/custom_query_test.go
```diff
@@ -0,0 +1,357 @@
...
+		sdk.NewInt64Coin("uosmo", 12000000),
+		sdk.NewInt64Coin("ustar", 240000000),
+	}
+	// 20 star to 1 osmo
+	starPool := preparePool(t, ctx, osmosis, actor, poolFunds)
+
+	reflect := instantiateReflectContract(t, ctx, osmosis, actor)
+	require.NotEmpty(t, reflect)
+
+	// query spot price
```
ethanfrey, line 128: Example querying spot price

app/wasm/test/queries_test.go
```diff
@@ -0,0 +1,489 @@
...
+	wasmbindings "github.com/osmosis-labs/osmosis/v7/app/wasm/bindings"
+	"github.com/stretchr/testify/assert"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/osmosis-labs/osmosis/v7/app/wasm"
+)
+
+func TestFullDenom(t *testing.T) {
```
ethanfrey, line 15: This file includes many table tests on the go functions (not touching the contract) to cover many edge cases faster. We ensure the basic happy/sad paths are covered with the contract to ensure the data is parsed properly, but this allows quicker table testing.

app/wasm/test/custom_msg_test.go
```diff
@@ -0,0 +1,469 @@
...
+	require.Equal(t, resp.Denom, coin.Denom)
+}
+
+type BaseState struct {
+	StarPool  uint64
+	AtomPool  uint64
+	RegenPool uint64
+}
+
+func TestSwapMsg(t *testing.T) {
```
ethanfrey, line 64: Probably the most interesting test for me.
Full stack table test where a caller requests the contract to perform a swap (while sending tokens to it), and then checking the result of the swap (balance remaining in the contract).

app/wasm/test/custom_msg_test.go
```diff
@@ -0,0 +1,469 @@
...
+							MinOutput: sdk.NewInt(100000),
+						},
+					},
+				}
+			},
+			initFunds: sdk.NewInt64Coin("uosmo", 13000000),
+			expectErr: true,
+		},
+
+		// FIXME: this panics in GAMM module !?! hits a known TODO
```
ethanfrey, line 160: I hit this two times with ExactOut. It seems that if you trade from A -> B where 1 A is worth > 2 B then the ExactOut swaps will all fail.

When I reverse it such that the output is the more valuable currency, then ExactOut passes.

app/wasm/test/custom_msg_test.go
```diff
@@ -0,0 +1,469 @@
...
+		//		sdk.NewInt64Coin("ustar", 120000000),
+		//	},
+		//},
+
+		"exact out: simple swap works": {
+			msg: func(state BaseState) *wasmbindings.SwapMsg {
+				return &wasmbindings.SwapMsg{
+					First: wasmbindings.Swap{
+						PoolId:   state.AtomPool,
+						DenomIn:  "uosmo",
```
ethanfrey, line 192: Like here, OSMO -> ATOM works with ExactOut.

The above is a bug in the GAMM code (with existing TODO). I think it would be nice to fix, but that is out of scope of this PR

app/wasm/test/custom_msg_test.go
```diff
@@ -0,0 +1,469 @@
...
+				}
+			},
+			initFunds: sdk.NewInt64Coin("ustar", 240000000),
+			finalFunds: []sdk.Coin{
+				// 240 STAR -> 6 OSMO
+				// 6 OSMO -> 2 ATOM (with minor rounding)
+				sdk.NewInt64Coin("uatom", 1999999),
+			},
+		},
+		"exact out: 2 step multi-hop": {
```
ethanfrey, line 240: Even when setting ExactOut (where we mangle the args to call into x/gamm)

app/wasm/bindings/pool.go
```diff
@@ -0,0 +1,8 @@
+package wasmbindings
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+type PoolAssets struct {
+	Assets []sdk.Coin
+	Shares sdk.Coin
+}
```
ValarDragon, line 8: As part of the refactor, we're trying to convert a lot of instances of `PoolAssets` in the code / other modules, to be PoolLiquidity. We should probably revisit this wasm binding pre-release

ethanfrey (commented):
A few comments to help orient the reviewer.

app/wasm/bindings/pool.go
ethanfrey: We can gladly rename everything. Changing the type I easy and this is only used when querying the gamm, so please do suggest new names.

We use the [Rust definitions as the canonical source](https://github.com/confio/osmosis-bindings/blob/main/packages/bindings/src/query.rs#L68-L74) for all types that hit the contract boundary, and this is more permanent (breaking API changes), so maybe you can double-check we use the proper names there first. Types that are only used within this package can easily be renamed by anyone

app/wasm/testdata/version.txt
ethanfrey: It is the version of the contract we use, so you can match it to which bindings: https://github.com/confio/osmosis-bindings/tree/v0.4.0/packages/bindings

We changed these names a few times, and use this to make sure the Go and Rust stay in sync. Once this is ready to go in a tagged release, we will make a 1.0 of osmo-bindings and this will be stable. But for now, it lets us track renames more easily

ethanfrey:
This is only failing linter, but not sure how to run that locally (and there is no clear error message in the CI).

Any tips?

ValarDragon:
So our original thinking re: native token minting was to make a "tokenfactory" module #584 / #1029 . Would it be possible / a blocker to comment out or otherwise disable the Mint tokens API for now, and re-enable that once the token factory gets on mainnet?

The goal right now is to get a mainnet upgrade out around April 8th / applied on chain week of April 15th. So we likely can't get token factory in on that time frame. Should we aim to get osmo bindings in?

Related: how careful do we need to be around backwards compatibility? (Are the bindings versioned)

ValarDragon:
Re linting: `make lint` lets you run it locally. You may need to merge main into the branch, since the linter had some updates in the last week.

The error here appears to be trailing newlines, so I think `make format` may even fix that for you

app/wasm/bindings/pool.go
ValarDragon: Awesome! Will review those in more depth then

app/wasm/message_plugin.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddress, contractIBCPortID string, msg wasmvmtypes.CosmosMsg) ([]sdk.Event, [][]byte, error) {
+	if msg.Custom != nil {
+		// only handle the happy path where this is really minting / swapping ...
+		// leave everything else for the wrapped version
+		var contractMsg wasmbindings.OsmosisMsg
+		if err := json.Unmarshal(msg.Custom, &contractMsg); err != nil {
+			return nil, nil, sdkerrors.Wrap(err, "osmosis msg")
+		}
+		if contractMsg.MintTokens != nil {
```
ValarDragon, line 45: Is it worth adding any error handling if multiple fields are entered (e.g. no result if MintTokens != nil and Swap != nil)

app/wasm/message_plugin.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		if swap.Amount.ExactIn.Input.IsNegative() {
+			return nil, wasmvmtypes.InvalidRequest{Err: "gamm perform swap negative amount in"}
+		}
+		tokenIn := sdk.Coin{
+			Denom:  swap.First.DenomIn,
+			Amount: swap.Amount.ExactIn.Input,
+		}
+		tokenOutMinAmount := swap.Amount.ExactIn.MinOutput
+		estimatedAmount, err := keeper.MultihopSwapExactAmountIn(ctx, contractAddr, routes, tokenIn, tokenOutMinAmount)
```
ValarDragon, line 124: I think this variable should be called `tokenOutAmount` right? Similar for the SwapExactAmountOut case?

app/wasm/bindings/query.go
```diff
@@ -0,0 +1,73 @@
...
+	PoolId uint64 `json:"id"`
+}
+
+type SpotPrice struct {
+	Swap        Swap `json:"swap"`
+	WithSwapFee bool `json:"with_swap_fee"`
+}
+
+type EstimateSwap struct {
+	Contract string     `json:"contract"`
```
ValarDragon, line 39: Whats the contract here?

app/wasm/message_plugin.go
maurolacy: Thought about that, and decided not to, as it seemed overkill. We can add it, for correctness.

app/wasm/bindings/query.go
maurolacy: This is the sender address. It is assumed the contract will set this to its own address, that's the reason of the name.

We can rename to `ContractAddr` for clarity. The idea is that that must be an address that has funds to try the swap, if not, swap estimation fails.

maurolacy:
> So our original thinking re: native token minting was to make a "tokenfactory" module #584 / #1029 . Would it be possible / a blocker to comment out or otherwise disable the Mint tokens API for now, and re-enable that once the token factory gets on mainnet?

Sure.

> 
> The goal right now is to get a mainnet upgrade out around April 8th / applied on chain week of April 15th. So we likely can't get token factory in on that time frame. Should we aim to get osmo bindings in for swaps, just without minting tokens?
> 
> Related: how careful do we need to be around backwards compatibility? (Are the bindings versioned)

The bindings are versioned. That's what the `version.txt` field in app/wasm/testdata is for. Every time the bindings are (breaking) changed, a new version of osmosis would need to be released, with the integration / support for the changes. There can be some non-breaking changes, basically depending on the json support of optional / omitted fields. But that will not be the case in general.

app/wasm/message_plugin.go
ethanfrey: It is impossible to do this in Rust if you use our bindings. Only if you make some custom types trying to break the caller.

If the worse we handle when you send invalid data is interpreting it as MintToken and ignoring the swap, I don't think this is an issue

app/wasm/bindings/query.go
ethanfrey: Let's rename to Sender for consistency

ethanfrey:
@ValarDragon 

> So our original thinking re: native token minting was to make a "tokenfactory" module https://github.com/osmosis-labs/osmosis/pull/584 / https://github.com/osmosis-labs/osmosis/issues/1029 . Would it be possible / a blocker to comment out or otherwise disable the Mint tokens API for now, and re-enable that once the token factory gets on mainnet?

This is commented out. Since the commits will be squashed and this will not be in the git history, I left this in as comments. When the token factory is added, we can add osmosis-v2 bindings that contain this in addition (and all v1 functionality)

app/wasm/bindings/query.go
```diff
@@ -0,0 +1,73 @@
...
+
+// OsmosisQuery contains osmosis custom queries.
+// See https://github.com/confio/osmosis-bindings/blob/main/packages/bindings/src/query.rs
+type OsmosisQuery struct {
+	// /// Given a subdenom minted by a contract via `OsmosisMsg::MintTokens`,
+	// /// returns the full denom as used by `BankMsg::Send`.
+	// FullDenom *FullDenom `json:"full_denom,omitempty"`
+	/// For a given pool ID, list all tokens traded on it with current liquidity (spot).
+	/// As well as the total number of LP shares and their denom.
+	PoolState *PoolState `json:"pool_state,omitempty"`
```
ValarDragon, line 15: Should we rename these types to have a Query suffix?

ValarDragon (approved):
This LGTM!

app/wasm/bindings/query.go
maurolacy: Is that the norm across the project? If not, I would leave it as it is.

ethanfrey:
Merged main and fixed conflicts.

I will merge PR once CI passes.

app/wasm/bindings/query.go
ValarDragon: That is the norm for all of the golang queries, I don't know whats the norm for the cosmwasm side

ethanfrey: We typically label them the same name as the field.

Again, as long as the json stays the same, you can gladly rename the types as you wish in another PR.
These are for use inside the osmosis Golang code, so it should be whatever idiom looks good to you.

Newman1515 (changes requested):
Ok and I am sorry I didn't put attention what I was reading but if the name need to change ok..work for me.

Issue #1121, opened 2022-03-21
Auto-swap non-osmo tx fees to osmo


AlpinYukseloglu:
Currently when a transaction fee is paid in a valid non-OSMO token, the token is collected and distributed as-is to stakers. While this doesn't affect the core user flow since gas fees are zero for the time being anyway, it causes confusing UX around claiming staking rewards (since epochs with too many additional denoms used as tx fees breach the auto set gas limit on Keplr) and dust for random denoms showing up in wallets. The goal of this change would be to implement the following logic:

- For osmo, do whatever it currently does (send to distribution module)
- For not osmo, send to a module account $FOO in txfees module
- On epoch, auto-swap the funds in module account $FOO and distribute that osmo to stakers.

More specifically, one relatively clean way of implementing this would be to take the following steps:

1. Create a second module account in the TxFees module to send non-OSMO fee tokens to
2. Add a check in the DeductFees function (sends tokens from fee payer to module account) that checks if the paid fee denom is OSMO and if not, make sure they are supported on the DEX (with enough liquidity) and send the tokens to the second module account
3. At the end of each epoch, iterate through all the denoms in the second module account, swapping each into OSMO and then transferring the total amount to the first module account to be distributed like normal

ValarDragon:
- [ ] Move DeductFeesAnteHandler code into txfeeskeeper
- [ ] Make new module account in txfees
- [ ] Send all non-native fee coins to that module account

- [ ] On epoch code, swap all non-native tx fees in that module to osmo, move that osmo to distribution for staking rewards

- [ ] Upstream the deduct fees keeper (no swap code) to mainline SDK to fix their gas problems

danwt:
Hi,

I was just wondering why the epoch part of this is important? Would it still be important even if the fees were not distributed to stakers?

Thanks a lot

czarcas7ic:
Its run at epoch in order for the fees that got transformed into osmo to the be distributed. It could happen at faster or slower intervals, but just makes sense to run at epoch prior to distribution.

danwt:
Thanks @czarcas7ic I understand it needs to be transformed into osmo, but I was just curious why the need to do this at epoch and not on the fly? thanks!

AlpinYukseloglu:
> Thanks @czarcas7ic I understand it needs to be transformed into osmo, but I was just curious why the need to do this at epoch and not on the fly? thanks!

[...]

PR #1145, opened 2022-03-25
Auto-swap non-osmo tx fees to osmo

AlpinYukseloglu:
Closes: #1121 

## Description

- Moved DeductFeeDecorator from auth -> txfees module
- Created second module account in txfees module to receive non-OSMO tx fees
- Implemented logic to route non-OSMO fees to second module account
- Set up AfterEpochEnd listener to trigger fee swaps into base denom + transfer to the main module account
- Did all the necessary wiring in app & txfees module (hook-related + misc.)
- Set up testing for all major changes/additions (primarily in `feedecorator_test.go` and `hooks_test.go`)

______

For contributor use:


x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,78 @@
...
+			// TO DO: figure out how to cast this or get the pool ID for the main OSMO paired pool
+			feetoken := coin.(txfeestypes.FeeToken)
+
+			// calculate spot price to determine minimum out for swap
+			// question: is the order of input denom and output denom correct?
+			spotPrice, _ := k.spotPriceCalculator.CalculateSpotPrice(ctx, feetoken.GetPoolID(), coin.Denom, baseDenom)
+
+			// swap into OSMO
+			// question: is spotPrice really the minimum out for the swap?
+			k.spotPriceCalculator.SwapExactAmountIn(ctx, addrFoo, feetoken.GetPoolId(), coin.Denom, baseDenom, k.spotPriceCalculator)
```
AlpinYukseloglu, line 43: @ValarDragon Is this an appropriate place to have the swap function call?

ValarDragon: Yeah!

Two notes though:
* I don't know why the last argument is k.spotPriceCalculator
* We should make a second keeper in expected_keeper called GammKeeper, that we use for the Swap methods

AlpinYukseloglu: Sounds good, thanks! The last argument is a placeholder (see the question in the comment above it). 

How would the minimum output amount supposed to be calculated in this case?

ValarDragon: For now I think its fine to allow max slippage (put sdk.Coins{} as the min out)

AlpinYukseloglu: I believe the argument is supposed to be an integer - would it just be zero in this case?

app/ante.go
```diff
@@ -31,6 +31,7 @@ func NewAnteHandler(
 ) sdk.AnteHandler {
 	mempoolFeeOptions := txfeestypes.NewMempoolFeeOptions(appOpts)
 	mempoolFeeDecorator := txfeeskeeper.NewMempoolFeeDecorator(*txFeesKeeper, mempoolFeeOptions)
+	deductFeeDecorator := txfeeskeeper.NewDeductFeeDecorator(*txFeesKeeper, ak, bankKeeper, nil)
```
alexanderbez, line 34: Just an FYI, I think I mentioned to @ValarDragon that ultimately ante-handler decorators/middleware should be defined in a root `ante` package, instead of in modules. However, we can address this in a future PR 👍

app/ante.go
```diff
@@ -43,7 +44,8 @@ func NewAnteHandler(
 		ante.TxTimeoutHeightDecorator{},
 		ante.NewValidateMemoDecorator(ak),
 		ante.NewConsumeGasForTxSizeDecorator(ak),
-		ante.NewDeductFeeDecorator(ak, bankKeeper, nil),
+		// Replaced with version from our txfees module from auth (previously "ante.NewDeductFeeDecorator(ak, bankKeeper, nil)")"
```
alexanderbez, line 47: nit: I think this comment is superfluous

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
 	}
 	return cfgMinGasPrice
 }
+
+// DeductFeeDecorator deducts fees from the first signer of the tx
+// If the first signer does not have the funds to pay for the fees, return with InsufficientFunds error
+// Call next AnteHandler if fees successfully deducted
+// CONTRACT: Tx must implement FeeTx interface to use DeductFeeDecorator
```
alexanderbez, line 130:
```suggestion
// DeductFeeDecorator deducts fees from the first signer of the tx.
// If the first signer does not have the funds to pay for the fees, we return an InsufficientFunds error.
// We call next AnteHandler if fees successfully deducted.
//
// CONTRACT: Tx must implement FeeTx interface to use DeductFeeDecorator
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+		ak:             ak,
+		bankKeeper:     bk,
+		feegrantKeeper: fk,
+		txFeesKeeper: tk,
+	}
+}
+
+func (dfd DeductFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	
+	// checks to make sure feeTx is of the right type
```
alexanderbez, line 149:
```suggestion
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	// checks to make sure the module account has been set to collect fees in OSMO
+	if addr := dfd.ak.GetModuleAddress(types.FeeCollectorName); addr == nil {
+		return ctx, fmt.Errorf("Fee collector module account (%s) has not been set", types.FeeCollectorName)
+	}
+
+	// checks to make sure a separate module account has been set to collect fees not in OSMO
+	if addrFoo := dfd.ak.GetModuleAddress(types.FooCollectorName); addrFoo == nil {
```
alexanderbez, line 161: Is `FooCollectorName` name a typo?

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+
+	// fee can be in any denom (checked for validity later)
+	fee := feeTx.GetFee()
+	feePayer := feeTx.FeePayer()
+	feeGranter := feeTx.FeeGranter()
+
+	// set the fee payer as the default address to deduct fees from
+	deductFeesFrom := feePayer
+
+	// if feegranter set deduct fee from feegranter account.
```
alexanderbez, line 173:
```suggestion
	// If a fee granter was set, deduct fee from the fee granter's account.
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	// this works with only when feegrant enabled.
```
alexanderbez, line 174:
```suggestion
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	if feeGranter != nil {
+		if dfd.feegrantKeeper == nil {
+			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "fee grants are not enabled")
```
alexanderbez, line 177:
```suggestion
			return ctx, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "fee grants is not enabled")
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+			if err != nil {
+				return ctx, sdkerrors.Wrapf(err, "%s not allowed to pay fees from %s", feeGranter, feePayer)
+			}
+		}
+
+		// if no errors, change the account that is charged for fees to the fee granter
+		deductFeesFrom = feeGranter
+	}
+
+	// pulls account from address and makes sure it exists
```
alexanderbez, line 190:
```suggestion
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+		return ctx, sdkerrors.Wrapf(sdkerrors.ErrUnknownAddress, "fee payer address: %s does not exist", deductFeesFrom)
+	}
+
+	// deducts the fees and transfer them to the module account
+	if !feeTx.GetFee().IsZero() {
+		err = DeductFees(dfd.txFeesKeeper, dfd.bankKeeper, ctx, deductFeesFromAcc, feeTx.GetFee())
+		if err != nil {
+			return ctx, err
+		}
+	} // no additional action/else statment needed if the fee is zero since zero-gas transactions are allowed
```
alexanderbez, line 202:
```suggestion
	}
```

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+
+	events := sdk.Events{sdk.NewEvent(sdk.EventTypeTx,
```
alexanderbez, line 204: No need to declare `events` variable that used once. Simply just pass the struct literal into `ctx.EventManager().EmitEvents(...)` 👍

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,122 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+		sdk.NewAttribute(sdk.AttributeKeyFee, feeTx.GetFee().String()),
+	)}
+	ctx.EventManager().EmitEvents(events)
+
+	return next(ctx, tx, simulate)
+}
+
+// DeductFees deducts fees from the given account and transfers them to the set module account
```
alexanderbez, line 212:
```suggestion
// DeductFees deducts fees from the given account and transfers them to the set module account.
```

alexanderbez (changes requested):
Nice work! Just left some initial minor feedback

app/keepers.go
```diff
@@ -158,6 +161,7 @@ type appKeepers struct {
 	MintKeeper           *mintkeeper.Keeper
 	PoolIncentivesKeeper *poolincentiveskeeper.Keeper
 	TxFeesKeeper         *txfeeskeeper.Keeper
+	FeeGrantKeeper 		 *feegrantkeeper.Keeper
```
alexanderbez, line 164: formatting is off here :)

app/keepers.go
```diff
@@ -358,13 +358,28 @@ func (app *OsmosisApp) InitNormalKeepers(
...
+		app.EpochsKeeper,
 		keys[txfeestypes.StoreKey],
 		app.GAMMKeeper,
+		app.GAMMKeeper,
+		txfeestypes.FeeCollectorName,
+		txfeestypes.FooCollectorName,
 	)
 	app.TxFeesKeeper = &txFeesKeeper
 
+	// Note: gammKeeper is expected to satisfy the SpotPriceCalculator interface parameter
```
mattverse, line 375:
```suggestion

```

mattverse, line 375:
```suggestion
```

x/txfees/keeper/keeper_test.go
```diff
@@ -49,13 +53,22 @@ func (suite *KeeperTestSuite) SetupTest(isCheckTx bool) {
 
 	suite.queryClient = queryClient
 
+	encodingConfig := osmosisapp.MakeEncodingConfig()
+	suite.clientCtx = client.Context{}.
+		WithInterfaceRegistry(encodingConfig.InterfaceRegistry).
+		WithTxConfig(encodingConfig.TxConfig).
+		WithLegacyAmino(encodingConfig.Amino).
+		WithJSONCodec(encodingConfig.Marshaler)
+
```
mattverse, line 62: is there a reason we need to set clientCtx?

x/txfees/keeper/keeper_test.go
```diff
@@ -23,7 +23,7 @@ type KeeperTestSuite struct {
 	suite.Suite
 
 	ctx sdk.Context
-	app *app.OsmosisApp
+	app *osmosisapp.OsmosisApp
```
mattverse, line 26: just curious, why made us change app to osmosisApp :p

x/txfees/types/expected_keepers.go
```diff
@@ -2,10 +2,57 @@ package types
...
+type GammKeeper interface {
+	SwapExactAmountIn(
+		ctx sdk.Context,
+		sender sdk.AccAddress,
+		poolId uint64,
+		tokenIn sdk.Coin,
+		tokenOutDenom string,
+		tokenOutMinAmount sdk.Int,
+	) (tokenOutAmount sdk.Int, spotPriceAfter sdk.Dec, err error)
+}
```
mattverse, line 25: Can we change so that GammKeeper has both SwapExactAmountIn, CalculateSpotPricce, instead of having `SpotPriceCalculator`?

x/txfees/types/expected_keepers.go
```diff
@@ -2,10 +2,57 @@ package types
...
+
+// AccountKeeper defines the contract needed for AccountKeeper related APIs.
+// Interface provides support to use non-sdk AccountKeeper for AnteHandler's decorators.
+type AccountKeeper interface {
+	GetParams(ctx sdk.Context) (params authtypes.Params)
```
mattverse, line 30: This doesn't seem like it's being used

x/txfees/types/expected_keepers.go
```diff
@@ -2,10 +2,57 @@ package types
...
+
+// TxFeesKeeper defines the expected transaction fee keeper
+type TxFeesKeeper interface {
+	ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin, error)
+	GetBaseDenom(ctx sdk.Context) (denom string, err error)
+	GetFeeToken(ctx sdk.Context, denom string) (FeeToken, error)
+
+	// add keepers here
+}
+
```
mattverse, line 54: There wouldn't need to be a interface of itself in expected_keepers afaik, unless we have a specific reason?

x/txfees/keeper/feedecorator.go
mattverse: No i think it's actually fooCollectorName

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,117 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	}
+
+	// deducts the fees and transfer them to the module account
+	if !feeTx.GetFee().IsZero() {
+		err = DeductFees(dfd.txFeesKeeper, dfd.bankKeeper, ctx, deductFeesFromAcc, feeTx.GetFee())
+		if err != nil {
+			return ctx, err
+		}
+	}
+
```
mattverse, line 201: does `UseGrantedFees` not deposit fees from the feePayer? I'm curious if it does not use fees twice if `!feeGranter.Equals(feePayer)`

mattverse (changes requested):
Left some comments, just my personal opinion: I think we should use a different name other than `FooCollector`.

I also have an architectural question: do we have a reason we are creating and managing two different module accounts for baseDenom and arbitrary denom assets?

x/gamm/keeper/math.go
```diff
@@ -98,6 +98,22 @@ func calcOutGivenIn(
 	return tokenAmountOut
 }
 
+// exported version
```
alexanderbez, line 101: nit: let's add proper, albeit it can be concise, godoc on what this does :)

```suggestion
// CalcOutGivenIn ...
```

x/gamm/keeper/math.go
```diff
@@ -98,6 +98,22 @@ func calcOutGivenIn(
...
+	tokenWeightIn,
+	tokenBalanceOut,
+	tokenWeightOut,
+	tokenAmountIn,
+	swapFee sdk.Dec,
+) sdk.Dec {
+	// deduct swapfee on the in asset
+	tokenAmountInAfterFee := tokenAmountIn.Mul(sdk.OneDec().Sub(swapFee))
+	// delta balanceOut is positive(tokens inside the pool decreases)
+	tokenAmountOut := solveConstantFunctionInvariant(tokenBalanceIn, tokenBalanceIn.Add(tokenAmountInAfterFee), tokenWeightIn, tokenBalanceOut, tokenWeightOut)
```
alexanderbez, line 113:
```suggestion
	tokenAmountOut := solveConstantFunctionInvariant(
	  tokenBalanceIn,
	  tokenBalanceIn.Add(tokenAmountInAfterFee),
	  tokenWeightIn,
	  tokenBalanceOut,
	  tokenWeightOut,
	)
```

x/txfees/keeper/feedecorator.go
alexanderbez: Any clarification on this @AlpinYukseloglu? I don't think we should have a variable and module account named `Foo` haha. It's a universal dummy variable name.

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,117 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	return DeductFeeDecorator{
+		ak:             ak,
+		bankKeeper:     bk,
+		feegrantKeeper: fk,
+		txFeesKeeper: tk,
+	}
+}
+
+func (dfd DeductFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+
```
alexanderbez, line 149:
```suggestion
```

x/txfees/types/expected_keepers.go
ValarDragon: I think we should have SpotPriceCalculator be separate, since thats what other non-osmosis repositories would use. (Different handlers depend on different expected_keepers)

x/txfees/keeper/feedecorator.go
```diff
@@ -119,3 +123,116 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	ctx.EventManager().EmitEvents(sdk.Events{sdk.NewEvent(sdk.EventTypeTx,
+		sdk.NewAttribute(sdk.AttributeKeyFee, feeTx.GetFee().String()),
+	)})
+
+	return next(ctx, tx, simulate)
+}
+
+// DeductFees deducts fees from the given account and transfers them to the set module account.
+func DeductFees(txFeesKeeper types.TxFeesKeeper, bankKeeper types.BankKeeper, ctx sdk.Context, acc authtypes.AccountI, fees sdk.Coins) error {
+
```
alexanderbez, line 210:
```suggestion
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,61 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	epochstypes "github.com/osmosis-labs/osmosis/v7/x/epochs/types"
+	txfeestypes "github.com/osmosis-labs/osmosis/v7/x/txfees/types"
+)
+
+func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
+}
```
alexanderbez, line 10:
```suggestion
func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) { }
```

x/txfees/keeper/keeper.go
```diff
@@ -17,19 +19,39 @@ type (
...
 
+		accountKeeper    types.AccountKeeper
+		bankKeeper       *bankkeeper.BaseKeeper
+		epochKeeper      types.EpochKeeper
+		gammKeeper		 types.GammKeeper
+
 		spotPriceCalculator types.SpotPriceCalculator
+
+		feeCollectorName string
+		altFeeCollectorName string
```
alexanderbez, line 30:
```suggestion

		accountKeeper    types.AccountKeeper
		bankKeeper       *bankkeeper.BaseKeeper
		epochKeeper      types.EpochKeeper
		gammKeeper		 types.GammKeeper
		spotPriceCalculator types.SpotPriceCalculator
		feeCollectorName string
		altFeeCollectorName string
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,61 @@
...
+
+var _ epochstypes.EpochHooks = Hooks{}
+
+// Return the wrapper struct
+func (k Keeper) Hooks() Hooks {
+	return Hooks{k}
+}
+
+// epochs hooks
+// Don't do anything pre epoch start
```
alexanderbez, line 55:
```suggestion
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,61 @@
...
+func (h Hooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
+}
```
alexanderbez, line 57:
```suggestion
func (h Hooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) {}
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,61 @@
...
+
+// at the end of each epoch, swap all non-OSMO fees into OSMO and transfer to fee module account
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
+
+	addrAltFee := k.accountKeeper.GetModuleAddress(txfeestypes.AltFeeCollectorName)
+
+	altFeeAccountBalances := k.bankKeeper.GetAllBalances(ctx, addrAltFee)
+
+	baseDenom, _ := k.GetBaseDenom(ctx)
```
alexanderbez, line 19:
```suggestion
	addrAltFee := k.accountKeeper.GetModuleAddress(txfeestypes.AltFeeCollectorName)
	altFeeAccountBalances := k.bankKeeper.GetAllBalances(ctx, addrAltFee)
	baseDenom, _ := k.GetBaseDenom(ctx)
```

x/gamm/keeper/math.go
```diff
@@ -98,6 +98,30 @@ func calcOutGivenIn(
 	return tokenAmountOut
 }
 
+// CalcOutGivenIn calculates token to be provided, fee added,
+// given the swapped out amount, using solveConstantFunctionInvariant.
+// This version is exported and can be used to calculate the exact output of a swap.
+func CalcOutGivenIn(
```
ValarDragon, line 104: Is this being used somewhere? This should probably not be exposed for any txfee purpose. (Its also not compatible post-API refactor)

x/txfees/keeper/feedecorator_test.go
```diff
@@ -157,23 +174,60 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-		_, err := antehandler(suite.ctx, tx, false)
+		dfd := keeper.NewDeductFeeDecorator(*suite.app.TxFeesKeeper, *suite.app.AccountKeeper, *suite.app.BankKeeper, *suite.app.FeeGrantKeeper)
+		antehandlerMFD := sdk.ChainAnteDecorators(mfd, dfd)
+		_, err := antehandlerMFD(suite.ctx, tx, false)
+
 		if tc.expectPass {
+			if tc.baseDenomGas && !tc.txFee.IsZero() {
+				moduleAddr := suite.app.AccountKeeper.GetModuleAddress(types.FeeCollectorName)
+				suite.Require().Equal(tc.txFee[0], suite.app.BankKeeper.GetBalance(suite.ctx, moduleAddr, baseDenom), tc.name)
+				suite.app.BankKeeper.SendCoinsFromModuleToAccount(suite.ctx, types.FeeCollectorName, addr0, tc.txFee)
```
ValarDragon, line 225: why are we sending funds out of the module to an account in the tests?

x/txfees/keeper/feedecorator_test.go
```diff
@@ -157,23 +174,60 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
+
+		simapp.FundAccount(suite.app.BankKeeper, suite.ctx, addr0, tc.txFee)
+		
+		txBuilder.SetMsgs(msgs[0])
+		txBuilder.SetSignatures(sigV2)
+		txBuilder.SetMemo("")
+		txBuilder.SetFeeAmount(tc.txFee)
+		txBuilder.SetGasLimit(gasLimit)
+
+		tx := txBuilder.GetTx()
```
ValarDragon, line 214: Thoughts on refactoring all of this to a helper method?

x/txfees/types/expected_keepers.go
ValarDragon: tbh, I do think its a good idea for giving other modules an easier import

x/txfees/types/keys.go
```diff
@@ -10,11 +14,28 @@ const (
 	// RouterKey is the message route for slashing
 	RouterKey = ModuleName
 
+	// FeeCollectorName the module account name for the fee collector account address.
+	FeeCollectorName = "fee_collector"
+
+	// AltFeeCollectorName the module account name for the alt fee collector account address (used for auto-swapping non-OSMO tx fees).
+	AltFeeCollectorName = "alt_fee_collector"
```
ValarDragon, line 21: Should we rename this to "NonNativeFeeCollector" ?

x/txfees/types/expected_keepers.go
mattverse: @AlpinYukseloglu

x/txfees/types/keys.go
mattverse: Agree with the renaming!

ValarDragon (changes requested):

x/txfees/keeper/hooks.go
```diff
@@ -21,15 +21,20 @@ func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumb
 		
 			feetoken, err := k.GetFeeToken(ctx, coin.Denom)
 			if err == nil {
```
ValarDragon, line 23: Can we switch this to `if err != nil { continue }` syntax?

This is definitely correct, but would prefer to keep more in line with the golang standard syntax around errors. Sorry I'm being a human linter :sweat_smile:

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,50 @@
...
+	addrNonNativeFee := k.accountKeeper.GetModuleAddress(txfeestypes.NonNativeFeeCollectorName)
+	nonNativeFeeAccountBalances := k.bankKeeper.GetAllBalances(ctx, addrNonNativeFee)
+	baseDenom, _ := k.GetBaseDenom(ctx)
+
+	for _, coin := range nonNativeFeeAccountBalances {
+		if coin.Denom == baseDenom {
+			continue
+		} else {
+		
+			feetoken, _ := k.GetFeeToken(ctx, coin.Denom)
```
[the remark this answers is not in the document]
ValarDragon: Yup it fixes it!

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// PLACEHOLDER – addrFee := k.accountKeeper.GetModuleAddress(txfeestypes.FeeCollectorName)
+
+	// Potential error to test for: do swaps to OSMO consolidate into a single denom in fooAccountBalances? Should be yes but keep an eye out
+
+	// send all OSMO to fee module account to be distributed in the next block
+	k.bankKeeper.SendCoinsFromModuleToModule(ctx, txfeestypes.FooCollectorName, txfeestypes.FeeCollectorName, fooAccountBalances)
+
+	// Should events be emitted at the end here?
```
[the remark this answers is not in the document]
ValarDragon: Hrmm, thats a good idea. I'm down to add an event for the total amount of osmo swapped, though feel free to do that in another PR

x/txfees/keeper/feedecorator_test.go
mattverse: https://github.com/osmosis-labs/osmosis/blob/main/app/apptesting/test_suite.go would be the perfect place, where we keep methods of useful helper methods

x/gamm/keeper/math.go
ValarDragon: Per call: Lets keep it, just switch to doing it off of the pool interface

x/txfees/keeper/feedecorator_test.go
ValarDragon: Ah that makes, yeah there is a better way!

At the beginning of every test cases processing code, do:
```
suite.SetupTest(false)
suite.ExecuteUpgradeFeeTokenProposal(uion, uionPoolId)
```

And then every test will start on a brand new, clean state

x/txfees/types/expected_keepers.go
ValarDragon: yeah!

x/txfees/keeper/feedecorator_test.go
AlpinYukseloglu: Awesome, adding in next commit!

x/gamm/keeper/math.go
AlpinYukseloglu: Forgot to resolve this – resolved in a commit a couple days ago!

x/txfees/keeper/feedecorator_test.go
AlpinYukseloglu: Moved the transaction building component into a helper method! It might be worth eventually creating a "basic transaction" helper method that fills in blank parameters for everything and builds a test tx, but in this case the fee and gas limit changes in each test case so the only part that could be cleanly abstracted was the building of the tx.

x/txfees/types/keys.go
```diff
@@ -10,11 +14,28 @@ const (
...
+	AddressStoreKeyPrefix = []byte{0x01}
+
+	// param key for global account number
+	GlobalAccountNumberKey = []byte("globalAccountNumber")
 )
+
+// AddressStoreKey turn an address to key used to get it from the account store
+func AddressStoreKey(addr sdk.AccAddress) []byte {
+	return append(AddressStoreKeyPrefix, addr.Bytes()...)
+}
```
alexanderbez, line 41: None of this is actually used from what I can tell. Why was this added?

x/txfees/types/expected_keepers.go
```diff
@@ -2,10 +2,58 @@ package types
...
+	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
+}
+
+// TxFeesKeeper defines the expected transaction fee keeper
+type TxFeesKeeper interface {
+	ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin, error)
+	GetBaseDenom(ctx sdk.Context) (denom string, err error)
+	GetFeeToken(ctx sdk.Context, denom string) (FeeToken, error)
+
+	// add keepers here
```
alexanderbez, line 53:
```suggestion
```

x/txfees/keeper/keeper.go
```diff
@@ -9,27 +9,45 @@ import (
...
+	gammKeeper                types.GammKeeper
+	spotPriceCalculator       types.SpotPriceCalculator
+	feeCollectorName          string
+	nonNativeFeeCollectorName string
+}
 
 func NewKeeper(
 	cdc codec.Codec,
+	accountKeeper types.AccountKeeper,
+	bankKeeper *bankkeeper.BaseKeeper,
```
alexanderbez, line 33: why are we relying on direct import of x/bank keeper? We should utilize the interface we defined in `expected_keepers.go`.

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+			minTokenOut := sdk.ZeroInt()
+			_, err = k.gammKeeper.SwapExactAmountIn(ctx, addrNonNativeFee, feetoken.PoolID, coin, baseDenom, minTokenOut)
+			if err != nil {
+				panic(err)
+			}
+		}
+	}
+
+	// Get all of the txfee payout denom in the module account
+	nonNativeFeeAccountBaseDenomBalance := sdk.NewCoins(k.bankKeeper.GetBalance(ctx, addrNonNativeFee, baseDenom))
```
alexanderbez, line 40:
```suggestion
	nonNativeCoins := sdk.NewCoins(k.bankKeeper.GetBalance(ctx, addrNonNativeFee, baseDenom))
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+
+	epochstypes "github.com/osmosis-labs/osmosis/v7/x/epochs/types"
+	txfeestypes "github.com/osmosis-labs/osmosis/v7/x/txfees/types"
+)
+
+func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) {}
+
+// at the end of each epoch, swap all non-OSMO fees into OSMO and transfer to fee module account
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
+	addrNonNativeFee := k.accountKeeper.GetModuleAddress(txfeestypes.NonNativeFeeCollectorName)
```
alexanderbez, line 14:
```suggestion
	nonNativeFeeAddr := k.accountKeeper.GetModuleAddress(txfeestypes.NonNativeFeeCollectorName)
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+	nonNativeFeeAccountBalances := k.bankKeeper.GetAllBalances(ctx, addrNonNativeFee)
```
alexanderbez, line 15:
```suggestion
	nonNativeBalances := k.bankKeeper.GetAllBalances(ctx, addrNonNativeFee)
```

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+	baseDenom, _ := k.GetBaseDenom(ctx)
+
+	for _, coin := range nonNativeFeeAccountBalances {
+		if coin.Denom == baseDenom {
+			continue
+		} else {
```
alexanderbez, line 21: there's no need to wrap this in an `else` block as the `if` condition, if matched, will just continue. I'm surprised the linter didn't signal this.

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,64 @@
...
+			if err != nil {
+				panic(err)
+			}
+
+			// We allow full slippage. Theres not really an effective way to bound slippage until TWAP's land,
+			// but even then the point is a bit moot.
+			// The only thing that could be done is a costly griefing attack to reduce the amount of osmo given as tx fees.
+			// However the idea of the txfees FeeToken gating is that the pool is sufficiently liquid for that base token.
+			minTokenOut := sdk.ZeroInt()
+			_, err = k.gammKeeper.SwapExactAmountIn(ctx, addrNonNativeFee, feetoken.PoolID, coin, baseDenom, minTokenOut)
```
alexanderbez, line 32: There's no need to set a `minTokenOut` variable as it's only used once (below).
```suggestion
			_, err = k.gammKeeper.SwapExactAmountIn(ctx, addrNonNativeFee, feetoken.PoolID, coin, baseDenom, sdk.ZeroInt())
```

alexanderbez (changes requested):

x/txfees/keeper/hooks.go
AlpinYukseloglu: Great point - should be fixed in next commit.

x/txfees/keeper/keeper.go
AlpinYukseloglu: Great catch - adding in next commit

x/txfees/types/keys.go
AlpinYukseloglu: Hmm I believe this is vestigial from when I was first migrating the relevant ante handler into our txfees module - removing in next commit

faddat:
this is really cool!

thank you!

app/keepers.go
```diff
@@ -103,6 +104,7 @@ type appKeepers struct {
 	MintKeeper           *mintkeeper.Keeper
 	PoolIncentivesKeeper *poolincentiveskeeper.Keeper
 	TxFeesKeeper         *txfeeskeeper.Keeper
+	FeeGrantKeeper       *feegrantkeeper.Keeper
```
ValarDragon, line 107: This PR is adding the FeeGrant module to Osmosis. I'm no longer opposed to adding it (I was last release, due to time since last security bug in the module, but a bit confusing why its bundled in this PR vs discussed elsewhere

ValarDragon: I would prefer we separate it out into a second issue to discuss its inclusion / if its what we want.

x/gamm/keeper/math.go
```diff
@@ -0,0 +1,202 @@
+package keeper
```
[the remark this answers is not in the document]
AlpinYukseloglu: Right - good point

app/keepers.go
ValarDragon: Lets handle this decision in a second PR

ValarDragon:
hrmm something got messed up in the imports for app/ante in the merge

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) {}
+
+// at the end of each epoch, swap all non-OSMO fees into OSMO and transfer to fee module account
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
+	nonNativeFeeAddr := k.accountKeeper.GetModuleAddress(txfeestypes.NonNativeFeeCollectorName)
+	nonNativeBalances := k.bankKeeper.GetAllBalances(ctx, nonNativeFeeAddr)
+	baseDenom, _ := k.GetBaseDenom(ctx)
+
+	for _, coin := range nonNativeBalances {
```
ValarDragon, line 18: We should change the GetAllBalances and this for loop, to instead be iterating over every fee token denom.

This is because `GetAllBalances` can actually take unbounded time, and is not suitable for usage in the state machine

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		// We allow full slippage. Theres not really an effective way to bound slippage until TWAP's land,
+		// but even then the point is a bit moot.
+		// The only thing that could be done is a costly griefing attack to reduce the amount of osmo given as tx fees.
+		// However the idea of the txfees FeeToken gating is that the pool is sufficiently liquid for that base token.
+		_, err = k.gammKeeper.SwapExactAmountIn(ctx, nonNativeFeeAddr, feetoken.PoolID, coin, baseDenom, sdk.ZeroInt())
+		if err != nil {
+			panic(err)
+		}
```
ValarDragon, line 34: We should wrap this in `osmoutils.ApplyFuncIfNoErr`, and only log the error. We can't panic for an error here, as that would cause a chain halt

x/txfees/keeper/feedecorator.go
```diff
@@ -121,3 +125,114 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	}
+
+	// checks to make sure the module account has been set to collect fees in OSMO
+	if addr := dfd.ak.GetModuleAddress(types.FeeCollectorName); addr == nil {
+		return ctx, fmt.Errorf("Fee collector module account (%s) has not been set", types.FeeCollectorName)
+	}
+
+	// checks to make sure a separate module account has been set to collect fees not in OSMO
+	if addrNonNativeFee := dfd.ak.GetModuleAddress(types.NonNativeFeeCollectorName); addrNonNativeFee == nil {
+		return ctx, fmt.Errorf("Alt Fee collector module account (%s) has not been set", types.NonNativeFeeCollectorName)
```
alexanderbez, line 163:
```suggestion
		return ctx, fmt.Errorf("non native fee collector module account (%s) has not been set", types.NonNativeFeeCollectorName)
```

x/txfees/keeper/feedecorator.go
```diff
@@ -121,3 +125,114 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	}
+
+	// pulls base denom from TxFeesKeeper (should be uOSMO)
+	baseDenom, err := txFeesKeeper.GetBaseDenom(ctx)
+	if err != nil {
+		return err
+	}
+
+	// checks if input fee is uOSMO (assumes only one fee token exists in the fees array (as per the check in mempoolFeeDecorator))
+	if fees[0].Denom == baseDenom {
```
alexanderbez, line 223: Is this invariant always true @ValarDragon ?

x/txfees/keeper/feedecorator_test.go
```diff
@@ -158,23 +176,61 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
+		txBuilder := suite.clientCtx.TxConfig.NewTxBuilder()
+		priv0, _, addr0 := testdata.KeyTestPubAddr()
+		acc1 := suite.App.AccountKeeper.NewAccountWithAddress(suite.Ctx, addr0)
+		suite.App.AccountKeeper.SetAccount(suite.Ctx, acc1)
+		msgs := []sdk.Msg{testdata.NewTestMsg(addr0)}
+		privs, accNums, accSeqs := []cryptotypes.PrivKey{priv0}, []uint64{0}, []uint64{0}
+		signerData := authsigning.SignerData{
+			ChainID:       suite.Ctx.ChainID(),
+			AccountNumber: accNums[0],
+			Sequence:      accSeqs[0]}
```
alexanderbez, line 206:
```suggestion
			Sequence:      accSeqs[0],
		}
```

x/txfees/keeper/feedecorator_test.go
```diff
@@ -158,23 +176,61 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
+
+		gasLimit := tc.gasRequested
+		sigV2, _ := clienttx.SignWithPrivKey(
+			1,
+			signerData,
+			txBuilder,
+			privs[0],
+			suite.clientCtx.TxConfig,
+			accSeqs[0])
```
alexanderbez, line 215:
```suggestion
			accSeqs[0],
		)
```

x/txfees/keeper/hooks.go
alexanderbez: Yes, this is a good point!

app/keepers.go
```diff
@@ -338,10 +338,17 @@ func (app *OsmosisApp) InitNormalKeepers(
...
 
+	// Note: gammKeeper is expected to satisfy the SpotPriceCalculator interface parameter
 	txFeesKeeper := txfeeskeeper.NewKeeper(
 		appCodec,
+		app.AccountKeeper,
+		app.BankKeeper,
+		app.EpochsKeeper,
 		keys[txfeestypes.StoreKey],
 		app.GAMMKeeper,
+		app.GAMMKeeper,
```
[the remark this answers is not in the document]
ValarDragon: one is as the spot price keeper, and a second is as a keeper we swap against

x/txfees/keeper/feedecorator.go
ValarDragon: The invariant is true! (For a given denom, there is at most one fee token)

I am confused as to how the invariant relates to these lines though

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,65 @@
...
+		if feetoken.Denom == baseDenom {
+			continue
+		}
+		coinBalance := k.bankKeeper.GetBalance(ctx, nonNativeFeeAddr, feetoken.Denom)
+		if coinBalance.Amount.IsZero() {
+			continue
+		}
+
+		// Do the swap of this fee token denom to base denom.
+		_ = osmoutils.ApplyFuncIfNoError(ctx, func(cacheCtx sdk.Context) error {
```
alexanderbez, line 29: should we log the error at least?

x/txfees/keeper/hooks.go
```diff
@@ -0,0 +1,65 @@
...
+			return err
+		})
+	}
+
+	// Get all of the txfee payout denom in the module account
+	baseDenomCoins := sdk.NewCoins(k.bankKeeper.GetBalance(ctx, nonNativeFeeAddr, baseDenom))
+
+	_ = osmoutils.ApplyFuncIfNoError(ctx, func(cacheCtx sdk.Context) error {
+		err := k.bankKeeper.SendCoinsFromModuleToModule(ctx, txfeestypes.NonNativeFeeCollectorName, txfeestypes.FeeCollectorName, baseDenomCoins)
+		return err
```
alexanderbez, line 45: ditto: should we log the error?

alexanderbez (approved):

x/txfees/keeper/hooks.go
ValarDragon: The error gets logged to ctx in ApplyFuncIfNoError https://github.com/osmosis-labs/osmosis/blob/main/osmoutils/cache_ctx.go#L26

ValarDragon (approved):

alexanderbez:
@mattverse wanna give this one more look?

x/txfees/keeper/hooks_test.go
```diff
@@ -0,0 +1,84 @@
...
+	suite.Require().True(suite.App.BankKeeper.HasBalance(suite.Ctx, moduleAddrNonNativeFee, coins[1]))
+	suite.Require().True(suite.App.BankKeeper.HasBalance(suite.Ctx, moduleAddrNonNativeFee, coins[2]))
+
+	params := suite.App.IncentivesKeeper.GetParams(suite.Ctx)
+	futureCtx := suite.Ctx.WithBlockTime(time.Now().Add(time.Minute))
+
+	suite.App.EpochsKeeper.AfterEpochEnd(futureCtx, params.DistrEpochIdentifier, int64(1))
+
+	suite.Require().Empty(suite.App.BankKeeper.GetAllBalances(suite.Ctx, moduleAddrNonNativeFee))
+	suite.Require().True(suite.App.BankKeeper.GetBalance(suite.Ctx, moduleAddrFee, baseDenom).Amount.GTE(fullExpectedOutput.Amount.TruncateInt()))
```
mattverse, line 83: Can we change this to testing w/ precise number instead of using `GTE`?

x/txfees/keeper/feedecorator.go
```diff
@@ -121,3 +125,114 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	}
+}
+
+func (dfd DeductFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	// checks to make sure the module account has been set to collect fees in OSMO
```
mattverse, line 156:
```suggestion
	// checks to make sure the module account has been set to collect fees in base token
```

x/txfees/keeper/feedecorator.go
```diff
@@ -121,3 +125,114 @@ func (mfd MempoolFeeDecorator) GetMinBaseGasPriceForTx(ctx sdk.Context, baseDeno
...
+	if addr := dfd.ak.GetModuleAddress(types.FeeCollectorName); addr == nil {
+		return ctx, fmt.Errorf("Fee collector module account (%s) has not been set", types.FeeCollectorName)
+	}
+
+	// checks to make sure a separate module account has been set to collect fees not in OSMO
```
mattverse, line 161:
```suggestion
	// checks to make sure a separate module account has been set to collect fees not in base token
```

x/txfees/keeper/feedecorator_test.go
```diff
@@ -158,23 +176,63 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
 
+		// reset pool and accounts for each test
+		suite.SetupTest(false)
+		uionPoolId := suite.PrepareUni2PoolWithAssets(
+			sdk.NewInt64Coin(sdk.DefaultBondDenom, 500),
+			sdk.NewInt64Coin(uion, 500),
+		)
+		suite.ExecuteUpgradeFeeTokenProposal(uion, uionPoolId)
+
 		suite.Ctx = suite.Ctx.WithIsCheckTx(tc.isCheckTx)
```
mattverse, line 193:
```suggestion
		suite.Ctx = suite.Ctx.WithIsCheckTx(tc.isCheckTx).WithMinGasPrices(tc.minGasPrices)
```

mattverse (commented):
Left some minor comments! 

Also, would it be possible to add documnetation in spec before we merge this?

x/txfees/keeper/hooks_test.go
```diff
@@ -0,0 +1,84 @@
...
+	expectedOutput2, err := atomPool.CalcOutAmtGivenIn(suite.Ctx,
+		sdk.Coins{sdk.Coin{Denom: atom, Amount: coins.AmountOf(atom)}},
+		baseDenom,
+		swapFee)
+	suite.Require().NoError(err)
+	expectedOutput3, err := ustPool.CalcOutAmtGivenIn(suite.Ctx,
+		sdk.Coins{sdk.Coin{Denom: ust, Amount: coins.AmountOf(ust)}},
+		baseDenom,
+		swapFee)
+	suite.Require().NoError(err)
```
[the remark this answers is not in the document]
ValarDragon: I actually think this is better than foo bar baz

ValarDragon:
Because this is a legacy PR thats been open so long, I'm in favor of merging it, and doing spec updates in a second issue/PR,

Right now theres not really a well formatted spec for this module to begin with, and I don't think its worth blocking the merge which has several useful updates, and will keep getting merge conflicts, until the spec is done

danwt:
Hiya, I was just curious what is the benefit of doing this in epoch as opposed to on the fly? Thanks

Issue #1142, opened 2022-03-24
Separate out mutative methods for joinPool, exitPool and swap functions


mattverse:
Currently we have non mutative operations, `CalcInAmtGivenOut`, ` CalcOutAmtGivenIn`, and mutative operations: `ApplySwap`, `JoinPool`, `ExitPool`. 

In the extent of allowing more code duplications, we want to have more options for the methods we want to have in our poolI interface, something that looks like such: 
```
* CalcInAmtGivenOut
* CalcOutAmtGivenIn
* SwapInAmtGivenOut
* SwapOutAmtGivenIn

* CalcJoinPoolShares
* CalcExitPoolShares
* JoinPool
* ExitPool
```

PR #1147, opened 2022-03-25
Add mutative, non mutative method for `PoolI`

mattverse:
Closes: #1142

## Description

This PR separates out mutative methods for joinPool, exitPool and swap functions for `PoolI`.

Further things to address in the future: changing the existing methods(ex. `SwapExactAmountIn` in swap.go ) so that they use the mutative methods.

______

For contributor use:


x/gamm/pool-models/balancer/amm.go
```diff
@@ -61,7 +63,28 @@ func (p Pool) CalcOutAmtGivenIn(
 	return sdk.NewDecCoinFromDec(tokenOutDenom, tokenAmountOut), nil
 }
 
-// calcInAmtGivenOut calculates token to be provided, fee added,
+// SwapOutAmtGivenIn is a mutative method for CalcOutAmtGivenIn, which includes the actual swap
+func (p Pool) SwapOutAmtGivenIn(
```
ValarDragon, line 67: p Needs to be mutative
```suggestion
func (p *Pool) SwapOutAmtGivenIn(
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -88,6 +111,27 @@ func (p Pool) CalcInAmtGivenOut(
...
+
+	err = p.ApplySwap(ctx, sdk.Coins{tokenInCoin}, tokensOut)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	return tokenInCoin, nil
+}
+
 // ApplySwap.
 func (p *Pool) ApplySwap(ctx sdk.Context, tokensIn sdk.Coins, tokensOut sdk.Coins) error {
```
ValarDragon, line 136: Lets make this unexposed 
```suggestion
func (p *Pool) applySwap(ctx sdk.Context, tokensIn sdk.Coins, tokensOut sdk.Coins) error {
```

mattverse:
Yeah agreed that we need to remove ApplySwap, was planning to make a second PR but seems that we could do it all in this PR!

x/gamm/pool-models/balancer/amm.go
```diff
@@ -61,7 +63,28 @@ func (p Pool) CalcOutAmtGivenIn(
 	return sdk.NewDecCoinFromDec(tokenOutDenom, tokenAmountOut), nil
 }
 
-// calcInAmtGivenOut calculates token to be provided, fee added,
+// SwapOutAmtGivenIn is a mutative method for CalcOutAmtGivenIn, which includes the actual swap
```
alexanderbez, line 66:
```suggestion
// SwapOutAmtGivenIn is a mutative method for CalcOutAmtGivenIn, which includes the actual swap.
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -61,7 +63,28 @@ func (p Pool) CalcOutAmtGivenIn(
...
+func (p *Pool) SwapOutAmtGivenIn(
+	ctx sdk.Context, tokensIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (
```
alexanderbez, line 68:
```suggestion
	ctx sdk.Context,
	tokensIn sdk.Coins,
	tokenOutDenom string,
	swapFee sdk.Dec,
) (
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -61,7 +63,28 @@ func (p Pool) CalcOutAmtGivenIn(
...
+	tokenOut sdk.Coin, err error,
+) {
+	tokenOutDecCoin, err := p.CalcOutAmtGivenIn(ctx, tokensIn, tokenOutDenom, swapFee)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	tokenOutCoin, _ := tokenOutDecCoin.TruncateDecimal()
+	if tokenOutCoin.Amount.LTE(sdk.ZeroInt()) {
```
alexanderbez, line 76:
```suggestion
	if !tokenOutCoin.Amount.IsPositive() {
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -61,7 +63,28 @@ func (p Pool) CalcOutAmtGivenIn(
...
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount is zero or negative")
```
alexanderbez, line 77:
```suggestion
		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -88,8 +111,29 @@ func (p Pool) CalcInAmtGivenOut(
 	return sdk.NewDecCoinFromDec(tokenInDenom, tokenAmountInBeforeFee), nil
 }
 
+// SwapInAmtGivenOut is a mutative method for CalcOutAmtGivenIn, which includes the actual swap
```
alexanderbez, line 114:
```suggestion
// SwapInAmtGivenOut is a mutative method for CalcOutAmtGivenIn, which includes the actual swap.
```

x/gamm/types/pool.go
```diff
@@ -28,21 +28,22 @@ type PoolI interface {
 	GetTotalShares() sdk.Int
 
 	CalcOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.DecCoin, err error)
-	CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.DecCoin, err error)
+	SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.Coin, err error)
```
alexanderbez, line 31: nit: not that is has to be done in these PRs, but I believe that these methods are the meat n bones of the AMM state machine, or are at least a large part of it. Being so, I think each of these methods in the interface here should be very very well documented with godocs.

x/gamm/pool-models/balancer/amm.go
```diff
@@ -88,8 +111,29 @@ func (p Pool) CalcInAmtGivenOut(
...
+func (p *Pool) SwapInAmtGivenOut(
+	ctx sdk.Context, tokensOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (
+	tokenIn sdk.Coin, err error,
+) {
+	tokenInDecCoin, err := p.CalcInAmtGivenOut(ctx, tokensOut, tokenInDenom, swapFee)
+	if err != nil {
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount is zero or negative")
+	}
+	tokenInCoin, _ := tokenInDecCoin.TruncateDecimal()
+	if tokenInCoin.Amount.LTE(sdk.ZeroInt()) {
```
alexanderbez, line 124:
```suggestion
	if !tokenInCoin.Amount.IsPositive() {
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -88,8 +111,29 @@ func (p Pool) CalcInAmtGivenOut(
...
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount is zero or negative")
```
alexanderbez, line 125:
```suggestion
		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
```

alexanderbez (approved):
🚀

x/gamm/pool-models/balancer/amm.go
```diff
@@ -222,35 +266,62 @@ func (p *Pool) maximalExactRatioJoin(tokensIn sdk.Coins) (numShares sdk.Int, rem
...
 	if !remCoins.Empty() {
 		for _, coin := range remCoins {
-			newShares, err := p.singleAssetJoin(coin, swapFee)
+			newShares, err := p.calcSingleAssetJoin(coin, swapFee)
 			if err != nil {
-				return sdk.ZeroInt(), err
+				return sdk.ZeroInt(), sdk.NewCoins(), err
 			}
-			p.updateLiquidity(newShares, sdk.Coins{coin})
+			newLiquidity = newLiquidity.Add(coin)
```
ValarDragon, line 299: So these single asset joins actual need the liquidity updated in order to work :/
This is because the pool reserves and therefore correct amounts out change at each of these steps.

Because the JoinPool and ExitPool methods have a lot of internal steps pool liquidity updates for correctness, I'd suggest doing this in one of two ways:
* Keeping JoinPool and ExitPool as before, but then make the calc methods do a deep copy of the struct. (Probably easier) I don't know if deepcopy / clone methods are going to be straight forward to make, or we'd have to make a function called "clonePoolAssets" to do it.
* Making a common method, that does all of this logic as before, but with pool assets passed in as an argument rather than read off the pool. (It can mutate this pool assets) Then make the calc method use a cloned pool assets struct, and the join pool use the 'real' pool assets struct.

ValarDragon (changes requested):

x/gamm/pool-models/balancer/amm.go
mattverse: Gotcha! How is this passing tests right now tho :/

ValarDragon: Uhhhh that's a very good question.

On first thought, maybe it's that we previously didn't support lping at arbitrary ratios, and therefore dont have tests for this? (Which is then concerning/  my bad for not adding tests in the refactoring)

mattverse:
So I took the approach where we keep the original structure, but pass in poolAssets, poolShares, as a parameter for calculating calcSingleAssetJoin. This way we can use the updated values of poolAssets and poolShares without mutating the original values, then mutate the original value only when joiningpool

p0mvn (approved):
LGTM

x/gamm/types/pool.go
```diff
@@ -28,21 +28,22 @@ type PoolI interface {
...
+	CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.DecCoin, err error)
+	SwapInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.Coin, err error)
 
 	SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom string) (sdk.Dec, error)
 
 	// JoinPool joins the pool, and uses all of the tokensIn provided.
 	// The AMM swaps to whatever the ratio should be and returns the number of shares created.
 	// Internally the pool updates its count for the number of shares in this function.
 	// If the function errors, or should not be mutative, then state must be reverted after this call.
+	CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, err error)
```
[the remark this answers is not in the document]
mattverse: yeah we would have to go through the comments for `PoolI`. Many of them seemed to be like temporary comments for @ValarDragon . I'll leave it to @ValarDragon  to go through / organize the final struct for `PoolI`: doesn't have to be in this PR

ValarDragon (approved):
LGTM!

mattverse:
@ValarDragon Sweet! Can you review the resolved conflicts?

PR #1150, opened 2022-03-25
Switch all create pool logic to use an interface

ValarDragon:
Closes: #1149

## Description

This makes an interface for CreatePool logic, allowing more code re-use / separation between general AMM logic and underlying pool implementation details.

After this change, as far as I know, the keeper struct and its methods should have no code using the balancer package!

______

For contributor use:


ValarDragon:
Going to merge this to unblock some other work around gamm. Please do still review this!

PR #1153, opened 2022-03-26
Fix: txfees module

mattverse:
## Description

This PR handles a possible bug in the txfees module. 
The current test cases could be the most intuitive example, where we created a pool with 100 base denom and 200 foo tokens. If we're using 10 foo tokens as the input fee, we should expect it to be converted to 5 base denom, not 20 base denom. 

This PR also includes deleting ceil for the calculated spot price. Using the same example above that was used for the test case, the user would be required 5.000000000005368710 base tokens, but this would be ceiled and would be required as 6 basetokens.

______

For contributor use:

### DIFF
--- a/x/txfees/keeper/feetokens.go
+++ b/x/txfees/keeper/feetokens.go
@@ -23,15 +23,34 @@ func (k Keeper) ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin
 		return sdk.Coin{}, err
 	}
 
-	spotPrice, err := k.spotPriceCalculator.CalculateSpotPrice(ctx, feeToken.PoolID, feeToken.Denom, baseDenom)
+	spotPrice, err := k.spotPriceCalculator.CalculateSpotPrice(ctx, feeToken.PoolID, baseDenom, feeToken.Denom)
+
 	if err != nil {
 		return sdk.Coin{}, err
 	}
 
-	return sdk.NewCoin(baseDenom, spotPrice.MulInt(inputFee.Amount).Ceil().RoundInt()), nil
+	return sdk.NewCoin(baseDenom, spotPrice.MulInt(inputFee.Amount).RoundInt()), nil
 }
 
-// GetFeeToken returns the fee token record for a specific denom.
+func (k Keeper) CalcFeeSpotPrice(ctx sdk.Context, inputDenom string) (sdk.Dec, error) {
+	baseDenom, err := k.GetBaseDenom(ctx)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	feeToken, err := k.GetFeeToken(ctx, inputDenom)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	spotPrice, err := k.spotPriceCalculator.CalculateSpotPrice(ctx, feeToken.PoolID, baseDenom, feeToken.Denom)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+	return spotPrice, nil
+}
+
+// GetFeeToken returns the fee token record for a specific denom
 func (k Keeper) GetBaseDenom(ctx sdk.Context) (denom string, err error) {
 	store := ctx.KVStore(k.storeKey)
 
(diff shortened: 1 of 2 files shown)


x/txfees/keeper/feetokens.go
```diff
@@ -23,15 +23,34 @@ func (k Keeper) ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin
 		return sdk.Coin{}, err
 	}
 
-	spotPrice, err := k.spotPriceCalculator.CalculateSpotPrice(ctx, feeToken.PoolID, feeToken.Denom, baseDenom)
+	spotPrice, err := k.spotPriceCalculator.CalculateSpotPrice(ctx, feeToken.PoolID, baseDenom, feeToken.Denom)
```
ValarDragon, line 26: Were the arguments here just in the wrong order before / we not have tests checking this?

x/txfees/keeper/feetokens.go
```diff
@@ -23,15 +23,34 @@ func (k Keeper) ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin
...
 	if err != nil {
 		return sdk.Coin{}, err
 	}
 
-	return sdk.NewCoin(baseDenom, spotPrice.MulInt(inputFee.Amount).Ceil().RoundInt()), nil
+	return sdk.NewCoin(baseDenom, spotPrice.MulInt(inputFee.Amount).RoundInt()), nil
 }
 
-// GetFeeToken returns the fee token record for a specific denom.
+func (k Keeper) CalcFeeSpotPrice(ctx sdk.Context, inputDenom string) (sdk.Dec, error) {
```
ValarDragon, line 35: This function looks unused? Is it supposed to be exposed over a query?

x/txfees/keeper/feetokens_test.go
```diff
@@ -177,7 +177,7 @@ func (suite *KeeperTestSuite) TestFeeTokenConversions() {
 			baseDenomPoolInput:  sdk.NewInt64Coin(baseDenom, 100),
 			feeTokenPoolInput:   sdk.NewInt64Coin("foo", 200),
 			inputFee:            sdk.NewInt64Coin("foo", 10),
-			expectedOutput:      sdk.NewInt64Coin(baseDenom, 20),
+			expectedOutput:      sdk.NewInt64Coin(baseDenom, 5),
```
ValarDragon, line 180: Ah good catch, this is definitely a bug

x/txfees/keeper/feetokens.go
ValarDragon: Ah your right, the argumernts were certainly reversed! (The test had the wrong expectation coded)

ValarDragon (approved):
LGTM! Nice catch! Can you add a test case showing the rounding situation as well?

x/txfees/keeper/feetokens.go
mattverse: I meant to take this out haha, wasn't expecting such a fast review! :)

mattverse:
> LGTM! Nice catch! Can you add a test case showing the rounding situation as well?

Hmm, The method itself is implemented with / without ceil, what's an approach I can take to prove it should be without ceil?
Added extra comment as of now!

x/txfees/keeper/feetokens_test.go
```diff
@@ -173,10 +173,11 @@ func (suite *KeeperTestSuite) TestFeeTokenConversions() {
...
-			name:                "unequal value",
-			baseDenomPoolInput:  sdk.NewInt64Coin(baseDenom, 100),
-			feeTokenPoolInput:   sdk.NewInt64Coin("foo", 200),
-			inputFee:            sdk.NewInt64Coin("foo", 10),
+			name:               "unequal value",
+			baseDenomPoolInput: sdk.NewInt64Coin(baseDenom, 100),
+			feeTokenPoolInput:  sdk.NewInt64Coin("foo", 200),
+			inputFee:           sdk.NewInt64Coin("foo", 10),
+			// expected to get 5.000000000005368710 baseDenom without rounding
 			expectedOutput:      sdk.NewInt64Coin(baseDenom, 5),
```
ValarDragon, line 181: oh interesting, this means theres some subtle rounding in the Calc Spot Price function, likely related to the pool weights somehow.

ValarDragon:
Oh setup a pool thats has spot price `1 foo_coin = .666 osmo` or smth, and then see the rounding behavior for how many foo coins it requires.

You can get that spot ratio with a supply of `300 foo, 200 osmo`

x/txfees/keeper/feetokens_test.go
ValarDragon: (Not an issue of any sort, just relates to the precision we quote reuslts as accurate until. We should probably just truncate precision of the quoted result in SpotPrice in the future)

ValarDragon:
I think the rounding behavior test is less important now with #1155 being worked on

PR #1154, opened 2022-03-26
Database stability improvements

p0mvn:
Closes: #XXX

## Description

- upgraded iavl and SDK with the most recent database stability improvements
- changelog updated

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,32 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Minor improvements & Bug Fixes
+
+* [#1052](https://github.com/osmosis-labs/osmosis/pull/1052) Eugen/cherry pick superfluid test scaffolding updates
+* [#1070](https://github.com/osmosis-labs/osmosis/pull/1070) Test improvisation for Superfluid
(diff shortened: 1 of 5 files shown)


ValarDragon (approved):
LGTM! Is the changing the hardcoded constant from 780k to 100k in this branch as well? (I didn't see it on skim in IAVL / cosmos SDK changes)

I think we should publish a release once thats in

p0mvn:
@ValarDragon I'm working on that today to represent the limit in terms of bytes and as a configurable value. Hoping to be done by EOD and merge in a separate PR.

If you'd like to get it out ASAP, I can hardcode to 100K instead of the configurable bytes alternative

ValarDragon:
Oh sweet, didn't realize your going for the limit in bytes at the moment, I thought that was something we'd do much later in the future. I think we should change it to 100k and get that into a release (the release can still come out tomorrow!)

I'm in favor of the limiting by bytes and a config value, but I don't think we should block the next release on that

p0mvn:
Sounds good. I will prepare the PRs with 100K right now.

Will still keep working on the configurable value. If that doesn't get in before the next release, that's okay.

ValarDragon:
My main thinking as to why, is because we may get unforeseen troubles that would delay a release to mid-week with the change to bytes. (E.g. some unexpected edge case, takes longer to review on time, etc.)

I would rather get this out sooner, so we can get more information around the app hash situation, and mitigate the ongoing RAM usage concerns

ValarDragon:
Yeah, we can still make a release as soon as the config gets done / merged!

p0mvn:
@ValarDragon updated sdk and iavl with smaller fast node cache size. Also, prepared changelog for `v7.1.0` release

Issue #1155, opened 2022-03-26
SpotPrice: Improve accuracy and reduce precision of quoted result.


ValarDragon:
Due to no accounting of significant figures in quoted results, we currently get certain things like SpotPrice subtly off. 

cref https://github.com/osmosis-labs/osmosis/pull/1153/files#diff-56292d6dacdcf98204375786e4dc828386beb827a7b61cf3283a86c864b4e6dbR177-R180 where a pool with spot price 'clearly' `.5` now has this complex `.500000000000536871` at the end of it. And then we further propogate these trailing digits that are incorrect.

We should reduce the precision of all quoted spot price results. It would be nice to base this on looking at how accurate / inaccurate our rounding is in the SpotPrice function. The main issue here is that divisions introduce new inaccuracies in the trailing bits, and multiplications amplify them.

The source of the inaccurancy in these common cases, is that the weights in our balancer code all have this factor of `2**20` in them by default. This can be mitigated though, by dividing the weights first, then multiplying by numerator over denominator.

I think we should do two things:
- [ ] Change balancer spotprice function in https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/pool-models/balancer/amm.go#L121-L123 to instead compute `weightRatio`, `supplyRatio` and multiply/divide them in a manner thats mathematically equivalent to the current expression. (With a comment showing they are equivalent, and that we're doing this to cancel out the common term in weight) Make some tests to see spot prices results / if these weird rounding errors persist
- [ ] Add a constant to the gamm module, for limiting the number of sig figs output by SpotPrice. I think there is no justifiable reason to output 18 sig figs in quoted results. I'd suggest limiting to 8 sig figs in results. This limiting can be done in `x/gamm/keeper`

PR #1156, opened 2022-03-26
Move around pool setup logic to re-importable infra

ValarDragon:
The purpose of this PR is to add in some scaffolding for testing the spot price. Adam mentioned that the part he wasn't sure how to do for #1155 / general osmosis code development was the testing infrastructure. So the goal of this PR is to:

- Update more of the testing code to use the `apptesting` module
- Make a table driven test in balancer that can be adapted for working on #1155 

______

For contributor use:


ValarDragon:
@czarcas7ic check out the second commit ( https://github.com/osmosis-labs/osmosis/pull/1156/commits/acbdb8668eb76132eda7a9e33229bd3a6043e7ab )

Basically the way you'd test #1155 with this change is then by editing / adding more test cases in that test, as you do the change. You can setup your editor to run those tests every time you save a file in the balancer package.

x/gamm/pool-models/balancer/suite_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stretchr/testify/suite"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+
+	"github.com/cosmos/cosmos-sdk/baseapp"
+
+	"github.com/osmosis-labs/osmosis/v7/app"
+	"github.com/osmosis-labs/osmosis/v7/app/apptesting"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
```
alexanderbez, line 14: formatting is off here 👍

alexanderbez (approved):
LGTM!

Question, `SetupTest()` is called by Testify's suite before _each_  test. Why do we also call it for each test case?

ValarDragon:
Re SetupTest, maybe we're doing it wrong!

The idea right now is we want it to start on a clean app state, and we set the clean app state on SetupTest. Should this per test-case reset state method have a different name?

alexanderbez:
> Re SetupTest, maybe we're doing it in a way thats against the standard!
> 
> The idea right now is we want it to start on a clean app state, and we set the clean app state on SetupTest. Should this per test-case reset state method have a different name?

Nope, that's totally fine and I assumed as much!

PR #1157, opened 2022-03-27
fix: Liveness CI Test

alexanderbez:
Fix the liveness CI test. It was a dep issue that was being caused by the import of `starport/cmd`. We don't really need this import as we don't use it directly anyway.

______

For contributor use:


ValarDragon (approved):
TY for fixing it!

PR #1176, opened 2022-03-29
Calculate SpotPrice with Ratios and SigFigs

czarcas7ic:
Closes: #1155 

## Description

* Calculates SpotPrice with ratios to prevent trailing digits from giving incorrect results
* Utilizes constant in x/gamm/types to determine how many significant figures we calculate SpotPrice to (other functions can also use this SigFigs constant now)
* Added a test to show a circumstance in which the 9th digit is greater than 5, and with 8 sigfigs set should remove the 9th digit however still round the 8th digit up
* Added comment to explain why this change was added

### DIFF
--- a/x/gamm/pool-models/balancer/amm.go
+++ b/x/gamm/pool-models/balancer/amm.go
@@ -122,9 +122,11 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
-	weightRatio := base.Weight.ToDec().Quo(quote.Weight.ToDec())
+        // spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
+        // spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
+	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
 	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
-	fullRatio := supplyRatio.Quo(weightRatio)
+	fullRatio := supplyRatio.Mul(invWeightRatio)
 	ratio := (fullRatio.Mul(types.SigFigs).RoundInt()).ToDec().Quo(types.SigFigs)
 	return ratio, nil
 }
--- a/x/gamm/pool-models/balancer/amm.go
+++ b/x/gamm/pool-models/balancer/amm.go
@@ -122,8 +122,8 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
-        // spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
-        // spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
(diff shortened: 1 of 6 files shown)


czarcas7ic:
Should I modify all the tests that failed to utilize the 8 sigfigs now?

czarcas7ic:
I basically would just need to change this test here 
https://github.com/osmosis-labs/osmosis/blob/main/x/pool-incentives/keeper/keeper_test.go#L96 
and here 
https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/keeper_test.go#L95

Just lmk what you would like and I will change accordingly

x/gamm/types/constants.go
```diff
@@ -9,6 +9,8 @@ const (
 	MaxPoolAssets = 8
 
 	OneShareExponent = 18
+
+	SigFigsExponent = 8
```
alexanderbez, line 13: Can we add a comment explaining what this variable is and what it's used for?

czarcas7ic: Done!

x/gamm/pool-models/balancer/amm.go
```diff
@@ -118,10 +122,10 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
-	numerator := base.Token.Amount.ToDec().Quo(base.Weight.ToDec())
-	denom := quote.Token.Amount.ToDec().Quo(quote.Weight.ToDec())
-	ratio := numerator.Quo(denom)
-
+	weightRatio := base.Weight.ToDec().Quo(quote.Weight.ToDec())
+	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
+	fullRatio := supplyRatio.Quo(weightRatio)
```
ValarDragon, line 127:
```suggestion
        // spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
        // spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
	fullRatio := supplyRatio.Mul(invWeightRatio)
```

czarcas7ic:
Test modified to include sigfigs, now pass

x/gamm/pool-models/balancer/amm.go
```diff
@@ -118,10 +122,12 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
-	numerator := base.Token.Amount.ToDec().Quo(base.Weight.ToDec())
-	denom := quote.Token.Amount.ToDec().Quo(quote.Weight.ToDec())
-	ratio := numerator.Quo(denom)
-
+	// spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
+	// spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
```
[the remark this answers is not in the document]
ValarDragon: They are equivalent! Its intended to show that they're the same, if you simplify the equation in that way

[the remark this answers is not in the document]
ValarDragon: The first equation is how its commonly quoted, but we're computing according to the second equation for the speed and improved precision due to the weight common denominator

ValarDragon (approved):
LGTM

PR #1178, opened 2022-03-30
Use latest ci on v3.x branch and do linting

faddat:
## Description

Adds latest CI to v3.x and enables linting.

______

For contributor use:


faddat:
I think that in order for this to work right, I'll need to remove caching from CI for anything older than our latest release.  Does anyone see a problem with this?

ValarDragon:
Can we make the liveness tests just be disabled for v3.x, v4.x, and v5.x branches?

ValarDragon:
Can we make the liveness tests just be disabled for v3.x, v4.x, and v5.x branches?

faddat:
yes-- I think I'll specify that in main since pretty sure that it supercedes the rest

p0mvn:
Hi! Seeing this has been open for a while. What is the status of this?

If this is still in progress, let's mark it as a draft. If not planning to merge, let's close please

faddat:
This is complete, and should be merged.  

It is running into some issues with ci, I can have a tinker but not sure they're fixable

czarcas7ic:
We would have to make some changes to the e2e test framework for this CI test to work on all prior version branches. I can look into after this next upgrade, but I would say it's not a priority.

faddat:
ok, I will just remove the file for now and then this one is ready

ValarDragon (approved):
Thank you for doing this to help with maintenance!

I generally don't think we should be updating all the dotfiles / CI / meta files for old versions. But thanks for doing it / most particular concerns I'd have over this have been handled!

Something thats wrong is updating tooling for go 1.18, whereas this old branch does need to support that old go version. I think its ok in this particular case since PR is here, but likely will not accept another PR of the same form for a general sync, unless theres a particular need.

PR #1184, opened 2022-03-31
x/lockup: Get updated lock for events

giansalex:
Current endtime in logs (MsgBeginUnlocking) is wrong: `0001-01-01 00:00:00 +0000 UTC`  because it is updated inside BeginUnlock and eventManager uses outdated lock, i wanted to return the updated lock from that function but it has many references, a quick fix is to get the lock again.

this also includes verifying the owner conditional before BeginUnlock is executed.

> endtime field could also be added to `MsgBeginUnlockingResponse`.
______

For contributor use:

### DIFF
--- a/x/lockup/keeper/msg_server.go
+++ b/x/lockup/keeper/msg_server.go
@@ -95,7 +95,7 @@ func (server msgServer) BeginUnlocking(goCtx context.Context, msg *types.MsgBegi
 	}
 
 	if msg.Owner != lock.Owner {
-		return nil, sdkerrors.Wrap(types.ErrNotLockOwner, fmt.Sprintf("msg sender(%s) and lock owner(%s) does not match", msg.Owner, lock.Owner))
+		return nil, sdkerrors.Wrap(types.ErrNotLockOwner, fmt.Sprintf("msg sender (%s) and lock owner (%s) does not match", msg.Owner, lock.Owner))
 	}
 
 	err = server.keeper.BeginUnlock(ctx, lock.ID, msg.Coins)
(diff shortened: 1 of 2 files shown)


x/lockup/keeper/msg_server.go
```diff
@@ -94,13 +94,18 @@ func (server msgServer) BeginUnlocking(goCtx context.Context, msg *types.MsgBegi
 		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
 	}
 
+	if msg.Owner != lock.Owner {
+		return nil, sdkerrors.Wrap(types.ErrNotLockOwner, fmt.Sprintf("msg sender(%s) and lock owner(%s) does not match", msg.Owner, lock.Owner))
```
alexanderbez, line 98:
```suggestion
		return nil, sdkerrors.Wrap(types.ErrNotLockOwner, fmt.Sprintf("msg sender (%s) and lock owner (%s) does not match", msg.Owner, lock.Owner))
```

alexanderbez (commented):
We should add a changelog entry for this too please 🙏

mattverse (approved):
Thanks for this PR  + the extra check for unlocking! LGTM!

ValarDragon (approved):
LGTM! I wonder if its worth reconsidering at some point how we do events. It does feel a bit silly that we have to do a second GetLockById in the main method. Maybe we should move towards `k.createUnlockEvent(ctx, lockId)` ? Or move events into the keeper methods rather than msg server?

PR #1174, opened 2022-03-29
tests: x/gamm pool JoinSwapExactAmountIn consistency tests

alexanderbez:
## Description

- Adds tests for `JoinSwapExactAmountIn` testing consistency.
- A wee bit of linting/cleanup

Closes: #1116

______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

PR #1186, opened 2022-04-01
chore: x/gamm docs + cleanup

alexanderbez:
## Description

- Cleanup spec/docs
- Cleanup x/gamm godocs and linting

______

For contributor use:

### DIFF
--- a/x/gamm/keeper/pool_service.go
+++ b/x/gamm/keeper/pool_service.go
@@ -79,38 +79,48 @@ func (k Keeper) validateCreatedPool(
 	return nil
 }
 
-func (k Keeper) CreatePool(
-	ctx sdk.Context,
-	msg types.CreatePoolMsg,
-) (uint64, error) {
+// CreatePool attempts to create a pool returning the newly created pool ID or
+// an error upon failure. The pool creation fee is used to fund the community
+// pool. It will create a dedicated module account for the pool and sends the
+// initial liquidity to the created module account.
+//
+// After the initial liquidity is sent to the pool's account, shares are minted
+// and sent to the pool creator. The shares are created using a denomination in
+// the form of gamm/pool/{poolID}. In addition, the x/bank metadata is updated
+// to reflect the newly created GAMM share denomination.
+func (k Keeper) CreatePool(ctx sdk.Context, msg types.CreatePoolMsg) (uint64, error) {
 	err := validateCreatePoolMsg(ctx, msg)
 	if err != nil {
 		return 0, err
 	}
+
 	sender := msg.PoolCreator()
 	initialPoolLiquidity := msg.InitialLiquidity()
+
 	// send pool creation fee to community pool
 	params := k.GetParams(ctx)
-	err = k.distrKeeper.FundCommunityPool(ctx, params.PoolCreationFee, sender)
+	if err := k.distrKeeper.FundCommunityPool(ctx, params.PoolCreationFee, sender); err != nil {
+		return 0, err
+	}
+
+	poolID := k.GetNextPoolNumberAndIncrement(ctx)
+	pool, err := msg.CreatePool(ctx, poolID)
 	if err != nil {
 		return 0, err
 	}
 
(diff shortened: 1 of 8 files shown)


x/gamm/keeper/pool_service.go
```diff
@@ -79,38 +79,48 @@ func (k Keeper) validateCreatedPool(
...
+// to reflect the newly created GAMM share denomination.
+func (k Keeper) CreatePool(ctx sdk.Context, msg types.CreatePoolMsg) (uint64, error) {
 	err := validateCreatePoolMsg(ctx, msg)
 	if err != nil {
 		return 0, err
 	}
+
 	sender := msg.PoolCreator()
 	initialPoolLiquidity := msg.InitialLiquidity()
+
```
mattverse, line 99: just out of curiosity, what are the standards for spacings in between codes? Maybe based on readability?

x/gamm/keeper/swap.go
```diff
@@ -9,26 +9,32 @@ import (
...
+// upon failure. Upon success, the resulting tokens swapped for are returned. A
+// swap fee is applied determined by the pool's parameters.
 func (k Keeper) SwapExactAmountIn(
 	ctx sdk.Context,
 	sender sdk.AccAddress,
 	poolId uint64,
 	tokenIn sdk.Coin,
 	tokenOutDenom string,
 	tokenOutMinAmount sdk.Int,
-) (tokenOutAmount sdk.Int, err error) {
```
mattverse, line 19: Nit: Can we keep this if possible? I think this way's way less confusing

mattverse (approved):
LGTM!
Gamm docs are definitely what we have been needing for a long time!

x/gamm/keeper/pool_service.go
alexanderbez: yup, readability. Break up "logical chunks" to make it easier to understand context.

x/gamm/keeper/swap.go
alexanderbez: Yeah, but here's the thing with named return variables. They're very prone to bugs if you're not careful.

E.g.: https://github.com/osmosis-labs/osmosis/pull/1174/files#diff-14148a5e967be907c5322344ffb4f10ebd6e85b59a995c9e6cc39409235f41de

Can you spot the bug? :p

mattverse: Wow I actually can't. What's the bug here?

alexanderbez: `shareOutAmount` was never actually set! So you'd always be returning the zero value. This is why named returns kind be bug-prone if you're not careful. I wish there was a linter for this because I do agree that named returns offer helpful additional context. I'm happy to keep the original change, but we just have to be very very careful.

I wouldn't be surprised if there's other methods that have this same bug.

mattverse: Wow. I'm seriously surprised. Yeah I think we should make an issue to go over / I can take some extra time to go over other code that might suspectedly have the same bug. 

I'm wondering how this was working before?

x/gamm/keeper/pool_service.go
```diff
@@ -79,38 +79,48 @@ func (k Keeper) validateCreatedPool(
...
 	params := k.GetParams(ctx)
-	err = k.distrKeeper.FundCommunityPool(ctx, params.PoolCreationFee, sender)
-	if err != nil {
+	if err := k.distrKeeper.FundCommunityPool(ctx, params.PoolCreationFee, sender); err != nil {
 		return 0, err
 	}
 
-	poolId := k.GetNextPoolNumberAndIncrement(ctx)
-	pool, err := msg.CreatePool(ctx, poolId)
+	poolID := k.GetNextPoolNumberAndIncrement(ctx)
```
ValarDragon, line 106: Thoughts on leaving poolId as is? We'd have to update the proto files to be consistent everywhere.

alexanderbez: Yup!

Issue #1124, opened 2022-03-22
proposal: introduce an e2e test suite


alexanderbez:
I propose we introduce a programatic e2e integration test suite that allows us to test functionality from the most basic operations like IBC transfers and native sends, to AMM operations.

Examples of this can be found here:

- https://github.com/cosmos/gaia/tree/main/tests/e2e
- https://github.com/umee-network/umee/tree/main/tests/e2e

If there is consensus on this, I'll be happy to tackle it or guide others on the approach used :)

ValarDragon:
This sounds awesome! Its something @nikever @czarcas7ic and @daniel-farina have been looking into

I think we should be really happy to use something that already exists! Is there some docs for how these end to end tests work? (Like how end-to-end are they? Is it starting a single node testnet, and then posting txs & running queries to it?)

alexanderbez:
Yeah, so in each implementation I've done, there is a `doc.go` that outlines what's going on, but I never spent that much time on that file/doc, but we totally could make it very elaborate.

The basic gist of it is:

- Setup data and files (in tmp dirs)
- Build and/or pull docker images (via dockertest)
- Run containers (via dockertest)
- Write your tests in Go! You have access to all gRPC, API and RPC endpoints.

ValarDragon:
Ah sweet, just checked the file! This sounds awesome, it even does IBC tests. I think we should totally use this

daniel-farina:
@czarcas7ic to investigate further with @alexanderbez.

PR #1193, opened 2022-04-05
Setup e2e tests on a single chain; add balances query test

p0mvn:
Closes: #1124

## Description

The package e2e defines an integration testing suite used for full end-to-end
testing functionality.

The file e2e_suite_test.go defines the testing suite and contains the core
bootstrapping logic that creates a testing environment via Docker containers.
A testing network is created dynamically with 2 test validators.

The file e2e_test.go contains the actual end-to-end integration tests that
utilize the testing suite.

Currently, there is a single test in `e2e_test.go` to query the balances of a validator.

More tests are to be added in future PRs

## Test Run
Proof of successful e2e test run: https://github.com/p0mvn/osmosis/runs/5830443179?check_suite_focus=true

## Next Steps

The following can be split into separate tasks / PRs and tackled one by one:

- run a second chain by reusing the current setup
- add ibc relayer functionality and enable ibc tests
- port all tests from:
   * https://github.com/cosmos/gaia/tree/main/tests/e2e
   * https://github.com/umee-network/umee/tree/main/tests/e2e
- identify more e2e tests specific for Osmosis and write up issues to add them

______

For contributor use:


.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Display go version
         run: go version
       - name: Run all tests
-        run: go test -mod=readonly -timeout 30m -coverprofile=coverage.txt -tags='norace' -covermode=atomic `go list ./... | grep -v simapp`
+        run: go test -mod=readonly -timeout 30m -coverprofile=coverage.txt -tags='norace' -covermode=atomic `go list ./... | grep -E -v 'simapp|e2e'`
```
alexanderbez, line 40: nit: we should instead encapsulate this in a make target.

See the changes I made here: https://github.com/cosmos/gaia/blob/main/Makefile#L167-L205

I don't think it's too much trouble to do that in this PR. WDYT?

go.mod
```diff
@@ -29,6 +29,27 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require (
```
alexanderbez, line 32: Did you run `go mod tidy --compat=1.17`? Weird, these should be bundled with the existing `indirect` require block below.

tests/e2e/README.md
```diff
@@ -0,0 +1,13 @@
+# End-to-end Tests
```
alexanderbez, line 1: Don't worry @ValarDragon I'll beef up this doc after merge.

alexanderbez (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

nikever:
Overall LGTM 🚀

I'm actually reverting to the previous `Dockerfile` to avoid these kind of permissions issues so we don't have to create a new `e2e.Dockerfile`

alexanderbez:
> Overall LGTM 🚀
> 
> I'm actually reverting to the previous `Dockerfile` to avoid these kind of permissions issues so we don't have to create a new `e2e.Dockerfile`

So the context on why I originally named it `e2e.Dockerfile` is because before it was just `Dockerfile` and validators/node operators _assumed_ that was the image they should use for mainnet, which it wasn't because it has (or can have) specific E2E stuff in it, not intended for mainnet.

nikever:
> So the context on why I originally named it `e2e.Dockerfile` is because before it was just `Dockerfile` and validators/node operators _assumed_ that was the image they should use for mainnet, which it wasn't because it has (or can have) specific E2E stuff in it, not intended for mainnet.

That makes absolutely sense but in our setup we are just running a _working_ copy of the `Dockerfile` that currently has permission problems if not run correctly.  

Ideally we should try to use the same `Dockerfile` for `prod` and `e2e` but if they will diverge in the future I'm happy with your solution.

go.mod
ValarDragon: Yeah, definitely looks like something went not-quite-right here

tests/e2e/e2e_setup_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	genFilePath := config.GenesisFile()
+	appGenState, genDoc, err := genutiltypes.GenesisStateFromGenFile(genFilePath)
+	s.Require().NoError(err)
+
+	var bankGenState banktypes.GenesisState
+	s.Require().NoError(cdc.UnmarshalJSON(appGenState[banktypes.ModuleName], &bankGenState))
+
+	bankGenState.DenomMetadata = append(bankGenState.DenomMetadata, banktypes.Metadata{
+		Description: "An example stable token",
```
ValarDragon, line 147: Oh wow, I didn't know there was a description field available in the bank module

ValarDragon (approved):
Didn't review scaffolding in depth. But the test looks great / its pretty exciting if it 'justworks.TM'.

Should we do the makefile change in this PR or a separate one?

p0mvn:
Making the makefile change here and then will merge

go.mod
p0mvn: Had to manually move them to the same block and run `go mod tidy`

p0mvn:
On another look, I think the makefile should be addressed in a separate PR. Otherwise, the scope ends up being too large. I will merge this and address it in a separate PR

Issue #1158, opened 2022-03-27
Add queries for getting Upcoming gauges by denom


ValarDragon:
A requested feature is adding a query to full nodes for getting all upcoming gauges by denomination. We can look at the code for 
[ActiveGaugesPerDenom](https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/incentives/query.proto#L42-L46), and basically add that for UpcomingGaugesPerDenom.

Whoever works on this, can use this as an example of how you add a new query: https://github.com/osmosis-labs/osmosis/pull/1136/files

The steps are roughly:
- [x] Add new request definitions in the proto. `UpcomingGaugesPerDenom, UpcomingGaugesPerDenomRequest, UpcomingGaugesPerDenomResponse`
- [x] make proto-all
- [x] Add a keeper method in x/incentives/keeper/grpc_query.go . We can look at this https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/grpc_query.go#L89 and https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/grpc_query.go#L112 for examples of how to do this
- [x] Ditto for adding a test for this in grpc_query_test.go

PR #1195, opened 2022-04-05
Adding upcoming gauges query

xBalbinus:
Closes: #1158 

## Description

Added new request definitions in the proto. UpcomingGaugesPerDenom, UpcomingGaugesPerDenomRequest, UpcomingGaugesPerDenomResponse

Made proto-all

Added a keeper method in x/incentives/keeper/grpc_query.go .

Ditto for adding a test for this in grpc_query_test.go: note: I implemented tests as well for querying active upcoming gauges per denom. At the moment, they're not table driven but still diving deeper in and will implement soon.

______

For contributor use:


proto/osmosis/incentives/query.proto
```diff
@@ -124,6 +131,16 @@ message UpcomingGaugesResponse {
   cosmos.base.query.v1beta1.PageResponse pagination = 2;
 }
 
+message UpcomingGaugesPerDenomRequest {
+  string denom = 1;
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+
+message UpcomingGaugesPerDenomResponse {
+  repeated Gauge data = 1 [ (gogoproto.nullable) = false ];
```
ValarDragon, line 140:
```suggestion
  repeated Gauge upcoming_gauges = 1 [ (gogoproto.nullable) = false ];
```

ValarDragon:
Nice work! I think the `mytestnet` folder got accidentally committed

x/incentives/keeper/grpc_query.go
```diff
@@ -131,6 +131,28 @@ func (k Keeper) UpcomingGauges(goCtx context.Context, req *types.UpcomingGaugesR
...
+				gauges = append(gauges, gauge)
+			}
+		}
+		return true, nil
+	})
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	return &types.UpcomingGaugesPerDenomResponse{Data: gauges, Pagination: pageRes}, nil
```
ValarDragon, line 153: On reviewing this PR, I'm realizing that every filtered paginate usage in this package is wrong. I think its fine to merge this as is though, since its not a real performance overhead / can be fixed later. (We should make an issue to track it)

Filtered paginate's method of working is actually pretty awkward / maybe we need to change it at the SDK level. It gives you the value in `[]bytes`, that this code should then de-marshal, and check if it meets the required constraints. If accumulate is true, only then add it to the gauges result. But this is a high amount of overhead tbh. I think its probably worth getting this refactored in the SDK to be simple, with go 1.18 Generics.

I'll add more detail on this / convert it to an issue, doesn't block this PR!

ValarDragon (approved):
Nice job on your first code change PR!!

LGTM! (It has the same behavior as ActiveGaugesPerDenom)

We should make two follow on issues:
* One brainstorming ways for fixing the erroneous pagination behavior (either we change things directly here, or add  better pagination code scaffolding)
* Steps for getting a better table driven test framework here. (We need to make it easy to make upcoming, present, and completed gauges in the test suite)

x/incentives/keeper/grpc_query_test.go
```diff
@@ -108,6 +108,39 @@ func (suite *KeeperTestSuite) TestGRPCActiveGauges() {
 	suite.Require().Equal(res.Data[0].String(), expectedGauge.String())
 }
 
+func (suite *KeeperTestSuite) TestGRPCActiveGaugesPerDenom() {
+	suite.SetupTest()
+
+		// initial check
+	res, err := suite.app.IncentivesKeeper.ActiveGaugesPerDenom(sdk.WrapSDKContext(suite.ctx), &types.ActiveGaugesPerDenomRequest{})
```
ValarDragon, line 115: oh you have to make a querier struct now for this to work now. (An update happened on main. 

Change to `suite.querier,ActiveGaugesPerDenom` and I think that should work in all places where this is erroring.

ValarDragon:
Can you run `make format` to fix the lint errors

xBalbinus:
Sweet, got you, checking it out!

ValarDragon:
PR LGTM, its having the liveness test timeout, but this confuses me, since this doesn't even touch the state machine? Any ideas @alexanderbez ?

proto/osmosis/incentives/query.proto
```diff
@@ -49,6 +49,13 @@ service Query {
     option (google.api.http).get =
         "/osmosis/incentives/v1beta1/upcoming_gauges";
   }
+
+  rpc UpcomingGaugesPerDenom(UpcomingGaugesPerDenomRequest)
```
alexanderbez, line 53: nit: add a godoc for this handler 👍

alexanderbez: Also, why is it structured like this? Makes it very hard to read. Did we not add the vscode proto formatting changes @ValarDragon?

x/incentives/client/cli/query.go
```diff
@@ -317,6 +318,47 @@ $ %s query incentives upcoming-gauges
 	return cmd
 }
 
+// GetCmdActiveGaugesPerDenom returns active gauges for specified denom.
+func GetCmdUpcomingGaugesPerDenom() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "upcoming-gauges-per-denom [denom]",
+		Short: "Query scheduled gauges per denom",
+		Long: strings.TrimSpace(
```
alexanderbez, line 326: nit: for concise/straight forward CLI handlers like this, I'd argue `Long` is not necessary as it does not offer any additional help or insight.

x/incentives/keeper/grpc_query.go
```diff
@@ -143,6 +143,28 @@ func (q Querier) UpcomingGauges(goCtx context.Context, req *types.UpcomingGauges
 	return &types.UpcomingGaugesResponse{Data: gauges, Pagination: pageRes}, nil
 }
 
+func (q Querier) UpcomingGaugesPerDenom(goCtx context.Context, req *types.UpcomingGaugesPerDenomRequest) (*types.UpcomingGaugesPerDenomResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	gauges := []types.Gauge{}
+	store := ctx.KVStore(q.Keeper.storeKey)
+	valStore := prefix.NewStore(store, types.KeyPrefixUpcomingGauges)
```
alexanderbez, line 150:
```suggestion
	prefixStore := prefix.NewStore(store, types.KeyPrefixUpcomingGauges)
```

x/incentives/keeper/grpc_query.go
```diff
@@ -143,6 +143,28 @@ func (q Querier) UpcomingGauges(goCtx context.Context, req *types.UpcomingGauges
...
+
+	pageRes, err := query.FilteredPaginate(valStore, req.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
+		upcomingGauges := q.Keeper.GetUpcomingGauges(ctx)
+		for _, gauge := range upcomingGauges {
+			if gauge.DistributeTo.Denom == req.Denom {
+				gauges = append(gauges, gauge)
+			}
+		}
+		return true, nil
```
alexanderbez, line 159: Hmmm, why are we returning true here?

x/incentives/keeper/grpc_query.go
```diff
@@ -143,6 +143,28 @@ func (q Querier) UpcomingGauges(goCtx context.Context, req *types.UpcomingGauges
 	return &types.UpcomingGaugesResponse{Data: gauges, Pagination: pageRes}, nil
 }
 
+func (q Querier) UpcomingGaugesPerDenom(goCtx context.Context, req *types.UpcomingGaugesPerDenomRequest) (*types.UpcomingGaugesPerDenomResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
```
alexanderbez, line 147: Nit: I know the other queries don't do this, which we need to address IMO, but we should have an explicit `nil` check here for `req`.

i.e.
```go
if req == nil {
  	if req == nil {
		return nil, status.Error(codes.InvalidArgument, "empty request")
	}
}

	if req.Denom == "" {
		return nil, status.Error(codes.InvalidArgument, "invalid denom")
	}

```

alexanderbez: Granted, we probably shouldn't use `InvalidArgument` for `req` being nil, but rather something else.

alexanderbez (changes requested):
Looks great @xBalbinus!

I left some minor feedback. I'm also taking a look at the liveness test.

proto/osmosis/incentives/query.proto
ValarDragon: Theres no vscode proto formatting AFAIK. I'm down for us to add it. (IIRC you did it via clang, not sure how that install works across machines / operating systems)

alexanderbez: https://github.com/osmosis-labs/osmosis/pull/1221

:P

alexanderbez:
Re liveness failures:

```
➜ starport chain serve --reset-once -v -c ./starport.ci.yml
[STARPORT] Cosmos SDK's version is: stargate - v0.45.1
[STARPORT] 
[STARPORT] 🛠️  Building proto...
[STARPORT] cannot build app:
[STARPORT] 
[STARPORT]      cannot build app:
[STARPORT] 
[STARPORT]      error while running command /var/folders/jl/532qskgn2y32m_qg3p4l8q440000gn/T/protoc832737936: osmosis/incentives/query.proto:136:54: Field number 1 has already been used in "osmosis.incentives.UpcomingGaugesPerDenomRequest" by field "denom".
[STARPORT] : exit status 1
```


Just an FYI, if liveness fails or timesout, you can always run it locally to find out:

```
$ starport chain serve --reset-once -v -c ./starport.ci.yml
```

proto/osmosis/incentives/query.proto
```diff
@@ -124,6 +131,16 @@ message UpcomingGaugesResponse {
   cosmos.base.query.v1beta1.PageResponse pagination = 2;
 }
 
+message UpcomingGaugesPerDenomRequest {
+  string denom = 1;
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
```
alexanderbez, line 136: We gotta be more careful with reviewing proto structs.

```suggestion
  string denom = 1;
  cosmos.base.query.v1beta1.PageRequest pagination = 2;
```

alexanderbez (changes requested):

proto/osmosis/incentives/query.proto
ValarDragon: ah cool! Make proto-format should fix this as well.

@xBalbinus can you run `make proto-format` and commit that change before we merge this in

proto/osmosis/incentives/query.proto
xBalbinus: Oop! Updated.

ValarDragon: Yeah your totally right. Is there a way we can lint for this more easily? I don't get why make proto-all even compiled with this as an issue.

x/incentives/keeper/grpc_query.go
ValarDragon: Hrmm, lets make a second issue for this? It'd be nice if theres something we can do one layer up to save if `if req == nil` in every single querier logic call

alexanderbez: Sounds good! If you create it, you can assign it to me and I'll knock it out 👍

xBalbinus:
make proto-format done; changes committed! Thanks @alexanderbez and @ValarDragon for your input & help.

x/incentives/keeper/grpc_query_test.go
```diff
@@ -109,6 +109,39 @@ func (suite *KeeperTestSuite) TestGRPCActiveGauges() {
...
+	suite.Require().NoError(err)
+	suite.Require().Len(res.Data, 0)
+
+	// create a gauge
+	gaugeID, _, coins, startTime := suite.SetupNewGauge(false, sdk.Coins{sdk.NewInt64Coin("stake", 10)})
+
+	// final check
+	res, err = suite.querier.ActiveGaugesPerDenom(sdk.WrapSDKContext(suite.ctx), &types.ActiveGaugesPerDenomRequest{})
+	suite.Require().NoError(err)
+	suite.Require().Len(res.Data, 1)
```
ValarDragon, line 126: This is failing right now, which is actually correct behavior at the moment. This is because gauges only move from upcoming -> active during the epoch logic at the moment. (We should change the state machine to make it possible for gauges to start out as active, not always as upcoming)

You should call `k.BeginDistribution(suite,ctx, gauge)` before this line to make it work. (Also define the gauge return value from `suite.SetupNewGauge`

We should do the same thing to fix the UpcomingGaugesPerDenom test

x/incentives/keeper/grpc_query.go
ValarDragon: Issue made #1224

x/incentives/client/cli/query.go
alexanderbez: bump @xBalbinus

x/incentives/keeper/grpc_query.go
alexanderbez: bump @xBalbinus

x/incentives/keeper/grpc_query.go
alexanderbez, line 147: Now that we've cleaned up the queries nil checking, we should add it here.

x/incentives/keeper/grpc_query.go
```diff
@@ -145,6 +145,16 @@ func (q Querier) UpcomingGauges(goCtx context.Context, req *types.UpcomingGauges
 
 func (q Querier) UpcomingGaugesPerDenom(goCtx context.Context, req *types.UpcomingGaugesPerDenomRequest) (*types.UpcomingGaugesPerDenomResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	if req == nil {
+		if req == nil {
+			return nil, status.Error(codes.InvalidArgument, "empty request")
+		}
```
ValarDragon, line 151: double if req == nil, lets remove the inner if statement

alexanderbez (approved):
🚀🚀🚀🚀🚀🚀🚀🚀🚀

alexanderbez:
I'm confused as to how CI is passing. Starport liveness test reports:

```
[STARPORT] # github.com/osmosis-labs/osmosis/v7/x/incentives/keeper
[STARPORT] ../../x/incentives/keeper/grpc_query.go:173:47: unknown field 'Data' in struct literal of type "github.com/osmosis-labs/osmosis/v7/x/incentives/types".UpcomingGaugesPerDenomResponse
[STARPORT] : exit status 2
```

xBalbinus:
Should we fix the liveliness tests first before committing? What might be causing this?

alexanderbez:
> Should we fix the liveliness tests first before committing? What might be causing this?

@xBalbinus the error above tells you:

```
[STARPORT] ../../x/incentives/keeper/grpc_query.go:173:47: unknown field 'Data' in struct literal of type "github.com/osmosis-labs/osmosis/v7/x/incentives/types".UpcomingGaugesPerDenomResponse

```

xBalbinus:
I just pushed out some fixes / changes & make all has been doing well for me (all tests passed). Let me know if the liveliness is still an issue.

ValarDragon:
So to fix that liveness error, can you do `make proto-all`, and then replace the usage of Data on line 173, with `UpcomingGauges` which is what should get written after `make proto-all` is ran? That should hopefully fix the liveness test

Issue #1185, opened 2022-03-31
[x/gamm][Misc improvements] Msg defined twice in different proto files


pyramation:
I'm working on a workflow to produce TS definitions and using `protoc` to generate them. In this process, I'm getting an error 

```
osmosis/gamm/v1beta1/tx.proto:9:9: "osmosis.gamm.v1beta1.Msg" is already defined in file "osmosis/gamm/pool-models/balancer/tx.proto".
```

https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/pool-models/balancer/tx.proto#L9

https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/v1beta1/tx.proto#L9

Seems that the `Msg` identifier is being used twice. Maybe not an issue in go, but in generating via `ts-proto` it seems to not be happy.

pyramation:
for more context, here is the command I'm running in the repo w all protobufs: https://github.com/pyramation/osmosis-protobufs/blob/master/bin/run-proto.sh

pyramation:
I saw the progress on the PR and tested out the protoc gen — the current PR works  🙌🏻

ValarDragon:
Matt had a good point: https://github.com/osmosis-labs/osmosis/pull/1208#discussion_r843559703

We should change the txs to be defined under the correct package, and just leave the balancer pool defined under the wrong proto package (which won't cause conflicts)

pyramation:
great, I like the idea of `osmosis.gamm.pool-models.BalacerMsg` or a path that makes sense 🙌🏻

PR #1208, opened 2022-04-06
Change the name of the balancer proto service from Msg -> BalancerMsg

ValarDragon:
This previously was causing problems for typescript proto generation,
because we have Balancer's proto Msg service and gamm's proto Msg service
existing at the same proto import path.

They existed at the same import path, because due to interface serialization
depending on the amino route (which depended on the proto import path),
changing the import path requires a state migration thats a bit cumbersome to
do.

Closes: #1185 

______

For contributor use:


proto/osmosis/txfees/v1beta1/query.proto
```diff
@@ -19,7 +19,8 @@ service Query {
 
   rpc DenomSpotPrice(QueryDenomSpotPriceRequest)
       returns (QueryDenomSpotPriceResponse) {
-    option (google.api.http).get = "/osmosis/txfees/v1beta1/spot_price_by_denom";
+    option (google.api.http).get =
+        "/osmosis/txfees/v1beta1/spot_price_by_denom";
```
ValarDragon, line 23: done by proto formatting automatically

x/gamm/pool-models/balancer/tx.pb.go
```diff
@@ -188,72 +188,72 @@ var _ grpc.ClientConn
...
+		return srv.(BalancerMsgServer).CreateBalancerPool(ctx, req.(*MsgCreateBalancerPool))
 	}
 	return interceptor(ctx, in, info, handler)
 }
 
-var _Msg_serviceDesc = grpc.ServiceDesc{
-	ServiceName: "osmosis.gamm.v1beta1.Msg",
-	HandlerType: (*MsgServer)(nil),
+var _BalancerMsg_serviceDesc = grpc.ServiceDesc{
+	ServiceName: "osmosis.gamm.v1beta1.BalancerMsg",
```
ValarDragon, line 251: Is this service name at all important? 

an alternative approach to fixing this problem entirely, is to migrate the balancer pools and not deal with this duplicate import situation

mattverse: why is still it creating BalancerMsg under v1beta1?  I was assuming it should be created as something like `osmosis.gamm.pool-models.BalacerMsg`

Wdym by migrating balancer pools?

proto/osmosis/gamm/pool-models/balancer/tx.proto
```diff
@@ -6,7 +6,7 @@ import "osmosis/gamm/pool-models/balancer/balancerPool.proto";
 
 option go_package = "github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer";
 
-service Msg {
+service BalancerMsg {
```
alexanderbez, line 9: Can you explain how/where this service is used? Does this actually process on-chain messages?

alexanderbez (approved):

x/gamm/pool-models/balancer/tx.pb.go
ValarDragon: oh thats a good point, we could make txs under the correct package, but just leave the pool proto definition under the legacy wrong one

proto/osmosis/gamm/pool-models/balancer/tx.proto
ValarDragon: It should process on-chain messages at least, I have not confirmed on a testnet. The service name as far as I can tell is just for generated protobuf structs, we have methods to register the proto service, linking the go struct to the proto message

x/gamm/pool-models/balancer/tx.pb.go
ValarDragon: I think your proposal is a better approach

proto/osmosis/gamm/pool-models/balancer/tx.proto
alexanderbez: Sorry if this is obvious, but why are they not all under a single `service?

mattverse: https://github.com/bufbuild/buf/issues/99#issuecomment-655841579

ValarDragon:
Going to close this as stale for now. Lets update with the fix matt pointed out

PR #1209, opened 2022-04-06
Add proto docs

mattverse:
## Description

This PR adds proto docs in docs/core directory, which is auto-generated whenever we run `make proto-gen`. 
Also modified protocgen.sh file which is triggered by make proto-gen. 

As of now, the proto-docs.md includes all proto documentations except `gamm/pool-models/balancer` 
cref: https://github.com/bufbuild/buf/issues/99#issuecomment-655834414
(Also explains why we’re currently getting no errors when we do `proto-gen`)

We have duplicate symbols in https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/pool-models/balancer/tx.proto#L9  and https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/v1beta1/tx.proto#L9, and as a getaway ( we might need a permanent solution for this), restricted the maxdepth of proto docs generation, which excludes `gamm/pool-models/balancer` 

______

For contributor use:

### DIFF
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -29,7 +29,6 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
 done
 
-
 # command to generate docs using protoc-gen-doc
 buf protoc \
   -I "proto" \
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -27,7 +27,6 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
   -I "$cosmos_sdk_dir/proto" \
   --grpc-gateway_out=logtostderr=true:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
-  # command to generate docs using protoc-gen-doc
 done
 
 
@@ -39,8 +38,6 @@ buf protoc \
   --doc_out=./docs/core \
   --doc_opt=./docs/protodoc-markdown.tmpl,proto-docs.md \
   $(find "$(pwd)/proto" -maxdepth 4 -name '*.proto')
-go mod tidy
-
 
 cp -r ./github.com/osmosis-labs/osmosis/v*/x/* x/
 rm -rf ./github.com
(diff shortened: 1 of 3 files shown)


czarcas7ic (approved):
This is great, @daniel-farina knows how to set it up to track this from the docs site so I believe he can set that up after this gets merged.

At some point, we should take the time to add descriptions to all these protos. I think a majority of the osmosis specific ones are null.

scripts/protocgen.sh
```diff
@@ -27,7 +27,20 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
...
+
+# command to generate docs using protoc-gen-doc
+buf protoc \
+  -I "proto" \
+  -I "$cosmos_sdk_dir/third_party/proto" \
+  -I "$cosmos_sdk_dir/proto" \
+  --doc_out=./docs/core \
+  --doc_opt=./docs/protodoc-markdown.tmpl,proto-docs.md \
+  $(find "$(pwd)/proto" -maxdepth 4 -name '*.proto')
+go mod tidy
```
alexanderbez, line 42: Why do we tidy here?

In any case, we should ensure this has the 1.17 compat:

```
$ go mod tidy -compat=1.17
```

scripts/protocgen.sh
```diff
@@ -27,7 +27,20 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
   -I "$cosmos_sdk_dir/proto" \
   --grpc-gateway_out=logtostderr=true:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
+  # command to generate docs using protoc-gen-doc
```
alexanderbez, line 30:
```suggestion
```

scripts/protocgen.sh
mattverse: actually no need to tidy here!

scripts/protocgen.sh
```diff
@@ -29,5 +29,15 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
 done
 
+
```
alexanderbez, line 32:
```suggestion
```

alexanderbez (approved):
🌮🌮🌮

PR #1214, opened 2022-04-06
Add boilerplate for stableswap

ValarDragon:
Adds boilerplate proto definitions for stableswap. This should not affect the current state machine
______

For contributor use:


x/gamm/pool-models/balancer/balancer_pool.go
```diff
@@ -517,47 +517,6 @@ func (params PoolParams) GetPoolExitFee() sdk.Dec {
 	return params.ExitFee
 }
 
-func ValidateFutureGovernor(governor string) error {
```
ValarDragon, line 520: This function was duplicated, and existed in the types package.

proto/osmosis/gamm/pool-models/stableswap/stableswapPool.proto
```diff
@@ -0,0 +1,65 @@
+syntax = "proto3";
```
alexanderbez, line 1: nit: can the file name be snake-cased? `stableswap_pool.proto`?

proto/osmosis/gamm/pool-models/stableswap/stableswapPool.proto
```diff
@@ -0,0 +1,65 @@
...
+    (gogoproto.nullable) = false
+  ];
+  string exitFee = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.moretags) = "yaml:\"exit_fee\"",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message Pool {
```
alexanderbez, line 32: nit: proto doc :p

proto/osmosis/gamm/pool-models/stableswap/tx.proto
```diff
@@ -0,0 +1,32 @@
...
+      returns (MsgCreateStableswapPoolResponse);
+}
+
+// ===================== MsgCreatePool
+message MsgCreateStableswapPool {
+  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
+
+  PoolParams poolParams = 2 [ (gogoproto.moretags) = "yaml:\"pool_params\"" ];
+
+  repeated cosmos.base.v1beta1.Coin initialPoolLiquidity = 3 [
```
alexanderbez, line 21:
```suggestion
  repeated cosmos.base.v1beta1.Coin initial_pool_liquidity = 3 [
```

proto/osmosis/gamm/pool-models/stableswap/tx.proto
```diff
@@ -0,0 +1,32 @@
...
+import "osmosis/gamm/pool-models/stableswap/stableswapPool.proto";
+
+option go_package = "github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/stableswap";
+
+service Msg {
+  rpc CreateStableswapPool(MsgCreateStableswapPool)
+      returns (MsgCreateStableswapPoolResponse);
+}
+
+// ===================== MsgCreatePool
```
alexanderbez, line 15:
```suggestion
```

alexanderbez (commented):
Curious, why the decision to have a sub-package called under `x/gamm`? Instead of:

- directly in x/gamm, OR
- separate module entirely

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (pa Pool) GetTotalShares() sdk.Int {
+	return pa.TotalShares.Amount
+}
+
+// These should all get moved to amm.go
+func (pa Pool) CalcOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.DecCoin, err error) {
+	return sdk.DecCoin{}, types.ErrNotImplemented
+}
+func (pa *Pool) SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.Coin, err error) {
```
p0mvn, line 58: Asking to learn:

I noticed that in the `balancer` package some of the methods that define `PoolI` are in `amm.go` and others are in `balancer_pool.go`.

Is there a reason for that?

Would it make sense to rename `balancer/balancer_pool.go` to `balancer/pool.go` so that the pool implementations are easier to navigate due to the consistency in naming across `pool-models`

alexanderbez: Agree with @p0mvn here. A single file `pool.go` makes sense to me.

ValarDragon: I think a single `pool.go` file makes sense.

I do think having all the math files in one location is preferrable though. If theres boilerplate wrapper / error checking, we could do that in pool.go and then call into a private function in `amm.go`

ValarDragon:
I'm down for changing how we do the proto package structure. I don't know whats standard. 

Go-side, I think the `x/gamm/pool-models/balancer` and `x/gamm/pool-models/stableswap; is right, we eventually may have dozens here, and this packages them by type. Don't want to be in a situation where we have `5+` pool models clogging the top level go structure, and they each have disparate names that makes the folder structure not as good.

proto/osmosis/gamm/pool-models/stableswap/stableswapPool.proto
ValarDragon: Not really sure what doc to put here yet, I think It'll become clearer when I go to make one go-side after an initial impl of the math. Added a placeholder for now though

mattverse (approved):
LGTM!

I also agree with @ValarDragon that we maintain the sub-package structure where different gamm models are structured like `x/gamm/pool-models/balancer` and `x/gamm/pool-models/stableswap`. One of the reasons I'm in support of this structure is that go import cycle can be rather tricky and keeping a sub-package is helpful in managing this import cycle structure between packages.

ValarDragon:
Lets discuss package structure in a separate issue, going to go ahead and merge this to simplify rebasing further logic work on stableswap!

Issue #1201, opened 2022-04-05
add ibc relayer functionality and enable ibc tests


p0mvn:
**Background**

We should have 2 chains running in our e2e tests. From there on, we can connect them via ibc.

The following steps need to be taken to port this functionality:
1. [Port hermes dockerfile](https://github.com/cosmos/gaia/blob/main/tests/e2e/docker/hermes.Dockerfile)
2. [ Port hermes setup](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/tests/e2e/scripts/hermes_bootstrap.sh) script
3. [Add hermes](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/tests/e2e/e2e_setup_test.go#L51) to the suite
4. Port [this call](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/tests/e2e/e2e_setup_test.go#L96) in the setup and [its implementation](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/tests/e2e/e2e_setup_test.go#L328)
5. add ibc test similar to [this one](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/tests/e2e/e2e_test.go#L11)
6. Add[ this makefile step](https://github.com/cosmos/gaia/blob/a82716775ef42b359f747be95f9cc5f627877afb/Makefile#L202-L205). In addition, address TODO by publishing the image to our dockr hub

**Acceptance Criteria**
- [ ] ibc setup added
- [ ] ibc test passes locally
- [ ] ibc test passes in CI
- [ ] hermes image is uploaded to dockerhub
- [ ] port all tests where applicable from:
       * https://github.com/cosmos/gaia/tree/main/tests/e2e
       * https://github.com/umee-network/umee/tree/main/tests/e2e

PR #1216, opened 2022-04-07
e2e with IBC tx and test

czarcas7ic:
Closes: #1201 

## Description

Please note, a majority of this is the original work of @alexanderbez and this is a simple port over from his work on gaia to osmosis.

- Added IBC functionality through hermes and added a single IBC test to ensure the tx was received on chainB from chainA
- Changed from `name` to `id` as requested by Dev
- Commented out TestQueryBalance. This is because we are currently relying on the order of tests for this test to pass. As we develop further understanding of how we want these tests to be ran, we can re-enable this and/or use it elsewhere.

______

For contributor use:


tests/e2e/e2e_util_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	var balancesResp banktypes.QueryAllBalancesResponse
+	if err := cdc.UnmarshalJSON(bz, &balancesResp); err != nil {
+		return nil, err
+	}
+
+	return balancesResp.Balances, nil
+}
+
+func queryOsmoDenomBalance(endpoint, addr, denom string) (sdk.Coin, error) {
```
p0mvn, line 160: Seems to be unused?

tests/e2e/e2e_util_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	txResp := result["tx_response"].(map[string]interface{})
+	if v := txResp["code"]; v.(float64) != 0 {
+		return fmt.Errorf("tx %s failed with status code %v", txHash, v)
+	}
+
+	return nil
+}
+
+func queryOsmoAllBalances(endpoint, addr string) (sdk.Coins, error) {
```
p0mvn, line 139: Is this used?

tests/e2e/e2e_util_test.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+	s.Require().NoErrorf(
+		err,
+		"failed to send IBC tokens; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
+	)
+
+	s.T().Log("successfully sent IBC tokens")
+}
+
+func queryOsmoTx(endpoint, txHash string) error {
```
p0mvn, line 114: Not used yet? Should we start committing things only if we are actually using them in tests?

tests/e2e/e2e_util_test.go
p0mvn: Also, from what I remember this wouldn't work. I was adapting `TestQueryBalances` from this test, and I think it wouldn't work in osmosis if we were to run it

tests/e2e/scripts/hermes_bootstrap.sh
```diff
@@ -0,0 +1,72 @@
...
+trusting_period = '14days'
+trust_threshold = { numerator = '1', denominator = '3' }
+EOF
+
+# import keys
+hermes keys restore ${OSMO_B_E2E_CHAIN_ID} -n "val01-gaia-b" -m "${OSMO_B_E2E_VAL_MNEMONIC}"
+hermes keys restore ${OSMO_A_E2E_CHAIN_ID} -n "val01-gaia-a" -m "${OSMO_A_E2E_VAL_MNEMONIC}"
+
+# start Hermes relayer
+hermes start
```
p0mvn, line 72: new line

tests/e2e/e2e_test.go
```diff
@@ -1,70 +1,71 @@
...
-		actualDenomsB = append(actualDenomsB, balanceB.GetDenom())
-		actualBalancesB = append(actualBalancesB, balanceB.Amount.Uint64())
-	}
-
-	s.Require().ElementsMatch(expectedDenoms, actualDenomsA)
-	s.Require().ElementsMatch(expectedBalancesA, actualBalancesA)
-	s.Require().ElementsMatch(expectedDenoms, actualDenomsB)
-	s.Require().ElementsMatch(expectedBalancesB, actualBalancesB)
-}
+// func (s *IntegrationTestSuite) TestQueryBalances() {
```
p0mvn, line 14: I'd prefer if we make this work right away instead of committing dead code that previously worked.

I think for now we should figure out how we can enforce a specific order and let tests run after the other so that we can expect specific values. Later, we can refactor in terms of setup/teardown before each test.

What do you think?

p0mvn (changes requested):
Great work on this. Left some comments, please take a look

Also

tests/e2e/docker/hermes.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM informalsystems/hermes:0.12.0 AS hermes-builder
```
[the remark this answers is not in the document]
czarcas7ic: Yes there is, great point, will update

tests/e2e/e2e_setup_test.go
```diff
@@ -340,6 +349,94 @@ func (s *IntegrationTestSuite) runValidators(c *chain, portOffset int) {
...
+	_, err = copyFile(
+		filepath.Join("./scripts/", "hermes_bootstrap.sh"),
+		filepath.Join(hermesCfgPath, "hermes_bootstrap.sh"),
+	)
+	s.Require().NoError(err)
+
+	s.hermesResource, err = s.dkrPool.RunWithOptions(
+		&dockertest.RunOptions{
+			Name:       fmt.Sprintf("%s-%s-relayer", s.chainA.id, s.chainB.id),
+			Repository: "ghcr.io/cosmos/hermes-e2e",
```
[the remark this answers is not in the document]
czarcas7ic: @nikever do you know how to add this as an image to our repo? I'm not positive of the process

Makefile
```diff
@@ -238,6 +238,11 @@ benchmark:
 docker-build-debug:
 	@docker build -t osmolabs/osmosisd-e2e --build-arg IMG_TAG=debug -f e2e.Dockerfile .
 
+# TODO: Push this to the Cosmos Dockerhub so we don't have to keep building it
```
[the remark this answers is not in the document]
czarcas7ic: @nikever can we link on getting this on our docker hub as well?

tests/e2e/e2e_util_test.go
czarcas7ic: agreed, removed

tests/e2e/e2e_util_test.go
czarcas7ic: no, removed!

tests/e2e/e2e_util_test.go
czarcas7ic: agreed, removed as well!

Makefile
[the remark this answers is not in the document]
p0mvn: I would also like to learn about how to get access to our docker hub. Please let me know if it's possible to get permissions for that

tests/e2e/e2e_setup_test.go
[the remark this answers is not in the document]
p0mvn: To avoid blocking this PR, let's create an issue to address this separately? 

It should be fine to use the `cosmos` image temporarily, The main concern about depending on external images, especially with the latest tag, is that whenever they make a change, our tests might get broken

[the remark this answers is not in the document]
nikever: The credentials for the `osmobot` Docker account are saved as GitHub Secrets. You can use them to build the Image via CI as in https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/docker.yml#L24-L29

I will also create personal accounts for both of you on our Docker account.

Lets use a new repository `osmolabs/osmosis-e2e` for all these new images

tests/e2e/e2e_test.go
czarcas7ic: Sounds good! I got it to work, but I decided to remove gas fees in order to make calculations simpler (still kept the const so we can add fees if we want in the future)

tests/e2e/e2e_test.go
```diff
@@ -11,45 +11,47 @@ import (
...
+	balancesA, err := queryBalances(chainAAPIEndpoint, s.chainA.validators[0].keyInfo.GetAddress().String())
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesA)
+	// s.Require().Equal(2, len(balancesA))
+
+	chainBAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainB.id][0].GetHostPort("1317/tcp"))
+	balancesB, err := queryBalances(chainBAPIEndpoint, s.chainB.validators[0].keyInfo.GetAddress().String())
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesB)
+	// s.Require().Equal(2, len(balancesB))
```
p0mvn, line 32: let's restore this

tests/e2e/e2e_test.go
```diff
@@ -11,45 +11,47 @@ import (
...
+		expectedDenomsB   = []string{osmoDenom, stakeDenom, ibcDenom}
+		expectedBalancesA = []uint64{osmoBalanceA - ibcSendAmount, stakeBalanceA - stakeAmountA}
+		expectedBalancesB = []uint64{osmoBalanceB, stakeBalanceB - stakeAmountB, ibcSendAmount}
+	)
+
+	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+	balancesA, err := queryBalances(chainAAPIEndpoint, s.chainA.validators[0].keyInfo.GetAddress().String())
+	s.Require().NoError(err)
+	s.Require().NotNil(balancesA)
+	// s.Require().Equal(2, len(balancesA))
```
p0mvn, line 26: and this one

tests/e2e/e2e_test.go
p0mvn: Looking good. Yeah let's add this later

p0mvn (approved):
Please restore commented-out lines in `TestQueryBalances`. Other than that LGTM!

Makefile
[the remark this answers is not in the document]
alexanderbez: Yes, good idea. It's very easy to push to our org. All you need is a github PAT (personal access token) and then you login with that and just push it. Note, just make sure the package set to public.

tests/e2e/e2e_setup_test.go
```diff
@@ -340,6 +349,94 @@ func (s *IntegrationTestSuite) runValidators(c *chain, portOffset int) {
...
+				"-c",
+				"chmod +x /root/hermes/hermes_bootstrap.sh && /root/hermes/hermes_bootstrap.sh",
+			},
+		},
+		noRestart,
+	)
+	s.Require().NoError(err)
+
+	endpoint := fmt.Sprintf("http://%s/state", s.hermesResource.GetHostPort("3031/tcp"))
+	s.Require().Eventually(
```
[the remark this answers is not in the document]
alexanderbez: Validators? You mean the node containers themselves? I believe we already have a health check for that. Are you suggesting that we implement a test that queries for validators?

tests/e2e/e2e_test.go
alexanderbez: So why is this commented out? Is this because or test execution order?

alexanderbez (approved):
🎉 🎉 🎉

tests/e2e/e2e_test.go
alexanderbez: Yeah, I'm a bit torn. Typically, the purist in me would state that test execution order should be independent. However, there are exceptions, and I feel like integration/e2e tests where there is a single global state is certainly one of them.

The only way I know for explicitly defining test execution order is by the test name as testify's suite package executes them based on alphabetical order. We can probably achieve this via a prefix and/or suffix to each test name.

czarcas7ic:
Had a call with @nikever, he was able to get our own hosted hermes image and integrated it into the e2e test. He will now attempt a small modification on the gh action and after he pushes this we can merge this pr

Makefile
```diff
@@ -238,11 +238,6 @@ benchmark:
 docker-build-debug:
 	@docker build -t osmolabs/osmosisd-e2e --build-arg IMG_TAG=debug -f e2e.Dockerfile .
 
-# TODO: Push this to the Cosmos Dockerhub so we don't have to keep building it
-# in CI.
-docker-build-hermes:
```
p0mvn: Should we still have this for convenience to build locally?

tests/e2e/docker/hermes.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM informalsystems/hermes:0.13.0 AS hermes-builder
+FROM osmolabs/hermes:0.13.0 AS hermes-builder
```
p0mvn, line 1: Why would we host our own `hermes` image?

I think we should only store `hermes-e2e` image because this is something with regards to the e2e tests that are "under our control". We also have Dockerfile for building the `hermes-e2e` image in this repository.

On the other hand, we are not developing or building hermes itself. So I think we should just reuse their image. Then, when there is a new version, we can just bump it up here. If we store our own, we have to upload a new one to our docker hub every time, and I don't really see a reason for doing that.

Let me know what you think

tests/e2e/e2e_setup_test.go
```diff
@@ -372,12 +372,19 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
 	s.hermesResource, err = s.dkrPool.RunWithOptions(
 		&dockertest.RunOptions{
 			Name:       fmt.Sprintf("%s-%s-relayer", s.chainA.id, s.chainB.id),
-			Repository: "ghcr.io/cosmos/hermes-e2e",
-			Tag:        "latest",
+			Repository: "osmolabs/hermes",
```
p0mvn, line 375: shouldn't this be `hermes-e2e`?

p0mvn (changes requested):

tests/e2e/e2e_setup_test.go
nikever: I don't think so.

They called it `hermes-e2e` because it was a modified version of the official `hermes` to be able to run tests (mainly due to permission problems). 

WIth the changes made we can use the official  `hermes`  image to run tests saving us the time needed to build the `hermes-e2e` image.

tests/e2e/docker/hermes.Dockerfile
nikever: Since we don't need to build the `hermes-e2e` image anymore we can use the official `hermes` image.

If for whatever reason in the future we need to deploy a specific `hermes-e2e` image for our e2e tests we can undo this change but I would try to stay as close to upstream as possible.

Regarding the other point. `osmolabs/hermes` it's just a mirror of the official `hermes` page. 
We could use the upstream one ofc but it's nice to have control of the artifacts that we use (e.g. they make the repo private / it is inaccessible / etc...). The burden of mantaining the mirror is minimal I think

nikever:
e2e running correctly with the last changes: https://github.com/osmosis-labs/osmosis/runs/5889286653?check_suite_focus=true 

🚀🚀

p0mvn (approved):
:ship:

tests/e2e/e2e_util_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	exec, err := s.dkrPool.Client.CreateExec(docker.CreateExecOptions{
+		Context:      ctx,
+		AttachStdout: true,
+		AttachStderr: true,
+		Container:    s.hermesResource.Container.ID,
+		User:         "root",
+		Cmd: []string{
+			"hermes",
+			"create",
```
adizere, line 27: Note that we plan to change the `create channel` invocation in Hermes. From

`create channel <CHAIN_A> <CHAIN_B> --port-a=transfer --port-b=transfert`

to 

`create channel <CHAIN-A-ID> <CONNECTION-A-ID> --port-a <PORT-ID> --port-b <PORT-ID>`.

In other words, you need to (1) `create connection` and then (2) pass the connection identifier when you invoke `create channel`. This mean you would need to adapt the CI to add an additional CLI command for creating a new connection, as follows:

`create connection s.chainA.id s.chainB.id`.

Assuming the chains are newly-spawned, the connection identifier will be `connection-0` on both chains. Then you would need to invoke create channel as follows:

`create channel s.chainA.id connection-0 --port-a=transfer --port-b=transfer`

adizere: PS. This change was motivated by relayer operators abusing the command (we got complaints that people create many channels in production with varying client security parameters, ref: https://github.com/informalsystems/ibc-rs/issues/1421).

Would be glad to take your requirements into account and do stuff differently if you have suggestions to change this.

ValarDragon: This seems like a great change to me! Thanks for letting us know / good design decision!

adizere: Great to hear it won't affect your setup adversely.

The change will land in Hermes v0.14, slated for end of April.

PR #1219, opened 2022-04-07
fix make test-cover

p0mvn:
Closes: #XXX

## Description

Quick fix for the failing `make-test`. The issue was that there were 4 spaces instead of a tab. 

I thought that another problem is that the CI job does not fail when running make does: https://github.com/osmosis-labs/osmosis/runs/5873348591?check_suite_focus=true

However, that is not a problem because due to spaces, the job was simply not found and makefile returned exit code 0. We just need to be careful and make sure that the makefile is written correctly.

______

For contributor use:


ValarDragon (approved):
Ah nice catch, it was an error locally as well.

This fixed it locally for me as well

PR #1221, opened 2022-04-07
chore: vscode setting updates

alexanderbez:
- JSON prettyify/format
- Add proto formatting settings

______

For contributor use:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,24 +1,28 @@
 {
-    "tabnine.experimentalAutoImports": true,
-    "go.lintTool": "golangci-lint",
-    "go.formatTool": "gofumports",
-    "go.testOnSave": true,
-    "go.useLanguageServer": true,
-    "[go.mod]": {
-        "editor.formatOnSave": true,
-        "editor.codeActionsOnSave": {
-          "source.organizeImports": true
-        }
-    },
-    "[go]": {
-        "editor.snippetSuggestions": "none",
-        "editor.formatOnSave": true,
-        "editor.codeActionsOnSave": {
-          "source.organizeImports": true
-        }
-    },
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):
I feel like this won't work out of the box on Linux systems that don't have clang installed, but I believe this should be fine since vscode shouldn't error if it doesn't exist.

I'll see what happens on my machine without clang

alexanderbez:
Yeah so you need clang-format installed, that is the caveat, but I think it's worth it for consistent proto formatting. WDYT?

ValarDragon:
I'll install it locally (clang was actually on my setup todo list), I'm just hesistant about making it a required step vs a sub-component of some automated commands, since its another thing to add to the complexities of new dev's setting up a working environment. (It also gets fixed in every `make proto-format` call)

Adding it to the default settings.json is the least controversial thing though, since thats a "if its installed, use it in the way we prefer" situation, which is ideal!

.vscode/settings.json
```diff
@@ -1,24 +1,28 @@
...
-        "editor.formatOnSave": true,
-        "editor.codeActionsOnSave": {
-          "source.organizeImports": true
-        }
-    },
-      "gopls": {
+  "tabnine.experimentalAutoImports": true,
+  "go.lintTool": "golangci-lint",
+  "go.formatTool": "gofumports",
+  "go.testOnSave": true,
```
p0mvn, line 5: I'm personally used to working with auto-save and just noticed the following:
image

Can we let everyone decide if they would like this enabled?

alexanderbez: Yes, wanna make a followup PR?

p0mvn: On it

PR #1222, opened 2022-04-07
chore: remove testOnSave vs code setting

p0mvn:
Closes: #XXX

## Description

https://github.com/osmosis-labs/osmosis/pull/1221#discussion_r845570324
______

For contributor use:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,6 @@
   "tabnine.experimentalAutoImports": true,
   "go.lintTool": "golangci-lint",
   "go.formatTool": "gofumports",
-  "go.testOnSave": true,
   "go.useLanguageServer": true,
   "[go.mod]": {
     "editor.formatOnSave": true,


alexanderbez (approved):

ValarDragon:
Why?

p0mvn:
> Why? This is a manual config that AFAICT, most people who I've helped setup vscode really liked,
> 
> Is this too intrusive or smth?

https://github.com/osmosis-labs/osmosis/pull/1221#discussion_r845570324

I noticed that it doesn't work well with auto-save

ValarDragon:
Hrmm, this is unfortunate. So the tradeoff here is picking between:

Option 1: No test on save / auto save config

Noone has to do manual overrides,, new people to the project have to enable test on save manually for better development experience

Option 2: Test on save set

* Better development experience for new folks
* People who want to have auto save, have to override it

Unfortunately, reading through https://code.visualstudio.com/docs/getstarted/settings, I don't see an easy way to override workspace settings :/, seems like its really meant for just linters. (The way to do it seems to me be to doing language specific, user-set setting overrides, which seems like a bit too complex to expect people to do). Given that, I'm more in favor of Option 1, though I do really think option 2 is the right thing in principle =/

What do others think?

alexanderbez:
I'm honestly indifferent. I've never had `testOnSave` set, so I don't know what it's like.

p0mvn:
I'm used to auto-save so I'm biased towards Option 1 where `testOnSave` is optional. However, if you think that it's better for everyone else, I can always override it locally, it's not a problem.

I'm also not sure about how `testOnSave` would work in the environment where we definitely don't want to run the tests on every save. For example, in e2e tests. I think running that concurrently or restarting too often might break the setup. That might be painful for new folks working on something like e2e tests if they don't know about this setting.

alexanderbez:
Yeah I don't see a reason for this to be `true` now that I think of it.

ValarDragon:
SGTM!

PR #1230, opened 2022-04-08
Stableswap CFMM equations

ValarDragon:
Closes: #XXX

## Description

Thus far, all thats been done is transcribing the solved CFMM equation. I manually split out all the parentheticals / divisions, etc. so I could track whats happening. 

Going to try to keep simplifying this, but thought I'd PR it in case people wanted to look at whats happening. This equation is massive :/

It may be worth us just solving the equation by hand, to make the more intermediate terms more easy to abstract into variable names.

______

For contributor use:


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -0,0 +1,73 @@
...
+//		  /(3 (b + y)
+// 			(-27 b^2 x^3 y - 27 b^2 x y^3
+// 				+ sqrt(
+//					(-27 b^2 x^3 y - 27 b^2 x y^3 - 54 b x^3 y^2 - 54 b x y^4 - 27 x^3 y^3 - 27 x y^5)^2
+//					+ 4 (3 b^4 + 12 b^3 y + 18 b^2 y^2 + 12 b y^3 + 3 y^4)^3)
+// 				- 54 b x^3 y^2 - 54 b x y^4 - 27 x^3 y^3 - 27 x y^5)^(1/3))}
+//      + {(b x + x y)/(b + y)}
+// Then notice that the two sqrt terms and surrounding items in the expression are the same.
+// So we replace them with a single term:
+// discriminant = sqrt(
```
ValarDragon, line 36: I don't think this is actually the discriminant, in particular because `a` isn't actually a polynomial in `x` and `y`. Maybe I should just rename this to radical

x/gamm/pool-models/stableswap/amm_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{
+			sdk.NewDec(100),
+			sdk.NewDec(100),
+			sdk.NewDec(1),
+		},
+		// {
+		// 	sdk.NewDec(100000),
+		// 	sdk.NewDec(100000),
+		// 	sdk.NewDec(10000),
+		// },
```
ValarDragon, line 35: This exceeds precision limits, because the equations have terms on the order of `x^6`.

We need to either (or do both of):
* Scale down all inputs to CFMM equations
* Make a new decimal type with a far larger decimal precision bound
* Ensure inputs are small enough, since we can't have these panicking in begin/end block. (Or we make all AMM calls in begin/end block have panic guards)

ValarDragon:
Thoughts on merging this in, and doing further updates against this code? (The CFMM equation is right, and I'd like to parallelize some of the remaining fronts for this)

ValarDragon:
Per call today: To merge after removing accidental contributing.md commit

PR #1231, opened 2022-04-11
Add `BigInt`, `BigDec` to osmomath

mattverse:
## Description
This PR adds BigDec & BigInt to osmomath package. 
sdk.Dec is implemented upon sdk.Int, which has a max bit length of 256 bit. sdk.Dec would then have max bit of 256 + 60 = 316 bits. 

The implementation of BigInt and BigDec changes these constants in the way that BigInt has the max bit length of 512 bit, 
 max BigInt being 2^512-1 = 1.340781e+154, min BigInt = -(2^512-1) = -1.340781e+154

and BigDec having the max bit len of 572(512 + 60). 

The naming of BigInt does bring confusion with big.Int though, open to recommendations for namings for BigInt

______

For contributor use:


alexanderbez (commented):
I don't really know how I feel bringing in the entire dec/int stuff over to osmosis. It's essentially 99% identical apart from the precision bits, right?

I'd be more inclined to either:

1. See if we can use an OSS library to accomplish our goals instead.
2. Modify our SDK fork to support providing custom precision, e.g. construct a decimal with precision bits defined on the type.

ValarDragon:
My main concern with an OSS library for BigDec is that we then have to deal with serialization overheads / longer term maintenance of that. I believe the SDK is working on trying to get a single serialization format for numeric types, and then have multiple decimal backends be deserializable from that. We'd have to maintain such a pairing if we adapted an open source lib like [apd](https://github.com/cockroachdb/apd)

mattverse (commented):
We would also have to add test cases for calculation between sdk.Dec and `BigInt`, but tbd after decision has been made upon how we want to push forward Dec bit

ValarDragon:
Why do we have the `BigInt` file? I think we should just have `BigDec` right, and make all methods truncate to sdk.Int?

mattverse:
sdk.Int is what contains the restriction for 256bit

mconcat:
How do you think about adding bit precision parameter to NewInt/NewDec on SDK fork, and use it for precise computation purpose? So normally 256-bit NewInt will be constructed using `sdk.NewInt(3)`, but for different precision something like `sdk.NewIntBitPrec(3, 512)`.

mattverse:
Yeah that was one of the main points per yesterday's discussion. I guess it's a trade off, one way or the other. With that being said, on the second thought, my thoughts are leaning towards @ValarDragon's opinion on using `BigDec` instead of fixing the original Dec type

fedekunze:
curious why do you need to support a bigger bit len on Osmo

alexanderbez:
> How do you think about adding bit precision parameter to NewInt/NewDec on SDK fork, and use it for precise computation purpose? So normally 256-bit NewInt will be constructed using `sdk.NewInt(3)`, but for different precision something like `sdk.NewIntBitPrec(3, 512)`.
> 
> (and make type alias for the 512-bit precision Int/Dec)

This was my proposal on the call.

alexanderbez:
I'm not sure I follow? I think it's pretty straightforward actually. All you do is move the private bit variable internally into the `Dec` type directly. No APIs change. You only introduce a few new APIs and change a bit of the logic where the bit variable is used.

mattverse:
Yeah your right. Changing Dec struct in sdk side looks like a more sustainable solution in software development pov.

I'm down to closing this PR and making changes in the sdk side, I'm also curious what @ValarDragon  thinks

alexanderbez:
As long as it doesn't change the marshaling, which it doesn't AFAIK (but we should confirm via a simple unit test), then it should be pretty straight forward. But yeah I'm curious to see what @ValarDragon thinks.

ValarDragon:
I'm not really in favor of doing a max precision field in the type itself. Its not how any other fixed precision integer libraries I've worked with before act, and I don't think its the right long term.

One thing thats common is templating / using a generic for fixed word sizes

mconcat:
We can do something like
```go
type Int[T BitSize] struct {
  *big.Int
}

type BitSize interface {
  BitSize() int
}

type BitSize256 struct {}
func (BitSize256) BitSize() int { return 256 }

type BitSize512 struct {}
func (BitSize512) BitSize() int { return 512 }

// inside of sdk.Int usage...
func (i Int[T]) SomeFunc() {
  bitsize := T.BitSize()
}
```

ValarDragon:
Yeah, I think thats the right generalized approach. Thats similar to what is done in Arkworks / other algebra libraries I'm familiar with. (https://github.com/arkworks-rs/algebra/blob/master/ff/src/biginteger/mod.rs#L25-L26) (If you look into the macros that implement these, are some optimizations for different word sizes / loop unrolling, but we likely won't need to chase speeds that fast)

I think we can leave that approach to much future work though.

alexanderbez:
I like the generics approach @mattverse! We should make those changes in the SDK though :)

ValarDragon (approved):
I recommend we go ahead and merge this, and then do further error analysis for bounds on this in stableswap review. (And potentially deleting bigint later)

Issue #944, opened 2022-02-22
x/lockup: Move messageServer logic for LockTokens into keeper method


ValarDragon:
Most of the logic for the lockup module's validations is in the message server, not the keeper.

However the keeper methods are where it should be, as its API safety that should be guaranteed for everything using the lockup module.

voluntatis-summ:
were you thinking of moving it to the x/lockup/keeper.go file?

stackman27:
Hi @ValarDragon, just making sure, the goal of this issue is to move [this](https://github.com/osmosis-labs/osmosis/blob/20cfa05c105c0fabef1150ce3d0687bbebb99d32/x/lockup/keeper/msg_server.go#L26-L53) logic to [this](https://github.com/osmosis-labs/osmosis/blob/20cfa05c105c0fabef1150ce3d0687bbebb99d32/x/lockup/keeper/lock.go#L118) function right?  What other validation checks are there?

PR #1233, opened 2022-04-12
x/lockup: Move messageServer logic for LockTokens into keeper method

stackman27:
Closes: #944

## Description
- Move `LockTokens` logic to its keeper method 
- I just want to make sure my approach is right before I modify test files 
- Also I'm wondering where I can find other validation logic for x/lockup
______

For contributor use:

- [x]  Move validation logic
- [x]  Move `LockTokens` logic
- [x]  Fix test cases


stackman27:
Fixed test cases. Ready for review!

x/lockup/keeper/lock.go
```diff
@@ -116,8 +117,18 @@ func (k Keeper) removeTokensFromLock(ctx sdk.Context, lock *types.PeriodLock, co
 
 // AddTokensToLock locks more tokens into a lockup
 // This also saves the lock to the store.
-func (k Keeper) AddTokensToLockByID(ctx sdk.Context, lockID uint64, coins sdk.Coins) (*types.PeriodLock, error) {
+func (k Keeper) AddTokensToLockByID(ctx sdk.Context, lockID uint64, owner sdk.AccAddress, coins sdk.Coins) (*types.PeriodLock, error) {
 	lock, err := k.GetLockByID(ctx, lockID)
+	
+	if lock.GetCoins().Len() != 1 {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
```
alexanderbez, line 124: This error has nothing to do with `err`. You should instead pass a string that alludes to the context of why we're erroring.

x/lockup/keeper/msg_server_test.go
```diff
@@ -43,16 +43,16 @@ func (suite *KeeperTestSuite) TestMsgLockTokens() {
...
-			name: "try creating multiple coin lockup",
-			param: param{
-				coinsToLock:         sdk.Coins{sdk.NewInt64Coin("stake", 10), sdk.NewInt64Coin("take", 10)}, // setup wallet
-				lockOwner:           sdk.AccAddress([]byte("addr1---------------")),                         // setup wallet
-				duration:            time.Second,
-				coinsInOwnerAddress: sdk.Coins{sdk.NewInt64Coin("stake", 10), sdk.NewInt64Coin("take", 10)},
-			},
-			expectPass: false,
-		},
+		// {
```
alexanderbez, line 46: Personally, I don't feel too comfortable having commented out code, especially test cases. Can we either remove this or un-comment it if it's useful?

x/lockup/types/msgs.go
```diff
@@ -31,6 +31,12 @@ func (m MsgLockTokens) ValidateBasic() error {
 	if m.Duration <= 0 {
 		return fmt.Errorf("duration should be positive: %d < 0", m.Duration)
 	}
+
+	// we only allow locks with one denom for now
+	if m.Coins.Len() != 1 {
+		return fmt.Errorf("Lockups can only have one denom per lockID, got %v", m.Coins)
```
alexanderbez, line 37:
```suggestion
		return fmt.Errorf("lockups can only have one denom per lock ID; got %v", m.Coins.Len())
```

x/lockup/keeper/msg_server_test.go
stackman27: I had to comment out the test case because i moved the check to `ValidateBasic()` in `msgs.go`, therefore it wasn't being tested when the function `LockTokens(...)` was called

x/lockup/keeper/lock.go
stackman27: I am a bit confused as to why this logic check is even required when we already check for coins length during validation check

alexanderbez: Maybe it isn't? :p

@ValarDragon wdyt?

x/lockup/keeper/msg_server_test.go
alexanderbez: So then let's remove it 👍

x/lockup/keeper/lock.go
ValarDragon: Yeah, sure. If we want tor remove the check, imo we should:

- [ ] Make the function private (rename to `addTokensToLockByID`)
- [ ] Change the function argument from `coins sdk.Coins` to `coin sdk.Coin`

ValarDragon:
Sorry for the delay in reply. I think the way I'd recommend architecting this, is we have a single point of entry function `LockTokens`. That function does what the message did before. 

Basically determining that:

- [ ] The owner is correct
- [ ] If there is a prior lock, call add tokens to lock
- [ ] If there is no prior lock, create a new lock. (This would involve renaming the existing ``LockTokens` function into a new function, perhaps called `CreateLock`)

x/lockup/keeper/lock.go
stackman27: how does making function private work? would it just be replacing `AddTokensToLockByID` to `addTokensToLockByID`?

Also if I change the function argument to `coin sdk.Coin` then i'm gonna have to change the rest of the functions that uses expects `coins sdk.Coins` like;` k.bk.SendCoinsFromAccountToModule(ctx, lock.OwnerAddress(), types.ModuleName, coins)` and `k.addTokensToLock(ctx, lock, coins)`, is that expected?

stackman27:
@ValarDragon @alexanderbez Made all the suggested changes!

x/lockup/keeper/lock.go
ValarDragon: and thats how making it private works as well!

stackman27: @ValarDragon tried to make `AddTokensToLockByID(...)` private but got this issue. What else do i need to refactor to accommodate  testing suite? 

Screen Shot 2022-05-03 at 4 22 03 PM

x/superfluid/keeper/stake_test.go
```diff
@@ -814,6 +815,7 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
 				}
 
 				// calculating the estimated delegation amount for multiplier change
+				fmt.Println(intermediaryDels)
```
mattverse, line 818: reminder to remove fmts in tests!

mattverse (changes requested):
I think the major goal of the issue was to move the logical gate of determining if it's an existing lock to a separate keeper method, whilst this PR only moves the basic checks that were done in `msg_server.go` to keeper.

I do think that this PR itself does contain useful changes that could be merged to main, but I do think that the main goal of the #https://github.com/osmosis-labs/osmosis/issues/944 would remain unsolved and would have to create a seperate PR for it later on.

x/lockup/keeper/msg_server.go
```diff
@@ -27,31 +27,16 @@ var _ types.MsgServer = msgServer{}
...
-		// if existing lock with same duration and denom exists, just add there
 		if len(locks) > 0 {
 			lock := locks[0]
-			if lock.Coins.Len() != 1 {
-				return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
-			}
-
-			if lock.Owner != owner.String() {
-				return nil, types.ErrNotLockOwner
-			}
```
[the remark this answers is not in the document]
stackman27: This check was moved [here](https://github.com/osmosis-labs/osmosis/pull/1233/files#diff-70d5ac756cc9fd42fda2ad97bd9c22a142e5deeb6e0a1b118784aac59030b332R132)

stackman27:
> I think the major goal of the issue was to move the logical gate of determining if it's an existing lock to a separate keeper method, whilst this PR only moves the basic checks that were done in `msg_server.go` to keeper.
> 
> I do think that this PR itself does contain useful changes that could be merged to main, but I do think that the main goal of the ##944 would remain unsolved and would have to create a seperate PR for it later on.

Thank you @mattverse for the insightful review. I was a bit confused whether to move the entire "existing lock" logic or just simple validation checks, but i've moved the entire logic as per your suggestion. Should be good now :)

x/lockup/types/msgs.go
alexanderbez, line 37:
```suggestion
		return fmt.Errorf("lockups can only have one denom per lock ID, got %v", m.Coins)
```

x/lockup/keeper/lock.go
```diff
@@ -114,18 +111,36 @@ func (k Keeper) removeTokensFromLock(ctx sdk.Context, lock *types.PeriodLock, co
 	return nil
 }
 
+func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) ([]types.PeriodLock, error) {
+	locks := k.GetAccountLockedDurationNotUnlockingOnly(ctx, owner, coin.Denom, duration)
+	// if existing lock with same duration and denom exists, just add there
+	if len(locks) > 0 {
+		lock := locks[0]
```
alexanderbez, line 118: Is this safe to always add to the first one?

stackman27: I believe so, I didn't change any logic from existing code. And locks usually are in this format: `sdk.Coins{sdk.NewInt64Coin("stake", 5)`

x/lockup/keeper/msg_server.go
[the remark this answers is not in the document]
mattverse: my bad must have missed it! Thanks for the asnwer!

mattverse (approved):
LGTM! 

Thanks for all the updates on the feedbacks!

alexanderbez (approved):

Issue #695, opened 2021-12-28
Rocksdb transition


faddat:
## Summary

The `whisky on the rocks` series of patches powers nearly every ibc relay operator's nodes on busy networks like Gaia, Terra, Osmosis and Juno.

But drunk nodes (nodes with incredibly high iavl and store caches) consume system resources the way I consume pizza after drinking: 40gb or so.  I would recommend to any relayer that they do indeed get their node very drunk.  

A sober node running rocksdb still does much better:

* at epoch
* while relaying
* After state sync on a raspberry pi
 
## Problem Definition

We should move sober Osmosis to a rocksdb by default build (binaries we distribute should use `-tags rocksdb` by default and release process so that Osmosis:

* syncs about 3x faster
	* save operator time
	* save operator money
	* grow network of operators due to lower time burden

## Proposal

I'm proposing that we use the following replace line:

https://github.com/notional-labs/osmosis/blob/54c0b7e2c1dd01dd46e7430c155aba9f78db6303/go.mod#L126

in go.mod until either:

* tecbot accepts a pr fixing his gorocksdb lib

or

* we change the module path for the cosmos gorocksdb lib and update tm-db so that it imports the cosmos gorocksdb lib.

The performance improvements to many aspects of operating gaia are quite incredible with this one little fix.  

In the next 2-4 weeks, Notional should have a quicksync.io style server set up at an unlimited 10gbps provider in the united states.  If anyone in the community is aware of hosting centers with true 10gbps connections in asia, preferably Singapore or Hong Kong, we'd glady provide direct asia coverage, too. 

This service will provide:

* Archive
* Default
* Fully Pruned

snapshots and should make it much simpler for operators to use rocksdb.  Since rocksdb also helps with read performance, analytics on Gaia can be improved, too. 

Because the improvements in performance are so dramatic across the board (observationally speaking) I am also suggesting that profiling and synthetic tests can mislead us in this case-- but are **surely** a part of this transition.  

I have largely copy/pasted this issue from one on Gaia, and I am heading over to Terra to do the same thing.  The issues are not at all dissimilar. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

faddat:
[...]

PR #1234, opened 2022-04-12
Update state sync scripts

faddat:
## Description

NB: Previously, osmosis would apply all snapshots, then crash.

Currently, osmosis applies all snapshots, and stays live.  Because it stays live, these scripts don't work correctly.

I have a machine doing a rocksdb sync now, and hope to prune that but nothing makes tiny snapshots like statesync.

Pursuant to #695 
______

For contributor use:


ValarDragon (approved):
Thanks for the script fixes!

Issue #164, opened 2021-05-07
x/lockup: Enable users to elongate the lockup period for locked tokens


ValarDragon:
Suppose a user has locked up 10 tokens for 2 days. We should allow them to increase the lockup period to 1 week, without having to unlock and relock. This will probably require a new message type in x/lockup to do this, as we need the user to specify which initial lockup time duration they want to increase.

(E.g. if they have 10 tokens locked for 2 days, 20 tokens locked for 4 days, and want to move the 20 tokens to a lockup of 1 week)

Also this should only support increasing the lockup duration, not decreasing it

ValarDragon:
Prelaunch we should just ensure that this will be easy to upgrade in the future / double check how we handle state / implicit assumptions around this.

mconcat:
Adding `MsgEditLockup` would work and we can check that lockup duration change is positive in `ValidateBasic`. This would be extensible in the future for other modifications too.

PR #1253, opened 2022-04-14
Add lockup duration edit method

mconcat:
## Description

Closes: #164

______

For contributor use:


ValarDragon:
Ah sweet! Lets add a hook for this?

Also thoughts on renaming `EditLockup` to `ExtendLockup`?

mconcat:
> Ah sweet! Lets add a hook for this?
> 
> Also thoughts on renaming `EditLockup` to `ExtendLockup`?

I was thinking about using this method for editing other part of the lockups in the future(e.g. cancelling unlock) so we can reduce the total number of messages. If you think they should be in different message, I agree ExtendLockup is a better name!

ValarDragon:
I think they should be different messages!

(It is an interesting API question if CancelUnlock should be expected behavior of extend lockup on an unbonding lockup, but I think thats a future question)

proto/osmosis/lockup/tx.proto
```diff
@@ -49,3 +51,20 @@ message MsgBeginUnlocking {
   ];
 }
 message MsgBeginUnlockingResponse { bool success = 1; }
+
+message MsgEditLockup {
```
alexanderbez, line 55: nit: protodoc

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,40 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
 	k.hooks.OnTokenUnlocked(ctx, owner, lock.ID, lock.Coins, lock.Duration, lock.EndTime)
 	return nil
 }
+
+func (k Keeper) EditLockup(ctx sdk.Context, lock types.PeriodLock, newDuration time.Duration) error {
+	// sanity check
```
alexanderbez, line 542:
```suggestion
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,40 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+
+	// check unlocking
```
alexanderbez, line 544:
```suggestion
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,40 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+		return fmt.Errorf("cannot edit unlocking lockup")
+	}
+
+	// check synthetic lockup exists
+	if k.HasAnySyntheticLockups(ctx, lock.ID) {
+		return fmt.Errorf("cannot edit lockup with synthetic lock")
+	}
+
+	if newDuration != 0 {
+		// check newDuration > duration
```
alexanderbez, line 555:
```suggestion
```

x/lockup/keeper/msg_server.go
```diff
@@ -155,3 +155,32 @@ func createBeginUnlockEvent(lock *types.PeriodLock) sdk.Event {
...
+func (server msgServer) EditLockup(goCtx context.Context, msg *types.MsgEditLockup) (*types.MsgEditLockupResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	lock, err := server.keeper.GetLockByID(ctx, msg.ID)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+
+	if msg.Owner != lock.Owner {
+		return nil, sdkerrors.Wrap(types.ErrNotLockOwner, fmt.Sprintf("msg sender (%s) and lock owner (%s) does not match", msg.Owner, lock.Owner))
```
alexanderbez, line 168: Nit: Use `Wrapf` instead of `Sprintf`

x/lockup/types/msgs.go
```diff
@@ -92,3 +93,27 @@ func (m MsgBeginUnlocking) GetSigners() []sdk.AccAddress {
...
+		Owner:    owner.String(),
+		ID:       id,
+		Duration: duration,
+	}
+}
+
+func (m MsgEditLockup) Route() string { return RouterKey }
+func (m MsgEditLockup) Type() string  { return TypeMsgEditLockup }
+func (m MsgEditLockup) ValidateBasic() error {
+	return nil
```
alexanderbez, line 109: There's zero validation we want to do here? What about checking the fields for basic bounds?

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,40 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+		for _, coin := range lock.Coins {
+			k.accumulationStore(ctx, coin.Denom).Decrease(accumulationKey(lock.Duration), coin.Amount)
+			k.accumulationStore(ctx, coin.Denom).Increase(accumulationKey(newDuration), coin.Amount)
+		}
+
+		lock.Duration = newDuration
+	}
+
+	// update lockup
+	err := k.setLockAndResetLockRefs(ctx, lock)
```
mattverse, line 570: Do we not need to run `deleteLockRefs` to delete old lock refs? I'm not quite sure if `setLockAndResetLockRefs` is actually deleting the old refs

x/lockup/types/msgs.go
mconcat: Yup seems like adding nil/empty checks will be needed!

x/lockup/keeper/lock.go
mconcat: Thanks for catching it! Will fix this

mconcat: Matt pointed out that `setLockAndResetLockRefs` is bit misleading, because the name and the comment
```
// setLockAndResetLockRefs sets the lock, and resets all of its lock references
// This puts the lock into a 'clean' state, aside from the AccumulationStore.
```
implies that it removes the existing lockrefs, but actually it doesn't. Should we fix this behaviour, or change the name of the method?

mattverse: I think we wouldn't need to fix this within this PR, but instead create an issue to investigate on this.

That being said, I think it would suffice for this PR to ensure that old lock refs are getting deleted using `deleteLockRefs`

mconcat: I have replaced setLockAnd.. to a pair of delete and add lockrefs. This works!

ValarDragon: Oh interesting, lets make an issue and investigate this

mattverse: https://github.com/osmosis-labs/osmosis/issues/1275

x/lockup/types/msgs.go
alexanderbez: What about making sure the owner is empty?

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,42 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
 	k.hooks.OnTokenUnlocked(ctx, owner, lock.ID, lock.Coins, lock.Duration, lock.EndTime)
 	return nil
 }
+
+func (k Keeper) ExtendLockup(ctx sdk.Context, lock types.PeriodLock, newDuration time.Duration) error {
+	if lock.IsUnlocking() {
+		return fmt.Errorf("cannot edit unlocking lockup")
```
alexanderbez, line 543:
```suggestion
		return fmt.Errorf("cannot edit unlocking lockup for lock %d", lock.ID)
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,42 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+	}
+
+	// check synthetic lockup exists
+	if k.HasAnySyntheticLockups(ctx, lock.ID) {
+		return fmt.Errorf("cannot edit lockup with synthetic lock")
```
alexanderbez, line 548:
```suggestion
		return fmt.Errorf("cannot edit lockup with synthetic lock %d", lock.ID)
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,42 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+	}
+
+	oldLock := lock
+
+	if newDuration != 0 {
+		fmt.Println(newDuration, lock.Duration)
```
alexanderbez, line 554:
```suggestion
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,42 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+		if !(newDuration > lock.Duration) {
```
alexanderbez, line 555:
```suggestion
		if newDuration < lock.Duration {
```

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,42 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+	// update lockup
+	err := k.deleteLockRefs(ctx, unlockingPrefix(oldLock.IsUnlocking()), oldLock)
+	if err != nil {
+		return err
+	}
+	err = k.addLockRefs(ctx, lock)
+	if err != nil {
+		return err
+	}
+	return k.setLock(ctx, lock)
```
alexanderbez, line 577:
```suggestion
	// update lockup
	err := k.deleteLockRefs(ctx, unlockingPrefix(oldLock.IsUnlocking()), oldLock)
	if err != nil {
		return err
	}

	err = k.addLockRefs(ctx, lock)
	if err != nil {
		return err
	}

	return k.setLock(ctx, lock)
```

stackman27:
Closes [#164](https://github.com/osmosis-labs/osmosis/issues/164) ?

alexanderbez (approved):

alexanderbez (approved):

alexanderbez:
@mattverse wanna give this another look?

x/lockup/keeper/msg_server.go
```diff
@@ -155,3 +155,32 @@ func createBeginUnlockEvent(lock *types.PeriodLock) sdk.Event {
...
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.TypeEvtLockTokens,
+			sdk.NewAttribute(types.AttributePeriodLockID, utils.Uint64ToString(lock.ID)),
+			sdk.NewAttribute(types.AttributePeriodLockOwner, lock.Owner),
+			sdk.NewAttribute(types.AttributePeriodLockDuration, lock.Duration.String()),
+		),
+	})
+
+	return &types.MsgExtendLockupResponse{}, nil
```
mattverse, line 185: We should either fill in the response field or delete it in the proto file if it's unused

x/lockup/keeper/lock_test.go
```diff
@@ -640,3 +640,63 @@ func (suite *KeeperTestSuite) TestSlashTokensFromLockByID() {
...
+	suite.Require().Error(err)
+
+	// duration increase should success
+	err = suite.App.LockupKeeper.ExtendLockup(suite.Ctx, *lock, time.Second*2)
+	suite.Require().NoError(err)
+
+	// check queries
+	lock, _ = suite.App.LockupKeeper.GetLockByID(suite.Ctx, lock.ID)
+	suite.Require().Equal(lock.Duration, time.Second*2)
+
```
mattverse, line 679: Can we add basic tests to ensure other fields of lock itself is preserved after `ExtendLockup`?
```suggestion
	suite.Require().Equal(1, lock.ID)
	suite.Require().Equal(coins, lock.Coins)
	
```

x/lockup/keeper/lock_test.go
```diff
@@ -640,3 +640,63 @@ func (suite *KeeperTestSuite) TestSlashTokensFromLockByID() {
...
+		Duration: time.Second,
+	})
+	suite.Require().Equal(int64(10), acc.Int64())
+
+	lock, _ := suite.App.LockupKeeper.GetLockByID(suite.Ctx, 1)
+
+	// duration decrease should fail
+	err = suite.App.LockupKeeper.ExtendLockup(suite.Ctx, *lock, time.Second/2)
+	suite.Require().Error(err)
+
```
mattverse, line 671: Could we also create test cases to err if the lock has same duration as the original lock?
```suggestion
	// extending lock with same duration should fail
	err = suite.App.LockupKeeper.ExtendLockup(suite.Ctx, *lock, time.Second)
	suite.Require().Error(err)
	
```

mattverse (commented):
Don't want to block this PR for so long being picky 😅 

Went through final review, should be good to merge after fix!

mattverse (approved):
🌮

proto/osmosis/lockup/tx.proto
```diff
@@ -49,3 +51,22 @@ message MsgBeginUnlocking {
...
+    (gogoproto.nullable) = false,
+    (gogoproto.stdduration) = true,
+    (gogoproto.jsontag) = "duration,omitempty",
+    (gogoproto.moretags) = "yaml:\"duration\""
+  ];
+
+  // extend for other edit, e.g. cancel unlocking
+}
+
+message MsgExtendLockupResponse { bool success = 1; }
```
ValarDragon, line 72: Should we remove this response field? (Can be done in a second PR)

Not sure what this would communicate relative to the general txs success / failure

x/lockup/keeper/lock.go
```diff
@@ -536,3 +537,51 @@ func (k Keeper) unlockInternalLogic(ctx sdk.Context, lock types.PeriodLock) erro
...
+		return err
+	}
+
+	err = k.setLock(ctx, lock)
+	if err != nil {
+		return err
+	}
+
+	k.hooks.OnTokenUnlocked(ctx, lock.OwnerAddress(), lock.ID, lock.Coins, oldLock.Duration, lock.EndTime)
+	k.hooks.OnTokenLocked(ctx, lock.OwnerAddress(), lock.ID, lock.Coins, lock.Duration, lock.EndTime)
```
ValarDragon, line 584: Lets re-assess these hook calls in a follow-up issue

x/lockup/keeper/msg_server_test.go
```diff
@@ -261,3 +262,78 @@ func (suite *KeeperTestSuite) TestMsgBeginUnlockingAll() {
 		}
 	}
 }
+
+func (suite *KeeperTestSuite) TestMsgEditLockup() {
```
ValarDragon, line 266: Great job on this test!

ValarDragon (approved):
Nice work on this!

Issue #1196, opened 2022-04-05
x/gamm: JoinSwapExactAmountIn/ExitSwapShareAmountIn Share Inconsistency Investigation


alexanderbez:
The `TestJoinSwapExactAmountInConsistency` test has a case there is an expected number of resulting "out" amount number of tokens returned from `ExitSwapShareAmountIn` after calling `JoinSwapExactAmountIn`. The current theory is the swap fee is being applied incorrectly and that the no-swap variants of these calls should be tried out to test this theory and further understand why this is happening.

ref: https://github.com/osmosis-labs/osmosis/pull/1174

cc @ValarDragon

p0mvn:
I will try to investigate this. Should help me understand gamm better

PR #1254, opened 2022-04-14
chore: inverse unit tests for CalcInAmtGivenOut and CalcOutAmtGivenIn with zero swap fee

p0mvn:
Closes: #XXX

## Description

Trying to investigate https://github.com/osmosis-labs/osmosis/issues/1196 by incrementally adding more unit tests

These tests do not allow us to conclude anything with regards to #1196 yet. The tests added so far have a swap fee of 0. I'm planning to continue adding unit tests with a non-zero swap fee. Also, test other functionality and methods that are being called in `ExitSwapShareAmountIn` and `JoinSwapExactAmountIn`

______

For contributor use:


alexanderbez (commented):
So what did we find out here? There's a test case commented out in `TestJoinSwapExactAmountInConsistency`.

p0mvn:
> So what did we find out here? There's a test case commented out in `TestJoinSwapExactAmountInConsistency`.

Nothing yet, I'm planning to continue increasing tests / coverage for all methods called within `JoinSwapExactAmountIn`.

Currently, it's hard to determine where exactly the error is happening because we are missing unit tests for each of the methods called by`JoinSwapExactAmountIn`. By increasing the number of unit tests for each of the called methods, we should make it more likely to find the problem

x/gamm/pool-models/balancer/util_test.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func createTestContext(t *testing.T) sdk.Context {
+	db := dbm.NewMemDB()
+	logger := log.NewNopLogger()
+
+	ms := rootmulti.NewStore(db, logger)
+
+	return sdk.NewContext(ms, tmtypes.Header{}, false, logger)
+}
```
mattverse, line 42: nit: we can also add these to app/app_testing/test_suite.go!

mattverse (commented):
Hrmm interesting. So I think these test cases tell us that running `CalcInAmtGivenOut` and `CalcOutAmtGiven` straight away returns the correct value we expect, and that the problem is coming from other steps?

p0mvn:
> Hrmm interesting. So I think these test cases tell us that running `CalcInAmtGivenOut` and `CalcOutAmtGiven` straight away returns the correct value we expect, and that the problem is coming from other steps?

It's unclear yet. I have tested with 0 swap fees so far. I'm planning to add the tests with a non-zero swap fee in a separate PR

mattverse:
Sounds great! Can we address that this is a test for zero fee either in the comment or in the test case name?

alexanderbez:
Can we leave this PR in Draft mode to indicate that it's still being worked on?

x/gamm/pool-models/balancer/util_test.go
p0mvn: The files in `test_suite.go` seem to belong to the keeper test suite.

On the other hand, these are low-level balancer pool test utilities. I briefly looked at the `test_uite.go`, and I don't think there are any applications for these utilities there. So I think we might want to keep them here. Let me know if you agree

p0mvn:
> Sounds great! Can we address that this is a test for zero fee either in the comment or in the test case name?

Done! I also created an issue to continue with the tests with a non-zero swap fee: https://github.com/osmosis-labs/osmosis/issues/1288

@alexanderbez this is ready for review. I'm trying to work on this in small parts / PRs to make it easier to review

x/gamm/pool-models/balancer/util_test.go
mattverse: oh yeah agree! Forgot that we're not using test suite while I was writing that suggestion!

alexanderbez (approved):
🌮

ValarDragon (approved):

Issue #1261, opened 2022-04-16
enrich existing PR template to smooth the review process


jlbit:
As discussed with @ValarDragon, currently Osmosis's existing [pr template](https://github.com/osmosis-labs/osmosis/blob/main/.github/pull_request_template.md) is a bit over simplified with only "Description" field required.

I'd like propose adding a bit more templating to help smooth the review process, as well as fostering a engineering habit of emphasizing documentations, e.g. when a new feature released or user-facing behavior changed, PR submitters should always keep in mind updating the docs at the same time in https://github.com/osmosis-labs/docs.

jblbit:
on the other hand, the template should not be overcomplicated like the one in cosmos SDK, which has too many checkboxes that makes it unreadable and thus no one taking it seriously anymore

jblbit:
this should be a very similar one to which I proposed in osmosis's doc repo https://github.com/osmosis-labs/docs/pull/68

PR #1264, opened 2022-04-16
enrich existing PR template to smooth the review process

jlbit:
Closes: https://github.com/osmosis-labs/osmosis/issues/1261

NOTE! I‘m using the proposed PR template for this PR, to demonstrate how it can work

Adding a bit more templating to help smooth the review process, as well as fostering a engineering habit of emphasizing documentations and tests.

## Brief change log

- updated PR template with more critical self-check aspects for PR submitters, while trying to keep it simple and lean

## PR Formatting

- Is PR targeted against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))  (yes)

## Testing and Verifying

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,18 +7,46 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
 
 Closes: #XXX
 
-## Description
+## What is the purpose of the change
 
-<!-- Add a description of the changes that this PR introduces and the files that
-are the most critical to review.
--->
+*Add a description of the overall background and high level changes that this PR introduces *
 
-______
+*(E.g.: This pull request improves documation of area A by adding ....*
 
-For contributor use:
 
-- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
-- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
-- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
-- [ ] Re-reviewed `Files changed` in the Github PR explorer
+## Brief change log
 
+*(for example:)*
+ 
+  - *The metadata is stored in the blob store on job creation time as a persistent artifact*
+  - *Deployments RPC transmits only the blob storage reference*
+  - *Daemons retrieve the RPC data from the blob cache*
+
+
+## Testing and Verifying
+
+*(Please pick one of the following options)*
+
+This change is a trivial rework / code cleanup without any test coverage.
+
+*(or)*
+
+This change is already covered by existing tests, such as *(please describe tests)*.
+
+*(or)*
+
+This change added tests and can be verified as follows:
+
+*(example:)*
+  - *Added unit test that validates ...*
+  - *Added integration tests for end-to-end deployment with ...*
+  - *Extended integration test for ...*
+  - *Manually verified the change by ...*
+
+## Documentation and Release Note
+
+  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
+  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
+  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)
+
+ 


.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
...
+ 
+  - *The metadata is stored in the blob store on job creation time as a persistent artifact*
+  - *Deployments RPC transmits only the blob storage reference*
+  - *Daemons retrieve the RPC data from the blob cache*
+
+
+## PR Formatting
+
+
+- Is PR targeted against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))  (yes / no)
```
p0mvn, line 29: instead of (yes/now) can we make this:
- [ ] Is PR targeted against correct branch?

So that the author can quickly toggle the desired checkboxes

.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
 
 Closes: #XXX
 
-## Description
+## What is the purpose of the change
 
-<!-- Add a description of the changes that this PR introduces and the files that
-are the most critical to review.
--->
+*Add a description of the overall background and high level changes that this PR introduces *
```
p0mvn, line 12: Can we keep these as comments? I personally would prefer to avoid having to delete these lines on every PR. With comments, the PR author can keep them and just start typing below

.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
...
 
-For contributor use:
 
-- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
-- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
-- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
-- [ ] Re-reviewed `Files changed` in the Github PR explorer
+## Brief change log
 
+*(for example:)*
```
p0mvn, line 19: Similar suggestion about making these comments

.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
...
+## PR Formatting
+
+
+- Is PR targeted against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))  (yes / no)
+
+
+
+## Testing and Verifying
+
+*(Please pick one of the following options)*
```
p0mvn, line 35: suggest making comments where applicable

.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
...
+
+This change is a trivial rework / code cleanup without any test coverage.
+
+*(or)*
+
+This change is already covered by existing tests, such as *(please describe tests)*.
+
+*(or)*
+
+This change added tests and can be verified as follows:
```
p0mvn, line 45:
```suggestion
- [ ] This change is a trivial rework/code cleanup without any test coverage.
- [ ] This change is already covered by existing tests, please provide examples of the test below:
   * < type here >
- [ ] This change added tests and can be verified as follows:
   * < type here >
```

.github/pull_request_template.md
```diff
@@ -7,18 +7,53 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
...
+
+*(example:)*
+  - *Added unit test that validates ...*
+  - *Added integration tests for end-to-end deployment with ...*
+  - *Extended integration test for ...*
+  - *Manually verified the change by ...*
+
+## Documentation and Release Note
+
+  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
```
p0mvn, line 55: Please consider making these as checkboxes for convenience

p0mvn (commented):
Thanks for this change and for putting so much thought into it. I added some minor suggestions, please take a look

.github/pull_request_template.md
jblbit: Hi Roman, the problem is, those comments won't show up when a PR is created, so submitters won't be able to see it at all. E.g. in https://github.com/osmosis-labs/osmosis/pull/1233, it just doesn't show up. 

Understand it takes a small effort to delete lines. But missing guidance can be quite confusing to new contributors like me who may not know what it's about.

Overall, hope it can offer a better practice to ask everyone, both existing and new contributors, to put more thoughts into PR formatting and content, that will eventually make us move faster as a group

jblbit (commented):
Hi @p0mvn , thanks for reviewing and suggesting ideas!

re: checkbox, background on why I'd prefer *trying* not having checkbox at least first is:
1. as mentioned in https://github.com/osmosis-labs/osmosis/issues/1261, one goal is to avoid the excessive checkboxes in template like cosmosSDK's, which is unreadable and resulting in people just completely ignore them, like [1](https://github.com/cosmos/cosmos-sdk/pull/11652) or [2](https://github.com/cosmos/cosmos-sdk/pull/11609)
2. this template proposal is mostly borrowed from Apache foundation projects where I've been heavily involved for the past few years, and it has been proven works well at scale of  thousand of contributors.

All being said, we can always see how it goes then iterate over it, and this should not be a final version we settle on. More than happy to iterate based on feedback, but highly encourage us to give the proposed format a try first and further polish it down the road.

.github/pull_request_template.md
p0mvn: > those comments won't show up when a PR is created, so submitters won't be able to see it at all

Could you help me understand - what exactly do you mean by not being able to see them as a submitter?

Here's an attempt to create a test PR where the comments are visible:
image

> But missing guidance can be quite confusing to new contributors like me who may not know what it's about.

Although I agree that these descriptions are valuable, I think they would be better kept as comments. The fundamental purpose of this text is for submitters to read them and understand what needs to be done. In my opinion, comments help to achieve that. In addition, I foresee future PRs not having these descriptions removed, making them less readable and more cluttered with text. With comments, the reviewers wouldn't see this detailed instruction text

p0mvn:
Appreciate all the clarifications @jblbit , I saw your example PR descriptions such as #1268 which looks great. I see the value in having to encourage everyone to put more thought into the descriptions by having fewer checkboxes. Open to trying that first

> to avoid the excessive checkboxes in template like cosmosSDK's, which is unreadable and resulting in people just completely ignore them

I think the major issue with people not utilizing the template in the SDK is stemming from the fact that the team is not encouraging its use. If we want to encourage all contributors to use our template, we should ultimately start reminding each other to fill in the description in every PR and not ignore it

mattverse (commented):
This is awesome! Yeah I agree that cosmos sdk's PR template and our current template with all the check box is easy to ignore.

However, I don't know if I quite agree that we should change it to the way an user answers yes or no to a question as this PR proposes (e.g `- Is PR targeted against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))  (yes / no)`)

I don't have strong opinions for this though and would follow along with what others think, but my suggested change would be keeping this sorting using the headers(e.g) what is the purpose of the change, brief change log etc) and then keeping a checkbox underneath it.

.github/pull_request_template.md
ValarDragon: idk, I think they're better off visible as well, you'd want them to show when you hit preview for instance.

I also don't think this is a big deal to change once we try something initially. Perhaps we start with it visible, and if we observe its not causing the effect we want, we make them comments in a future PR?

ValarDragon:
I think part of the problem with checkboxes, is they don't work too well in the PR creation template. (You have to manually type x, or do a thing after hitting submit)

.github/pull_request_template.md
p0mvn: Discussed offline - let's make it visible and try :+1:

.github/pull_request_template.md
alexanderbez, line 55: Can this entire section be converted to a checklist? Similar to what we do for the SDK.

.github/pull_request_template.md
alexanderbez, line 29: This should also be added to the checklist below.

alexanderbez (commented):
Love the motivation for this! I'm a bit hesitant because it makes the template more verbose for the PR author. Can we find a happy medium similar to what we do for the SDK?

Essentially, it's a checklist, one for the PR author and one for the repo maintainers. You can find it here: https://github.com/cosmos/cosmos-sdk/blob/master/.github/PULL_REQUEST_TEMPLATE/other.md

jlbit:
Hi @alexanderbez ! 

> Love the motivation for this! I'm a bit hesitant because it makes the template more verbose for the PR author. 

Good question. To clarify, it's not aimed to save time for PR author, but for PR reviewers. So please evaluate it from a reviewer angle, e.g. would you be able to review someone else's PR with this template more efficiently and confidently, even with little context? Honestly authors should take the extra work to make it easy to review by others, and don't count on reviewers to check for them.

> Can we find a happy medium similar to what we do for the SDK?

As mentioned [here](https://github.com/osmosis-labs/osmosis/pull/1264#pullrequestreview-943885282), we discussed and wanted to try sth new rather than what cosmosSDK does. I suspect why the PR review process in cosmosSDK is not effective is mostly counting too much on reviewers rather than authors, and also checkbox in practice don't work as most expected.

jlbit:
Hi all, based on feedback, I'm gonna take out the part of "Is PR targeted against correct branch" which is inherited. This may just be a bylaw and no need to check for every PR.

BTW, Roman opened [a PR with this new template](https://github.com/osmosis-labs/osmosis/pull/1286) as a good example. It makes it much easier at least for me to ramp up and review, compared to most other opened PRs.

All being said, understand it's a bit different from what we have so far, but I highly recommend trying it which has been proven at scale. We can always make changes based on real feedback afterwards rather than debating beforehand :)

ValarDragon (approved):

ValarDragon:
I'm down to try this and iterate on it!

I actually like the non-checkbox system, and want to try it out at least. I feel like I've been in projects using checkboxes for PRs for years now, (not just in Cosmos), and none have had great success with getting contributors to follow the checkboxes. (in part, because imo the markdown syntax is confusing in preview)

We may want to cut down on the intro blurb so more of the relevant text automatically fills in, but I think that will become more apparent as we do more PR's and iterate on the templtae

mattverse:
Yeah totally down to at least try it as well!

jlbit:
Hi all, thank you for trying it out！Love to hear your feedback in a few weeks. 

Can you help merge it, or grant me write access to merge it?

btw, in order to see it takes effect, all of us will be leading the change here, as nothing comes for granted :) e.g. we have to be the role model to follow it ourselves; if someone is not following it, please politely point it out and ask them to reformat.

alexanderbez:
I'm highly skeptical PR authors will fill this out in detail, but I suppose it can't hurt to try.

PR #1270, opened 2022-04-17
Speedrun

faddat:
## Speedrun

We guess a bit too much.  To reduce the amount of guessing about Osmosis performance, here's a speedrun script, which will (at least) record how long v3-6 take to sync.

______

For contributor use:

### DIFF
--- a/scripts/archivenode.bash
+++ b/scripts/archivenode.bash
@@ -5,16 +5,33 @@
 # Archive node script
 # NB:  you can also download archives at quicksync:
 # https://quicksync.io/networks/osmosis.html
-# 2nd NB:  may explode but works for me. - Jacob
+# 2nd NB: you can change OSMOSISD_PRUNING=nothing to OSMOSISD_PRUNING=default OR you could also set the pruning settings manually with OSMOSISD_PRUNING=custom
+# 3rd NB: you might want to use this to test different databases, and to do that my recommended technique is like:
+# go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb' -tags rocksdb ./...
+# if you do not use the ldflags thing you won't use the chosen db for everything, so best use it.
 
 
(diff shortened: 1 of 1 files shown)


scripts/archivenode.bash
```diff
@@ -5,16 +5,33 @@
...
 
 
-git checkout v1.0.1
+export OSMOSISD_PRUNING=nothing
+
+# VERSION THREE
+echo "v3 took" > howlong
+git checkout v3.x
+time osmosisd start 
+make install
```
ValarDragon, line 20: How does this work? The prior code that downloaded the genesis file made sense to me

faddat: it wouldn't!

thanks @ValarDragon

scripts/archivenode.bash
```diff
@@ -5,16 +5,36 @@
...
+export OSMOSISD_DB_BACKEND=goleveldb
+
+# VERSION THREE
+echo "v3 took" > howlong
+git checkout v3.x
+osmosisd init speedrun
 wget -O ~/.osmosisd/config/genesis.json https://github.com/osmosis-labs/networks/raw/main/osmosis-1/genesis.json
-osmosisd start --pruning nothing --p2p.seeds 085f62d67bbf9c501e8ac84d4533440a1eef6c45@95.217.196.54:26656,f515a8599b40f0e84dfad935ba414674ab11a668@osmosis.blockpane.com:26656
-git checkout v3.1.0
+time osmosisd start
```
ValarDragon, line 22: Does it work without seeds here? Do the seeds in v3.x still work now?

faddat: they don't :/

I'll download them from quicksync, that is what I just did ~20m ago.  

I'll also throw a PR into the v3.x branch that updates to the latest seeds from v7

ValarDragon (approved):

PR #1282, opened 2022-04-18
Antehandler for blocking validator configured sender

mconcat:
Closes: #XXX

need to change to checktx-only

## Description

______

For contributor use:


faddat (approved):
handy!

I assume that this is to help with the fact that we keep committing to the chain many failed txns per block?

ValarDragon:
The desired goal imo is for validators to have a mempool block for addresses only being able to send to other addresses, for when they wish to help with scammed mnemonic recovery, recover staked / bonded assets.

The prior state of the art involved was brittle - relying on spamming the mempool, with a race condition to get your tx first. (And getting as many validators to be locally spamming near the time of unlock)

The idea of this method is to allow validators who agree to add a mempool filter locally.

ValarDragon (approved):

PR #1285, opened 2022-04-18
support rocksdb v7

faddat:
Works towards: #695 

## Description

This PR migrates us away from gorocksdb wrapper library, which is no longer maintained, and to grocksdb, which is.  Along the way we get support for rocksdb v7:

https://github.com/facebook/rocksdb/releases/tag/v7.0.1

______

For contributor use:


alexanderbez (commented):
Shouldn't support for this version of RocksDB come from the (our) SDK dependency?

faddat:
oh, if it is notional, that's in error

ValarDragon:
Have you checked if we can upgrade an existing rocksDB database to this new import, and if things still work? If you've tested that and it works, then imo its good to merge!

If that doesn't work, we should call this db something else in tm-db / makefile OR communicate to all validators on the old RocksDB to switch in the interrim. (I would prefer not doing the latter option, or at minimum only considering it after state sync is up)

faddat:
@ValarDragon 

* I think it is compatible with v6
* I think that we were the only ones to use rocksdb as a validator, and tbh, it did not go well (yet).  The biggest issue was state sync, and that broke (for rocks specifically) when we added wasm to osmosis

This is kinda too many "I thinks" for me, I'd like to suggest that we merge this when this is merged:

https://github.com/tendermint/tm-db/pull/239

That way, our tm-db continues to match the one upstream (our iavl currently relies on that repo as well)

p0mvn:
Hi @faddat . I'm reviewing open PRs. Are we still blocked on the PR in `tm-db`?

faddat:
Yes.  We are still blocked on that, here’s a tracker:

https://github.com/tendermint/tm-db/issues/248

faddat:
@marbar3778 I would like to try to resolve this.

faddat:
Regarding this one I intend to bring up the core issues at play on Friday's SDK architecture review call. 

I'd very much like to get it in.

ValarDragon:
Adding the blocked label for now!

ValarDragon:
Closing since we're not actively working on this rn / its gone stale.

Lets re-open when we can allocate attention again?

Seems like this is actually super close to being in a state we can use? We basically needed to:
* Update our TM-DB version
* See what happens running a rocksDB instance on mainnet / it still works
* See what happens if I had a rocks DB v6 version, and we switched to this for v7. (Did things break?)

Issue #1284, opened 2022-04-18
libwasmvm.so is missing for non-Docker setup


vorotech:
### What is the issue?

`libwasmvm.so` is missing when executing `osmosisd` binary downloaded from the releases page.

This is not a Docker based, but Full Node setup

```sh
$ osmosisd init moniker-1 --home=/home/cosmos/.osmosisd --chain-id=osmosis-1
...
osmosisd: error while loading shared libraries: libwasmvm.so: cannot open shared object file: No such file or directory
```

### System

```sh
$ uanme -a
Linux #23~20.04.2-Ubuntu SMP Thu Mar 31 11:36:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ go version
go version go1.18.1 linux/amd64
```

`osmosisd` is https://github.com/osmosis-labs/osmosis/releases/download/v7.2.0/osmosisd-7.2.0-linux-amd64

Last known working version for both downloaded binary as well as built from source without any issues is v6.4.0

### Workaround 

The only workaround I've found is to build from sources, but make extra steps similar to those in Docker file

```sh
wget https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0-beta7/libwasmvm_muslc.a -O /lib/libwasmvm_muslc.a
git clone https://github.com/osmosis-labs/osmosis
cd osmosis
BUILD_TAGS=muslc make build
cp ./build/osmosisd /usr/local/bin
```

NOTE: Only version `v1.0.0-beta7` of CosmWasm/wasmvm is working, other (newer) has breaking interface changes

### Expected result

The `osmosisd` downloaded from the releases page works without any additional requirements or at lease this process is documented.

p0mvn:
Thank you for this issue. It seems that our [builder image](https://github.com/osmosis-labs/osmosis/blob/298ad00dc9f792dfccf10e864002a02bcdd0b3fb/contrib/images/rbuilder/Dockerfile) was not updated with the  cosm wasm related dependencies. I will look into it

alexanderbez:
Does this have anything to do with cross-compiling? If so, enabling CGO and shared libs in automated builds that cross compile is not possible AFAIK (e.g. this is why Ledger doesn't work on cross-compiled builds).

ValarDragon:
My understanding is that cosmwasm statically links the wasmvm binary on supported systems, but on any unsupported system it wouldn't, and requires it to be dynamically linked.
[...]

PR #1286, opened 2022-04-18
fix: release build scripts

p0mvn:
Closes: #1284

This change fixes our `make build-reproducible` Makefile step to generate release binaries. As described in the issue #1284, our current binaries do not have cosm wasm dependencies correctly installed. Wasmvm binaries need to be statically linked.

On initial investigation, I observed that our `rbuilder` is broken when built locally. The version from docker hub that is used by default when local version is not available has not been updated in around a year and does not work with the `main` branch of our repository.

As a result, our `rbuilder` was fixed in this PR by referencing the updated version here: https://github.com/tendermint/images/tree/master/rbuilder

Additionally, it does not make sense to rely on the cosmoshub image (`cosmossdk/rbuilder:latest`). That is not built by our team.

I propose to push the final result of the `rbuilder` in this PR to our Dockerhub (i.e `osmolabs/rbuilder`) CC: @nikever 

## Brief change log
- fix `make build-reproducible` and all related scripts by referencing: https://github.com/tendermint/images/tree/master/rbuilder
- add logic for statically linking wasmvm
- fix permission errors related to cosm wasm library running in container as non-root

## PR Formatting

- Is PR targeted against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))  yes

## Testing and Verifying

Built the binary with `make built-reproducible` on main branch and successfully ran it without getting the error described in #1284:
```
./osmosisd-7.0.0-123-g33a436a2-linux-amd64 start
7:09PM INF starting ABCI with Tendermint
```

  - How is the feature or change documented? added a README in `contrib/images/README.md`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ endif
 ifeq (,$(findstring nostrip,$(OSMOSIS_BUILD_OPTIONS)))
   ldflags += -w -s
 endif
-ifeq ($(LEDGER_ENABLED),true)
+ifeq ($(LINK_STATICALLY),true)
 	ldflags += -linkmode=external -extldflags "-L/lib/ -lwasmvm_muslc -Wl,-z,muldefs -static"
 endif
 ldflags += $(LDFLAGS)
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ build-reproducible: go.sum
 	--env COMMIT=$(COMMIT) \
 	--env LEDGER_ENABLED=$(LEDGER_ENABLED) \
 	--name latest-build osmolabs/rbuilder:latest
-	$(DOCKER) cp -a latest-build:/root/artifacts/ $(CURDIR)/
+	$(DOCKER) cp -a latest-build:/home/builder/artifacts/ $(CURDIR)/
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -85,7 +88,7 @@ all: install lint test
 
 BUILD_TARGETS := build install
 
-build: BUILD_ARGS=-o $(BUILDDIR)/
+ build: BUILD_ARGS=-o $(BUILDDIR)/
```
ValarDragon, line 91: is this leading space right?  I thought make file commands have to have no leading spaces for the definition

p0mvn: This is a typo that is fixed now. Thanks for catching that

It worked locally for me. I remember we were having issues in CI when having 4 spaces instead of tabs. That seems to be breaking things

ValarDragon (approved):
This LGTM, but want to leave open in case others who know more than me about the build system get a chance to take a look.

p0mvn:
This is ready now

nikever (approved):
LGTM

PR #1287, opened 2022-04-18
Add boilerplate logic for stableswap swaps and spot price

ValarDragon:
Adds logic for stableswap implementing the swap and spot price interfaces. The PR is marked as draft until some test cases are added.
______

For contributor use:o


x/gamm/pool-models/stableswap/pool.go
```diff
@@ -68,34 +69,85 @@ func (pa Pool) getPoolAmts(denoms ...string) ([]sdk.Int, error) {
...
 func (pa *Pool) SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.Coin, err error) {
-	return sdk.Coin{}, types.ErrNotImplemented
+	tokenOutDec, err := pa.CalcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	// we ignore the decimal component, as token out amount must round down
+	tokenOut, _ = tokenOutDec.TruncateDecimal()
+	if !tokenOut.Amount.IsPositive() {
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
```
p0mvn, line 106: does it make sense to format what it actually was in the message?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -68,34 +69,85 @@ func (pa Pool) getPoolAmts(denoms ...string) ([]sdk.Int, error) {
...
+// It requires caller to validate that tokensIn and tokensOut only consist of
+// denominations in the pool.
+// The function sanity checks this, and panics if not the case.
+func (p *Pool) applySwap(tokensIn sdk.Coins, tokensOut sdk.Coins) {
+	l := p.PoolLiquidity.Len()
+	// update liquidity
+	p.PoolLiquidity = p.PoolLiquidity.Add(tokensIn...).Sub(tokensOut)
+	// sanity check that no new denoms were added
+	if len(p.PoolLiquidity) != l {
+		panic("applySwap changed number of tokens in pool")
```
p0mvn, line 82: q: At first, I thought that we can propagate the error up top. However, it seems that the panic here is for a reason. We don't want to let the caller handle the error if the number of tokens happens to change. Is my reasoning correct?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -68,34 +69,85 @@ func (pa Pool) getPoolAmts(denoms ...string) ([]sdk.Int, error) {
...
+	}
+	pa.applySwap(tokenIn, sdk.NewCoins(tokenOut))
+
+	return tokenOut, nil
 }
 
 func (pa Pool) CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.DecCoin, err error) {
-	return sdk.DecCoin{}, types.ErrNotImplemented
+	if tokenOut.Len() != 1 {
+		return sdk.DecCoin{}, errors.New("stableswap CalcInAmtGivenOut: tokenOut is of wrong length")
```
p0mvn, line 115: nit: format the length if makes sense

x/gamm/pool-models/stableswap/pool.go
ValarDragon: yeah, this should never happen. (It should have been caught by the methods that call it in this package, and not be able to be triggered from functions outside the package)

ValarDragon:
Decision from call: Add test for this module, don't amend the more general testing Roman recently added right now

x/gamm/pool-models/balancer/amm.go
```diff
@@ -131,7 +131,11 @@ func (p *Pool) SwapInAmtGivenOut(
 	if err != nil {
 		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount is zero or negative")
 	}
-	tokenInCoin, _ := tokenInDecCoin.TruncateDecimal()
+	tokenInCoin, tokenInDecimal := tokenInDecCoin.TruncateDecimal()
+	// if tokenInDecimal is non-zero, we add 1 to the tokenInCoin
+	if tokenInDecimal.Amount.IsPositive() {
+		tokenInCoin.Amount.AddRaw(1)
```
mattverse, line 137: don't we have to do `tokenInCoin = tokenInCoin.Amount.AddRaw(1)` ?

mattverse: Also, I'm curious why we add 1 in the case of non-zeros!

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -68,34 +69,85 @@ func (pa Pool) getPoolAmts(denoms ...string) ([]sdk.Int, error) {
...
+// denominations in the pool.
+// The function sanity checks this, and panics if not the case.
+func (p *Pool) updatePoolLiquidityForSwap(tokensIn sdk.Coins, tokensOut sdk.Coins) {
+	numTokens := p.PoolLiquidity.Len()
+	// update liquidity
+	p.PoolLiquidity = p.PoolLiquidity.Add(tokensIn...).Sub(tokensOut)
+	// sanity check that no new denoms were added
+	if len(p.PoolLiquidity) != numTokens {
+		panic("updatePoolLiquidityForSwap changed number of tokens in pool")
+	}
```
mattverse, line 83: love this! 🚀

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -188,3 +188,31 @@ func spotPrice(baseReserve, quoteReserve sdk.Dec) sdk.Dec {
...
+func (pa *Pool) calcInAmtGivenOut(tokenOut sdk.Coin, tokenInDenom string, swapFee sdk.Dec) (sdk.Dec, error) {
+	reserves, err := pa.getPoolAmts(tokenInDenom, tokenOut.Denom)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+	tokenInSupply := reserves[0].ToDec()
+	tokenOutSupply := reserves[1].ToDec()
+	// We are solving for the amount of token in, cfmm(x,y) = cfmm(x + x_in, y - y_out)
+	// x = tokenInSupply, y = tokenOutSupply, yIn = -tokenOutAmount
+	inAmtRaw := solveCfmm(tokenInSupply, tokenOutSupply, tokenOut.Amount.ToDec().Neg())
```
mattverse, line 215: Could you explain why we use a negative value for y in?

mattverse (commented):
Left some minor comments! (mostly comments for the purpose of actually understanding what's happening rather than changes requests)

x/gamm/pool-models/balancer/amm.go
ValarDragon: ah yeah your right, good catch.

The idea here is that I want to get exactly `100` tokens out after the trade. But to do that, I have to provide say `50.82` tokens as input. Tokens are stored as integers, so before we were only requiring `50` tokens to be sent. But actually it should have been `51` when its converted to integers.

mattverse (approved):
Core function LGTM once https://github.com/osmosis-labs/osmosis/pull/1287/files#r855833475 is fixed.

I propose we merge this to main without tests as of now, since this PR is a big blocker in gamm generalization parallelization!

x/gamm/pool-models/balancer/amm.go
ValarDragon: I'm going to revert this change for Balancer, and fix it in a second PR that fixes CalcSpotPrice to be an integer API

Issue #756, opened 2022-01-13
Querying account's LP shares by lockup duration


czarcas7ic:
Currently, there is no way to query an account's locked LP shares by denom and lock length. The closest we can get to this is by querying  ```osmosisd query lockup account-locked-longer-duration-denom [address] [duration] [denom]``` 
This however separates all the outputs by lock-id.

The reason for this (thank you to @ValarDragon for the explanation) is that during launch, there was a gas bug whose mitigation in the front-end involved creating a slight "jitter" in the lock duration by adding slight ms to the end of the duration. 

We can counteract this jitter by either creating a `--jitter` flag that takes into account plus or minus the defined jitter value (I think the jitter only happened in the positive direction but just saying plus or minus in case this is not true) and returns all bonded tokens within the range of the duration +- the defined jitter OR find the max jitter in either direction and simply implement that in the query (i.e if someone queries their locked tokens for denom x for 1 day lockup period, it automatically assumes you want to include any LP shares of 1day +- max jitter).

ValarDragon:
yeah jitter should've only happened in the positive direction.

PR #1292, opened 2022-04-19
Feat: CLI account-locked-duration

CmpHDL:
Closes: #756 

## Description

Adds CLI command to query pool lockups with given duration. The jitter issue outlined in #756 no longer applies due to +2 weeks (longest lockup time) passing since v7 has been out.

Example Usage:
```bash
Query account locked records with a specific duration.

Example:
$ osmosisd query lockup account-locked-duration <address> <duration>

Usage:
  osmosisd query lockup account-locked-duration <address> <duration> [flags]
```
______

For contributor use:


CmpHDL:
This was based off of v7.0.2, I can re-base this off of main if needed.

ValarDragon:
Please do rebase the change off of main

Knstld:
@ValarDragon I had rebased off main after that comment starting with commit [93fd8c3](https://github.com/osmosis-labs/osmosis/pull/1292/commits/93fd8c397afb85bebdec37f963d25264996e090a), were you wanting to remerge main based on new commits since then? Also I refixed the typo that somehow got reversed during the rebase so a rerun of workflows should satisfy the linter.

CmpHDL:
@ValarDragon I had rebased off main after that comment starting with commit [93fd8c3](https://github.com/osmosis-labs/osmosis/pull/1292/commits/93fd8c397afb85bebdec37f963d25264996e090a), were you wanting to remerge main based on new commits since then? Also I refixed the typo that somehow got reversed during the rebase so a rerun of workflows should satisfy the linter.

proto/osmosis/lockup/query.proto
```diff
@@ -95,6 +95,14 @@ service Query {
     option (google.api.http).get =
         "/osmosis/lockup/v1beta1/account_locked_longer_duration/{owner}";
   }
+
+  // Returns account locked records with a specific duration
+  rpc AccountLockedDuration(AccountLockedDurationRequest)
+      returns (AccountLockedDurationResponse) {
+    option (google.api.http).get =
+        "/osmosis/lockup/v1beta1/account_locked_duration/{owner}";
```
alexanderbez, line 103: Since `AccountLockedDurationRequest` defines the query params, is there a point in having `{owner}` as part of the resource URL?

proto/osmosis/lockup/query.proto
```diff
@@ -243,6 +251,18 @@ message AccountLockedLongerDurationResponse {
   repeated PeriodLock locks = 1 [ (gogoproto.nullable) = false ];
 };
 
+message AccountLockedDurationRequest {
```
alexanderbez, line 254: nit: proto doc comment :)

proto/osmosis/lockup/query.proto
```diff
@@ -243,6 +251,18 @@ message AccountLockedLongerDurationResponse {
...
+  string owner = 1 [ (gogoproto.moretags) = "yaml:\"owner\"" ];
+  google.protobuf.Duration duration = 2 [
+    (gogoproto.stdduration) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"duration\""
+  ];
+};
+message AccountLockedDurationResponse {
```
alexanderbez, line 262: nit: proto doc comment :)

```suggestion
};

message AccountLockedDurationResponse {
```

x/lockup/client/cli/query.go
```diff
@@ -535,7 +536,48 @@ $ %s query lockup account-locked-longer-duration <address> <duration>
...
-// GetCmdAccountLockedLongerDurationNotUnlockingOnly returns account locked records with longer duration from unlocking only queue.
+// GetCmdAccountLockedDuration returns account locked records with a specific duration
+func GetCmdAccountLockedDuration() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "account-locked-duration <address> <duration>",
+		Short: "Query account locked records with a specific duration",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Query account locked records with a specific duration.
+Example:
+$ %s query lockup account-locked-duration <address> <duration>
```
alexanderbez, line 547: If we're gonna include an example in the `Long` description, we should actually make it a true example. I.e. fill out the `address` and `duration` values 👍

x/lockup/client/cli/query.go
```diff
@@ -535,7 +536,48 @@ $ %s query lockup account-locked-longer-duration <address> <duration>
 	return cmd
 }
 
-// GetCmdAccountLockedLongerDurationNotUnlockingOnly returns account locked records with longer duration from unlocking only queue.
+// GetCmdAccountLockedDuration returns account locked records with a specific duration
```
alexanderbez, line 539:
```suggestion
// GetCmdAccountLockedDuration returns account locked records with a specific duration.
```

x/lockup/client/cli/query.go
```diff
@@ -535,7 +536,48 @@ $ %s query lockup account-locked-longer-duration <address> <duration>
...
+			}
+
+			duration, err := time.ParseDuration(args[1])
+			if err != nil {
+				return err
+			}
+
+			queryClient := types.NewQueryClient(clientCtx)
+
+			res, err := queryClient.AccountLockedDuration(cmd.Context(), &types.AccountLockedDurationRequest{Owner: args[0], Duration: duration})
```
alexanderbez, line 566: Do we have a type constructor for `AccountLockedDurationRequest`?

x/lockup/keeper/iterator.go
```diff
@@ -36,6 +36,7 @@ func (k Keeper) iteratorBeforeTime(ctx sdk.Context, prefix []byte, maxTime time.
 	return store.Iterator(prefix, storetypes.PrefixEndBytes(key))
 }
 
+// @TODO: confirm this existing iterator is accurate
```
alexanderbez, line 39: What are the TODOs for? We should create issues for any TODOs we introduce in code.

x/lockup/keeper/store.go
```diff
@@ -184,7 +184,15 @@ func (k Keeper) GetAccountLockedLongerDuration(ctx sdk.Context, addr sdk.AccAddr
 	return combineLocks(notUnlockings, unlockings)
 }
 
-// GetAccountLockedLongerDurationNotUnlockingOnly Returns account locked with duration longer than specified.
+// GetAccountLockedDuration Returns account locked with a specific duration
+func (k Keeper) GetAccountLockedDuration(ctx sdk.Context, addr sdk.AccAddress, duration time.Duration) []types.PeriodLock {
+	// it does not matter started unlocking or not for duration query
+	unlockings := k.getLocksFromIterator(ctx, k.AccountLockIteratorDuration(ctx, true, addr, duration))
+	notUnlockings := k.getLocksFromIterator(ctx, k.AccountLockIteratorDuration(ctx, false, addr, duration))
```
p0mvn, line 191: This comment doesn't need to be directly addressed in this PR: 

I'm wondering if there is a way to avoid these double iterations with false and true values? If only we could specify to the iterator to consider any values for the next X bytes in the prefix, that would prevent us from having to iterate multiple times

x/lockup/keeper/iterator.go
```diff
@@ -148,7 +150,13 @@ func (k Keeper) AccountLockIteratorLongerDuration(ctx sdk.Context, isUnlocking b
 	return k.iteratorLongerDuration(ctx, combineKeys(unlockingPrefix, types.KeyPrefixAccountLockDuration, addr), duration)
 }
 
-// AccountLockIteratorShorterThanDuration returns iterator used for getting all locks by account longer than duration.
+// AccountLockIteratorDuration returns iterator used for getting all locks by account for a specific duration
```
p0mvn, line 153:
```suggestion
// AccountLockIteratorDuration returns an iterator used for getting all locks for a given account, isUnlocking, and specific duration.
```

x/lockup/keeper/store.go
```diff
@@ -184,7 +184,15 @@ func (k Keeper) GetAccountLockedLongerDuration(ctx sdk.Context, addr sdk.AccAddr
...
+	return combineLocks(notUnlockings, unlockings)
```
p0mvn, line 192:
```suggestion
	unlockedLocks := k.getLocksFromIterator(ctx, k.AccountLockIteratorDuration(ctx, true, addr, duration))
	lockedLocks := k.getLocksFromIterator(ctx, k.AccountLockIteratorDuration(ctx, false, addr, duration))
	return combineLocks(unlockedLocks, lockedLocks)
```

p0mvn (changes requested):
Thank you for this change @CmpHDL 

I left some comments, please take a look

CmpHDL:
Thanks for all of the comments and feedback guys, hoping to be able to contribute more to the AMM in the near future, not feeling well today so I'll do a more detailed review tomorrow.

x/lockup/client/cli/query.go
CmpHDL: Are we wanting the example to work? We would need an address that always had bonded assets for 2 different durations.

x/lockup/client/cli/query.go
CmpHDL: Like a `NewAccountLockedDurationRequest()`? Can you elaborate further on this?

x/lockup/keeper/store.go
CmpHDL: I like these variable names much better, great suggestion, much clearer.

x/lockup/client/cli/query.go
alexanderbez: Yup, let's use constructors if they exist.

x/lockup/keeper/iterator.go
```diff
@@ -47,6 +48,7 @@ func (k Keeper) iteratorLongerDuration(ctx sdk.Context, prefix []byte, duration
 	store := ctx.KVStore(k.storeKey)
 	durationKey := getDurationKey(duration)
 	key := combineKeys(prefix, durationKey)
+	// @TODO: ask for elaboration on how this works
 	// inclusive on longer side, means >= (longer or equal)
```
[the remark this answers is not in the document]
CmpHDL: > Thanks for this PR!
> 
> I might have misunderstood issue #756, but seems to me that the direction this PR points to vs the original issue mentioned in the issue seems different.
> 
> How I understood issue #756 was that we can't query total amount of specific denom that's over specific amount of duration, since the query `account-locked-longer-duration-denom` returns all lockups separately instead of adding up the what's in individual lockups.
> 
> This PR, however seems to have the same problem, seems like it's closer to adding a query to query all denoms instead of specific denom over specified time. (I still think this query is superuseful, just wanted to confirm my understanding on the existing issue and the current PR submitted!)

The ticket changed a little due to jitter no longer being an issue and yes I planned todo the same command but for denoms after this one. Rather than do both of them at once I wanted to see the process etc for making a change and what kinds of feedback you would have, and better understand things as a whole, rather than making the same mistakes twice and having to fix them twice in separate PRs. 

tl;dr
Once this goes through I will be making the command you're talking about with `denom` as an additional parameter, this one was easier so decided to give it a go right before its brother which includes `denom`.

CmpHDL:
> Thanks for this PR!
> 
> I might have misunderstood issue #756, but seems to me that the direction this PR points to vs the original issue mentioned in the issue seems different.
> 
> How I understood issue #756 was that we can't query total amount of specific denom that's over specific amount of duration, since the query `account-locked-longer-duration-denom` returns all lockups separately instead of adding up the what's in individual lockups.
> 
> This PR, however seems to have the same problem, seems like it's closer to adding a query to query all denoms instead of specific denom over specified time. (I still think this query is superuseful, just wanted to confirm my understanding on the existing issue and the current PR submitted!)

The ticket changed a little due to jitter no longer being an issue and yes I planned todo the same command but for denoms after this one. Rather than do both of them at once I wanted to see the process etc for making a change and what kinds of feedback you would have, and better understand things as a whole, rather than making the same mistakes twice and having to fix them twice in separate PRs. 

tl;dr
Once this goes through I will be making the command you're talking about with `denom` as an additional parameter, this one was easier so decided to give it a go right before its brother which includes `denom`.

x/lockup/keeper/iterator.go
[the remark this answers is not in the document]
CmpHDL: Do you have discord? I'd like to add you there to discuss further, this is really insightful.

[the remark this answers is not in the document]
p0mvn: Always happy to discuss: `r0man#8579`

CHANGELOG.md
```diff
@@ -107,7 +107,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Minor improvements & Bug Fixes
 
 * [#1022](https://github.com/osmosis-labs/osmosis/pull/1022) upgrade iavl to v0.17.3-osmo-v4 - fix state export at an old height
```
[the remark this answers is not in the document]
CmpHDL: Thanks, couldn't find the unreleased portion until rebasing off of main for obvious reasons.

proto/osmosis/lockup/query.proto
CmpHDL: Can you elaborate on this? `AccountLockedPastTimeDenom`, `AccountLockedPastTimeNotUnlockingOnlyRequest`, and most of the locked RPC commands have the {owner} within the resource URL, I had assumed it was for the address parameter and maintained practice.

x/lockup/client/cli/query.go
CmpHDL: Gotcha, so the command itself doesn't have to return results that satisfy an example usage's result, we just need an example usage.

x/lockup/client/cli/query.go
CmpHDL: I don't believe we currently have a constructor for this type and all other CLI commands seem to mimic this usage of passing the reference of a direct type initialization.

x/lockup/client/cli/query.go
CmpHDL: @alexanderbez If we want to go this route for `Long` descriptions, we should have constants that signify commonly used arguments. Eg: `EXP_BECH_32` `EXP_DURATION` `EXP_POOL`

mattverse (approved):
LGTM!

x/lockup/keeper/store.go
mattverse: Are they double iterating? It seems to me that they're not, since unlocking locks and locked locks don't have the same prefix and iterator only iterates over given prefix

CmpHDL:
@mattverse Ended up fixing it now, `query.go` should pass, looks like a cat had walked along mid commit.

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* [#1292](https://github.com/osmosis-labs/osmosis/pull/1292) Feat: CLI account-locked-duration
```
alexanderbez, line 44: Add this under `Features` and remove the `Feat` prefix 👍

x/lockup/client/cli/query.go
```diff
@@ -535,6 +536,47 @@ $ %s query lockup account-locked-longer-duration <address> <duration>
...
 
+// GetCmdAccountLockedDuration returns account locked records with a specific duration.
+func GetCmdAccountLockedDuration() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "account-locked-duration <address> <duration>",
+		Short: "Query account locked records with a specific duration",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Query account locked records with a specific duration.
+Example:
+$ %s query lockup account-locked-duration <address> <duration>
```
alexanderbez, line 547: I would remove this entirely  and utilize Cobra's `Example` field instead. We can then get rid of `Long` too.

CmpHDL: Done

CHANGELOG.md
CmpHDL: Done

CmpHDL:
@alexanderbez Made your requested changes.

x/lockup/client/cli/query.go
p0mvn: Can we introduce these constants here?

proto/osmosis/lockup/query.proto
p0mvn: @alexanderbez I'm not very familiar with this RPC library - does it make sense for us to create an issue and remove potentially redundant `{owner}`?

x/lockup/keeper/iterator.go
```diff
@@ -148,7 +148,13 @@ func (k Keeper) AccountLockIteratorLongerDuration(ctx sdk.Context, isUnlocking b
...
 }
 
-// AccountLockIteratorShorterThanDuration returns iterator used for getting all locks by account longer than duration.
+// AccountLockIteratorDuration returns an iterator used for getting all locks for a given account, isUnlocking, and specific duration.
+func (k Keeper) AccountLockIteratorDuration(ctx sdk.Context, isUnlocking bool, addr sdk.AccAddress, duration time.Duration) sdk.Iterator {
+	unlockingPrefix := unlockingPrefix(isUnlocking)
+	return k.iteratorDuration(ctx, combineKeys(unlockingPrefix, types.KeyPrefixAccountLockDuration, addr), duration)
+}
+
+// AccountLockIteratorShorterThanDuration returns iterator used for getting all locks by account longer than duration
```
p0mvn, line 157:
```suggestion
// AccountLockIteratorShorterThanDuration returns an iterator used for getting all locks by account shorter than the specified duration
```

p0mvn (approved):
Please address the remaining minor comments but otherwise LGTM. Great job!

proto/osmosis/lockup/query.proto
mattverse: These urls are mostly used in front end, but we definitely don't need to have them both in the request url and the request itself. Although I don't know if it's worth changing considering the fact that all the urls and queries in the frontend would have to be updated along as well

x/lockup/keeper/iterator.go
CmpHDL: @p0mvn This is outside this ticket, I'll make the change though.

x/lockup/client/cli/query.go
CmpHDL: @p0mvn Make a new ticket for this as this would impact basically all commands.

Issue #1235, opened 2022-04-12
[E2E - Upgrade Testing] Add CI pipelines to test upgrades and chain integrity post-upgrade


nikever:
**Background**

Now that we have a e2e testing process in place to the test basic functionality of the chain, we can expand our test suite to test chain upgrades and make sure that basic properties are being preserved after the update

**Tasks**

- #1293 
- #1330
- #1332 
- #1338
- #1333 
- #1414 
- #1415

**Acceptance Criteria**

- Have an automated flow to test an upgrade and perform basic tests after the upgrade to ensure chain integrity.

faddat:
@nikever do you have a way around the github self-hosted runners lack of concurrency / extreme expense of the permium runners they sell?

basically, since their runners really do not match up to the performance of what most node operators use, I am not sure that this path can work, unless we are restricting the state that is used somehow (genesis export maybe?)

ValarDragon:
Can we close this issue? cc @p0mvn

PR #1293, opened 2022-04-19
refactor: begin modularizing e2e tests in preparation for upgrade

p0mvn:
Closes: #XXX

This is a spike related to #1235 to determine a test plan needed for implementing the upgrade. 

Currently, there is no way to create genesis or configs such as `app.toml` for an older version of Osmosis because we directly depend on the branch that executes e2e tests. This is required to implement the logic for testing the upgrade.

This change decouples the functionality that directly depends on the Osmosis codebase from the e2e test runner by creating a `chain` package that is independent of the `e2e` package.

By decoupling the logic for configuring the chain (creating genesis + config files), we can use an older Docker image of the Osmosis codebase to configure the chains for the desired version. Then, we may use the created genesis and configs for the old version to execute the upgrade and perform e2e tests from the `main` branch. 

Please note that this PR does not introduce any functional changes and only modularizes and decouples the code to make it possible to add the new upgrade functionality.

## Brief change log

- create the `chain` package
- move some utility and io logic that is unrelated to the `chain` to `util` package
- move all remaining logic that is directly dependent on Osmosis codebase to `chain` package
- refactor `chain` package so that:
   * `main.go` - contains a single `Init()` function that initializes a chain with 2 validators
      * This function is called from the test setup in `e2e` package
   * `chain.go` and `validator.go` are mostly unchanged and only contain the implementations for `Chain` and `Validator` structs respectively
   * `util.go` is moved to `chain/chain_util.go`
   * config.go has constants and configuration logic
- refactor `e2e_setup.go` to have the following flow in setup:
   * create 2 chains by calling the `chain.Init(<id>)` method for each
      * the plan is to do this in a Docker container for a specific version of Osmosis in the future PR
   * configure Docker resources
   * run the 2 chains
   * run relayer

## Testing and Verifying

- run the e2e tests several times without issues

  - How is the feature or change documented? not applicable - will be documented when the upgrade implementation is finalized


czarcas7ic (approved):
Tested this locally as well and all prior tests still pass. Totally agree with pushing chain to a separate library. Great work!

I think the next step is figuring out what changes we need to make to the current osmosis docker image in order to facilitate these upgrade tests. After that, I think the next steps should be more straight forward.

alexanderbez:
Looks like this just moved a few things around to isolate logic in packages with a few tweaks. LGTM.

Issue #1148, opened 2022-03-25
Make every module use apptesting for its test suites


ValarDragon:
We made the apptesting helper package, so we dont have modules re-implementing each others test scaffolding. We need to propogate this change to all of our existing modules.

Right now its being a painpoint for a lot of the GAMM updates I'm working on. (Modules like pool-incentives & txfees have copy paste methods of setup code from gamm)

catShaark:
so there's differences between each module's `KeeperTestSuite`, for example some modules use `keeper.Querier` for grpc query test, while some others use `types.queryClient` for grpc query test. Should we have a standard for `KeeperTestSuite` that every modules has to follow.

PR #1297, opened 2022-04-19
Test suites refactor

catShaark:
Closes: #1148

## Description
I try to put all the necessary test suite logic of modules into apptesting package, like @ValarDragon describing in #1148

______

For contributor use:


catShaark:
So I put some the logic commonly used in all the modules test, put them into method of KeeperTestHelper, I also add test accounts as a field in KeeperTestHelper to replace acc1, acc2, acc3 in keeper_test.go

alexanderbez (approved):

ValarDragon:
Thanks so much for making this! Reviewing now

ValarDragon:
Oh, I wasn't sure about the usage of the static addresss list that was changed in a lot of the tests. Let me make a separate issue for that then

catShaark:
> Oh, I wasn't sure about the usage of the static addresss list that was changed in a lot of the tests. Let me keep investigating that issue then.
> 
> Basically `suite.TestAccs[0]` being used instead of acc1, and if thats safe or not

The suite.TestAccs[] thing is inspired from `ibc-test`'s `testchain` where they store the account and its key to do action on the `testchain`, also some helper code in `apptesting` like func `PrepareBalancerPoolWithPoolParams` needs to have a `pool creating account` that is accessible to `gamm`'s `suite` so that the main reason I create `TestAccs` . 

> 
> I guess if the policy is that all re-usable helper code should use new temporary keys, it could be fine?

PR #1301, opened 2022-04-20
Remove claim module

faddat:
Closes: #903 

This PR:

* Removes the claim module
* changes the module path of osmosis to github.com/osmosis-labs/osmosis/v8

PR #1304, opened 2022-04-20
Add upgrade handler for v8

ValarDragon:

This PR adds boilerplate for the v8 upgrade handler. Simplifies the logic for upgrade store handler updates. Intended as a precursor for #1301 

## Brief change log

- Adds boilerplate for the v8 upgrade handler. I don't think this necessitates an entry in the actual changelog.

## Testing and Verifying

This change should be tested in upcoming upgrade handler testing on testnets / in framework. It needs an end to end test to properly test changes to the store loaders.

  - How is the feature or change documented? NA


mattverse (commented):
LGTM! I was wondering if there's any logic we need to handle within the upgrade handler for v8. 

Should we also create an issue to change main to v8?

alexanderbez (commented):
Template LGTM!

Issue #1260, opened 2022-04-15
[x/gamm] implement JoinSwapShareAmountOut


p0mvn:
**Background**

We need to refactor `JoinSwapShareAmountOut` as part of #1050  and re-enable the tests, including here:
https://github.com/osmosis-labs/osmosis/blob/023928013a8c3b32171f68694013cdab084d2dcc/x/gamm/keeper/pool_service_test.go#L481-L486

**Acceptance Criteria**
- implement `JoinSwapShareAmountOut` 
- enable all tests for `JoinSwapShareAmountOut` as well as the inverse tests with `ExitSwapExternAmountOut`
- add more unit tests
- all unit tests pass

p0mvn:
@ValarDragon FYI created this issue but it is missing implementation details. Please add them when you can

Issue #1288, opened 2022-04-18
add inverse unit tests for CalcInAmtGivenOut and CalcOutAmtGivenIn with non-zero swap fee


p0mvn:
**Background**

In #1254, we added inverse tests with a non-zero swap fee. We should continue by adding more related tests where a fee is non-zero.

**Acceptance Criteria**
- inverse unit tests for CalcInAmtGivenOut and CalcOutAmtGivenIn with non-zero swap fee added
- all unit tests pass

PR #1244, opened 2022-04-13
refactor: ExitSwapExternAmountOut

p0mvn:
Closes: #1130

## Description

- Re-implemented `ExitSwapExternAmountOut` with similar logic to v7
- introduced `PoolExitSwapExternAmountOutExtension`

## Risks

I was confused about the following part:
https://github.com/osmosis-labs/osmosis/blob/523ed353f033e8a7c0e9f86c72e6400c19918980/x/gamm/keeper/pool_service.go#L533-L543

I think that instead of doing that, it is possible to call `k.applyExitPoolStateChange(...)` so that it takes care of the fee since it does `k.BurnPoolShareFromAccount(ctx, pool, exiter, numShares)` where `numShares` do not have the `exitFee` subtracted. 

It seems to me that in the old logic subtracting the fee only to burn it separately was redundant, and I'm not sure if I'm missing some algebra reasoning behind it.

## Next Steps

I'm planning to spend some time writing more unit tests for this in a separate PR. Then, spearhead more e2e tests by joining a swap and exiting it using this method. 

______

For contributor use:

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

PR #1307, opened 2022-04-21
refactor: JoinSwapShareAmountOut

p0mvn:
Closes: #1260

This PR introduces the implementation of `JoinSwapShareAmountOut` on the balancer pool as part of the pool interface refactor.

Similar to #1244 , it uses an interface extension to add to the original `PoolI` interface. The bulk of the logic for the implementation was copied over from the `v7.x` branch.

More about this method can be found in our [docs](https://docs.osmosis.zone/developing/modules/spec-gamm.html#exit-swap-extern-amount-out)

## Brief change log

- Renamed `PoolExitSwapExactAmountOutExtension` to `PoolAmountOutExtension`
- Added new `JoinSwapShareAmountOut` method on the interface
- Implemented it in `balancer/gamm.go`
- refactored gamm keeper to call `JoinSwapShareAmountOut` on the pool
- uncommented the existing tests for `JoinSwapShareAmountOut` 
- performed clean up
  * reuse error messages as constants in amm.go
  * format values in the error messages
  * add comments for exported functions
  * `go mod tidy`

## Testing and Verifying

This change is already covered by existing tests, such as:
- `TestActiveBalancerPool`
- `TestNewJoinSwapShareAmountOutCmd`

More unit tests are needed and will be added after #1288

  - How is the feature or change documented? 
     * [docs](https://docs.osmosis.zone/developing/modules/spec-gamm.html#exit-swap-extern-amount-out)
     * [Comments](https://github.com/osmosis-labs/osmosis/blob/4439f0c1b4cd9df462305b8606a498ced360c07a/x/gamm/types/pool.go#L87-L88)

 ## Refactoring Ideas
Create the following proposal issues:
- refactor balancer package
   * rename `balancer_pool.go` to `pool.go` to match the naming pattern across various pool (balancer /stableswap)
   * issue to extract `math.go`
       * move some of the math related methods from `amm.go` into its own file `math.go` such as 
          * `feeRatio`
          * `solveConstantFunctionInvariant`
          * `calcSingleAssetInInGivenPoolSharesOut`
          * `calcPoolSharesInGivenSingleAssetOut`
          * `calcPoolSharesOutGivenSingleAssetIn`
       * These methods are not defined on the balancer Pool struct and are defined as functions. This change would make the `amm.go` less cluttered and only contain the balancer amm related Pool interface method implementations
- rename `JoinSwapShareAmountOut` to `JoinPoolShareAmountOut`
- rename `ExitSwapExactAmountOut` to `ExitPoolExactAmountOut`
   * It seems to me that a "JoinPool" / "ExitPool" is a better abstraction for these methods because that's essentially what's happening
   * This change should include the rename from the lowest level of abstraction in the balancer up until the CLI


x/gamm/keeper/pool_service.go
```diff
@@ -296,25 +295,14 @@ func (k Keeper) JoinSwapShareAmountOut(
...
-	if tokenInAmount.LTE(sdk.ZeroInt()) {
-		return sdk.Int{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount is zero or negative")
+	extendedPool, ok := pool.(types.PoolAmountOutExtension)
+	if !ok {
+		return sdk.Int{}, fmt.Errorf("pool with id %d does not support this kind of join", poolId)
 	}
 
-	if tokenInAmount.GT(tokenInMaxAmount) {
-		return sdk.Int{}, sdkerrors.Wrapf(types.ErrLimitMaxAmount, "%s token is larger than max amount", tokenInDenom)
+	tokenInAmount, err = extendedPool.JoinSwapShareAmountOut(ctx, tokenInDenom, shareOutAmount, tokenInMaxAmount)
```
ValarDragon, line 303: We should add the `tokenInAmount < tokenInMaxAmount` check at the keeper layer here

x/gamm/types/pool.go
```diff
@@ -75,15 +75,26 @@ type PoolI interface {
...
+type PoolAmountOutExtension interface {
 	PoolI
 
+	// JoinSwapShareAmountOut add liquidity to a specified pool with a maximum amount of tokens in (tokenInMaxAmount)
+	// and swaps to an exact number of shares (shareOutAmount).
+	JoinSwapShareAmountOut(
+		ctx sdk.Context,
+		tokenInDenom string,
+		shareOutAmount sdk.Int,
+		tokenInMaxAmount sdk.Int,
```
ValarDragon, line 93: I suggest we remove `tokenInMaxAmount` as an argument here

ValarDragon: Yeah I think so. Ideally, slippage bounds checks are something that the keeper layer is responsible for handling, so we know its done correctly independent of the underlying pools implementation.

p0mvn: It seems that I've encountered a similar rounding error to #1196. More investigation is needed to determine if these are related but I'm just documenting the findings for now.

The issue here is related to the panic in the code below:
```
func (k Keeper) JoinSwapShareAmountOut(
	ctx sdk.Context,
	sender sdk.AccAddress,
	poolId uint64,
	tokenInDenom string,
	shareOutAmount sdk.Int,
	tokenInMaxAmount sdk.Int,
	pool, err := k.getPoolForSwap(ctx, poolId)
	if err != nil {
		return sdk.Int{}, err
	}

	extendedPool, ok := pool.(types.PoolAmountOutExtension)
	if !ok {
		return sdk.Int{}, fmt.Errorf("pool with id %d does not support this kind of join", poolId)
	}

	tokenInAmount, err = extendedPool.CalcTokenInShareAmountOut(ctx, tokenInDenom, shareOutAmount, pool.GetSwapFee(ctx))
	if err != nil {
		return sdk.Int{}, err
	}

	if tokenInAmount.GT(tokenInMaxAmount) {
		return sdk.Int{}, sdkerrors.Wrapf(types.ErrLimitMaxAmount, "%d resulted tokens is larger than the max amount of %d", tokenInAmount.Int64(), tokenInMaxAmount.Int64())
	}

	tokenIn := sdk.Coins{sdk.NewCoin(tokenInDenom, tokenInAmount)}

	sharesCreated, err := extendedPool.JoinPool(ctx, tokenIn, pool.GetSwapFee(ctx))
	if sharesCreated != shareOutAmount {
		panic(fmt.Sprintf("%d shares created is not equal to the requested amount of %d", sharesCreated, shareOutAmount))
	}

	err = k.applyJoinPoolStateChange(ctx, pool, sender, shareOutAmount, tokenIn)
	if err != nil {
		return sdk.ZeroInt(), err
	}
	return tokenInAmount, nil
}
```

As per suggestion, I attempted to move validation logic to the keeper layer by refactoring to create a non-mutative method `CalcTokenInShareAmountOut`. Then, I noticed that it should be possible to reuse the existing `JoinPool` of the `PoolI` interface after validation on the keeper layer.

The resulting shares, returned from `JoinPool`, should be equal to the requested `shareOutAmount`. However, the above functionality panics with:
```
test panicked: {824636220864} shares created is not equal to the requested amount of {824633847808}
```

I will investigate further if these are related with #1196  or if this is an issue specific to my implementation

ValarDragon: Thanks, yeah this does look something has gone wrong

alexanderbez:
@p0mvn do you wanna tag/pair on this to figure out the rounding issue?

p0mvn:
@alexanderbez yes, please. That would be helpful

The way I was approaching this so far is by writing more unite tests:
- #1344 
- #1254

Currently, we don't have many unit tests so it's hard to identify the issue. I was planning to just continue incrementally increasing the number of tests, hoping to find the problem

Also, I'm waiting for #1345 to be merged to see if that helps mitigate this. It seems that it has a related fix.

Please let me know if you have any ideas on what else we can try

x/gamm/pool-models/balancer/amm.go
```diff
@@ -412,26 +396,106 @@ func (p *Pool) CalcExitPoolShares(ctx sdk.Context, exitingShares sdk.Int, exitFe
...
+	swapFee sdk.Dec,
+) (tokenInAmount sdk.Int, err error) {
+	_, poolAssetIn, err := p.getPoolAssetAndIndex(tokenInDenom)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+
+	normalizedWeight := poolAssetIn.Weight.ToDec().Quo(p.GetTotalWeight().ToDec())
+
+	tokenInAmount = calcSingleAssetInGivenPoolSharesOut(
```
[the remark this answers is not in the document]
p0mvn: I think yes because we want to err on the side of caution and round up to avoid charging less than needed for shareOutAmount. @ValarDragon wdyt?

x/gamm/types/pool.go
mattverse: Oh wow, surprised that we had such rounding issues! Thanks for investigating on this by comparing it with v7.x!

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
alexanderbez: That makes sense to me.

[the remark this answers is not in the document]
p0mvn: Thanks for the feedback @alexanderbez . Updated and going to merge this

PR #1310, opened 2022-04-21
fix: minor security findings

jim380:
Closes: #XXX

This PR addresses a couple of minor security findings when I was getting ready for a fuzzing session.

- fix implicit memory aliasing in for loop
- fix file inclusion via variable
- fix unhandled errors
- fix gosec in github workflow

### Implicit Memory Aliasing in `for` Loop

- osmosis/x/lockup/keeper/msg_server.go:142
- osmosis/x/incentives/genesis.go:16

### Description
In a for loop in Golang the iterator variable is reused in each iteration. In other words, only the value of v changes upon each iteration; its memory address stays the same. Storing the address of the iterator variable could cause issues because if/when you had to dereference the pointer, the value that the pointer points to might have changed.

### Mitigation
- Index the ranged slice
```go
for i := range unlocks {
      events = events.AppendEvent(createBeginUnlockEvent(&unlocks[i]))
}
```
- Reassign inside the loop
```go
for _, lock := range unlocks {
     lock := lock
     events = events.AppendEvent(createBeginUnlockEvent(&lock))
}
```
 ### Potential File Inclusion via Variable

- osmosis/cmd/osmosisd/cmd/balances_from_state_export.go:123

### Description
`getGenStateFromPath` is called with a filePath that's user-supplied and insufficiently validated.

### Mitigation
Use `filepath.Clean` to find the shortest path.
```go
genesisFile, err := os.Open(filepath.Clean(genesisFilePath))
```

 ### Unhandled Errors
- osmosis/store/tree.go:101
- osmosis/app/export.go:175

### Description
`iter.Close()` returns an err and it’s not currently being handled.

### Mitigation
Properly handle the error.

```go
err := iter.Close()
if err != nil {
     panic(err)
}
```


ValarDragon (changes requested):
Thanks for catching these problems! Can we keep gosec disabled for now, and make an issue about it? (The informal fork needs to be updated for go 1.17 & go 1.18 to not crash, and we can't merge it in anyway until a PR fixes all the issues)

I'm confused as to why the errcheck linter didn't catch the untested iterator error.

For the Event loop issue, lets change those functions to not have pointer receiver. The complexity of having to deref / use a different loop structure is not something I'd like us to pay.

Ditto for SetGaugeWithRefKey.

In both cases at the moment, theres no current issue, But good anyways to just pass the direct struct.

jim380:
@ValarDragon Regarding the event loop I actually think changing the function signature changes the semantics in a way that it hides the cost. 

If we were to change [`createBeginUnlockEvent`](https://github.com/osmosis-labs/osmosis/blob/main/x/lockup/keeper/msg_server.go#L149-L157) to not having a pointer receiver, then on [this line](https://github.com/osmosis-labs/osmosis/blob/main/x/lockup/keeper/msg_server.go#L113) we'd have to dereference `lock` anyway as most variables returned by the keeper methods are in fact pointers.

Same thing with [`SetGaugeWithRefKey`](https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/gauge.go#L76-L96). If we were to change it, then we'd have to either change the signature of  [CreateGaugeRefKeys](https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/gauge.go#L64-L74) too, or dereference `gauge` again [in the body](https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/gauge.go#L88-L94) of `SetGaugeWithRefKey`.

Either way I think it preserves the mechanics (how things work) at the cost of sacrificing the semantics (how things behave). As an alternative solution, in the latest commit I simply re-assign the the iterator variable inside the loop. This way we keep the semantics, avoiding a chain of reactions in other related methods, and mitigate the security risks. This is also what's typically recommended by Go's idiom.

app/export.go
```diff
@@ -172,9 +172,12 @@ func (app *OsmosisApp) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddr
 		counter++
 	}
 
-	iter.Close()
+	err := iter.Close()
+	if err != nil {
+		panic(err)
```
alexanderbez, line 177: nit: just to be consistent :) 


```suggestion
		log.Fatal(err)
```

Issue #1276, opened 2022-04-18
Stableswap: Createpool logic


ValarDragon:
We need to add CreatePool logic to the gamm keeper for stableswap. This has a few steps:
- [x] Implement the msg.CreatePool logic https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/pool-models/stableswap/msgs.go#L86-L88 , see Balancer logic for an example of how to do it
- [x] Add a new function to the message server, similar to https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/msg_server.go#L34-L37
- [x] Add stableswap to relevant register functions here; https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/module.go

alexanderbez:
@ValarDragon are you going to add a description to the issue? :)

ValarDragon:
Thanks for the reminder! This came up during a call today, forgot to come back and add descriptions

PR #1312, opened 2022-04-21
Stableswap: Createpool logic

AlpinYukseloglu:
Closes: #1276 

Add CreatePool logic to the gamm keeper for stableswap.

Comment for review: the Balancer `CreatePool` method has a number of set functions that adjust pools based on the weights of each asset. It seems like these methods aren't relevant for stableswap, but I have two concerns I think should be kept in mind during review:
1. Does not adding these functions cause the pool to not be set up properly (in case I missed something)?
2. All of them have checks/validate various parts of the `CreatePool` inputs that don't seem to be necessary in this case, but I want to make sure we don't accidentally remove guardrails that were necessary

(For reference) 
Balancer's `CreatePool`: https://github.com/osmosis-labs/osmosis/blob/03b13c960710242f48e49f3044890df3e3b848c3/x/gamm/pool-models/balancer/msgs.go#L98 Relevant checks in Balancer: https://github.com/osmosis-labs/osmosis/blob/03b13c960710242f48e49f3044890df3e3b848c3/x/gamm/pool-models/balancer/balancer_pool.go#L42-L56

## Brief change log

- Implement the msg.CreatePool logic https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/pool-models/stableswap/msgs.go#L86-L88 , see Balancer logic for an example of how to do it
- Add a new function to the message server, similar to https://github.com/osmosis-- labs/osmosis/blob/main/x/gamm/keeper/msg_server.go#L34-L37
- Add stableswap to relevant register functions here; https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/module.go

## Testing and Verifying

This change will add tests that will be listed here when complete


p0mvn:
I think this change deserves a changelog entry. What do you think @AlpinYukseloglu ?

x/gamm/keeper/msg_server.go
```diff
@@ -26,16 +27,28 @@ func NewBalancerMsgServerImpl(keeper *Keeper) balancer.MsgServer {
...
+		keeper: keeper,
+	}
+}
+
 var (
-	_ types.MsgServer    = msgServer{}
-	_ balancer.MsgServer = msgServer{}
+	_ types.MsgServer      = msgServer{}
+	_ balancer.MsgServer   = msgServer{}
+	_ stableswap.MsgServer = msgServer{}
```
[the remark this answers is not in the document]
ValarDragon: Why this change? The msgServer{} struct satisfies the relevant interfaces?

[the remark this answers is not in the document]
p0mvn: This is to avoid the allocation of `msgServer{}` by creating a struct. The suggested pattern is used to ensure that the struct implements the interface but doesn't do the allocation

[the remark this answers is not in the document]
ValarDragon: This doesn't even get compiled though, it's just a compile time type check. I don't think we should worry about optimizing compile times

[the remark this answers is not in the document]
ValarDragon: Golangs dead code elimination should prevent this from entering the binary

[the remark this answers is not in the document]
p0mvn: Yeah, I was wrongly assuming that only this pattern is right. I looked at: https://go.dev/doc/faq#guarantee_satisfies_interface which also suggests both. Thanks for pointing this out

x/gamm/pool-models/stableswap/msgs.go
```diff
@@ -83,6 +83,16 @@ func (msg MsgCreateStableswapPool) InitialLiquidity() sdk.Coins {
...
-func (msg MsgCreateStableswapPool) CreatePool(ctx sdk.Context, poolID uint64) (types.PoolI, error) {
-	return &Pool{}, types.ErrNotImplemented
+func (msg MsgCreateStableswapPool) CreatePool(ctx sdk.Context, poolId uint64) (types.PoolI, error) {
+	// We assume that FuturePoolGovernor is valid.
+	pool := &Pool{
+		Address:            types.NewPoolAddress(poolId).String(),
+		Id:                 poolId,
+		PoolParams:         *msg.PoolParams,
+		TotalShares:        sdk.NewCoin(types.GetPoolShareDenom(poolId), types.InitPoolSharesSupply),
+		PoolLiquidity:      nil,
```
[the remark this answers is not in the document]
alexanderbez: Good catch!

x/gamm/keeper/msg_server.go
```diff
@@ -26,16 +27,28 @@ func NewBalancerMsgServerImpl(keeper *Keeper) balancer.MsgServer {
...
 )
 
 func (server msgServer) CreateBalancerPool(goCtx context.Context, msg *balancer.MsgCreateBalancerPool) (*balancer.MsgCreateBalancerPoolResponse, error) {
 	poolId, err := server.CreatePool(goCtx, msg)
 	return &balancer.MsgCreateBalancerPoolResponse{PoolID: poolId}, err
 }
 
+func (server msgServer) CreateStableswapPool(goCtx context.Context, msg *stableswap.MsgCreateStableswapPool) (*stableswap.MsgCreateStableswapPoolResponse, error) {
+	poolId, err := server.CreatePool(goCtx, msg)
+	return &stableswap.MsgCreateStableswapPoolResponse{PoolID: poolId}, err
```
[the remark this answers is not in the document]
alexanderbez: Yes, this is preferable. Catch and handle errors early as possible ^^^

ValarDragon:
Highlighted concern areas don't seem applicable to me for Stableswap! Thanks for pointing them out.

Once the current PR comments are resolved, PR LGTM

x/gamm/pool-models/stableswap/msgs.go
```diff
@@ -83,6 +83,16 @@ func (msg MsgCreateStableswapPool) InitialLiquidity() sdk.Coins {
 	return msg.InitialPoolLiquidity
 }
 
-func (msg MsgCreateStableswapPool) CreatePool(ctx sdk.Context, poolID uint64) (types.PoolI, error) {
-	return &Pool{}, types.ErrNotImplemented
+func (msg MsgCreateStableswapPool) CreatePool(ctx sdk.Context, poolId uint64) (types.PoolI, error) {
+	// We assume that FuturePoolGovernor is valid.
+	pool := &Pool{
```
p0mvn, line 88: Just noticed that in balancer we have [`NewBalancerPool`](https://github.com/osmosis-labs/osmosis/blob/031a6b57e949f7d3e8684141e70020cc74c3aa60/x/gamm/pool-models/balancer/balancer_pool.go#L27) that is being called in its [`MsgCreateBalancerPool`](https://github.com/osmosis-labs/osmosis/blob/031a6b57e949f7d3e8684141e70020cc74c3aa60/x/gamm/pool-models/balancer/msgs.go#L98)

Do you think it makes sense to do something similar here?

p0mvn (approved):
LGTM

x/gamm/pool-models/stableswap/msgs.go
alexanderbez: Yeah we should have a constructor that sets fields to default values and takes a poolID as an arg. Good idea.

ValarDragon: Agreed, we should do a `NewStableswapPool` function

ValarDragon (approved):
LGTM once theres a NewStableswapPool function

x/gamm/pool-models/stableswap/msgs.go
AlpinYukseloglu: Agreed - added in recent commit!

AlpinYukseloglu:
Should be good to go! @ValarDragon @p0mvn @alexanderbez

x/gamm/pool-models/stableswap/msgs.go
```diff
@@ -84,15 +84,10 @@ func (msg MsgCreateStableswapPool) InitialLiquidity() sdk.Coins {
...
-	pool := &Pool{
-		Address:            types.NewPoolAddress(poolId).String(),
-		Id:                 poolId,
-		PoolParams:         *msg.PoolParams,
-		TotalShares:        sdk.NewCoin(types.GetPoolShareDenom(poolId), types.InitPoolSharesSupply),
-		PoolLiquidity:      msg.InitialPoolLiquidity,
-		FuturePoolGovernor: msg.FuturePoolGovernor,
+	stableswapPool, err := NewStableswapPool(poolId, *msg.PoolParams, msg.InitialPoolLiquidity, msg.FuturePoolGovernor, ctx.BlockTime())
+	if err != nil {
+		panic(err)
```
ValarDragon, line 89: This should return an error, not panic

x/gamm/pool-models/stableswap/stableswap_pool.go
```diff
@@ -0,0 +1,29 @@
...
+
+var _ types.PoolI = &Pool{}
+
+// NewStableswapPool returns a stableswap pool
+// Invariants that are assumed to be satisfied and not checked:
+// * len(initialLiquidity) = 2
+// * FutureGovernor is valid
+// * poolID doesn't already exist
+func NewStableswapPool(poolId uint64, stableswapPoolParams PoolParams, initialLiquidity sdk.Coins, futureGovernor string, blockTime time.Time) (Pool, error) {
+	pool := &Pool{
```
ValarDragon, line 19: lets remove the reference taken here, and dereference taken on on line 28?

x/gamm/pool-models/stableswap/stableswap_pool.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+)
+
+var _ types.PoolI = &Pool{}
+
+// NewStableswapPool returns a stableswap pool
+// Invariants that are assumed to be satisfied and not checked:
+// * len(initialLiquidity) = 2
+// * FutureGovernor is valid
```
ValarDragon, line 16: Is the idea that these two are checked by msg.ValidateBasic?

x/gamm/pool-models/stableswap/msgs.go
alexanderbez: bump @AlpinYukseloglu :)

alexanderbez (commented):
@AlpinYukseloglu I think there are 2-3 core comments from @ValarDragon that need to be addressed prior to merging this 👍

x/gamm/pool-models/stableswap/msgs.go
AlpinYukseloglu: Thanks for the bump @alexanderbez :) Should be fixed in next commit!

x/gamm/pool-models/stableswap/stableswap_pool.go
AlpinYukseloglu: Should be added in next commit!

x/gamm/pool-models/stableswap/stableswap_pool.go
AlpinYukseloglu: Yes, but it looks like the initial liquidity validation [isn't implemented yet](https://github.com/osmosis-labs/osmosis/blob/e9b9804bcef341a2b22fa90d7722a2fadb738d7e/x/gamm/pool-models/stableswap/msgs.go#L46) – do you think this something that should be added in this PR or in a separate issue @ValarDragon?

AlpinYukseloglu:
@alexanderbez Once @ValarDragon responds to my question in the last comment this we should hopefully be good to go!

x/gamm/pool-models/stableswap/stableswap_pool.go
ValarDragon: We can add that in a second PR

AlpinYukseloglu: Created an issue here: https://github.com/osmosis-labs/osmosis/issues/1364

Issue #1089, opened 2022-03-14
Pool-params query is broken


UnityChaos:
`pool-params` query seems to be returning raw byte string, instead of structured result.
```
osmosisd q gamm pool-params 1
typeurl: /osmosis.gamm.v1beta1.PoolParams
value:
- 10
- 16
- 51
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 48
- 18
- 1
- 48
xxx_nounkeyedliteral: {}
xxx_unrecognized: []
xxx_sizecache: 0
```
```
osmosisd q gamm pool-params 1 --output=json
Error: unable to resolve type URL /osmosis.gamm.v1beta1.PoolParams
```

PR #1315, opened 2022-04-21
fix: pool params query

nghuyenthevinh2000:
Closes: #1089 

The code here can solve issues addressed at: https://github.com/osmosis-labs/osmosis/issues/1089

Here is a picture to prove it

Screenshot from 2022-04-22 01-20-06

*Add a description of the overall background and high level changes that this PR introduces *

## Brief change log

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


x/gamm/pool-models/balancer/codec.go
```diff
@@ -26,6 +26,7 @@ func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 	registry.RegisterImplementations(
 		(*sdk.Msg)(nil),
 		&MsgCreateBalancerPool{},
+		&PoolParams{},
```
ValarDragon, line 29: PoolParams doesnt implement the message interface though?

x/gamm/client/cli/query.go
```diff
@@ -13,6 +13,8 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+
```
alexanderbez, line 16:
```suggestion
```

x/gamm/pool-models/balancer/msgs.go
```diff
@@ -98,3 +99,22 @@ func (msg MsgCreateBalancerPool) CreatePool(ctx sdk.Context, poolID uint64) (typ
 	poolI, err := NewBalancerPool(poolID, *msg.PoolParams, msg.PoolAssets, msg.FuturePoolGovernor, ctx.BlockTime())
 	return &poolI, err
 }
+
+var _ sdk.Msg = &PoolParams{}
```
alexanderbez, line 103: Why are we making this change?

ValarDragon:
The approach of this PR is making pool params satisfy the message interface. What should likely instead be done is just registering this struct onto the proto codec directly.

x/gamm/pool-models/balancer/codec.go
nghuyenthevinh2000: yes, I have added PoolParams directly to codec like your suggestion

mattverse: I think what @ValarDragon was saying was that `PoolParams` does not follow implementations of proto.Message. 

Structs should be implemented upon specific interfaces via `RegisterImplementations`, @ValarDragon  do you think we should make an interface for pool params?

ValarDragon:
So the problem being faced here, is that the PoolParams is getting serialized and then deserialized as a types.Any: https://github.com/osmosis-labs/osmosis/blob/6f50a357b2f0c441f64e61cd2f4da9066e1af1e0/x/gamm/types/query.pb.go#L351-L353

This means that we need to register it as implementing any interface. I think it suffices to say the struct implements `interface{}` rather than sdk.Msg. Can you try that and see if it fixes the issue?

Also btw, `coverage.txt` needs to be deleted

mattverse:
I think a better approach is creating a pool params interface rather than using a emtpy interface wdyt?

ValarDragon:
Why would we have an interface? The problem here is the codec doesn't know about this route, and it can learn about the route by registering it as an implementation of any interface, such as `interface{}`

nghuyenthevinh2000:
wouldn't it be better to use `proto.Message` interface rather than just `interface{}` cause `PoolParams{}` already satisfied `proto.Message` by default when it was first generated.

`registry.RegisterImplementations(
		(*proto.Message)(nil),
		&PoolParams{},
	)
`

x/gamm/client/cli/query.go
```diff
@@ -216,7 +217,10 @@ $ %s query gamm pool-params 1
 			}
 
 			if clientCtx.OutputFormat == "text" {
-				out, err := yaml.Marshal(res.GetParams())
+				poolParams := &balancer.PoolParams{}
+				poolParams.Unmarshal(res.GetParams().Value)
```
alexanderbez, line 221: we need to check the error here.

ValarDragon:
oh good point, using `proto.Message` as the interface makes sense as well.

mattverse (approved):
Awesome! LGTM!

p0mvn:
Hi @nghuyenthevinh2000 . Would you mind updating the PR name to something like "fix: pool params query", please? I'm requesting this so that it is easier to follow in Github history, once merged. Thank you!

nghuyenthevinh2000:
I have renamed as requested

x/gamm/client/cli/query.go
```diff
@@ -216,7 +217,13 @@ $ %s query gamm pool-params 1
 			}
 
 			if clientCtx.OutputFormat == "text" {
-				out, err := yaml.Marshal(res.GetParams())
+				poolParams := &balancer.PoolParams{}
+				err := poolParams.Unmarshal(res.GetParams().Value)
+				if err != nil {
```
p0mvn, line 222:
```suggestion
				if err := poolParams.Unmarshal(res.GetParams().Value); err != nil {
```

p0mvn (approved):

PR #1321, opened 2022-04-22
Move forceprune logic into functions, make db open/closes all in a single scope

ValarDragon:

Follow up PR to #1262, it shuffles the logic into functions, so each db opening and closing all happens in a single scope. (And therefore defer closes work as expected)

Beyond DB closing behavior, the logic should be identical 

cc @moniya12

## Brief change log

No logic change, just code structure updates to the forceprune command. Also updates variables names from snake case to camel case (matching the rest of the repo)

## Testing and Verifying

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no (not needed)


cmd/osmosisd/cmd/forceprune.go
```diff
@@ -69,92 +69,122 @@ func forceprune() *cobra.Command {
...
+	if err = db.CompactRange(*util.BytesPrefix([]byte{})); err != nil {
+		return err
+	}
+	return nil
+}
+
+func forcepruneStateStore(dbPath string, startHeight, currentHeight, minHeight, fullHeight int64) error {
+	opts := opt.Options{
+		DisableSeeksCompaction: true,
+	}
```
ValarDragon, line 138: I chose to duplicate this in each function, since its pretty unclear to me that longer term, these won't be different for every function

p0mvn (approved):

Issue #504, opened 2021-10-05
Fix make proto-gen when you have a vendor folder


ValarDragon:
`make proto-gen` fails if you use a `vendor` folder. @robert-zaremba pointed out in https://github.com/osmosis-labs/osmosis/pull/496#issuecomment-934709990 that the proto script searches for all `.proto` files and attempts to compile them. This is done in this line https://github.com/osmosis-labs/osmosis/blob/main/scripts/protocgen.sh#L12

When I tried to add a `--prune "*/vendor/*"` or `-not -path "*/vendor/*"` as an argument to that find command, the problem still persisted. Namely I get:
```
make proto-gen
Generating Protobuf files
/third_party/proto: does not exist
make: *** [proto-gen] Error 1
```

Not sure how we fix this. (Its possible I did the ignoring wrong!)

When fixed, we should upstream to the SDK.

robert-zaremba:
In fact, the find command searches only in the `./proto` director (and it's subdirectories):

    find ./proto -path -prune -o -name '*.proto' -print0

The issues sounds like something is importing `/third_party/proto` but can't find it.

PR #1326, opened 2022-04-22
make proto gen scripts compatible with vendor

catShaark:
Closes: #504

So these 3 scripts for generating protobuf ( [generate-proto.sh](https://github.com/osmosis-labs/osmosis/blob/main/scripts/generate-proto.sh),
[generate-docs.sh](https://github.com/osmosis-labs/osmosis/blob/main/scripts/generate-docs.sh), [protocgen.sh](https://github.com/osmosis-labs/osmosis/blob/main/scripts/protocgen.sh) ) use [go list](https://github.com/osmosis-labs/osmosis/blob/039f686ed40008228db65352ad09d1f38a8f88cb/scripts/protocgen.sh#L11) to get the `cosmos-sdk` dependency in go/pkg/mod. But `go list` doesn't work properly if we have a vendor folder, so I move the vendor folder to a temp folder to run `go list` and then move it back right after running `go list`.

### DIFF
--- a/scripts/generate-docs.sh
+++ b/scripts/generate-docs.sh
@@ -4,8 +4,20 @@ set -eo pipefail
 
 mkdir -p ./tmp-swagger-gen
 
+# move the vendor folder to a temp dir so that go list works properly
+temp_dir="f29ea6aa861dc4b083e8e48f67cce"
+if [ -d vendor ]; then
+  mv ./vendor ./$temp_dir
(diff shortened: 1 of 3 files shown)


ValarDragon (approved):
Awesome, LGTM! Thanks for adding this, saves me a lot of vendor directory nuking & redownloading =)

Issue #1319, opened 2022-04-22
Move appKeepers struct to a different package


ValarDragon:
# Summary

The `appKeepers` struct in `app/keepers.go` is quite useful. However, it would be even better if we could have it be defined in its own package that other modules could import. Right now the upgrades package can't import this due to import cycles, and this is blocking standardization / simplification of many of the upgrade logic API's.

# Acceptance Criteria

- [ ] New package called `keepers` at `app/keepers`, and the logic from the Osmosis `app/keepers.go` being moved to that package, and functions being correctly exported.

PR #1327, opened 2022-04-22
Move appKeepers struct to a different package

nguyenvuong1122000:
Closes: #1319 

 - Move appKeepers struct to a different package

## Brief change log
 - Create new package called `keepers` at `app/keepers`,
 - Move logic from `app/keepers.go` to new module

## Testing and Verifying

 * This change is a trivial rework / code cleanup without any test coverage.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -158,6 +158,7 @@ func NewOsmosisApp(
 	bApp.SetInterfaceRegistry(interfaceRegistry)
 
 	app := &OsmosisApp{
+		AppKeepers:        keepers.AppKeepers{},
 		BaseApp:           bApp,
 		cdc:               cdc,
 		appCodec:          appCodec,
--- a/app/app.go
+++ b/app/app.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/CosmWasm/wasmd/x/wasm"
-	transfer "github.com/cosmos/ibc-go/v2/modules/apps/transfer"
 	"github.com/gorilla/mux"
 	"github.com/rakyll/statik/fs"
 	"github.com/spf13/cast"
@@ -37,7 +36,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
 	authrest "github.com/cosmos/cosmos-sdk/x/auth/client/rest"
 	authtx "github.com/cosmos/cosmos-sdk/x/auth/tx"
-	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 	"github.com/cosmos/cosmos-sdk/x/crisis"
 	paramstypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
@@ -121,15 +119,9 @@ type OsmosisApp struct {
 	interfaceRegistry types.InterfaceRegistry
 	invCheckPeriod    uint
 
-	// keys to access the substores
-	keys    map[string]*sdk.KVStoreKey
-	tkeys   map[string]*sdk.TransientStoreKey
-	memKeys map[string]*sdk.MemoryStoreKey
-
-	transferModule transfer.AppModule
-	mm             *module.Manager
-	sm             *module.SimulationManager
-	configurator   module.Configurator
+	mm           *module.Manager
+	sm           *module.SimulationManager
+	configurator module.Configurator
 }
 
 func init() {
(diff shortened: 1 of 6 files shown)


ValarDragon:
Thanks for making this!

app/app.go
```diff
@@ -192,9 +193,33 @@ func NewOsmosisApp(
 		panic(fmt.Sprintf("error while reading wasm config: %s", err))
 	}
 
-	app.InitSpecialKeepers(skipUpgradeHeights, homePath, invCheckPeriod)
+	app.InitSpecialKeepers(
+		appCodec,
+		bApp,
+		keys,
+		tkeys,
+		memKeys,
```
ValarDragon, line 201: We should be able to generate these in the keepers package right? Perhaps we have a new file in there called storekeys, which has functions for these?

nguyenvuong1122000: yes, i think generate these keys in keepers package is better.

nguyenvuong1122000: i think  transfer module should be in keepers package instead of what i did.

app/keepers/keepers.go
```diff
@@ -103,384 +103,401 @@ type appKeepers struct {
...
-	app.ParamsKeeper = &paramsKeeper
+	// transfer module
+	TransferModule transfer.AppModule
 
-	// set the BaseApp's parameter store
-	bApp.SetParamStore(app.ParamsKeeper.Subspace(baseapp.Paramspace).WithKeyTable(paramskeeper.ConsensusParamsKeyTable()))
+	// keys to access the substores
+	keys    map[string]*sdk.KVStoreKey
+	tkeys   map[string]*sdk.TransientStoreKey
+	memKeys map[string]*sdk.MemoryStoreKey
```
ValarDragon, line 112: Oh interesting, right we do still need the map stored. Good call on moving it to this package!

app/keepers/keepers.go
```diff
@@ -508,3 +525,41 @@ func KVStoreKeys() []string {
...
+func (appKeepers *AppKeepers) GetTKey(storeKey string) *sdk.TransientStoreKey {
+	return appKeepers.tkeys[storeKey]
+}
+
+// GetMemKey returns the MemStoreKey for the provided mem key.
+//
+// NOTE: This is solely used for testing purposes.
+func (appKeepers *AppKeepers) GetMemKey(storeKey string) *sdk.MemoryStoreKey {
+	return appKeepers.memKeys[storeKey]
+}
```
ValarDragon, line 565: Perhaps we move this to another file, e.g. `getters.go` or `keys.go` ?

(If going with `keys.go`, we should put the key initialization there as well)

nguyenvuong1122000: I moved these functions to `keys.go`. Thanks for reviewing !

app/app.go
```diff
@@ -164,37 +157,40 @@ func NewOsmosisApp(
...
-	keys := sdk.NewKVStoreKeys(KVStoreKeys()...)
-
-	// Define transient store keys
-	tkeys := sdk.NewTransientStoreKeys(paramstypes.TStoreKey)
-
-	// MemKeys are for information that is stored only in RAM.
-	memKeys := sdk.NewMemoryStoreKeys(capabilitytypes.MemStoreKey)
-
 	app := &OsmosisApp{
 		BaseApp:           bApp,
```
ValarDragon, line 161: We should similarly add an AppKeepers empty struct initialization here right?

app/keepers/keepers.go
```diff
@@ -103,384 +103,388 @@ type appKeepers struct {
...
+func (appKeepers *AppKeepers) InitSpecialKeepers(
+	appCodec codec.Codec,
+	bApp *baseapp.BaseApp,
+	wasmDir string,
+	cdc *codec.LegacyAmino,
+	invCheckPeriod uint,
+	skipUpgradeHeights map[int64]bool,
+	homePath string,
+) {
+	appKeepers.GenerateKeys()
```
ValarDragon, line 376: Perhaps we should have an appKeepers.Init() function, and we do this Generate Keys call in there? (Related to comment about initializing the struct in app.go)

ValarDragon (approved):
LGTM! Nice job on this!

I checked each of the replaces / lines, and it all looked good to me! (Aside from the initailization question for the OsmosisApp struct)

app/app.go
nguyenvuong1122000: yes, i added it in osmosis app initialization

go.mod
```diff
@@ -224,7 +224,7 @@ require (
 	github.com/stretchr/objx v0.3.0 // indirect
 	github.com/subosito/gotenv v1.2.0 // indirect
 	github.com/sylvia7788/contextcheck v1.0.4 // indirect
-	github.com/syndtr/goleveldb v1.0.1-0.20200815110645-5c35d600f0ca // indirect
+	github.com/syndtr/goleveldb v1.0.1-0.20200815110645-5c35d600f0ca
```
mattverse, line 227: What's the reason behind changing goleveldb from inderect -> direct import?

nguyenvuong1122000: I think this change not from my PR. `go.mod` have same change in main branch when i run `go mod tidy`.

Issue #1341, opened 2022-04-25
[stableswap]: Switch logic to binary search


AlpinYukseloglu:
## Background

Once our binary search solver implementations are merged (#2816 and #2825), we should point our core solver functions (`solveCfmm` and `solveCfmmMulti`) to them to complete the loop on the core CFMM work for stableswap. Further cleanup to tie up loose ends is described in #2826

## Suggested Design

- Replace bodies of `solveCfmm` and `solveCfmmMulti` to use their respective binary search solvers once those are merged

## Acceptance Criteria

- All two-asset and multi-asset tests pass

AlpinYukseloglu:
Closed by #2845

PR #1342, opened 2022-04-25
Switch cfmm logic to binary search

AlpinYukseloglu:
Closes: #1341 

Change the `solveCfmm` function to call `solveCFMMBinarySearch` and update the original one to be called `solveCfmmDirectEquation`

## Brief change log
 
  - *Change the `solveCfmm` function to call `solveCFMMBinarySearch`*
  - *Change the original `solveCfmm` to `solveCfmmDirectEquation`*

## Testing and Verifying

This change is already covered by existing tests, such as *(existing CFMM tests since the changes shouldn't affect how any other code functions)*.


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -168,6 +168,10 @@ func solveCFMMBinarySearch(constantFunction func(sdk.Dec, sdk.Dec) sdk.Dec) func
 	}
 }
 
+func solveCfmm(xReserve, yReserve, yIn sdk.Dec) sdk.Dec {
+	return solveCFMMBinarySearch(cfmmConstant)(xReserve, yReserve, yIn)
+}
```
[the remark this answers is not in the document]
AlpinYukseloglu: Bump @ValarDragon :)

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -180,6 +188,10 @@ func solveCFMMBinarySearch(constantFunction func(sdk.Dec, sdk.Dec) sdk.Dec) func
 	}
 }
 
+func solveCfmm(xReserve, yReserve, yIn sdk.Dec) sdk.Dec {
```
p0mvn, line 191: Why do we need this wrapper and not use `solveCFMMBinarySearch` directly?

mattverse: Same question here! Maybe we have plans to have different implementations in the future?

mattverse (commented):
Apart from the question of why we need a secondary wrapper `solveCfmm` instead of having to use `solveCFMMBinarySearch` directly, LGTM!

x/gamm/pool-models/stableswap/amm.go
ValarDragon: so we can easily swap the backend implementation, without replacing the function everywhere else

mattverse: I'm not sure I follow, is it because of the time dependencies and other priorities over replacing `solveCfmm` implementation everywhere?

mattverse: @AlpinYukseloglu

AlpinYukseloglu: The idea was to have a general solveCfmm function that abstracts away the specific solving methodology in case we change it in the future. It also felt cleaner to keep it as a general solver function and keep it as the only one instead of having two functions floating around with only one actually used (we would still need to keep the original solver for testing reasons.

mattverse (changes requested):
So it seems that we changed from the direct equation in solving stableswap to using binary search, which is causing the test to fail. 

We should fix this before merging || create a new branch to fix this and then come back to this PR

AlpinYukseloglu:
> So it seems that we changed from the direct equation in solving stableswap to using binary search, which is causing the test to fail. 
> 
> 
> 
> We should fix this before merging || create a new branch to fix this and then come back to this PR

Yeah this is interesting/weird - looking into it right now!

AlpinYukseloglu:
The main issue left to investigate in this PR is a rounding error that surfaces with larger `k` values

In any case, tabling this PR until we pick stableswap work back up

PR #1345, opened 2022-04-25
Change Pool interface calc methods to using integers

ValarDragon:
Solves this comment: https://github.com/osmosis-labs/osmosis/pull/1287#discussion_r856316806

The purpose of this change is two-fold:
* Fixes a logic bug where TokenInAmounts were rounded down, instead of rounded up
* Simplifies rounding logic & API, by making pool interface calc methods return sdk.Coin. We removed several ad-hoc rounding truncations (e.g. in multi-hop) 

## Brief change log
 
* Fixes logic bug where TokenInAmounts were rounded down, instead of rounded up
* Makes pool interface use sdk.Coin return type for Calc methods

## Testing and Verifying

This change is already covered by existing tests, see the existing test updates

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no, only in gamm update spec
  - How is the feature or change documented? Needs docs for GAMM to append this to


ValarDragon:
The test failures are blocked on a different bug in GAMM, I'm addressing that one in a separate PR

ValarDragon:
Ah rip, its not a simple change if we care about convergence bounds (we do) cref #1346

p0mvn (commented):
Sorry if I'm early with comments - was working on the same part of the code and ended up looking at this change

x/gamm/pool-models/balancer/amm.go
```diff
@@ -67,7 +67,12 @@ func (p Pool) CalcOutAmtGivenIn(
 		poolAssetOut.Weight.ToDec(),
 	)
```
mattverse, line 69:
```suggestion

	// we ignore the decimal component, as token out amount must round down
```

x/gamm/keeper/swap_test.go
```diff
@@ -125,7 +125,7 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleSwapExactAmountOut() {
 				tokenInDenom:          "foo",
 				tokenInMaxAmount:      sdk.NewInt(900000000),
 				tokenOut:              sdk.NewCoin("bar", sdk.NewInt(100000)),
-				expectedTokenInAmount: sdk.NewInt(206164),
+				expectedTokenInAmount: sdk.NewInt(206165),
```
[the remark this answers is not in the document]
mattverse: Let's address this in https://github.com/osmosis-labs/osmosis/issues/1349

mattverse (approved):
LGTM! Feels to me like test case improvements should be handled in a separate PR since we have a lot of other missing tests for stableswap rn.

x/gamm/pool-models/balancer/amm.go
```diff
@@ -119,27 +120,24 @@ func (p Pool) CalcInAmtGivenOut(
 	// Thus in order to give X amount out, we solve the invariant for the invariant input. However invariant input = (1 - swapfee) * trade input.
 	// Therefore we divide by (1 - swapfee) here
 	tokenAmountInBeforeFee := tokenAmountIn.Quo(sdk.OneDec().Sub(swapFee))
-	// TODO: Once we make Calc methods return integers
+
 	// if tokenInDecimal is non-zero, we add 1 to the tokenInCoin
```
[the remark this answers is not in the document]
p0mvn: I think `// if tokenInDecimal is non-zero, we add 1 to the tokenInCoin` doesn't apply anymore too

ValarDragon:
All comments / tests resolved. Please check out the last few commits. basically what changed is:

- For cosmwasm message tests, updated the values so they don't run into the half the pool being swapped out problem. (Problem not new to this PR, just the boundary condition of exactly 50% swapping out, now gets rounded up and causes the issue)
- For cosmwasm query tests, update the error tolerance
- For gamm keeper inverse calc in <> calc out consistency tests, added an error tolerance, and added another test case for the failing size, that has much higher values. The latter convinces us that this was a rounding error, not a constant fraction error.
- Fixing the type in hooks_test, for the calc method.

ValarDragon:
All comments / tests resolved. Please check out the last few commits. basically what changed is:

- For cosmwasm message tests, updated the values so they don't run into the half the pool being swapped out problem. (Problem not new to this PR, just the boundary condition of exactly 50% swapping out, now gets rounded up and causes the issue)
- For cosmwasm query tests, update the error tolerance
- For gamm keeper inverse calc in <> calc out consistency tests, added an error tolerance, and added another test case for the failing size, that has much higher values. The latter convinces us that this was a rounding error, not a constant fraction error.
- Fixing the type in hooks_test, for the calc method.

osmoutils/dec_helper.go
```diff
@@ -0,0 +1,14 @@
+package osmoutils
+
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// intended to be used with require/assert:  require.True(DecEq(...))
+// TODO: Replace with function in SDK types package when we update
```
alexanderbez, line 10: RIP

x/gamm/keeper/multihop.go
```diff
@@ -79,14 +79,14 @@ func (k Keeper) createMultihopExpectedSwapOuts(ctx sdk.Context, routes []types.S
 			return nil, err
 		}
 
+		// fmt.Printf("calc in given out %s, pool liquidity %s\n", sdk.NewCoins(tokenOut), pool.GetTotalPoolLiquidity(ctx))
```
alexanderbez, line 82:
```suggestion
```

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -84,59 +84,60 @@ func (p *Pool) updatePoolLiquidityForSwap(tokensIn sdk.Coins, tokensOut sdk.Coin
...
+		return sdk.Coin{}, err
+	}
+
+	// we ignore the decimal component, as token out amount must round down
+	tokenOutAmt := outAmtDec.TruncateInt()
+	if !tokenOutAmt.IsPositive() {
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
 	}
-	return sdk.DecCoin{Denom: tokenOutDenom, Amount: amt}, nil
+	return sdk.Coin{Denom: tokenOutDenom, Amount: tokenOutAmt}, nil
```
alexanderbez, line 101: Can we use `NewCoin` if possible please?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -84,59 +84,60 @@ func (p *Pool) updatePoolLiquidityForSwap(tokensIn sdk.Coins, tokensOut sdk.Coin
...
 	}
 	// TODO: Refactor this later to handle scaling factors
 	amt, err := pa.calcInAmtGivenOut(tokenOut[0], tokenInDenom, swapFee)
 	if err != nil {
-		return sdk.DecCoin{}, err
+		return sdk.Coin{}, err
+	}
+
+	// We round up tokenInAmt, as this is whats charged for the swap, for the precise amount out.
+	// (else the pool would under-charge by this rounding error)
```
alexanderbez, line 126:
```suggestion
	// Otherwise, the pool would under-charge by this rounding error.
```

alexanderbez (commented):
🚀

osmoutils/dec_helper.go
ValarDragon: This made me sad as well :(

x/gamm/keeper/multihop.go
ValarDragon: Thank you! Forgot to delete that debug code

alexanderbez (approved):
🌮 👩‍🎤

PR #1365, opened 2022-04-28
Partial ordering for endblock

ValarDragon:

This pull request introduces a partial ordering for EndBlocker logic. The first commit indicates every module that is a no-op in the EndBlocker. 

## Brief change log

* Utilize partial orderings for EndBlocker orderings

## Testing and Verifying

* If all logic is correct, all ordering should work as before.
  * The first commit indicates every module whose end blocker code is a no-op, and proposes the new ordering. Please double check this
  * The second commit deletes the boilerplate old code, and integrates into the SetOrderEndBlocker logic

  - Does this pull request introduce a new feature or user-facing behavior changes? no, purely developer UX
  - How is the feature or change documented? It is utilization of code structure alteration, meant to help with mitigating SDK overhead


app/keepers/modules.go
```diff
@@ -0,0 +1,84 @@
+package keepers
```
alexanderbez, line 1: What's the motivation for moving this to a dedicated keepers package instead of `app`?

ValarDragon: The goal is to get more of the keeper/module scaffolding outside of the app package, so its importable by other packages. Maybe this doesn't make sense though.

The `AppKeepers` struct was needed in its own package, to fix API complexity problems in our upgrades registration code. (We just want to give it access to all keepers it can modify, while having each upgrade be its own package)

app/modules.go
```diff
@@ -220,37 +170,49 @@ func orderBeginBlockers() []string {
...
+	return ord.TotalOrdering()
+}
+
 // orderEndBlockers Tell the app's module manager how to set the order of
 // EndBlockers, which are run at the end of every block.
 var orderEndBlockers = []string{
-	lockuptypes.ModuleName,
+	lockuptypes.ModuleName, // no-op
 	crisistypes.ModuleName,
 	govtypes.ModuleName,
```
p0mvn: What about these 2?

app/modules.go
```diff
@@ -220,37 +170,49 @@ func orderBeginBlockers() []string {
...
-	bech32ibctypes.ModuleName,
-	txfeestypes.ModuleName,
+	stakingtypes.ModuleName,        // no-op
+	capabilitytypes.ModuleName,     // no-op
+	authtypes.ModuleName,           // no-op
+	banktypes.ModuleName,           // no-op
+	distrtypes.ModuleName,          // no-op
+	slashingtypes.ModuleName,       // no-op
+	minttypes.ModuleName,           // no-op
+	genutiltypes.ModuleName,        // (can't tell if no-op)
```
p0mvn: Did you confirm that this one is a no-op? Just making sure it wasn't missed

alexanderbez (approved):
Super helpful. Nice work @ValarDragon

app/modules.go
ValarDragon: No, they do work. However, their ordering does not matter as far as I know

app/modules.go
ValarDragon: Just found it: https://github.com/cosmos/cosmos-sdk/blob/v0.45.4/x/genutil/module.go#L85

I missed finding it on my first pass, because I missed that line

Issue #1223, opened 2022-04-07
x/incentives: Rename BeginDistribution and FinishDistribution


ValarDragon:
We should rename `k.BeginDistribution(ctx, gauge)` to `k.moveUpcomingGaugeToActiveGauge`, and `k.FinishDistribution(ctx, gauge)` to `k.moveActiveGaugeToFinishedGauge(ctx, gauge)`

These are really internal methods, that shouldn't be called externally, that just update the 'state' of the gauge.

PR #1366, opened 2022-04-28
Closing #1223 - BeginDistribution / FinishDistribution Changes

xBalbinus:
Closes: #1223 

## Purpose of the change

Renames k.BeginDistribution(ctx, gauge) to k.moveUpcomingGaugeToActiveGauge, and k.FinishDistribution(ctx, gauge) to k.moveActiveGaugeToFinishedGauge(ctx, gauge)

## Brief change log

- Added function definitions to export_test.go
- Renamed the original functions
- Fixed function calls on tests to match the names of the current functions


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/CosmWasm/wasmd v0.24.0
-	github.com/cosmos/cosmos-sdk v0.45.3
+	github.com/cosmos/cosmos-sdk v0.45.4
```
p0mvn, line 7: What is the reason for this? Is this necessary?

We usually upgrade our SDK fork in the replace directive here:
https://github.com/osmosis-labs/osmosis/blob/d8918936a898bcc982147f8e42d096a0f9a473db/go.mod#L271

alexanderbez (approved):

docs/core/proto-docs.md
```diff
@@ -4,31 +4,6 @@
 
 ## Table of Contents
 
-- [osmosis/claim/v1beta1/claim.proto](#osmosis/claim/v1beta1/claim.proto)
```
p0mvn, line 7: I don't see any protos updated in this PR. Why is this changed?

go.mod
xBalbinus: I'll change this back - I think I got this as a part of pulling from main before making these changes but I could be wrong

PR #1382, opened 2022-05-01
add voting period to init image

czarcas7ic:
This PR needs to be merged in order to create the docker image that will be served on docker hub to initialize the v7 genesis/configs for the upgrade testing. The current image I uploaded [here](https://hub.docker.com/layers/205814749/osmolabs/osmosis-init/v7.2.1/images/sha256-4f1aa2797d06dbf72b235f8a9beb20de46cc179041d39711c5c0b76f24abdb5c?context=repo) does not change the voting period, so I needed to make this change and then will replace this docker image.


czarcas7ic:
Actually, I realized that I can just push this image without it being merged. I will instead do that and then backport everything to v7.x once complete.

Issue #1333, opened 2022-04-23
[E2E - Upgrade Testing] logic for testing upgrade process


czarcas7ic:
This issue will lay out initial thoughts on what the upgrade process must entail to complete issue #1235  based on discussion with @p0mvn 

- [x] Ensure version X+1's [upgrades.go and constants.go](https://github.com/osmosis-labs/osmosis/tree/main/app/upgrades) exist in the current branch
- [x] Figure out how long gov submission / passing takes
  - [x] Ensure we modify genesis to have fast voting periods
  - [x] Transfer this time to number of blocks, then add some for a buffer. This will be our upgrade height
- [x] Chain will be initialized on current stable prod release (version X)
- [x] Start chain with halt height = upgrade height
- [x] Gov logic executes
  - [x] Prop submission
  - [x] Prop fund
  - [x] Prop vote
  - [x] Prop passes
- [x] Once daemon stops due to halt height, start new docker container using osmosis:debug (locally compiled docker image based off main). 
  - [x] Ensure volume used on version X is mounted to version X+1
 
From here, we can run tests to ensure nothing broke.

alexanderbez:
Awesome, thanks for writing this up @czarcas7ic.

Question though, why do we need this step:

>  Start chain with halt height = upgrade height

czarcas7ic:
@alexanderbez  From what I remember, the chain doesn't actually halt on upgrade height right? It displays a message saying the that the new binary is needed and will just sit idly until the daemon is stopped and the binary is changed (I could be wrong and it could panic and halt). When I thought up this process, it just seemed simpler to use the halt height flag since the upgrade height must be hardcoded via the gov prop anyway and wont change. I'm sure there are other ways to do this, this was just the first one that came to mind.

alexanderbez:
@czarcas7ic you might be right. Either way, it can't hurt to add a halt-height configuration, so I'm all for it 👍

czarcas7ic:
Just noting here that the chain does in fact not panic, you get a log like the following
```
6:03PM ERR UPGRADE "v8" NEEDED at height: 100: 
6:03PM ERR CONSENSUS FAILURE!!! err="UPGRADE \"v8\" NEEDED at height: 100:
```
and then it goes back to searching for peers. No issues with this, just noting for future reference.

alexanderbez:
Well that is a panic, but the node does not _halt_. So yes, I guess we need to add that config 👍

p0mvn:
[...]

PR #1389, opened 2022-05-02
test: upgrade testing automation

czarcas7ic:
Closes: #1333 , #1235

This pull request follows the methodology laid out in #1333 to complete the minimum viable product needed for #1235 

## Process
  - Modified the v7 init docker image to reduce the vote time to 30 seconds
  - Checks out a v7 init docker image from dockerhub and initializes the genesis and config files
  - Checks out a v7 osmosis image from dockerhub and utilizes the initialized genesis and config to spin up two separate chains, along with a hermes image that connects the two chains via IBC
  - Submits a v8 upgrade prop, funds, and votes for it on both chains
  - Ensures upgrade height is reached and then no longer hits blocks
  - Stops all containers and replaces them with the locally compiled v8 image
  - Ensures blocks are hit on both chains
  - Proceeds with remaining tests

## Testing and Verifying

This change is already covered by existing e2e test suite. This was tested locally and through github CI

[Here is an example of the test output](https://github.com/osmosis-labs/osmosis/runs/6265997089?check_suite_focus=true#step:6:125)

  - How is the feature or change documented? none

## Notes / Next Steps

From here, we can create an environment variable that allows us to skip upgrade testing and instead only compile locally and go though the chain tests. This will save us time instead of having to go through the upgrade process for every commit. 

Another important note, I [removed the locally compile osmosis-init-image step](https://github.com/osmosis-labs/osmosis/commit/f142e1af7363daa2ec01184c617f2256384fd4b1) here since it is not needed at the moment. When we introduce the environment variable addressed above, we will need to re-add this to compile the v8 init image instead of compiling with the v7 init image from docker hub. We should split these into two separate make tests so that way we are only compiling the docker images we absolutely need. The full upgrade testing takes only 8 minutes though.

Another note, I need to backport the config changes to v7.x that changes the gov voting period

Last note, we need to [change the test.yml](https://github.com/osmosis-labs/osmosis/commit/2305866480bc45383fd520f4b08b0f04230bbacf) because if you can make a commit that does not pass the e2e-test at first, but if you do a second commit that does not trigger this `enc.GIT_DIFF` and it will skip the test despite actually not passing it. cc @daniel-farina if you know how to go about fixing this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,7 @@ benchmark:
 	@go test -mod=readonly -bench=. $(PACKAGES_UNIT)
 
 docker-build-debug:
-	@docker build -t osmosis:debug --build-arg BASE_IMG_TAG=debug -f Dockerfile .
+	@docker build -t osmosis:debug --build-arg BASE_IMG_TAG=debug -f tests/e2e/Dockerfile .
 
 docker-build-e2e-chain-init:
 	@docker build -t osmosis-e2e-chain-init:debug -f tests/e2e/chain_init/chain-init.Dockerfile .
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ test-sim:
 	@VERSION=$(VERSION) go test -mod=readonly $(PACKAGES_SIM)
 
 test-e2e:
-	@VERSION=$(VERSION) go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
+	@VERSION=$(VERSION) go test -mod=readonly -timeout=25m -v github.com/osmosis-labs/osmosis/v7/tests/e2e
 
 benchmark:
 	@go test -mod=readonly -bench=. $(PACKAGES_UNIT)
--- a/tests/e2e/Dockerfile
+++ b/tests/e2e/Dockerfile
@@ -15,7 +15,7 @@ RUN sha256sum /lib/libwasmvm_muslc.a | grep d0152067a5609bfdfb3f0d5d6c0f2760f79d
 RUN BUILD_TAGS=muslc make build
 
 ## Deploy image
-FROM gcr.io/distroless/base-debian11:${BASE_IMG_TAG}
+FROM ubuntu
 
 COPY --from=build /osmosis/build/osmosisd /bin/osmosisd
 
--- a/tests/e2e/chain/config.go
+++ b/tests/e2e/chain/config.go
@@ -13,6 +13,7 @@ import (
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/cosmos/cosmos-sdk/x/genutil"
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/spf13/viper"
 	tmconfig "github.com/tendermint/tendermint/config"
 	tmjson "github.com/tendermint/tendermint/libs/json"
@@ -27,6 +28,7 @@ const (
 	IbcDenom      = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
 	MinGasPrice   = "0.000"
 	IbcSendAmount = 3300000000
+	VotingPeriod  = 180000000000
 	// chainA
 	ChainAID      = "osmo-test-a"
 	OsmoBalanceA  = 200000000000
@@ -157,6 +159,21 @@ func initGenesis(c *internalChain) error {
 		return err
 	}
 	appGenState[banktypes.ModuleName] = bz
+
+	var govGenState govtypes.GenesisState
+	if err := util.Cdc.UnmarshalJSON(appGenState[govtypes.ModuleName], &govGenState); err != nil {
+		return err
+	}
+
+	govGenState.VotingParams = govtypes.VotingParams{
+		VotingPeriod: VotingPeriod,
+	}
+
+	gz, err := util.Cdc.MarshalJSON(&govGenState)
+	if err != nil {
+		return err
+	}
+	appGenState[govtypes.ModuleName] = gz
 
 	var genUtilGenState genutiltypes.GenesisState
 	if err := util.Cdc.UnmarshalJSON(appGenState[genutiltypes.ModuleName], &genUtilGenState); err != nil {
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -109,8 +109,11 @@ func (s *IntegrationTestSuite) runValidators(c *chain.Chain, portOffset int) {
 			Mounts: []string{
 				fmt.Sprintf("%s/:/osmosis/.osmosisd", val.ConfigDir),
 			},
-			Repository: "osmosis",
-			Tag:        "debug",
+			Repository: "osmolabs/osmosis-dev",
+			Tag:        "v7.0.4-debug",
+			Cmd: []string{
+				"start",
+			},
 		}
 
 		// expose the first validator for debugging and communication
@@ -267,8 +270,8 @@ func (s *IntegrationTestSuite) configureChain(chainId string) {
 	s.initResource, err = s.dkrPool.RunWithOptions(
 		&dockertest.RunOptions{
 			Name:       fmt.Sprintf("%s", chainId),
-			Repository: "osmosis-e2e-chain-init",
-			Tag:        "debug",
+			Repository: "osmolabs/osmosis-init",
+			Tag:        "v7.2.1",
 			NetworkID:  s.dkrNet.Network.ID,
 			Cmd: []string{
 				fmt.Sprintf("--data-dir=%s", tmpDir),
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -334,14 +334,13 @@ func noRestart(config *docker.HostConfig) {
 func (s *IntegrationTestSuite) initUpgrade() {
 	for x := range s.chains {
 		c := s.chains[x]
-		err := os.Chmod(c.Validators[0].ConfigDir, 0777)
-		if err != nil {
-			fmt.Println(err)
-		}
-		_, err2 := util.CopyFile(
+		_, err := util.CopyFile(
 			filepath.Join("./scripts/", "osmosis_upgrade.sh"),
 			filepath.Join(c.Validators[0].ConfigDir, "osmosis_upgrade.sh"),
 		)
+		s.Require().NoError(err)
+		cmd := fmt.Sprintf("chmod +x %s/osmosis_upgrade.sh", c.Validators[0].ConfigDir)
+		_, err2 := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
 		s.Require().NoError(err2)
 		s.T().Logf("submitting upgrade proposal for chain-id: %s", c.ChainMeta.Id)
 		cmdStr := fmt.Sprintf("docker exec %v bash -c 'chmod +x ~/.osmosisd/osmosis_upgrade.sh && ~/.osmosisd/osmosis_upgrade.sh \"%s\"'", s.valResources[c.ChainMeta.Id][0].Container.ID, c.ChainMeta.Id)
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -334,11 +334,15 @@ func noRestart(config *docker.HostConfig) {
 func (s *IntegrationTestSuite) initUpgrade() {
 	for x := range s.chains {
 		c := s.chains[x]
-		_, err := util.CopyFile(
+		err := os.Chmod(c.Validators[0].ConfigDir, 0777)
+		if err != nil {
+			fmt.Println(err)
+		}
+		_, err2 := util.CopyFile(
 			filepath.Join("./scripts/", "osmosis_upgrade.sh"),
 			filepath.Join(c.Validators[0].ConfigDir, "osmosis_upgrade.sh"),
 		)
-		s.Require().NoError(err)
+		s.Require().NoError(err2)
 		s.T().Logf("submitting upgrade proposal for chain-id: %s", c.ChainMeta.Id)
 		cmdStr := fmt.Sprintf("docker exec %v bash -c 'chmod +x ~/.osmosisd/osmosis_upgrade.sh && ~/.osmosisd/osmosis_upgrade.sh \"%s\"'", s.valResources[c.ChainMeta.Id][0].Container.ID, c.ChainMeta.Id)
 		exec.Command("/bin/sh", "-c", cmdStr).Output()
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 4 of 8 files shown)


Makefile
```diff
@@ -237,13 +237,13 @@ test-sim:
 	@VERSION=$(VERSION) go test -mod=readonly $(PACKAGES_SIM)
 
 test-e2e:
-	@VERSION=$(VERSION) go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
+	@VERSION=$(VERSION) go test -mod=readonly -timeout=25m -v github.com/osmosis-labs/osmosis/v7/tests/e2e
```
alexanderbez, line 240: so we're no longer testing the current code base?

tests/e2e/chain/config.go
```diff
@@ -27,6 +28,7 @@ const (
 	IbcDenom      = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
 	MinGasPrice   = "0.000"
 	IbcSendAmount = 3300000000
+	VotingPeriod  = 180000000000
```
alexanderbez, line 31: nit: can we add a small comment to make this more human readable? 

```suggestion
	VotingPeriod  = 180000000000 // X seconds or days or something
```

tests/e2e/scripts/osmosis_upgrade.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+osmosisd tx gov submit-proposal software-upgrade v8 --title "v8 upgrade" --description "v8 upgrade proposal" --upgrade-height 75 --upgrade-info "" --chain-id $1 --from val -b block --yes --keyring-backend test
```
alexanderbez, line 3: why not do these programmatically like we do for the other commands?

tests/e2e/chain/config.go
```diff
@@ -158,6 +160,21 @@ func initGenesis(c *internalChain) error {
...
+	var govGenState govtypes.GenesisState
+	if err := util.Cdc.UnmarshalJSON(appGenState[govtypes.ModuleName], &govGenState); err != nil {
+		return err
+	}
+
+	govGenState.VotingParams = govtypes.VotingParams{
+		VotingPeriod: VotingPeriod,
+	}
+
+	gz, err := util.Cdc.MarshalJSON(&govGenState)
```
p0mvn, line 172: So we unmarsal and then marshal back only to add `VotingPeriod`? Is it possible to instead add this flag when we are initializing genesis in the chain_init container?

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+	}
+
+	for x := range s.chains {
+		c := s.chains[x]
+		s.T().Logf("waiting to reach upgrade height for chain-id: %s", c.ChainMeta.Id)
+		type status struct {
+			LatestHeight string `json:"latest_block_height"`
+		}
+
+		type syncInfo struct {
```
p0mvn, line 351: please define types at the top of the file

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
 		Name: "no",
 	}
 }
+
+func (s *IntegrationTestSuite) initUpgrade() {
+	for x := range s.chains {
+		c := s.chains[x]
```
p0mvn, line 338:
```suggestion
	for _, chain := range s.chains {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+		s.T().Logf("submitting upgrade proposal for chain-id: %s", c.ChainMeta.Id)
+		cmdStr := fmt.Sprintf("docker exec %v bash -c 'chmod +x /scripts/osmosis_upgrade.sh && /scripts/osmosis_upgrade.sh \"%s\"'", s.valResources[c.ChainMeta.Id][0].Container.ID, c.ChainMeta.Id)
+		exec.Command("/bin/sh", "-c", cmdStr).Output()
+	}
+
+	for x := range s.chains {
+		c := s.chains[x]
```
p0mvn, line 345:
```suggestion
	for _, chain := range s.chains {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+		s.T().Logf("waiting to reach upgrade height for chain-id: %s", c.ChainMeta.Id)
+		type status struct {
```
p0mvn, line 347: move to the top please

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+			LatestHeight string `json:"latest_block_height"`
+		}
+
+		type syncInfo struct {
+			SyncInfo status `json:"SyncInfo"`
+		}
+
+		s.Require().Eventually(
```
p0mvn, line 355: Should we be doing this check for all validators in a for loop?

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+			2*time.Minute,
+			5*time.Second,
+		)
+	}
+
+	for x := range s.chains {
+		c := s.chains[x]
+		for i := range c.Validators {
+			s.Require().NoError(s.dkrPool.RemoveContainerByName(s.valResources[c.ChainMeta.Id][i].Container.Name))
+		}
```
p0mvn, line 376: I see a lot of for loops in this function, please add a comment before each to explain what is it that is happening there

p0mvn: or a log message

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+		for i := range c.Validators {
+			type status struct {
+				LatestHeight string `json:"latest_block_height"`
+			}
+
+			type syncInfo struct {
+				SyncInfo status `json:"SyncInfo"`
+			}
+
+			s.Require().Eventually(
```
p0mvn, line 413: Should this be done for every validator?

tests/e2e/e2e_setup_test.go
```diff
@@ -326,3 +332,100 @@ func noRestart(config *docker.HostConfig) {
...
+				func() bool {
+					cmdStr := fmt.Sprintf("docker exec %v bash -c 'osmosisd status'", s.valResources[c.ChainMeta.Id][0].Container.ID)
```
p0mvn, line 415: Should this be done this way to make it be checked for every validator?

```suggestion
					cmdStr := fmt.Sprintf("docker exec %v bash -c 'osmosisd status'", s.valResources[c.ChainMeta.Id][i].Container.ID)
```

p0mvn (changes requested):
Nice work, left some comments. Please take a look

Makefile
czarcas7ic: As per our discussion, will revert

tests/e2e/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+## Build Image
+FROM golang:1.18-bullseye as build
+
+WORKDIR /osmosis
+COPY . /osmosis
+
+# From https://github.com/CosmWasm/wasmd/blob/master/Dockerfile
+# For more details see https://github.com/CosmWasm/wasmvm#builds-of-libwasmvm 
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0-beta7/libwasmvm_muslc.a /lib/libwasmvm_muslc.a
```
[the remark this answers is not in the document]
czarcas7ic: So initially I checked the height via a `docker exec` through bash. This required an ubuntu image rather than distroless (the difference shown here). I just added a bunch of util functions and one of these utilizes a dockertest to check the height instead, which in turn allows me to use distroless again. So I will revert this

tests/e2e/e2e_test.go
```diff
@@ -14,7 +14,45 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/tests/e2e/util"
 )
 
-func (s *IntegrationTestSuite) TestQueryBalances() {
+func (s *IntegrationTestSuite) TestAIBCTokenTransfer() {
```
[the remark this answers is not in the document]
czarcas7ic: The tests in e2e test are run in alphabetical order. Since we token transfer first and then query the balance, I prefaced them with TestA and TestB respectively. Please let me know if there is another way we should go about this.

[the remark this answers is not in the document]
czarcas7ic: I just realized that the query balances fails when using distroless. Should I change the query function to instead use dockertest or should we just role with plain ubuntu for now? @p0mvn

tests/e2e/e2e_setup_test.go
czarcas7ic: Correct, will do

tests/e2e/e2e_setup_test.go
czarcas7ic: Correct, thanks

czarcas7ic (commented):
Thanks for the review, resolved everything that I added in the following commit. Just had a few clarifying questions.

tests/e2e/e2e_test.go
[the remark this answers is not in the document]
ValarDragon: uhhh, can we find some other abbreviations here. `AIBC` looks way to close to `ABCI` :sweat_smile:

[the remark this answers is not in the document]
ValarDragon: Happy for that to be relegated to a separate github issue tho

tests/e2e/e2e_util_test.go
```diff
@@ -106,3 +108,155 @@ func (s *IntegrationTestSuite) sendIBC(srcChainID, dstChainID, recipient string,
...
+	)
+
+	err = s.dkrPool.Client.StartExec(exec.ID, docker.StartExecOptions{
+		Context:      ctx,
+		Detach:       false,
+		OutputStream: &outBuf,
+		ErrorStream:  &errBuf,
+	})
+
+	s.Require().NoErrorf(
```
czarcas7ic, line 141: I think regardless of the outcome of the command this never error never gets triggered. My thought was to instead marshal the response and only unmarshal the error code and require code 0. Is this a good approach or is there a better way of doing this @p0mvn @alexanderbez

czarcas7ic: I instead opted to just check the outBuf.String() for code 0, I think it works and is simpler.

tests/e2e/chain_init/main.go
```diff
@@ -11,12 +11,30 @@ import (
...
+		chainId                string
+		numVal                 int
+		pruning                string
+		pruningKeepRecent      string
+		pruningInterval        string
+		snapshotInterval       uint64
+		snapshotKeepRecent     uint
+		telemetryEnabled       bool
+		telemetryRetentionTime int64
+		prometheusEnabled      bool
```
p0mvn, line 24: Why would we ever need these in a CI job?

tests/e2e/chain/config.go
p0mvn: I see, I misinterpreted how it all works. Thanks for clarifying

tests/e2e/chain/config.go
```diff
@@ -249,7 +266,7 @@ func initNodes(c *internalChain) error {
 	return nil
 }
 
-func initValidatorConfigs(c *internalChain) error {
+func initValidatorConfigs(c *internalChain, pruning string, pruningKeepRecent string, pruningInterval string, snapshotInterval uint64, snapshotKeepRecent uint32, telemetryEnabled bool, telemetryRetentionTime int64, prometheusEnabled bool) error {
```
p0mvn, line 269: I think these configs should instead be per-validator. Can we do this in a separate PR? 

I think we should limit the scope of this one to just the upgrade and keep adding on top of it after. Otherwise, we might have to keep iterating on this long PR for some time. Let me know your thoughts

tests/e2e/e2e_test.go
```diff
@@ -14,7 +14,45 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/tests/e2e/util"
 )
 
-func (s *IntegrationTestSuite) TestQueryBalances() {
+func (s *IntegrationTestSuite) Test1IBCTokenTransfer() {
```
p0mvn, line 17: Let's name `001`. I'm just expecting that we might have more tests relying on order soon

tests/e2e/e2e_setup_test.go
```diff
@@ -325,3 +336,85 @@ func noRestart(config *docker.HostConfig) {
...
+	s.voteProposal(s.chains[0])
+	s.voteProposal(s.chains[1])
+
+	// wait till all chains halt at upgrade height
+	for _, chain := range s.chains {
+		for i := range chain.Validators {
+			s.T().Logf("waiting to reach upgrade height for chain-id: %s", chain.ChainMeta.Id)
+			s.Require().Eventually(
+				func() bool {
+					out := s.chainStatus(chain, i)
```
p0mvn, line 355: Passing an index to a function like this is a code smell. It basically couples the function on its parameter. I'd like to suggest passing `s.valResources[c.ChainMeta.Id][i].Container.ID` here directly

tests/e2e/e2e_setup_test.go
```diff
@@ -325,3 +336,85 @@ func noRestart(config *docker.HostConfig) {
...
+	s.submitProposal(s.chains[1])
+	s.depositProposal(s.chains[0])
+	s.depositProposal(s.chains[1])
+	s.voteProposal(s.chains[0])
+	s.voteProposal(s.chains[1])
+
+	// wait till all chains halt at upgrade height
+	for _, chain := range s.chains {
+		for i := range chain.Validators {
+			s.T().Logf("waiting to reach upgrade height for chain-id: %s", chain.ChainMeta.Id)
```
p0mvn, line 352: let's also add a container id log?

tests/e2e/e2e_setup_test.go
```diff
@@ -325,3 +336,85 @@ func noRestart(config *docker.HostConfig) {
...
+			s.Require().Eventually(
+				func() bool {
+					out := s.chainStatus(chain, i)
+					var syncInfo syncInfo
+					json.Unmarshal(out, &syncInfo)
+					if syncInfo.SyncInfo.LatestHeight != "75" {
+						fmt.Printf("current block height is %v, waiting for block 75\n", syncInfo.SyncInfo.LatestHeight)
```
p0mvn, line 359: Let's use `s.T().Logf(...)` to be consistent

p0mvn: Also suggest logging container id

tests/e2e/e2e_setup_test.go
```diff
@@ -325,3 +336,85 @@ func noRestart(config *docker.HostConfig) {
...
+		s.T().Logf("starting upgrade for chain-id: %s...", chain.ChainMeta.Id)
+		for i, val := range chain.Validators {
+			runOpts := &dockertest.RunOptions{
+				Name:       val.Name,
+				Repository: "osmosis",
+				Tag:        "debug",
+				NetworkID:  s.dkrNet.Network.ID,
+				User:       "root:root",
+				Mounts: []string{
+					fmt.Sprintf("%s/:/osmosis/.osmosisd", val.ConfigDir),
```
p0mvn, line 390: `s.T().Logf(...)

p0mvn: Please remove all occurrences of `fmt.Sprintf()` if possible

p0mvn (changes requested):
Great work on the updates. Left some comments, let me know what you think. Also, already DMed you about potentially moving chain parameters to a separate PR if that's not too much trouble

tests/e2e/e2e_setup_test.go
czarcas7ic: Didn't know this but it makes total sense, thanks for explaining the reasoning behind it!

p0mvn (approved):
LGTM

Issue #1372, opened 2022-04-28
[x/gamm][Testing] balancer/CalcJoinPoolShares


p0mvn:
**Background**

We have observed rounding errors in balancer post-refactor (#1307). There are several balancer pool methods being called from the function in #1307 under the hood that are not unit-tested.

As a result, it is non-trivial to identify the source of the problem. All methods should be covered by unit tests. One of such untested methods is [`CalcJoinPoolShares`](https://github.com/osmosis-labs/osmosis/blob/69811b3f40367a162f093cc437a4a95cd504ab9e/x/gamm/pool-models/balancer/amm.go#L316)

On initial investigation, this method does many things, violating the single responsibility principle in [SOLID](https://en.wikipedia.org/wiki/Single-responsibility_principle). As a result, it is difficult to reason about and test.

The person taking on this issue should investigate if it is worth refactoring this method to extract a few others to be called under the hood to mitigate the violation. If any, all newly extracted methods should be unit tested, including `CalcJoinPoolShares` itself.

**Acceptance Criteria**
- investigate the need for extracting several methods from `CalcJoinPoolShares` to ease unit testing and improve readability
- ensure all logic in `CalcJoinPoolShares` and newly extracted functions/methods, if any, are covered by unit tests
- edge cases are well thought out
- all old and new unit tests pass

p0mvn:
@alexanderbez would you be interested in working on this? Thanks for offering to help with the tests

PR #1430, opened 2022-05-05
chore: Increase CalcJoinPoolShares Test Coverage

alexanderbez:
- Lint/cleanup `CalcJoinPoolShares`
- Add `NewPoolParams`
- Add `TestCalcJoinPoolShares` test

closes: #1372


x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -324,3 +318,65 @@ func TestCalcSingleAssetInAndOut_InverseRelationship(t *testing.T) {
...
+		// 	swapFee: sdk.MustNewDecFromStr("0.1"),
+		// },
+		// {
+		// 	name: "equal weights with 0.99 swap fee",
+		// 	swapFee: sdk.MustNewDecFromStr("0.99"),
+		// },
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
p0mvn, line 365: Is this to ensure that we are operating on the desired copy when the for loop advances to the next tc?

alexanderbez: Correct, since we run the test in a closure.

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -324,3 +318,65 @@ func TestCalcSingleAssetInAndOut_InverseRelationship(t *testing.T) {
...
+					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+			},
+			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
+			expectErr:    false,
+			expectShares: sdk.NewInt(2499999968800),
+			expectLiq:    sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50000)),
+		},
+		// {
```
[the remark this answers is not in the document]
alexanderbez: No I'm just waiting on feedback on https://github.com/osmosis-labs/osmosis/pull/1430#issuecomment-1122616623

p0mvn:
> So I can add a few more test cases @p0mvn @ValarDragon, but when in order to determine expected outputs, I'm just using what the function returns, which doesn't induce a large amount of confidence lol. Is there a way to know what the expected values _should_ be without just running the function?

I agree, and I'm also not sure how we can estimate the values without relying on our code. Maybe we should manually calculate 1-3 tests? @ValarDragon wdyt?

ValarDragon:
cc @AlpinYukseloglu can you help with using the balancer invariant to calculate exact expected quantities for test vectors here

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -324,3 +318,65 @@ func TestCalcSingleAssetInAndOut_InverseRelationship(t *testing.T) {
...
+					Weight: sdk.NewInt(100),
+				},
+				{
+					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+			},
+			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
+			expectErr:    false,
+			expectShares: sdk.NewInt(2499999968800),
```
AlpinYukseloglu, line 347: How was this calculated @alexanderbez?

Naser531900: 😎😎

alexanderbez: By calling the function @AlpinYukseloglu 😆

ValarDragon: I'm down to have this merged as is, and handle more verification of these numbers in Alpin's follow-up PR

Issue #1414, opened 2022-05-04
[E2E - Upgrade Testing] add nodes with various configs to CI


czarcas7ic:
When spinning up the pre-upgrade and post upgrade testnet, we want to have nodes with various configs (in order of importance):
- [x] Pruning
- [x] Snapshot
- [ ] #1878

In order to complete the upgrade testing issue, the testing framework should be able to do the following:
- [x] Create state on v7 (or whatever the pre upgrade version may be)
  - [x] Create a pool
  - [x] Create a lockup
  - [x] Other potential messages we want to send to create required state
- [x] Upgrade to locally compiled v8 (or whatever the post upgrade version may be) 
- [x] Run similar tests that were used to create prior state on v7 (or whatever the pre upgrade version was)

Post upgrade completion, it would be nice to:
- [ ] Have an option to retain the nodes and ssh into them post test

PR #1437, opened 2022-05-05
test: expose individual validator options

czarcas7ic:
Completes Part Of: #1414 

This pull request allows the user to express how many validators each chain should have, and then exposes the following options on each individual validator
- Pruning
- PruningKeepRecent
- PruningInterval
- SnapshotInterval
- SnapshotKeepRecent

I wanted this portion to be approved before finishing the rest of #1414 to ensure I went about this in the best possible way.

## Brief change log
 
  - New default is running three nodes on each chain, each with different pruning options
    - Side note, what do we want the defaults for snapshot interval and keep recent to be?
  - Now every validator votes for a proposal in order to reach quorum

## Testing and Verifying

This change is already covered by the existing e2e test framework

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? not yet but will


tests/e2e/chain/config.go
```diff
@@ -21,6 +21,15 @@ import (
...
 )
 
+type ValidatorConfig struct {
+	NumVal             int
+	Pruning            []string
+	PruningKeepRecent  []string
+	PruningInterval    []string
+	SnapshotInterval   []uint64
+	SnapshotKeepRecent []uint32
+}
```
p0mvn, line 31:
```suggestion
type ValidatorConfig struct {
	Pruning            string
	PruningKeepRecent  string
	PruningInterval    string
	SnapshotInterval   uint64
	SnapshotKeepRecent uint32
}
```

tests/e2e/chain/config.go
```diff
@@ -266,7 +275,7 @@ func initNodes(c *internalChain) error {
 	return nil
 }
 
-func initValidatorConfigs(c *internalChain) error {
+func initValidatorConfigs(c *internalChain, pruning []string, pruningKeepRecent []string, pruningInterval []string, snapshotInterval []uint64, snapshotKeepRecent []uint32) error {
```
p0mvn, line 278:
```suggestion
func initValidatorConfigs(c *internalChain, validatorConfigs []*ValidatorConfig) error {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -38,6 +55,15 @@ type IntegrationTestSuite struct {
...
 }
 
+type ValidatorConfig struct {
+	NumVal             int
+	Pruning            []string
+	PruningKeepRecent  []string
+	PruningInterval    []string
+	SnapshotInterval   []uint64
+	SnapshotKeepRecent []uint32
+}
```
p0mvn, line 65: Why do we need 2? We can import the other one

```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -24,7 +24,24 @@ import (
...
+var (
+	// common
+	maxRetries = 10 // max retries for json unmarshalling
+	// chainA
+	NumValA             int = 3
+	PruningA                = []string{"nothing", "default", "custom"} // default, nothing, everything, or custom
+	PruningKeepRecentA      = []string{"0", "0", "10000"}              // keep all of the last N states (only used with custom pruning)
+	PruningIntervalA        = []string{"0", "0", "13"}                 // delete old states from every Nth block (only used with custom pruning)
+	SnapshotIntervalA       = []uint64{1500, 1500, 1500}               // statesync snapshot every Nth block (0 to disable)
+	SnapshotKeepRecentA     = []uint32{2, 2, 2}                        // number of recent snapshots to keep and serve (0 to keep all)
```
p0mvn, line 36: syntax is likely to be off but the idea is to have something like the following:
```suggestion
        validatorConfigsChainA = []*chain.ValidatorConfig{
             {
                 PruningStrategy: "default",
                 ....
             },
             {
                 PruningStrategy: "nothing",
                 ....
             },
             ...
        }
```

tests/e2e/e2e_setup_test.go
```diff
@@ -268,17 +294,43 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
 }
 
 func (s *IntegrationTestSuite) configureChain(chainId string) {
```
p0mvn, line 296:
```suggestion
func (s *IntegrationTestSuite) configureChain(chainId, validatorConfigs []*chain.ValidatorConfig) {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -268,17 +294,43 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
...
+	var config ValidatorConfig
```
p0mvn, line 297:
```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -268,17 +294,43 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
...
+	} else if chainId == chain.ChainBID {
+		config.NumVal = NumValB
+		config.Pruning = PruningB
+		config.PruningKeepRecent = PruningKeepRecentB
+		config.PruningInterval = PruningIntervalB
+		config.SnapshotInterval = SnapshotIntervalB
+		config.SnapshotKeepRecent = SnapshotKeepRecentB
+	} else {
+		s.T().Log("unexpected chainId")
+	}
```
p0mvn, line 320:
```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -268,17 +294,43 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
...
+
+	b, _ := json.Marshal(config)
```
p0mvn, line 322:
```suggestion
	b, _ := json.Marshal(validatorConfig)
```

tests/e2e/chain/config.go
```diff
@@ -308,8 +317,13 @@ func initValidatorConfigs(c *internalChain) error {
 		appCfgPath := filepath.Join(val.configDir(), "config", "app.toml")
 
 		appConfig := srvconfig.DefaultConfig()
+		appConfig.BaseConfig.Pruning = pruning[i]
```
p0mvn, line 320:
```suggestion
		appConfig.BaseConfig.Pruning = validatorConfigs[i].Pruning
```

tests/e2e/chain/main.go
```diff
@@ -1,17 +1,17 @@
 package chain
 
-func Init(id, dataDir string) (*Chain, error) {
+func Init(id, dataDir string, numVal int, pruning []string, pruningKeepRecent []string, pruningInterval []string, snapshotInterval []uint64, snapshotKeepRecent []uint32) (*Chain, error) {
```
p0mvn, line 3:
```suggestion
func Init(id, dataDir string, validatorConfigs []*ValidatorConfig) (*Chain, error) {
```

tests/e2e/chain/main.go
```diff
@@ -1,17 +1,17 @@
...
 	chain, err := new(id, dataDir)
 	if err != nil {
 		return nil, err
 	}
-	if err := initNodes(chain); err != nil {
+	if err := initNodes(chain, numVal); err != nil {
```
p0mvn, line 8:
```suggestion
	if err := initNodes(chain, len(validatorConfigs)); err != nil {
```

p0mvn (commented):
Nice work, would like to suggest the following structure to avoid passing a lot of parameters across functions and duplicating struct definitions

tests/e2e/chain_init/main.go
```diff
@@ -5,19 +5,38 @@ import (
...
+		dataDir   string
+		chainId   string
 	)
 
 	flag.StringVar(&dataDir, "data-dir", "", "chain data directory")
 	flag.StringVar(&chainId, "chain-id", "", "chain ID")
+
 	flag.Parse()
+	file := fmt.Sprintf("%v/%v-configEncode", dataDir, chainId)
+	for i := 0; i < 10; i++ {
```
p0mvn, line 25: Why is this needed? Is it because we are mounting on a volume? 

If yes, can we instead pass a string via CLI? These files are expected to be quite small. I think it would be a cleaner solution than doing repeats

tests/e2e/e2e_setup_test.go
```diff
@@ -267,18 +316,26 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
...
-func (s *IntegrationTestSuite) configureChain(chainId string) {
+func (s *IntegrationTestSuite) configureChain(chainId string, validatorConfigs []*chain.ValidatorConfig) {
+
 	s.T().Logf("starting e2e infrastructure for chain-id: %s", chainId)
 	tmpDir, err := ioutil.TempDir("", "osmosis-e2e-testnet-")
 
 	s.T().Logf("temp directory for chain-id %v: %v", chainId, tmpDir)
 	s.Require().NoError(err)
 
+	b, _ := json.Marshal(validatorConfigs)
```
p0mvn, line 327: error check

tests/e2e/e2e_setup_test.go
```diff
@@ -267,18 +316,26 @@ func (s *IntegrationTestSuite) runIBCRelayer() {
...
+	file := fmt.Sprintf("%v/%v-configEncode", tmpDir, chainId)
```
p0mvn, line 328: As per my previous comment, instead of creating a file and writing to disk, we could pass in the byte stream via CLI

p0mvn (commented):
Good job on the clean-up! Would like to suggest passing the bytes via CLI instead of writing to disk to avoid the hack with repeat reads from disk

tests/e2e/e2e_setup_test.go
czarcas7ic: As in, pass it in as a flag via Cmd via dockertest?

p0mvn: Yes

tests/e2e/e2e_setup_test.go
```diff
@@ -24,7 +24,56 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/tests/e2e/util"
 )
 
-const maxRetries = 10 // max retries for json unmarshalling
+var (
+	// common
+	maxRetries             = 10 // max retries for json unmarshalling
+	validatorConfigsChainA = []*chain.ValidatorConfig{
+		{
+			Pruning:            "default", // default, nothing, everything, or custom
```
p0mvn, line 32: let's move the comments to the struct definition in chain package instead please

p0mvn (approved):
LGTM

PR #1445, opened 2022-05-07
Enable wasm mint bindings

maurolacy:
Closes / Related to: #1029

Re-instate the CoswmWasm message handlers for Mint and FullDenom. Related to https://github.com/confio/osmosis-bindings/pull/25 on the Rust side.

## Brief change log

  - Uncomment the Mint-related code.
  - Implemented `Mint` using `TokenFactory`.
  - Implemented `FullDenom` in terms of `TokenFactory`.

## Testing and Verifying

Uncommented and adapted already existing tests.

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes. For the CosmWasm side)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (will do)


app/wasm/message_plugin.go
```diff
@@ -12,22 +11,27 @@ import (
 	wasmbindings "github.com/osmosis-labs/osmosis/v7/app/wasm/bindings"
 	gammkeeper "github.com/osmosis-labs/osmosis/v7/x/gamm/keeper"
 	gammtypes "github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+
+	tokenfactorykeeper "github.com/osmosis-labs/osmosis/v7/x/tokenfactory/keeper"
+	tokenfactorytypes "github.com/osmosis-labs/osmosis/v7/x/tokenfactory/types"
 )
 
-func CustomMessageDecorator(gammKeeper *gammkeeper.Keeper, bank *bankkeeper.BaseKeeper) func(wasmkeeper.Messenger) wasmkeeper.Messenger {
+func CustomMessageDecorator(gammKeeper *gammkeeper.Keeper, bank *bankkeeper.BaseKeeper, tokenFactory *tokenfactorykeeper.Keeper) func(wasmkeeper.Messenger) wasmkeeper.Messenger {
```
ethanfrey, line 19: :+1:

app/wasm/message_plugin.go
```diff
@@ -40,52 +44,70 @@ func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddre
...
+		return wasmvmtypes.InvalidRequest{Err: "mint token null mint"}
+	}
+	rcpt, err := parseAddress(mint.Recipient)
+	if err != nil {
+		return err
+	}
+
+	// Check if denom is valid
+	denom, err := GetFullDenom(contractAddr.String(), mint.SubDenom)
+	if err != nil {
```
ethanfrey, line 76: Missing Rust's `?` operator here

app/wasm/message_plugin.go
```diff
@@ -40,52 +44,70 @@ func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddre
...
+		return wasmvmtypes.InvalidRequest{Err: "mint token zero amount"}
+	}
+	if mint.Amount.IsNegative() {
+		return wasmvmtypes.InvalidRequest{Err: "mint token negative amount"}
+	}
+	coin := sdk.NewCoin(denom, mint.Amount)
+
+	msgServer := tokenfactorykeeper.NewMsgServerImpl(*f)
+
+	// Check if denom already exists
```
ethanfrey, line 90: Okay, this is the critical question.

Assuming we never set any metadata that is not present in the MintTokens command, this works great and makes a simple API.

However, if they plan to add more metadata there, we will want a separate command to run once to register the metadata (and another to query it?).

ethanfrey: I would merge this as is (we can test stuff with it), but have a discussion with @ValarDragon and @sunnya97 before tagging an Osmosis release.

If they want to add more required metadata, they should add it in Go and we should expose it in Rust now. Otherwise, it will most likely be a breaking contract API change, which we want very much to avoid.

app/wasm/message_plugin.go
```diff
@@ -155,44 +177,28 @@ func PerformSwap(keeper *gammkeeper.Keeper, ctx sdk.Context, contractAddr sdk.Ac
...
-// 		return nil, sdkerrors.Wrap(err, "address from bech32")
-// 	}
-// 	err = sdk.VerifyAddressFormat(parsed)
-// 	if err != nil {
-// 		return nil, sdkerrors.Wrap(err, "verify address format")
-// 	}
-// 	return parsed, nil
-// }
-
-// const reSubdenomStr = `^[a-zA-Z][a-zA-Z0-9]{2,31}$`
```
ethanfrey, line 185: Yes, this is done in token factory now, no need for us to do so.

app/wasm/test/custom_msg_test.go
```diff
@@ -14,46 +15,50 @@ import (
 	wasmbindings "github.com/osmosis-labs/osmosis/v7/app/wasm/bindings"
 )
 
-// func TestMintMsg(t *testing.T) {
-// 	creator := RandomAccountAddress()
-// 	osmosis, ctx := SetupCustomApp(t, creator)
+func TestMintMsg(t *testing.T) {
```
ethanfrey, line 18: This is a good sanity check of the happy path.

I would add something else to check for metadata, and also mint a few times from the same contract, With same and different denoms (eg. A, A, B) to ensure metadata is set properly and in all cases. (esp. that the "only create once" guard is correct)

app/wasm/test/messages_test.go
```diff
@@ -7,102 +7,107 @@ import (
...
-// 				return
-// 			}
-// 			require.NoError(t, gotErr)
-// 		})
-// 	}
+	amount, ok := sdk.NewIntFromString("8080")
+	require.True(t, ok)
 
-// }
+	specs := map[string]struct {
```
ethanfrey, line 33: Ok, this covers most of what I wished above... this hits many branches.

I would add tests for metadata being set properly. And I would add a test where the same SubDenom is minted twice (without resetting the context). This is the main area I think there could be an integration failure

(Your code looks clean, just saying this is where some minor change to an if statement would break contracts, so let's provide a test to protect against it)

p0mvn:
Hi @maurolacy. Thank you for this PR. Is it ready for review? If in progress, could you mark it as a draft, please?

ethanfrey (approved):
Nice one.

One test I would like to see (minting the same subdenom twice). Otherwise good to merge.

One open question for the Osmosis core devs before this is tagged. But please don't block merging this, we can enhance it in a follow up.

app/wasm/message_plugin.go
```diff
@@ -40,52 +44,70 @@ func (m *CustomMessenger) DispatchMsg(ctx sdk.Context, contractAddr sdk.AccAddre
...
-// 		return sdkerrors.Wrap(err, "sending newly minted coins from message")
-// 	}
-// 	return nil
-// }
+func (m *CustomMessenger) mintTokens(ctx sdk.Context, contractAddr sdk.AccAddress, mint *wasmbindings.MintTokens) ([]sdk.Event, [][]byte, error) {
+	err := PerformMint(m.tokenFactory, m.bank, ctx, contractAddr, mint)
+	if err != nil {
+		return nil, nil, sdkerrors.Wrap(err, "perform mint")
+	}
+	return nil, nil, nil
```
p0mvn, line 62: Do we need to return `[]sdk.Event` and `[][]byte` if these are always nil?

app/wasm/testdata/version.txt
```diff
@@ -1 +1 @@
-v0.5.0
```
p0mvn, line 1: asking to learn - how is the tesdata generated? Is there a repo/code I can look at?

app/wasm/test/custom_msg_test.go
```diff
@@ -2,6 +2,7 @@ package wasm
```
p0mvn, line 2: Can we make this a `wasm_test` package as well as all of the other test files:
- `custom_msg_test.go`
- `custom_query_test.go`
- `messages_test.go`
- `queries_test.go`

Let's not block merge on this but creating an issue for that would be great!

p0mvn (approved):
LGTM once CI is passing

app/wasm/message_plugin.go
ethanfrey: It's part of the required interface for DispatchMsg. And saves us from rewriting in the caller above.

That said, it would be interesting to define some standard data/events to return to contracts for better integration. But that would be a larger discussion with the Osmosis team on API design

app/wasm/testdata/version.txt
ethanfrey: https://github.com/osmosis-labs/osmosis/blob/main/app/wasm/testdata/download_releases.sh

app/wasm/test/custom_msg_test.go
ethanfrey: What is the reasoning here?

Basically, the default is usually the same package, and there are valid reasons to make it a separate `wasm_test` package, but it is good to discuss which one to prefer

app/wasm/message_plugin.go
maurolacy: Yeah, me too.

app/wasm/test/custom_msg_test.go
maurolacy: Will do.

app/wasm/testdata/version.txt
maurolacy: As part of this, can create a new release of osmosis-bindings, and update version.txt here. Will do.

app/wasm/test/messages_test.go
maurolacy: Done in the latest version.

mconcat (approved):
LGTM!

ValarDragon:
I don't have permission to update the branch, but once its updated lets go ahead and merge this in

maurolacy:
I just updated it manually. That's why the version.txt file has a commit id (from the osmosis-bindings repository).

If you use the committed version of the wasm file, you should be good to go for tests.

I'll update this after releasing a new osmosis-bindings version.

nicolaslara:
@maurolacy thanks for the quick reply! ...Using the committed version I get the a "variant missing" error. Which makes sense, since that wasm file doesn't have information about the new type I've added on `osmosis-bindings`

```
=== RUN   TestLockTokensMsg
    custom_msg_test.go:638:
        	Error Trace:	custom_msg_test.go:638
        	Error:      	Received unexpected error:

        	            	github.com/CosmWasm/wasmd/x/wasm/keeper.Keeper.execute
        	            		/Users/nicolas/go/pkg/mod/github.com/osmosis-labs/wasmd@v0.27.0-rc2.0.20220517191021-59051aa18d58/x/wasm/keeper/keeper.go:354
        	            	github.com/CosmWasm/wasmd/x/wasm/keeper.PermissionedKeeper.Execute
        	            		/Users/nicolas/go/pkg/mod/github.com/osmosis-labs/wasmd@v0.27.0-rc2.0.20220517191021-59051aa18d58/x/wasm/keeper/contract_keeper.go:51
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.executeCustom
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_msg_test.go:543
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.TestLockTokensMsg
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_msg_test.go:637
        	            	Error parsing into type osmo_reflect::msg::ExecuteMsg: unknown variant `lock_tokens`, expected `mint_tokens` or `swap`: execute wasm contract failed
        	Test:       	TestLockTokensMsg
--- FAIL: TestLockTokensMsg (0.36s)
FAIL
FAIL	github.com/osmosis-labs/osmosis/v7/app/wasm/test	0.952s
FAIL
```

nicolaslara:
ahh, I think I know what the problem is. I hadn't finished the mock in `multitest.rs`. I'll fix that and report back.

Issue #642, opened 2021-12-07
State Sync is broken on mainnet


faddat:
So basically the Block.Protocol.Version is set wrong in the block headers in osmosis-- I think.  Here's the result:

(pending script completing a run)

faddat:
```
230","port":2000}
8:11AM INF Added peer module=p2p peer={"Data":{},"Logger":{}}
8:11AM INF Stopping queryMaj23Routine for peer module=consensus peer={"Data":{},"Logger":{}}
panic: Failed to process committed block (2284501:147C5444A238F247A79F08807FD33BE86D155E3385DB807165DACAFF06DD3C60): wrong Block.Header.Version. Expected {11 0}, got {11 1}

goroutine 6946 [running]:
github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).poolRoutine(0x140005d2000, 0x1)
	/Users/faddat/go/pkg/mod/github.com/tendermint/tendermint@v0.34.14/blockchain/v0/reactor.go:401 +0x10e4
created by github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).SwitchToFastSync
	/Users/faddat/go/pkg/mod/github.com/tendermint/tendermint@v0.34.14/blockchain/v0/reactor.go:125 +0xe4
```

dualsystems:
on emoney chain sometimes when sync via statesync, I get error:
9:54AM ERR WARNING: Attempt to allocate proposer rewards to unknown proposer . This should happen only if the proposer unbonded completely within a single block, which generally should not happen except in exceptional circumstances (or fuzz testing). We recommend you investigate immediately. module=x/distribution
9:54AM INF Endblock: Block 287401 was proposed by 51C818CB49A0B0C8FE8CDD7F217ABFF15AAE9988 module=emz
9:54AM INF executed block height=287401 module=state num_invalid_txs=0 num_valid_txs=0
9:54AM INF commit synced commit=436F6D6D697449447B5B31333020333220353320313635203134372031333420393320353820323620363620353520313420323037203131312032303320323438203230322031333520323620373420323020342032303220313337203135352031393620323320313833203136302031323620313038203134325D3A34363241397D
9:54AM INF committed state app_hash=822035A593865D3A1A42370ECF6FCBF8CA871A4A1404CA899BC417B7A07E6C8E height=287401 module=state num_txs=0
9:54AM INF indexed block height=287401 module=txindex
[...]

PR #1449, opened 2022-05-08
Update test.yml

faddat:
Enables #1178 

Runs liveness test only on changes to the main branch.

## Brief change log

* modify liveness test github action

 
Need to fix the conditional on that, maybe today or maybe tomorrow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,7 @@ jobs:
 
   liveness-test:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
     timeout-minutes: 10
     steps:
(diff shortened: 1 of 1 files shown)


ValarDragon:
Does it not make sense for future release branches?

faddat:
@ValarDragon I think that I should actually close this.

I have something better in mind, that would run many tests at once, and it is half-developed:

https://github.com/notional-labs/tm-db/runs/6438218175


This corresponds to #642 : if we have state sync, and we use buildjet, we can do way, way more testing, faster and easier.

czarcas7ic:
We also removed liveness test in favor of e2e test. @faddat is there a reason we should keep liveness test as a check still?

PR #1458, opened 2022-05-10
test: more e2e upgrade testing

czarcas7ic:
Completes all required tasks for: #1414 

Adds further functionality to the e2e and upgrade test framework.

## Brief Changelog
 
  - Created IBC send and pool creation functions
    - These function are called pre upgrade to create state and post upgrade to test if it still works as intended
  - Adds four pool json files required for pool creation command
  - Submit, deposit, vote, chain height, and query utility functions created
  - Now required code 0 (success) from chain. Before only checked that the commands were sent successfully to the docker container which is only half the picture.
  - Modified the ibc send test so that it will pass regardless of what order it is called in.
  - Bug fixes to allow for a variety of validator combinations

## Testing and Verifying

This change is already covered by the existing e2e test framework and expands upon it. It was also tested locally.


tests/e2e/chain/config.go
```diff
@@ -33,7 +33,8 @@ const (
 	// common
 	OsmoDenom     = "uosmo"
 	StakeDenom    = "stake"
-	IbcDenom      = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
+	OsmoIbcDenom  = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
```
alexanderbez, line 36:
```suggestion
	OsmoIBCDenom  = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
```

tests/e2e/e2e_setup_test.go
```diff
@@ -25,8 +26,11 @@ import (
 )
 
 var (
-	// common
-	maxRetries             = 10 // max retries for json unmarshalling
+	propHeight  int
+	propHeightA int  // upgrade proposal height for chain A
+	propHeightB int  // upgrade proposal height for chain B
+	maxRetries  = 60 // max retries for json unmarshalling
+	// whatever number of validator configs get posted here are how many validators that will spawn on chain A and B respectively
```
alexanderbez, line 33: Can we put the comment above the variable so it's easier to read?

tests/e2e/e2e_util_test.go
```diff
@@ -102,27 +107,44 @@ func (s *IntegrationTestSuite) sendIBC(srcChainID, dstChainID, recipient string,
...
 	})
+
 	s.Require().NoErrorf(
 		err,
 		"failed to send IBC tokens; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
 	)
 
+	s.Require().Truef(
+		strings.Contains(outBuf.String(), "Success"),
+		"tx returned non code 0; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
```
p0mvn, line 118:
```suggestion
		"tx returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
```

tests/e2e/e2e_util_test.go
```diff
@@ -102,27 +107,44 @@ func (s *IntegrationTestSuite) sendIBC(srcChainID, dstChainID, recipient string,
...
 }
 
-func (s *IntegrationTestSuite) submitProposal(c *chain.Chain) {
+func (s *IntegrationTestSuite) submitProposal(c *chain.Chain, upgradeHeight int) {
+	upgradeHeightStr := strconv.Itoa(upgradeHeight)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	s.T().Logf("submitting upgrade proposal for chain-id: %s", c.ChainMeta.Id)
+	s.T().Logf("submitting upgrade proposal on %s container: %s", s.valResources[c.ChainMeta.Id][0].Container.Name[1:], s.valResources[c.ChainMeta.Id][0].Container.ID)
```
p0mvn, line 139: Why do we need to log both name and id?

tests/e2e/e2e_util_test.go
```diff
@@ -146,7 +168,17 @@ func (s *IntegrationTestSuite) submitProposal(c *chain.Chain) {
 
 	s.Require().Truef(
 		strings.Contains(outBuf.String(), "code: 0"),
-		"tx returned non code 0",
+		"tx returned non code 0; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
```
p0mvn, line 171:
```suggestion
		"tx returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
```

tests/e2e/e2e_util_test.go
```diff
@@ -146,7 +168,17 @@ func (s *IntegrationTestSuite) submitProposal(c *chain.Chain) {
...
+	)
+
+	currentHeight := s.chainHeight(s.valResources[c.ChainMeta.Id][0].Container.ID)
+
+	s.Require().Eventually(
+		func() bool {
+			return s.chainHeight(s.valResources[c.ChainMeta.Id][0].Container.ID) > currentHeight+2
```
p0mvn, line 178: What is happening here? Why are we getting `currentHeight` and then trying to eventually reach 2 above current?

tests/e2e/e2e_util_test.go
```diff
@@ -188,14 +220,25 @@ func (s *IntegrationTestSuite) depositProposal(c *chain.Chain) {
 
 	s.Require().Truef(
 		strings.Contains(outBuf.String(), "code: 0"),
-		"tx returned non code 0",
+		"tx returned non code 0; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
+	)
+
+	currentHeight := s.chainHeight(s.valResources[c.ChainMeta.Id][0].Container.ID)
+
+	s.Require().Eventually(
```
p0mvn, line 228: What is the purpose of this?

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +387,116 @@ func noRestart(config *docker.HostConfig) {
...
 	s.depositProposal(s.chains[0])
 	s.depositProposal(s.chains[1])
-	s.voteProposal(s.chains[0])
-	s.voteProposal(s.chains[1])
-
+	var wg sync.WaitGroup
+	wg.Add(2)
+	// use goroutines to vote on both chains concurrently (no possibility of sequence mismatch)
+	go s.voteProposal(s.chains[0], &wg)
+	go s.voteProposal(s.chains[1], &wg)
```
p0mvn, line 409: Since these are 2 independent chains, why does it matter if we are voting concurrently?

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +387,116 @@ func noRestart(config *docker.HostConfig) {
...
+						s.T().Logf("current block height on %s is %v, waiting for block %v container: %s", s.valResources[c.ChainMeta.Id][i].Container.Name[1:], currentHeight, propHeight, s.valResources[c.ChainMeta.Id][i].Container.ID)
 					}
-					return syncInfo.SyncInfo.LatestHeight == "75"
+					if currentHeight > propHeight {
+						panic("chain did not halt at upgrade height")
+					}
+					if currentHeight == propHeight {
+						counter++
+					}
+					return counter == 3
```
p0mvn, line 434: What is the purpose for having both a counter and a 5 minute timeout?

By having a counter running 3 times max every second , we essentially have a 3 second timeout which is much lower than `5*time.Minute`. We can then remove the counter, always return true, and change timeout from `5*time.Minute` to `3*time.Second`

tests/e2e/e2e_util_test.go
```diff
@@ -276,3 +319,107 @@ func (s *IntegrationTestSuite) chainStatus(containerId string) []byte {
 	return errBufByte
 
 }
+
+func (s *IntegrationTestSuite) chainHeight(containerId string) int {
```
p0mvn, line 323:
```suggestion
func (s *IntegrationTestSuite) getCurrentChainHeight(containerId string) int {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -35,20 +39,8 @@ var (
...
-			SnapshotKeepRecent: 2,
-		},
-		{
-			Pruning:            "custom",
-			PruningKeepRecent:  "10000",
-			PruningInterval:    "13",
-			SnapshotInterval:   1500,
-			SnapshotKeepRecent: 2,
-		},
+		{Pruning: "nothing", PruningKeepRecent: "0", PruningInterval: "0", SnapshotInterval: 1500, SnapshotKeepRecent: 2},
```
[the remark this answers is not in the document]
p0mvn: I also prefer multi-laine for all. I think readability is preferable to having large files. We can extract these configs to JSON files if we end up having 50 validators

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +391,116 @@ func noRestart(config *docker.HostConfig) {
...
 }
 
-func (s *IntegrationTestSuite) initUpgrade() {
+func (s *IntegrationTestSuite) upgrade() {
 	// submit, deposit, and vote for upgrade proposal
-	s.submitProposal(s.chains[0])
-	s.submitProposal(s.chains[1])
+	currentHeightA := s.chainHeight(s.valResources[s.chains[0].ChainMeta.Id][0].Container.ID)
+	currentHeightB := s.chainHeight(s.valResources[s.chains[1].ChainMeta.Id][0].Container.ID)
+	propHeightA = currentHeightA + 40 + len(s.chains[0].Validators)
```
p0mvn, line 398: Is it possible to instead programmatically force the chain to pause at a certain height so that we can deterministically supply the desired upgrade height?

I don't know how to do it or if it is possible. The reason for this suggestion is to prevent having non-determinism such as where `currentHeightA` is different from each test run. That is likely to cause bugs in the future. From my experience, such things are very painful to debug. 

Also, the choice of 40. It might work today but not tomorrow if block time changes. That's why programmatically enforcing a certain height might be more stable long-term

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +391,116 @@ func noRestart(config *docker.HostConfig) {
 	}
 }
 
-func (s *IntegrationTestSuite) initUpgrade() {
+func (s *IntegrationTestSuite) upgrade() {
 	// submit, deposit, and vote for upgrade proposal
-	s.submitProposal(s.chains[0])
-	s.submitProposal(s.chains[1])
+	currentHeightA := s.chainHeight(s.valResources[s.chains[0].ChainMeta.Id][0].Container.ID)
```
p0mvn, line 396: Can we instead move all these calls to a for loop? We can iterate over all chains and do these actions instead of duplicating calls for every chain

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +391,116 @@ func noRestart(config *docker.HostConfig) {
...
-			s.Require().NoError(s.dkrPool.RemoveContainerByName(s.valResources[chain.ChainMeta.Id][i].Container.Name))
+			var opts docker.RemoveContainerOptions
+			opts.ID = s.valResources[chain.ChainMeta.Id][i].Container.ID
+			opts.Force = true
+			s.dkrPool.Client.RemoveContainer(opts)
+			s.T().Logf("removed container: %s", s.valResources[chain.ChainMeta.Id][i].Container.Name[1:])
 		}
 	}
+	// upgrade to locally compiled binary concurently
+	wg.Add(2)
```
p0mvn, line 454: Why do we have to do this concurrently?

Also, we should avoid hardcoding and duplicating calls

2 == len(s.chains)

and `go s.upgradeContainers(s.chains[0], &wg)` can be done in a loop

tests/e2e/e2e_test.go
```diff
@@ -1,135 +1,34 @@
...
-	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
-
 	"github.com/osmosis-labs/osmosis/v7/tests/e2e/chain"
-	"github.com/osmosis-labs/osmosis/v7/tests/e2e/util"
 )
 
-func (s *IntegrationTestSuite) Test001IBCTokenTransfer() {
-	var ibcStakeDenom string
-
+func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
```
p0mvn, line 9: Why do we not care about the order anymore?

p0mvn (changes requested):
Awesome job. Left some comments, please take a look.

The biggest question I have right now is why do we have to do certain things concurrently between chains? Why does it matter?

tests/e2e/e2e_setup_test.go
```diff
@@ -192,7 +188,7 @@ func (s *IntegrationTestSuite) runValidators(c *chain.Chain, portOffset int) {
 		s.Require().NoError(err)
 
 		s.valResources[c.ChainMeta.Id][i] = resource
-		s.T().Logf("started Osmosis %s validator container: %s", c.ChainMeta.Id, resource.Container.ID)
+		s.T().Logf("started %s validator container: %s", resource.Container.Name[1:], resource.Container.ID)
```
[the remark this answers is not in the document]
czarcas7ic: Container.Name starts with a "/", removing this provides a clean log of the container name

tests/e2e/e2e_setup_test.go
czarcas7ic: Can you point out which lines you are referring to?

tests/e2e/e2e_setup_test.go
czarcas7ic: Yes, after thinking about this for a while I think the method you suggested is better. What I liked about this method is it forced the chain to prove that it halted at the specified height, but what we can instead do is halt at the upgrade height and if on upgrade we dont hit blocks we can assume something wrong happened with the proposal

tests/e2e/e2e_setup_test.go
czarcas7ic: It's more of an opportunity than a need. If we have 100 validators on each chain, instead of looping through 200 validators we instead can do them at the same time since there is no risk of account sequence mismatches.

tests/e2e/e2e_setup_test.go
czarcas7ic: The purpose of this was to return True 3 times before accepting it as fact. For instance if upgrade height is 72 its possible to return 72 twice and still hit 73 but 3 seconds is too long to hit the same block 3 times in a row without it having been stopped for upgrade. But this is now moot since we will use a hard chain halt height as suggested by you in your previous comment

tests/e2e/e2e_setup_test.go
czarcas7ic: Same as before, if we have to upgrade 200 containers I would rather them be done at the same time instead of waiting for each to be finished since they are not dependent on each other.

tests/e2e/e2e_test.go
czarcas7ic: The previous implementation checked for Len==3, which is what we used as just a proof of concept. This function now stores the coin struct of the wallet pre send and post send and then compares them to each other. The difference between the structs should be the amount that was sent. If not, something went wrong.

tests/e2e/e2e_util_test.go
czarcas7ic: I like having both, the name is human readable and the container ID is used the check those containers logs and other docker functions. If you still would like me to remove one I can

tests/e2e/e2e_util_test.go
czarcas7ic: You need to hit a few blocks before the rpc server is started, otherwise the following test (ibc send) will fail. The three blocks here have passed every time and I have run it many times, so I feel safe with it hard coded. Lmk if you think of a better solution (I guess we could continue to ping the nodes rpc until it responds successfully, lmk if thats a better option)

tests/e2e/e2e_util_test.go
czarcas7ic: Without this, you run into account sequence mismatches. The only way around this would be to round robin multiple wallets on the hermes docker container for chain A and chain B instead of only using 1 for each.

czarcas7ic (commented):
Thank you so much for the thorough review @p0mvn ! Just had a few questions, the others I will address now!

tests/e2e/e2e_setup_test.go
```diff
@@ -391,84 +391,116 @@ func noRestart(config *docker.HostConfig) {
...
-	s.submitProposal(s.chains[1])
+	currentHeightA := s.chainHeight(s.valResources[s.chains[0].ChainMeta.Id][0].Container.ID)
+	currentHeightB := s.chainHeight(s.valResources[s.chains[1].ChainMeta.Id][0].Container.ID)
+	propHeightA = currentHeightA + 40 + len(s.chains[0].Validators)
+	propHeightB = currentHeightB + 40 + len(s.chains[1].Validators)
+
+	s.submitProposal(s.chains[0], propHeightA)
+	s.submitProposal(s.chains[1], propHeightB)
 	s.depositProposal(s.chains[0])
 	s.depositProposal(s.chains[1])
```
p0mvn, line 404:
```suggestion
        for _, chain := range chains {
           currentHeight := s.chainHeight(s.valResources[chain.ChainMeta.Id][0].Container.ID)
           propHeight := currentHeight + 40 + len(chain.Validators)
           s.submitProposal(chain, propHeight)
	  s.depositProposal(chain)
        }
```

tests/e2e/e2e_setup_test.go
p0mvn: Similar idea to: https://github.com/osmosis-labs/osmosis/pull/1458/files/dace939d3b2461b62fbd5c407393b6905339a21d#r869717359

tests/e2e/e2e_setup_test.go
p0mvn: https://github.com/osmosis-labs/osmosis/pull/1458/files/dace939d3b2461b62fbd5c407393b6905339a21d#r869717359

tests/e2e/e2e_setup_test.go
p0mvn: I think that my earlier proposed solution would functionally the same way as right now, just without a counter. But yeah, if it's going to get change - it doesn't matter :)

tests/e2e/e2e_setup_test.go
p0mvn: Do we have plans to run that many, and would the CI even handle it?

The reason I'm careful about concurrency in CI is that it always adds a layer of complexity. Since we are not unit testing this setup code directly, it's easy to miss something and struggle in the future with "unintended consequences".

Imo CI should be as simple as possible and things like concurrencyy added when there is a concrete need for it.

I don't have a strong opinion on this. Up to you to decide. Just sharing my thoughts :+1:

tests/e2e/e2e_test.go
p0mvn: Awesome, nice wok!

tests/e2e/e2e_setup_test.go
czarcas7ic: Via DMs with @p0mvn 

Decided to more deterministically predict how long it takes to create, deposit, and vote for a proposal. This will translate into a variable voting period that will get created via a flag in the init container

tests/e2e/e2e_setup_test.go
alexanderbez: If voting doesn't take that much time (e.g. <1m) we should probably just make it sequential.

tests/e2e/e2e_util_test.go
czarcas7ic: So I tested out the round robin wallet idea. While it ended up working, it felt overly convoluted. I realized the best solution was an extremely obvious one; spam txs with a single account until you get a successful send code. Will post what I have soon

tests/e2e/e2e_setup_test.go
czarcas7ic: Okay, I reverted it back to sequential. The only downside is that VotingPeriod for B now has to be the originally calculated VotingPeriod for B plus VotingPeriod for A since A votes first now, but its not a huge deal.

czarcas7ic:
@alexanderbez @p0mvn I addressed a majority of your comments, however one or two might not be addressed since I arranged how some of works now. 

The most drastic of changes is that IBC sends now continually send until a success code is received. Additionally, there is a check to ensure that IBC send actually arrives on the destination chain before continuing. This has mitigated all issues I was having with account sequence mismatching. Please let me know what you think when you have the time.

tests/e2e/chain/config.go
```diff
@@ -130,7 +133,7 @@ func addAccount(path, moniker, amountStr string, accAddr sdk.AccAddress) error {
 	return genutil.ExportGenesisFile(genDoc, genFile)
 }
 
-func initGenesis(c *internalChain) error {
+func initGenesis(c *internalChain, votingPeriod time.Duration) error {
```
alexanderbez, line 136: In the future, this can get messy. We should revise this such that we can instead pass the entire module genesis state. But for now, this is OK.

p0mvn (approved):

PR #1469, opened 2022-05-11
Change proto service for Balancer Msg

mattverse:
Closes: [#1185](https://github.com/osmosis-labs/osmosis/issues/1185)

With proto Msg service defined twice(both in Balancer proto and gamm package proto), this is currently causing problems in typescript proto generation. 

As @ValarDragon mentioned in https://github.com/osmosis-labs/osmosis/pull/1208,
> They existed at the same import path, because due to interface serialization
depending on the amino route (which depended on the proto import path),
changing the import path requires a state migration thats a bit cumbersome to
do.

This PR solves this by making txs under the correct package, but leaving balacner pool proto definition under the legacy wrong one, `package osmosis.gamm.v1beta1`. Solving the problem this way allows us to solve without having to need state breaking changes or additional migration for existing pools.

In addition to solving the exisitng problem of duplicated Msg in one package, this PR also changes stableswap pool proto package to the correct package.

## Brief Changelog
- Change proto package for balancer msg

  - Does this pull request introduce a new feature or user-facing behavior changes? maybe


proto/osmosis/gamm/pool-models/balancer/tx.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package osmosis.gamm.v1beta1;
+package osmosis.gamm.poolmodels.balancer;
```
ValarDragon, line 2: should we still have the `v1beta1`?

alexanderbez: It should be versioned, yes.

alexanderbez (approved):

ValarDragon (approved):

PR #1473, opened 2022-05-11
hack: replace superfluid intermediary account hash(denom + old acc) with hash(denom + new acc) for init genesis

p0mvn:
Closes: #XXX

DO NOT MERGE

Superfluid intermediary account address = hash(denom + validator address) ([link](https://github.com/osmosis-labs/osmosis/blob/2d8d32c5e28779bc059768ce47f4463e88e6d6c1/x/superfluid/types/superfluid.go#L38))

To spin up a testnet, we export genesis and replace a validator address to the one we have control of in the following script:
https://github.com/osmosis-labs/osmosis/blob/84e7726a4de52539b0fe3c0c5fc68b8d84131835/cmd/osmosisd/cmd/testnetify/testnetify.py#L102

However, since Superfluid intermediary account address = hash(denom + validator address), that does not get replaces since this address is a hash of 2 values where one gets changed.

As a result, we need a way to programmatically replace Superfluid intermediary account with a new version so that we can initialize genesis from genesis that had validator addresses replaced.

This is only a temporary solution to be able to spin up a testnet. We should think of a better way to change Superfluid intermediary account addresses in the future via testnetify script.

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/x/superfluid/genesis.go
+++ b/x/superfluid/genesis.go
@@ -1,6 +1,8 @@
 package superfluid
 
 import (
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/osmosis-labs/osmosis/v7/x/superfluid/keeper"
 	"github.com/osmosis-labs/osmosis/v7/x/superfluid/types"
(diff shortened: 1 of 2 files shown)


ValarDragon:
Lets make an issue describing this + alternate options for making EditValidatorMsg's / if thats a priority.

ValarDragon:
Closing, some context in #1655.

PR #1480, opened 2022-05-13
feat: Enable Unpooling for whitelisted pools

mattverse:
Closes: prop226

**This PR is to get readiness for the circumstances when prop 226 has been passed via governance.** 

The procedures to unpooling existing locks is as follows:
- Check if the pool is whitelisted for unpooling
- Check if the lock has been superfluid staked 
   - if so, superfluid undelegate, superfluid unbond
 - Finish unlocking directly for locked locks
 - Run ExitPool for the gamm share from the lock
 - Check if the lock was already unbonding previously
    - if so, run logic to preserve unbonding end time
    - if not, start initial unbonding
 - if not, create lock for the coins from ExitPool
 
 
 
The logic for this is built-in `x/superfluid`, mainly because of the restrictions caused due to go lang import cycle, where we have the architecture of superfluid module importing both the gamm module and the lockup module. Since Unpooling logic requires both functionalities of superfluid, gamm, lockup module, decision made to place logic for unpooling within the superfluid module.

This implementation includes one additional state entry, where we would keep state of whitelisted pools that are allowed to unpool directly, and a Tx Msg to trigger unpooling.

## Brief Changelog

  - *Allow unpooling for whitelisted pools*

## Testing and Verifying

TestCases WIP


x/superfluid/keeper/msg_server.go
```diff
@@ -97,3 +97,27 @@ func (server msgServer) LockAndSuperfluidDelegate(goCtx context.Context, msg *ty
 		ID: lockupRes.ID,
 	}, err
 }
+
+func (server msgServer) UnPoolWhitelistedPool(goCtx context.Context, msg *types.MsgUnPoolWhitelistedPool) (*types.MsgUnPoolWhitelistedPoolResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
```
ValarDragon, line 103: Lets add a height gating here. `if ctx.BlockHeight() < vx.UpgradeHeight  {return nil, errors.New("message not activated"))}`

Will need to confirm thats state compatible until upgrade height

x/superfluid/keeper/unpool.go
```diff
@@ -0,0 +1,108 @@
...
+	if !allowed {
+		return 0, types.ErrUnpoolNotAllowed
+	}
+
+	// check if lock is unlocking
+	lock, err := k.lk.GetLockByID(ctx, lockId)
+	if err != nil {
+		return 0, err
+	}
+
```
Thunnini, line 33: I think we should check the sender and owner of lock is same.

app/upgrades/vx/forks.go
```diff
@@ -0,0 +1,29 @@
...
+func RunForkLogic(ctx sdk.Context, keepers *keepers.AppKeepers) {
+	ctx.Logger().Info("Applying Osmosis vx upgrade." +
+		"Allowing direct unbonding for whitelisted pools")
+	RegisterWhitelistedDirectUnbondPools(ctx, keepers.SuperfluidKeeper)
+}
+
+// RegisterWhitelistedDirectUnbondPools registers pools that are allowed to unpool
+// https://www.mintscan.io/osmosis/proposals/226
+func RegisterWhitelistedDirectUnbondPools(ctx sdk.Context, superfluid *superfluidkeeper.Keeper) {
+	whitelistedPoolShares := []int64{1, 2, 3, 4}
```
ValarDragon, line 21: TODO here right?
```suggestion
	// TODO: Get from proposal
	whitelistedPoolIDs := []int64{1, 2, 3, 4}
```

x/superfluid/keeper/unpool.go
```diff
@@ -0,0 +1,108 @@
...
+	newLock, err := k.lk.CreateLock(ctx, sender, exitCoins, lock.Duration)
+	if err != nil {
+		return 0, err
+	}
+
+	initializedTime := time.Time{}
+	// if the lock was unlocking, run separate logic to preserve lock endTime
+	if lock.EndTime != initializedTime {
```
Thunnini, line 71: I think `initializedTime` variable is not neccessary, we can check the Time struct is initialized by checking "if lock.EndTime != (time.Time{})"

x/superfluid/keeper/unpool.go
```diff
@@ -0,0 +1,108 @@
...
+		err = k.lk.BeginForceUnlockWithEndTime(ctx, lockId, lock.EndTime)
```
Thunnini, line 72: If I understand your intention properly, I think we should force unlock `newLock.ID`

Thunnini (changes requested):
- Register new msgs to codec.go

Superfluid is too complex to do a brief review, so I will continue to look around.
For now, please check my current comments first.

docs/core/proto-docs.md
```diff
@@ -3935,6 +3953,38 @@ specified validator addr.
...
+
+### MsgUnPoolWhitelistedPool
+===================== MsgUnPoolWhitelistedPool
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| `sender` | [string](#string) |  |  |
+| `poolId` | [uint64](#uint64) |  |  |
+| `lockId` | [uint64](#uint64) |  |  |
```
mconcat, line 3967: The proposal does not state whether the msg should unpool the entire set of lockups or not. However it might be more straightforward to unpool the entire set of lockup for an account. There could be multiple locks for a single gamm denom per bonding period.

x/lockup/keeper/lock.go
```diff
@@ -493,6 +493,41 @@ func (k Keeper) beginForceUnlock(ctx sdk.Context, lock types.PeriodLock, coins s
...
+	lock, err := k.GetLockByID(ctx, lockID)
+	if err != nil {
+		return err
+	}
+	return k.beginForceUnlockWithEndTime(ctx, *lock, endTime)
+}
+
+func (k Keeper) beginForceUnlockWithEndTime(ctx sdk.Context, lock types.PeriodLock, endTime time.Time) error {
+	// remove lock refs from not unlocking queue if exists
+	err := k.deleteLockRefs(ctx, types.KeyPrefixNotUnlocking, lock)
```
mconcat, line 506: BeginForceUnlockWithEndTime is called when the lock is unlocking(in the UnpoolAllowedPool function if-then-else branch), but KeyPrefixNotUnlocking is used here

EDIT: no this is right because newLock is provided which is just created with notunlocking

x/superfluid/keeper/unpool.go
```diff
@@ -0,0 +1,108 @@
...
+	lock, err := k.lk.GetLockByID(ctx, lockId)
+	if err != nil {
+		return 0, err
+	}
+
+	gammShare := lock.Coins[0]
+	if gammShare.Denom != gammtypes.GetPoolShareDenom(poolId) {
+		return 0, types.ErrLockUnpoolNotAllowed
+	}
+
```
ValarDragon, line 38:
```suggestion
// Steps for unpooling
// 1) If superfluid delegated, superfluid undelegate
// 2) Break underlying lock. This will clear any metadata if things are superfluid unbonding
// 3) Get duration from {} (Consider if we can handle complexity for already unbonding Locks)
// 4) ExitPool with these unlocked LP shares
// 5) Make 1 new lock for every asset in collateral. Many code paths need this assumption to hold
// 6) Make new lock begin unlocking
```

ValarDragon: Then clarify / perhaps document each component further. Worth considering splitting them into separate functions

x/superfluid/keeper/unpool.go
ValarDragon: I'd recommend the variable, but it being `defaultInitializedTime` perhaps with a line link to the relevant part of lockup that initalizes it.

x/superfluid/keeper/unpool.go
```diff
@@ -0,0 +1,108 @@
...
+		if poolId == allowedPoolId {
+			allowed = true
+		}
+	}
+
+	if !allowed {
+		return 0, types.ErrUnpoolNotAllowed
+	}
+
+	// check if lock is unlocking
```
mconcat, line 28:
```suggestion
```

x/lockup/keeper/lock.go
```diff
@@ -525,6 +560,10 @@ func (k Keeper) ForceUnlock(ctx sdk.Context, lock types.PeriodLock) error {
 	return k.unlockInternalLogic(ctx, lock)
 }
 
+func (k Keeper) BreakLockForUnpool(ctx sdk.Context, lock types.PeriodLock) error {
+	return k.unlockInternalLogic(ctx, lock)
```
mconcat, line 564: UnlockInternalLogic assumes the lock is unlocking(see KeyPrefixUnlocking usage). This is not always true for unpooling.

mconcat (commented):
first iteration

mattverse:
Resloving currently opened comments, closing this PR, opening a new one based off v8.x

x/superfluid/keeper/msg_server.go
mattverse: yeah good catch! We can't use vx package directly due to import cycle, going to put this as antoher constant in keeper for now

x/superfluid/keeper/unpool.go
mattverse: Tysm for this!

docs/core/proto-docs.md
mattverse: I think adding an additional logic for partial unlockings is too over-engineering atm. For having multiple locks for a single denom per bonding period, it's not a problem since we're getting passed the lockID as well

mconcat: agreed with partial unpooling, we dont need to support it. but could we make lockID a repeated field so there can be a single transaction that unpools the entire pool?

x/lockup/keeper/lock.go
mattverse: good catch!

mattverse:
Closing this PR since branch needs to be made off v8.x

mattverse:
Closing this PR since branch needs to be made off v8.x

mattverse:
replaced with https://github.com/osmosis-labs/osmosis/pull/1481

PR #1483, opened 2022-05-13
feat: add height-gated AnteHandler message filter

alexanderbez:
Add an AnteHandler decorator that provides (height-gated) message filtering for the v8 upgrade.

### DIFF
--- a/app/upgrades/v8/msg_filter_ante.go
+++ b/app/upgrades/v8/msg_filter_ante.go
@@ -1,0 +1,36 @@
+package v8
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	superfluidtypes "github.com/osmosis-labs/osmosis/v7/x/superfluid/types"
+)
+
+// MsgFilterDecorator defines an AnteHandler decorator for the v8 upgrade that
+// provide height-gated message filtering acceptance.
+type MsgFilterDecorator struct{}
+
+// AnteHandle performs an AnteHandler check that returns an error if the current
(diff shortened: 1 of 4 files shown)


ValarDragon (approved):

app/upgrades/v8/msg_filter_ante.go
```diff
@@ -0,0 +1,36 @@
...
+// provide height-gated message filtering acceptance.
+type MsgFilterDecorator struct{}
+
+// AnteHandle performs an AnteHandler check that returns an error if the current
+// block height is less than the v8 upgrade height and contains messages that are
+// not supported until the upgrade height is reached.
+func (mfd MsgFilterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	currHeight := ctx.BlockHeight()
+
+	if currHeight < UpgradeHeight && hasInvalidMsgs(tx.GetMsgs()) {
```
ValarDragon, line 20:
```suggestion
	if currHeight < v8.UpgradeHeight && hasInvalidMsgs(tx.GetMsgs()) {
```

x/superfluid/keeper/msg_server.go
```diff
@@ -124,5 +124,6 @@ func (server msgServer) UnPoolWhitelistedPool(goCtx context.Context, msg *types.
 		),
 	})
 
+	x := &typesMsgUnPoolWhitelistedPool{}
```
ValarDragon, line 127:
```suggestion
```

PR #1484, opened 2022-05-14
Lock tokens via cosmwasm

nicolaslara:

This adds support for locking to cosmwasm. The code currently works but it's not in a state where it can be merged. This PR is mostly to ask if this is of interest and if I should go through and clean it up, add tests, docs, etc. 

For this to be usable, some simple changes to https://github.com/confio/osmosis-bindings are needed. I've added a [PR](https://github.com/confio/osmosis-bindings/pull/28) there too. 

## Brief Changelog

 - Expose the lock-tokens endpoint to cosmwasm

## Testing and Verifying

Tests are needed and I will include them if there is interest in getting this (or something similar) merged.

  - Does this pull request introduce a new feature or user-facing behavior changes? YES, but only exposed via cosmwasm
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? Not yet
  - How is the feature or change documented? Not documented yet, but should be under the cosmwasm bindings documentation.


ValarDragon:
Whats the motivation for this vs the create lock message?

nicolaslara:
@ValarDragon AFAICT the locking messages are not exposed to wasm (https://github.com/osmosis-labs/osmosis/blob/28761d0b69509d1fc3513ded8a513c178787630d/app/keepers/keepers.go#L327), but maybe they get registered automatically. Did I miss something?

ValarDragon:
Ah they are exposed to wasm, but this is still useful. I blanked when I replied, that these bindings serve the dual purpose of decoupling cosmwasm contracts from things like proto type routes, and other oddities, for the SDK messages.

nicolaslara:
Right, they are "not exposed" was probably not the right way to phrase it either. Just don't have a clean interface. 

Are all modules exposed by default? I don't see any config specifying which ones should be accessible from cosmwasm. 

This may not be the case for osmosis, but I can imagine some chains may want to have privileged messages that can't be accessed by contracts.

ValarDragon:
All SDK messages as far as I know are exposed to cosmwasm, but these bindings give better APIs that don't break with every upgrade.

I now think the lock tokens binding approach makes sense!

nicolaslara:
Sounds good. I'll clean this up so it can be merged then. Might add some of the other endpoints if I have time too.

ValarDragon:
Any chance you could move your cosmwasm binding PR to [osmosis-labs/osmosis-bindings](https://github.com/osmosis-labs/osmosis-bindings) ?

Sorry for that inconvenience, I just cloned and remade a repo, rather than transferring it

nicolaslara:
Just saw that. I'll shuffle my git branches to move the PR there.

app/keepers/keepers.go
```diff
@@ -318,7 +318,7 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 	// if we want to allow any custom callbacks
 	supportedFeatures := "iterator,staking,stargate,osmosis"
 
-	wasmOpts = append(owasm.RegisterCustomPlugins(appKeepers.GAMMKeeper, appKeepers.BankKeeper, appKeepers.TokenFactoryKeeper), wasmOpts...)
+	wasmOpts = append(owasm.RegisterCustomPlugins(appKeepers.GAMMKeeper, appKeepers.BankKeeper, appKeepers.TokenFactoryKeeper, appKeepers.LockupKeeper), wasmOpts...)
```
alexanderbez, line 321: Can we break this up onto multi-line to make it easier to read what's being passed in?

app/wasm/bindings/msg.go
```diff
@@ -23,3 +30,32 @@ type SwapMsg struct {
 	Route  []Step              `json:"route"`
 	Amount SwapAmountWithLimit `json:"amount"`
 }
+
+type Duration time.Duration
```
alexanderbez, line 34: Why is there an alias like this?

app/wasm/test/custom_msg_test.go
```diff
@@ -544,3 +543,60 @@ func executeCustom(t *testing.T, ctx sdk.Context, osmosis *app.OsmosisApp, contr
 	_, err = contractKeeper.Execute(ctx, contract, sender, reflectBz, coins)
 	return err
 }
+
+//func TestLockTokensMsg(t *testing.T) {
```
alexanderbez, line 547: Is there a reason this is commented out?

nicolaslara: Basically cause I haven't finished writing the tests. I'm having some issues with the wasm bindings used as testdata (https://github.com/osmosis-labs/osmosis/pull/1445#issuecomment-1135695165). I should have some free time to look into this this week.

alexanderbez: Gotcha. Should be move this into Draft mode until it's all good-to-go?

app/wasm/bindings/msg.go
nicolaslara: That was my way of adding marshal/unmarshal to Duration (https://stackoverflow.com/a/48051946). Happy to change it if there's a better way.

nicolaslara:
Spent some time experimenting with different approaches to the test to see if I could get it to pass. Still stuck with the same two errors though.

If I don't replace `testdata/osmo_reflect.wasm`, then I get 

```
=== RUN   TestLockTokensMsg
    custom_msg_test.go:641:
        	Error Trace:	custom_msg_test.go:641
        	Error:      	Received unexpected error:

        	            	github.com/CosmWasm/wasmd/x/wasm/keeper.Keeper.execute
        	            		/Users/nicolas/go/pkg/mod/github.com/osmosis-labs/wasmd@v0.27.0-rc2.0.20220517191021-59051aa18d58/x/wasm/keeper/keeper.go:354
        	            	github.com/CosmWasm/wasmd/x/wasm/keeper.PermissionedKeeper.Execute
        	            		/Users/nicolas/go/pkg/mod/github.com/osmosis-labs/wasmd@v0.27.0-rc2.0.20220517191021-59051aa18d58/x/wasm/keeper/contract_keeper.go:51
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.executeCustom
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_msg_test.go:544
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.TestLockTokensMsg
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_msg_test.go:640
        	            	Error parsing into type osmo_reflect::msg::ExecuteMsg: unknown variant `lock_tokens`, expected `mint_tokens` or `swap`: execute wasm contract failed
        	Test:       	TestLockTokensMsg
--- FAIL: TestLockTokensMsg (0.35s)
FAIL
FAIL	github.com/osmosis-labs/osmosis/v7/app/wasm/test	0.980s
FAIL
```

 which makes to some extent sense to me, since the contract old `.wasm` contract doesn't know about `lock_tokens`.

If I compile the contract and replace it under `testdata` I get 

```
=== RUN   TestLockTokensMsg
    custom_query_test.go:312:
        	Error Trace:	custom_query_test.go:312
        	            				custom_query_test.go:37
        	            				custom_msg_test.go:608
        	Error:      	Received unexpected error:

        	            	github.com/cosmos/cosmos-sdk/x/gov/keeper.Keeper.SubmitProposal
        	            		/Users/nicolas/go/pkg/mod/github.com/osmosis-labs/cosmos-sdk@v0.45.1-0.20220517190359-30ebc413ddff/x/gov/keeper/proposal.go:24
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.storeReflectCode
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_query_test.go:311
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.SetupCustomApp
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_query_test.go:37
        	            	github.com/osmosis-labs/osmosis/v7/app/wasm/test.TestLockTokensMsg
        	            		/Users/nicolas/devel/osmosis/app/wasm/test/custom_msg_test.go:608
        	            	code bytes cannot be longer than 819200 bytes: exceeds limit: invalid request: invalid proposal content
        	Test:       	TestLockTokensMsg
--- FAIL: TestLockTokensMsg (0.07s)
FAIL
FAIL	github.com/osmosis-labs/osmosis/v7/app/wasm/test	0.819s
FAIL
```

That breaks the other tests too (Mint, for example). From the error, it almost seems like the gov module changed in a way that the version of cosmwasm used in `osmosis-bindings` doesn't support. 

If anyone has guesses as to what may be happening here, I'd appreciate them. Maybe @ValarDragon or @alexanderbez ?

app/wasm/test/custom_msg_test.go
nicolaslara: Fixed the tests. Might still be good to write some more, but closing this conversation for now

nicolaslara:
Moving out of draft now that I've managed to fix the tests.

ValarDragon:
Hey! Thanks so much for the work on this! Closing for now, as we want to go with an approach of using StargateMsg and StargateQuery. This is because we don't want to maintain a separate interface layer in go just for cosmwasm -- we should just promise interface compatability at one point, or make a native backwards compatible abstraction.

I think you did a great job on this, and if we were going wtih the binding-for-everything approach should definitely merge this.

nicolaslara:
@ValarDragon did some research here and `StargateQuery` is disabled by default on wasmd "because of non-determinism" (though I'm still trying to figure out the exact reason on the cosmwasm discord). This means that there will still be a need for changes that either (i) activates StargateQuery with code similar to what existed in wasmd before it was disabled, or (ii) creates custom bindings for all queries which will have to be maintained. 

If I understand correctly the reason why this was removed from wasmd, then there should be no issue in adding it back on osmosis since cosmwasm is permissioned, and code with non-deterministic queries can be denied (or we could just have a list of non-deterministic queries that are excluded from cosmwasm calls). But I'll get to the bottom of the issue with the queries and report back when I know more.

Issue #1225, opened 2022-04-07
Make pool API fulfilling methods all in single file


ValarDragon:
At the moment, methods that fulfill the Pool API for balancer are split between amm.go and balancer_pool.go.

We should change this to having all pool interface methods in a single `pool.go`.

All the AMM / math methods should call out to a separate private function inside of `amm.go`. The top -level exported method can handle some of the input sanitization, so that the private methods can assume cleaner invariants though.

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/1214#discussion_r845584996_

PR #1498, opened 2022-05-16
refactor: consolidate pool implementation

alexanderbez:
Closes: #1225

- Move balancer and amm logic into a single `pool.go` file
- Move `PoolParams` to `pool_params.go` file
- Update test package to `balancer_test`
- Make `SetInitialPoolAssets` a public API (to satisfy the `TestBalancerPoolUpdatePoolAssetBalance` test)


alexanderbez:
@ValarDragon the question remains, do we want `pool_test.go` to be in package `balancer` or `balancer_test`? If the latter, we need to make `Pool#setInitialPoolAssets` a public API.

ValarDragon:
I'm down for it to be in either package tbh

x/gamm/pool-models/balancer/pool_params.go
```diff
@@ -0,0 +1,78 @@
+package balancer
```
mattverse, line 1: I love how we made a separate file for pool params!

mattverse (commented):
Per deleting `amm.go`, did we decide to remove separating the amm logic and the pool logic by using two different files `amm.go` and `pool.go`?

mattverse (changes requested):
Can we also change stableswap package from stableswap_pool.go -> pool.go to better achieve consolidation between the packages? I'm also down to leaving balancer_pool.go

alexanderbez:
I'm going to need to rebase this again unfortunately, a bit annoying but w.e. The idea is just to consolidate the pool logic into a single file along with the tests.

mattverse:
Oh interesting, I thought we were trying to separate out the pool methods and the core math logic in two different files? 

If we were to change the decision to move towards a single pool.go file, I think we should make the same changes in stableswap package as well or do you think that change is out of scope for this PR?

alexanderbez:
Conflicts are too large here to warrant resolving them. Instead re-opened a new PR: https://github.com/osmosis-labs/osmosis/pull/1868

Issue #1460, opened 2022-05-10
[x/gamm][StableSwap - Post V8] Add stableswap `scaling factor governor`


ValarDragon:
There is a desire to use stableswaps with yield accumulating assets. There are a few designs for how you could do this, but the simplest one seems to be to utilize the 'scaling factors' present in the Stableswap algorithm. If someone wanted to adjust the scaling factors according to yield, then this would achieve altering the price that its being stable around.

We should add a new field in the stableswap proto message, called `scaling_factor_governor`, which can be either blank, or an osmo address.

We should also add a new message called "MsgStableSwapAdjustScalingFactors", that can only be sent from the `scaling_factor_governor` address for the respective poolID. (if the address is blank, noone can send this message for the given poolID). This message adjusts the scaling factors to the provided ones.

It is intentioned that cosmwasm contracts could be used to set these in the future. There should also be a query for this, as it will be important for front-ends / users to know about a governor risk prior to LP'ing.

This is a post v8 feature

vuong177:
Could i try doing that ?

PR #1499, opened 2022-05-16
Scaling factor governor

vuong177:
Closes: #1460

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


x/gamm/keeper/msg_server.go
```diff
@@ -280,3 +280,13 @@ func (server msgServer) ExitSwapShareAmountIn(goCtx context.Context, msg *types.
...
+
+// func (server msgServer) StableSwapAdjustScalingFactors(goCtx context.Context, msg *stableswap.MsgStableSwapAdjustScalingFactors) (*stableswap.MsgStableSwapAdjustScalingFactorsResponse, error) {
+// 	ctx := sdk.UnwrapSDKContext(goCtx)
+
+// 	if err := server.keeper.setStableSwapScalingFactors(ctx, msg.ScalingFactors, msg.PoolID, msg.ScalingFactorGovernor); err != nil {
+// 		return nil, err
+// 	}
+
+// 	return &stableswap.MsgStableSwapAdjustScalingFactorsResponse{}, nil
+// }
```
ValarDragon, line 292: Why's this commented out?

proto/osmosis/gamm/pool-models/stableswap/tx.proto
```diff
@@ -29,3 +31,16 @@ message MsgCreateStableswapPool {
 message MsgCreateStableswapPoolResponse {
   uint64 pool_id = 1 [ (gogoproto.customname) = "PoolID" ];
 }
+
+message MsgStableSwapAdjustScalingFactors {
+  string scaling_factor_governor = 1
+      [ (gogoproto.moretags) = "yaml:\"scaling_factor_governor\"" ];
```
ValarDragon, line 37:
```suggestion
  // Sender must be the pool's scaling_factor_governor in order for the tx to succeed
  string sender = 1
      [ (gogoproto.moretags) = "yaml:\"sender\"" ];
```

ValarDragon (commented):
Awesome, thanks so much for adding this!

Want to just add a test that ensures the factors do get updated in state, then we can merge it in?

x/gamm/keeper/msg_server.go
vuong177: Because stableswap msg server still disable in main branch. I think better if we comment out that code !

ValarDragon: Fair enough

vuong177:
I added test but commented it out because stableswap still disable

x/gamm/keeper/msg_server.go
```diff
@@ -280,3 +280,13 @@ func (server msgServer) ExitSwapShareAmountIn(goCtx context.Context, msg *types.
...
 }
+
+// func (server msgServer) StableSwapAdjustScalingFactors(goCtx context.Context, msg *stableswap.MsgStableSwapAdjustScalingFactors) (*stableswap.MsgStableSwapAdjustScalingFactorsResponse, error) {
+// 	ctx := sdk.UnwrapSDKContext(goCtx)
+
+// 	if err := server.keeper.setStableSwapScalingFactors(ctx, msg.ScalingFactors, msg.PoolID, msg.ScalingFactorGovernor); err != nil {
+// 		return nil, err
+// 	}
+
+// 	return &stableswap.MsgStableSwapAdjustScalingFactorsResponse{}, nil
```
mattverse, line 291: `stableswap` -> `types`

vuong177: Is this necessary ? I see we use `stableswap` in function `CreateStableswapPool` (line 51).

mattverse: yeah my bad it looks fine 👍

mattverse (approved):
LGTM

Issue #1448, opened 2022-05-08
Add tests for Binary Search


mattverse:
There's a room for improvement for test for binary search introduced in https://github.com/osmosis-labs/osmosis/pull/1419. 
Some extra testing that we need, originally mentioned by @p0mvn  would be 

- testing binary search on more complex function(currently testing with a straight line function)
- Non-zero tolerance test cases

p0mvn:
Hi @mattverse ! Thanks for creating this issue, could you please add it to: https://github.com/osmosis-labs/osmosis/issues/1452

Also, please add a "T:task" label

PR #1524, opened 2022-05-17
Stableswap Test additions

xBalbinus:
Closes: #1448

This was meant to be added on along with the Stableswap LP logic PR but accidentally got left out.

## Brief Changelog

*(for example:)*
 
  - Added additional non-linear function and non-linear non-zero-tolerance tests to stableswap logic.

## Testing and Verifying

This change added tests and can be verified via. make all

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / **no**)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / **no**)
  - How is the feature or change documented? (**not applicable**   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

### DIFF
--- a/x/tokenfactory/keeper/admins.go
+++ b/x/tokenfactory/keeper/admins.go
@@ -19,8 +19,8 @@ func (k Keeper) GetAuthorityMetadata(ctx sdk.Context, denom string) (types.Denom
 	return metadata, nil
 }
 
-// SetAuthorityMetadata stores authority metadata for a specific denom
-func (k Keeper) SetAuthorityMetadata(ctx sdk.Context, denom string, metadata types.DenomAuthorityMetadata) error {
+// setAuthorityMetadata stores authority metadata for a specific denom
+func (k Keeper) setAuthorityMetadata(ctx sdk.Context, denom string, metadata types.DenomAuthorityMetadata) error {
 	err := metadata.Validate()
 	if err != nil {
 		return err
@@ -45,5 +45,5 @@ func (k Keeper) setAdmin(ctx sdk.Context, denom string, admin string) error {
 
 	metadata.Admin = admin
 
-	return k.SetAuthorityMetadata(ctx, denom, metadata)
+	return k.setAuthorityMetadata(ctx, denom, metadata)
 }
--- a/x/tokenfactory/keeper/createdenom.go
+++ b/x/tokenfactory/keeper/createdenom.go
@@ -44,7 +44,7 @@ func (k Keeper) CreateDenom(ctx sdk.Context, creatorAddr string, denomNonce stri
 	authorityMetadata := types.DenomAuthorityMetadata{
 		Admin: creatorAddr,
 	}
-	err = k.SetAuthorityMetadata(ctx, denom, authorityMetadata)
+	err = k.setAuthorityMetadata(ctx, denom, authorityMetadata)
 	if err != nil {
 		return "", err
 	}
--- a/x/tokenfactory/keeper/genesis.go
+++ b/x/tokenfactory/keeper/genesis.go
@@ -1,15 +1,14 @@
-package tokenfactory
+package keeper
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
-	"github.com/osmosis-labs/osmosis/v7/x/tokenfactory/keeper"
 	"github.com/osmosis-labs/osmosis/v7/x/tokenfactory/types"
 )
 
 // InitGenesis initializes the tokenfactory module's state from a provided genesis
 // state.
-func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
+func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
 	k.CreateModuleAccount(ctx)
 
 	if genState.Params.DenomCreationFee == nil {
@@ -26,7 +25,7 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) 
 		if err != nil {
 			panic(err)
 		}
-		err = k.SetAuthorityMetadata(ctx, genDenom.GetDenom(), genDenom.GetAuthorityMetadata())
+		err = k.setAuthorityMetadata(ctx, genDenom.GetDenom(), genDenom.GetAuthorityMetadata())
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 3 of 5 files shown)


osmoutils/binary_search_test.go
```diff
@@ -41,10 +41,11 @@ func TestBinarySearch(t *testing.T) {
...
 
 func TestBinarySearchNonlinear(t *testing.T) {
-	// straight line function that returns input. Simplest to binary search on,
+	// x^2 function, a test example for a monotonic increasing function
 	// binary search directly reveals one bit of the answer in each iteration with this function.
-	lineF := func(a sdk.Int) (sdk.Int, error) {
-		return a, nil
+	expF := func(a sdk.Int) (sdk.Int, error) {
+		output := a.MulRaw(2)
+		return output, nil
```
ValarDragon, line 48: I'm confused, this replaced a line, with another line, not x^2 or an exponential function?

xBalbinus:
Oh shoot that's totally my bad on the comment there. The exponential is
supposed to be for the non-linear non-zero tolerance test! Changing.

xBalbinus:
Actually never mind I think I completely messed up with the line not sure why I did that.

Function has been changed to x^3 just like the non-zero tolerance test

osmoutils/binary_search_test.go
```diff
@@ -41,10 +41,13 @@ func TestBinarySearch(t *testing.T) {
...
-	// straight line function that returns input. Simplest to binary search on,
+	// x^2 function, a test example for a monotonic increasing function
 	// binary search directly reveals one bit of the answer in each iteration with this function.
-	lineF := func(a sdk.Int) (sdk.Int, error) {
-		return a, nil
+	expF := func(a sdk.Int) (sdk.Int, error) {
+		calculation := sdk.Dec(a)
+		result := calculation.Power(3)
+		output := sdk.Int(result)
+		return output, nil
```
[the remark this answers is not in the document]
ValarDragon: (So two functions being tested)

mattverse (commented):
Can we add comments to the newly added testing functions how the hard coded numbers are resulted? (e.g `{"Nonzero tolerance: =", NonZeroErrTolerance, sdk.NewInt(1011), sdk.NewInt(1001), 0},`) IMO documenting how the hard coded numbers are resulted in test cases are crucial for code maintainence

We can also create a separate issue for this, I dont want to block merging :p

xBalbinus:
No prob I can actually add this in pre merging of this PR, wouldn’t be much
of a problem.

xBalbinus:
Updates made!

osmoutils/binary_search_test.go
```diff
@@ -41,11 +48,17 @@ func TestBinarySearch(t *testing.T) {
 }
 
 func TestBinarySearchNonlinear(t *testing.T) {
-	// straight line function that returns input. Simplest to binary search on,
-	// binary search directly reveals one bit of the answer in each iteration with this function.
 	lineF := func(a sdk.Int) (sdk.Int, error) {
 		return a, nil
 	}
+	// x^2 function, a test example for a monotonic increasing function
```
p0mvn, line 54:
```suggestion
	// x^3 function, a test example for a monotonic increasing function
```

osmoutils/binary_search_test.go
```diff
@@ -130,7 +154,7 @@ func TestErrTolerance_Compare(t *testing.T) {
 }
 
 func TestErrToleranceNonzero_Compare(t *testing.T) {
-	// Nonzero error tolerance test
+	// Nonzero error tolerance test, tolerance is used in context of searching for a solution for stable swap over a cfmm function
 	NonZeroErrTolerance := ErrTolerance{AdditiveTolerance: sdk.NewInt(10), MultiplicativeTolerance: sdk.Dec{}}
```
p0mvn, line 158: Are we testing multiplicative tolerance anywhere? I think that should be tested as well

p0mvn: We should test:
- only multiplicative is non zero
- both multiplicative and additive are non-zero

osmoutils/binary_search_test.go
```diff
@@ -74,12 +89,19 @@ func TestBinarySearchNonlinear(t *testing.T) {
 }
 
 func TestBinarySearchNonlinearNonzero(t *testing.T) {
```
p0mvn, line 91: Why is this called "NonLinear" when we have both linear and non-linear functions?

osmoutils/binary_search_test.go
```diff
@@ -41,11 +48,17 @@ func TestBinarySearch(t *testing.T) {
 }
 
 func TestBinarySearchNonlinear(t *testing.T) {
```
p0mvn, line 50: Same here about the non-linear name

osmoutils/binary_search_test.go
```diff
@@ -24,6 +25,12 @@ func TestBinarySearch(t *testing.T) {
 
 		expectedSolvedInput sdk.Int
 		expectErr           bool
+		// This binary searches inputs to a monotonic increasing function F
```
p0mvn, line 28: Thanks for this comment

osmoutils/binary_search_test.go
p0mvn: What is the difference from `TestBinarySearch`? Why can't we have a single test `TestBinarySearch` that combines all of `TestBinarySearch`, `TestBinarySearchNonlinear`, `TestBinarySearchNonlinearNonzero`.

For each test case, we can have a name that describes whether it is a linear function and non-zero tolerance

p0mvn (commented):
Nice work, left some comments. Please take a look

osmoutils/binary_search_test.go
p0mvn: This is to avoid code duplication, each test fixture is essentially the same right now:
```
	for _, tc := range tests {
		actualSolvedInput, err := BinarySearch(tc.f, tc.lowerbound, tc.upperbound, tc.targetOutput, tc.errTolerance, tc.maxIterations)
		if tc.expectErr {
			require.Error(t, err)
		} else {
			require.NoError(t, err)
			require.True(sdk.IntEq(t, tc.expectedSolvedInput, actualSolvedInput))
		}
	}
```

Naser531900 (approved):

osmoutils/binary_search_test.go
xBalbinus: Ah this is called nonlinear because it tests nonzero tolerance with a nonlinear function, I can just change it to TestBinarySearchNonzero if that's better?

osmoutils/binary_search_test.go
xBalbinus: This is a great idea, I'll do that.

alexanderbez (approved):
epic

Issue #1512, opened 2022-05-16
Make tests for the generic CalcExitPool logic


ValarDragon:
We need to add unit tests for the generic `CalcExitPool` Logic in `x/gamm/pool-models/internal/cfmm_common/lp.go`.

TBD to see if something can be easily reused from existing balancer tests.

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/1419#discussion_r867283261_

PR #1540, opened 2022-05-19
add test for CalcExitPool

catShaark:
Closes: #1512 

- Unit test for `CalcExitPool`.

## Brief Changelog

- Add unit test for the generic `CalcExitPool` Logic in `x/gamm/pool-models/internal/cfmm_common/lp.go`.
- Change [this error](https://github.com/osmosis-labs/osmosis/blob/bcd59423fe0c939a64ac593692326f4cd91a6081/x/gamm/pool-models/internal/cfmm_common/lp.go#L19) returned by `CalcExitPool` when `exitingShares` is grater or equal to pool's `total shares` so that there's no `Int64() out of bound` panic.


x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/internal/cfmm_common"
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/stableswap"
+	gammtypes "github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func TestCalcExitPool(t *testing.T) {
+	var (
+		two_asset_pool = &stableswap.Pool{
```
p0mvn, line 18: Suggest using `NewStableswapPool(...)`

x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,117 @@
...
+			exitFee:       sdk.ZeroDec(),
+			expExitCoins:  sdk.Coins{},
+			expError:      true,
+		},
+		{
+			name:          "two-asset pool, valid exiting shares",
+			pool:          two_asset_pool,
+			exitingShares: gammtypes.InitPoolSharesSupply.QuoRaw(2),
+			exitFee:       sdk.ZeroDec(),
+			expExitCoins:  sdk.Coins{sdk.NewInt64Coin("foo", 500000000), sdk.NewInt64Coin("bar", 500000000)},
```
p0mvn, line 78: How were these calculated / estimated?

Is it possible to calculate these based on the initial values when we initialize the pools? That would make it clearer how we got these values and why they are expected. Otherwise, one only can guess where they are coming from

p0mvn (commented):
Awesome work on this! The tests look great. Had a couple of minor comments, please let me know what you think

x/gamm/pool-models/internal/cfmm_common/lp_test.go
catShaark: The pools used in this test is not fully initialized because they only need to provide `GetTotalShares()` and `GetTotalPoolLiquidity()`. Fields like `Address` or `Id` are not relevant in this test so I don't want to use `NewStableswapPool()` which requires those fields.

p0mvn: The reason why I suggest using `NewStableswapPool(...)` is because we are likely to keep adding initialization and validation logic in the constructor. Over time, it might start being invalid initializing a pool with or without certain parameters.

Using the constructor from the start would help to ensure that this initialization remains valid going forward. 

We can create:
```
const (
     testAddress = "testAddress"
     testId = "testId"
)
```
and provide that to `NewStableswapPool(...)`

x/gamm/pool-models/internal/cfmm_common/lp_test.go
catShaark: > How were these calculated / estimated?

This is calculated simply by hand.
 
> Is it possible to calculate these based on the initial values when we initialize the pools?

U mean defining a function to calculate expected values. I don't know how to elegantly do so without that function looking similar to the `CalcExitPool` function that we're testing.

x/gamm/pool-models/internal/cfmm_common/lp_test.go
catShaark: ohh that makes sense, I'll apply the change accordingly

x/gamm/pool-models/internal/cfmm_common/lp_test.go
catShaark: I think I got your idea now

x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,138 @@
...
+				{Token: sdk.NewInt64Coin("bar", 3000000000), Weight: sdk.NewIntFromUint64(5)},
+				{Token: sdk.NewInt64Coin("baz", 4000000000), Weight: sdk.NewIntFromUint64(5)},
+			},
+			"",
+			time.Now(),
+		)
+	)
+
+	// make sure there're no error creating those pools
+	require.NoError(t, err1)
```
p0mvn, line 75: nit: 

Checking errors right after pool initialization might be preferable because we wouldn't have to wait until all pools are initialized to validate them all. Imagine, err1 had an error but we proceeded to initialize all pools redundantly to fail anyways.

In short, I'm pointing at something like the following:
```
twoAssetPool, err := stableswap.NewStableswapPool(
			1,
			stableswap.PoolParams{ExitFee: sdk.ZeroDec()},
			sdk.NewCoins(sdk.NewInt64Coin("foo", 1000000000),
				sdk.NewInt64Coin("bar", 1000000000)),
			"",
			time.Now(),
		)
require.NoError(t, err)

threeAssetPool, err = balancer.NewBalancerPool(
			1,
			balancer.PoolParams{SwapFee: sdk.ZeroDec(), ExitFee: sdk.ZeroDec()},
			[]balancer.PoolAsset{
				{Token: sdk.NewInt64Coin("foo", 2000000000), Weight: sdk.NewIntFromUint64(5)},
				{Token: sdk.NewInt64Coin("bar", 3000000000), Weight: sdk.NewIntFromUint64(5)},
				{Token: sdk.NewInt64Coin("baz", 4000000000), Weight: sdk.NewIntFromUint64(5)},
			},
			"",
			time.Now(),
		)
require.NoError(t, err)
```

x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,138 @@
...
+		{
+			name:          "three-asset pool with exit fee, valid exiting shares",
+			pool:          &threeAssetPoolWithExitFee,
+			exitingShares: sdk.NewIntFromUint64(7000000000000),
+			expError:      false,
+		},
+	}
+
+	for _, test := range tests {
+		// using empty context since the context won't be used anyway
```
p0mvn, line 125:
```suggestion
		// using empty context since, currently, the context is not used anyway. This might be changed in the future
```

x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,138 @@
...
+		exitFee := test.pool.GetExitFee(emptyContext)
+		exitCoins, err := cfmm_common.CalcExitPool(emptyContext, test.pool, test.exitingShares, exitFee)
+		if test.expError {
+			require.Error(t, err, "test: %v", test.name)
+		} else {
+			require.NoError(t, err, "test: %v", test.name)
+
+			// exitCoins = ( (1 - exitFee) * exitingShares / poolTotalShares ) * poolTotalLiquidity
+			expExitCoins := mulCoins(test.pool.GetTotalPoolLiquidity(emptyContext), (sdk.OneDec().Sub(exitFee)).MulInt(test.exitingShares).QuoInt(test.pool.GetTotalShares()))
```
p0mvn, line 134: Nice work!

x/gamm/pool-models/internal/cfmm_common/lp_test.go
```diff
@@ -0,0 +1,138 @@
...
+			"",
+			time.Now(),
+		)
+		threeAssetPool, err2 = balancer.NewBalancerPool(
+			1,
+			balancer.PoolParams{SwapFee: sdk.ZeroDec(), ExitFee: sdk.ZeroDec()},
+			[]balancer.PoolAsset{
+				{Token: sdk.NewInt64Coin("foo", 2000000000), Weight: sdk.NewIntFromUint64(5)},
+				{Token: sdk.NewInt64Coin("bar", 3000000000), Weight: sdk.NewIntFromUint64(5)},
+				{Token: sdk.NewInt64Coin("baz", 4000000000), Weight: sdk.NewIntFromUint64(5)},
```
p0mvn, line 48: Seems that pool assets have the same values between the version with fee and the one without. Maybe we extract a `threePoolAssets` variable to avoid duplication?

p0mvn (commented):
Awesome work on making the calculations for expected exit coins! Left a few minor comments

x/gamm/pool-models/internal/cfmm_common/lp_test.go
p0mvn: Thank you, looks great

x/gamm/pool-models/internal/cfmm_common/lp_test.go
p0mvn: same with `twoPoolAssets`, assuming those are also duplicates

x/gamm/pool-models/internal/cfmm_common/lp_test.go
catShaark: Tbf, I don't know why I did this, thanks for pointing out

p0mvn (approved):
LGTM

PR #1535, opened 2022-05-18
chore: upgrade wasmd to v0.27.0.rc3-osmo and ibc-go to v3 (#1527)

p0mvn:
Closes: #XXX

## What is the purpose of the change

Context: #1527 

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? yes
  - How is the feature or change documented? not applicable

PR #1541, opened 2022-05-19
Add arm64 support to Docker images

nikever:
Closes: #XXX

This PR extends the work done in #1535 to introduce support for `arm64` architecture making osmosis docker image multi-architecture (`amd64` and `arm64`).

I have also updated the CI to build and push the image for multiple architectures.

Please note that it takes ~20 minutes for the build complete but it would run only on every new tag so I think it's acceptable.

## Brief Changelog

- Modify Dockerfile to add `arm64` support
- Update docker CI to build and push for `arm64`

## Testing and Verifying

You can build for arm64 with:

```bash
docker buildx build --platform linux/arm64 --tag osmosis:arm64 .
```

I tested the CI in my fork: https://github.com/nikever/osmosis/runs/6505857265?check_suite_focus=true

  - Does this pull request introduce a new feature or user-facing behavior changes? yes (`arm64` support!)


nikever:
Should I target `v9.x` instead?

nikever:
The same logic here could be used to create a "releaser" docker image that creates the binary for both `amd64` and `arm64`. Let me know if that could interest you!

alexanderbez (approved):
<3

ValarDragon (approved):

ValarDragon:
Targetting main is fine, we can backport the change!

p0mvn (commented):
@nikever could you also update the following image please:
https://github.com/osmosis-labs/osmosis/blob/82e58f8f5f7e200e6659de6490526391811dd9ea/contrib/images/rbuilder/Dockerfile

It is needed for building the release binaries

nikever:
> @nikever could you also update the following image please: https://github.com/osmosis-labs/osmosis/blob/82e58f8f5f7e200e6659de6490526391811dd9ea/contrib/images/rbuilder/Dockerfile
> 
> It is needed for building the release binaries

Since now we support different architecture and each one would have to be linked against its own CosmWasm library, what do you think to use `goreleaser` for this part instead of the Makefile? 

It's pretty cool and automation friendly: https://goreleaser.com/quick-start/ 

If you want to test it, I think it's best to track this part in another issue. lmk!

p0mvn:
> > @nikever could you also update the following image please: https://github.com/osmosis-labs/osmosis/blob/82e58f8f5f7e200e6659de6490526391811dd9ea/contrib/images/rbuilder/Dockerfile
> > It is needed for building the release binaries
> 
> Since now we support different architectures, and each one would have to be linked against its corresponding CosmWasm library, what do you think about using `goreleaser` for this part instead of the Makefile?
> 
> It's pretty cool and automation-friendly: https://goreleaser.com/quick-start/
> 
> If you want to give it a shot, I think it's best to track this releasing part in another issue. lmk!

Yeah I'm all for looking into that later. We still need to make the current release image consistent with the main one so that we can make releases for all supported platforms. We can either do it here or in a separate issue

nikever:
I'm tracking the release part in another issue: https://github.com/osmosis-labs/osmosis/issues/1550 

I have already started doing some work in this branch: https://github.com/osmosis-labs/osmosis/compare/main...nikever:feat/add-go-releaser?expand=1

We can merge this one since the Docker part should be okay.

p0mvn (approved):

PR #1546, opened 2022-05-19
bug: tx flags for txfees update prop

czarcas7ic:
Can't submit txfees prop without fees flags


p0mvn:
@czarcas7ic  Was it not working before and just not used? Just curious how this was never caught

czarcas7ic:
My guess is all the current submissions were done via lcd @p0mvn

p0mvn (approved):

PR #1551, opened 2022-05-20
Make token factory handle empty genesis configurations better

ValarDragon:

A problem @czarcas7ic ran into was getting panics on token factory messages, when the genesis wasn't initialized.

## Brief Changelog

- Improve nil handling of token factory InitGenesis

## Testing and Verifying

- TODO

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no, still pre-release for token factory
  - How is the feature or change documented? not applicable, removes a thing to document tbh

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,3 +43,4 @@ EXPOSE 26657
 EXPOSE 1317
 
 ENTRYPOINT ["osmosisd"]
+CMD [ "start" ]
--- a/x/tokenfactory/genesis.go
+++ b/x/tokenfactory/genesis.go
@@ -12,6 +12,9 @@ import (
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
 	k.CreateModuleAccount(ctx)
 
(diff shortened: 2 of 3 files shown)


ValarDragon:
e2e test failure seems like some sort of docker issue?

czarcas7ic:
The local compilation is not hitting blocks, so essentially the update failed. I will go through it locally to see what the specific issue is for this pr

czarcas7ic (approved):

alexanderbez (commented):
Curious why it'd ever be nil to begin with. Does default genesis state not use `sdk.NewCoins`?

czarcas7ic:
I ran into it when modifying the v8 genesis on LocalOsmosis to work on v9. Will the average user ever run into this case? Probably not. But its just a nice check to have since if it is null (like it was in my case) and it still gets through init genesis, it causes untraceable errors.

alexanderbez:
We should add safe guards, I agree @czarcas7ic 👍

PR #1554, opened 2022-05-21
feat: local dev environment

czarcas7ic:

As it stands, it is not simple to quickly test a change in a local dev environment. This introduces a modified LocalOsmosis that is lightweight and makes it very easy to spin up a local net to test your changes. The docker image also creates a new genesis on build, preventing any issues that have arose in the past from hardcoded genesis files.

The basic process is as follows:

1. Make your local changes
2. Run `make localnet-build`
3. Run `make localnet-start`. Your local network has now started. Query it with your local daemon (i.e `osmosisd status`)
4. Run `make localnet-keys` to add all 10 keys to your keyring
5. Run `make localnet-remove` to remove all block data

## Testing and Verifying

This code was tested locally in an arm64 environment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,18 @@ test-e2e:
 test-e2e-skip-upgrade:
 	@VERSION=$(VERSION) OSMOSIS_E2E_SKIP_UPGRADE=True go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
 
+lo-keys:
+	. tests/localosmosis/keys.sh
+
+lo-build:
+	@docker build -t local:osmosis --build-arg BASE_IMG_TAG=debug -f tests/localosmosis/Dockerfile .
+
+lo-start:
+	@docker-compose -f tests/localosmosis/docker-compose.yml up
+
+lo-remove:
+	@docker-compose -f tests/localosmosis/docker-compose.yml down
+
 benchmark:
 	@go test -mod=readonly -bench=. $(PACKAGES_UNIT)
 
--- a/Makefile
+++ b/Makefile
@@ -242,18 +242,6 @@ test-e2e:
 test-e2e-skip-upgrade:
 	@VERSION=$(VERSION) OSMOSIS_E2E_SKIP_UPGRADE=True go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
 
-lo-keys:
-	. tests/localosmosis/keys.sh
-
-lo-build:
-	@docker build -t local:osmosis --build-arg BASE_IMG_TAG=debug -f tests/localosmosis/Dockerfile .
-
-lo-start:
-	@docker-compose -f tests/localosmosis/docker-compose.yml up
-
-lo-remove:
-	@docker-compose -f tests/localosmosis/docker-compose.yml down
-
 benchmark:
 	@go test -mod=readonly -bench=. $(PACKAGES_UNIT)
 
@@ -279,6 +267,17 @@ format:
 ###                                Localnet                                 ###
 ###############################################################################
 
+localnet-keys:
+	. tests/localosmosis/keys.sh
+
+localnet-build:
+	@docker build -t local:osmosis -f tests/localosmosis/Dockerfile .
+
+localnet-start:
+	@docker-compose -f tests/localosmosis/docker-compose.yml up
+
+localnet-remove:
+	@docker-compose -f tests/localosmosis/docker-compose.yml down
 
 .PHONY: all build-linux install format lint \
 	go-mod-cache draw-deps clean build build-contract-tests-hooks \
--- a/tests/localosmosis/Dockerfile
+++ b/tests/localosmosis/Dockerfile
@@ -32,14 +32,16 @@ RUN BUILD_TAGS=muslc LINK_STATICALLY=true make build
 # Runner
 # --------------------------------------------------------
 
-FROM ubuntu
+FROM alpine
 
 COPY --from=build /osmosis/build/osmosisd /bin/osmosisd
 COPY /tests/localosmosis/setup.sh /setup.sh
 
 ENV HOME /osmosis
 WORKDIR $HOME
-RUN apt-get update && apt-get install -y jq
+RUN apk update \
+ && apk add jq \
+ && rm -rf /var/cache/apk/*
 RUN chmod +x /setup.sh
 RUN /setup.sh
 EXPOSE 26656
--- a/tests/localosmosis/Dockerfile
+++ b/tests/localosmosis/Dockerfile
@@ -1,7 +1,5 @@
 # syntax=docker/dockerfile:1
 
-ARG BASE_IMG_TAG=nonroot
-
 # --------------------------------------------------------
 # Build
 # --------------------------------------------------------
@@ -39,9 +37,10 @@ COPY /tests/localosmosis/setup.sh /setup.sh
 
 ENV HOME /osmosis
 WORKDIR $HOME
-RUN apk update \
- && apk add jq \
- && rm -rf /var/cache/apk/*
+RUN apk update
+RUN apk add jq
+RUN apk add moreutils
+RUN rm -rf /var/cache/apk/*
 RUN chmod +x /setup.sh
 RUN /setup.sh
 EXPOSE 26656
--- a/tests/localosmosis/Dockerfile
+++ b/tests/localosmosis/Dockerfile
@@ -1,0 +1,91 @@
+# syntax=docker/dockerfile:1
+
+ARG BASE_IMG_TAG=nonroot
+
+# --------------------------------------------------------
+# Build
+# --------------------------------------------------------
+
+FROM golang:1.18.2-alpine3.15 as build
+
+RUN set -eux; apk add --no-cache ca-certificates build-base;
+RUN apk add git
+# Needed by github.com/zondax/hid
+RUN apk add linux-headers
+
+WORKDIR /osmosis
+COPY . /osmosis
+
+# CosmWasm: see https://github.com/CosmWasm/wasmvm/releases
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
+RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep 7d2239e9f25e96d0d4daba982ce92367aacf0cbd95d2facb8442268f2b1cc1fc
+RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep f6282df732a13dec836cda1f399dd874b1e3163504dbd9607c6af915b2740479
+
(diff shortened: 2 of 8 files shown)


tests/localosmosis/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+# CosmWasm: copy the right library according to architecture. The final location will be found by the linker flag `-lwasmvm_muslc`
+RUN cp /lib/libwasmvm_muslc.$(uname -m).a /lib/libwasmvm_muslc.a
+
+RUN BUILD_TAGS=muslc LINK_STATICALLY=true make build
+
+# --------------------------------------------------------
+# Runner
+# --------------------------------------------------------
+
+FROM ubuntu
```
p0mvn, line 34: Does it have to be ubuntu? Maybe something more lightweight?

tests/localosmosis/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+COPY --from=build /osmosis/build/osmosisd /bin/osmosisd
+
+ENV HOME /osmosis
+WORKDIR $HOME
+RUN apt-get update && apt-get install -y jq
+# change staking denom to uosmo
+RUN osmosisd init --chain-id=localosmosis val
+RUN echo "satisfy adjust timber high purchase tuition stool faith fine install that you unaware feed domain license impose boss human eager hat rent enjoy dawn" | osmosisd keys add val --recover --keyring-backend=test
+RUN cat $HOME/.osmosisd/config/genesis.json | jq '.app_state["staking"]["params"]["bond_denom"]="uosmo"' > $HOME/.osmosisd/config/tmp_genesis.json && mv $HOME/.osmosisd/config/tmp_genesis.json $HOME/.osmosisd/config/genesis.json
+RUN osmosisd add-genesis-account osmo1phaxpevm5wecex2jyaqty2a4v02qj7qmlmzk5a 100000000000uosmo,100000000000uion
```
p0mvn, line 45: Each RUN in a Dockerfile adds an additional layer, making the images large in size:
https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-which-is-better

We should extract all these in a bash script and execute that instead

tests/localosmosis/keys.sh
```diff
@@ -0,0 +1,12 @@
...
+echo "spatial forest elevator battle also spoon fun skirt flight initial nasty transfer glory palm drama gossip remove fan joke shove label dune debate quick" | osmosisd keys add lo-test6 --recover --keyring-backend test
+echo "noble width taxi input there patrol clown public spell aunt wish punch moment will misery eight excess arena pen turtle minimum grain vague inmate" | osmosisd keys add lo-test7 --recover --keyring-backend test
+echo "cream sport mango believe inhale text fish rely elegant below earth april wall rug ritual blossom cherry detail length blind digital proof identify ride" | osmosisd keys add lo-test8 --recover --keyring-backend test
+echo "index light average senior silent limit usual local involve delay update rack cause inmate wall render magnet common feature laundry exact casual resource hundred" | osmosisd keys add lo-test9 --recover --keyring-backend test
+echo "prefer forget visit mistake mixture feel eyebrow autumn shop pair address airport diesel street pass vague innocent poem method awful require hurry unhappy shoulder" | osmosisd keys add lo-test10 --recover --keyring-backend test
```
p0mvn, line 12: nit: newline at the end

tests/localosmosis/keys.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+echo "satisfy adjust timber high purchase tuition stool faith fine install that you unaware feed domain license impose boss human eager hat rent enjoy dawn" | osmosisd keys add lo-val --recover --keyring-backend test
```
p0mvn, line 2: Nice, I was just doing all that manually like an hour ago

tests/localosmosis/Dockerfile
czarcas7ic: I think it's really important to generate the genesis at compile time, and to do this it requires jq. I attempted with distroless but was unable to find and replace that way. If you have any ideas I would be more than happy to give them a shot

tests/localosmosis/Dockerfile
czarcas7ic: Ah, see I didn't know that. I had these in a script file to begin with but then felt like I might as well combine them in the required dockerfile as well. I will separate and instead call that bash file from the dockerfile, thanks!

tests/localosmosis/Dockerfile
p0mvn: Ahh interesting, I think  jq should be available on alpine images that are more lightweight than ubuntu

czarcas7ic: Alpine worked, thank you!!

p0mvn (approved):

Makefile
```diff
@@ -242,6 +242,18 @@ test-e2e:
 test-e2e-skip-upgrade:
 	@VERSION=$(VERSION) OSMOSIS_E2E_SKIP_UPGRADE=True go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
 
+lo-keys:
```
nikever, line 245: Two small suggestions:

- What if we move group all these Makefile targets in a `localnet` section replacing the one below? 
- To the cost of being move verbose, I would be more explicit with the naming `lo` --> `localnet` . `make localnet-start` sounds a bit better

tests/localosmosis/Dockerfile
```diff
@@ -0,0 +1,52 @@
+# syntax=docker/dockerfile:1
+
+ARG BASE_IMG_TAG=nonroot
```
nikever, line 3: This line can be remove. 

`BASE_IMG_TAG` is not used in this Dockefile

tests/localosmosis/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3"
+
+services:
+  osmosisd:
```
nikever, line 4: You could add:

```
osmosisd:
  build:
      context: .
```

In this way when you run `docker-compose start` it would automatically build the image for you! :)

Makefile
```diff
@@ -242,6 +242,18 @@ test-e2e:
...
+	. tests/localosmosis/keys.sh
+
+lo-build:
+	@docker build -t local:osmosis --build-arg BASE_IMG_TAG=debug -f tests/localosmosis/Dockerfile .
```
nikever, line 249: `--build-arg BASE_IMG_TAG=debug` can be removed as it's not used.

If you like the idea of adding the `build/context` option in the `docker-compose` as suggested in a previous command, the command could be made consistent to the others replacing it with:

 `@docker-compose -f tests/localosmosis/docker-compose.yml build`

tests/localosmosis/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+FROM alpine
+
+COPY --from=build /osmosis/build/osmosisd /bin/osmosisd
+COPY /tests/localosmosis/setup.sh /setup.sh
+
+ENV HOME /osmosis
+WORKDIR $HOME
+RUN apk update \
+ && apk add jq \
+ && rm -rf /var/cache/apk/*
```
nikever, line 44: Split this in multiple lines:

```
RUN apk update
RUN apk add jq
...
```

It's generally best to split commands in `Dockerfile` as it allows to cache and reuse layers more easily

tests/localosmosis/setup.sh
```diff
@@ -0,0 +1,46 @@
...
+osmosisd add-genesis-account osmo12rr534cer5c0vj53eq4y32lcwguyy7nndt0u2t 100000000000uosmo,100000000000uion
+osmosisd add-genesis-account osmo1nt33cjd5auzh36syym6azgc8tve0jlvklnq7jq 100000000000uosmo,100000000000uion
+osmosisd add-genesis-account osmo10qfrpash5g2vk3hppvu45x0g860czur8ff5yx0 100000000000uosmo,100000000000uion
+osmosisd add-genesis-account osmo1f4tvsdukfwh6s9swrc24gkuz23tp8pd3e9r5fa 100000000000uosmo,100000000000uion
+osmosisd add-genesis-account osmo1myv43sqgnj5sm4zl98ftl45af9cfzk7nhjxjqh 100000000000uosmo,100000000000uion
+osmosisd add-genesis-account osmo14gs9zqh8m49yy9kscjqu9h72exyf295afg6kgk 100000000000uosmo,100000000000uion
+osmosisd gentx val 500000000uosmo --keyring-backend=test --chain-id=localosmosis
+osmosisd collect-gentxs
+# update staking genesis
+cat $HOME/.osmosisd/config/genesis.json | jq '.app_state["staking"]["params"]["unbonding_time"]="240s"' > $HOME/.osmosisd/config/tmp_genesis.json && mv $HOME/.osmosisd/config/tmp_genesis.json $HOME/.osmosisd/config/genesis.json
```
nikever, line 21: If in the `Dockerfile` we install the `moreutils` package:

```Dockerfile
RUN apk update 
RUN apk add jq 
RUN apk add moreutils
```

We could use `sponge` command to make this command a little less verbose:

```bash
cat genesis.json | jq <query> | sponge genesis.json
```

example:

```bash
cat $HOME/.osmosisd/config/genesis.json | jq '.app_state["staking"]["params"]["unbonding_tim
e"]="999s"' | sponge $HOME/.osmosisd/config/genesis.json
```

To increase readibilty we could also do:

```bash
GENESIS=$HOME/.osmosisd/config/genesis.json

cat $GENESIS | jq '.app_state["staking"]["params"]["unbonding_tim
e"]="999s"' | sponge $GENESIS
...
```

tests/localosmosis/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN chmod +x /setup.sh
+RUN /setup.sh
```
nikever, line 46: One little improvement could be not to run this script during the `docker build` phase but running as an entrypoint script of the docker container. 

In this way we can avoid rebuilding the image if we want to modify a parameter of the genesis. 
Building takes 1-2 mins while running the script should be in the order of seconds

The way to do that would be to modifying the Dockerfile like this:

```bash

...
WORKDIR $HOME
RUN apk update \
 && apk add jq \
 && rm -rf /var/cache/apk/*
RUN chmod +x /setup.sh

# Remove the RUN /setup

EXPOSE 26656
EXPOSE 26657
EXPOSE 1317

ENTRYPOINT ["/setup.sh"]
CMD ["osmosisd", "start"]
```

And in the `setup.sh`:

```bash
#!/usr/bin/env sh
set -eu

<STANDARD CONTENT>

exec "$@" 
```

The `exec` part will execute the `CMD` part of the Dockerfile after is has executed the script

nikever (approved):
LGTM! Nice stuff! It's going to be very useful!

I added some improvements here and there. 
As always, take what you like.

I'm approving the PR as it's working as it is.

tests/localosmosis/setup.sh
czarcas7ic: I didn't know about this command, awesome!

tests/localosmosis/Dockerfile
czarcas7ic: The only problem with this is the user would be recreating the genesis process on every start right? This process should only be done once at compilation time I think, and if the user wanted to modify a genesis param, they are probably better off recompiling anyway right?

czarcas7ic:
@nikever @p0mvn thanks so much for the review, both had awesome suggestions that I ended up implementing

PR #1564, opened 2022-05-23
Adding ICA Host module

ValarDragon:
Closes: #XXX

Adds Interchain account host module

## Brief Changelog

- Add interchain account host module

## Testing and Verifying

References I'm using in writing this:
* https://github.com/cosmos/ibc-go/blob/v3.0.0/docs/migrations/v2-to-v3.md
* https://github.com/cosmos/gaia/blob/v7.0.2/app/app.go

As long as we can verify that IBC channel creation and sends still work post this getting merged, I think were fine.

  - How is the feature or change documented? https://github.com/cosmos/ibc-go/blob/v3.0.0/docs/migrations/v2-to-v3.md


app/upgrades/v9/upgrades.go
```diff
@@ -13,8 +17,24 @@ func CreateUpgradeHandler(
...
+		// set the ICS27 consensus version so InitGenesis is not run
+		fromVM[icatypes.ModuleName] = mm.Modules[icatypes.ModuleName].ConsensusVersion()
+
+		// create ICS27 Controller submodule params, controller module not enabled.
+		controllerParams := icacontrollertypes.Params{}
+
+		// create ICS27 Host submodule params
+		hostParams := icahosttypes.Params{
+			HostEnabled:   true,
+			AllowMessages: []string{"/cosmos.bank.v1beta1.MsgSend"},
```
alexanderbez, line 33: To be safe here, what we can/should do, is the use the types directly.

i.e.

```go
AllowMessages: []string{
  sdk.MsgTypeURL(&banktypes.MsgSend{}),
  // ...
}
```

alexanderbez (commented):
🙌

ValarDragon:
We should use this PR as a reference for more boilerplate we need to eliminate

app/upgrades/v9/upgrades.go
ValarDragon: such sad that we have to do this :/

ValarDragon: I'm going to add send, gov vote, stake, reward claim, and gamm messages.

I think we just go change the IBC interface to make it more sane to include everything by default >:( (Aditya mentioned this is on their roadmap, may be we can just help with a PR here, if they'd be down to release sooner)

ValarDragon: And Authz, they can take the L of executing two messages via authz'ing to themself lol

PR #1567, opened 2022-05-24
Use balancer invariant to calculate exact expected quantities of test vectors

AlpinYukseloglu:

Use balancer invariant to calculate exact expected quantities of test vectors.

## Testing and Verifying

All calculations rely directly on the Balancer whitepaper but need to be reviewed manually for correctness.


mattverse (changes requested):
Oh interesting, it seems that the actual output of the method is different from the actual calculation by 50? I think we should probably investigate and document why we have this difference / where this is coming from

ValarDragon:
The concerning part imo is that its rounding up, the actual error magnitude seems within error margins at least

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -324,3 +318,76 @@ func TestCalcSingleAssetInAndOut_InverseRelationship(t *testing.T) {
...
+			// P_issued = P_supply * ((1 + (A_t / B_t))^W_t - 1),
+			// where:
+			// 	P_supply = initial pool supply = 100 * 10^18 (set at pool creation, same for all new pools)
+			//	A_t = amount of deposited asset = 50,000
+			//	B_t = existing balance of deposited asset in the pool prior to deposit = 1,000,000,000,000
+			//	W_t = normalized weight of deposited asset in pool = 0.5 (equally weighted two-asset pool)
+			// Plugging all of this in, we get:
+			// 	Full solution: https://www.wolframalpha.com/input?i=100000000000000000000*%28%281+%2B+%2850000%2F1000000000000%29%29%5E0.5+-+1%29
+			// 	Simplified:  P_issued = 2,499,999,968,750
+			expectShares: sdk.NewInt(2499999968800),
```
AlpinYukseloglu, line 358: @ValarDragon

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -324,3 +318,76 @@ func TestCalcSingleAssetInAndOut_InverseRelationship(t *testing.T) {
 		}
 	}
 }
+
+func TestCalcJoinPoolShares(t *testing.T) {
```
mattverse, line 322: I think these changes are good to be deleted for now since we got most of this testing covered in 
https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/pool-models/balancer/amm_joinpool_test.go

mattverse (changes requested):
imo, should be good to remove the changes in this PR regarding `amm_test.go`, other than that lgtm

PR #1571, opened 2022-05-24
Add msg_filter_ante_handler to block IBCTimeoutOnClose

ValarDragon:
Closes: #XXX

Add a msg_filter_ante for v9, and block MsgTimeoutOnClose, due to edge cases that can happen if Terra re-enables the previously closed IBC channels. (This is because Closed channels were never meant to be re-opened)

## Brief Changelog

Adds a ante handler filter for IBC Timeout close messages. This is a port of Bez' old work.

## Testing and Verifying

Covered by provided tests/

  - How is the feature or change documented? TODO

### DIFF
--- a/ante/msg_filter_ante.go
+++ b/ante/msg_filter_ante.go
@@ -1,0 +1,35 @@
+package ante
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+)
+
+// MsgFilterDecorator defines an AnteHandler decorator for the v8 upgrade that
+// provide height-gated message filtering acceptance.
+type MsgFilterDecorator struct{}
+
+// AnteHandle performs an AnteHandler check that returns an error if the tx contains a message
+// that is blocked.
+// Right now, we block MsgTimeoutOnClose due to incorrect behavior that could occur if a packet is re-enabled.
+func (mfd MsgFilterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	if hasInvalidMsgs(tx.GetMsgs()) {
+		currHeight := ctx.BlockHeight()
+		return ctx, fmt.Errorf("tx contains unsupported message types at height %d", currHeight)
+	}
+
+	return next(ctx, tx, simulate)
+}
+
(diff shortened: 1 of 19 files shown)


ante/msg_filter_ante.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+)
+
+// MsgFilterDecorator defines an AnteHandler decorator for the v8 upgrade that
+// provide height-gated message filtering acceptance.
+type MsgFilterDecorator struct{}
```
p0mvn, line 12: Do we need to initialize it similar to how we did this in v8?

https://github.com/osmosis-labs/osmosis/blob/b82dae462c072aa6eebfa5c56ea3d22b9bb63473/app/ante.go#L39

ante/msg_filter_ante.go
```diff
@@ -0,0 +1,35 @@
+package ante
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+)
+
+// MsgFilterDecorator defines an AnteHandler decorator for the v8 upgrade that
```
p0mvn, line 10:
```suggestion
// MsgFilterDecorator defines an AnteHandler decorator for the v9 upgrade that
```

ante/msg_filter_ante.go
```diff
@@ -0,0 +1,35 @@
...
+// provide height-gated message filtering acceptance.
+type MsgFilterDecorator struct{}
+
+// AnteHandle performs an AnteHandler check that returns an error if the tx contains a message
+// that is blocked.
+// Right now, we block MsgTimeoutOnClose due to incorrect behavior that could occur if a packet is re-enabled.
```
p0mvn, line 16: Is it possible to run into other issues from disabling this message?

p0mvn (approved):
utACK. It makes sense based on earlier Bez's PR in v8 and the description on Slack. 

However, I'm not very confident in my understanding of how this change behaves with other parts of the system

ante/msg_filter_ante.go
ValarDragon: No, we'll be fine. Just forces any packet sent on that channel to have to wait until its timeout before it can be closed.

ante/msg_filter_ante.go
ValarDragon: yes, ty

PR #1582, opened 2022-05-25
osmolbp: cli updates & default genesis

harish551:
-  updates to CLI tx commands (create lbp with json input & finalize lbp cmd)
-  added default genesis with params


x/osmolbp/genesis.go
```diff
@@ -0,0 +1,18 @@
...
+
+// InitGenesis initializes the capability module's state from a provided genesis
+// state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState api.GenesisState) {
+	// k.SetParams(ctx, genState.Params)
+}
+
+// ExportGenesis returns the osmolbp module's exported genesis.
+func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *api.GenesisState {
+	return &api.GenesisState{}
```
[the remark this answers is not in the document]
robert-zaremba: Also, shall we move this function to Keeper (as a method)?

x/osmolbp/api/keys.go
```diff
@@ -0,0 +1,12 @@
...
+const (
+	// ModuleName is the module name constant used in many places
+	ModuleName = "osmolbp"
+
+	// RouterKey is the message route for osmolbp
+	RouterKey = ModuleName
+
+	// QuerierRoute is the querier route for osmolbp
+	QuerierRoute = ModuleName
+)
```
robert-zaremba, line 12: Why copying this? we have `x/osmolbp/keys.go`. I think we should not have this consts in `api`.

app/app.go
```diff
@@ -128,6 +128,7 @@ import (
 
 	osmolbpkeeper "github.com/osmosis-labs/osmosis/x/osmolbp/keeper"
 	osmolbpmodule "github.com/osmosis-labs/osmosis/x/osmolbp/module"
+	osmolbptypes "github.com/osmosis-labs/osmosis/x/osmolbp/api"
```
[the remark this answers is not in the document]
harish551: sure will rename

x/osmolbp/genesis.go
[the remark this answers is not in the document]
harish551: will mv genesis file to keepers

app/app.go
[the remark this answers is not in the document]
harish551: we don't need it actually if we remove api/keys.go file

x/osmolbp/client/cli/parse.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/spf13/pflag"
+)
+
+type XCreateLBPInputs createLBPInputs
```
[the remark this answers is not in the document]
harish551: no reason just copied gamm/pool create cmd parse code and updated
will remove X

x/osmolbp/api/keys.go
robert-zaremba: yes , that was the idea

x/osmolbp/client/cli/parse.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/spf13/pflag"
+)
+
+type CreateLBPInputs createLBPInputs
```
robert-zaremba, line 12: why type rename?

x/osmolbp/client/cli/parse.go
```diff
@@ -0,0 +1,53 @@
...
+
+type XCreateLBPInputsExceptions struct {
+	CreateLBPInputs
+	Other *string // Other won't raise an error
+}
```
robert-zaremba, line 17: Why do we need this wrapper type? Why do we need `Other *sting`?

If we really need it then it should be a not exported type and we should remove X from the prefix

x/osmolbp/client/cli/parse.go
harish551: without renaming type if i directly use `createLBPInputs`
getting 
`runtime: goroutine stack exceeds 1000000000-byte limit`
I'm not sure why this is happening but with renaming & type conversion it's working

x/osmolbp/client/cli/parse.go
harish551: yeah we can directly use `CreateLBPInputs` type, will remove this

robert-zaremba (approved):

Issue #1578, opened 2022-05-25
Speedup CI


ValarDragon:
## Background

We should invest time in speeding up CI. Lots of time is lost in just waiting for CI to finish executing. If theres any reason we cannot do all of the above in github actions, imo we should investigate moving to a paid CI service, perhaps Circle, where they can all be done. Getting the feedback loop right for rapid merging multiple related PR's is important.

First where we stand on latencies. As of time of issue creation, we are blocked on:
- CodeQL (Takes 10 minutes to complete)
- Two simultaneous end to end tests (takes around 8 minutes to complete, half of which is in docker building)
- (Queued) the two go tests, each of which takes around 7 minutes to complete, half of which is spent in downloading go packages.

Furthermore, many github actions are stuck in 'queued', and take awhile before they begin execution, further increasing this dead time cost.

## Components

- [ ] Remove duplicate push / pull request jobs. #1573 , hopefully improves queue delay
- [ ] Remove less useful CLI tests (~2 minute saving on inspection of CLI logs)
- [ ] Cache Docker build
- [ ] Cache go tests #1229 
- [ ] Cache go package downloads in CI
- [ ] Figure out how to improve CodeQL times, or just remove it.

## Acceptance Criteria

- [ ] CI completes in under 5 minutes

ValarDragon:
This repo has examples for how we can cache go package downloads in CI: https://github.com/mvdan/github-actions-golang

It also has the caching needed to get incremental builds, which would likely also help some of this CI time.

ValarDragon:
oh thats a great idea. Perhaps we can make that variable also get run if theres any change in the upgrades sub folder as well?

ValarDragon:
oh thats a great idea. Perhaps we can make that variable also get run if theres any change in the upgrades sub folder as well?

ValarDragon:
For docker build times, I think we can replace the `make docker-build-debug` step with a direct docker build invocation, and use `cache` fields, e.g. https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api

nikever:
Added the cache option in https://github.com/osmosis-labs/osmosis/pull/1602 
It gives a nice speed up but I would like to be even faster.

There are some redundant steps between `go_test` step and `e2e_test` that could be improved. 
I'm working on re-structuring a little bit the setup to avoid redoing work.

PR #1583, opened 2022-05-25
Add CI cache

ValarDragon:

Try adding a cache for go build data, to help speedup CI. Copied from mvdan here: https://github.com/mvdan/github-actions-golang/blob/master/.github/workflows/test.yml

cref #1578 

## Brief Changelog

- Adds go cache to CI

## Testing and Verifying

We will need to keep an eye on if there are edge cases that make this not work / give incorrect results, or things don't speedup. A bit hard to know this without it being in place though.

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no (I don't think CI is in scope)
  - How is the feature or change documented? We need to figure out a strategy for if/how we want to document CI.


ValarDragon:
Compare the two runs
1) (First run, cache miss) https://github.com/osmosis-labs/osmosis/runs/6591778494?check_suite_focus=true
2) (Second run, cache hit) https://github.com/osmosis-labs/osmosis/runs/6591903758?check_suite_focus=true

This cut out over half the time! (No tests were cached here, presumably due to me doing a change in app.go which under current infra requires re-running every test!)

This suggests that the go build time was around 3 minutes in every CI run, perhaps getting something similar for docker tests would help as well! We'd likely want a different cache key for the docker tests though, since we don't want it to override the go test cache.

ValarDragon:
Last commit is trying to add a cache to test-e2e, may need a couple of go edits before we can see if it helps

ValarDragon:
Awesome, tests are still getting rebuilt / re-run! (Evident by last commit)

ValarDragon:
Hrmm these docker test fails seem unrelated (see lint super-linter failure. Appears that docker itself may be down :/, as it failed on docker pull)

ValarDragon:
For Docker, Compare the two runs

(First run, cache miss) https://github.com/osmosis-labs/osmosis/runs/6592152484?check_suite_focus=true
(Second run, cache hit) https://github.com/osmosis-labs/osmosis/runs/6592518772?check_suite_focus=true

Time went from `9m31s -> `8m23s`. However, concurrently it appeared github actions is having network latency problems. (The lint job was intermittently failing to connect to the internet for docker pull, and checkout the repo)

The e2e upgrade testing time shaved 1minute 40 seconds, however docker image build increased by 30 seconds. So we still need to separately figure out how to speedup docker image building (perhaps incremental build) and further speeding up the e2e tests.

ValarDragon:
Linter speed difference:

(Run with no cache) https://github.com/osmosis-labs/osmosis/runs/6591900239?check_suite_focus=true
(First run, cache miss) https://github.com/osmosis-labs/osmosis/runs/6593255081?check_suite_focus=true
(Second run, cache hit) https://github.com/osmosis-labs/osmosis/runs/6593340103?check_suite_focus=true

Time difference is `2m14s -> 36s`. (The no cache run took 2m6s, so huge time saving regardless.)

app/modules.go
```diff
@@ -235,6 +237,11 @@ func simulationModules(
 ) []module.AppModuleSimulation {
 	appCodec := encodingConfig.Marshaler
 
+	// appModules := appModules(app, encodingConfig, skipGenesisInvariants)
```
alexanderbez, line 240: can we just remove this if we're commenting it out?

ValarDragon: ah yeah, that was purely there for getting CI cache to run, I moved some of my WIP code from another branch here lol

ValarDragon: deleted!

p0mvn (approved):

PR #1586, opened 2022-05-25
tests: superfluid delegator voting

czarcas7ic:

  - Changes genesis creation to be closer to production (uses `osmo` instead of `stake` as staking token as well as all other occurrences of using stake as default token)
  - Adds a test to ensure a delegator can override their share of the validator's voting power.

## Testing and Verifying

This change is already covered by existing e2e test framework. Was also run locally on arm64 arch.

  - Does this pull request introduce a new feature or user-facing behavior changes? not


tests/e2e/e2e_util_test.go
```diff
@@ -180,14 +187,101 @@ func (s *IntegrationTestSuite) submitProposal(c *chain.Chain, upgradeHeight int)
...
+			})
+			s.Require().NoError(err)
+
+			err = s.dkrPool.Client.StartExec(exec.ID, docker.StartExecOptions{
+				Context:      ctx,
+				Detach:       false,
+				OutputStream: &outBuf,
+				ErrorStream:  &errBuf,
+			})
+			return strings.Contains(outBuf.String(), "code: 0")
```
ValarDragon, line 267: Lets refactor all of this into a helper function, this is too much boilerplate to be repeating so much

ValarDragon: On skim, it appears to me that line 242 through 271 an be refactored into a single helper function, and re-used in most of these tests.

ValarDragon: Can later make a separate helper method that lets for more control of various arguments, e.g. timeouts and container id

ValarDragon: If you prefer to keep the diff for this PR small, perhaps add the helper, and then just use it for new tests, and make an issue  and/or second PR for using it for everything else.

(I'm also happy with using it for everything in this PR as well though)

tests/e2e/e2e_util_test.go
```diff
@@ -401,3 +592,220 @@ func (s *IntegrationTestSuite) createPool(c *chain.Chain, poolFile string) {
...
+					ErrorStream:  &errBuf,
+				})
+				if err != nil {
+					return false
+				}
+				return true
+			},
+			time.Minute,
+			time.Second,
+		)
```
ValarDragon, line 759: Ditto for making a helper for this generic query boilerplate.

tests/e2e/e2e_util_test.go
ValarDragon: You can even make the `Cmd` in the helper just be a string thats formatted, and the helper can handle splitting it on spaces.

tests/e2e/chain/chain.go
```diff
@@ -41,6 +41,9 @@ func (c *internalChain) createAndInitValidators(count int) error {
 		if err := node.createKey("val"); err != nil {
 			return err
 		}
+		if err := node.createKey("val2"); err != nil {
```
alexanderbez, line 44: Why do we create two keys per validator?

tests/e2e/chain/export.go
```diff
@@ -8,16 +8,21 @@ type ChainMeta struct {
...
-	ConfigDir     string `json:"configDir"`
-	Index         int    `json:"index"`
-	Mnemonic      string `json:"mnemonic"`
-	PublicAddress string `json:"publicAddress"`
+	Name           string `json:"name"`
+	ConfigDir      string `json:"configDir"`
+	Index          int    `json:"index"`
+	Mnemonic       string `json:"mnemonic"`
+	PublicAddress  string `json:"publicAddress"`
+	PublicAddress2 string `json:"publicAddress2"`
```
alexanderbez, line 16: Can we describe or comment why we need two keys?

tests/e2e/chain/export.go
```diff
@@ -8,16 +8,21 @@ type ChainMeta struct {
...
+	PublicKey      string `json:"publicKey"`
+	OperAddress    string `json:"operAddress"`
 }
 
 type Chain struct {
 	ChainMeta  ChainMeta    `json:"chainMeta"`
 	Validators []*Validator `json:"validators"`
+	PropNumber int          `json:"propNumber"`
```
alexanderbez, line 24: What's prop number?

tests/e2e/e2e_setup_test.go
```diff
@@ -33,6 +33,14 @@ type syncInfo struct {
 	SyncInfo status `json:"SyncInfo"`
 }
 
+type valinfo struct {
+	VotingPower string `json:"VotingPower"`
```
alexanderbez, line 37: json tag should be `voting_power`, no?

tests/e2e/e2e_setup_test.go
```diff
@@ -33,6 +33,14 @@ type syncInfo struct {
...
+}
+
+type operInfo struct {
+	Bech32Val string `json:"Bech32 Val"`
```
alexanderbez, line 41: json tag should be `bech32_val`, no?

tests/e2e/e2e_util_test.go
czarcas7ic: @p0mvn and I had discussed moving all of this boilerplate to a separate function this morning, but we first wanted to ensure we got something that just works prior to the upgrade (we were hitting a bunch of issues getting our tests to work in the first place). Now that this is in, we can go back and redo this (we can do it pre PR merge as well, I have no issue with that)

tests/e2e/chain/chain.go
czarcas7ic: I did this because I didn't want to use a validator's self bonded wallet, since if we did this, both the superfluid stake that was bonded as well as the validators vote would change. Now that the water is less murky though, I feel like I could probably do this and just calculate what the new tally result would be with the validator re-voting

tests/e2e/chain/export.go
czarcas7ic: Prop Number keeps what the current prop number is. This way when we vote for the current prop, we can simply call this and then once it passes, increase this number by one. Please let me know if there is a better way for us to query what the most recent prop number is

tests/e2e/e2e_setup_test.go
czarcas7ic: This was originally being pulled from `osmosisd status` which shows as `VotingPower` even when queried as json. I actually no longer use this, so will remove it.

tests/e2e/chain/chain.go
p0mvn: Do we need a delegator account /wallet other than the validator's? I think we should name it accordingly since having 2 validator keys per node is confusing

tests/e2e/chain/config.go
```diff
@@ -171,6 +180,157 @@ func initGenesis(c *internalChain, votingPeriod time.Duration) error {
 	}
 	appGenState[banktypes.ModuleName] = bz
 
+	var stakeGenState staketypes.GenesisState
```
p0mvn, line 183: Let's add comments about what is happening below

tests/e2e/chain/export.go
p0mvn: Why do we need this passed from chain initialization? We should minimize the changes to this struct as it is only used for genesis / initial state initialization.

For the purposes of superfluid, we can create proposals either in `e2e_setup.go` it in `TestSuperfluidVoting` to avoid modifying this struct

tests/e2e/chain/export.go
```diff
@@ -8,16 +8,21 @@ type ChainMeta struct {
...
-	Mnemonic      string `json:"mnemonic"`
-	PublicAddress string `json:"publicAddress"`
+	Name           string `json:"name"`
+	ConfigDir      string `json:"configDir"`
+	Index          int    `json:"index"`
+	Mnemonic       string `json:"mnemonic"`
+	PublicAddress  string `json:"publicAddress"`
+	PublicAddress2 string `json:"publicAddress2"`
+	PublicKey      string `json:"publicKey"`
+	OperAddress    string `json:"operAddress"`
```
p0mvn, line 18: Same as with `Chain`, we should minimize the changes here as much as possible for backward compatibility. Why do we need all these?

tests/e2e/chain_init/chain-init.Dockerfile
```diff
@@ -13,14 +13,20 @@ RUN apk add linux-headers
...
+RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep f6282df732a13dec836cda1f399dd874b1e3163504dbd9607c6af915b2740479
+
+# CosmWasm: copy the right library according to architecture. The final location will be found by the linker flag `-lwasmvm_muslc`
+RUN cp /lib/libwasmvm_muslc.$(uname -m).a /lib/libwasmvm_muslc.a
+
 RUN BUILD_TAGS=muslc LINK_STATICALLY=true make build-e2e-chain-init
 
 ## Deploy image
 FROM ubuntu
+RUN apt-get update && apt-get install -y jq moreutils
```
p0mvn, line 29: Why is this needed?

tests/e2e/e2e_setup_test.go
```diff
@@ -167,6 +175,7 @@ func (s *IntegrationTestSuite) SetupSuite() {
 
 	for i, chainConfig := range s.chainConfigs {
 		s.runValidators(chainConfig, s.dockerImages.OsmosisRepository, s.dockerImages.OsmosisTag, i*10)
+		s.extractOperAddress(chainConfig)
```
p0mvn, line 178: What is `Oper`?

tests/e2e/e2e_test.go
```diff
@@ -1,14 +1,67 @@
 package e2e
 
 import (
-	"github.com/osmosis-labs/osmosis/v7/tests/e2e/chain"
+	"fmt"
+	"strconv"
+	"time"
 )
 
-func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
```
p0mvn, line 7: Let's keep this test

tests/e2e/e2e_test.go
```diff
@@ -1,14 +1,67 @@
...
+func (s *IntegrationTestSuite) TestSuperfluidVoting() {
 	chainA := s.chainConfigs[0].chain
-	chainB := s.chainConfigs[1].chain
+	s.Run("superfluid_vote_chainA", func() {
```
p0mvn, line 11: `s.Run(...)` is normally used for table drive tests to iterate over several test cases and create a runner for each. Here, we already have the test runner created with `TestSuperfluidVoting`, the name also makes sense so I think `s.Run(...)` is not needed

tests/e2e/e2e_test.go
```diff
@@ -1,14 +1,67 @@
...
+		s.depositProposal(chainA)
+		s.voteProposal(s.chainConfigs[0])
+		// set delegator vote to no
+		s.voteNoProposal(chainA, 0, "val2")
 
-	// compare coins of reciever pre and post IBC send
-	// diff should only be the amount sent
-	s.sendIBC(chainA, chainB, chainB.Validators[0].PublicAddress, chain.OsmoToken)
+		chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[chainA.ChainMeta.Id][0].GetHostPort("1317/tcp"))
+		sfProposalNumber := strconv.Itoa(chainA.PropNumber)
```
p0mvn, line 29: We shouldn't store this proposal number on the chain struct, instead, it should be stored in the test suite

tests/e2e/e2e_util_test.go
```diff
@@ -401,3 +592,220 @@ func (s *IntegrationTestSuite) createPool(c *chain.Chain, poolFile string) {
...
+		5*time.Minute,
+		time.Second,
+		"tx returned non code 0; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
+	)
+
+	s.T().Logf("successfully sent tx from %s container: %s", s.valResources[c.ChainMeta.Id][i].Container.Name[1:], s.valResources[c.ChainMeta.Id][i].Container.ID)
+
+}
+
+func (s *IntegrationTestSuite) extractOperAddress(chainConfig *chainConfig) {
```
p0mvn, line 715:
```suggestion
func (s *IntegrationTestSuite) extractValidatorOperatorAddress(chainConfig *chainConfig) {
```

tests/e2e/chain/export.go
p0mvn: Seems that we need a second address for the delegator to overwrite the vote, can we create it during the test itself? That way, we don't have to modify these structs to ensure backward compatibility and also wouldn't break the Validator's abstraction

tests/e2e/chain/export.go
```diff
@@ -8,16 +8,21 @@ type ChainMeta struct {
...
 }
 
 type Chain struct {
 	ChainMeta  ChainMeta    `json:"chainMeta"`
 	Validators []*Validator `json:"validators"`
+	PropNumber int          `json:"propNumber"`
+	LockNumber int          `json:"lockNumber"`
```
p0mvn, line 25: Just checked that both `PropNumber` and `LockNumber` are currently not used during initialization. So they should be in the test suite instead. These structs are primarily for passing chain initialization data between docker `chain_init` container and the test runner

tests/e2e/chain/chain.go
czarcas7ic: Went for adding it via the test instead of during initialization, thanks for the feedback!

tests/e2e/e2e_test.go
```diff
@@ -4,64 +4,76 @@ import (
...
+	// diff should only be the amount sent
+	s.sendIBC(chainA, chainB, chainB.Validators[0].PublicAddress, chain.OsmoToken)
+}
+
+func (s *IntegrationTestSuite) TestSuperfluidVoting() {
+	chainA := s.chainConfigs[0]
+	s.submitSuperfluidProposal(chainA, "gamm/pool/1")
+	s.depositProposal(chainA)
+	s.voteProposal(chainA)
+	s.ExecTx(chainA.chain.ChainMeta.Id, 0, []string{"chmod", "+x", "./add_account.sh"}, "")
```
p0mvn, line 25: How much more difficult is it to execute this transaction via RPC? I'm afraid this is not very extensive adding files for transaction

czarcas7ic: Do you know how to sign a tx with a key that isn't on your keyring? I spend some hours trying to figure out how to add it and I just kept running into problems

p0mvn: Since we are using a test keyring backend, it is storing keys insecurely on a disk. Since we are mounting a volume, it should be accessible on the host, should it not?

p0mvn: https://github.com/osmosis-labs/cosmos-sdk/blob/889979c0e22a744cd1e4de82140d915863aca219/crypto/keyring/doc.go#L42

czarcas7ic: Right but the only way I have been able to figure it out is via bash "osmosisd keys add NAME --recover" and haven't been able to get the key added through just golang (I'm positive its possible, but my lack of experience with it has made it challenging)

p0mvn: No worries, I'm not sure what the best way is either - just trying to brainstorm. 

After looking more into it - I don't think there is an RPC for that.

We could use some `testutil` functions directly in the test runner to create an account: https://github.com/osmosis-labs/cosmos-sdk/blob/0b8e530add916d724f4ea077bb174c275861647b/testutil/key.go#L31

or even import the `NewKeyring(...)` and `NewAccount(...)` functions directly: https://github.com/osmosis-labs/cosmos-sdk/blob/0b8e530add916d724f4ea077bb174c275861647b/crypto/keyring/keyring.go#L77

However, we've been trying to avoid depending on the codebase directly in tests so these are not the best way to go forward.

I think what we discussed offline is best - try to avoid piping the mnemonic and use an auto-generated one so that we can execute the command directly via dockertest

czarcas7ic: Ended up going with your suggestion of simply adding a new random key since it didn't need to have a balance initially (had to still use regex expression sadly to pull the address out). Thanks for the suggestion!

czarcas7ic:
This should be good to go now, unsure of why linter is failing though, looks like its having trouble pulling the linter docker image suddenly

tests/e2e/e2e_util_test.go
```diff
@@ -135,269 +142,215 @@ func (s *IntegrationTestSuite) sendIBC(srcChain *chain.Chain, dstChain *chain.Ch
...
 
-	outBufByte := outBuf.Bytes()
-	var balancesResp banktypes.QueryAllBalancesResponse
-	if err := util.Cdc.UnmarshalJSON(outBufByte, &balancesResp); err != nil {
-		return nil, err
-	}
+}
 
-	return balancesResp.GetBalances(), nil
+func (s *IntegrationTestSuite) extractValidatorOperatorAddress(chainConfig *chainConfig) {
```
p0mvn, line 308:
```suggestion
func (s *IntegrationTestSuite) extractValidatorOperatorAddresses(chainConfig *chainConfig) {
```

p0mvn (commented):
Really great job on minimizing code duplication and extracting these reusable methods!

A few minor questions and comments from me

tests/e2e/scripts/pool2A.json
```diff
@@ -1,6 +1,6 @@
 {
-    "weights": "5ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518,5stake",
```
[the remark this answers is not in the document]
czarcas7ic: It's likely that the account that normally would have the ibc asset due to the pre state creation txs does not have the required assets when we skip this step. This can easily be fixed by ensuring we set the receiver address to the account making the pools for when we skip the upgrade (and indirectly skip pre upgrade state creation)

tests/e2e/scripts/.bash_history
```diff
@@ -0,0 +1,8 @@
+osmosisd keys list --keyring-backend test
+osmosisd keys add wallet
+exit
+ls
+osmosisd keys list --keyring-backend test
+osmosisd q bank balances osmo16ssxe0l899487ye2rjss9yf0y7lgz9djucyadn
+osmosisd tx bank send val osmo133uej2dmeukrsx9du6kvgmltcg2xg7dsyssf7q 100000000000000000000gamm/pool/1 --chain-id osmo-test-a --from val -b block --keyring-backend test
+exit
```
ValarDragon: Accidentally committed?

tests/e2e/e2e_util_test.go
```diff
@@ -135,269 +142,215 @@ func (s *IntegrationTestSuite) sendIBC(srcChain *chain.Chain, dstChain *chain.Ch
...
-	defer cancel()
-
+func (s *IntegrationTestSuite) submitUpgradeProposal(config *chainConfig) {
+	c := config.chain
+	upgradeHeightStr := strconv.Itoa(config.propHeight)
 	s.T().Logf("submitting upgrade proposal on %s container: %s", s.valResources[c.ChainMeta.Id][0].Container.Name[1:], s.valResources[c.ChainMeta.Id][0].Container.ID)
+	cmd := []string{"osmosisd", "tx", "gov", "submit-proposal", "software-upgrade", upgradeVersion, fmt.Sprintf("--title=\"%s upgrade\"", upgradeVersion), "--description=\"upgrade proposal submission\"", fmt.Sprintf("--upgrade-height=%s", upgradeHeightStr), "--upgrade-info=\"\"", fmt.Sprintf("--chain-id=%s", c.ChainMeta.Id), "--from=val", "-b=block", "--yes", "--keyring-backend=test", "--log_format=json"}
+	s.ExecTx(c.ChainMeta.Id, 0, cmd, "code: 0")
+	s.T().Log("successfully submitted upgrade proposal")
+	config.propNumber = config.propNumber + 1
```
ValarDragon, line 158: amazing, this is nice

tests/e2e/chain/export.go
```diff
@@ -13,6 +13,8 @@ type Validator struct {
 	Index         int    `json:"index"`
 	Mnemonic      string `json:"mnemonic"`
 	PublicAddress string `json:"publicAddress"`
+	PublicKey     string `json:"publicKey"`
+	OperAddress   string `json:"operAddress"`
```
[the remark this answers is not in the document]
czarcas7ic: We still need to store this value for each validator so we are able to SFS to their operator address

tests/localosmosis/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN BUILD_TAGS=muslc LINK_STATICALLY=true make build
 # Runner
 # --------------------------------------------------------
 
-FROM alpine
```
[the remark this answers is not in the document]
czarcas7ic: Was accidental, changing back

czarcas7ic:
Changes made, let me know if we anything further done here. Once we merge this, CI will be ~10 minute for this until I merge in Roman's changes which should make non PRs to main much faster

ValarDragon (approved):
LGTM! We should chat about some more steps to help eliminate boilerplate / overhead further in a followup issue/PR

tests/e2e/chain/export.go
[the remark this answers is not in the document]
p0mvn: I was hoping to keep these `Validator` and `Chain` structs only used for transferring state between the "chain-initialization" part (that happens inside a Docker container) and the test runner part (in `e2e` package). The goal of this is to decouple the chain initialization and test runner components as much as possible. 

I know there is still some coupling present right now but the goal is to eventually get to completely independent components with incremental refactoring.

So this `OperAddress` is never used during chain initialization. I think if we want to keep adding more state to `Validator` or `Chain` structs, we should wrap them into structs that are local to the `e2e` package:
```
type validatorConfig struct {
	validator chain.Validator
}
```

We already do something similar with `chain.Chain`: https://github.com/osmosis-labs/osmosis/blob/360abcf3b256f8ebff41d1eddf04d96ffd5fd8b6/tests/e2e/e2e_setup_test.go#L40-L52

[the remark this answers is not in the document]
ValarDragon: Should we move this to an issue?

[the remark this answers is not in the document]
czarcas7ic: I can get to it now, if it proves to be a difficult change (it shouldn't be) then I will make an issue instead and merge

PR #1589, opened 2022-05-26
chore: Update Proto + ORM Support

alexanderbez:
A re-attempt of https://github.com/osmosis-labs/osmosis/pull/1455, but I've reached the capacity of my knowledge here. Maybe @robert-zaremba or @aaronc can chime in on how to get this to the finish line.

Currently, running `make proto-gen` yields:

```
make proto-gen
Generating Protobuf files
Generating gogo proto code
WARNING: field Pool.ScalingFactor is a repeated non-nullable native type, nullable=false has no effect
WARNING: field Pool.ScalingFactor is a repeated non-nullable native type, nullable=false has no effect
WARNING: field MsgStableSwapAdjustScalingFactors.ScalingFactors is a repeated non-nullable native type, nullable=false has no effect
Cleaning API directory
./scripts/protocgen2.sh: cd: line 14: can't cd to api: No such file or directory
make: *** [proto-gen] Error 2
```

It also produces an erroneous `v7/` dir in the root.


aaronc:
Seems like the `api` directory is missing

aaronc:
Doesn't hurt to try a mkdir -p

robert-zaremba:
If you don't use `pulsar` then you don't need to generate `api` content.

robert-zaremba:
Originally, I was thinking to not include the ./scripts/protocgen2.sh until you will like to use pulsar and ORM.

alexanderbez:
> Originally, I was thinking to not include the ./scripts/protocgen2.sh until you will like to use pulsar and ORM.

@ValarDragon stated we want to introduce support for ORM, but idk what that looks like.

ValarDragon:
Should we use Pulsar? I didn't realize ORM and Pulsar came together, my bad then. Would like to not have to force a switch of everything to the API module, I was under the impression from the regen codebase that just ORM things went there, and the normal proto files were being compiled as they are today

ValarDragon:
Those warnings are present on main as well

aaronc:
> Should we use Pulsar? I didn't realize ORM and Pulsar came together, my bad then. Would like to not have to force a switch of everything to the API module, I was under the impression from the regen codebase that just ORM things went there, and the normal proto files were being compiled as they are today

That's right. Currently there's a hybrid approach

alexanderbez:
I think we should still merge this as it uses the new buf stuff (heh), but just remove the `protogen2` script. WDYT @ValarDragon?

proto/buf.md
```diff
@@ -0,0 +1,3 @@
+# Protobufs
+
+This is the public protocol buffers API for the [Osmosis](https://github.com/osmosis-labse/osmosis).
```
p0mvn, line 3:
```suggestion
This is the public protocol buffers API for the [Osmosis](https://github.com/osmosis-labs/osmosis).
```

p0mvn (approved):

ValarDragon (approved):

Issue #1588, opened 2022-05-26
Add back balances_from_state_export.go


ValarDragon:
## Background

The `balances_from_state_export.go` file in #1301 got moved out of the CLI, and renamed to `balances_from_state_export.go.history` and we did not catch this in review. Its definitely needed in main / v8 / releases. It needs the `.history` extension deleted, and to be added back to cmd.

PR #1594, opened 2022-05-26
Add back balances_from_state_export_command

ValarDragon:
Closes: #1588 

This was erroneously commented out.


alexanderbez (approved):

p0mvn:
Should this be backported to v9 instead?

ValarDragon:
Should be backported to v9 as well

Issue #1601, opened 2022-05-27
Cache / incrementally Docker build


nikever:
For docker build times, I think we can replace the make docker-build-debug step in test.yml with a direct docker build invocation, and use cache fields, e.g. https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/issues/1578#issuecomment-1138822581_

PR #1602, opened 2022-05-27
[e2e] Add GitHub cache to docker build

nikever:
Closes: #1601 

This PR adds experimental GitHub Cache backend API to the `e2e_test` job.

This results in the following speed-up in `Build e2e image` step:

- Build without cache (4m 4s): https://github.com/nikever/osmosis/runs/6622443417?check_suite_focus=true
- Build after the cache it's filled (3m 10s) https://github.com/nikever/osmosis/runs/6622514813?check_suite_focus=true

## Brief Changelog

- AddGitHub Cache backend API to the e2e workflows

## Testing and Verifying

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,8 @@ jobs:
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
     steps:
-      - id: skip_check
+      -
+        id: skip_check
         uses: fkirc/skip-duplicate-actions@master
         with:
           cancel_others: "true" # workflow-runs from outdated commits will be cancelled.
@@ -23,7 +24,8 @@ jobs:
           skip_after_successful_duplicate: "true"
           paths: '["**/*.go", "**/*.mod", "**/*.sum"]'
           do_not_skip: '["pull_request", "workflow_dispatch", "schedule"]'
-      - name: Skipping test
+      -
+        name: Skipping test
         run: echo Should I skip tests? ${{ steps.skip_check.outputs.should_skip }}
 
   go_test:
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):
awesome, thanks! Another minute of wait time shaved :fire:

alexanderbez (approved):
🚀 🎸

alexanderbez:
I've noticed `Add new issues to project / Add issue to chain dev project` fail on many PRs -- what's up with that @p0mvn ?

ValarDragon:
I think it fails on anyone making a PR from a fork

p0mvn:
@czarcas7ic I think you are right, good catch. The purpose of e2e tests is to test the current/latest change. Unfortunately, if we are caching the older `osmosis:debug` image, we are losing the main purpose of the e2e tests.

I don't think this part can be optimized with the cache. We must be building the image every time. However, we can optimize by improving the image itself

nikever:
It builds a new image every time, however in this process it might re-use some cached docker layers (e.g. dependencies installed `apk add...` etc..). 

In the links above you can see that the `go build` happens in both cases (cached and not cached version).

p0mvn:
@nikever  I see, that's awesome! I misunderstood how this works at first. It seems that many steps are being cached for the Docker image-building process. We still end up copying the updated codebase and rebuilding the image every time: https://github.com/osmosis-labs/osmosis/runs/6627351333?check_suite_focus=true#step:8:261

czarcas7ic:
I assume I was just misunderstanding, nice work @nikever! Do we know why this fails on anyone making a PR from a fork / is there something we can do to fix that?

PR #1607, opened 2022-05-27
Token Factory: tests for burn

mattverse:
Closes: #XXX

Pulling as a draft as the test is failing for unknown reasons.

Tests for burning...Investigation is needed on why these tests are not succeeding. My guess is that it has something to do with the message sender of binding msgs are being called with the reflect address, not the sender address that gets passed in as a parameter in `executeCustom`


app/wasm/test/custom_msg_test.go
```diff
@@ -173,6 +176,50 @@ func TestMintMsg(t *testing.T) {
...
+	}}
+	err = executeCustom(t, ctx, osmosis, reflect, lucky, msg, sdk.Coin{})
+	require.NoError(t, err)
+
+	msg = wasmbindings.OsmosisMsg{BurnTokens: &wasmbindings.BurnTokens{
+		Denom:           sunDenom,
+		Amount:          amount,
+		BurnFromAddress: lucky.String(),
+	}}
+	err = executeCustom(t, ctx, osmosis, reflect, lucky, msg, sdk.Coin{})
```
mattverse, line 219: where error occurs

app/wasm/test/custom_msg_test.go
```diff
@@ -173,6 +176,50 @@ func TestMintMsg(t *testing.T) {
...
+
+	// lucky was broke
+	balances := osmosis.BankKeeper.GetAllBalances(ctx, lucky)
+	require.Empty(t, balances)
+
+	// Create denom for minting
+	msg := wasmbindings.OsmosisMsg{CreateDenom: &wasmbindings.CreateDenom{
+		Subdenom: "SUN",
+	}}
+	err := executeCustom(t, ctx, osmosis, reflect, lucky, msg, sdk.Coin{})
```
mattverse, line 199: the admin of the denom is not set to "lucky"( the msg sender), but the reflect address here for some reason

ValarDragon: This is correct. This is how the reflect contract works

ValarDragon: It executes the messages here / becomes the admin, lucky is the one who calls reflect contract.

app/wasm/test/custom_msg_test.go
ValarDragon: also correct, it should be BurnFromAddress = "", or reflect.String()

You would have to send the tokens from Lucky back to the reflect contract first, to get the burn to happen.

ValarDragon:
ah tests didn't run because of a github actions outage :/ https://www.githubstatus.com/

ValarDragon (approved):

Issue #1609, opened 2022-05-27
Panic on RegisterWhitelistedDirectUnbondPools (on non-mainnet)


joeabbey:
`RegisterWhitelistedDirectUnbondPools` checked for gamms that existed on mainnet

Other ChainIDs are unlikely to have the same pools, and may experience a halt, due to a panic in the safety check.

PR #1610, opened 2022-05-27
fix: Only apply v8 fork logic on osmosis-1

joeabbey:
Closes: #1609

The v8 fork contained parameter changes that were only applicable on the `"osmosis-1"` chainID.

## Brief Changelog

Limiting the `RunForkLogic` to osmosis-1

## Testing and Verifying

### DIFF
--- a/app/upgrades/v8/forks.go
+++ b/app/upgrades/v8/forks.go
@@ -9,6 +9,12 @@ import (
 // RunForkLogic executes height-gated on-chain fork logic for the Osmosis v8
 // upgrade.
 func RunForkLogic(ctx sdk.Context, appKeepers *keepers.AppKeepers) {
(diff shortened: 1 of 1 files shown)


andrewl33:
Would a test that runs through all upgrades catch regressions like this going forward? 


The tests could be fast if upgrade heights were in sequential.
It would mean [these values would have to be overridable though](https://github.com/osmosis-labs/osmosis/blob/4bf3099d8efd5f1d7c91a4473dfb9efd8f582ec9/app/upgrades/v8/constants.go#L14=), which might've been done before.

ValarDragon:
I think what should be done going forward is we add a flag / genesis file value for whether a given fork is following the Osmosis mainnet-only upgrades (things that refer to specific pool ID's, or incentive adjustments, etc.)

PR #1619, opened 2022-05-30
Try to add a check to prevent github secret call on PR's from fork

ValarDragon:

Attempt to fix erroneous CI errors on PR's from forks.
Solution taken from https://github.community/t/have-github-action-only-run-on-master-repo-and-not-on-forks/140840/14 , but not sure if I'm doing it right.

## Testing and Verifying

- Check if the job still run on this PR?
- See what happens on PRs from forks.


ValarDragon:
Doesn't look like this worked

PR #1622, opened 2022-05-30
Remove codecov

### LINKED ISSUE #1229: Share Go's test cache in CI between jobs to reduce latency for getting the results
**Context**

We would like to investigate and implement, if possible, the solution for sharing Go's test cache between test CI jobs to reduce the number of redundant runs

**Initial Investigation**
- The cache is stored under `GOCACHE` environment variable
   * Confirmed this by deleting the folder and rerunning the tests to see of they get cached.
- GitHub actions have artifacts and cache, we can probably store it there.
- Here's some info about caching and artifacts in GitHub Actions: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

**Risks**
- The proposed solution has never been tested
- From quick google, there are some issues with getting it work in Gitlab: https://stackoverflow.com/questions/45537023/gitlab-ci-cache-keep-golang-packages-between-stages
- However, nothing specific to GitHub

**Acceptance Criteria**
- investigate the solution and alternatives
- implement reusable test cache
- the solution is stable
   * it reduces the number of test runs across jobs
   * it runs the tests when  a change was made in a package

ValarDragon:
Closes #1229

Remove codecov / cover profile, enables test caches to be much more shared.

## Brief Changelog

Remove codecov from CI


ValarDragon:
I'm not sure why this isn't hitting cache. It hits cache for me locally :/

ValarDragon:
I'm still in favor of merging this personally, just to get rid of that codecov message. We should leave the underliyng issue open though. (Though perhaps deprioritize)

p0mvn (approved):

ValarDragon:
Going to merge, but we should keep an eye on:

1) if cache eventually gets used
2) build dir size doesn't massively grow. (If it does, we just introduce deleting it once a week or smth)

Issue #1623, opened 2022-05-30
E2E tests, make better CLI cmd arg formatting


ValarDragon:
## Background

Currently CLI commands have a very manually intensive formatting in the code, e.g. 
```golang
	cmd := []string{"osmosisd", "tx", "bank", "send", sendAddress, receiveAddress, amount, fmt.Sprintf("--chain-id=%s", c.meta.Id), "--from=val", "-b=block", "--yes", "--keyring-backend=test"}
```

Because everything routes through an `ExecTx` helper, we can actually apply better formatting / argument assumptions here. (We should make a generic `ExecCLI` function that `ExecTx` calls.

## Suggested Design

Make ExecTx automatically fill in the `fmt.Sprintf("--chain-id=%s", c.meta.Id), "-b=block", "--yes", "--keyring-backend=test"` arguments.

## Acceptance Criteria

- [ ] Remove this overhead from each tx invocation

ValarDragon:
We can also replace the `success` string for `ExecTx` with `expSuccess` as a boolean. And then the command maps that to `"code: 0"`, rather than every invokation doing it

PR #1625, opened 2022-05-30
Small test e2e updates / boilerplate notes

ValarDragon:

Some small things I noticed on reading the code, in addition to issues I'm making on the side!

## Brief Changelog

* Being a human linter
* Add some TODO's
* introduce a `NewGenesisEpochInfo` function, and use to eliminate overhead.

## Testing and Verifying

### DIFF
--- a/tests/e2e/chain/config.go
+++ b/tests/e2e/chain/config.go
@@ -69,6 +69,7 @@ var (
 	InitBalanceStrB = fmt.Sprintf("%d%s,%d%s", OsmoBalanceB, OsmoDenom, StakeBalanceB, StakeDenom)
 	OsmoToken       = sdk.NewInt64Coin(OsmoDenom, IbcSendAmount)  // 3,300uosmo
 	StakeToken      = sdk.NewInt64Coin(StakeDenom, IbcSendAmount) // 3,300ustake
+	tenOsmo         = sdk.Coins{sdk.NewInt64Coin(OsmoDenom, 10_000_000)}
 )
 
 func addAccount(path, moniker, amountStr string, accAddr sdk.AccAddress) error {
@@ -86,6 +87,7 @@ func addAccount(path, moniker, amountStr string, accAddr sdk.AccAddress) error {
 	balances := banktypes.Balance{Address: accAddr.String(), Coins: coins.Sort()}
 	genAccount := authtypes.NewBaseAccount(accAddr, nil, 0, 0)
 
+	// TODO: Make the SDK make it far cleaner to add an account to GenesisState
 	genFile := config.GenesisFile()
 	appState, genDoc, err := genutiltypes.GenesisStateFromGenFile(genFile)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_util_test.go
```diff
@@ -61,10 +61,6 @@ func (s *IntegrationTestSuite) ExecTx(chainId string, validatorIndex int, comman
 				return false
 			}
 
-			if err != nil {
-				return false
-			}
```
ValarDragon, line 66: Duplicate of lines 60-62

czarcas7ic (approved):
Thanks so much for cleaning up some of the mess I made! All looks good to me, one thing though; can you further explain what you are expecting in regards to this message https://github.com/osmosis-labs/osmosis/issues/1623#issuecomment-1141538814, I wasn't quite following.

ValarDragon:
You did a great job on these! Not really a mess here at all. Don't expect to build things perfectly the first time lol, getting something that works, and then iterating to improve it is the way. A lot of this highlights general improvements we need to make to the code overall!

For that particular comment, we see in `s.ExecTx(config.meta.Id, 0, cmd, "code: 0")` , "code: 0" as the sentinel for correctness in ~every call. Can just replace that with a bool for do we expect the tx to succeed vs fail. (Still need to provide our own sentinel for "ExecCLI")

czarcas7ic:
If thats the case though, we cannot use this function for hermes as well, which looks for a "Success" string or something along those lines instead of a "code: 0". So I think that is why I opted for the string so I could use it across the two. If this is not good practice though, would be happy to implement your suggestion!

czarcas7ic:
Here https://github.com/osmosis-labs/osmosis/blob/main/tests/e2e/e2e_util_test.go#L116
and here https://github.com/osmosis-labs/osmosis/blob/main/tests/e2e/e2e_util_test.go#L122

ValarDragon:
Those Hermes calls would be calls to `ExecCLI` not `ExecTx` !

czarcas7ic:
Gotcha, makes sense now, thanks!

tests/e2e/chain/config.go
```diff
@@ -86,6 +87,7 @@ func addAccount(path, moniker, amountStr string, accAddr sdk.AccAddress) error {
 	balances := banktypes.Balance{Address: accAddr.String(), Coins: coins.Sort()}
 	genAccount := authtypes.NewBaseAccount(accAddr, nil, 0, 0)
 
+	// TODO: Make the SDK make it far cleaner to add an account to GenesisState
```
p0mvn, line 90: Going to create an issue instead
```suggestion
```

p0mvn (approved):

tests/e2e/chain/config.go
p0mvn: https://github.com/osmosis-labs/osmosis/issues/1628

ValarDragon: I made one on the sdk earlier

ValarDragon: Also tracked here https://github.com/cosmos/cosmos-sdk/issues/12091

tests/e2e/chain/validator.go
```diff
@@ -270,8 +272,9 @@ func (v *internalValidator) signMsg(msgs ...sdk.Msg) (*sdktx.Tx, error) {
 
 	txBuilder.SetMemo(fmt.Sprintf("%s@%s:26656", v.nodeKey.ID(), v.instanceName()))
 	txBuilder.SetFeeAmount(sdk.NewCoins())
-	txBuilder.SetGasLimit(200000)
+	txBuilder.SetGasLimit(uint64(200000 * len(msgs)))
```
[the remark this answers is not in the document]
ValarDragon: If we do multi-messages gas consumption will increase notably and the limit won't suffice.

PR #1626, opened 2022-05-31
tests/e2e: Try using generics for improving config initialization

ValarDragon:

Aim to reduce boilerplate in config initialization via generics. 

## Brief Changelog

* Use generics to reduce config initialization overhead

## Testing and Verifying

### DIFF
--- a/tests/e2e/chain/config.go
+++ b/tests/e2e/chain/config.go
@@ -17,6 +17,7 @@ import (
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/gogo/protobuf/proto"
 	"github.com/spf13/viper"
 	tmconfig "github.com/tendermint/tendermint/config"
 	tmjson "github.com/tendermint/tendermint/libs/json"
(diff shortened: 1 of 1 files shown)


ValarDragon:
lol, looks like we just have to delete the dedicated `gofmt` CI job, as its not updated for go 1.18. (no update since 2020) Confused about the golangci failure, it should've been updated for go 1.18 support IIRC

ValarDragon:
Does this PR seem like a good idea? Its marginally net saving lines. I do like that its moving out some of the err checking.

If so, do we prefer the version in the first commit (with the param update in the main body) or the second commit, with that fully relegated to a separate function

tests/e2e/chain/config.go
```diff
@@ -416,6 +317,90 @@ func initGenesis(c *internalChain, votingPeriod time.Duration) error {
...
+		DistrEpochIdentifier: "day",
+	}
+}
+
+func updateMintGenesis(mintGenState *minttypes.GenesisState) {
+	mintGenState.Params.MintDenom = OsmoDenom
+	mintGenState.Params.EpochIdentifier = "day"
+}
+
+func updateGenUtilGenesis(c *internalChain) func(*genutiltypes.GenesisState) {
```
p0mvn, line 375: There is a PR in progress that aims to allow initializing state for single full nodes to be added to existing chains. Just linking it to help with reconciling the differences after one of these is merged: https://github.com/osmosis-labs/osmosis/pull/1587

p0mvn (approved):
LGTM

tests/e2e/chain/config.go
```diff
@@ -142,6 +143,21 @@ func addAccount(path, moniker, amountStr string, accAddr sdk.AccAddress) error {
 	return genutil.ExportGenesisFile(genDoc, genFile)
 }
 
+func updateModuleGenesis[V proto.Message](appGenState map[string]json.RawMessage, moduleName string, protoVal V, updateGenesis func(V)) error {
```
p0mvn, line 146: Seems that the problem is that it doesn't recognize generics syntax so we can just disable it?
```suggestion
// nolint: typecheck
func updateModuleGenesis[V proto.Message](appGenState map[string]json.RawMessage, moduleName string, protoVal V, updateGenesis func(V)) error {
```

ValarDragon:
I'm confused by golangci-lint in CI failing. It works for me locally, both with and without the nolint line :confused:

alexanderbez:
Ohhh, the problem is we're using the github actions job, whereas we need to use the makefile target, i.e `make lint`. E.g. https://github.com/umee-network/umee/blob/main/.github/workflows/lint.yml#L23-L33

alexanderbez:
Pushed a commit...let's see what happens

ValarDragon:
Oh sweet, were set then! I think we just have to remove the gofmt job (its checked anyway by golangci-lint)

The gofmt repo hasn't been updated since 2020

PR #1644, opened 2022-06-01
Add release pipeline to create artifacts for osmosis releases

nikever:
Continues: #1550 

This PR introduces an automatic release workflow via GitHub Actions to create all the artifacts for an osmosis release.
This workflow runs on every new tag or when manually triggered and it creates a release folder with all the osmosis binaries as workflow output.

## Brief Changelog

- Introduce automatic release workflow via GitHub Actions

## Testing and Verifying

Tested this workflow on a fork: https://github.com/nikever/osmosis/actions/runs/2422822418
The release folder is present in the `Artifacts` section:

image


alexanderbez (commented):
Does this use goreleaser behind the scenes?

nikever:
Not currently, the script creates all the release files manually.

We could switch to goreleaser but we need the PRO version as we would need a custom build process which afaik is not available in the free version.

ValarDragon (approved):
LGTM! I'm not quite sure that goreleaser helps us any tbh, docker file seems great

nikever:
I feel the same way, `goreleaser` won't add much at the moment.

We would still need to use this Docker process to create the binaries and then add `goreleaser` after it.

Issue #701, opened 2021-12-29
The proto structs for Balancer should use snake_case


sunnya97:
Currently, the field names all use camelCase, but I believe the standard we use for everything else is snake_case

pwang00:
Where are the proto structs located?  I tried searching under `osmosis/proto` but I wasn't able to find any relevant info, but I may have missed something.

ValarDragon:
They're located in files in this directory: https://github.com/osmosis-labs/osmosis/tree/main/proto/osmosis/gamm

hieuvubk:
Did anyone done this issue? 
@ValarDragon Could i countinue your work and close this issue ?

hieuvubk:
> Yes please go for it!

#1656
could u review this pr

PR #1656, opened 2022-06-02
Proto: camelCase to snake_case

hieuvubk:
Closes: #701


alexanderbez (approved):
utACK (did a quick skim)

ValarDragon (approved):
Lgtm, ty!!

ValarDragon:
Can you add a breaking change entry in the changelog? (Also in the future can you fill out the PR template please!)

Btw, also added you to the 'notional' github team, so in the future you can branch off the repo directly

pyramation:
utACK as well! @Thunnini can you take a look to verify this will change amino encoding on FE? I have a pretty good idea of the implications but have not tested.

This change would like require we update FE amino converter implementations (to be more like how it's done with cosmos package with underscore<->camel instead of camel<->camel). So in a sense, this is a good move in the right direction, and how other packages seem to be doing it.

hieuvubk:
> Can you add a breaking change entry in the changelog? (Also in the future can you fill out the PR template please!)
> 
> Btw, also added you to the 'notional' github team, so in the future you can branch off the repo directly

thank u, i edited changelog

pyramation:
thanks for updating all of these! super helpful. Let's coordinate a migration plan with @Thunnini @jonator for v9 or v10 so we can test this with FE.

hieuvubk:
@ValarDragon Va @pyramation 
I just gone through proto files and encountered some special cases in the _txfees_ module
Ảnh chụp Màn hình 2022-06-05 lúc 13 04 04
Ảnh chụp Màn hình 2022-06-05 lúc 13 04 25
Neither camelCase nor snake_case, so any change will affect to other code. What do u think?

pyramation:
> @ValarDragon Va @pyramation I just gone through proto files and encountered some special cases in the _txfees_ module Ảnh chụp Màn hình 2022-06-05 lúc 13 04 04 Ảnh chụp Màn hình 2022-06-05 lúc 13 04 25 Neither camelCase nor snake_case, so any change will affect to other code. What do u think?

I don't believe we're using those fields in FE yet — would it make sense to rename those to `fee_token` and `pool_id`?

hieuvubk:
> > @ValarDragon Va @pyramation I just gone through proto files and encountered some special cases in the _txfees_ module Ảnh chụp Màn hình 2022-06-05 lúc 13 04 04 Ảnh chụp Màn hình 2022-06-05 lúc 13 04 25 Neither camelCase nor snake_case, so any change will affect to other code. What do u think?
> 
> I don't believe we're using those fields in FE yet — would it make sense to rename those to `fee_token` and `pool_id`?

Oh i mean in chain code

ValarDragon:
Hey, this probably shouldn't have been merged yet, adjusting perms to improve this in the future.

Going to make a follow-up issue, to ensure we have a plan for how significant is this breaking change, and if we should revert it or bundle it with another large breaking change in the future.

ValarDragon:
Eh, its not that huge of a deal.

Going to make the issue for us to decide in

PR #1658, opened 2022-06-02
chore: unify docker proto container name

robert-zaremba:
Unify docker container name with the name used by Cosmos SDK.
The container is using the same docker image, so no need to duplicate hanging containers (each of them is few hundreds of MB).

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? rather no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,8 +174,8 @@ docs:
 
 protoVer=v0.7
 protoImageName=tendermintdev/sdk-proto-gen:$(protoVer)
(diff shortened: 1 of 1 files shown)


p0mvn (approved):

ValarDragon (approved):
LGTM, though admittedly I have a rather shallow understanding of our proto build process.

I'm down to merge and then if it breaks something we just revert

ValarDragon:
This PR breaks `make proto-all` for me, I think we should revert

PR #1662, opened 2022-06-03
Try adding github labeller actions for app-wiring and wasm

ValarDragon:

Adds a github labeller action for 'app-wiring' and 'wasm'.

## Testing and Verifying

I will submit a commit and see if it applies.

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? not applicable, CI / github tooling only

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,12 +1,7 @@
 "C:app-wiring":
   # Match all of app folder except wasm sub-folder
-  - 'app/*.go'
-  - 'app/apptesting/**/*'
-  - 'app/keepers/**/*'
-  - 'app/params/**/*'
-  - 'app/upgrades/**/*'
   # Also match any module.go update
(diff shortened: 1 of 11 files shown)


alexanderbez (approved):

ValarDragon:
I am very confused as to why this is not adding the `C:app-wiring` label. Its not obvious to me from the labeller docs how you're supposed to do OR's, but this seems to be right to me?

ValarDragon:
Last commit ( https://github.com/osmosis-labs/osmosis/pull/1662/commits/6d13a658b319a52d0ca8cabf1b0e49e32c24e76c ) seems very clearly like it should work to me, so I suspect that this is then a problem of these changes need to get into main for it to work?

ValarDragon:
ah yeah yhis is what it was. Post-merge, PRs are getting labelled as expected!

PR #1664, opened 2022-06-03
De-duplicate sim module setup

ValarDragon:

The simulation module setup code is an unnecessary large duplication of logic. This PR makes it get most of the simulation modules from run-time type casting, and allowing overrides where necessary.

This should ideally be upstreamed to the SDK, and then be a really simple helper method we call.

## Brief Changelog

- De-duplicate logic in simulation module setup

## Testing and Verifying

This change is a minor rework. Also not that impactful, as we don't really use the simulator within osmosis.

### DIFF
--- a/app/modules.go
+++ b/app/modules.go
@@ -20,7 +20,6 @@ import (
 	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
 	"github.com/cosmos/cosmos-sdk/x/authz"
 	authzmodule "github.com/cosmos/cosmos-sdk/x/authz/module"
-	"github.com/cosmos/cosmos-sdk/x/bank"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/cosmos/cosmos-sdk/x/capability"
 	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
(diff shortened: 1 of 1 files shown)


app/modules.go
```diff
@@ -237,35 +236,26 @@ func simulationModules(
...
-			app.StakingKeeper,
-			app.LockupKeeper,
-			app.GAMMKeeper,
-			app.EpochsKeeper,
-		),
-		app.TransferModule,
+	// recreate list of modules, to ensure no issues with overriding prior module structs.
+	modules := appModules(app, encodingConfig, skipGenesisInvariants)
+	overrideModules := map[string]module.AppModuleSimulation{
+		authtypes.ModuleName: auth.NewAppModule(appCodec, *app.AccountKeeper, authsims.RandomGenesisAccounts),
```
mattverse, line 242: asking to learn: is auth module the only "overriding" module as of right now? If so, what makes it an overriding module?

app/modules.go
```diff
@@ -237,35 +236,26 @@ func simulationModules(
...
-			app.AccountKeeper,
-			app.BankKeeper,
-			app.StakingKeeper,
-			app.LockupKeeper,
-			app.GAMMKeeper,
-			app.EpochsKeeper,
-		),
-		app.TransferModule,
+	// recreate list of modules, to ensure no issues with overriding prior module structs.
+	modules := appModules(app, encodingConfig, skipGenesisInvariants)
```
mattverse, line 240: Seems like the modules returned from `appModules` is different from the existing modules that were hard coded? Is this change intended / is there context that I missed here?

app/modules.go
ValarDragon: Yes it is the only overriding module!

Its an 'overriden' one, because it has a different argument than the 'native' version, the `authsims.RandomGenesisAccounts` arg. Every other module in Osmosis (and the SDK) is the same between the simulation and non-simulation version

app/modules.go
ValarDragon: The prior ones that weren't here were just forgotten tbh, because this was too much of a pain to maintain

app/modules.go
```diff
@@ -237,35 +236,26 @@ func simulationModules(
...
+	overrideModules := map[string]module.AppModuleSimulation{
+		authtypes.ModuleName: auth.NewAppModule(appCodec, *app.AccountKeeper, authsims.RandomGenesisAccounts),
+	}
+
+	simModules := []module.AppModuleSimulation{}
+	for _, appModule := range modules {
+		// for every module, see if we override it. If so, use override.
```
alexanderbez, line 247:
```suggestion
		// For every module, see if we override it. If so, use override.
```

alexanderbez (approved):

PR #1665, opened 2022-06-03
Delete the app interface

ValarDragon:

Deletes the app interface. I have no idea what this is for, and am not familiar with other apps using it. If they are, this is an interface that should be defined in the SDK, that we than enforce compatability with here.

## Brief Changelog

  - Delete App interface.

## Testing and Verifying

Deadcode deletion


ValarDragon:
Ah this came as an erroneous copy paste from: https://github.com/cosmos/cosmos-sdk/blob/main/simapp/types.go, so just need to change this interface to that.

PR #1666, opened 2022-06-03
Fix : inconsistent proto names

vuong177:
Closes: #XXX

 - Run `buf generate --template buf.gen.pulsar.yaml` have inconsistent proto error. 
 ```
protoc-gen-go-cosmos-orm: Go package "github.com/osmosis-labs/osmosis/api/osmosis/gamm/pool-models/balancer" has inconsistent names gammv1beta1 (osmosis/gamm/pool-models/balancer/balancerPool.proto) and balancerv1beta1 (osmosis/gamm/pool-models/balancer/tx.proto)
protoc-gen-go-grpc: Go package "github.com/osmosis-labs/osmosis/api/osmosis/gamm/pool-models/balancer" has inconsistent names gammv1beta1 (osmosis/gamm/pool-models/balancer/balancerPool.proto) and balancerv1beta1 (osmosis/gamm/pool-models/balancer/tx.proto)
protoc-gen-go-pulsar: Go package "github.com/osmosis-labs/osmosis/api/osmosis/gamm/pool-models/balancer" has inconsistent names gammv1beta1 (osmosis/gamm/pool-models/balancer/balancerPool.proto) and balancerv1beta1 (osmosis/gamm/pool-models/balancer/tx.proto)
```
 - This PR for  fix that bug.

## Brief Changelog

- Change package name in `balancerPool.proto` from `osmosis.gamm` to `osmosis.gamm.poolmodels.balancer.v1beta1`

## Testing and Verifying


mattverse (changes requested):
If I remember correctly, the reason we have package inconsistency for the balancer pool is that we have to have separate migration logic since changing the proto package would cause breaking changes. I'm personally in favor of keeping this the old way (cref: https://github.com/osmosis-labs/osmosis/pull/866)

For the superfluid module, we should've had the structure of `v1beta1` in the first place, but since we didn't I don't think we should break things and change to `v1beta1` now, I'm in favor of leaving that the way it is as well, and if we need further changes using `v1beta2` and so on.

What's the possible disadvantages of leaving `buf generate --template buf.gen.pulsar.yaml` errors / is there any other possible solutions available to get around this without state breaking?

vuong177:
> If I remember correctly, the reason we have package inconsistency for the balancer pool is that we have to have separate migration logic since changing the proto package would cause breaking changes. I'm personally in favor of keeping this the old way (cref: #866)
> For the superfluid module, we should've had the structure of `v1beta1` in the first place, but since we didn't I don't think we should break things and change to `v1beta1` now, I'm in favor of leaving that the way it is as well, and if we need further changes using `v1beta2` and so on.

Thanks for make me understand.

> What's the possible disadvantages of leaving `buf generate --template buf.gen.pulsar.yaml` errors / is there any other possible solutions available to get around this without state breaking?

I think we need fix for ORM/Pulsar support.

mattverse:
Oh i see this is a gnarly one given the context that we had multiple fixes before as well to solve the legacy problem of proto files( https://github.com/osmosis-labs/osmosis/pull/1469) do you mind if I close this and take this one over?

vuong177:
Yes. I close it now.

mattverse:
thanks for bringing up this issue @vuong177  btw, we want orm support asap also for twap 🔥

Issue #1657, opened 2022-06-02
chore: remove version from Proto schema go paths


alexanderbez:
1. Remove all versions from Protobuf go package declarations. E.g. https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/v1beta1/tx.proto#L7
2. Update `protogen.sh` to remove the `v7` suffix

ValarDragon:
We couldn't do this due to cross-module imports having things break

Issue #1256, opened 2022-04-15
Make CI for testing generated proto correctness


ValarDragon:
We should have a CI job that does `make proto-all`,and returns an error if there is a diff betrween the post `make proto-all` output, and any `.proto` or `.pb.go` file. This will help us catch whenever make proto-all wasn't ran

alexanderbez:
Great idea!

catShaark:
so we should check if .proto files is different, if so .pb.go files must be also different right

ValarDragon:
We need to check differences in both `.pb.go` files, and `.proto` files

ValarDragon:
Someone want to see if we can just copy what Tendermint's workflow here is: https://github.com/tendermint/tendermint/blob/mjf/unsocketing/.github/workflows/check-generated.yml

xBalbinus:
May be closed by [#1892](https://github.com/osmosis-labs/osmosis/pull/1892)

PR #1668, opened 2022-06-04
remove version from Proto schema go paths

AlpinYukseloglu:
Closes: #1657 

Removes version from Proto schema go paths.

## Brief Changelog

- Removed `v7` suffix from from all Protobuf go package declarations
- Updated protocgen.sh to remove the v7 suffix

## Testing and Verifying

This change is a trivial rework.

### DIFF
--- a/proto/osmosis/epochs/genesis.proto
+++ b/proto/osmosis/epochs/genesis.proto
@@ -5,7 +5,7 @@ import "gogoproto/gogo.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 
-option go_package = "github.com/osmosis-labs/osmosis/v7/x/epochs/types";
+option go_package = "github.com/osmosis-labs/osmosis/x/epochs/types";
 
 message EpochInfo {
   string identifier = 1;
--- a/proto/osmosis/epochs/query.proto
+++ b/proto/osmosis/epochs/query.proto
@@ -6,7 +6,7 @@ import "google/api/annotations.proto";
 import "cosmos/base/query/v1beta1/pagination.proto";
 import "osmosis/epochs/genesis.proto";
 
-option go_package = "github.com/osmosis-labs/osmosis/v7/x/epochs/types";
+option go_package = "github.com/osmosis-labs/osmosis/x/epochs/types";
 
 // Query defines the gRPC querier service.
 service Query {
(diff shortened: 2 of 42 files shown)


mattverse (approved):
Super Nice!! What do folks think aboout running `make proto-all` to apply all proto changes to main before merging? Disadvantage of doing so would be that it would break all open PRs, the advantage of doing so would be that if there's a PR that has to deal with proto changes, we would have to make the change anyway

ValarDragon:
We definitely should require this, should be enforced via CI: https://github.com/osmosis-labs/osmosis/issues/1256

The converse makes too many unrelated changes on other PRs

Did you run `make proto-all` on this PR @AlpinYukseloglu ?

AlpinYukseloglu:
> We definitely should require this, should be enforced via CI: #1256
> 
> The converse makes too many unrelated changes on other PRs

Agreed

> Did you run `make proto-all` on this PR @AlpinYukseloglu ?

No I did not – would you like me to?

ValarDragon:
yeah, every proto change must run this. In particular, we need to be sure its still working since were changing something here!

AlpinYukseloglu:
@ValarDragon I'm not sure if this is an issue with `make proto-all` as we were discussing in #1669, but when I run `make format` after running `make proto-all` I get a bunch of errors that look something like:

```
x/incentives/types/gauge.go:15:22: cannot use distrTo (variable of type "github.com/osmosis-labs/osmosis/v7/x/lockup/types".QueryCondition) as "github.com/osmosis-labs/osmosis/x/lockup/types".QueryCondition value in struct literal (typecheck)
                DistributeTo:      distrTo,
                                   ^
x/incentives/types/msgs.go:25:22: cannot use distributeTo (variable of type "github.com/osmosis-labs/osmosis/v7/x/lockup/types".QueryCondition) as "github.com/osmosis-labs/osmosis/x/lockup/types".QueryCondition value in struct literal (typecheck)
                DistributeTo:      distributeTo,
                                   ^
x/incentives/keeper/keeper.go:42:16: invalid operation: cannot compare k.hooks != nil (operator != not defined on untyped nil) (typecheck)
        if k.hooks != nil {
                      ^
```

Do you have any idea how I might be able to resolve this? Running `make format` on the most recent commit should replicate this.

ValarDragon:
@alexanderbez I'm pretty sure this doesn't work. This is making references to other proto files in the repo use their osmosis `v1` version (import with no version prefix)

PR #1669, opened 2022-06-04
Revert #1658

ValarDragon:

#1658 broke `make proto-all` on my machine. Is there something I need to do update the containers? Could others check if it still works for them on main?

## Testing and Verifying

Checkout the branch, make a proto comment change and try `make proto-all`. Check that a difference is generated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,8 +174,8 @@ docs:
 
 protoVer=v0.7
 protoImageName=tendermintdev/sdk-proto-gen:$(protoVer)
-containerProtoGen=cosmos-sdk-proto-gen-$(protoVer)
-containerProtoFmt=cosmos-sdk-proto-fmt-$(protoVer)
+containerProtoGen=osmosis-proto-gen-$(protoVer)
+containerProtoFmt=osmosis-proto-fmt-$(protoVer)
 
 proto-gen:
 	@echo "Generating Protobuf files"


mattverse:
hrrm `make proto-all` seems to work for me atm(getting few warnings on nullable options in the process but still works)

mattverse:
What's the error you get when you run `make proto-all`?

robert-zaremba:
hmm, it was working fine for me. Could you share the error message?

ValarDragon:
I don't get an error, the make commands just don't do anything. They print some lines, but don't generate any code. I tried editing comments in proto files, and `make proto-gen` and `make proto-all` would act as no-ops.

Changing this back to `osmosis` gets the commands to execute for me.

robert-zaremba:
I've tested on `main` and it worked without a problems:
+ added few comments 
+ run `make proto-all`
+ go files were updated - see attached diff
[proto-check.txt](https://github.com/osmosis-labs/osmosis/files/8848082/proto-check.txt)

robert-zaremba:
I think the problem @ValarDragon has is with some conflict - you probably were testing some versions and didn't delete it. It happens sometimes no matter which image you are using. Could you remove your running containers? Eg: `docker system prune`

ValarDragon:
Ah your spot on @robert-zaremba ! Thank you! This fixed it for me =)

PR #1672, opened 2022-06-05
Fix proto services for balancer and superfluid module

mattverse:

This fix is necessary in order to support ORM.

The legacy problem with proto package for balancer has been a problem for a long time, which has been fixed and addressed over multiple PRs(https://github.com/osmosis-labs/osmosis/pull/1208, https://github.com/osmosis-labs/osmosis/pull/1469). The latest decision that has been made upon this legacy package issue was to change the `tx.proto`(https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/gamm/pool-models/balancer/tx.proto) to the correct package while leaving `balacerPool.proto` under the legacy wrong package so that we wouldn't have to deal with migration logic. 

Even though such decision has fixed duplicate package issue that has been a blocker for typescript proto generation, it is now an issue with orm support, as protoc does not allow one package path to have different packages within the protofiles(ex. protoc-gen-go-grpc: Go package "github.com/osmosis-labs/osmosis/api/osmosis/gamm/pool-models/balancer" has inconsistent names gammv1beta1 (osmosis/gamm/pool-models/balancer/balancerPool.proto) and balancerv1beta1 (osmosis/gamm/pool-models/balancer/tx.proto)).

There would be multiple ways to fix this problem, open to other ways of fixing this as well.  

The way this PR solves it is physically creating a new folder for the `tx.proto` file under the `proto/gamm/pool-models/balancer` folder to avoid the problem stated above. 

---

As for the superfluid module, we have had inconsistency with the packages as well proto package for `gov.proto` set to `osmosis.superfluid.v1beta`, while the other proto packages have the proto package of `osmosis.superfluid`. The suggested fix in this PR is creating a new folder under `proto/superfluid` to fix the duplicate proto package path issue. 

## Brief Changelog

- Change proto package for `proto/gamm/pool-models/balancer/tx.proto` and `proto/superfluid/gov.proto`

## Testing and Verifying

Additional testing needed that this fix does not break unknown unknowns in the state side.

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?no


ValarDragon (approved):
Oh wow, yikes, thanks for doing this!

Looks correct to me! Can we add a quite readme in the `balancer` proto folder explaining the situation?

Issue #1418, opened 2022-05-05
Add table-driven tests to x/tokenfactory


sunnya97:
Currently, most of the tests for x/tokenfactory are "story-driven tests", but we should use more table driven tests with more edge cases to maximize code coverage.

ValarDragon:
I don;t think they are story driven tests at the moment either. (At least not for common definitions of story driven testing / acceptance criteria testing)

ValarDragon:
To consider this issue complete, we should have tests for:
* MsgCreateDenom
  * Try with subdenom too long
  * Subdenom having invalid characters
  * Subdenom & creator pair already exists
  * Success case
* MsgMintDenom
  * Denom does not exist
  * Mint is not by the admin
  * Try to overflow int limits with mint
  * Success
(Same as above for burn)
* MsgChangeAdminDenom
  * After setting to "", creator / prior admin can't mint
  * non-admins can't change the existing admin
  
There are more edge cases, but I think we can add them separately

PR #1691, opened 2022-06-06
Add table-driven tests to x/tokenfactory

vuong177:
Closes: #1418

I add table-driven tests for x/tokenfactory. It includes `TestMintDenom`, `TestBurnDenom`, `TestChangeAdminDenom`, `TestCreateDenom`


x/tokenfactory/keeper/admins_test.go
```diff
@@ -72,3 +74,221 @@ func (suite *KeeperTestSuite) TestAdminMsgs() {
...
+func (suite *KeeperTestSuite) TestMintDenom() {
+	suite.SetupTest()
+	var (
+		denom     string
+		msgServer types.MsgServer
+		addr0bal  = int64(0)
+	)
+
+	// Create a denom
+	msgServer = keeper.NewMsgServerImpl(*suite.App.TokenFactoryKeeper)
```
alexanderbez, line 87: 1. We should probably define and setup `msgServer` in the `suite` itself upon `SetupTest`
2. Also, why are we making explicit calls to `SetupTest`? Testify will automatically call `SetupTest` prior to each test.

vuong177: Sure. I'm going to change it

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -7,13 +7,16 @@ import (
...
+	"github.com/osmosis-labs/osmosis/v7/x/tokenfactory/keeper"
 	"github.com/osmosis-labs/osmosis/v7/x/tokenfactory/types"
 )
 
 type KeeperTestSuite struct {
 	apptesting.KeeperTestHelper
 
-	queryClient types.QueryClient
+	queryClient  types.QueryClient
+	msgServer    types.MsgServer
```
ValarDragon, line 18: oh interesting, we should definitely add this to our default test suite code structure everywhere!

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -7,13 +7,16 @@ import (
...
+	defaultDenom string
```
ValarDragon, line 19:
```suggestion
	// defaultDenom is on the suite, as it depends on the creator test address.
	defaultDenom string
```

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -50,11 +48,70 @@ func (suite *KeeperTestSuite) TestMsgCreateDenom() {
...
 	// Make sure that an address with a "/" in it can't create denoms
-	res, err = msgServer.CreateDenom(sdk.WrapSDKContext(suite.Ctx), types.NewMsgCreateDenom("osmosis.eth/creator", "bitcoin"))
+	res, err = suite.msgServer.CreateDenom(sdk.WrapSDKContext(suite.Ctx), types.NewMsgCreateDenom("osmosis.eth/creator", "bitcoin"))
 	suite.Require().Error(err)
 }
+
+func (suite *KeeperTestSuite) TestCreateDenom() {
+	for _, tc := range []struct {
+		desc     string
+		malleate func()
```
ValarDragon, line 63:
```suggestion
		setup func()
```

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -50,11 +48,70 @@ func (suite *KeeperTestSuite) TestMsgCreateDenom() {
...
+		subdenom string
+		valid    bool
+	}{
+		{
+			desc:     "subdenom too long",
+			malleate: func() {},
```
ValarDragon, line 69: Lets delete the function arg here, and just do it in the main body. (e.g. if `tc.setup != nil { tc.setup() }`)

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,270 @@
...
+func (suite *KeeperTestSuite) TestChangeAdminDenom() {
+	var (
+		denom    string
+		addr0bal int64
+		addr1bal int64
+	)
+
+	for _, tc := range []struct {
+		desc     string
+		malleate func()
```
ValarDragon, line 200:
```suggestion
		setup func()
```

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,270 @@
...
+				suite.Require().True(suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom).Amount.Int64() == addr0bal, suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom))
+			} else {
+				suite.Require().Error(err)
+				suite.Require().True(suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom).Amount.Int64() == addr0bal, suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom))
+			}
+		})
+	}
+}
+
+func (suite *KeeperTestSuite) TestChangeAdminDenom() {
```
ValarDragon, line 191: This tests logic due to the complex malleate's isn't very easy to follow. I'll spend some more time tmrw thinking about how to restructure it

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,270 @@
...
+
+	// Create a denom and mint
+	suite.CreateDefaultDenom()
+
+	// Make sure that the admin is set correctly
+	queryRes, err := suite.queryClient.DenomAuthorityMetadata(suite.Ctx.Context(), &types.QueryDenomAuthorityMetadataRequest{
+		Denom: suite.defaultDenom,
+	})
+	suite.Require().NoError(err)
+	suite.Require().Equal(suite.TestAccs[0].String(), queryRes.AuthorityMetadata.Admin)
```
ValarDragon, line 133: Can this be deleted?

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,270 @@
...
+
+	_, err = suite.msgServer.Mint(sdk.WrapSDKContext(suite.Ctx), types.NewMsgMint(suite.TestAccs[0].String(), sdk.NewInt64Coin(suite.defaultDenom, 10)))
+	addr0bal += 10
+	suite.Require().NoError(err)
+	suite.Require().True(suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom).Amount.Int64() == addr0bal, suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], suite.defaultDenom))
```
ValarDragon, line 138: What is this for? Making some tokens to burn?

These lines are hard to follow / the last seems non-standard equality checking syntax)

ValarDragon:
Thank you for making this PR! Apologies for how long it took to get another response, with the chain downtime and restart.

Left some comments on it!

x/tokenfactory/keeper/admins_test.go
vuong177: Sure. I deleted it.

vuong177:
@ValarDragon, thanks for your reviews. I've changed code for a better look.

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,262 @@
...
+		},
+	} {
+		suite.Run(fmt.Sprintf("Case %s", tc.desc), func() {
+			// setup test
+			suite.SetupTest()
+			addr0bal = 0
+			addr1bal = 0
+
+			// Create a denom and mint
+			res, err := suite.msgServer.CreateDenom(sdk.WrapSDKContext(suite.Ctx), types.NewMsgCreateDenom(suite.TestAccs[0].String(), "bitcoin"))
```
mattverse, line 250: same as comment left in `keeper_test.go` we can either utilzie that better or we can replace that by directly calling the methods from the test, i would prefer the latter option

x/tokenfactory/keeper/admins_test.go
```diff
@@ -1,74 +1,262 @@
...
 	suite.Require().NoError(err)
 	queryRes, err = suite.queryClient.DenomAuthorityMetadata(suite.Ctx.Context(), &types.QueryDenomAuthorityMetadataRequest{
-		Denom: res.GetNewTokenDenom(),
+		Denom: suite.defaultDenom,
 	})
 	suite.Require().NoError(err)
 	suite.Require().Equal("", queryRes.AuthorityMetadata.Admin)
 }
+
+func (suite *KeeperTestSuite) TestMintDenom() {
```
ValarDragon, line 69:
```suggestion
// TestMintDenom ensures the following properties of the MintMessage:
// * Noone can mint tokens for a denom that doesn't exist
// * Only the admin of a denom can mint tokens for it
// * The admin of a denom can mint tokens for it
func (suite *KeeperTestSuite) TestMintDenom() {
```

x/tokenfactory/keeper/admins_test.go
ValarDragon: What if instead of setup(), we added three fields:
* `ChangeAdminMsg`
* `exoectedAdmin`
* `*MintMsg` (and can be nil)
* `expMintPass`

Sorry for the delay it took me to come back to this

ValarDragon (approved):
Thanks for the PR! I suggest we merge it, and then do the followup on ChangeAdmin tests in a second PR

vuong177:
Yes sir. I'll make a second PR including your suggestion.

ValarDragon:
Thanks so much!

PR #1693, opened 2022-06-06
Fix make format

ValarDragon:

Fixes `make format`, needed `mvdan.cc` added to tools directory.

## Brief Changelog

* Add mvdan.cc/gofumpt to tools folder
* Restrict it to only run on `app/ x/ ante/`, it has this annoying problem that it cannot run on the entire root folder because it reads the `.git` folder. It inherits the problem from `gofmt` that it cannot ignore a particular directory.

## Testing and Verifying

  - How is the feature or change documented? Should land into contributing.md once its here


Makefile
```diff
@@ -261,7 +261,7 @@ lint:
 
 format:
 	@go run github.com/golangci/golangci-lint/cmd/golangci-lint run ./... --fix
-	@go run mvdan.cc/gofumpt -l -w .
+	@go run mvdan.cc/gofumpt -l -w x/ app/ ante/
```
ValarDragon, line 264: The unfortunate makefile update, gofmt / gofumpt have no way to ignore a directory from the path as far as I can tell

p0mvn: `tests/e2e` could also be useful
```suggestion
	@go run mvdan.cc/gofumpt -l -w x/ app/ ante/ tests/e2e
```

p0mvn (approved):

Makefile
ValarDragon: Great call!

PR #1699, opened 2022-06-07
Fix spotprice sigfig rounding for small spot prices

ValarDragon:

Fixes query bug on mainnet, for pools with an asset with `10**18` precision. We passed all spot prices through 8 sig fig rounding, however our sig fig rounding was incorrect for inputs with value less than 1.

This fixes a bug with spot price being called on values less than 1.

## Brief Changelog

- Fixes SpotPrice rounding on extremely small inputs

## Testing and Verifying

This change added a new test for a large input from mainnet, that was returning 0.
I also checked that spot price is not called anywhere in the state machine, except in the cosmwasm bindings.
We will have to ensure that no contract gets uploaded with spot price calls in the v9.x release series, which was also the general plan since contracts should depend on TWAP not spot price.

  - How is the feature or change documented? Correct code to the spec

### DIFF
--- a/osmomath/sigfig_round.go
+++ b/osmomath/sigfig_round.go
@@ -6,7 +6,7 @@ var pointOne = sdk.OneDec().QuoInt64(10)
 
 func SigFigRound(d sdk.Dec, tenToSigFig sdk.Int) sdk.Dec {
 	// for d > .1, we do round(d * 10^sigfig) / 10^sigfig
-	// for k, where 10^k*d > 1 && 10^{k-1}*d < .1, we do:
+	// for k, where 10^k*d > .1 && 10^{k-1}*d < .1, we do:
 	// (round(10^k * d * 10^sigfig) / (10^sigfig * 10^k)
 	// take note of floor div, vs normal div
 	k := uint64(0)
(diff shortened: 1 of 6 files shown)


osmomath/sigfig_round.go
```diff
@@ -0,0 +1,26 @@
...
+	k := uint64(0)
+	dTimesK := d
+	for ; dTimesK.LT(pointOne); k += 1 {
+		dTimesK.MulInt64Mut(10)
+	}
+	// d * 10^k * 10^sigfig
+	dkSigFig := dTimesK.MulInt(tenToSigFig)
+	numerator := dkSigFig.RoundInt().ToDec()
+	denominator := tenToSigFig
+	if k != 0 {
```
p0mvn, line 21: if k == 0, then `tenTok` is 1 so the conditional might be redundant

Please consider removing it. I found translating the comment above directly to code to be very helpful. At first, this check got me confused.

p0mvn (approved):
Nice work!

osmomath/sigfig_round.go
ValarDragon: Yeah I guess thats fair, don't need to optimize on two multiplications right now / there are better ways to do that.

PR #1702, opened 2022-06-07
thought experiment: zeroconf osmosis

faddat:
Basically, I wanted to put this idea out there to know if it is worth pursuing further.  It embeds the genesis.json file into osmosis. 

* let's do to genesis configuration what we did to seeds

....it's only ~~15~~, ~~30~~, 60? mins/dev/searchforfile/year right?

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

This code is incomplete and I'll update this section going forward.

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,5 +23,6 @@
   "[proto3]": {
     "editor.defaultFormatter": "xaver.clang-format"
   },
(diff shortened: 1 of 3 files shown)


ValarDragon:
Can we do:
```
genesis file does not exist. Download mainnet genesis from {...} ? (Y/N)
```

Though I do feel like the intended flow should be we just tell everyone to use get.osmosis.zone

ValarDragon:
Generally aligned with making the default config use Osmosis mainnet

Lets re-open the PR with some incremental steps towards this? (Not sure whats happening right now / seems stale)

I would also generally prefer we make the `get.osmosis.zone` UX as good as possible and direct everyone to install via that

PR #1708, opened 2022-06-08
fix: Fix `MaximalExactRatioJoin` to return correct share amount

mattverse:

This PR contains fix for ensuring lp shares from joining pool returns the correct amount, previously `minShareAmount` has been calculated off of `tokenIn` coin amount, not the pool liquidity, which has been returning LTE to 50% of lp shares than expected. This PR also includes unit test for `MaximalExactRatioJoin` to ensure we have the correct calculation.

## Testing and Verifying
Added test cases on `MaximalExactRatioJoin`, confirmed that the bug has been fixed by comparing test results with previous calculation and the updated calculation.


faddat (approved):
This looks great as a starting point.  Notional and DSRV team members have observed that this error is present on testnet.  Maybe a good next move is to cut a release containing only this change, and deploy that to testnet?

meetrick:
> This looks great as a starting point. Notional and DSRV team members have observed that this error is present on testnet. Maybe a good next move is to cut a release containing only this change, and deploy that to testnet?

I agree with you

hydrogen18 (approved):

alexanderbez (approved):
LGTM (utACK)...testing locally

rootwarp:
To check this problem, our team(DSRV) setup local network and confirm this problem.
As this PR described, `usedAmount` becomes too small because wrong value multiplied to ratio.

If you need more support, let us know. We are ready to test.

mconcat (approved):
utACK

heejin-github:
I think this change is reasonable.
Since `minShareRatio` is the ratio of the input coins to the total pool liquidity, `usedAmount` must be calculated as the multiply of the `minShareRatio` and the amount of Pool Liquidity.

However before this change, the `usedAmount` was calculated as the multiply of the `minShareRatio` and the amount of input coins.
This could be caused unexpectedly leaving a larger remaining coin.

ValarDragon (approved):
tACK

Issue #1710, opened 2022-06-08
Interchain Accounts module registration


ant1g:
#### System information

Osmosis version: `v9` or `main`
OS & Version: Linux ubuntu 22.04
Commit hash: 9165ac2

#### Expected behaviour
It should be possible to send ICA commands to Osmosis (via `MsgCreateBalancerPool` for instance), given that the ICA Host controller has already been integrated.

#### Actual behaviour
When sending the `MsgCreateBalancerPool` over ICA to Osmosis, the following error is reported:
```recovered: no concrete type registered for type URL /ibc.applications.interchain_accounts.v1.InterchainAccount against interface *types.AccountI"```

Meaning the ICA message codecs are not registered in the Osmosis application.

#### Steps to reproduce the behaviour
1. Start chain A (Quasar chain in my case, but can be any)
2. Start chain B (Osmosis on v9 tag, or main)
3. Run Hermes relayer connecting the two chains
4. Register an Interchain Account on Osmosis from chain A
5. Properly fund the registered ICA on Osmosis to be able to create a pool
6. Send the `MsgCreateBalancerPool` from chain A to Osmosis via ICA.

#### Backtrace

Error trace in the relayer (Hermes) output:

````
ERROR ThreadId(165) packet_cmd{src_chain=quasar src_port=icacontroller-quasar1sqlsc5024sszglyh7pswk5hfpc5xtl77gqjwec src_channel=channel-1 dst_chain=osmosis}:
	will retry: 
	schedule execution encountered error:
	failed with underlying error: gRPC call failed with status: status: Unknown, message: 
"recovered: no concrete type registered for type URL /ibc.applications.interchain_accounts.v1.InterchainAccount against interface *types.AccountI
stack:
goroutine 5122 [running]:
runtime/debug.Stack()
	/usr/lib/go-1.18/src/runtime/debug/stack.go:24 +0x65
github.com/cosmos/cosmos-sdk/baseapp.newDefaultRecoveryMiddleware.func1({0x1c29900, 0xc00458e810})
	/home/johndoe/go/pkg/mod/github.com/osmosis-labs/cosmos-sdk@v0.45.1-0.20220524162204-830f277f8259/baseapp/recovery.go:71 +0x27
github.com/cosmos/cosmos-sdk/baseapp.newRecoveryMiddleware.func1({0x1c29900?, 0xc00458e810?})
[...]

PR #1711, opened 2022-06-08
Register ICA module for ICA Host keeper

ant1g:
Closes: #1710

Register the ICA module basics in order to complete the ICA host integration to Osmosis, that is currently not working properly, as described in issue #1710 

## Brief Changelog

- *Register ICA module basics*

## Testing and Verifying

### DIFF
--- a/app/keepers/modules.go
+++ b/app/keepers/modules.go
@@ -3,6 +3,7 @@ package keepers
 import (
 	"github.com/CosmWasm/wasmd/x/wasm"
 	wasmclient "github.com/CosmWasm/wasmd/x/wasm/client"
+	ica "github.com/cosmos/ibc-go/v3/modules/apps/27-interchain-accounts"
 	transfer "github.com/cosmos/ibc-go/v3/modules/apps/transfer"
 	ibc "github.com/cosmos/ibc-go/v3/modules/core"
 	ibcclientclient "github.com/cosmos/ibc-go/v3/modules/core/02-client/client"
@@ -83,4 +84,5 @@ var AppModuleBasics = []module.AppModuleBasic{
 	tokenfactory.AppModuleBasic{},
 	bech32ibc.AppModuleBasic{},
 	wasm.AppModuleBasic{},
+	ica.AppModuleBasic{},
 }


ValarDragon:
Can you investigate go test failures?

ant1g:
> Can you investigate go test failures?

What I found: by adding the ICA module app basics to the osmosis app, it enables the ICA module InitGenesis function for the init chain sequence, which Binds the ICA port.
While running the upgrade later, the ICA module InitModule function is called, which tries to bind the same port.

Should the upgrade tests run the InitGenesis sequence in the first place? I am trying to understand if it is an issue in the testing, or in the app.

ant1g:
> > Can you investigate go test failures?
> 
> What I found: by adding the ICA module app basics to the osmosis app, it enables the ICA module InitGenesis function for the init chain sequence, which Binds the ICA port. While running the upgrade later, the ICA module InitModule function is called, which tries to bind the same port.
> 
> Should the upgrade tests run the InitGenesis sequence in the first place? I am trying to understand if it is an issue in the testing, or in the app.

I think we need to check if the port is not already bound before calling InitModule

ant1g:
> > > Can you investigate go test failures?
> > 
> > 
> > What I found: by adding the ICA module app basics to the osmosis app, it enables the ICA module InitGenesis function for the init chain sequence, which Binds the ICA port. While running the upgrade later, the ICA module InitModule function is called, which tries to bind the same port.
> > Should the upgrade tests run the InitGenesis sequence in the first place? I am trying to understand if it is an issue in the testing, or in the app.
> 
> I think we need to check if the port is not already bound before calling InitModule

I can do that, however I read [here](https://github.com/osmosis-labs/osmosis/blob/main/app/upgrades/v9/upgrades.go#L42) that InitGenesis should not run, or the ICA module InitGenesis is clearly being called when running the upgrade test.

Any help would be appreciated here! Thanks

ValarDragon:
Ah those are different, init genesis will still run at chain genesis, but not at the upgrade time

Anything we need to do to make init genesis work? (I haven't been looking at tests)

This is viable to get into v10 if its sufficiently simple.

ant1g:
> Ah those are different, init genesis will still run at chain genesis, but not at the upgrade time
> 
> Anything we need to do to make init genesis work? (I haven't been looking at tests)
> 
> This is viable to get into v10 if its sufficiently simple.

I believe that the issue is either the Upgrade tests are running the InitGenesis while they shouldn't OR we need to call the ICA InitModule, only if it has not been initialized first (meaning checking if the ICA port is already bound).

mattverse (commented):
So I looked into the test failures and the reason it's failing is because we're calling `InitGenesis` for the ica module twice (once in the init chainer, once more when we're calling the v9 upgrade test). 

My suggested take here is commenting out the v9 upgrade_test.go for versions higher than v9.

ValarDragon:
Ah interesting, because our app setup calls InitGenesis for the entire chain, not the v8-only modules. That makes sense.

This is unfortunate that theres not a clean way to keep this test around. I agree with the conclusion that for now we can comment the test out. We should make a backlog item for coming back and understanding how to re-instate it though, as this is an error we can get for testing addition of any new module in the future.

ValarDragon:
Can you take a look at #1825? Fixes the test failure as well

PR #1717, opened 2022-06-08
Add inverse relationship test on JoinPool and ExitPool

mattverse:
This PR adds inverse relationship test to ensure we are guaranteed the same amount returned upon joinpool and exitpool.

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/x/gamm/keeper/pool_service_test.go
+++ b/x/gamm/keeper/pool_service_test.go
@@ -430,6 +430,86 @@ func (suite *KeeperTestSuite) TestExitPool() {
 	}
 }
 
+// TestJoinPoolExitPool_InverseRelationship tests that joining pool and exiting pool
+// guarantees same amount in and out
+func (suite *KeeperTestSuite) TestJoinPoolExitPool_InverseRelationship() {
+	testCases := []struct {
+		name             string
+		pool             balancertypes.MsgCreateBalancerPool
+		joinPoolShareAmt sdk.Int
+	}{
+		{
+			name: "pool with same token ratio",
+			pool: balancer.NewMsgCreateBalancerPool(nil, balancer.PoolParams{
+				SwapFee: sdk.ZeroDec(),
+				ExitFee: sdk.ZeroDec(),
+			}, []balancertypes.PoolAsset{
+				{
+					Weight: sdk.NewInt(100),
+					Token:  sdk.NewCoin("foo", sdk.NewInt(10000)),
+				},
+				{
+					Weight: sdk.NewInt(100),
+					Token:  sdk.NewCoin("bar", sdk.NewInt(10000)),
+				},
+			}, defaultFutureGovernor),
+			joinPoolShareAmt: types.OneShare.MulRaw(50),
+		},
+		{
+			name: "pool with different token ratio",
+			pool: balancer.NewMsgCreateBalancerPool(nil, balancer.PoolParams{
+				SwapFee: sdk.ZeroDec(),
+				ExitFee: sdk.ZeroDec(),
+			}, []balancertypes.PoolAsset{
+				{
+					Weight: sdk.NewInt(100),
+					Token:  sdk.NewCoin("foo", sdk.NewInt(7000)),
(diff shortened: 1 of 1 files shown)


x/gamm/keeper/pool_service_test.go
```diff
@@ -430,6 +430,86 @@ func (suite *KeeperTestSuite) TestExitPool() {
...
+		balanceBeforeJoin := suite.App.BankKeeper.GetAllBalances(suite.Ctx, joinPoolAcc)
+		fmt.Println(balanceBeforeJoin.String())
+
+		err = suite.App.GAMMKeeper.JoinPoolNoSwap(suite.Ctx, joinPoolAcc, poolId, tc.joinPoolShareAmt, sdk.Coins{})
+		suite.Require().NoError(err)
+
+		_, err = suite.App.GAMMKeeper.ExitPool(suite.Ctx, joinPoolAcc, poolId, tc.joinPoolShareAmt, sdk.Coins{})
+
+		balanceAfterExit := suite.App.BankKeeper.GetAllBalances(suite.Ctx, joinPoolAcc)
+		deltaBalance, _ := balanceBeforeJoin.SafeSub(balanceAfterExit)
```
AlpinYukseloglu, line 502: Does depositing and withdrawing tokens preserve the order in which they were initially deposited? If not, does multi-asset deposit require a specific order for the input address? If any order suffices, this test could be potentially be circumvented by depositing multiple tokens in at different weights in a different order than what the pool outputs upon withdrawal.

In any case, we should add brief test case(s) that cover deposits/withdrawals with shuffled order of input assets on pools that aren't 50/50

mattverse: wdym by "ordering", and "depositing" and "withdrawing' tokens?

AlpinYukseloglu: Oh nvm I think JoinPoolNoSwap handles ordering etc.

mattverse: Oh yeah, this is directly using the method `JoinPool` no need to worry about stuff here

mconcat (approved):
LGTM

alexanderbez (approved):

x/gamm/keeper/pool_service_test.go
```diff
@@ -430,6 +430,86 @@ func (suite *KeeperTestSuite) TestExitPool() {
...
+				{
+					Weight: sdk.NewInt(100),
+					Token:  sdk.NewCoin("foo", sdk.NewInt(7000)),
+				},
+				{
+					Weight: sdk.NewInt(100),
+					Token:  sdk.NewCoin("bar", sdk.NewInt(10000)),
+				},
+			}, defaultFutureGovernor),
+			joinPoolShareAmt: types.OneShare.MulRaw(50),
```
p0mvn, line 473: Should we also have cases with different weights, with swap and exit fees?

x/gamm/keeper/pool_service_test.go
```diff
@@ -430,6 +430,86 @@ func (suite *KeeperTestSuite) TestExitPool() {
 	}
 }
 
+// TestJoinPoolExitPool_InverseRelationship tests that joining pool and exiting pool
+// guarantees same amount in and out
+func (suite *KeeperTestSuite) TestJoinPoolExitPool_InverseRelationship() {
```
p0mvn, line 435: Let's name it TestJoinPoolExitPool_Balancer_InverseRelationship since we can have a similar test for a stable swap pool?

p0mvn (approved):
LGTM. Had some questions about whether we need more test cases with different inputs

p0mvn:
@mattverse going to merge this to reduce cognitive overhead for tracking several PRs. Here, we have a few approvals and only my comments are left. My comments are unrelated to correctness and only suggest more test cases and renaming. If you agree with the suggestions, please follow up :+1:

PR #1713, opened 2022-06-08
chore: additional test cases for TestCalcJoinPoolShares (multi-coin swap)

alexanderbez:
Add two additional test cases to `TestCalcJoinPoolShares` where `tokensIn` contains multiple coins where a swap is performed.

PR #1721, opened 2022-06-08
Extract functions and methods from CalcJoinPoolShares; unit tests

p0mvn:
Closes: #XXX

Extract functions and methods in `CalcJoinPoolShares` to be able to unit test each one separately. Add more comments explaining steps in detail. Add unit tests for the newly extracted functions.

Functions that were created:
- `getPoolAssetsByDenom`
- `updateIntermediaryPoolAssets`
- `calcJoinMultipleSingleAssetTokensIn`

Additionally, added `calcSingleAssetJoin` test cases with various inputs. For each test case, verified the expected value with Wolfram. Most values are mismatched by an insignificant number that is checked with  multiplicative `ErrTolerance` of `"0.0000001"`
- Spotted that there was a bug with `ErrTolerance` related to when the additive difference is nil so made a quick fix

## Testing and Verifying

- `go test -timeout 30s -run ^TestGetPoolAssetsByDenom$ github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer`
- `go test -timeout 30s -run ^TestUpdateIntermediaryPoolAssets$ github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer`
- `go test -timeout 30s -run ^TestCalcSingleAssetJoin$ github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer`
- `go test -timeout 30s -run ^TestCalcSingleAssetJoin$ github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer`
- `go test -timeout 30s -run ^TestCalcJoinSingleAssetTokensIn$ github.com/osmosis-labs/osmosis/v7/x/gamm/pool-models/balancer`


osmoutils/binary_search.go
```diff
@@ -32,17 +32,17 @@ func (e ErrTolerance) Compare(expected sdk.Int, actual sdk.Int) int {
...
-		}
-	}
-
 	// Check additive tolerance equations
 	if !e.AdditiveTolerance.IsNil() {
+		// if no error accepted, do a direct compare.
+		if e.AdditiveTolerance.IsZero() {
+			if expected.Equal(actual) {
+				return 0
+			} else {
```
alexanderbez, line 41: You can remove the else clause

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+	return nil
+}
+
+// calcJoinMultipleSingleAssetTokensIn attempts to calculate single
+// asset join for all tokensIn given totalSharesSoFar,
+// poolAssetsByDenom and swapFee.
+//
+// Returns totalNumShares and totalNewLiquidity from joining all tokensIn
+// or error if fails to calculate join for any of the tokensIn.
+func (p *Pool) calcJoinMultipleSingleAssetTokensIn(tokensIn sdk.Coins, totalSharesSoFar sdk.Int, poolAssetsByDenom map[string]PoolAsset, swapFee sdk.Dec) (sdk.Int, sdk.Coins, error) {
```
alexanderbez, line 364:
```suggestion
func (p *Pool) calcJoinMultipleSingleAssetTokensIn(tokensIn sdk.Coins, currTotalShares sdk.Int, poolAssetsByDenom map[string]PoolAsset, swapFee sdk.Dec) (sdk.Int, sdk.Coins, error) {
```

alexanderbez, line 364: I personally think this reads a bit funny -- multiple and single. Hard to evaluate

p0mvn: renamed to `CalcJoinSingleAssetTokensIn`, Token**s** implies many

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+
+		// update total shares for accurate calcSingleAssetJoin calculation
+		newTotalShares := totalShares.Add(numShares)
+
+		newNumSharesFromRemaining, newLiquidityFromRemaining, err := p.calcJoinSingleAssetTokensIn(remainingLiquidity, newTotalShares, poolAssetsByDenom, swapFee)
+		if err != nil {
+			return sdk.ZeroInt(), sdk.NewCoins(), err
+		}
+		numShares.Add(newNumSharesFromRemaining)
+		newLiquidity = newLiquidity.Add(newLiquidityFromRemaining...)
```
p0mvn, line 312: Unit tests for `CalcJoinPoolShares` validating that these updates happen should be added in  a separate PR

p0mvn: https://github.com/osmosis-labs/osmosis/issues/1722

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -436,3 +436,683 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+				{
+					Token:  sdk.NewInt64Coin("uosmo", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+				{
+					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+			},
+			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000), sdk.NewInt64Coin("uatom", 50_000)),
```
stackman27, line 936: Wait is this multi asset deposit test? Does it relate to #1713?

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -436,3 +436,683 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+			expectShares: sdk.NewInt(2_499_999_968_750 * 2),
```
stackman27, line 937: Are you sure you can multiply (* 2) to calculate shares for two asset deposit with same amount of tokens In? 

cc @AlpinYukseloglu @alexanderbez

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -436,3 +436,683 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+					Token:  sdk.NewInt64Coin("uosmo", 2_000_000_000),
+					Weight: sdk.NewInt(500),
+				},
+				{
+					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+			},
+			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000), sdk.NewInt64Coin("uatom", 100_000)),
+			expectShares: sdk.NewInt(2_072_912_400_000_000 + 1_624_999_900_000),
```
stackman27, line 1047: Should this be `(2_072_912_400_000_000 + 812_483_500_000_000)` as P_issed is  P_issued = 812483500000000?

stackman27 (commented):
Thank you for the PR @p0mvn. Left few notes, really curious whether the calculation for multi-asset is accurate as i was trying to figure that out for hours today 😅

Thank you for all the human readable equation + wolfram too. Really helpful!

x/gamm/pool-models/balancer/amm_test.go
stackman27: > Are you sure you can add both shares to calculate total shares for two asset deposit ?
> 
> cc @AlpinYukseloglu @alexanderbez

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
 }
 
+// getPoolAssetsByDenom return a mapping from pool asset
+// denom to the pool asset itself. There must be no duplicates.
+// Returns error, if any found.
+func getPoolAssetsByDenom(poolAssets []PoolAsset) (map[string]PoolAsset, error) {
+	poolAssetsByDenom := make(map[string]PoolAsset)
+	for _, poolAsset := range poolAssets {
+		_, ok := poolAssetsByDenom[poolAsset.Token.Denom]
+		if ok {
```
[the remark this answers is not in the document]
p0mvn: Here, the goal is to check that there are no duplicates so if we detect that there is an item in the map indexed by an existing denom, we would like to error out

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+	totalNumShares := sdk.ZeroInt()
+	totalNewLiquidity := sdk.NewCoins()
+	for _, coin := range tokensIn {
+		newShares, err := p.calcSingleAssetJoin(coin, swapFee, poolAssetsByDenom[coin.Denom], currTotalShares)
+		if err != nil {
+			return sdk.ZeroInt(), sdk.Coins{}, err
+		}
+
+		totalNewLiquidity = totalNewLiquidity.Add(coin)
+		currTotalShares = currTotalShares.Add(newShares)
```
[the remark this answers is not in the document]
p0mvn: It is needed a few lines above for `calcSingleAssetJoin(...)`. We would like to provide all issued shares in pool so far, even the ones that were created by previous calls  to `calcSingleAssetJoin(...)`

However, instead of keeping this value as an accumulator and increasing it every iteration, we can provide `totalShares.Add(totalNewShares)` to `calcSingleAssetJoin(...)` so I changed it. 

Also, added comments explaining the difference between `totalShares` and `totalNewShares` to avoid confusion

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+		totalNumShares = totalNumShares.Add(newShares)
```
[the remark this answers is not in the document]
p0mvn: Yes, you are right. Changed the logic, names and added comments to better distinguish between the 2 as per https://github.com/osmosis-labs/osmosis/pull/1721/files#r893458890

x/gamm/pool-models/balancer/amm_test.go
p0mvn: It does not. After we do multi-asset maximal equal ratio join in`CalcJoinPoolShares`, there might be coins remaining if the share ratios [are not equal](https://github.com/osmosis-labs/osmosis/blob/dbefea1c6a4af4cbfd2e23d9d1dc8941e64e87d0/x/gamm/pool-models/internal/cfmm_common/lp.go#L82)

So we perform a single asset join on all remaining coins

x/gamm/pool-models/balancer/amm_test.go
p0mvn: As per my comment above, https://github.com/osmosis-labs/osmosis/pull/1721/files#r893463274

here, we can add shares because these are essentially done by single asset joining each token.

x/gamm/pool-models/balancer/amm_test.go
p0mvn: Technically, what multi-asset join in #1713 does is:
1. [maximal equal ration join](https://github.com/osmosis-labs/osmosis/blob/dbefea1c6a4af4cbfd2e23d9d1dc8941e64e87d0/x/gamm/pool-models/balancer/amm.go#L286) (it might leave some remaining tokens unjoined)
2. If any tokensIn remain, do a [single asset join on each individually](https://github.com/osmosis-labs/osmosis/blob/dbefea1c6a4af4cbfd2e23d9d1dc8941e64e87d0/x/gamm/pool-models/balancer/amm.go#L298-L313).

So, technically, this test is a step 2 of #1713 only

x/gamm/pool-models/balancer/amm_test.go
p0mvn: This is correct but the comment is wrong, nice catch

p0mvn:
> Thank you for the PR @p0mvn. Left few notes, really curious whether the calculation for multi-asset is accurate as i was trying to figure that out for hours today sweat_smile
> 
> Thank you for all the human readable equation + wolfram too. Really helpful!

Thanks for the detailed review, and great questions.

All comments are addressed now and ready for another round of :eyes:

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,100 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
-			if err != nil {
-				return sdk.ZeroInt(), sdk.NewCoins(), err
-			}
-
-			newLiquidity = newLiquidity.Add(coin)
-			newTotalShares = newTotalShares.Add(newShares)
-			numShares = numShares.Add(newShares)
+	if !remainingLiquidity.Empty() {
+		// update pool assets for accurate calcSingleAssetJoin calculation
+		if err := updateIntermediaryPoolAssets(newLiquidity, poolAssetsByDenom); err != nil {
```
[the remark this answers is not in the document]
mattverse: Can we make sure we document this or make it explicit that the `poolAssetByDenom` gets changed within the method?

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+// all liqidity coins must exist in poolAssetsByDenom. Returns error, if not.
+//
+// This is a helper function that is useful for updating the pool asset amounts
+// as an intermediary step in a multi-join methods such as CalcJoinPoolShares.
+// In CalcJoinPoolShares with multi-asset joins, we first attempt to do
+// a MaximalExactRatioJoin that might leave out some tokens in.
+// Then, for every remaining tokens in, we attempt to do a single asset join.
+// Since the first step (MaximalExactRatioJoin) affects the pool liqudity due to slippage,
+// we would like to account for that in the subsequent steps of single asset join
+func updateIntermediaryPoolAssets(liquidity sdk.Coins, poolAssetsByDenom map[string]PoolAsset) error {
```
mattverse, line 345: Can we also add within the comment that `poolAssetsByDenom` would get updated internally within whatever method calls this?

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
 		}
+
+		// update total shares for accurate calcSingleAssetJoin calculation
+		newTotalShares := totalShares.Add(numShares)
+
+		newNumSharesFromRemaining, newLiquidityFromRemaining, err := p.calcJoinSingleAssetTokensIn(remainingLiquidity, newTotalShares, poolAssetsByDenom, swapFee)
+		if err != nil {
+			return sdk.ZeroInt(), sdk.NewCoins(), err
+		}
+		numShares.Add(newNumSharesFromRemaining)
```
alexanderbez, line 311: there's no assignment here

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -436,3 +436,683 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+
+	errTolerance := osmoutils.ErrTolerance{
+		MultiplicativeTolerance: allowedErrRatioDec,
+	}
+
+	require.Equal(
+		t,
+		0,
+		errTolerance.Compare(expectedShares, actualShares),
+		fmt.Sprintf("expectedShares: %d, actualShares: %d", expectedShares.Int64(), actualShares.Int64()))
```
alexanderbez, line 1117:
```suggestion
		fmt.Sprintf("expectedShares: %d, actualShares: %d", expectedShares.Int64(), actualShares.Int64()),
	)
```

x/gamm/pool-models/balancer/amm_test.go
alexanderbez: Seems reasonable to me.

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
 	}
 
 	// Add all exact coins we can (no swap). ctx arg doesn't matter for Balancer.
-	numShares, remCoins, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
+	numShares, remainingLiquidity, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
```
mattverse, line 286: can we use the context we take in as the parameter?

x/gamm/pool-models/balancer/amm.go
p0mvn: Good catch, there is not test for this yet, it is coming in: https://github.com/osmosis-labs/osmosis/issues/1722

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -436,3 +436,683 @@ func TestCalcJoinPoolShares(t *testing.T) {
 		})
 	}
 }
+
```
mattverse, line 441:
```suggestion

// TestGetPoolAssetsByDenom tests if `GetPoolAssetsByDenom` succesfully creates a map of denom to pool asset
// given pool asset as parameter
```

x/gamm/pool-models/balancer/amm_test.go
p0mvn: Hmm, I tried this locally and it causes a syntax error

x/gamm/pool-models/balancer/amm.go
mattverse: Yeah should be `numShares = numShares.Add(newNumSharesFromRemaining)` 

We need test cases for this case as well seeing that this wasn't caught in the test cases

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -439,6 +439,669 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+			if tc.err != nil {
+				return
+			}
+
+			require.Equal(t, tc.expectedPoolAssetsByDenom, actualPoolAssetsByDenom)
+		})
+	}
+}
+
```
mattverse, line 523:
```suggestion

// TestUpdateIntermediaryPoolAssets tests if `updateIntermediaryPoolAssets` returns poolAssetsByDenom map
// with the updated liquidity given by the parameter
```

mattverse (commented):
Can we add additional test cases for `		numShares.Add(newNumSharesFromRemaining)
` that was not caught with the previous test cases?

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+//
+// all liqidity coins must exist in poolAssetsByDenom. Returns error, if not.
+//
+// This is a helper function that is useful for updating the pool asset amounts
+// as an intermediary step in a multi-join methods such as CalcJoinPoolShares.
+// In CalcJoinPoolShares with multi-asset joins, we first attempt to do
+// a MaximalExactRatioJoin that might leave out some tokens in.
+// Then, for every remaining tokens in, we attempt to do a single asset join.
+// Since the first step (MaximalExactRatioJoin) affects the pool liqudity due to slippage,
+// we would like to account for that in the subsequent steps of single asset join
```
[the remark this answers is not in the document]
p0mvn: Do you mean that the caller will have to update the internal state of the pool, meaning that this function is non-mutative?

Wouldn't you think that is implied since this is a function (not a method) and does not have access to the pool state?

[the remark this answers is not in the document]
p0mvn: Nvm, I see what you mean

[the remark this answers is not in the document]
p0mvn: The comment already mentions that on the first line: https://github.com/osmosis-labs/osmosis/blob/9c1a2da3e0a9b875680022f4a77c5daf65d6d294/x/gamm/pool-models/balancer/amm.go#L334

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
stackman27: I see i was just a bit confused with the name `ok`,  like if it's`ok` why would it give an error

x/gamm/pool-models/balancer/amm_test.go
ValarDragon: If the linters are happy with it, imo we leave it as is

x/gamm/pool-models/balancer/amm.go
p0mvn: Coming in #1722

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
 	}
 
 	// Add all exact coins we can (no swap). ctx arg doesn't matter for Balancer.
-	numShares, remCoins, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
+	numShares, remainingLiquidity, err := cfmm_common.MaximalExactRatioJoin(p, ctx, tokensIn)
```
ValarDragon, line 286: Lets not call this remainingLiquidity, but remainingTokensIn

p0mvn:
Moved `TestCalcSingleAsset` to a separate PR to ease the review process: https://github.com/osmosis-labs/osmosis/pull/1735

x/gamm/pool-models/balancer/amm.go
```diff
@@ -258,11 +261,10 @@ func (p *Pool) JoinPool(_ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (
 }
 
 // CalcJoinPoolShares
```
alexanderbez, line 263: We should finish the godoc :P

```suggestion
// CalcJoinPoolShares
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -281,39 +283,101 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
 		}
+
+		// update total shares for accurate calcSingleAssetJoin calculation
+		newTotalShares := totalShares.Add(numShares)
+
+		newNumSharesFromRemaining, newLiquidityFromRemaining, err := p.calcJoinSingleAssetTokensIn(remainingTokensIn, newTotalShares, poolAssetsByDenom, swapFee)
+		if err != nil {
+			return sdk.ZeroInt(), sdk.NewCoins(), err
+		}
+		numShares = numShares.Add(newNumSharesFromRemaining)
```
alexanderbez, line 311:
```suggestion
		numShares = numShares.Add(newNumSharesFromRemaining)

```

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
alexanderbez: `*, ok := foo()` is idiomatic Go for returning `<someType, bool>`, `ok` indicating true or false.

alexanderbez (approved):

p0mvn:
Seems that this has gotten quite a few eyes. Going to merge once CI passes and continue unit testing in follow-up PRs

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -445,6 +445,234 @@ func TestCalcJoinPoolShares(t *testing.T) {
 	}
 }
 
+// TestGetPoolAssetsByDenom tests if `GetPoolAssetsByDenom` succesfully creates a map of denom to pool asset
+// given pool asset as parameter
+func TestGetPoolAssetsByDenom(t *testing.T) {
+	testCases := []struct {
+		name                      string
+		poolAssets                []balancer.PoolAsset
+		expectedPoolAssetsByDenom map[string]balancer.PoolAsset
```
ValarDragon, line 454: This field seems redundant?

p0mvn: Why would you think so?

`poolAssets` is the input to the function we are testing, `expectedPoolAssetsByDenom` is the expected output.

and we are, essentially, testing the conversion from `[]balancer.PoolAsset` to `map[string]balancer.PoolAsset`

p0mvn: Moved this logic to: https://github.com/osmosis-labs/osmosis/pull/1738

mattverse (approved):

x/gamm/pool-models/balancer/amm.go
```diff
@@ -331,6 +333,53 @@ func getPoolAssetsByDenom(poolAssets []PoolAsset) (map[string]PoolAsset, error)
...
+// all liqidity coins must exist in poolAssetsByDenom. Returns error, if not.
+//
+// This is a helper function that is useful for updating the pool asset amounts
+// as an intermediary step in a multi-join methods such as CalcJoinPoolShares.
+// In CalcJoinPoolShares with multi-asset joins, we first attempt to do
+// a MaximalExactRatioJoin that might leave out some tokens in.
+// Then, for every remaining tokens in, we attempt to do a single asset join.
+// Since the first step (MaximalExactRatioJoin) affects the pool liqudity due to slippage,
+// we would like to account for that in the subsequent steps of single asset join.
+func updateIntermediaryPoolAssets(liquidity sdk.Coins, poolAssetsByDenom map[string]PoolAsset) error {
```
AlpinYukseloglu, line 347: suggestion for new name - see comment above

x/gamm/pool-models/balancer/amm.go
```diff
@@ -331,6 +333,53 @@ func getPoolAssetsByDenom(poolAssets []PoolAsset) (map[string]PoolAsset, error)
...
+	for _, coin := range liquidity {
+		poolAsset, ok := poolAssetsByDenom[coin.Denom]
```
AlpinYukseloglu, line 349: This line makes sure that the trust assumption isn't broken – all the validation for the assets passed in is done by this map.

x/gamm/pool-models/balancer/amm.go
```diff
@@ -282,34 +284,34 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
-			newShares, err := p.calcSingleAssetJoin(coin, swapFee, poolAssetsByDenom[coin.Denom], newTotalShares)
-			if err != nil {
-				return sdk.ZeroInt(), sdk.NewCoins(), err
-			}
-
-			newLiquidity = newLiquidity.Add(coin)
-			newTotalShares = newTotalShares.Add(newShares)
-			numShares = numShares.Add(newShares)
+	if !remainingTokensIn.Empty() {
+		// update pool assets for accurate calcSingleAssetJoin calculation
```
AlpinYukseloglu, line 300: Doesn't calling `MaximalExactRatioJoin` already update the pool? Why do we need to add assets into the pool in a separate function?

x/gamm/pool-models/balancer/amm.go
```diff
@@ -282,34 +284,34 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
+		if err := updateIntermediaryPoolAssets(newLiquidity, poolAssetsByDenom); err != nil {
+			return sdk.ZeroInt(), sdk.NewCoins(), err
+		}
+
+		// update total shares for accurate calcSingleAssetJoin calculation
+		newTotalShares := totalShares.Add(numShares)
+
+		newNumSharesFromRemaining, newLiquidityFromRemaining, err := p.calcJoinSingleAssetTokensIn(remainingTokensIn, newTotalShares, poolAssetsByDenom, swapFee)
```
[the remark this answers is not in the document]
AlpinYukseloglu: After digging around a bit it seems like it relies on `poolAssetsByDenom[coin.Denom]` throwing an error if the asset is not in the pool, but worth confirming this here just to make sure

x/gamm/pool-models/balancer/amm.go
```diff
@@ -331,6 +333,53 @@ func getPoolAssetsByDenom(poolAssets []PoolAsset) (map[string]PoolAsset, error)
...
+
+// calcJoinSingleAssetTokensIn attempts to calculate single
+// asset join for all tokensIn given totalShares in pool,
+// poolAssetsByDenom and swapFee. totalShares is the number
+// of shares in pool before beginnning to join any of the tokensIn.
+//
+// Returns totalNewShares and totalNewLiquidity from joining all tokensIn
+// by mimicking individually single asset joining each.
+// or error if fails to calculate join for any of the tokensIn.
+func (p *Pool) calcJoinSingleAssetTokensIn(tokensIn sdk.Coins, totalShares sdk.Int, poolAssetsByDenom map[string]PoolAsset, swapFee sdk.Dec) (sdk.Int, sdk.Coins, error) {
```
AlpinYukseloglu, line 368: Am I correct in saying that this function assumes `len(tokensIn)` < `len(poolAssets)` i.e. that the number of tokens in `tokensIn` is less than the total number of assets/denoms in the pool since otherwise, we would run `MaximalExactRatioJoin` instead of running `calcSingleAssetJoin` on each asset?

If so, it might be worth enforcing this with a check at the beginning of the function so that we only have one valid way of doing a multi-asset join: `MaximalExactRatioJoin` following by `calcJoinSingleAssetTokensIn` on the remaining assets

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"math/rand"
 	"testing"
-	"time"
+	time "time"
```
AlpinYukseloglu, line 7: nit: this change doesn't actually do anything right?

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -445,6 +445,150 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+	testCases := []struct {
+		name string
+
+		// returns newLiquidity, originalPoolAssetsByDenom, expectedPoolAssetsByDenom
+		setup func() (sdk.Coins, map[string]balancer.PoolAsset, map[string]balancer.PoolAsset)
+
+		err error
+	}{
+		{
+			name: "regular case with multiple pool assets and a subset of newLqiduity to update",
```
AlpinYukseloglu, line 460: nit: "newLqiduity" -> "newLiquidity"

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -445,6 +445,150 @@ func TestCalcJoinPoolShares(t *testing.T) {
 	}
 }
 
+// TestUpdateIntermediaryPoolAssets tests if `updateIntermediaryPoolAssets` returns poolAssetsByDenom map
+// with the updated liquidity given by the parameter
+func TestUpdateIntermediaryPoolAssets(t *testing.T) {
```
AlpinYukseloglu, line 450: see comment above about `len(tokensIn)`

x/gamm/pool-models/balancer/amm_test.go
AlpinYukseloglu: I'm confused as to why `expectShares` is double what is in the previous test when the total liquidity isn't being doubled. From the [balancer whitepaper](https://balancer.fi/whitepaper.pdf):

> An “all-asset” deposit has to follow the distribution of existing assets in the pool. If the deposit contains 10%
of each of the assets already in the pool, then the Value Function will increase by 10% and the depositor will
be minted 10% of the current outstanding pool token supply.

Also as per my comment above, it's not clear to me why we're implementing/testing a case where we single asset join two assets into a 2-asset pool when `MaximalExactRatioJoin` would do the trick with no fees (and would be the intended/expected behavior when a pool is joined)

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -653,6 +797,236 @@ func TestCalcSingleAssetJoin(t *testing.T) {
...
+					Token:  sdk.NewInt64Coin("uosmo", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+				{
+					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
+					Weight: sdk.NewInt(100),
+				},
+			},
+			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000), sdk.NewInt64Coin("uatom", 50_000)),
+			expectShares: sdk.NewInt(2_487_500_000_000 * 2),
```
AlpinYukseloglu, line 925: Same question as my comment above

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -653,6 +797,236 @@ func TestCalcSingleAssetJoin(t *testing.T) {
...
+			// 	swapFeeRatio = (1 - (1 - W_t) * swapFee)
+			// Plugging all of this in, we get:
+			// 	Full solution: https://www.wolframalpha.com/input?i=100+*10%5E18*%28%281+%2B+%2850000*%281+-+%281-%28500+%2F+%28500+%2B+100%29%29%29+*+0.03%29%2F2000000000%29%29%5E%28500+%2F+%28500+%2B+100%29%29+-+1%29
+			// 	Simplified:  P_issued = 2_072_912_400_000_000
+			//
+			//
+			// For uatom:
+			//
+			// Expected output from Balancer paper (https://balancer.fi/whitepaper.pdf) using equation (25) on page 10:
+			// P_issued = P_supply * ((1 + (A_t * swapFeeRatio  / B_t))^W_t - 1)
```
AlpinYukseloglu, line 949: The equation in the whitepaper linked (eq (25) on page 10) shows this exact equation but without the `swapFeeRatio` – am I looking at the wrong whitepaper or is the version with swap fees derived elsewhere?

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -653,6 +797,236 @@ func TestCalcSingleAssetJoin(t *testing.T) {
...
+			require.NoError(t, err)
+
+			// estimate expected liquidity
+			expectedNewLiquidity := sdk.NewCoins()
+			for _, tokenIn := range tc.tokensIn {
+				expectedNewLiquidity = expectedNewLiquidity.Add(tokenIn)
+			}
+
+			totalNumShares, totalNewLiquidity, err := balancerPool.CalcJoinSingleAssetTokensIn(tc.tokensIn, pool.GetTotalShares(), poolAssetsByDenom, tc.swapFee)
+			// It is impossible to set up a test case with error here so we omit it.
```
AlpinYukseloglu, line 1014: If we have a length check (as described in my comment above), an error case could be when a set with too many tokens is passed in

x/gamm/pool-models/balancer/amm_test.go
p0mvn: > I'm confused as to why expectShares is double what is in the previous test when the total liquidity isn't being doubled

The logic was the following - here we are adding 2 tokens in that have equal weights and amounts in, so the liquidity would double. In the previous case, there was only one token in.

However, it seems that these should still not be equal because after the first join we would have more shares created for one of the tokens, right? As a result, the second join would have fewer shares returned than the first one. Seems that this test is passing because of the error tolerance

x/gamm/pool-models/balancer/amm.go
p0mvn: While I agree that this check is useful, I think that we should have it in the caller of this function, not inside this function itself.

A function should work independently of its caller. Otherwise, we would add an invariant that might restrict its use in other cases.

I will investigate this further

x/gamm/pool-models/balancer/amm_test.go
p0mvn: fixed

x/gamm/pool-models/balancer/amm_test.go
p0mvn: Replied in that comment

x/gamm/pool-models/balancer/amm.go
mattverse: AFAIK, `MaximalExactRatioJoin` shouldnt be updating pool itself, just calculate the maximal amount of share available for join

x/gamm/pool-models/balancer/amm_test.go
p0mvn: > Also as per my comment above, it's not clear to me why we're implementing/testing a case where we single asset join two assets into a 2-asset pool when MaximalExactRatioJoin would do the trick with no swap fees (and would be the intended/expected behavior when a pool is joined) – am I missing something

The goal was to make the function be independent of what the caller might do. I think the suggested check is useful but should belong to the caller of this function. Let me know what you think

x/gamm/pool-models/balancer/amm_test.go
p0mvn: swapFeeRatio is based on our discussions offline in terms of how to integrate the swap fee into the calculation

p0mvn: I just copied your original format. That's why it might be referring to the calculation without the swap fee

p0mvn: I can add a comment saying "with feeRatioAdded" if this is confusing

x/gamm/pool-models/balancer/amm.go
p0mvn: What do you think about having this check here: 
https://github.com/osmosis-labs/osmosis/commit/f6071c9757448dc929f3933b3ead336efcbcd9e3

x/gamm/pool-models/balancer/amm.go
p0mvn: Yes, I believe @mattverse is correct here

x/gamm/pool-models/balancer/amm.go
p0mvn: Going with your suggestion

x/gamm/pool-models/balancer/amm_test.go
p0mvn: Let's continue the conversation about this in that thread :+1:

x/gamm/pool-models/balancer/amm.go
```diff
@@ -282,34 +284,34 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
-			if err != nil {
-				return sdk.ZeroInt(), sdk.NewCoins(), err
-			}
-
-			newLiquidity = newLiquidity.Add(coin)
-			newTotalShares = newTotalShares.Add(newShares)
-			numShares = numShares.Add(newShares)
+	if !remainingTokensIn.Empty() {
+		// update pool assets for accurate calcSingleAssetJoin calculation
+		if err := updateIntermediaryPoolAssets(newLiquidity, poolAssetsByDenom); err != nil {
```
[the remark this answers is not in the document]
p0mvn: Name changed

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
p0mvn: Will add this check for extra security and a unit test.

I think that this should also be done on the keeper layer but I don't think that it is always ensured.
An example where I think it's not checked when it should be: https://github.com/osmosis-labs/osmosis/blob/111c1747b213ce4993cde7e4d3629f05ba8cf252/x/gamm/keeper/pool_service.go#L262

@AlpinYukseloglu @mattverse @alexanderbez  @ValarDragon wdyt?

x/gamm/pool-models/balancer/amm_test.go
p0mvn: Will resolve this once figure out an appropriate path forward

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
p0mvn: `calcSingleAssetJoin` is updated with check though:
https://github.com/osmosis-labs/osmosis/pull/1721/commits/63334c8ca85b16dfd5314d8c4e509fab55ddaa54

Nice catch, Alpin

x/gamm/pool-models/balancer/amm_test.go
p0mvn: comment added to every test case where applicable

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
AlpinYukseloglu: LGTM - good check to have as a guardrail imo so we don't rely on SDK-level map implementation throwing errors

x/gamm/pool-models/balancer/amm.go
AlpinYukseloglu: > A function should work independently of its caller. Otherwise, we would add an invariant that might restrict its use in other cases.

Generally agree with this - probably a good idea to leave it as it is to not break abstraction

> What do you think about having this check here: [f6071c9](https://github.com/osmosis-labs/osmosis/commit/f6071c9757448dc929f3933b3ead336efcbcd9e3)

This addresses a separate but important point & lets us have an error case we can use for testing. Great addition

x/gamm/pool-models/balancer/amm_test.go
AlpinYukseloglu: > > I'm confused as to why expectShares is double what is in the previous test when the total liquidity isn't being doubled
> 
> The logic was the following - here we are adding 2 tokens in that have equal weights and amounts in, so the liquidity would double. In the previous case, there was only one token in.
> 
> However, it seems that these should still not be equal because after the first join we would have more shares created for one of the tokens, right? As a result, the second join would have fewer shares returned than the first one. Seems that this test is passing because of the error tolerance

This is interesting – I wonder if we can set up a case where this falls outside of error tolerance just for proper coverage? An example could be deposit amounts that are very large relative to the pool liquidity (i.e. in this case we're depositing 50_000 uosmo into a pool with 1_000_000_000_000 uosmo – what if the amount we deposit was much higher so the 2nd transaction generates significantly fewer LP shares than the first?)

AlpinYukseloglu: > > Also as per my comment above, it's not clear to me why we're implementing/testing a case where we single asset join two assets into a 2-asset pool when MaximalExactRatioJoin would do the trick with no swap fees (and would be the intended/expected behavior when a pool is joined) – am I missing something
> 
> The goal was to make the function be independent of what the caller might do. I think the suggested check is useful but should belong to the caller of this function. Let me know what you think

I hear you on this, agreed!

x/gamm/pool-models/balancer/amm_test.go
AlpinYukseloglu: Makes sense - thanks for adding the comments!

x/gamm/pool-models/balancer/amm.go
```diff
@@ -282,34 +289,34 @@ func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk
...
 		return sdk.ZeroInt(), sdk.NewCoins(), err
 	}
 
-	// update liquidity for accurate calcSingleAssetJoin calculation
-	newLiquidity = tokensIn.Sub(remCoins)
-	for _, coin := range newLiquidity {
-		poolAsset := poolAssetsByDenom[coin.Denom]
-		poolAsset.Token.Amount = poolAssetsByDenom[coin.Denom].Token.Amount.Add(coin.Amount)
-		poolAssetsByDenom[coin.Denom] = poolAsset
-	}
```
ValarDragon, line 296: Comment for posterity, these lines were moved to line 306

x/gamm/pool-models/balancer/amm.go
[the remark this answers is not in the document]
alexanderbez: LGTM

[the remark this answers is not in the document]
p0mvn: Moving this change to its own PR: https://github.com/osmosis-labs/osmosis/pull/1751

ValarDragon:
LGTM to merge after the last few commits

Issue #548, opened 2021-10-23
Add governance proposal type to change pool params


sunnya97:
Can be used to standardize Swap Fee and Exit Fees before per-pool governance exists.

daniel-farina:
@romelukaku  @sunnya97 
Is this ticket still work in progress or was it already merged with a PR? 

Thank you!

sunnya97:
It was not already merged.  @romelukaku are you still working on it?

romelukaku:
no I'm not, sorry I haven't done this already
I'll start now

mconcat:
What is the status on this issue? If it is available I would like to start working on it

sunnya97:
@hieuvubk Which PR?

hieuvubk:
> Change pool params (swap fee, exit fee) via proposal #1725

https://github.com/osmosis-labs/osmosis/pull/1725

PR #1725, opened 2022-06-09
Change pool params (swap fee, exit fee) via proposal

hieuvubk:
Closes: #548

> Add change pool swap fee proposal

> Add change pool exit fee proposal

> Add test


x/gamm/keeper/gov/gov_test.go
```diff
@@ -0,0 +1,191 @@
...
+					Title:       "tittle",
+					Description: "des",
+					Content: types.SetSwapFeeContent{
+						PoolId:  1,
+						SwapFee: sdk.MustNewDecFromStr("1.1"),
+					},
+				})
+
+				suite.Require().Error(err, types.ErrTooMuchSwapFee)
+			},
```
mattverse, line 57: Can we take these out and have them tested separately instead of testing them inside the function? Also, would it be possible to have more test cases?(superfluid module gov tests are a good example)

mattverse (changes requested):
Apologies for the late review here, haven't noticed that this PR was left open for so long.

Nice work on adding these proposals overall! 

Left some feedbacks and comments, please take a look!

x/gamm/keeper/gov/gov_test.go
hieuvubk: Thank you for the comments. I'll fix them

mattverse: no problem! lmk once its ready for review!

hieuvubk: Yeah I think it's ready. Could u pls take a look

PR #1737, opened 2022-06-09
Add specified documentation and comments on JoinPool flow

mattverse:
Closes: #XXX

Adds specified comments on join pool flow. 

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -42,6 +42,7 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/app/keepers"
 	appparams "github.com/osmosis-labs/osmosis/v7/app/params"
 	"github.com/osmosis-labs/osmosis/v7/app/upgrades"
+	v10 "github.com/osmosis-labs/osmosis/v7/app/upgrades/v10"
 	v3 "github.com/osmosis-labs/osmosis/v7/app/upgrades/v3"
 	v4 "github.com/osmosis-labs/osmosis/v7/app/upgrades/v4"
 	v5 "github.com/osmosis-labs/osmosis/v7/app/upgrades/v5"
@@ -87,7 +88,7 @@ var (
 	_ sdksimapp.App = (*OsmosisApp)(nil)
 
 	Upgrades = []upgrades.Upgrade{v4.Upgrade, v5.Upgrade, v7.Upgrade, v9.Upgrade}
-	Forks    = []upgrades.Fork{v3.Fork, v6.Fork, v8.Fork}
+	Forks    = []upgrades.Fork{v3.Fork, v6.Fork, v8.Fork, v10.Fork}
 )
 
 // GetWasmEnabledProposals parses the WasmProposalsEnabled and
--- a/app/upgrades/v10/constants.go
+++ b/app/upgrades/v10/constants.go
@@ -1,16 +1,39 @@
-package v9
+package v10
 
 import (
 	"github.com/osmosis-labs/osmosis/v7/app/upgrades"
 
-	store "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// UpgradeName defines the on-chain upgrade name for the Osmosis v9 upgrade.
+// Last executed block on the v9 code was 4713064.
+// Last committed block is assumed to be 4713064, as we have block proposals that were not precommitted upon
+// for 4713065.
+const ForkHeight = 4713065
+
+// UpgradeName defines the on-chain upgrade name for the Osmosis v9-fork for recovery.
+// This is not called v10, due to this bug that would require a state migration to fix:
 const UpgradeName = "v10"
 
-var Upgrade = upgrades.Upgrade{
-	UpgradeName:          UpgradeName,
-	CreateUpgradeHandler: CreateUpgradeHandler,
-	StoreUpgrades:        store.StoreUpgrades{},
+// RecoveryAddress that the irregular state change transfers to.
+// TODO: Include derivation of this
+var RecoveryAddress, recoveryAddressErr = sdk.AccAddressFromBech32("osmo1rdkpu0tfnp3vx7vg4gxhjr0gt9rtydtv4fsrd0")
+
+func init() {
+	if recoveryAddressErr != nil {
+		panic("recovery address decoding failure")
+	}
+}
+
+// Created synthetically via fork
+// var Upgrade = upgrades.Upgrade{
+// 	UpgradeName:          UpgradeName,
+// 	CreateUpgradeHandler: CreateUpgradeHandler,
+// 	StoreUpgrades:        store.StoreUpgrades{},
+// }
+
+var Fork = upgrades.Fork{
+	UpgradeName:    UpgradeName,
+	UpgradeHeight:  ForkHeight,
+	BeginForkLogic: RunForkLogic,
 }
--- a/app/upgrades/v10/fork.go
+++ b/app/upgrades/v10/fork.go
@@ -1,0 +1,26 @@
+package v10
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	"github.com/osmosis-labs/osmosis/v7/app/keepers"
+)
+
+func RunForkLogic(ctx sdk.Context, appKeepers *keepers.AppKeepers) {
+	if ctx.BlockHeight() != ForkHeight {
+		panic(fmt.Sprintf("current height %d, expected it to be the fork height %d", ctx.BlockHeight(), ForkHeight))
+	}
+	plan := upgradetypes.Plan{
+		Name:   UpgradeName,
+		Height: ForkHeight,
+		Info:   "",
+	}
+	err := appKeepers.UpgradeKeeper.ScheduleUpgrade(ctx, plan)
+	if err != nil {
+		panic(err)
+	}
+	appKeepers.UpgradeKeeper.SetUpgradeHandler(plan.Name, CreateUpgradeHandler(appKeepers))
+}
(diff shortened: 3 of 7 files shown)


x/gamm/keeper/pool_service.go
```diff
@@ -177,6 +177,9 @@ func (k Keeper) CreatePool(ctx sdk.Context, msg types.CreatePoolMsg) (uint64, er
 // JoinPoolNoSwap determines the maximum amount that can be LP'd without any swap,
 // by looking at the ratio of the total LP'd assets. (e.g. 2 osmo : 1 atom)
 // It then finds the maximal amount that can be LP'd.
+//
+// All pool interfaces handled within this method are pointer references, `JoinPool` directly updates the pool interface.
```
p0mvn, line 181: I think it makes sense to have this comment inside the method but not as a spec.

Method specs should not be concerned with the implementation details so that we can treat methods as black boxes

p0mvn, line 181: an interface cannot be a pointer so let's simplify to avoid confusion

```suggestion
// All pools handled within this method are pointer references, `JoinPool` directly updates the pools
```

x/gamm/keeper/pool_service.go
```diff
@@ -189,12 +192,13 @@ func (k Keeper) JoinPoolNoSwap(
 		return err
 	}
 
+	// we do an abstract calculation on the lp liquidity coins needed to have the designated amount of given shares of the pool
```
p0mvn, line 195: Does it make sense to elaborate by talking about the fact that no swap would be done?

x/gamm/keeper/pool_service.go
```diff
@@ -189,12 +192,13 @@ func (k Keeper) JoinPoolNoSwap(
...
 	neededLpLiquidity, err := getMaximalNoSwapLPAmount(ctx, pool, shareOutAmount)
 	if err != nil {
 		return err
 	}
 
-	// if neededLPLiquidity >= tokenInMaxs, return err
+	// check that needed lp liquidity does not exceed `tokenInMaxs`. Return error if so.
```
p0mvn, line 201: I think it makes it clearer that this parameter is provided by the caller

```suggestion
	// check that needed lp liquidity does not exceed the given `tokenInMaxs` parameter. Return error if so.
```

x/gamm/keeper/pool_service.go
mattverse: make sense. added a line here, lmk what you think or if you have better ideas for the wording!

x/gamm/keeper/msg_server.go
```diff
@@ -84,6 +84,22 @@ func (server msgServer) CreatePool(goCtx context.Context, msg types.CreatePoolMs
 	return poolId, nil
 }
 
+// JoinPool routes `JoinPoolNoSwap` where we do an abstract calculation on needed lp liquidity coins to get the designated
+// amount of share for the pool. (This is done by taking the number of shares we want and then using total number of shares
```
p0mvn, line 88:
```suggestion
// amount of shares for the pool. (This is done by taking the number of shares we want and then using the total number of shares
```

x/gamm/keeper/msg_server.go
```diff
@@ -84,6 +84,22 @@ func (server msgServer) CreatePool(goCtx context.Context, msg types.CreatePoolMs
...
+// to get the ratio of the pool it accounts for. Using this ratio, we iterate through all pool assets to get how much tokens we need
```
p0mvn, line 89:
```suggestion
// to get the ratio of the pool it accounts for. Using this ratio, we iterate over all pool assets to get the number of tokens we need
```

x/gamm/keeper/msg_server.go
```diff
@@ -84,6 +84,22 @@ func (server msgServer) CreatePool(goCtx context.Context, msg types.CreatePoolMs
...
+// `JoinPool` takes in the tokensIn calculated above as the parameter rather than using the number of shares provided in the msg.
+// This can result in negotiable difference between the number of shares provided within the msg
+// and the actual number of share amount resulted from joining pool.
+// Internal logic flow for each pool model is as follows:
+// Balancer: TokensIn provided as the argument must be either a single token or tokens containing all assets in the pool.
+// 			 For the case of a single token, we simply perform single asset join (balancer notation: pAo, pool shares amount out,
+// 			 given single asset in).
+//			 For the case of multi-asset join, we first calculate the maximal amount of tokens that can be joined whilst maintaining
+// 			 pool asset's ratio without swap. We then iterate through the remaining coins that couldn't be joined
+// 			 and perform single asset join on each token.
```
p0mvn, line 102: What do you think about moving this to a README? 

While this is very useful for a doc, I don't think it should be a method spec. For the same reason I mentioned earlier - method/function specs should not be concerned about implementation details to avoid breaking the black box.

I understand that the goal here is to simplify the review process for the community/validators. However, while doing so, we should aim to avoid violating engineering practices

p0mvn: So I suggest keeping a short description for the spec with regards to inputs and outputs and moving this excellent write-up to a README

x/gamm/pool-models/balancer/amm.go
```diff
@@ -250,16 +250,25 @@ func (p *Pool) calcSingleAssetJoin(tokenIn sdk.Coin, swapFee sdk.Dec, tokenInPoo
 	).TruncateInt(), nil
 }
 
+// JoinPool calculates the number of shares needed and the updated liquidity for joining pool
```
p0mvn, line 253: Also, consider elaborating on what kind of errors are returned and given which inputs / state of the pool

```suggestion
// JoinPool calculates the number of shares needed given tokensIn with swapFee applied.
// It updates the liquidity if the pool is joined successfully. If not, returns error.
```

x/gamm/pool-models/balancer/amm.go
```diff
@@ -250,16 +250,25 @@ func (p *Pool) calcSingleAssetJoin(tokenIn sdk.Coin, swapFee sdk.Dec, tokenInPoo
...
 }
 
-// CalcJoinPoolShares
+// CalcJoinPoolShares calculates the number of shares created to join pool with the provided amount of `tokenIn`.
+// When a single token is provided as an argument, we simply perform single asset join with the token.
+// If tokenIn provided as an argument isn't a sinlge token, it must contain all the tokens in the pool.
+// For the case of multi-asset join for a pool, we first calculate the maximum amount we can join a pool without swap, then
+// perform single asset join for the remaining coins.
+// CalcJoinPoolShares does not directly alter the state of the pool, but only does the calculation for shares for joining the pool.
 func (p *Pool) CalcJoinPoolShares(_ sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, err error) {
```
p0mvn, line 272: This is not specific to this PR. I just kept thinking about this while working on this method over these past few days. Basically, I think that we should have separate methods for single tokenIn calc and multi tokensIn calc.

The way it is done currently is violating the "Single responsibility principle" in SOLID where the method does 2 things.
It makes it less testable and less abstract IMO. We should consider addressing this post v10. Please create an issue if you agree. If not, please ignore

x/gamm/pool-models/internal/cfmm_common/lp.go
```diff
@@ -50,8 +50,16 @@ func CalcExitPool(ctx sdk.Context, pool types.PoolI, exitingShares sdk.Int, exit
...
+// MaximalExactRatioJoin calculates the maximal amount of tokens that can be joined whilst maintaining pool asset's ratio
+// returning the number of shares that'd be and how many coins would be left over.
+// 		e.g) suppose we have a pool of 10 foo tokens and 10 bar tokens, with the total amount of 100 shares.
+//			 if `tokensIn` provided is 1 foo token and 2 bar tokens, `MaximalExactRatioJoin`
+//			 would be returning (10 shares, 1 bar token, nil)
+// This can be used when `tokensIn` are not guaranteed the same ratio as assets in the pool.
+// Calculation for this is done in the following steps.
+//		1. iterate through all the tokens provided as an argument, calculate how much ratio it accounts for the asset in the pool
+//		2. get the minimal share ratio that would work as the benchmark for all tokens.
+//		3. calculate the number of shares that could be joined (total share * min share ratio), return the remaining coins
```
p0mvn, line 62: Also feels like something that should belong in a README, not a spec. Consider moving please

p0mvn (approved):
Excellent write-up. Had some suggestions about spec vs doc. Overall, LGTM

x/gamm/pool-models/balancer/amm.go
ValarDragon: Agree

x/gamm/pool-models/internal/cfmm_common/lp.go
ValarDragon: Why is a code comment bad here? Information can also live in a readme, but I dont think anything gets around the helpfulness of code comments that are literally right there when you review code

x/gamm/keeper/msg_server.go
ValarDragon: I think we make a second README / spec file with the detail, but the overall logic being here seems good

ValarDragon (approved):
Great job on improving docs here

x/gamm/keeper/msg_server.go
ValarDragon: I also don't think this violated any engineering principles, this describes how the result is computed, which is something a caller should care about

Nothing sticks out to me as a logic violation

ValarDragon: If this were a single balancer implementation.

It does stick out as abstraction violation given the pool abstraction

ValarDragon: I think we should iterate towards a better long term structure!

x/gamm/pool-models/balancer/amm.go
mattverse: Agreed

PR #1741, opened 2022-06-10
Add v10 forking logic

ValarDragon:
Closes: #XXX

This PR adds logic for the v10 fork

In BeginBlock it registers a new upgrade handler for cosmovisor to be able to hook into for future syncing nodes, and then instantiates an upgrade for the irregular state change (not yet implemented)

We update the SDK fork to add the ability to do this.

## Brief Changelog

- Adds v10 forking logic hooks

## Testing and Verifying

TODO

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #1747, opened 2022-06-10
Add v10 halting logic to v9 (cref #1741)

ValarDragon:

This PR introduces v10 halting logic to the v9 series. It makes a v9 state machine compatible version, that will halt at the upgrade height, and correctly signal cosmovisor to upgrade.

## Brief Changelog

- Add halt at v10 fork height logic

## Testing and Verifying

The test attempts running this block height, and it panics then. (as expected)

  - How is the feature or change documented? Should be

### DIFF
--- a/app/upgrades/v10/upgrades_test.go
+++ b/app/upgrades/v10/upgrades_test.go
@@ -1,0 +1,61 @@
+package v10_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/osmosis-labs/osmosis/v9/app/apptesting"
+	v10 "github.com/osmosis-labs/osmosis/v9/app/upgrades/v10"
+)
+
+type UpgradeTestSuite struct {
+	apptesting.KeeperTestHelper
+}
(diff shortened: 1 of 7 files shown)


app/upgrades/v10/upgrades_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(UpgradeTestSuite))
+}
+
+func (suite *UpgradeTestSuite) TestUpgradePayments() {
+	testCases := []struct {
+		msg         string
+		pre_update  func()
```
mattverse, line 28: Nit: Is there any reason we put the fields `pre_update` and `post_update`? If not could we delete these fields?

mattverse (approved):
LGTM, we just have to make sure the instructions are clear that this version halts the V9 at fork height

app/upgrades/v10/upgrades_test.go
ValarDragon: good call, done

Issue #1752, opened 2022-06-10
fix: Change `PowApprox` to return one when exponent is 0


mattverse:
## Background

`PowApprox` currently returns zero when the exponent is zero, whilst imo should be returning 1 when the exponent over is zero.
https://github.com/osmosis-labs/osmosis/blob/main/osmomath/math.go#L86-L88

We should change this to return `sdk.OneDec()`

## Acceptance Criteria

> Goals & criteria for success

- [ ] confirm that this `PowApprox` should be returning `1` instead of `0` in the case of having zero exponent. 
- [ ] create test cases for this specific case

ValarDragon:
oof, can you PR a fix. Lets get this backported.

PR #1755, opened 2022-06-10
fix: Change PowApprox to return one when exponent is 0 

mattverse:
Closes: #1752 

We were basically returning Zero Dec when the exponent was zero for `PowApprox` when in fact we should be returning one when any given exponent is zero.

Currently added a simple test case of testing when exponent is one but we should definitely revist osmomath as a while for better testing in a near future.

### DIFF
--- a/osmomath/math.go
+++ b/osmomath/math.go
@@ -81,7 +81,7 @@ func Pow(base sdk.Dec, exp sdk.Dec) sdk.Dec {
 }
 
 // Contract: 0 < base <= 2
-// 0 < exp < 1.
+// 0 <= exp < 1.
 func PowApprox(base sdk.Dec, exp sdk.Dec, precision sdk.Dec) sdk.Dec {
 	if exp.IsZero() {
 		return sdk.OneDec()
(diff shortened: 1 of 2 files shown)


osmomath/math.go
```diff
@@ -84,7 +84,7 @@ func Pow(base sdk.Dec, exp sdk.Dec) sdk.Dec {
 // 0 < exp < 1.
```
ValarDragon, line 84:
```suggestion
// 0 <= exp < 1.
```

ValarDragon (approved):
LGTM, thankfully under the used API this would never occur (due to line 74-76)

osmomath/math_test.go
```diff
@@ -37,6 +37,21 @@ func TestPowApprox(t *testing.T) {
...
+
+	s = PowApprox(base, exp, powPrecision)
+	expectedDec = sdk.OneDec()
+	require.NoError(t, err)
+
+	require.True(
+		t,
+		expectedDec.Sub(s).Abs().LTE(powPrecision),
+		"expected value & actual value's difference should less than precision",
+	)
```
p0mvn, line 54: Commenting to create an issue for post v10 work (not in this PR):

Make `TestPowApprox` and `TestPow` table-driven and add more test cases

p0mvn (approved):

Issue #1218, opened 2022-04-07
Fix filtered paginate in the incentives module


xBalbinus:
FilteredPaginate does pagination of all the results in the PrefixStore based on the provided PageRequest. 

We need to make the incentives module follow this

https://github.com/cosmos/cosmos-sdk/blob/master/types/query/filtered_pagination.go

Please feel free to add more details @ValarDragon

hieuvubk:
@xBalbinus 
did this issue close? I see incentives grpc_query already followed filter logic above

xBalbinus:
I don't think this issue was officially closed, if it already follows then closing is no problem

ValarDragon:
Theres a few queries in the incentives modules that just do filtered paginate incorrectly, https://github.com/osmosis-labs/osmosis/blob/main/x/incentives/keeper/grpc_query.go#L124-L132

Its getting all active gauges and filtering, inside the loop.

That loop should instead be reading off the key / value, and seeing if its valid. Rather than reading all of it.

But to be honest, most of the filtered paginate calls in that file are suspicious / wrong. (E.g. a lot that don't do any filtering!)

hieuvubk:
@xBalbinus @ValarDragon could u take a look at this pr. I think it can close this issue. Thank u so much!

xBalbinus:
A kind reminder that Bez commented something you may want to change, otherwise I defer to @ValarDragon’s judgement because I have less context

PR #1759, opened 2022-06-10
Fix incentives query filtering

hieuvubk:
Closes: #1218

Fix gauges query filter:
> handle accumulate
> no use _GetActiveGauges_ anymore in _ActiveGaugesPerDenom_ query
> no use _GetUpcomingGauges_ anymore in _UpcommingGaugesPerDenom_ query
> Add more test


x/incentives/keeper/grpc_query.go
```diff
@@ -122,10 +126,16 @@ func (q Querier) ActiveGaugesPerDenom(goCtx context.Context, req *types.ActiveGa
...
+		newGauges, err := q.getGaugeFromIDJsonBytes(ctx, value)
+		if err != nil {
+			panic(err)
+		}
+		for i := 0; i < len(newGauges); i++ {
+			if newGauges[i].DistributeTo.Denom != req.Denom {
+				return false, nil
+			}
+			if accumulate {
+				gauges = append(gauges, newGauges...)
```
alexanderbez, line 138: Shouldn't we be appending a single element, i.e. `newGauges[i]` instead of the entire slice each time?

hieuvubk: yes sure, my mistake

x/incentives/keeper/grpc_query.go
```diff
@@ -122,10 +126,16 @@ func (q Querier) ActiveGaugesPerDenom(goCtx context.Context, req *types.ActiveGa
...
 	pageRes, err := query.FilteredPaginate(valStore, req.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
-		activeGauges := q.Keeper.GetActiveGauges(ctx)
-		for _, gauge := range activeGauges {
-			if gauge.DistributeTo.Denom == req.Denom {
-				gauges = append(gauges, gauge)
+		newGauges, err := q.getGaugeFromIDJsonBytes(ctx, value)
+		if err != nil {
+			panic(err)
+		}
+		for i := 0; i < len(newGauges); i++ {
```
alexanderbez, line 133:
```suggestion
		}

		for i, g := range newGauges) {
```

ValarDragon (changes requested):
Thanks for the PR!

Since your touching this area, would you be open to de-duplicating logic across:

`Gauges`, `ActiveGauges`, `UpcomingGauges`, `ActiveGaugesByDenom`, etc?

Basically make a generic method that takes in the prefix, and an optional parameter for 'denom', and re-uses all of the logic here?

Also can you add a comment above the 
```
newGauges, err := q.getGaugeFromIDJsonBytes(ctx, value)
```
line saying to the extent of "This may return multiple gauges at once if two gauges start at the same time. For now this is treated as an edge case that is not of importance"

hieuvubk:
> Thanks for the PR!
> 
> Since your touching this area, would you be open to de-duplicating logic across:
> 
> `Gauges`, `ActiveGauges`, `UpcomingGauges`, `ActiveGaugesByDenom`, etc?
> 
> Basically make a generic method that takes in the prefix, and an optional parameter for 'denom', and re-uses all of the logic here? It would be a very welcome 100 line deletion here :)
> 
> Also can you add a comment above the
> 
> ```
> newGauges, err := q.getGaugeFromIDJsonBytes(ctx, value)
> ```
> 
> line saying to the extent of "This may return multiple gauges at once if two gauges start at the same time. For now this is treated as an edge case that is not of importance"

thanks for the suggestion. 
I moved duplicate logic to a separate function, filtering by prefix types and denom.

x/incentives/keeper/grpc_query.go
```diff
@@ -253,3 +195,33 @@ func (q Querier) getGaugeFromIDJsonBytes(ctx sdk.Context, refValue []byte) ([]ty
...
+	gauges := []types.Gauge{}
+	store := ctx.KVStore(q.Keeper.storeKey)
+	valStore := prefix.NewStore(store, prefixType)
+
+	pageRes, err := query.FilteredPaginate(valStore, pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
+		// This may return multiple gauges at once if two gauges start at the same time.
+		// For now this is treated as an edge case that is not of importance
+		newGauges, err := q.getGaugeFromIDJsonBytes(ctx, value)
+		if err != nil {
+			panic(err)
```
ValarDragon, line 210:
```suggestion
			return false, err
```

ValarDragon (approved):
Thanks, this LGTM! Can you add a bug fix changelog entry, for making gauge queries respect pagination?

Then I think this is good to go!

hieuvubk:
> Thanks, this LGTM! Can you add a bug fix changelog entry, for making gauge queries respect pagination?
> 
> Then I think this is good to go!

Yes I did

alexanderbez (approved):

ValarDragon (approved):

Issue #1733, opened 2022-06-09
Ensure multiplicative tolerance for share unit tests is as `actualShares <= expectedShares`


p0mvn:
We should also ensure that `actualShares <= expectedShares` right?

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/1731#discussion_r893878529_

mattverse:
Yeah alot of the test cases get broken when we add `	require.LessOrEqual(t, actualShares, expectedShares)`

further investigating

ValarDragon:
The existing situation should be safe. Spent awhile investigating this. First, details of the situation:
* PowApprox for exponents that != .5 has an insufficient accuracy threshold, 10^-8. This ends up translating to errors on the order of 10^-7 % - 10^-8% of total share amount in these tests.
* PowApprox is called in single asset join methods, and swaps
* This means they can all be off (in the positive direction) by 10^-7 to 10^-8
* This is fine, because in all cases they are used, the input is also charged swap fee, on the order of 10^-3. We just have to be extra careful this is the case in JoinPoolNoSwap.

My thought was then to do:
* Thus we amend the tests to special case the 0 swap fee to allow the positive error, and otherwise disallow positive error

however, that didn't work, as there is rounding error still, just you shouldn't be able to exit pool off of it to get more than the input due to the swap fee dominance. :sweat: 

best way to test this would be off the pool, shares pair on all the test vectors, or a new randomized test

AlpinYukseloglu:
Hmm even though there's still a rounding error in the swap fee case, shouldn't the swap fee eclipse it and cause those tests to pass?

Or do you mean the no swap fee case still errors?

In any case, would it be too much to implement a minimum swap fee of 1e-7 (or just make it so a 0 swap fee effectively still takes a 1e-7 fee) to prevent a pool from outputting more than it takes in in any situation?

ValarDragon:
I think we can close this, our current behavior is correct

In the swap fee case, we still could be greater than expected in the wolfram calculation test vector. But the point is that the error is tolerable due to the swap fee.

The Exit(Join(input)) < input invariant will hold due to swap fee, even with the roundup behavior.

We should increase the accuracy from 10**-7 generally though

PR #1767, opened 2022-06-11
Tests: Join pool for non-even ratio

mattverse:
Closes: #XXX

Adds tests for join pool for non even ratio assets, so that the tests hit the side where even join is done to the maximal amount able to, and then do a single pool join for the remaining coins. 

All the results used within this test codes are hand derived, attached in comment the calculations done via wolfram

### DIFF
--- a/x/gamm/pool-models/balancer/amm_test.go
+++ b/x/gamm/pool-models/balancer/amm_test.go
@@ -786,23 +786,6 @@ func TestCalcJoinPoolShares(t *testing.T) {
 				sdk.NewInt64Coin("uatom", 100_000),
 			),
 		},
-		{
-			name:    "equal weights with 0.001 swap fee",
-			swapFee: sdk.MustNewDecFromStr("0.001"),
-			poolAssets: []balancer.PoolAsset{
-				{
-					Token:  sdk.NewInt64Coin("uosmo", 1_000_000_000_000),
-					Weight: sdk.NewInt(100),
-				},
-				{
-					Token:  sdk.NewInt64Coin("uatom", 1_000_000_000_000),
-					Weight: sdk.NewInt(100),
-				},
-			},
-			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
-			expectShares: sdk.NewInt(2498749968800),
-			expectLiq:    sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
-		},
 	}
 	testCases = append(testCases, calcSingleAssetJoinTestCases...)
 
--- a/x/gamm/pool-models/balancer/amm_test.go
+++ b/x/gamm/pool-models/balancer/amm_test.go
@@ -683,6 +683,126 @@ func TestCalcJoinPoolShares(t *testing.T) {
 				sdk.NewInt64Coin("uatom", 25_000),
 			),
 		},
+		{
+			// For uosmos and uatom
+			// Expected output from Balancer paper (https://balancer.fi/whitepaper.pdf) using equation (25) on page 10:
+			// P_issued = P_supply * ((1 + (A_t * swapFeeRatio  / B_t))^W_t - 1)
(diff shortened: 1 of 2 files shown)


x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -799,6 +799,129 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+			),
+
+			expectShares: sdk.NewInt(2.5e12 + 1243750000000),
+			expectLiq: sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 25_000),
+				sdk.NewInt64Coin("uatom", 50_000),
+			),
+		},
+		{
+			// join pool is first done to the extent where the ratio can be preserved, which is 25,000 uosmo and 12,500 uatom
```
stackman27, line 885: aah so the ratio is 2usmo for 1uatom?

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -799,6 +799,129 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+			// then we perfrom single asset deposit for the remaining 37,500 uatom with the equation below
+			//
+			// For uatom:
+			// Expected output from Balancer paper (https://balancer.fi/whitepaper.pdf) using equation (25) on page 10:
+			// P_issued = P_supply * ((1 + (A_t * swapFeeRatio  / B_t))^W_t - 1)
+			// 609,374,990,000 = (100 * 10^18 + 1,250,000,000,000) *  (( 1 + (37,500 * (1 - (1 - 1/6) * 0.03) / 10,000,00,025,000))^1/6 - 1)
```
stackman27, line 891: do you happen to have the calculation for `1,250,000,000,000`?

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -683,6 +683,129 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+				sdk.NewInt64Coin("uatom", 50_000),
+			),
+		},
+		{
+			// For uosmos and uatom
+			// join pool is first done to the extent where the ratio can be preserved, which is 25,000 uosmo and 25,000 uatom
+			// then we perfrom single asset deposit for the remaining 25,000 uatom with the equation below
+			// Expected output from Balancer paper (https://balancer.fi/whitepaper.pdf) using equation (25) on page 10:
+			// P_issued = P_supply * ((1 + (A_t * swapFeeRatio  / B_t))^W_t - 1)
+			// 1_243_750_000_000 = 100 * 10^18 *  (( 1 + (25000 * (1 - (1 - 0.5) * 0.01) / 1000000025000))^0.5 - 1)
```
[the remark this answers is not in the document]
mattverse: I think this is right?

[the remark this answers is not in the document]
mattverse: your wolfram uses 12,500 as A_t whilst we should be using 25,000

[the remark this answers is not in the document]
stackman27: aah gotchu!

x/gamm/pool-models/balancer/amm_test.go
mattverse: we're not accounting weight here, the only thing that gets accounted here is token in and the amount of each pool asset

x/gamm/pool-models/balancer/amm_test.go
mattverse: yeah! just added in the commetns

stackman27: Is there an equation for minimal shares that i'm not aware of?

mattverse: yeah check `MaximalExactRatioJoin`

ValarDragon (approved):
I checked the test vectors.

I'm overall pretty worried about error tolerances and #1733 , working on verifying more things in those.

Also test fvectors in this file are kind of crazy. Need to reduce boilerplate in them or move them to their own file

mattverse:
current plan is to merge this and then create a new PR to remove all the noise in `amm_test.go`

PR #1776, opened 2022-06-11
Tests: Edge case tests for CalcJoinPoolShares

AlpinYukseloglu:

Adds three edge case tests for `CalcJoinPoolShares`:
- minimum input with two assets and minimum liquidity
- minimum input single asset equal liquidity
- minimum input single asset imbalanced liquidity

where "minimum input" means e.g. 1 uosmo


x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -833,6 +833,96 @@ func TestCalcJoinPoolShares(t *testing.T) {
...
+			tokensIn: sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 1),
+			),
+			expectShares: sdk.NewInt(50_000_000),
+			expectLiq: sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 1),
+			),
+		},
+		{
+			name:    "minimum input single asset imbalanced liquidity",
```
p0mvn, line 881: Let's move single asset to this variable:
https://github.com/osmosis-labs/osmosis/blob/bf8ef5b13b7f3a5ce58ad1c30cb5ded86d7927cf/x/gamm/pool-models/balancer/amm_test.go#L62

So that the test case is shared between multiple test cases that test similar functionality of different functions and methods

x/gamm/pool-models/balancer/amm_test.go
```diff
@@ -833,6 +833,96 @@ func TestCalcJoinPoolShares(t *testing.T) {
 			),
 			expectShares: sdk.NewInt(1250000000000 + 609374990000),
 		},
+		{
+			name:    "minimum input with two assets and minimum liquidity",
```
p0mvn, line 837: Is there a Wolfram link I could use to verify this?

x/gamm/pool-models/balancer/amm_test.go
alexanderbez: yup!

AlpinYukseloglu: Good call!

x/gamm/pool-models/balancer/amm_test.go
AlpinYukseloglu: This one doubles the liquidity in the pool from (1uosmo, 1uatom) to (2uosmo, 2uatom), so it will generate the base number of LP shares for a pool upon creation which is 1e20 (or 100 * 10^18).

I'll go ahead and add a comment to clarify but we can reach the correct final result just by using this property of Balancer pools

p0mvn (approved):
LGTM

AlpinYukseloglu:
> Something went wrong with merge conflict resolution, @AlpinYukseloglu can you investigate? Likely its that you have to move tests you added to amm_joinpool_test.go 

Looking into it!

AlpinYukseloglu:
@ValarDragon Looks like [this PR](https://github.com/osmosis-labs/osmosis/pull/1781) broke these changes. Should be fixed now!

ValarDragon (approved):

PR #1781, opened 2022-06-11
Move join pool tests into own file

ValarDragon:
Move JoinPool tests into their own file


p0mvn (approved):

p0mvn:
I will merge this. Otherwise, this will get out of sync soon

PR #1809, opened 2022-06-15
tests: fork logic for e2e tests

czarcas7ic:
Closes: #XXX

Adds ability to test forks instead of only vanilla upgrades. @p0mvn and I agreed that for now we will hardcode a height gate of 60, and later find a better way to change this value based on state (number of validators present)

## Brief Changelog

- adds isFork bool
- changes testing logic when isFork is set to true

## Testing and Verifying


app/forks.go
```diff
@@ -1,15 +1,24 @@
...
+		if !strings.Contains(ctx.ChainID(), "e2e") {
+			if ctx.BlockHeight() == fork.UpgradeHeight {
+				fork.BeginForkLogic(ctx, &app.AppKeepers)
+				return
+			}
+		} else {
+			if ctx.BlockHeight() == TestFork.UpgradeHeight {
+				TestFork.BeginForkLogic(ctx, &app.AppKeepers)
+				return
+			}
```
ValarDragon, line 21: Hrmm, I want to spend a bit more time tmrw on the space of options here, this feels unappealing. (Complicating state machine to accomodate testing hooks. Feels like theres cleaner ways to achieve what we want)

czarcas7ic: Would love advice on how to make this cleaner! Wanted something that just worked for now so we have at least some kind of test coverage since right now our CI is still testing V8

app/app.go
```diff
@@ -89,6 +90,7 @@ var (
 
 	Upgrades = []upgrades.Upgrade{v4.Upgrade, v5.Upgrade, v7.Upgrade, v9.Upgrade}
 	Forks    = []upgrades.Fork{v3.Fork, v6.Fork, v8.Fork, v10.Fork}
```
p0mvn, line 92: Then, logic in `app/forks.go` can be reverted

```suggestion
	Forks    = []upgrades.Fork{test.Fork, v3.Fork, v6.Fork, v8.Fork, v10.Fork}
```

app/forks.go
alexanderbez: I agree here with @ValarDragon -- let's discuss alternatives. Maybe we can use build tags or something.

tests/e2e/e2e_setup_test.go
```diff
@@ -185,10 +188,16 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		}
 	}
 
-	if !skipUpgrade {
+	if !skipUpgrade && !isFork {
```
alexanderbez, line 191: once you have more than one `else if`, it's best to convert to a `switch` clause for better readability

app/app.go
czarcas7ic: If I did this though, wouldn't it run at block 60 on mainnet without the app/forks.go logic?

p0mvn: yeah I did not consider the fact that there are use cases for replaying from 0. I think other discussed suggestions with build tags/genesis tweaks are more solid options than this

czarcas7ic:
Now that the genesis fix is merged to our sdk fork I was able to complete the fork upgrade logic. At this point however I need to refactor it again in order to catch up with the spike you did @p0mvn. After that I will be good for review!

tests/e2e/docker/image_config.go
```diff
@@ -28,10 +28,10 @@ const (
 	// It should be uploaded to Docker Hub. OSMOSIS_E2E_SKIP_UPGRADE should be unset
 	// for this functionality to be used.
 	previousVersionOsmoRepository = "osmolabs/osmosis-dev"
-	previousVersionOsmoTag        = "v8.0.0-2-debug"
+	previousVersionOsmoTag        = "v9.0.0-debug"
 	// Pre-upgrade repo/tag for osmosis initialization (this should be one version below upgradeVersion)
-	previousVersionInitRepository = "osmolabs/osmosis-init"
-	previousVersionInitTag        = "v8.0.0-4-osmo"
+	previousVersionInitRepository = "osmolabs/osmosis-e2e-init-chain"
+	previousVersionInitTag        = "v9.0.0"
```
p0mvn: just making sure - is the v9.x branch updated with the most recent "init" changes?

.github/workflows/test.yml
```diff
@@ -121,6 +121,17 @@ jobs:
           cache-to: type=gha,mode=max
           build-args: |
             BASE_IMG_TAG=debug
+      -
+        name: Build e2e init image
+        uses: docker/build-push-action@v3
+        with:
+          load: true
+          context: .
+          file: tests/e2e/chain_init/chain-init.Dockerfile
```
p0mvn, line 130: Why do we need to build this now?

tests/e2e/e2e_setup_test.go
```diff
@@ -228,6 +241,46 @@ func (s *IntegrationTestSuite) runValidators(chainConfig *chainConfig, dockerRep
 	s.valResources[chainConfig.meta.Id] = make([]*dockertest.Resource, len(chainConfig.validators)-len(chainConfig.skipRunValidatorIndexes))
 	pwd, err := os.Getwd()
 	s.Require().NoError(err)
+	if isFork == true {
```
p0mvn, line 229:
```suggestion
	if isFork {
```

tests/e2e/e2e_setup_test.go
```diff
@@ -185,10 +191,17 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		}
 	}
 
-	if !skipUpgrade {
+	switch {
```
p0mvn, line 189: I don't understand the conditions of this switch.

`!skipUpgrade && !isFork` -> we are supposed to do both but, in fact, only regular upgrade is run

`skipUpgrade` fork can be either skipped or unskipped -> we only run post upgrade tests without doing the fork upgrade

What's going on here?

tests/e2e/e2e_setup_test.go
```diff
@@ -228,6 +241,46 @@ func (s *IntegrationTestSuite) runValidators(chainConfig *chainConfig, dockerRep
 	s.valResources[chainConfig.meta.Id] = make([]*dockertest.Resource, len(chainConfig.validators)-len(chainConfig.skipRunValidatorIndexes))
 	pwd, err := os.Getwd()
 	s.Require().NoError(err)
+	if isFork == true {
+		for i, val := range chainConfig.validators {
+			// Skip some validators from running during set up.
```
p0mvn, line 231: I don't see any skip logic being run?

tests/e2e/e2e_setup_test.go
```diff
@@ -228,6 +241,46 @@ func (s *IntegrationTestSuite) runValidators(chainConfig *chainConfig, dockerRep
...
+			// This is needed for testing functionality like
+			// state-sunc where we might want to start some validators during tests.
+			s.T().Logf("changing %s validator genesis with index %d...", val.validator.Name, i)
+			genesis := fmt.Sprintf("%s/config/genesis.json", val.validator.ConfigDir)
+			byteValue, err := ioutil.ReadFile(genesis)
+			if err != nil {
+				panic(err)
+			}
```
p0mvn, line 239: Same with other panics
```suggestion
                        s.Require().NoError(err)
```

tests/e2e/e2e_setup_test.go
```diff
@@ -55,7 +56,13 @@ type chainConfig struct {
...
 const (
 	// osmosis version being upgraded to (folder must exist here https://github.com/osmosis-labs/osmosis/tree/main/app/upgrades)
-	upgradeVersion = "v9"
+	upgradeVersion string = "v10"
+	// is this version a fork
+	isFork bool = true
+	// if isFork is true, this is the forkHeight
+	forkHeight int = 4713065
+	// if not skipping upgrade, how many blocks we allow for fork to run pre upgrade state creation
+	forkHeightOffset int = 60
```
p0mvn, line 65: I know it's verbose but I'd prefer verbosity for clarity here. I was wondering if that's what the offset was for before seeing the comment
```suggestion
	forkHeightPreUpgradeOffset int = 60
```

tests/e2e/e2e_setup_test.go
```diff
@@ -547,12 +600,42 @@ func (s *IntegrationTestSuite) upgrade() {
 		}
 	}
 
-	// remove all containers so we can upgrade them to the new version
 	for _, chainConfig := range s.chainConfigs {
 		s.upgradeContainers(chainConfig, chainConfig.propHeight)
 	}
 }
 
+func (s *IntegrationTestSuite) upgradeFork() {
```
p0mvn, line 500: Please add to README documenting how the fork logic works, including changing `initial_height` in genesis

.github/workflows/test.yml
czarcas7ic: If we skip the upgrade shouldn't we need to initialize with the same version we run? If you are asking why we didn't do this before, I am unsure

tests/e2e/docker/image_config.go
czarcas7ic: I did, thanks for checking. We still need to work on automating this as well so we don't need to manually create this every time.

tests/e2e/e2e_setup_test.go
czarcas7ic: Accidental comment copy, thanks

tests/e2e/e2e_setup_test.go
p0mvn: Is it possible to decouple `runValidators` from any fork or upgrade logic?

From single responsibility principle, `runValidators` should be responsible for one task - running the validators. Any other setup or configuration logic should be happening separately. It feels that this coupling makes it more difficult to reason about the relationship of this configuration to how and when the upgrade and fork logic is done.

p0mvn (commented):
I know this is a draft but I thought of a few suggestions when replying to an earlier discussion. Hope you don't mind

czarcas7ic:
> I know this is a draft but I thought of a few suggestions when replying to an earlier discussion. Hope you don't mind

No problem, thank you! I first need to get past these permission issues (I think I just need to modify the fork height directly from the init function instead of after. Really annoying how it will pass locally but then will fail on GitHub 😢 ), then I will address your comments

tests/e2e/docker/image_config.go
czarcas7ic: L, I lied, it wasn't 😞 thanks for calling this out

tests/e2e/e2e_setup_test.go
```diff
@@ -290,7 +334,7 @@ func (s *IntegrationTestSuite) runValidators(chainConfig *chainConfig, portOffse
 func (s *IntegrationTestSuite) runIBCRelayer(chainA *chainConfig, chainB *chainConfig) {
 	s.T().Log("starting Hermes relayer container...")
 
-	tmpDir, err := ioutil.TempDir("", "osmosis-e2e-testnet-hermes-")
+	tmpDir, err := os.MkdirTemp("", "osmosis-e2e-testnet-hermes-")
```
p0mvn, line 337: Why? I remember seeing an issue in `cosmos-sdk` that one of these is causing trouble in CI. Is that the reason for this change?

tests/e2e/e2e_setup_test.go
```diff
@@ -368,14 +412,20 @@ func (s *IntegrationTestSuite) configureChain(chainId string, validatorConfigs [
 	// If upgrade is skipped, we can use the chain initialization logic from
 	// current branch directly. As a result, there is no need to run this
 	// via Docker.
+	if !s.isFork {
+		s.forkHeight = 0
+	}
```
p0mvn, line 417: `s.forkHeight` should be 0 by default and set to something else if fork is enabled when parsing environment variables
```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -368,14 +412,20 @@ func (s *IntegrationTestSuite) configureChain(chainId string, validatorConfigs [
...
-		initializedChain, err := chain.Init(chainId, tmpDir, validatorConfigs, time.Duration(newChainConfig.votingPeriod))
+		initializedChain, err := chain.Init(chainId, tmpDir, validatorConfigs, time.Duration(newChainConfig.votingPeriod), s.forkHeight)
 		s.Require().NoError(err)
 		s.initializeChainConfig(&newChainConfig, initializedChain)
 		return
 	}
 
-	initResource, err := s.containerManager.RunChainInitResource(chainId, int(newChainConfig.votingPeriod), validatorConfigBytes, tmpDir)
+	s.forkHeight = s.forkHeight - forkHeightPreUpgradeOffset
+	fmt.Printf("FORK HEIGHT %v", s.forkHeight)
```
p0mvn, line 427: let's use a proper log if this is useful or remove if not

Makefile
```diff
@@ -237,7 +237,7 @@ test-sim:
 	@VERSION=$(VERSION) go test -mod=readonly $(PACKAGES_SIM)
 
 test-e2e:
-	@VERSION=$(VERSION) go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
```
p0mvn, line 240: Wdyt about only having `OSMOSIS_E2E_FORK_HEIGHT`?

If it is set, then we can set `s.isFork` to true in `e2e_setup_test.go`. If not, we can keep it as default false.

The boolean seems redundant because having a height is idiomatic with the True value of `OSMOSIS_E2E_IS_FORK`

tests/e2e/containers/config.go
```diff
@@ -40,22 +40,25 @@ const (
...
 // Otherwise, returns images for running non-upgrade e2e tests.
-func NewImageConfig(isUpgrade bool) ImageConfig {
+func NewImageConfig(isUpgrade, isFork bool) ImageConfig {
 	config := ImageConfig{
 		RelayerRepository: relayerRepository,
 		RelayerTag:        relayerTag,
 	}
 
-	if isUpgrade {
+	if isUpgrade && !isFork {
```
p0mvn, line 49: Was playing with this and adding comments to make sure I understand. Feel free to adapt as you see fit.

```
    if !isUpgrade {
	// If upgrade is not tested, we do not need InitRepository and InitTag
	// because we directly call the initialization logic without
	// the need for Docker.
        config.OsmosisRepository = CurrentBranchOsmoRepository
	config.OsmosisTag = CurrentBranchOsmoTag
        return isUpgrade
    }

    // If upgrade is tested, we need to utilize InitRepository and InitTag
    // to initialize older state with Docker
    config.InitRepository = previousVersionInitRepository
    config.InitTag = previousVersionInitTag

    if isFork {
	// Forks are state compatible with earlier versions before fork height.
	// Normally, validators switch the binaries pre-fork height
	// Then, once the fork height is reached, the state breaking-logic
	// is run.
	config.OsmosisRepository = CurrentBranchOsmoRepository
	config.OsmosisTag = CurrentBranchOsmoTag
    } else {
	// Upgrades are run at the time when upgrade height is reached
	// and are submitted via a governance proposal. Thefore, we
	// must start running the previous Osmosis version. Then, the node
        // should auto-upgrade, at which point we can restart the updated
        // Osmosis validator container.
        config.OsmosisRepository = previousVersionOsmoRepository
	config.OsmosisTag = previousVersionOsmoTag
    }
```

tests/e2e/containers/config.go
```diff
@@ -40,22 +40,25 @@ const (
 // Returns ImageConfig needed for running e2e test.
 // If isUpgrade is true, returns images for running the upgrade
```
p0mvn, line 41: godoc needs to be updated

tests/e2e/chain_init/main.go
```diff
@@ -39,7 +41,7 @@ func main() {
 		panic(err)
 	}
 
-	createdChain, err := chain.Init(chainId, dataDir, valConfig, votingPeriod)
+	createdChain, err := chain.Init(chainId, dataDir, valConfig, votingPeriod, forkHeight)
```
p0mvn, line 44: Assuming that such logic exists, If I were to spin up a single node for state-sync when a chain is already running, what should I be specifying for `forkHeight`?

tests/e2e/e2e_setup_test.go
```diff
@@ -210,15 +247,22 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
+	case !s.skipUpgrade && !s.isFork:
 		s.createPreUpgradeState()
 		s.upgrade()
 		s.runPostUpgradeTests()
+	case !s.skipUpgrade && s.isFork:
+		s.createPreUpgradeState()
+		s.upgradeFork()
+		s.runPostUpgradeTests()
+	case s.skipUpgrade:
+		s.runPostUpgradeTests()
```
p0mvn, line 260: `s.runPostUpgradeTests()` should be in its own fixture `func TestPostUpgrade(t *testing.T)` in `e2e_test.go`. We can do this in a future PR

p0mvn (commented):
Good job, this looks great. Had a couple more minor comments and questions

tests/e2e/e2e_setup_test.go
czarcas7ic: As of go 1.17, ioutil.TempDir just calls os.MkdirTemp. The change was made originally to see if it would fix the permissions issue I was having (it didn't and I solved it another way), but I didn't revert this change because it seems like we might as well call the function directly instead of inadvertently.

tests/e2e/e2e_setup_test.go
czarcas7ic: Oof I left this behind from debugging, thanks

tests/e2e/chain_init/main.go
p0mvn: Makes sense, thank you

tests/e2e/README.md
```diff
@@ -11,9 +11,15 @@ via Docker files. As a result, the test suite may provide the desired
...
-performing the chain upgrade, and running the latest test suite.
-
-The file e2e\_suite\_test.go defines the testing suite and contains the
+performing the chain upgrade, and running the latest test suite. When
+testing a normal upgrade, the e2e test suite submits an upgrade proposal at
+an upgrade height, ensures the upgrade happens at the desired height, and
+then checks that operations that worked before still work as intended. If
+testing a fork, the test suite instead starts the chain a few blocks before
+the set fork height and ensures the chain continues after the fork triggers
+the upgrade.
```
p0mvn, line 20:
```suggestion
the upgrade. Note that a regular upgrade and a fork upgrade are mutually exclusive. 
```

tests/e2e/README.md
```diff
@@ -111,6 +117,13 @@ tests require IBC logic.
 - `OSMOSIS_E2E_SKIP_CLEANUP` - when true, avoids cleaning up the e2e Docker
 containers.
 
+- `OSMOSIS_E2E_IS_FORK` - when true, utilizes fork logic over traditional upgrade logic
+
```
p0mvn, line 121:
```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -210,15 +237,22 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 	}
 }
```
p0mvn, line 252:
```suggestion
if !s.skipUpgrade {
	s.createPreUpgradeState()

	if s.isFork {
		s.upgradeFork()
	} else {
		s.upgrade()
	}
}

s.runPostUpgradeTests()
```
```suggestion
	switch {
	case !s.skipUpgrade && !s.isFork:
		s.createPreUpgradeState()
		s.upgrade()
		s.runPostUpgradeTests()
	case !s.skipUpgrade && s.isFork:
		s.createPreUpgradeState()
		s.upgradeFork()
		s.runPostUpgradeTests()
	case s.skipUpgrade:
		s.runPostUpgradeTests()
	}
}
```

tests/e2e/e2e_setup_test.go
```diff
@@ -484,12 +520,46 @@ func (s *IntegrationTestSuite) upgrade() {
...
+				func() bool {
+					currentHeight := s.getCurrentChainHeight(chainConfig, i)
+					if currentHeight < s.forkHeight {
+						s.T().Logf("current block height on %s is %v, waiting for block %v container: %s", containerName, currentHeight, s.forkHeight, containerId)
+						return false
+					}
+					if currentHeight > s.forkHeight {
+						return true
+					}
+					return true
```
p0mvn, line 553: Seems that this is true either way
```suggestion
					return true
```

tests/e2e/README.md
p0mvn: I think it was removed as per: https://github.com/osmosis-labs/osmosis/pull/1809/files#r905583441

tests/e2e/e2e_setup_test.go
p0mvn: Depends on the case. 

With only a few conditionals, the choice between if and switch depends on readability.

Here, we are repeating a lot of the same code - `s.createPreUpgradeState()` and `s.runPostUpgradeTests()`. Additionally, cases are not as simple with 2 conditions, making this logic harder to grok.

If there are many conditions, switch is preferred because it is more performant: https://stackoverflow.com/questions/767821/is-else-if-faster-than-switch-case

tests/e2e/README.md
p0mvn: The link in my earlier comment was broken, updated now. Basically, the presence of `OSMOSIS_E2E_FORK_HEIGHT signifies whether this is a fork upgrade now

tests/e2e/e2e_setup_test.go
p0mvn: The biggest reason for my comment here is that the last case (`case s.skipUpgrade`) is redundant because `s.runPostUpgradeTests()` can be applied to all cases.

Having the following also works imo:
```
	switch {
	case !s.skipUpgrade && !s.isFork:
		s.createPreUpgradeState()
		s.upgrade()
	case !s.skipUpgrade && s.isFork:
		s.createPreUpgradeState()
		s.upgradeFork()
        }
        s.runPostUpgradeTests()
```

czarcas7ic:
Once this gets merged, going to follow this up with already completed PR for switching to v10 upgrade to v11 logic instead of v8 to v10 which we are doing now (also fixes a couple of upgrade logic bugs that got introduced from this PR)

tests/e2e/e2e_setup_test.go
```diff
@@ -212,13 +239,20 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
+
+		if s.isFork {
+			s.upgradeFork()
+		} else {
+			s.upgrade()
+		}
 	}
+
+	s.runPostUpgradeTests()
+
```
p0mvn, line 251:
```suggestion
```

tests/e2e/e2e_setup_test.go
```diff
@@ -184,7 +203,15 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		}
 	}
 
-	s.containerManager, err = containers.NewManager(!s.skipUpgrade)
+	if str := os.Getenv(upgradeVersionEnv); len(str) > 0 {
+		s.upgradeVersion = str
+
+		if s.upgradeVersion != "" {
+			s.T().Log(fmt.Sprintf("upgrade version set to %s", s.upgradeVersion))
+		}
```
p0mvn, line 211: It is alsways non-empty because we check `len(str) > 0` above

```suggestion
		s.T().Log(fmt.Sprintf("upgrade version set to %s", s.upgradeVersion))
```

tests/e2e/e2e_setup_test.go
```diff
@@ -171,6 +179,17 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 
+	if str := os.Getenv(forkHeightEnv); len(str) > 0 {
+		forkHeight64, err = strconv.ParseInt(str, 0, 64)
+		s.forkHeight = int(forkHeight64)
+		s.Require().NoError(err)
+		s.isFork = true
+
+		if s.forkHeight != 0 {
+			s.T().Log(fmt.Sprintf("%s was set to height %v", forkHeightEnv, s.forkHeight))
+		}
```
p0mvn, line 190: I think we should always print the height since someone might provide 0 as the value of the `forkHeightEnv`
```suggestion
		s.T().Log(fmt.Sprintf("fork upgrade is enabled, %s was set to height %v", forkHeightEnv, s.forkHeight))
```

tests/e2e/e2e_setup_test.go
```diff
@@ -171,6 +179,17 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		}
 	}
 
+	if str := os.Getenv(forkHeightEnv); len(str) > 0 {
+		forkHeight64, err = strconv.ParseInt(str, 0, 64)
+		s.forkHeight = int(forkHeight64)
+		s.Require().NoError(err)
```
p0mvn, line 185: nit:
```suggestion
		s.Require().NoError(err)
		s.forkHeight = int(forkHeight64)
```

p0mvn (approved):
LGTM

Issue #1773, opened 2022-06-11
Create TestCalcJoinPoolShares test case to fall outside of the error tolerance


p0mvn:
> > I'm confused as to why expectShares is double what is in the previous test when the total liquidity isn't being doubled
> 
> The logic was the following - here we are adding 2 tokens in that have equal weights and amounts in, so the liquidity would double. In the previous case, there was only one token in.
> 
> However, it seems that these should still not be equal because after the first join we would have more shares created for one of the tokens, right? As a result, the second join would have fewer shares returned than the first one. Seems that this test is passing because of the error tolerance

This is interesting – I wonder if we can set up a case where this falls outside of error tolerance just for proper coverage? Or would it always pass due to error tolerance?

_Originally posted by @AlpinYukseloglu in https://github.com/osmosis-labs/osmosis/pull/1721#discussion_r894167070_

p0mvn:
We need a test case to fall outside of the error tolerance for when we are testing multi token calcs by joining one after another in `calcJoinSingleAssetTokensIn`. 

We should be able to achieve that by having large token in values relative to the pool liquidity

stackman27:
Hi @p0mvn can i give this a shot?

p0mvn:
Hey @stackman27 . Yes, please, that would be awesome

PR #1822, opened 2022-06-19
[Test] Create TestCalcJoinSingleAssetTokensIn test case to fall outside of the error tolerance

stackman27:
Closes: #1773

Check if the `tokenshares` fall outside of the error tolerance

- [x] fix err tolerance check function to accommodate  error 
- [x] test for single token in that falls outside error tolerance 
- [x]  test for multi tokens in that falls outside the error tolerance
- [x]  Add tolerance check for `TestCalcJoinPoolShares` 

## Brief Changelog
n/a

## Testing and Verifying
Add extra test to `TestCalcJoinSingleAssetTokensIn`


x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -820,11 +854,12 @@ func TestCalcJoinSingleAssetTokensIn(t *testing.T) {
...
-			name:         "one token in - equal weights with zero swap fee",
-			swapFee:      sdk.MustNewDecFromStr("0"),
-			poolAssets:   oneTrillionEvenPoolAssets,
-			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
-			expectShares: sdk.NewInt(2_499_999_968_750),
+			name:                 "one token in - equal weights with zero swap fee",
+			swapFee:              sdk.MustNewDecFromStr("0"),
+			poolAssets:           oneTrillionEvenPoolAssets,
+			tokensIn:             sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
+			expectShares:         sdk.NewInt(2_499_999_968_750), // ACTUAL: 249_999_996_8800
```
mattverse, line 861:
```suggestion
			expectShares:         sdk.NewInt(2_499_999_968_750), // ACTUAL: 249_999_996_880
```

x/gamm/pool-models/balancer/util_test.go
```diff
@@ -39,19 +39,19 @@ func createTestContext(t *testing.T) sdk.Context {
...
 		MultiplicativeTolerance: allowedErrRatioDec,
 	}
 
-	require.Equal(
-		t,
-		0,
-		errTolerance.Compare(expectedShares, actualShares),
-		fmt.Sprintf("expectedShares: %s, actualShares: %s", expectedShares.String(), actualShares.String()))
+	if errTolerance.Compare(expectedShares, actualShares) != 0 {
+		return errors.New("shares doesnot fall within the tolerance range")
```
mattverse, line 51:
```suggestion
		return errors.New("shares does not fall within the tolerance range")
```

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -939,24 +977,51 @@ func TestCalcJoinSingleAssetTokensIn(t *testing.T) {
...
+		{
+			// |a - b| / min(a, b) <= allowedErrRatio
+			// a = expected shares, b = actual shares
+			// 0.04000000047920000 = abs(2_599_999_968_750-2_499_999_968_800)/2_499_999_968_800
+			// 	Full solution: https://www.wolframalpha.com/input?i=abs%282599999968750-2499999968800%29%2F2499999968800
+			name: "Error Tolerance: One token in, expected > actual",
+			swapFee:              sdk.MustNewDecFromStr("0"),
+			poolAssets:           oneTrillionEvenPoolAssets,
+			tokensIn:             sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
+			expectShares:         sdk.NewInt(2_599_999_968_750),// ACTUAL: 2_499_999_968_800
```
mattverse, line 993: wdyt about adding comments for the calculations on why these are outside the range of balancer module's `allowedErrRatio`?

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -80,6 +81,7 @@ var calcSingleAssetJoinTestCases = []calcJoinSharesTestCase{
 		poolAssets:   oneTrillionEvenPoolAssets,
 		tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
 		expectShares: sdk.NewInt(2_499_999_968_750),
+		expectToleranceError: false,
```
p0mvn, line 84: No need to specify `expectToleranceError: false,` since non-initialized bool is equivalent to false in Go

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -101,6 +103,7 @@ var calcSingleAssetJoinTestCases = []calcJoinSharesTestCase{
 		poolAssets:   oneTrillionEvenPoolAssets,
 		tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000)),
 		expectShares: sdk.NewInt(2_487_500_000_000),
+		expectToleranceError: false,
```
p0mvn, line 106: ditto and for all other instances of `expectToleranceError: false,`

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -603,8 +626,13 @@ func (suite *KeeperTestSuite) TestCalcJoinPoolShares() {
 					require.Equal(t, sdk.NewCoins(), liquidity)
 				} else {
 					require.NoError(t, err)
-					assertExpectedSharesErrRatio(t, tc.expectShares, shares)
 					assertExpectedLiquidity(t, tc.tokensIn, liquidity)
+					err = assertExpectedSharesErrRatio(t, tc.expectShares, shares)
+					if tc.expectToleranceError {
```
p0mvn, line 631: I don't see a single test case where `expectToleranceError` is true, do we need to have this conditional?

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -786,7 +814,12 @@ func TestCalcSingleAssetJoin(t *testing.T) {
 				}
 
 				require.NoError(t, err)
-				assertExpectedSharesErrRatio(t, tc.expectShares, shares)
+				err = assertExpectedSharesErrRatio(t, tc.expectShares, shares)
+				if tc.expectToleranceError {
```
p0mvn, line 818: I don't see a single test case where `expectToleranceError` is true, do we need to have this conditional?

x/gamm/pool-models/balancer/amm_joinpool_test.go
```diff
@@ -798,13 +831,14 @@ func TestCalcSingleAssetJoin(t *testing.T) {
...
 func TestCalcJoinSingleAssetTokensIn(t *testing.T) {
 	testCases := []struct {
-		name           string
-		swapFee        sdk.Dec
-		poolAssets     []balancer.PoolAsset
-		tokensIn       sdk.Coins
-		expectShares   sdk.Int
-		expectLiqudity sdk.Coins
-		expErr         error
+		name                 string
```
p0mvn, line 834: can we make this be `calcJoinSharesTestCase`? Seems that most of the fields are the same

p0mvn (commented):
Nice work so far. Haven't checked the calculations in detail yet but had some comments. Please take a look

x/gamm/pool-models/balancer/amm_joinpool_test.go
stackman27: i have added two tests so far; 
1. https://github.com/osmosis-labs/osmosis/pull/1822/files#diff-d90d7fb125be0247c48c167de4adedf2ddbdc76b1ad06bc243d329bc8dbff76dR985
2. https://github.com/osmosis-labs/osmosis/pull/1822/files#diff-d90d7fb125be0247c48c167de4adedf2ddbdc76b1ad06bc243d329bc8dbff76dR997

x/gamm/pool-models/balancer/amm_joinpool_test.go
stackman27: I can do that

x/gamm/pool-models/balancer/amm_joinpool_test.go
p0mvn: But these are defined in a different test fixture - `TestCalcJoinSingleAssetTokensIn`, are they not?

The test cases for this particular fixture (`TestCalcJoinPoolShares`) seem to be unmodified

stackman27: aaah i see! I thought i was only creating tolerance check tests for `TestCalcJoinSingleAssetTokensIn` function. I can create tests for `TestCalcJoinPoolShares` too

stackman27:
@p0mvn @mattverse Thank you for the review! =) 

Addressed all your comments [6d69ca4](https://github.com/osmosis-labs/osmosis/pull/1822/commits/6d69ca4c566202f657de54bd7bdf155dce1caec0)

mattverse (approved):
Nice work on the vecotrs here! LGTM!

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -354,6 +351,42 @@ func TestCalcJoinSingleAssetTokensIn(t *testing.T) {
 			tokensIn:     sdk.NewCoins(sdk.NewInt64Coin("uosmo", 50_000), sdk.NewInt64Coin("uatom", 100_000)),
 			expectShares: sdk.NewInt(2_072_912_400_000_000 + 1_624_999_900_000),
 		},
+		{
+			// Compares expected shares and actual shares to make sure the multiplicativeTolerance ratio below is maintained
+			// |a - b| / min(a, b) <= allowedErrRatio
+			// 0.04000000047920000 = abs(2_599_999_968_750-2_499_999_968_800)/2_499_999_968_800
```
p0mvn, line 357: Identifying the test cases is an important step toward fixing these issues. You've done a great job at this! Now, we should understand the sources of these errors so that we can fix them.

On that note, I'm wondering: 

- How did you choose these values so that they are outside of the error tolerance? In other words, for each test case, what choice of inputs causes the delta to be large? 

Could you please document this directly in the comments? I think it would be incredibly helpful so that we can come back to it later and know what causes these issues.

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -51,14 +51,15 @@ var (
...
-	expErr       error
+	name                 string
+	swapFee              sdk.Dec
+	poolAssets           []balancer.PoolAsset
+	tokensIn             sdk.Coins
+	expectShares         sdk.Int
+	expectLiq            sdk.Coins
+	expectPanic          bool
+	expErr               error
+	expectToleranceError bool
```
p0mvn, line 62:
```suggestion
        // There are known cases where the actual shares might fall outside of the
        // tolerable range from the expected shares. Each such test case documents
        // the reason for the fallout. 
	expectToleranceError bool
```

p0mvn (commented):
Great work! I'm currently not sure what inputs cause the delta to be large. Can we document that for each test case, please?

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -650,6 +654,56 @@ func (suite *KeeperTestSuite) TestCalcJoinPoolShares() {
...
+			// allowedErrRatio = 0.0000001
+			// 	Full solution: https://www.wolframalpha.com/input?i=abs%286243749992287-3743749992293%29%2F3743749992293
+			name:       "Error tolerance: Multi token with unequal amounts, with equal weights and 0.01 swap fee",
+			swapFee:    sdk.MustNewDecFromStr("0.01"),
+			poolAssets: oneTrillionEvenPoolAssets,
+			tokensIn: sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 50_000),
+				sdk.NewInt64Coin("uatom", 75_000),
+			),
+			expectShares:         sdk.NewInt(2.5e12 + 1350000000000), // ACTUAL: 6_243_749_992_287
```
p0mvn, line 676: Also, is there a Wolfram calculation for `expectShares`? How was this value selected?

stackman27: I thought about this previously as well, but if i calculate expShares(through our logic)  and insert the right value, it wouldn't cause tolerance err right? So i purposely tried to mess around the numbers

p0mvn: So our goal is to find out if it is possible to find test cases with valid inputs / settings that cause us to fall outside of the error tolerance.

That would mean having a valid calculation for `expectShares` , similar to previous test cases that, at the same time, fall outside of the error tolerance.

How can we claim that calculated `shares` fall outside of the error tolerance from `expectShares` when the `expectShares are randomly selected, right?

So, for this to be accepted, I would like to know how `expectShares` are estimated. Preferably, with a Wolfram link similar to previous test cases.

stackman27: hmm i tried lots of scenarios and i wasn't able to generate scenarios to get values that fall outside the error tolerance. In face the larger the `tokensIn` the better the bound became. Maybe we might wanna make the error tolerance higher?  

This is the largest value i tried; 
```
 Desc: one token in - equal weights with zero swap fee
 P_issued = P_supply * ((1 + (A_t / B_t))^W_t - 1)
 4_880_884_817_015_154_699 =  1e20 * (( 1 + (100000000000 / 1e12))^0.5 - 1) 

|a - b| / min(a, b) <= allowedErrRatio
0.000000000000000000204 = abs(4_880_884_817_015_154_699-4_880_884_817_015_154_700)/4_880_884_817_015_154_699

0.000000000000000000204 <= 0.0000001 (WITHIN ACCEPTED RATIO)
```

stackman27: cc @p0mvn

p0mvn: That makes sense, thanks for looking deep into this.

If that's the case, let's do the following:
- add any test cases that you've tried and think make sense to add
- make comments consistent with the rest of the test cases
- remove prints and `expectToleranceError bool`

Then, we can close this :+1:  The conclusion is that we could not find a case to fall outside of the tolerance

Issue #1139, opened 2022-03-24
Make swapfees count towards required minimum tx fee on Osmosis


ValarDragon:
With the take that tx fees are purely for Spam resistance, and not validator profit, swap fee and capital lockup txs can count towards minimum tx fees.

This issue is around swap fees.

We should make swapfees count towards the minimum tx fees per unit has that a node has, for swaps on pools accepted by 'txfees' module.

ValarDragon:
Some notes from a call I had on the feature: https://hackmd.io/p_ERhcv5SeSN4yPjg5-lSg

RusAkh:
hello @ValarDragon is this issue still open to work on?

PR #1675, opened 2022-06-06
feat: Swap Fees used For Sybil Resistant Tx Fees  

lostak:
Closes: #1139
New PR: #1823 

## What is the purpose of the change

Swap fees paid for a msg are added to Tx fee when determining if the fee is sufficient for the gas price
- If the tx fee attached for a swap message is not sufficient to pay for gas, the swap fees paid can make up the difference
- The attached tx fee is always deducted in its entirety if it exists
- If the gas price is zero, no tx fee needs to be attached
- If the message satisfies the SybilResistantFee interface, add the swap fees paid to the sybil struct
- For all tx, add tx fee to sybil fees paid (for non-applicable msgs, feed paid = 0)
- Compare sybil fees paid to gas price when determining if a fee is sufficient

## Brief Changelog

- Add `GetSwapFee(ctx sdk.Context, poolId uint64) (sdk.Dec, error)` to `swap.go` and to the `GammKeeper` in the `expected_keepers` for the `txfees` module
- Add interface `SybilResistantFee` in a new file `msg_sybil.go` in `gamm/types/` 
- Add implement `SybilResistantFee` in all msgs involving a swap, including multihops
- Add file `sybil.go` for sybil resistant fees struct and methods
- `AnteHandle` now treats `SybilResistantFee` msgs differently when checking a fee but other msgs use the same logic as before

## Testing and Verifying

`make test`, `make test-sim` and `make build` are all successful but more tests are needed

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (not documented)

PR #1823, opened 2022-06-19
feat: Swap Fees Paid Contribute to Tx Fee for Sybil Resistant Fees

lostak:
Closes: #1139   
[Previous PR](https://github.com/osmosis-labs/osmosis/pull/1675)

Swap fees paid for a msg are added to Tx fee when determining if the fee is sufficient for the gas price
- If the tx fee attached for a swap message is not sufficient to pay for gas, the swap fees paid can make up the difference
- The attached tx fee is always deducted in its entirety if it exists
- If the gas price is zero, no tx fee needs to be attached
- If the message satisfies the SybilResistantFee interface, add the swap fees paid to the sybil struct
- For all tx, add tx fee to sybil fees paid (for non-applicable msgs, feed paid = 0)
- Compare sybil fees paid to gas price when determining if a fee is sufficient

## Brief Changelog

- Add `GetSwapFee(ctx sdk.Context, poolId uint64) (sdk.Dec, error)` to `swap.go` and to the `GammKeeper` in the `expected_keepers` for the `txfees` module
- Add interface `SybilResistantFee` in a new file `msg_sybil.go` in `gamm/types/` 
- Add implement `SybilResistantFee` in all msgs involving a swap, including multihops
- Add file `sybil.go` for sybil resistant fees struct and methods
- `AnteHandle` now treats `SybilResistantFee` msgs differently when checking a fee but other msgs use the same logic as before

## Testing and Verifying

`make test`, `make test-sim`, `make test-unit` and `make build` are all successful and specific tests are being written

### DIFF
--- a/x/gamm/types/msg_sybil.go
+++ b/x/gamm/types/msg_sybil.go
@@ -1,0 +1,97 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// SybilResistantFee Swap Msg defines an interface for getting
+// - poolIds on a swap msg route
+// - token to apply the sybil resistant fees to
+type SybilResistantFee interface {
+	GetTokenDenomsOnPath() []string
+	GetPoolIdOnPath() []uint64
+	GetTokenToFee() sdk.Coin
+}
+
+var (
+	_ SybilResistantFee = MsgSwapExactAmountOut{}
+	_ SybilResistantFee = MsgSwapExactAmountIn{}
+)
+
+// MsgSwapExactAmountOut implements SybilResistantFee
+func (msg MsgSwapExactAmountOut) GetTokenDenomsOnPath() []string {
+	if msg.Routes == nil {
+		return []string{}
+	}
+
+	if len(msg.Routes) == 0 {
+		return []string{}
+	}
+
+	denoms := make([]string, 0, len(msg.Routes)+1)
+	for i := 0; i < len(msg.Routes); i++ {
+		denoms = append(denoms, msg.Routes[i].TokenInDenom)
+	}
+
+	denoms = append(denoms, msg.TokenOutDenom())
+
+	return denoms
+}
+
+func (msg MsgSwapExactAmountOut) GetTokenToFee() sdk.Coin {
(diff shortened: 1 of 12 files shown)


lostak:
[Diagram Comparing Proposed Changes in Logic and Files to the Current Fee Decorator Process](https://github.com/osmosis-labs/osmosis/files/8934664/SybilResistantFeeDiagram.pdf)

x/gamm/types/msg_sybil.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// SybilResistantFee Swap Msg defines an interface for getting
+// - poolIds on a swap msg route
+// - token to apply the sybil resistant fees to
+type SybilResistantFee interface {
+	GetTokenDenomsOnPath() []string
+	GetPoolIdOnPath() []uint64
+	GetTokenToFee() sdk.Coin
+}
```
ValarDragon, line 14: I think we should just add the `GetPoolIdOnPath` to the existing `msg_swap.go` interface. Not sure why "GetTokenToFee" exists

lostak: By using a separate interface, `SybilResistantFee` can be more abstractly applied to any message involving a swap, whereas `SwapMsgRoute` is only satisfied by messages with swap routes in their proto file.
[These lines](https://github.com/osmosis-labs/osmosis/blob/d7fa2fd172668b4a0eb7d8207b3e2d0c952d43b6/x/txfees/keeper/feedecorator.go#L167-L185) of `feedecorator.go` use this abstraction and could work for msgs like `MsgJoinSwapExternAmountIn` properly satisfying `SybilResistantFee`. 
`GetTokenToFee()` returns the `sdk.Coin` that would have paid the swap fees, in its initial amount. This is then used to check if the token is applicable for use as a `feeToken`, and to compute the fees paid in the base denom.
If the swap is a `multihop`, then only swap fees from pools with a `feeToken` are added to the total swap fee.

lostak: If you only see this as being used for swaps, not joins and exits, I'll move the new functionality to the existing interface.

PR #1852, opened 2022-06-22
Mergify: Make backports automatically assign PR author

ValarDragon:
This change has been made by @ValarDragon from https://mergify.com config editor.


ValarDragon:
Going to backport this just to sanity check it works, even though mergify says the config is valid

PR #1853, opened 2022-06-22
feat: mutation testing setup

alexanderbez:
First stab at [mutation testing](https://en.wikipedia.org/wiki/Mutation_testing) using the [`go-mutesting`](https://github.com/zimmski/go-mutesting) tool/binary.

A quick summary of mutation testing and the output the binary produces:

- Mutation testing attempts to modify a program or binary, which are called mutants, and executes existing tests against those mutants.
- If tests fail, the mutant is considered "killed".
- Having mutants killed is a _good_ thing, which means the software properly tests the various possible execution paths and should lead to strong confidence levels in the program's correctness.
- The `go-mutesting` produces an output such as the following: `The mutation score is 0.750000 (6 passed, 2 failed, 0 skipped, total is 8)`. This means 6 mutants were killed yielding a score or confidence level of 75%, which is not terrible. The higher the score the better.
- False positives are probably going to be a common, i.e. a mutant that isn't killed but should be ignored. The most common example is "early exits" in a function/method. These are going to be pretty common I imagine and need to be added to a "blacklist" file that we will maintain as times goes on.

This PR sets up the scaffolding in CI for mutation testing and uses various parts of x/tokenfactory as the starting point. Subsequent PRs should be made which add more and more files/packages to be executed. Note, if the confidence score is low, that means our tests are not sufficient and should be improved prior to merging said PR.


mutation.blacklist
```diff
@@ -0,0 +1,17 @@
+5e97df8a85d3b16dbc06853634d1098d
```
ValarDragon, line 1: I'm a bit worried about this being committed and a source of many merge conflicts

.vscode/settings.json
```diff
@@ -23,5 +23,25 @@
...
+    "statusBar.background": "#2d2755",
+    "statusBar.foreground": "#e7e7e7",
+    "statusBarItem.hoverBackground": "#403778",
+    "statusBarItem.remoteBackground": "#2d2755",
+    "statusBarItem.remoteForeground": "#e7e7e7",
+    "titleBar.activeBackground": "#2d2755",
+    "titleBar.activeForeground": "#e7e7e7",
+    "titleBar.inactiveBackground": "#2d275599",
+    "titleBar.inactiveForeground": "#e7e7e799"
+  },
```
ValarDragon, line 45: Is there some sort of conflict happening with these colors? (Feels a bit odd to me to commit color preferences in the vscode config :sweat_smile: )

scripts/mutation-test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+DISABLED_MUTATORS='branch/*'
+
+# Only consider the following:
+# * go files in types or keeper packages
+# * ignore test and Protobuf files
+MUTATION_SOURCES=$(find ./x -type f \( -path '*/keeper/*' -or -path '*/types/*' \) \( -name '*.go' -and -not -name '*_test.go' -and -not -name '*pb*' \))
```
ValarDragon, line 10: This is some interestintg file specification syntax lol

Makefile
```diff
@@ -237,6 +238,9 @@ test-e2e:
 test-e2e-skip-upgrade:
 	@VERSION=$(VERSION) OSMOSIS_E2E_SKIP_UPGRADE=True go test -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
 
+test-mutation:
+	@go run github.com/zimmski/go-mutesting/cmd/go-mutesting --blacklist=mutation.blacklist $(MUTATION_SOURCES)
```
ValarDragon, line 242: The script you made seems way better

ValarDragon:
This is awesome! Ran the script for the token factory module, and this seems like qutie valid testing gaps found!

(That we currently aren't verifying codec changes, which even led to concerns on PRs #1972 #1994 )

Only errror type I;'m struggling to understand right now is:
```
-
-	_, err = GetTokenDenom(m.Sender, m.Subdenom)
+	_, _, _ = err, m.Sender, m.Subdenom
 	if err != nil {
 		return sdkerrors.Wrap(ErrInvalidDenom, err.Error())
 	}
```

wdyt the steps here should be?  (Beyond understanding ^ and if its a false positive)

I feel like going ahead and merging this, and then maybe making CI run this on a schedule (perhaps once a week at first?) for a couple modules, and generate a list of failures into a github issue we tackle? (If details of that are hard, we can do so manually first couple of times)

scripts/mutation-test.sh
alexanderbez: lol this took me ages to figure out FYI

.vscode/settings.json
alexanderbez: I wanted to commit these for use with Peacock but oh well, reverted :(

czarcas7ic:
@ValarDragon simulation is failing in CI (and locally) every now and then when the simulator decides to use the same lock ID twice. I can look into it, but for now if we rerun this should clear

ValarDragon (approved):

ValarDragon:
Really excited about this!

Issue #1753, opened 2022-06-10
refactor: separate methods for single / multi token calculation in `CalcJoinPoolShares`


mattverse:
## Background

Currently `CalcJoinPoolShares` handles two different methods within the methods:
1. calculating join pool shares for single token in 
2. calculating join pool shares for multiple tokens(namely all tokens within the pool).

Currently these two different roles of calculating is separated by a single logic gate of `if` statement, we should create two separate methods to call within the method `CalcJoinPoolShares`

cref: https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/pool-models/balancer/amm.go#L273 

## Suggested Design

- Create two different sub-methods for the two different cases for calculating join pool shares mentioned above(something like `calcSingleTokenJoinPoolShare` and `calcMultipleTokenPoolShare`)
- call these two methods within `CalcJoinPoolShares` based on token length.

## Acceptance Criteria

- [ ] Separate the two method, refactor so that `CalcJoinPoolShares` calls those two methods 
- [ ] Create and seperate current test cases so that each methods are tested better

hieuvubk:
@mattverse 
There are some changes to the CalcJoinPoolShares function. Do we still intend to separate it up?

PR #1750, opened 2022-06-10
prototype: JoinPoolNoSwap pool interface method

p0mvn:
Closes: #XXX

## What is the purpose of the change

> Add a description of the overall background and high level changes that this PR introduces

*(E.g.: This pull request improves documation of area A by adding ....*

## Brief Changelog

*(for example:)*
 
  - *The metadata is stored in the blob store on job creation time as a persistent artifact*
  - *Deployments RPC transmits only the blob storage reference*
  - *Daemons retrieve the RPC data from the blob cache*

## Testing and Verifying

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*
  - *Manually verified the change by ...*

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #1854, opened 2022-06-22
Separate calc join pool shares

hieuvubk:
Closes: #1753

Refactor code
Create two different methods calcSingleTokenJoinPoolShare and calcMultipleTokenPoolShare called by CalcJoinPoolShares base on tokenIns length


x/gamm/pool-models/balancer/amm.go
```diff
@@ -349,27 +349,44 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
 	// more complex / don't just alter the state.
 	// We should simplify this logic further in the future, using balancer multi-join equations.
 
+	if tokensIn.Len() == 1 {
```
p0mvn, line 352: Thanks for this PR @hieuvubk 

I think we might want to delegate this decision to the caller of this function. More specifically, have 2 separate exported methods:
- `CalcMultiTokenInJoinPoolShares`
- `CalcSingleTokenInJoinPoolShares`

However, I'm not sure if defining these 2 methods on the pool interface would make sense. Would these methods also work for stable swap?

Let's wait on @ValarDragon 's feedback on this before making any changes, please.

alexanderbez: I think this idea makes sense 👍

hieuvubk: Thanks for the suggestion!
I have a question. In amm pool model, **JoinPool** function calls to **CalcJoinPoolShares** internally, should we separate **JoinPool** to single and multi join or based on tokensIn length to call

p0mvn: Good point. Let's keep `JoinPool` as is for now and make it decide which "calc" method to call

For more context and as per https://github.com/osmosis-labs/osmosis/pull/1750, we are considering splitting the `JoinPool` abstraction to `JoinPool` and `JoinPoolNoSwap`.

If we make single and multi-asset join methods and also adopt #1750, there will be 4 different join pool variations.

We can investigate later to determine the best path forward. For now, to limit the scope, let's only deal with the calc methods, please

x/gamm/types/pool.go
```diff
@@ -59,9 +59,13 @@ type PoolI interface {
...
 
-	// CalcJoinPoolShares returns how many LP shares JoinPool would return on these arguments.
+	// CalcSingleTokenInJoinPoolShares returns how many LP shares JoinPool would return on these arguments( single token join).
 	// This does not mutate the pool, or state.
-	CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, err error)
+	CalcSingleTokenInJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, err error)
+
+	// CalcMultiTokenInJoinPoolShares returns how many LP shares JoinPool would return on these arguments( multi tokens join).
+	// This does not mutate the pool, or state.
+	CalcMultiTokenInJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, err error)
```
ValarDragon, line 68: hrmm, what if we kept the same `CalcJoinPoolShares` API, but then made the first lines of the function doing the split you wrote

hieuvubk: https://github.com/osmosis-labs/osmosis/pull/1854/commits/909568221759a11b14cc69ca25f8fcc4039134bd
should revert to this commit?

AlpinYukseloglu: I agree with @ValarDragon on adding two functions for `CalcJoinPoolShares` to split into based on its input, and I think splitting into single-token and multi-token joins would be be appropriate abstractions

I'm personally not convinced on the purpose of removing and replacing `CalcJoinPoolShares` altogether though, as it seems to me like we would just be floating the same logic in it up to `JoinPool`, which is already complex enough as it is

So +1 for keeping `CalcJoinPoolShares` and making it internally route to `CalcSingleTokenInJoinPoolShares` and `CalcMultiTokenInJoinPoolShares` based on the passed in token set

hieuvubk: https://github.com/osmosis-labs/osmosis/pull/1854/commits/1be8c4b359c2e395cde657403711424ea3e89629
Can u check this commit

hieuvubk: Thanks for the suggestion

alexanderbez:
Hey @hieuvubk! Where are we at with this PR? Do you need any assistance on pushing it through? Or is this blocked or perhaps no longer relevant?

hieuvubk:
> Hey @hieuvubk! Where are we at with this PR? Do you need any assistance on pushing it through? Or is this blocked or perhaps no longer relevant?

Sorry i just missed this pr :(((. Still keep CalcJoinPoolShares and call internal function depend on tokenIns length

ValarDragon:
Closing this in light of ^ 

Thanks so much for the work on this!

PR #1865, opened 2022-06-24
Update chain.schema.json

ValarDragon:

Add v10.0.1 to compatible versions


xBalbinus (approved):
Thanks! Excited to see the outcome later today

PR #1872, opened 2022-06-26
chore: x/mint genesis - audit, clean up and tests

p0mvn:
Subset of: https://github.com/osmosis-labs/osmosis/issues/1839

Audit, clean up, and tests for mint genesis.

## Brief Changelog
- Added extra validation for init genesis and removes what seemed to be a redundant check
- Added tests for export genesis
- Remove unused `integration_test.go`
- Refactored and tested `SetInitialSupplyOffsetDuringMigration` and `CreateDeveloperVestingModuleAccount` to better understand how these affect genesis tests
- Improved genesis tests


x/mint/keeper/genesis.go
```diff
@@ -12,12 +12,10 @@ func (k Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, bk types.Ba
 	k.SetMinter(ctx, data.Minter)
 	k.SetParams(ctx, data.Params)
 
-	if !ak.HasAccount(ctx, ak.GetModuleAddress(types.ModuleName)) {
-		ak.GetModuleAccount(ctx, types.ModuleName)
-		totalDeveloperVestingCoins := sdk.NewCoin(data.Params.MintDenom, sdk.NewInt(225_000_000_000_000))
-		k.CreateDeveloperVestingModuleAccount(ctx, totalDeveloperVestingCoins)
-		bk.AddSupplyOffset(ctx, data.Params.MintDenom, sdk.NewInt(225_000_000_000_000).Neg())
-	}
```
ValarDragon, line 20: This old logic seems correct? The account would get exported in ExportGenesis?

p0mvn: You are right. I assumed that this check would cover it: https://github.com/osmosis-labs/osmosis/blob/f793aaaf252cf15c81a793cc4e4fbcf8677a92ce/x/mint/keeper/keeper.go#L54-L56

On another look, it is completely unrelated

p0mvn:
The reason I focused on refactoring and testing`SetInitialSupplyOffsetDuringMigration` and `CreateDeveloperVestingModuleAccount` that might not be as important is to better understand how offsets work.

Testing `CreateDeveloperVestingModuleAccount` also helped me with genesis tests.

x/mint/module_test.go
```diff
@@ -1,29 +0,0 @@
...
-	abcitypes "github.com/tendermint/tendermint/abci/types"
-	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
-
-	simapp "github.com/osmosis-labs/osmosis/v7/app"
-	"github.com/osmosis-labs/osmosis/v7/x/mint/types"
-
-	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
-)
-
-func TestItCreatesModuleAccountOnInitBlock(t *testing.T) {
```
p0mvn, line 16: This is now tested in `TestMintInitGenesis`

p0mvn:
This PR is ready for another round of :eyes:

x/mint/keeper/keeper.go
```diff
@@ -53,25 +63,46 @@ func NewKeeper(
 }
 
 // SetInitialSupplyOffsetDuringMigration sets the supply offset based on the balance of the
-// Develop rVesting Module Account.  It should only be called one time during the initial
-// migration to v7.
-func SetInitialSupplyOffsetDuringMigration(ctx sdk.Context, k Keeper) {
+// developer vesting module account. CreateDeveloperVestingModuleAccount must be called
+// prior to calling this method. That is, developer vesting module account must exist when
+// SetInitialSupplyOffsetDuringMigration is called. Also, SetInitialSupplyOffsetDuringMigration
+//  should only be called one time during the initial migration to v7. This is done so because
```
mattverse, line 69:
```suggestion
// should only be called one time during the initial migration to v7. This is done so because
```

x/mint/keeper/keeper_test.go
```diff
@@ -26,6 +28,33 @@ func (suite *KeeperTestSuite) SetupTest() {
...
+// Specifically, it removes the developer vesting module account
+// from account keeper if isDeveloperModuleAccountCreated is true.
+// Additionally, it initializes suite's Ctx with blockHeight
+func (suite *KeeperTestSuite) setupDeveloperVestingModuleAccountTest(blockHeight int64, isDeveloperModuleAccountCreated bool) {
+	suite.Setup()
+	// Reset height to the desired value since test suite setup initialized
+	// it to 1.
+	suite.Ctx = suite.Ctx.WithBlockHeader(tmproto.Header{Height: blockHeight})
+
+	// If developer module account is created, the suite.Setup() also correctly sets the offset
```
mattverse, line 42:
```suggestion
	// If developer module account is not created, the suite.Setup() also correctly sets the offset
```

Maybe not? didn't fully understand what the comment here was indicating

x/mint/keeper/keeper_test.go
```diff
@@ -179,3 +208,96 @@ func (suite *KeeperTestSuite) TestDistrAssetToCommunityPoolWhenNoDeveloperReward
...
+
+		expectedError error
+	}{
+		"valid call": {
+			blockHeight:                     1,
+			isDeveloperModuleAccountCreated: true,
+		},
+		"dev vesting module account does not exist": {
+			blockHeight: 1,
+
```
mattverse, line 272:
```suggestion

```

x/mint/keeper/export_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+const (
+	DeveloperVestingAmount = developerVestingAmount
+)
+
+var (
+	ErrAmountCannotBeNilOrZero               = errAmountCannotBeNilOrZero
+	ErrDevVestingModuleAccountAlreadyCreated = errDevVestingModuleAccountAlreadyCreated
+	ErrDevVestingModuleAccountNotCreated     = errDevVestingModuleAccountNotCreated
+)
```
mattverse, line 11: nit: 
IMO, I don't think errors have to private variables that need secondary export for tests. 
what do you think about moving these to `errors.go` as other packages do and making them public functions?

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+	testCases := map[string]struct {
+		mintGenesis                     *types.GenesisState
+		mintDenom                       string
+		ctxHeight                       int64
+		isDeveloperModuleAccountCreated bool
+
+		expectPanic             bool
+		expectedEpochProvisions sdk.Dec
+		// Deltas represent by how much a certain paramets
+		// has changeda after calling InitGenesis()
```
mattverse, line 57:
```suggestion
		// has changed after calling InitGenesis()
```

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
-func TestMintInitGenesis(t *testing.T) {
-	app := simapp.Setup(false)
-	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
+var customGenesis = types.NewGenesisState(
+	types.NewMinter(sdk.ZeroDec()),
+	types.NewParams(
+		"uosmo",                  // denom
+		sdk.NewDec(200),          // epoch provisions
+		"year",                   // epoch identifier
+		sdk.NewDecWithPrec(5, 1), //reduction factor
```
mattverse, line 20:
```suggestion
		sdk.NewDecWithPrec(5, 1), // reduction factor
```

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+		5,                        // reduction perion in epochs
```
mattverse, line 21:
```suggestion
		5,                        // reduction period in epochs
```

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+		expectPanic             bool
+		expectedEpochProvisions sdk.Dec
+		// Deltas represent by how much a certain paramets
+		// has changeda after calling InitGenesis()
+		expectedSupplyOffsetDelta           sdk.Int
+		expectedSupplyWithOffsetDelta       sdk.Int
+		expectedDeveloperVestingAmountDelta sdk.Int
+		expectedHalvenStartedEpoch          int64
+	}{
+		"default genesis - developer module account is not created priot to InitGenesis() - created during the call": {
```
mattverse, line 63:
```suggestion
		"default genesis - developer module account is not created prior to InitGenesis() - created during the call": {
```

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+			mintGenesis: types.DefaultGenesisState(),
+			mintDenom:   sdk.DefaultBondDenom,
+
+			expectedEpochProvisions:             types.DefaultGenesisState().Params.GenesisEpochProvisions,
+			expectedSupplyOffsetDelta:           sdk.NewInt(keeper.DeveloperVestingAmount).Neg(),
+			expectedSupplyWithOffsetDelta:       sdk.ZeroInt(),
+			expectedDeveloperVestingAmountDelta: sdk.NewInt(keeper.DeveloperVestingAmount),
+		},
+		"default genesis - developer module account is created priot to InitGenesis() - not created during the call": {
```
mattverse, line 72:
```suggestion
		"default genesis - developer module account is created prior to InitGenesis() - not created during the call": {
```

mattverse (approved):
Left some minor comments(mostly typo fixes)

Otherwise, lgtm!

x/mint/keeper/keeper_test.go
p0mvn: [Addressed](https://github.com/osmosis-labs/osmosis/pull/1872/commits/906f06f962c1b197130b21c346cbe74d143eaa92)

x/mint/keeper/export_test.go
p0mvn: These exports are only compiled for tests. Any file that ends with `_test.go` does not get compiled into the regular binary.

I think a good practice is to have package-private variables and constants if they are not used anywhere else. If we see that we reuse the same variables/errors across multiple packages, then, I agree, it constitutes the need for moving into global scope / made exported. Let me know what you think, happy to discuss further

p0mvn:
Waiting for one more ACK since this is a > 100 LOC change

x/mint/keeper/genesis.go
```diff
@@ -6,17 +6,31 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/x/mint/types"
 )
 
+const developerVestingAmount = 225_000_000_000_000
+
 // InitGenesis new mint genesis.
 func (k Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, bk types.BankKeeper, data *types.GenesisState) {
+	if data == nil {
+		panic("nil mint genesis state")
+	}
```
czarcas7ic, line 15: Just for my learning, is there a reason you explicitly panic for a nil data input but not for any of the other inputs?

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
 
-func TestMintInitGenesis(t *testing.T) {
-	app := simapp.Setup(false)
-	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
+var customGenesis = types.NewGenesisState(
+	types.NewMinter(sdk.ZeroDec()),
+	types.NewParams(
+		"uosmo",                  // denom
+		sdk.NewDec(200),          // epoch provisions
+		"year",                   // epoch identifier
```
czarcas7ic, line 19: why is the epoch identifier set to "year" here?

czarcas7ic (approved):
👍

x/mint/keeper/genesis.go
```diff
@@ -6,17 +6,31 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/x/mint/types"
 )
 
+const developerVestingAmount = 225_000_000_000_000
+
 // InitGenesis new mint genesis.
 func (k Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, bk types.BankKeeper, data *types.GenesisState) {
```
ValarDragon, line 12: We should delete the `ak` and `bk` params here, and use them directly from `k`.

x/mint/keeper/genesis.go
```diff
@@ -6,17 +6,31 @@ import (
...
 	data.Minter.EpochProvisions = data.Params.GenesisEpochProvisions
 	k.SetMinter(ctx, data.Minter)
 	k.SetParams(ctx, data.Params)
 
-	if !ak.HasAccount(ctx, ak.GetModuleAddress(types.ModuleName)) {
-		ak.GetModuleAccount(ctx, types.ModuleName)
-		totalDeveloperVestingCoins := sdk.NewCoin(data.Params.MintDenom, sdk.NewInt(225_000_000_000_000))
-		k.CreateDeveloperVestingModuleAccount(ctx, totalDeveloperVestingCoins)
-		bk.AddSupplyOffset(ctx, data.Params.MintDenom, sdk.NewInt(225_000_000_000_000).Neg())
+	ak.GetModuleAccount(ctx, types.ModuleName)
```
ValarDragon, line 21: We should be able to delete this line, your right static check would help catch this =)

x/mint/keeper/genesis.go
ValarDragon: The keeper inputs should get removed

ValarDragon (approved):
Nice job on this!

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+			if tc.expectPanic {
+				suite.Panics(func() {
+					mintKeeper.InitGenesis(ctx, accountKeeper, bankKeeper, tc.mintGenesis)
+				})
+				return
+			}
+
+			suite.NotPanics(func() {
+				mintKeeper.InitGenesis(ctx, accountKeeper, bankKeeper, tc.mintGenesis)
+			})
```
ValarDragon, line 123: We should make a helper for this pattern. e.g. `osmoutils.ConditionalPanics(suite, expPanic, func)`

Lets file an issue for this?

x/mint/keeper/genesis_test.go
```diff
@@ -4,26 +4,183 @@ import (
...
+	}
+
+	for name, tc := range testCases {
+		suite.Run(name, func() {
+			// Setup.
+			suite.setupDeveloperVestingModuleAccountTest(tc.ctxHeight, tc.isDeveloperModuleAccountCreated)
+			ctx := suite.Ctx
+			accountKeeper := suite.App.AccountKeeper
+			bankKeeper := suite.App.BankKeeper
+			mintKeeper := suite.App.MintKeeper
```
ValarDragon, line 105: Something we could adapt our test standards is aliasing relevant keeper in the module's suite, if we find this helpful

x/mint/keeper/genesis_test.go
p0mvn: Just to test something other than the default of "day". For every field, I tried to choose something other than default.

My reasoning for this is to make sure that correctness does not depend only on a single possible value of the input.

say I did something like:
```
if epochIdentifier == "day" {
    // correct logic
}
```

Then, if we were to add a different epoch identifier in the future, it wouldn't make sense why it wouldn't work. Testing with at least one different input helps to cover such cases quickly

x/mint/keeper/genesis.go
p0mvn: Good catch, I don't know why I didn't see it

x/mint/keeper/genesis.go
p0mvn: I think this is likely to be needed. The name is confusing but, basically, this method creates a module account from the module name. If we remove it, the module account will not be created.

The line that starts failing if I remove this: https://github.com/osmosis-labs/osmosis/blob/c3081a9e10e73925db9ac89345d0e2953905c10d/x/mint/keeper/genesis_test.go#L127-L129

I ported this check from this removed file:
https://github.com/osmosis-labs/osmosis/blob/d9a4f94c15c9237289408689131acb04023f3c63/x/mint/module_test.go#L27

p0mvn: I will add a comment explaining why this is needed

x/mint/keeper/genesis_test.go
p0mvn: I like the idea. I will document this in an issue to further think about

x/mint/keeper/genesis_test.go
p0mvn: https://github.com/osmosis-labs/osmosis/issues/1946

x/mint/keeper/genesis.go
ValarDragon: oh wtf is the account keeper API if the get causes state changes to be saved :/

Make sense

x/mint/keeper/genesis_test.go
p0mvn: https://github.com/osmosis-labs/osmosis/issues/1947

x/mint/keeper/genesis.go
```diff
@@ -6,17 +6,34 @@ import (
...
 
-	if !ak.HasAccount(ctx, ak.GetModuleAddress(types.ModuleName)) {
-		ak.GetModuleAccount(ctx, types.ModuleName)
-		totalDeveloperVestingCoins := sdk.NewCoin(data.Params.MintDenom, sdk.NewInt(225_000_000_000_000))
-		k.CreateDeveloperVestingModuleAccount(ctx, totalDeveloperVestingCoins)
-		bk.AddSupplyOffset(ctx, data.Params.MintDenom, sdk.NewInt(225_000_000_000_000).Neg())
+	// The call to GetModuleAccount creates a module account if it does not exist.
+	k.accountKeeper.GetModuleAccount(ctx, types.ModuleName)
+
+	// k.accountKeeper.SetModuleAccount(ctx)
```
p0mvn, line 24:
```suggestion
```

Issue #1845, opened 2022-06-20
Improve tokenfactory/keeper/admins_test ChangeAdmin test


ValarDragon:
Within `xtokenfactory/keeper/admins_test.go`, I suggest we change the test case structure to improve readability.

Instead of `tc.setup()`, I think we should add four fields:
* `ChangeAdminMsg`
* `exoectedAdmin`
* `*MintMsg` (and can be nil)
* `expMintPass`

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/1691#discussion_r900979369_

PR #1876, opened 2022-06-27
Improve readability in x/tokenfactory test

mattverse:
Closes: [#1845 ](https://github.com/osmosis-labs/osmosis/issues/1845)

The purpose of this change is to have better readability for `admins_test.go`, where previously `TestChangeAdminDenom` had a structure where everything tested would be grouped in a single method of setup(), where now it has separate fields of testing, which consequently improves overall code readability.

## Brief Changelog

- Change `admins_test.go` to better follow tdd to improve overall readability.
- 
## Testing and Verifying

This change refactored existing tests.

### DIFF
--- a/x/tokenfactory/keeper/admins_test.go
+++ b/x/tokenfactory/keeper/admins_test.go
@@ -185,20 +185,11 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
 }
 
 func (suite *KeeperTestSuite) TestChangeAdminDenom() {
-	var (
-		addr0bal  int64
-		testDenom string
-		// addr1bal int64
-	)
-
-	// testDenom := "bitcoin"
 	for _, tc := range []struct {
-		desc                 string
-		msgChangeAdmin       func(denom string)
-		expectedAdminIndex   int
-		expectedAdminAddress string
-		msgMint              func(denom string)
-		expMintPass          bool
+		desc               string
+		msgChangeAdmin     func(denom string)
+		expectedAdminIndex int
+		msgMint            func(denom string)
 	}{
 		{
 			desc: "creator admin can't mint after setting to '' ",
@@ -211,7 +202,6 @@ func (suite *KeeperTestSuite) TestChangeAdminDenom() {
 				_, err := suite.msgServer.Mint(sdk.WrapSDKContext(suite.Ctx), types.NewMsgMint(suite.TestAccs[0].String(), sdk.NewInt64Coin(denom, 5)))
 				suite.Require().Error(err)
 			},
-			expMintPass: false,
 		},
 		{
 			desc: "non-admins can't change the existing admin",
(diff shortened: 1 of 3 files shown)


alexanderbez (approved):
🧀

x/tokenfactory/keeper/admins_test.go
```diff
@@ -185,82 +185,76 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
...
-	)
-
 	for _, tc := range []struct {
-		desc  string
-		setup func()
-		valid bool
+		desc               string
+		msgChangeAdmin     func(denom string)
+		expectedAdminIndex int
+		msgMint            func(denom string)
```
ValarDragon, line 192: Lets make the functions returns the message type directly?

Then LGTM, ty for doing this

mattverse: The reason I used a separate function is bc we need to take in the changed denom as a parameter of a function to used the updated denom. 

A possible option perhaps is changing the field to become `msgChangeAdmin func(denom string) *types.MsgChangeAdmin`, where the function directly returns the message?

ValarDragon: yeah!

mattverse: yeah agreed that testing layers should be all together, not in separate fields!

x/tokenfactory/keeper/admins_test.go
```diff
@@ -185,82 +185,72 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
...
 
-			tc.setup()
+			// expectedAdminIndex with negative value is assumed as admin with value of ""
+			if tc.expectedAdminIndex < 0 {
+				suite.Require().Equal("", queryRes.AuthorityMetadata.Admin)
+			} else {
+				suite.Require().Equal(suite.TestAccs[tc.expectedAdminIndex].String(), queryRes.AuthorityMetadata.Admin)
+			}
+
+			if tc.msgMint != nil {
```
p0mvn, line 250: Let's add a comment explaining why we are also testing mint when the test spec is focused on the change admin message. I think the goal here is to make sure that we can mint after changing admin?

x/tokenfactory/keeper/admins_test.go
```diff
@@ -185,82 +185,72 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
...
-					Denom: denom,
-				})
-				suite.Require().NoError(err)
-				suite.Require().Equal("", queryRes.AuthorityMetadata.Admin)
-
-				// can't mint.
-				_, err = suite.msgServer.Mint(sdk.WrapSDKContext(suite.Ctx), types.NewMsgMint(suite.TestAccs[0].String(), sdk.NewInt64Coin(denom, 5)))
-				suite.Require().Error(err)
-				suite.Require().Equal(suite.App.BankKeeper.GetBalance(suite.Ctx, suite.TestAccs[0], denom).Amount.Int64(), addr0bal)
+			msgChangeAdmin: func(denom string) *types.MsgChangeAdmin {
```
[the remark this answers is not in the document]
p0mvn: Ahh I see, we can probably still initialize the variable inside the loop after calling `suite.SetupTest()`. Wdyt?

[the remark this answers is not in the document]
mattverse: makes sense, but we would still have to pass in the denom as a parameter right?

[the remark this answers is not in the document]
p0mvn: You are totally right! I didn't notice that

I'll let you decide what you think is best. This minor readability improvement is proving to be more effort than I originally expected so I think it's okay to leave this as is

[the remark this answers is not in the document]
ValarDragon: Also we could extend in the future to cover other perceived edge cases wrt denoms, e.g. if we were worried about creator parsing.

I;m in favor of current approach

x/tokenfactory/keeper/admins_test.go
ValarDragon: Yeah, basically this is our "test" for the authority actually being correct

x/tokenfactory/keeper/admins_test.go
```diff
@@ -185,82 +185,76 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
...
-		addr1bal int64
-	)
-
 	for _, tc := range []struct {
-		desc  string
-		setup func()
-		valid bool
+		desc               string
+		msgChangeAdmin     func(denom string) *types.MsgChangeAdmin
+		expectedAdminIndex int
```
ValarDragon, line 191: Can we just make this `expectedAdmin` directly?

ValarDragon (approved):

x/tokenfactory/keeper/admins_test.go
mattverse: You mean and then have them set as suite.TestAccs[0] and so on right?

Here we have the same problem where TestAccs are reseted to random accounts every iteration for each test, while the values for these fields are initialized at the very beginning

ValarDragon: Hrmm, not sure I get it, but not worth blocking this imo

x/tokenfactory/keeper/admins_test.go
mattverse: added a comment here!

x/tokenfactory/keeper/admins_test.go
```diff
@@ -185,82 +185,72 @@ func (suite *KeeperTestSuite) TestBurnDenom() {
...
+			// expectedAdminIndex with negative value is assumed as admin with value of ""
+			if tc.expectedAdminIndex < 0 {
+				suite.Require().Equal("", queryRes.AuthorityMetadata.Admin)
+			} else {
+				suite.Require().Equal(suite.TestAccs[tc.expectedAdminIndex].String(), queryRes.AuthorityMetadata.Admin)
+			}
+
+			if tc.msgMint != nil {
+				_, err := suite.msgServer.Mint(sdk.WrapSDKContext(suite.Ctx), tc.msgMint(testDenom))
+				suite.Require().NoError(err)
```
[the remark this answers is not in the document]
mattverse: The sencond test case would be the case!

Issue #1877, opened 2022-06-27
Add invariant to `x/lockup` for module balance


mattverse:
## Background

`x/lockup` keeper should have an invariant to ensure that the module balance and the sum of all tokens within all locks have the equivalent amount of tokens.

## Suggested Design

Add this invariant check in `invariants.go`

PR #1884, opened 2022-06-27
Locks amount invariant

hieuvubk:
Closes: #1877

> Add an invariant to ensure that the module balance and the sum of all tokens within all locks have the equivalent amount of tokens.


mattverse (changes requested):
This is awesome! Thanks!

Adding invariants require alot of precaution, as wrong invariants can lead to chain halt. 

Can we test this invariant before merging? I won't mind testing it through either simapp or directly off current state from mainnet

ValarDragon:
Wrong invariants can't lead to a chain halt! Invariants causing a chain halt actually never worked

mattverse:
Oh wow, didn't know that invariants causing chain halt never worked. If that's the case, what's the exact role of invariants?

alexanderbez:
> Oh wow, didn't know that invariants causing chain halt never worked. If that's the case, what's the exact role of invariants?

They were meant to but don't actually do that. Invariants are very useful for simulations and other types of simulated/e2e testing, to ensure your state machine is in a state that is considered correct. We should fix x/crisis, but that's orthogonal IMO.

alexanderbez (approved):

ValarDragon:
(Merging based on matt's invariant logic LGTM message + Bez review)

mattverse:
Oh interesting!

We should still have this invariant tested locally at least before merging?

PR #1901, opened 2022-06-30
x/txfees: Internal audit and test cleanup

stackman27:
added documentation to x/txfees keeper methods and changed test to follow table driven approach. 

Very new to tdd tests so apologies for any simple errors 😅

## Brief Changelog
n/a

## Testing and Verifying
Changed `hook_test.go` to table driven 


x/txfees/keeper/hooks_test.go
```diff
@@ -37,49 +37,66 @@ func (suite *KeeperTestSuite) TestTxFeesAfterEpochEnd() {
...
+		suite.App.BankKeeper.SendCoinsFromAccountToModule(suite.Ctx, addr0, types.NonNativeFeeCollectorName, sdk.Coins{tc.coin})
+
+		suite.Require().True(suite.App.BankKeeper.HasBalance(suite.Ctx, moduleAddrNonNativeFee, tc.coin))
+
+		suite.App.EpochsKeeper.AfterEpochEnd(futureCtx, params.DistrEpochIdentifier, int64(1))
+		moduleBaseDenomBalance := suite.App.BankKeeper.GetBalance(suite.Ctx, moduleAddrFee, tc.baseDenom)
+		suite.Require().Empty(suite.App.BankKeeper.GetAllBalances(suite.Ctx, moduleAddrNonNativeFee))
+		// Tried to make use Equal() to check the exact same value but moduleBaseDenomBalance.Amount seems to be incrementing
+		// by 2000000 per test hmm
+		suite.Require().True(moduleBaseDenomBalance.Amount.GTE(finalOutputAmount))
```
stackman27, line 100: Tried to make use Equal() to check the exact same value but moduleBaseDenomBalance.Amount seems to be incrementing by 2000000 per test 🤔

mattverse: it's probably because the variable `finalOutputAmount` is defined outside the test cases iteration. Try moving the variable within the test loops! OR this assertion can be moved outside the loops

x/txfees/keeper/hooks_test.go
```diff
@@ -37,49 +37,66 @@ func (suite *KeeperTestSuite) TestTxFeesAfterEpochEnd() {
...
+
+		finalOutputAmount = finalOutputAmount.Add(expectedOutput.Amount)
+		_, _, addr0 := testdata.KeyTestPubAddr()
+		simapp.FundAccount(suite.App.BankKeeper, suite.Ctx, addr0, sdk.Coins{tc.coin})
+		suite.App.BankKeeper.SendCoinsFromAccountToModule(suite.Ctx, addr0, types.NonNativeFeeCollectorName, sdk.Coins{tc.coin})
+
+		suite.Require().True(suite.App.BankKeeper.HasBalance(suite.Ctx, moduleAddrNonNativeFee, tc.coin))
+
+		suite.App.EpochsKeeper.AfterEpochEnd(futureCtx, params.DistrEpochIdentifier, int64(1))
+		moduleBaseDenomBalance := suite.App.BankKeeper.GetBalance(suite.Ctx, moduleAddrFee, tc.baseDenom)
```
mattverse, line 96: maybe we want to add test for individual balance here?

mattverse (changes requested):
It seems that the `hooks_test.go` change came directly from changing the existing tests into table format, but I think tdd should better test and account for individual test cases, rather than focusing on collective, sequential testings of the test cases.

x/txfees/keeper/hooks_test.go
stackman27: Actually there was a bug in the test and the test was testing a different `AfterEpochEnd` method. 
It's supposed to be testing `suite.App.TxFeesKeeper.AfterEpochEnd` but it was testing suite.App.EpochsKeeper.AfterEpochEnd` which made the native-basedenom increment by 2000000

x/txfees/keeper/feetokens.go
```diff
@@ -32,6 +32,10 @@ func (k Keeper) ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin
 	return sdk.NewCoin(baseDenom, spotPrice.MulInt(inputFee.Amount).RoundInt()), nil
 }
 
+// CalcFeeSpotPrice convert the provided tx fees into their equivalent value in the base denomination.
```
alexanderbez, line 35:
```suggestion
// CalcFeeSpotPrice converts the provided tx fees into their equivalent value in the base denomination.
```

x/txfees/keeper/feetokens.go
```diff
@@ -32,6 +32,10 @@ func (k Keeper) ConvertToBaseToken(ctx sdk.Context, inputFee sdk.Coin) (sdk.Coin
...
+// Spot Price Calculation: spotPrice / (1 - swapFee),
+// where spotPrice is defined as:
+// `(tokenBalanceIn / tokenWeightIn) / (tokenBalanceOut / tokenWeightOut)`
```
alexanderbez, line 38:
```suggestion
// (tokenBalanceIn / tokenWeightIn) / (tokenBalanceOut / tokenWeightOut).
```

AlpinYukseloglu:
Thanks for taking a crack at this @stackman27! This new set of table driven tests is very useful, but it tests a slightly different thing than what the original tests were testing (which means we'll have more coverage when it's done!)

The original tests ran the following logic (total # of epochs passed: 1)
1. Run multiple transactions with different non-osmo fees and collect fee tokens in the module account
2. Make sure all non-osmo fees are correctly swapped into osmo at epoch end 

Your tests seem to run the following similar but different logic (total # of epochs passed: n, where n = # test cases)
1. Run a tx with a non-osmo fee and collect it in the module account
2. Make sure that fee is swapped into osmo at epoch end
3. Repeat for a couple other tokens, checking to make sure the module osmo balance increases by the correct amount each time

In general, we want each test case in a set of table driven tests to represent an independent case, so your current setup has three separate cases for the same thing (single non-osmo fee swap at epoch end). I would suggest adding a case (or a new set of cases if it's too messy to change inputs to the current one) that tests the situation where there are multiple non-osmo fee tokens in the module account at the end of an epoch

x/txfees/keeper/hooks_test.go
```diff
@@ -37,49 +37,71 @@ func (suite *KeeperTestSuite) TestTxFeesAfterEpochEnd() {
...
+		{
+			name:      "TxFees AfterEpochEnd for atom",
+			coin:      sdk.NewInt64Coin(atom, 20),
+			baseDenom: baseDenom,
+			denom:     atom,
+			poolType:  atomPool,
+			swapFee:   sdk.NewDec(0),
+		},
+		{
+			name:      "TxFees AfterEpochEnd for ust :(",
```
AlpinYukseloglu, line 66:
```
:(
```

:( indeed

x/txfees/keeper/hooks_test.go
AlpinYukseloglu: Great catch

stackman27:
> Thanks for taking a crack at this @stackman27! This new set of table driven tests is very useful, but it tests a slightly different thing than what the original tests were testing (which means we'll have more coverage when it's done!)
> 
> The original tests ran the following logic (total # of epochs passed: 1)
> 
> 1. Run multiple transactions with different non-osmo fees and collect fee tokens in the module account
> 2. Make sure all non-osmo fees are correctly swapped into osmo at epoch end
> 
> Your tests seem to run the following similar but different logic (total # of epochs passed: n, where n = # test cases)
> 
> 1. Run a tx with a non-osmo fee and collect it in the module account
> 2. Make sure that fee is swapped into osmo at epoch end
> 3. Repeat for a couple other tokens, checking to make sure the module osmo balance increases by the correct amount each time
> 
> In general, we want each test case in a set of table driven tests to represent an independent case, so your current setup has three separate cases for the same thing (single non-osmo fee swap at epoch end). I would suggest adding a case (or a new set of cases if it's too messy to change inputs to the current one) that tests the situation where there are multiple non-osmo fee tokens in the module account at the end of an epoch

Thank you for the suggestion Alp! I added Multi non-osmo fee token just like you suggested. Please let me know if there's anything else to add 

Few things that I want to clarify; 
1. Should the osmo amount in module account reset after each epoch? My guess is no, but i just want to clarify 
2. Is there a way in which we can test for failure in `AfterEpochEnd`? I guess the only failure would be the funds not actually being transferred to osmo module account but is there anything else you suggest

AlpinYukseloglu:
> Thank you for the suggestion Alp! I added Multi non-osmo fee token just like you suggested. Please let me know if there's anything else to add
> 
> Few things that I want to clarify;
> 
> 1. Should the osmo amount in module account reset after each epoch? My guess is no, but i just want to clarify
> 2. Is there a way in which we can test for failure in `AfterEpochEnd`? I guess the only failure would be the funds not actually being transferred to osmo module account but is there anything else you suggest

1. I believe if you are only triggering the `AfterEpochEnd` event in the TxFees module then it shouldn't change. If you're triggering the event globally, then the fee tokens should be distributed at epoch end and leave the account (haven't dug in that part of the codebase for a while so not 100% sure)
2. The case you mentioned is a good one, but an even more important one is when the tokens are sent as-is to the osmo module (i.e. 100 foo tokens get sent and distributed) since this is how it operated before we implemented this auto-swap feature. Another failure case could be if swap outputs are off by a lot (e.g. 0 osmo is collected when we expect 100)

Not sure if this is helpful/what you were asking about but happy to discuss offline too

AlpinYukseloglu (approved):
LGTM!

mattverse (commented):
Don't want to be a huge blocker for this PR, but IMO I think we need to test the balance of the module account as well for the tx fees, not only checking if it's not empty

stackman27:
> Don't want to be a huge blocker for this PR, but IMO I think we need to test the balance of the module account as well for the tx fees, not only checking if it's not empty

Thank you @mattverse! should be good now

alexanderbez (approved):

PR #1923, opened 2022-06-30
e2e: remove unused skipRunValidatorIndexes

p0mvn:
Closes: #XXX

This was originally meant to be used for initializing genesis and configs for state-sync nodes and be able to run them after the rest of the chain has started. However, a new design was chosen over this where we have the ability to initialize the configs for a single node, separate from the rest of the chain

As a result, this logic is not needed anymore

### DIFF
--- a/tests/e2e/containers/containers.go
+++ b/tests/e2e/containers/containers.go
@@ -88,10 +88,6 @@ func (m *Manager) ExecCmd(t *testing.T, chainId string, validatorIndex int, comm
 				return false
 			}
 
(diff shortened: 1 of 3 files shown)


czarcas7ic (approved):
I am glad we are doing away with this, it was just one more thing to remember to do when going through the setup process (also it oddly took a long time to just skip the validator, at least when looking at the logs)

p0mvn:
@czarcas7ic thanks for quick reviews everywhere

PR #1902, opened 2022-06-30
fix: e2e flakiness related to err checking containers

p0mvn:
Closes: #XXX

## What is the purpose of the change

#1898 introduced flakiness. Apparently, Hermes sometimes produces errors to stderr when it functions correctly. Therefore, the added logic prevents IBC from functioning correctly.

For some reason, sometimes Hermes does not produce these logs so the CI passes.

This logic is not very useful so can be removed to fix CI. Test locally several times, and it works.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

PR #1898, opened 2022-06-29
e2e: chain initialization refactor for extensibility

p0mvn:
Part of: https://github.com/osmosis-labs/osmosis/issues/1879

## What is the purpose of the change

Our current chain initialization design does not allow for spinning up single nodes. This functionality is useful for testing feautes like state-sync.

This PR introduced the barebones structure to initialize single nodes via Docker. To build a single node initialization image, run:
```
make docker-build-e2e-init-node
```
Additional information about the new design can be found in `tests/e2e/initialization/README.md`

## Brief Changelog

- move initialization logic to `initialzation` package
- refactor init Dockerfile to support scripts provided by `ARG`
- create a stub script for single node initialization

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? 
      * https://github.com/osmosis-labs/osmosis/blob/0f3677649a43ca8ea6a906f49fb70d4565b514e9/tests/e2e/initialization/README.md

PR #1924, opened 2022-06-30
e2e: ExecCmd debug log on error

p0mvn:
Closes: #XXX

This change was originally introduced in #1898. However, in #1898 if "error" was found, the system would fail on the following line:
https://github.com/osmosis-labs/osmosis/blob/8f31b5afdbc9dfdbfbc20045dcb7f7004ae72b53/tests/e2e/containers/containers.go#L101

That started causing issues because some applications such as Hermes log "error" to stderr in the normal mode of operation.  So the change was reverted in #1902

This log proved to be incredibly helpful, making me manually change every PR for debugging. So, a modified version is introduced here. This version does not fail but only logs when "error is encountered. 

## Testing and Verifying

### DIFF
--- a/tests/e2e/containers/containers.go
+++ b/tests/e2e/containers/containers.go
@@ -106,7 +106,7 @@ func (m *Manager) ExecCmd(t *testing.T, chainId string, validatorIndex int, comm
 			}
 
 			if success != "" {
-				return strings.Contains(outBuf.String(), success) || strings.Contains(errBuf.String(), success)
+				return strings.Contains(outBuf.String(), success) || strings.Contains(errBufString, success)
 			}
 
 			return true
--- a/tests/e2e/containers/containers.go
+++ b/tests/e2e/containers/containers.go
@@ -109,15 +109,11 @@ func (m *Manager) ExecCmd(t *testing.T, chainId string, validatorIndex int, comm
 				return strings.Contains(outBuf.String(), success) || strings.Contains(errBuf.String(), success)
 			}
 
-			if success != "" {
-				return strings.Contains(outBuf.String(), success) || strings.Contains(errBuf.String(), success)
-			}
-
 			return true
 		},
 		time.Minute,
 		time.Second,
-		"tx returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
+		"tx returned a non-zero code",
 	)
 
 	return outBuf, errBuf, nil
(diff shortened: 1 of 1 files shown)


czarcas7ic:
Can you give an example where this would be useful? I am just lacking context, since as it stands when an error occurs it eventually halts the e2e test suite and prints the error to the terminal right?

p0mvn:
> eventually halts the e2e test suite and prints the error to the terminal right

It does not print anything. 

The arguments are evaluated before running the callback so it prints empty. In fact, we should remove this line because it always prints empt for errBuf and outBuf:
```
"tx returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String()
```


```
        // 1. We start here with empty errBuf 
	require.Eventually(
		t,
		func() bool {
	                      // 3. We update errBug here (after it was evaluated in 2.)
		},
		time.Minute,
		time.Second,
		"tx returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),  // 2. This gets evaluated so this is always empty
	)
```

p0mvn:
I made some updates, let me know what you think. 

This change also allows us to see if there is an error right away as opposed to waiting for 1 minute to get a log without stdout or stderr

tests/e2e/containers/containers.go
```diff
@@ -106,7 +106,7 @@ func (m *Manager) ExecCmd(t *testing.T, chainId string, validatorIndex int, comm
 			}
 
 			if success != "" {
-				return strings.Contains(outBuf.String(), success) || strings.Contains(errBuf.String(), success)
+				return strings.Contains(outBuf.String(), success) || strings.Contains(errBufString, success)
```
czarcas7ic, line 109: The only problem I have with this is, wouldn't this not trigger if the error said something like "this tx was not successful" since the string contains success?

czarcas7ic: Oh wait my bad, thats not the case

czarcas7ic (approved):
LGTM now!

tests/e2e/containers/containers.go
p0mvn: but this is exactly as it used to be. In some cases, we expect stderr to containers success results. That's why compare stderr against success

Also note that we are comparing against `success` variable, not "success" string

p0mvn: I think I accidentally duplicated that success code in my very first commit. Sorry if that was confusing

czarcas7ic: Yeah this is acceptable, I misread it on first pass

ValarDragon (approved):

Issue #1929, opened 2022-07-01
x/gamm: Add check to ensure JoinPoolNoSwap cannot be called with arbitrary denoms


AlpinYukseloglu:
## Background

Our current implementation of `JoinPoolNoSwap` only checks for whether the input set of tokens _includes_ the tokens in the pool, but does not have any check that prevents someone from putting in an arbitrary number of arbitrary tokens. Since this is a no-swap/all-asset join, there are no other immediate checks so an invalid input can make it directly to the state-changing `JoinPool` function.

https://github.com/osmosis-labs/osmosis/blob/add306ebc9f588648cfa8c626350c85ad28da6e4/x/gamm/keeper/pool_service.go#L180

An example of how this could potentially be exploited for the OSMO/USDC pool (1 OSMO = 1 USDC):
1. Create arbitrary spam/poisoned token FOO
2. Create a `JoinPoolNoSwap` tx and send in a set of [1 OSMO, 1 USDC, 1 FOO] as your input
3. The following check doesn't catch the existence of FOO, since we only check if the input includes 1 OSMO and 1 USDC:
https://github.com/osmosis-labs/osmosis/blob/add306ebc9f588648cfa8c626350c85ad28da6e4/x/gamm/keeper/pool_service.go#L203

## Suggested Design

- Add `&& tokenInMaxs.DenomSubsetOf(neededLpLiquidity)` to x/gamm/keeper/pool_service.go/L203 so that input tokens for a no-swap pool join don’t include denoms that aren’t in the pool:
https://github.com/osmosis-labs/osmosis/blob/add306ebc9f588648cfa8c626350c85ad28da6e4/x/gamm/keeper/pool_service.go#L203

## Acceptance Criteria

- All existing and new tests pass

PR #1930, opened 2022-07-01
x/gamm: Add check to ensure JoinPoolNoSwap cannot be called with arbitrary denoms

AlpinYukseloglu:
Closes: #1929

Our current implementation of JoinPoolNoSwap only checks for whether the input set of tokens includes the tokens in the pool, but does not have any check that prevents someone from putting in an arbitrary number of arbitrary tokens. Since this is a no-swap/all-asset join, there are no other immediate checks so an invalid input can make it directly to the state-changing JoinPool function.

## Brief Changelog
 
  - Add && tokenInMaxs.DenomSubsetOf(neededLpLiquidity) to x/gamm/keeper/pool_service.go/L203 so that input tokens for a no-swap pool join don’t include denoms that aren’t in the pool:

## Testing and Verifying

This change is already covered by existing tests in our GAMM module

### DIFF
--- a/x/gamm/keeper/pool_service.go
+++ b/x/gamm/keeper/pool_service.go
@@ -200,9 +200,12 @@ func (k Keeper) JoinPoolNoSwap(
 	// check that needed lp liquidity does not exceed the given `tokenInMaxs` parameter. Return error if so.
 	// if tokenInMaxs == 0, don't do this check.
 	if tokenInMaxs.Len() != 0 {
-		if !(neededLpLiquidity.DenomsSubsetOf(tokenInMaxs) && tokenInMaxs.IsAllGTE(neededLpLiquidity) && tokenInMaxs.DenomsSubsetOf(neededLpLiquidity)) {
+		if !(neededLpLiquidity.DenomsSubsetOf(tokenInMaxs) && tokenInMaxs.IsAllGTE(neededLpLiquidity)) {
 			return sdkerrors.Wrapf(types.ErrLimitMaxAmount, "TokenInMaxs is less than the needed LP liquidity to this JoinPoolNoSwap,"+
 				" upperbound: %v, needed %v", tokenInMaxs, neededLpLiquidity)
+		} else if !(tokenInMaxs.DenomsSubsetOf(neededLpLiquidity)) {
+			return sdkerrors.Wrapf(types.ErrDenomNotFoundInPool, "TokenInMaxs includes tokens that are not part of the target pool,"+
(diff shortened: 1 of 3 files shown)


x/gamm/keeper/pool_service.go
```diff
@@ -200,7 +200,7 @@ func (k Keeper) JoinPoolNoSwap(
 	// check that needed lp liquidity does not exceed the given `tokenInMaxs` parameter. Return error if so.
 	// if tokenInMaxs == 0, don't do this check.
 	if tokenInMaxs.Len() != 0 {
-		if !(neededLpLiquidity.DenomsSubsetOf(tokenInMaxs) && tokenInMaxs.IsAllGTE(neededLpLiquidity)) {
+		if !(neededLpLiquidity.DenomsSubsetOf(tokenInMaxs) && tokenInMaxs.IsAllGTE(neededLpLiquidity) && tokenInMaxs.DenomsSubsetOf(neededLpLiquidity)) {
```
p0mvn, line 203: It feels that the error message does not apply here:
```
TokenInMaxs is less than the needed LP liquidity to this JoinPoolNoSwap
```

What do you think about having a separate check that produces a relevant error message?

p0mvn, line 203: Also, please add a test case for the new condition

ValarDragon:
LGTM once a test case is added

x/gamm/keeper/pool_service.go
AlpinYukseloglu: Agreed – refactored the new logic to have a new error message that says the following and returns an `ErrDenomNotFoundInPool` error:

```
"TokenInMaxs includes tokens that are not part of the target pool,"
```

AlpinYukseloglu: MB – added!

AlpinYukseloglu:
> LGTM once a test case is added

Done!

ValarDragon (approved):

ValarDragon:
Awesome! Can you add a changelog entry under bug fix, and then merge this?

p0mvn (approved):

AlpinYukseloglu:
> Awesome! Can you add a changelog entry under bug fix, and then merge this?

Ready to merge (not authorized to merge myself)!

ValarDragon:
Err can you fix your markdown editor locally, to not overwrite all the bulletpoints :sweat_smile: 

I'll commit to this branch to get this changelog change through

Issue #1906, opened 2022-06-30
x/gamm: Add check for correct input asset denoms to CalcJoinPoolShares


AlpinYukseloglu:
## Background

Our [current implementation](https://github.com/osmosis-labs/osmosis/blob/85529c76683f2d2f2ffb2121125baf0e39463f84/x/gamm/pool-models/balancer/amm.go#L336) of `CalcJoinPoolShares` checks to make sure the _number_ of denoms sent in for an all-asset join is correct ([here](https://github.com/osmosis-labs/osmosis/blob/85529c76683f2d2f2ffb2121125baf0e39463f84/x/gamm/pool-models/balancer/amm.go#L368-L370)), but has no check for whether those assets are actually the correct denoms. This makes it so that entirely invalid inputs (i.e. correct number of input tokens but incorrect denoms) successfully make it through the entire set of all-asset join calculations until they are incidentally caught in the tail-end single-asset join section in a [check](https://github.com/osmosis-labs/osmosis/blob/85529c76683f2d2f2ffb2121125baf0e39463f84/x/gamm/pool-models/balancer/amm.go#L390) in `updateIntermediaryPoolAssetsLiquidity`.

I suggest we add an additional check to the input tokens that makes sure they are of the correct denoms so that the function never assumes invalid inputs are correct for calculating all-asset joins.

## Suggested Design

- Add a check [here](https://github.com/osmosis-labs/osmosis/blob/85529c76683f2d2f2ffb2121125baf0e39463f84/x/gamm/pool-models/balancer/amm.go#L368-L370) that makes sure that the denoms are equivalent as well (simplest way is probably just to use DenomSubsetOf to ensure both are denom subsets of each other)

## Acceptance Criteria

- All existing and new tests should pass[
](https://github.com/osmosis-labs/osmosis/pull/1858)

PR #1931, opened 2022-07-01
x/gamm: Add check for correct input asset denoms to CalcJoinPoolShares

AlpinYukseloglu:
Closes: #1906 

Our current implementation of CalcJoinPoolShares checks to make sure the _number_ of denoms sent in for an all-asset join is correct, but has no check for whether those assets are actually the correct denoms. This makes it so that entirely invalid inputs (i.e. correct number of input tokens but incorrect denoms) successfully make it through the entire set of all-asset join calculations until they are incidentally caught in the single-asset join section due to a check in `updateIntermediaryPoolAssetsLiquidity`.

https://github.com/osmosis-labs/osmosis/blob/85529c76683f2d2f2ffb2121125baf0e39463f84/x/gamm/pool-models/balancer/amm.go#L368-L370

This PR adds an additional check to the input tokens that makes sure they are of the correct denoms so that the function never assumes invalid inputs are correct for calculating all-asset joins.

## Brief Changelog

- Adds a check similar to `DenomsSubsetOf` that makes sure that the input denoms are part of the pool being joined (Note: we cannot use `DenomsSubsetOf` here because we have a denom->PoolAsset map instead of a set of `sdk.Coins` to represent pool liquidity)

## Testing and Verifying

This change is already covered by existing GAMM tests.


mattverse (approved):
I do think that the existing code base does have checks within the methods of `CalcJoinPoolShares` already, but Im generally agreeing on this extra explicit sanity check!

alexanderbez (approved):
LGTM

1. Do we have a test case already testing this path? If not, we should definitely add one.
2. We should get a changelog entry in.

AlpinYukseloglu:
> LGTM
> 
> 1. Do we have a test case already testing this path? If not, we should definitely add one.

I believe this test should cover it but happy to add another one with more assets/mixed assets if you think it would be useful:
https://github.com/osmosis-labs/osmosis/blob/2007babdee01e9bb7743ccadeff097cdab9b7236/x/gamm/pool-models/balancer/pool_test.go#L121

> 2. We should get a changelog entry in.

Added in recent commit!

p0mvn:
> > LGTM
> > 
> > 1. Do we have a test case already testing this path? If not, we should definitely add one.
> 
> I believe this test should cover it but happy to add another one with more assets/mixed assets if you think it would be useful:
> 
> https://github.com/osmosis-labs/osmosis/blob/2007babdee01e9bb7743ccadeff097cdab9b7236/x/gamm/pool-models/balancer/pool_test.go#L121
> 
> > 2. We should get a changelog entry in.
> 
> Added in recent commit!

Just making sure we don't miss anything - was that commit pushed? Not seeing a new test

p0mvn (approved):
LGTM

AlpinYukseloglu:
> Just making sure we don't miss anything - was that commit pushed? Not seeing a new test

This test was already there so it's not a part of this PR. The link I included was directly from main

alexanderbez (approved):

Issue #476, opened 2021-09-13
Make all incentive keeper gauge tests table driven


ValarDragon:
Currently all of our incentives keeper gauge tests are incredibly ad-hoc scripts, that are unclear to grok at a glance. They reuse lots of setup code as well.

Several of these tests even have very common logic, e.g. Perpetual vs NonPerpetual

We should instead switch to using table driven tests for each of these, which will dramatically simplify the readability of the tests, and ease adding new cases.

PR #1932, opened 2022-07-01
x/incentives: Simplify gauge tests in module

rrrliu:
Closes: #476

Combined three separate tests that shared significant boilerplate into a single table-driven test that tests both perpetual / non-perpetual gauges and their denoms.

## Brief Changelog

- Modified `x/incentive/keeper/gauge_test.go`

## Testing and Verifying

This change is a no-op in terms of test functionality, so it's already covered since the logic is the same.

  - How is the feature or change documented? n/a


x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +117,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
...
 
 		// check gauge ids by other denom
-		gaugeIds = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lpt")
-		suite.Require().Len(gaugeIds, 0)
+		gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lpt")
+		suite.Require().Len(gaugeIdsByDenom, 0)
 
-		// distribute coins to stakers
-		_, err = suite.App.IncentivesKeeper.Distribute(suite.Ctx, []types.Gauge{*gauge})
+		// distribute coins to stakers, since it's perpetual distribute everything on single distribution
```
mattverse, line 132: seems like its a deprecated comment

mattverse (changes requested):
Overall, I feel like more generalization could be done to better achieve the purpose of tdd! (maybe Im getting this impression bc we don't have much test cases)

Left some minor comments! Please take a look

rrrliu:
> Overall, I feel like more generalization could be done to better achieve the purpose of tdd! (maybe Im getting this impression bc we don't have much test cases)
> 
> Left some minor comments! Please take a look

Yeah, I agree that having just two cases (perpetual and non-perpetual gauges) is underutilizing table-driven testing a bit. Couple ideas I was thinking, curious to get your thoughts + any other ideas you had!

---

Add more test cases where we vary the other initialization parameters of the gauge beyond just `isPerpetual`. We could add `coins` or `numEpochsPaidOver` as input fields in the testing struct, which gives us a lot more "rows" in our table to test.

---

We have code like this checking for # of gauges per denom:

```go
// check gauge ids by denom
gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lptoken")
suite.Require().Len(gaugeIdsByDenom, 1, "perpetual %b", tc.isPerpetual)

// start distribution
// ...

// check gauge ids by denom
gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lptoken")
suite.Require().Len(gaugeIdsByDenom, 1)

// check gauge ids by other denom
gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lpt")
suite.Require().Len(gaugeIdsByDenom, 0)

// distribute coins to stakers
// ...

gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lptoken")
suite.Require().Len(gaugeIdsByDenom, expectedNumGauges)
```

Was initially thinking about generalizing each `check gauge ids by denom` into a struct since it's so repetitive and making it a sub-field in our `testCases` struct or a separate array altogether within the main `for` loop.

```go
struct {
    denom: string
    expectedGaugeIds: []uint64
}
```

A similar struct could also be made for checking upcoming, active, and finished gauges.

The main problem was that in between many of the `check gauge ids by denom`, we have code that mutates state (starting distribution, distributing coins, finishing, etc.) which creates dependencies between different "rows" of the table. Each row would no longer be its own independent unit test, we'd probably have to add some callback function to the struct to perform this mutation. At this point, it seemed like refactoring would be less clean than just leaving it as is. However, could prove useful if we do end up checking for way more denoms / cases.

alexanderbez (approved):

x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +114,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
...
+
+		// check active gauges
+		gauges = suite.App.IncentivesKeeper.GetActiveGauges(suite.Ctx)
+		suite.Require().Len(gauges, 1)
+
+		// check gauge ids by denom
+		gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lptoken")
+		suite.Require().Len(gaugeIdsByDenom, 1)
 
 		// finish distribution for non perpetual gauge
```
AlpinYukseloglu, line 162: This whole section should be moved under the non-perpetual gauge check on line 169 since it has the same condition right?

x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +114,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
...
 		if !gauge.IsPerpetual {
 			err = suite.App.IncentivesKeeper.MoveActiveGaugeToFinishedGauge(suite.Ctx, *gauge)
 			suite.Require().NoError(err)
 		}
 
+		// check finished gauges
```
AlpinYukseloglu, line 168:
```suggestion
		// check non-perpetual/finished gauges
```

x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +114,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
 		err = suite.App.IncentivesKeeper.MoveUpcomingGaugeToActiveGauge(suite.Ctx, *gauge)
 		suite.Require().NoError(err)
```
AlpinYukseloglu, line 116:
```suggestion
		gauges = suite.App.IncentivesKeeper.GetActiveGauges(suite.Ctx)
		suite.Require().Len(gauges, 1)
```

We want to make sure the gauge was successfully moved to the set of active gauges. This also serves as our active gauge check for non-perpetual gauges, since the current implementation only has one for perpetual ones (line 154)

x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +114,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
...
-		gaugeIds = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lpt")
-		suite.Require().Len(gaugeIds, 0)
+		gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lpt")
+		suite.Require().Len(gaugeIdsByDenom, 0)
 
 		// distribute coins to stakers
-		_, err = suite.App.IncentivesKeeper.Distribute(suite.Ctx, []types.Gauge{*gauge})
+		distrCoins, err := suite.App.IncentivesKeeper.Distribute(suite.Ctx, []types.Gauge{*gauge})
 		suite.Require().NoError(err)
+		suite.Require().Equal(sdk.Coins{sdk.NewInt64Coin("stake", int64(10/tc.expectedNumEpochsPaidOver))}, distrCoins)
```
AlpinYukseloglu, line 132:
```suggestion
		// SetupLockAndGauge hard-codes 10 "stake" tokens
		suite.Require().Equal(sdk.Coins{sdk.NewInt64Coin("stake", int64(10/tc.expectedNumEpochsPaidOver))}, distrCoins)
```

x/incentives/keeper/gauge_test.go
```diff
@@ -43,235 +43,69 @@ func (suite *KeeperTestSuite) TestNonExistentDenomGaugeCreation() {
...
+		suite.Require().Equal(expectedGauge.String(), gauges[0].String())
+
+		// check gauge ids by denom
+		gaugeIdsByDenom = suite.App.IncentivesKeeper.GetAllGaugeIDsByDenom(suite.Ctx, "lptoken")
+		suite.Require().Len(gaugeIdsByDenom, 1, "perpetual %b", tc.isPerpetual)
+		suite.Require().Equal(gaugeID, gaugeIdsByDenom[0])
+
+		// check rewards estimation
+		rewardsEst := suite.App.IncentivesKeeper.GetRewardsEst(suite.Ctx, lockOwner, []lockuptypes.PeriodLock{}, 100)
+		suite.Require().Equal(coins.String(), rewardsEst.String())
```
AlpinYukseloglu, line 99: I wonder if directly using `coins` here is appropriate given that it represents all the tokens in the gauge. Unless I'm missing something, the only reason this should be passing is because `SetupLockAndGauge` only sets up one  one lock (i.e. one account to distributes rewards to), and `coins / 1` = `coins` so it doesn't error. We might be able to increase coverage in general by setting this whole set of tests up with multiple gauges (or maybe even adding a "number of gauges" field to each test case).

The multi-lock case was originally covered in the non-perpetual gauge test, which set up many locks using `SetupManyLocks` and then directly filled the gauge using `SetupNewGauge` (see below). I'll leave it up to you to decide whether that kind of an implementation makes sense for the TDD version of these tests, but we should definitely be testing the scenario where multiple locks are being distributed to (for context, the correct expected value here would be `coins / n` where n is the number of locks).

The original implementation of this was inconsistent across perpetual and non-perp gauges so this is a completely understandable issue.

https://github.com/osmosis-labs/osmosis/blob/8f018a6c638b64912c1b22c7df04f84abe667ba0/x/incentives/keeper/gauge_test.go#L57

https://github.com/osmosis-labs/osmosis/blob/8f018a6c638b64912c1b22c7df04f84abe667ba0/x/incentives/keeper/gauge_test.go#L74

x/incentives/keeper/gauge_test.go
```diff
@@ -280,33 +114,92 @@ func (suite *KeeperTestSuite) TestGaugesByDenom() {
...
+			suite.Require().Len(gauges, 0)
+
+			// check rewards estimation
+			rewardsEst = suite.App.IncentivesKeeper.GetRewardsEst(suite.Ctx, lockOwner, []lockuptypes.PeriodLock{}, 100)
+			suite.Require().Equal(sdk.Coins(nil), rewardsEst)
+		}
+
 		expectedNumGauges := 1
-		if !isPerpetual {
+		if !tc.isPerpetual {
```
AlpinYukseloglu, line 197: Since we now have separate perpetual/non-perpetual blocks, why not just move these instances of `expectedNumGauges` to the end of each of those?

e.g. put `expectedNumGauges := 1` at the end of the `if tc.isPerpetual {...}` section and `expectedNumGauges := 0` at the end of the `if !tc.isPerpetual {...}` section

This would also have the added benefit of reducing our # of if statements to two, one for perp gauges and one for non-perp, which imo improves code readability a lot

AlpinYukseloglu (changes requested):

AlpinYukseloglu:
Fantastic work, Richard. I left some suggestions in places where I saw inconsistencies between the old tests and the new ones, but overall this test restructuring looks great. Generally agree with the rest of the comments/discussion as well so I tried not to repeat the stuff mentioned there

AlpinYukseloglu:
Also cc: @czarcas7ic since this relates to the incentives module audit

x/incentives/keeper/gauge_test.go
rrrliu: Great catch, ended up adding `numLocks` as a column for our TDD

x/incentives/keeper/gauge_test.go
rrrliu: Makes sense -- the additional if statement just adds more confusion.

I made the change (putting the declaration outside the if/else blocks because of variable scope), but it might just be cleaner to put the denom checks directly at the end of both blocks? Lmk if you feel strongly either way

x/incentives/keeper/suite_test.go
```diff
@@ -41,8 +41,6 @@ func (suite *KeeperTestSuite) setupAddr(addrNum int, prefix string, balance sdk.
 		prefixBz := make([]byte, 8)
 		_, _ = rand.Read(prefixBz)
 		prefix = string(prefixBz)
-	} else {
-		prefix = fmt.Sprintf("%8.8s", prefix)
```
rrrliu, line 45: I removed this because the width and precision format specifier options for %s are based on unicode runes, not bytes. So essentially it was causing a strange bug where it would append to a certain []byte, bumping it from 8 bytes to 9 bytes to get it up to 8 runes. Then, the assert for lock creation would [fail](https://github.com/osmosis-labs/osmosis/runs/7176310153?check_suite_focus=true#step:6:56), because it would be a 21 byte address instead of 20.

If we do actually mean to make this exactly 8 unicode runes, I can change the assert statement when we create locks to check for rune length, not byte/string length.

More context: https://stackoverflow.com/a/60681952

x/incentives/keeper/gauge_test.go
AlpinYukseloglu: Denom checks at the end of both blocks would be the most appropriate in this case imo

AlpinYukseloglu (approved):

x/incentives/keeper/suite_test.go
ValarDragon: Thanks for fixing this! I'm pretty sure its some copy/paste issue from someone pulling something of stackoverflow and not looking into the details

mattverse (approved):
Nice fixes! Thanks for addressing the comments in so much detail! LGTM!

@rrrliu  Can you fix the merge conflicts? Once merge conflicts are fixed, we can go ahead and merge!

rrrliu:
@mattverse conflicts are fixed!

mattverse:
Sweet! Going ahead and merging this!

Issue #1838, opened 2022-06-20
x/lockup - internal audit


ValarDragon:
To be done in three seperate PR
- [ ] #1828 
- [ ] https://github.com/osmosis-labs/osmosis/pull/1900
- [ ] https://github.com/osmosis-labs/osmosis/pull/1936
- [ ] https://github.com/osmosis-labs/osmosis/pull/1937
- [ ] https://github.com/osmosis-labs/osmosis/pull/1951
- [ ] #1968 
- [ ] https://github.com/osmosis-labs/osmosis/pull/1979

mattverse:
To be done in three seperate PR

- [ ] One PR for deleting unused methods, renaming methods, and the documentations cleanup
- [ ] One PR for refactoring `Lock` method.
- [ ] One PR for ensuring API safety for `ExtendLock`, where we used to take lock of pointer references as parameter. This is bad for ensuring API safety, now it takes in lockID as parameter.

mattverse:
All the other components are finished as well in the branch `mattverse/audit-lockup`, but holding on those until #1900 gets closed since likely to cause too much merge conflicts

PR #1936, opened 2022-07-01
Refactor `lock` method

mattverse:
Sub component of: #1838 

The lock method in the lockup module was a public method before current change, which was an imo dangerous design, as it is and should be only called by either of the two methods
- `CreateLock`
- `AddTokensToLockByID`

That being said, there shouldn't be any cases where the `Lock` method is called directly, but called only through the two entrypoints mentioned above, not to mention the code duplication that we had for the two methods above. 

This PR changes the Lock method into a private method and refactors code duplication logic as well.

## Brief Changelog
- API change for the `Lock` method in x/lockup module.

### DIFF
--- a/x/lockup/keeper/lock.go
+++ b/x/lockup/keeper/lock.go
@@ -14,29 +14,14 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/x/lockup/types"
 )
 
-// TODO: Reorganize functions in this file
-
 // WithdrawAllMaturedLocks withdraws every lock thats in the process of unlocking, and has finished unlocking by
 // the current block time.
 func (k Keeper) WithdrawAllMaturedLocks(ctx sdk.Context) {
 	k.unlockFromIterator(ctx, k.LockIteratorBeforeTime(ctx, ctx.BlockTime()))
 }
 
-func (k Keeper) getCoinsFromLocks(locks []types.PeriodLock) sdk.Coins {
-	coins := sdk.Coins{}
-	for _, lock := range locks {
-		coins = coins.Add(lock.Coins...)
-	}
-	return coins
-}
-
-func (k Keeper) accumulationStore(ctx sdk.Context, denom string) store.Tree {
-	return store.NewTree(prefix.NewStore(ctx.KVStore(k.storeKey), accumulationStorePrefix(denom)), 10)
-}
-
-// GetModuleBalance Returns full balance of the module.
+// GetModuleBalance returns full balance of the module.
 func (k Keeper) GetModuleBalance(ctx sdk.Context) sdk.Coins {
-	// TODO: should add invariant test for module balance and lock items
 	acc := k.ak.GetModuleAccount(ctx, types.ModuleName)
 	return k.bk.GetAllBalances(ctx, acc.GetAddress())
 }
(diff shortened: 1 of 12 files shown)


x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
-	if err != nil {
-		return nil, err
+	for _, synthlock := range k.GetAllSyntheticLockupsByLockup(ctx, lock.ID) {
+		k.accumulationStore(ctx, synthlock.SynthDenom).Increase(accumulationKey(synthlock.Duration), tokensToAdd.Amount)
 	}
 
 	if k.hooks == nil {
 		return lock, nil
 	}
+	k.hooks.AfterAddTokensToLock(ctx, lock.OwnerAddress(), lock.GetID(), sdk.NewCoins(tokensToAdd))
```
alexanderbez, line 92:
```suggestion

	k.hooks.AfterAddTokensToLock(ctx, lock.OwnerAddress(), lock.GetID(), sdk.NewCoins(tokensToAdd))
```

alexanderbez (approved):

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
 // Tokens locked are sent and kept in the module account.
 // This method alters the lock state in store, thus we do a sanity check to ensure
 // lock owner matches the given owner.
-func (k Keeper) AddTokensToLockByID(ctx sdk.Context, lockID uint64, owner sdk.AccAddress, coin sdk.Coin) (*types.PeriodLock, error) {
+func (k Keeper) AddTokensToLockByID(ctx sdk.Context, lockID uint64, owner sdk.AccAddress, tokensToAdd sdk.Coin) (*types.PeriodLock, error) {
 	lock, err := k.GetLockByID(ctx, lockID)
+	if err != nil {
+		return nil, err
+	}
```
ValarDragon, line 73: huh, why did the `errcheck` linter not catch this before. Its enabled?

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
 	if err != nil {
 		return nil, err
 	}
-	if err := k.bk.SendCoinsFromAccountToModule(ctx, lock.OwnerAddress(), types.ModuleName, sdk.NewCoins(coin)); err != nil {
-		return nil, err
-	}
 
-	err = k.addTokenToLock(ctx, lock, coin)
-	if err != nil {
-		return nil, err
```
ValarDragon, line 85: This looks wrong?  addTokenToLock did more, e.g. handling of the main accumulation store? Why was this deleted?

ValarDragon: Oh the claim is that all functionality is there between the synth lock line, and lock.

Why not keep `addTokenToLock` and just have that function as a subcomponent call `lock`? (But retain the synth lock functionality)

That would be maintaining the standard in this package of having the exported method be doing validation logic & hook / return type logic, and the private method doing the state change

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
-	store := ctx.KVStore(k.storeKey)
-	bz, err := proto.Marshal(&lock)
-	if err != nil {
-		return err
-	}
-	store.Set(lockStoreKey(lock.ID), bz)
-
-	// add lock refs into not unlocking queue
-	err = k.addLockRefs(ctx, lock)
+	err = k.setLock(ctx, lock)
```
ValarDragon, line 132: ty!

x/lockup/keeper/lock.go
p0mvn: `golangci-lint linters` shows it as enabled. Maybe a bug in the linter?

x/lockup/keeper/lock_test.go
```diff
@@ -144,47 +144,17 @@ func (suite *KeeperTestSuite) TestUnlockPeriodLockByID() {
 	suite.Require().Len(locks, 0)
 }
 
-func (suite *KeeperTestSuite) TestLock() {
```
p0mvn, line 147: Does it make sense to do either or both:
-  test `CreateLock`
- use the [`export_test.go`](https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd) trick that will allow us to
    * export `lock` for tests only and
    * keep testing the now unexported `lock` method

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
+	if err != nil {
+		return nil, err
+	}
 
 	if lock.GetOwner() != owner.String() {
 		return nil, types.ErrNotLockOwner
 	}
 
+	lock.Coins = lock.Coins.Add(tokensToAdd)
+	err = k.lock(ctx, *lock, sdk.NewCoins(tokensToAdd))
```
AlpinYukseloglu, line 80: Out of curiosity, why do we use `sdk.NewCoins(tokensToAdd)` here instead of directly passing `tokensToAdd` in like the `Lock` function does?

x/lockup/keeper/lock.go
AlpinYukseloglu: I still don't fully follow the logic here. There are two calls that `addTokenToLock` makes that are not related to the synth lock related accumulation store update, which are:
1. `setLock` to update state with the new lock instance:
https://github.com/osmosis-labs/osmosis/blob/8f018a6c638b64912c1b22c7df04f84abe667ba0/x/lockup/keeper/lock.go#L100
2. Update to `accumulationStore`:
https://github.com/osmosis-labs/osmosis/blob/8f018a6c638b64912c1b22c7df04f84abe667ba0/x/lockup/keeper/lock.go#L106

How/where are these being accounted for in the new implementation?

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
 	if err != nil {
 		return nil, err
 	}
-	if err := k.bk.SendCoinsFromAccountToModule(ctx, lock.OwnerAddress(), types.ModuleName, sdk.NewCoins(coin)); err != nil {
-		return nil, err
-	}
```
AlpinYukseloglu, line 81: Just to clarify, this was removed because we're now calling `lock` above, which already sends new locked tokens to the module account?

x/lockup/keeper/lock.go
AlpinYukseloglu: Oops, I see the `lock` call above now. The change that adds an addition parameter to `lock` also makes sense now. Clever refactor :)

x/lockup/keeper/lock.go
```diff
@@ -66,100 +66,76 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
 
-	err = k.addTokenToLock(ctx, lock, coin)
-	if err != nil {
-		return nil, err
+	for _, synthlock := range k.GetAllSyntheticLockupsByLockup(ctx, lock.ID) {
+		k.accumulationStore(ctx, synthlock.SynthDenom).Increase(accumulationKey(synthlock.Duration), tokensToAdd.Amount)
 	}
```
AlpinYukseloglu, line 87: Is it okay that we're no longer doing the single coin check that was originally in `addTokenToLock`? Is it checked elsewhere or was it not necessary to begin with?
https://github.com/osmosis-labs/osmosis/blob/8f018a6c638b64912c1b22c7df04f84abe667ba0/x/lockup/keeper/lock.go#L108-L111

AlpinYukseloglu (commented):
Left a few clarifying questions but overall LGTM! Love this refactor, cleans things up quite a bit :)

x/lockup/keeper/lock.go
mattverse: oh its just because we weren't using the `lock` method previosuly!

x/lockup/keeper/lock.go
mattverse: we're using a pointer reference here, editing the lock directly makes it impossible to track the newly added tokens to the lock, which is needed for accumulation store, lock refs etc!

x/lockup/keeper/lock.go
mattverse: yeah!

x/lockup/keeper/lock.go
mattverse: @ValarDragon  wdym by ?

> (But retain the synth lock functionality)

x/lockup/keeper/lock.go
mattverse: this has been checked in the ValidateBasic in the msg server!

x/lockup/keeper/lock_test.go
mattverse: Ah missed the test for `CreateLock` thanks for catching this! Just added this in the most recent commit!

AlpinYukseloglu (approved):

mattverse:
merging this upon being blocker for future work for adding testing for lockup!

mattverse:
Apologies if I went ahead and rushed into merging this, I supposed there weren't extra reviews.

Is there something you're concerned about or would like to give a review on? We can either revert this commit or create a new PR depending on the review

mattverse:
A Note to reviewers to clarify no logic has been removed / changed per this PR

**Previous logic for `AddTokensToLockByID`**
- SendCoinsFromAccountToModule
- addTokenToLock
     - lock.Coins = lock.Coins.Add(coin)
     - setLock
     - accumulation store increase
     - synth denom accumulation store increase
     - hooks.AfterAddTokensToLock
- hooks.OnTokenLocked

**Refactored Logic for `AddTokensToLockByID`**
- lock.Coins = lock.Coins.Add(coin)
- lock
   - SendCoinsFromAccountToModule
   - setLock
   - accumulation store increase
   - hooks.OnTokenLocked 
 - synth denom accumulation store increase
 -  hooks.AfterAddTokensToLock
 
---------
 **Previous logic for `CreateLock`**
 - Lock
   - SendCoinsFromAccountToModule
   - store.Set(lock)
   - addlockRefs
   - accumulation store increase
   - hooks.OnTokenLocked
  - setLastLockID
  
  **Refactored Logic for `CreateLock`**
  - lock 
      - SendCoinsFromAccountToModule
     - setLock
     - accumulation store increase
      - hooks.OnTokenLocked  
   - addLockRefs
   - SetLastLockID

Issue #1943, opened 2022-07-01
x/gamm: refactor all pool_service.go tests to be properly table-driven


AlpinYukseloglu:
## Background

All of the current tests in `pool_service.go`, while technically structured as table-driven tests, have complex nested "setup" functions that defeat the purpose of structuring them in a tabular way. We should refactor all of these to be properly table-driven by grouping test cases by logic and setting them up in such a way that none of the test cases have unique complex logic in the form of a nested function.

## Suggested Design

- Group test cases by similar logic and factor out the nested input functions

## Acceptance Criteria

- All test cases work correctly and pass

PR #1949, opened 2022-07-03
x/gamm: refactor all pool_service.go tests to be properly table-driven

AlpinYukseloglu:
Closes: #1943 

All of the current tests in pool_service.go, while technically structured as table-driven tests, have complex nested "setup" functions that defeat the purpose of structuring them in a tabular way. This PR refactors all of these to be properly table-driven by grouping test cases by logic and setting them up in such a way that none of the test cases have unique complex logic in the form of a nested function.

## Brief Changelog

1. Remove the `fn` function from the `TestCreateBalancerPool` test struct and replace it with `name`, `msg` (of type `MsgCreateBalancerPool`), and `expectPass` fields
2. Move all relevant test logic that existed in the nested function into the test case for loop
3. Create new `TestPoolCreationFee` tests to isolate `PoolCreationFee` related tests `TestCreateBalancerPool`
4. Refactor `TestJoinPoolNoSwap` by replacing `fn` field with `JoinPoolNoSwap` inputs and factoring out common logic
5. Refactor `TestExitPool` by replacing `fn` field with `ExitPool` inputs and factoring out common logic

## Testing and Verifying


x/gamm/keeper/pool_service_test.go
```diff
@@ -61,73 +61,38 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
 	// (And should be moved to balancer package)
```
AlpinYukseloglu, line 61: @ValarDragon what did you mean by this comment?

x/gamm/keeper/pool_service_test.go
```diff
@@ -41,93 +41,53 @@ var (
...
-		_, err := keeper.CreatePool(suite.Ctx, msg)
-		suite.Require().Error(err)
-	}()
-
-	// TODO: Refactor this to be more sensible.
-	// The struct should contain a MsgCreateBalancerPool.
-	// then the scaffolding should test pool creation, and check if it was a success or not.
-	// (And should be moved to balancer package)
-	// PoolCreationFee tests should get their own isolated test in this package.
+	// TODO: should be moved to balancer package
```
AlpinYukseloglu, line 53: @ValarDragon what did you mean by this comment?

x/gamm/keeper/pool_service_test.go
```diff
@@ -203,241 +151,347 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+
+		// fund sender test account
+		sender, err := sdk.AccAddressFromBech32(test.msg.Sender)
+		suite.Require().NoError(err, "test: %v", test.name)
+		if sender.String() != emptySender.String() {
+			suite.FundAcc(sender, defaultAcctFunds)
+		}
+
+		// note starting balances for community fee pool and pool creator account
+		feePoolBeforeNewPool := suite.App.DistrKeeper.GetFeePoolCommunityCoins(suite.Ctx)
```
mattverse, line 171: nit: consider changing to:

```suggestion
		feePoolBalBeforeNewPool := suite.App.DistrKeeper.GetFeePoolCommunityCoins(suite.Ctx)
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -203,241 +151,347 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		poolCreationFeeDecCoins = poolCreationFeeDecCoins.Add(sdk.NewDecCoin(coin.Denom, coin.Amount))
+	}
+
+	tests := []struct {
+		name            string
+		poolCreationFee sdk.Coins
+		msg             balancertypes.MsgCreateBalancerPool
+		expectPass      bool
+	}{
+		{
```
mattverse, line 227: can we add test cases that suffices `expectPass` false?

x/gamm/keeper/pool_service_test.go
```diff
@@ -203,241 +151,347 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		// Create the pool at first
+		msg := balancer.NewMsgCreateBalancerPool(test.txSender, balancer.PoolParams{
+			SwapFee: sdk.NewDecWithPrec(1, 2),
+			ExitFee: sdk.NewDec(0),
+		}, defaultPoolAssets, defaultFutureGovernor)
+		poolId, err := suite.App.GAMMKeeper.CreatePool(suite.Ctx, msg)
+
+		// If we are testing insufficient pool share balances, switch sender to empty account
+		if test.txSender.String() == emptySender.String() {
+			test.txSender = suite.TestAccs[2]
```
mattverse, line 475: What's ther reason for changing emptySender from suite.TestAccs[1] to suite.TestAccs[2]? From my understanding, didn't we reserve suite.TestAccs[1] for the empty sender?

mattverse (commented):
Overall looks nice and lgtm! 

Left some minor comments, please take a look!

x/gamm/keeper/pool_service_test.go
AlpinYukseloglu: I think this was a hacky approach to begin with – I refactored in recent commits to have an `emptySender` field for tests that need it.

For this specific case, we cannot use the same account we used to create the pool since that one has pool shares in it. If we create the pool with a completely independent account, then we would need to find and transfer the gamm shares for every other test, which felt like a messier approach than just switching the tx sender to an empty account when we want to test insufficient shares

x/gamm/keeper/pool_service_test.go
AlpinYukseloglu: The original version of this was just `feePool` so I felt like `feePoolBeforeNewPool` was already pretty verbose. If you feel like this would have made it more readable for you then definitely happy to change

mattverse: yeah would prefer changing to keep consistency with `senderBalBefore`! Thnx!

x/gamm/keeper/pool_service_test.go
mattverse: ah makes sense, can we write this down in the comment here that we're using a third account here because we can't use the same account we used to create pool?

AlpinYukseloglu: of course, I expanded the existing comment in my recent commit!

x/gamm/keeper/pool_service_test.go
AlpinYukseloglu: sounds good, just changed! (couldn't commit your suggestion directly since it would break things hope that's okay)

x/gamm/keeper/pool_service_test.go
ValarDragon: That it should be moved to the balancer package?

(Seeing as its a create Balancer pool message =p )

AlpinYukseloglu:
Accidental close – merge conflicts should be fixed now!

alexanderbez (approved):

Issue #1897, opened 2022-06-29
proposal: enable staticcheck


p0mvn:
## Background

I noticed that our linters do not catch unused variables. Upon running `staticcheck` with:
```
staticcheck $(go list ./...)
```
I discovered around 30 lint issues related to unused code. We should consider enabling `staticcheck` as this is a useful static analysis tool.

## Suggested Design

We can enable it via `golanglint-ci` without much effort:
https://github.com/osmosis-labs/osmosis/blob/a785de13eebcfd91b6807458fbcf8ffaff105bb5/.golangci.yml#L71

## Acceptance Criteria

- `staticcheck` is enabled
- all dead code / unused variables are removed

ValarDragon:
IIRC, staticcheck is super slow

p0mvn:
You're right. It's slow. Took 5 seconds on the first pass. Reused cache a second time, taking approximately 1 second.

To mitigate, what do you think about the following:
- adding a makefile step that is separate from our main `golangci-lint` linter i.e `make staticcheck`
   * This should only be used if someone sees CI failing and wants to reproduce locally
   * Our regular linting is done separately by `golangci-lint` where it has `staticcheck` disabled
- separate CI job for `staticcheck` that runs in parallel with other jobs
    * Our bottlenecks are tests, both e2e and unit so 5 seconds should not be a problem
- Disabling noise like deprecation linter and anything else we might see later

I just shot myself in the foot in https://github.com/osmosis-labs/osmosis/pull/1857 by leaving a lot of unused code. Since we tend to be working on many PRs concurrently while waiting on feedback, it's easy to miss things like this when context switching.

`staticcheck` in CI should be able to automate away this cognitive overhead for both the PR author and the reviewer.

ValarDragon:
Why not just add the deadcode linter?

p0mvn:
I've been framing this in terms of the dead code detection benefit so far because of recency bias.

However, `staticcheck` is also useful for [the variety of other checks](https://staticcheck.io/docs/checks/) it supports, helping spot erroneous errors and maintain code standards. It would also help reduce manual linting in PRs.

With only a quick skim, these are the linters that I've seen to be manually pointed out in our reviews that can be covered by `staticcheck` instead:
- Simplify error construction with fmt.Errorf - https://staticcheck.io/docs/checks/#S1028
- Merge variable declaration and assignment - https://staticcheck.io/docs/checks/#S1021
[...]

PR #1956, opened 2022-07-04
lint: enable staticcheck

p0mvn:
Closes: #XXX

Follow-up to: https://github.com/osmosis-labs/osmosis/issues/1897

Please note that this is a `staticcheck` set of rules (a subset of rules from `staticcheck` binary

## Brief Changelog

- Removed variables that were not being reused after assignment
- Handled errors from `defer Close()`
   * More context: https://www.joeshaw.org/dont-defer-close-on-writable-files/
- Swapped deprecated `GenerateSaveCoinKey`
- Removed redundant check against 0 for `uint`s

## Testing and Verifying


x/lockup/types/genesis.go
```diff
@@ -13,8 +11,5 @@ func DefaultGenesis() *GenesisState {
 // Validate performs basic genesis state validation returning an error upon any
 // failure.
 func (gs GenesisState) Validate() error {
-	if gs.LastLockId < 0 {
```
alexanderbez, line 16: Why was this removed?

x/incentives/types/genesis.go
```diff
@@ -45,8 +45,5 @@ func (gs GenesisState) Validate() error {
 	if gs.Params.DistrEpochIdentifier == "" {
 		return errors.New("epoch identifier should NOT be empty")
 	}
-	if gs.LastGaugeId < 0 {
```
alexanderbez, line 48: ditto

cmd/osmosisd/cmd/balances_from_state_export.go
```diff
@@ -179,11 +179,6 @@ Example:
 
 			authGenesis := authtypes.GenesisState{}
 			clientCtx.Codec.MustUnmarshalJSON(genState["auth"], &authGenesis)
-			accounts, err := authtypes.UnpackAccounts(authGenesis.Accounts)
```
alexanderbez, line 182: Was this removed on purpose?

x/lockup/types/genesis.go
p0mvn: This is a `uint` so it cannot be less than 0

x/incentives/types/genesis.go
p0mvn: Same reason - `uint`

cmd/osmosisd/cmd/balances_from_state_export.go
p0mvn: Yes, this was unused other than unpacking

p0mvn: I just double-checked, and I think the two lines above can also be removed:
```
authGenesis := authtypes.GenesisState{}
clientCtx.Codec.MustUnmarshalJSON(genState["auth"], &authGenesis)
```

Generally, I'm having difficulty understanding what's happening in that command. We might want to consider having an issue to determine what is the need for it and potentially refactor

x/epochs/handler.go
```diff
@@ -13,8 +13,6 @@ import (
 // NewHandler returns a handler for epochs module messages.
 func NewHandler(k keeper.Keeper) sdk.Handler {
 	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
-		ctx = ctx.WithEventManager(sdk.NewEventManager())
```
alexanderbez, line 16: IMO we shouldn't even have these `handler.go` files -- it's all legacy code.

x/incentives/types/genesis.go
alexanderbez: ACK

alexanderbez (approved):

x/epochs/handler.go
p0mvn: I'm guessing this: https://github.com/cosmos/cosmos-sdk/pull/9650 would be the right way to do it?

cmd/osmosisd/cmd/forceprune.go
```diff
@@ -128,13 +139,21 @@ func compactBlockStore(dbPath string) error {
...
 	if err = db.CompactRange(*util.BytesPrefix([]byte{})); err != nil {
 		return err
 	}
+	// N.B: We duplicate the call to db.Close() on top of
+	// the call in defer statement above to make sure that the resources
+	// are properly released and any potential error from Close()
+	// is handled. Close() should be idempotent so this is acceptable.
+	if err := db.Close(); err != nil {
+		return err
+	}
```
ValarDragon, line 156: I don't get why are we doing this? The defer handles this more cleanly?

p0mvn: If we do `defer db.Close()`, it does not do error handling.

The new change validates the error while keeping `defer db.Close()` as well. The code in `defer` runs even if the code panics, making sure that we always close the resources. Calling this twice should not be a problem because `Close()` is idempotent

ValarDragon: Can we instead do a named return error, (e.g. `func compactBlockStore(dbPath string) (rerr error)`, and then do `defer rerr = db.Close()`. This is the general pattern for handling errors in a panic.

Encouraging db closing being isolated scope / not handled multiple times is ideal imo.

ValarDragon (approved):
LGTM after we resolve the db.Close

p0mvn:
Merging since db.Close addressed + 2 approvals

PR #1960, opened 2022-07-04
lint: enable stylecheck

p0mvn:
Closes: #XXX

Follow-up to: https://github.com/osmosis-labs/osmosis/issues/1897

Linter that enforces style rules. A subset of rules from the staticcheck binary.

## Brief Changelog

- disable annoying check requiring "Id" to be "ID"
- ensure that receivers are named consistently
- remove duplicate imports that had different identifiers
- remove redundant types in var declarations
- errors messages are not capitalized
- godoc starts with var name

## Testing and Verifying

### DIFF
--- a/x/gamm/pool-models/stableswap/pool.go
+++ b/x/gamm/pool-models/stableswap/pool.go
@@ -15,64 +15,64 @@ import (
 
 var _ types.PoolI = &Pool{}
 
-func (pa Pool) GetAddress() sdk.AccAddress {
-	addr, err := sdk.AccAddressFromBech32(pa.Address)
+func (p Pool) GetAddress() sdk.AccAddress {
+	addr, err := sdk.AccAddressFromBech32(p.Address)
 	if err != nil {
-		panic(fmt.Sprintf("could not bech32 decode address of pool with id: %d", pa.GetId()))
+		panic(fmt.Sprintf("could not bech32 decode address of pool with id: %d", p.GetId()))
 	}
 	return addr
 }
 
-func (pa Pool) String() string {
-	out, err := json.Marshal(pa)
+func (p Pool) String() string {
+	out, err := json.Marshal(p)
 	if err != nil {
 		panic(err)
 	}
 	return string(out)
 }
 
-func (pa Pool) GetId() uint64 {
-	return pa.Id
+func (p Pool) GetId() uint64 {
+	return p.Id
 }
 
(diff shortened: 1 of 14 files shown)


p0mvn:
Anecdotally, it feels that this linter caused the largest increase in time it takes to run `golangci-lint run`

However, it does seem that it was helpful in enforcing consistency so might still be worth trying it.

I think it should be another candidate to be potentially removed if linter times become a problem. Instead, we can run it every now and then to clean up the codebase

p0mvn:
> Anecdotally, it feels that this linter caused the largest increase in time it takes to run `golangci-lint run`
> 
> However, it does seem that it was helpful in enforcing consistency so might still be worth trying it.
> 
> I think it should be another candidate to be potentially removed if linter times become a problem. Instead, we can run it every now and then to clean up the codebase

On another look, seems I was wrong - the difference was likely due to cache. After timing these again:
```
git checkout main
time golangci-lint run

real    0m1.441s
user    0m5.352s
sys     0m1.170s
```
```
git checkout roman/lint-stylecheck
time golangci-lint run

real    0m1.465s
user    0m5.423s
sys     0m1.231s
```

Going to leave this here still in case anyone has thoughts

alexanderbez (approved):

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -196,59 +196,59 @@ func (pa Pool) CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDen
 	return sdk.NewCoin(tokenInDenom, tokenInAmt), nil
 }
 
-func (pa *Pool) SwapInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.Coin, err error) {
-	tokenIn, err = pa.CalcInAmtGivenOut(ctx, tokenOut, tokenInDenom, swapFee)
+func (p *Pool) SwapInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.Coin, err error) {
```
mattverse, line 199: yay pa -> p !

mattverse (approved):

mattverse:
doesn't seem like a big increase in time / a big overhead just from the stats you provided?

Issue #1879, opened 2022-06-27
e2e: enable state-sync


p0mvn:
## Background

We would like to be testing state-sync in e2e. To make it work, a refactor needs to be done.

It has been started in:

- https://github.com/osmosis-labs/osmosis/pull/1844
- https://github.com/osmosis-labs/osmosis/pull/1824
- https://github.com/osmosis-labs/osmosis/pull/1591
- [`roman/e2e-chaininit-refactor`](https://github.com/osmosis-labs/osmosis/tree/roman/e2e-chaininit-refactor)
- https://github.com/osmosis-labs/osmosis/pull/1587
- https://github.com/osmosis-labs/osmosis/pull/1572

These need to be rebased, split into smaller PRs where applicable and merged.

## Acceptance Criteria
- state-sync works in e2e

PR #1963, opened 2022-07-04
e2e: refactor initialization with single node logic

p0mvn:
Part of: #1879 

Building up on https://github.com/osmosis-labs/osmosis/pull/1898

This refactor separates the initialization logic into 2 parts:
- `initialization.InitChain()` - for spinning up a chain with several validators
- `initialization.InitSingleNode()` - for spinning up a single node (e.g. for testing state-sync)

The documentation for this is already merged and available here: https://github.com/osmosis-labs/osmosis/tree/main/tests/e2e/initialization

## Testing and Verifying


tests/e2e/initialization/init.go
```diff
@@ -0,0 +1,75 @@
...
+				return nil, err
+			}
+		}
+	}
+	return chain.export(), nil
+}
+
+func InitSingleNode(chainId, dataDir string, existingGenesisDir string, nodeConfig *NodeConfig, votingPeriod time.Duration, trustHeight int64, trustHash string, stateSyncRPCServers []string) (*Node, error) {
+	if nodeConfig.IsValidator {
+		return nil, errors.New("creating individual validator nodes after starting up chain is not currently supported")
```
czarcas7ic, line 49: Should we make an issue for this?

tests/e2e/initialization/init_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		dataDir, err = ioutil.TempDir("", "osmosis-e2e-testnet-test")
+	)
+
+	// Setup
+	existingChain, err := initialization.InitChain(id, dataDir, existingChainNodeConfigs, time.Second*3, forkHeight)
+	require.NoError(t, err)
+
+	actualNode, err := initialization.InitSingleNode(existingChain.ChainMeta.Id, dataDir, filepath.Join(existingChain.Nodes[0].ConfigDir, "config", "genesis.json"), expectedConfig, time.Second*3, 3, "testHash", []string{"some server"})
+	require.NoError(t, err)
```
czarcas7ic, line 116: Do you think we should also check that the actual genesis.json file itself is the same between the single node and a randomly selected node from the chain, or do you think this is a good enough check already?

tests/e2e/initialization/node.go
```diff
@@ -256,22 +290,72 @@ func createMnemonic() (string, error) {
 	return mnemonic, nil
 }
 
+func (n *internalNode) initValidatorConfigs(c *internalChain, persistendtPeers []string) error {
```
czarcas7ic, line 293: Was this supposed to be persistentPeers?

tests/e2e/initialization/init_test.go
```diff
@@ -0,0 +1,143 @@
...
+			IsValidator:        false,
+		}
+		dataDir, err = ioutil.TempDir("", "osmosis-e2e-testnet-test")
+	)
+
+	// Setup
+	existingChain, err := initialization.InitChain(id, dataDir, existingChainNodeConfigs, time.Second*3, forkHeight)
+	require.NoError(t, err)
+
+	actualNode, err := initialization.InitSingleNode(existingChain.ChainMeta.Id, dataDir, filepath.Join(existingChain.Nodes[0].ConfigDir, "config", "genesis.json"), expectedConfig, time.Second*3, 3, "testHash", []string{"some server"})
```
czarcas7ic, line 115: I assume the next step here is automating pulling a specified height hash from one of the validator's RPCs? Shouldn't we have enough here already to start filling in the RPC server address? IMO we should just fill this in with a single RPC twice ("1.23.45.67:26657,1.23.45.67:26657"), just makes it simpler to debug. Unless you feel we should be testing with two separate RPCs

czarcas7ic (approved):
Incredible job with this! Was honestly really easy to follow despite the size. I say we merge asap and can address smaller bugs that may come up through watching it run on prod as we discussed in todays meeting

tests/e2e/initialization/init.go
p0mvn: I think we can address this when there is a need for it in the future. Currently, for testing state sync we should be able to use a regular full node instead of a validator.

I don't think there's is a current use case for us to have the validator nodes

tests/e2e/initialization/init_test.go
p0mvn: I think this would be necessary to have if this was the actual system but since we are "testing a test", what we have now is sufficient.

The goal for this test was to just make sure the expected initialization files are created because I was finding myself doing that manually a lot while refactoring

tests/e2e/initialization/init_test.go
p0mvn: You are right, the next step would be to start actually utilizing this initialization logic for spinning up a state-synching full node. I don't have that completed yet. Before I continue with this, there is one more large refactoring PR to merge.

And thanks for the advice, I will make sure to use the same one twice

Issue #1944, opened 2022-07-01
x/gamm: refactor all pool.go tests to be properly table-drive


AlpinYukseloglu:
## Background

Similar to #1943. These tests should not include unique test logic in case-specific functions.

## Suggested Design

- Group tests by similar logic and factor out case-specific functions

## Acceptance Criteria

- All tests work as intended and pass

PR #1969, opened 2022-07-05
x/gamm: refactor balancer pool_test.go tests to be properly table-driven

AlpinYukseloglu:
Closes: #1944 

Several tests in our balancer `pool_test.go` had setup functions as fields in test cases. These functions repeated code and overall were functionally not really "table-driven". This PR factors out the logic in these setup functions and replaces the function itself with relevant fields.

## Brief Changelog

- Remove setup function field from `TestUpdateIntermediaryPoolAssetsLiquidity` and replace it with `newLiquidity`,
`originalPoolAssets`, and `expectPass` fields
- Move logic for finding expected assets to test-wide for loop
- Clean up error checks using `expectPass`

## Testing and Verifying

### DIFF
--- a/x/gamm/pool-models/balancer/pool_test.go
+++ b/x/gamm/pool-models/balancer/pool_test.go
@@ -42,11 +42,11 @@ func TestUpdateIntermediaryPoolAssetsLiquidity(t *testing.T) {
 		weight = 100
 	)
 	testCases := []struct {
-		name               string
-		newLiquidity       sdk.Coins
-		originalPoolAssets map[string]balancer.PoolAsset
-		expectPass         bool
-		err                error
+		name         string
+		newLiquidity sdk.Coins
+		poolAssets   map[string]balancer.PoolAsset
+		expectPass   bool
+		err          error
 	}{
 		{
 			name: "regular case with multiple pool assets and a subset of newLiquidity to update",
@@ -54,7 +54,7 @@ func TestUpdateIntermediaryPoolAssetsLiquidity(t *testing.T) {
 				sdk.NewInt64Coin("uosmo", 1_000),
 				sdk.NewInt64Coin("atom", 2_000),
 				sdk.NewInt64Coin("ion", 3_000)),
-			originalPoolAssets: map[string]balancer.PoolAsset{
+			poolAssets: map[string]balancer.PoolAsset{
 				"uosmo": {
 					Token:  sdk.NewInt64Coin("uosmo", uosmoValueOriginal),
 					Weight: sdk.NewInt(weight),
@@ -73,7 +73,7 @@ func TestUpdateIntermediaryPoolAssetsLiquidity(t *testing.T) {
 		{
 			name:         "new liquidity has no coins",
 			newLiquidity: sdk.NewCoins(),
-			originalPoolAssets: map[string]balancer.PoolAsset{
+			poolAssets: map[string]balancer.PoolAsset{
 				"uosmo": {
 					Token:  sdk.NewInt64Coin("uosmo", uosmoValueOriginal),
 					Weight: sdk.NewInt(weight),
@@ -93,7 +93,7 @@ func TestUpdateIntermediaryPoolAssetsLiquidity(t *testing.T) {
 			name: "newLiquidity has a coin that poolAssets don't",
 			newLiquidity: sdk.NewCoins(
 				sdk.NewInt64Coin("juno", 1_000)),
-			originalPoolAssets: map[string]balancer.PoolAsset{
+			poolAssets: map[string]balancer.PoolAsset{
 				"uosmo": {
 					Token:  sdk.NewInt64Coin("uosmo", uosmoValueOriginal),
 					Weight: sdk.NewInt(weight),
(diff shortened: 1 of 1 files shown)


x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -33,131 +33,97 @@ var (
...
 		t.Run(tc.name, func(t *testing.T) {
-			newLiquidity, originalPoolAssetsByDenom, expectedPoolAssetsByDenom := tc.setup()
-
-			err := balancer.UpdateIntermediaryPoolAssetsLiquidity(newLiquidity, originalPoolAssetsByDenom)
+			expectedPoolAssetsByDenom := map[string]balancer.PoolAsset{}
+			for i, asset := range tc.originalPoolAssets {
+				expectedValue := balancer.PoolAsset{Token: asset.Token, Weight: asset.Weight}
+				expectedValue.Token.Amount = expectedValue.Token.Amount.Add(tc.newLiquidity.AmountOf(i))
+				expectedPoolAssetsByDenom[i] = expectedValue
+			}
```
p0mvn, line 114: This suggestion is twofold:
- we are iterating over a map so "i" feels confusing because this identifier is a convention for counters in a loop, not map keys. Suggesting to change to either `denom` or `key`
- `expectedValue := asset` is a copy by value so we do not need to initialize a new `balancer.PoolAsset{...}` directly, can copy instead to simplify

```suggestion
			for denom, asset := range tc.originalPoolAssets {
				expectedValue := asset
				expectedValue.Token.Amount = expectedValue.Token.Amount.Add(tc.newLiquidity.AmountOf(denom))
				expectedPoolAssetsByDenom[denom] = expectedValue
			}
```

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -33,131 +33,97 @@ var (
...
 
-			require.Equal(t, tc.err, err)
+			err := balancer.UpdateIntermediaryPoolAssetsLiquidity(tc.newLiquidity, tc.originalPoolAssets)
 
-			if tc.err != nil {
-				return
+			if tc.expectPass {
+				require.NoError(t, tc.err, "test: %v", tc.name)
```
p0mvn, line 119: nit: consider asserting on what error specifically is returned

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -33,131 +33,97 @@ var (
...
 func TestUpdateIntermediaryPoolAssetsLiquidity(t *testing.T) {
-	testCases := []struct {
-		name string
-
-		// returns newLiquidity, originalPoolAssetsByDenom, expectedPoolAssetsByDenom
-		setup func() (sdk.Coins, map[string]balancer.PoolAsset, map[string]balancer.PoolAsset)
+	const (
+		uosmoValueOriginal = 1_000_000_000_000
+		atomValueOriginal  = 123
+		ionValueOriginal   = 657
```
p0mvn, line 39: Do you think it makes sense to have one test case with initial pool assets other than these?

p0mvn (approved):
LGTM, Looks much cleaner, nice work. Only had a few nit suggestions

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: I added `require.Equal(t, tc.err, err)` in the case where we expect an error (a few lines down) as a defacto assertion – does that suffice or do you think we should add something even more explicit?

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: Great points

x/gamm/pool-models/balancer/pool_test.go
p0mvn: I see, I missed that line. Thanks for clarifying

alexanderbez (approved):

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -33,131 +33,97 @@ var (
...
 
-			if tc.err != nil {
-				return
+			if tc.expectPass {
+				require.NoError(t, tc.err, "test: %v", tc.name)
+				// make sure actual pool assets aren't changed
+				require.Equal(t, expectedPoolAssetsByDenom, tc.originalPoolAssets)
+			} else {
+				require.Error(t, tc.err, "test: %v", tc.name)
+				require.Equal(t, tc.err, err)
```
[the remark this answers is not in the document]
AlpinYukseloglu: added a check to make sure pool assets don't change!

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -33,131 +33,98 @@ var (
...
 
-			require.Equal(t, tc.err, err)
+			err := balancer.UpdateIntermediaryPoolAssetsLiquidity(tc.newLiquidity, tc.originalPoolAssets)
 
-			if tc.err != nil {
-				return
+			if tc.expectPass {
+				require.NoError(t, tc.err, "test: %v", tc.name)
+				// make sure actual pool assets aren't changed
+				require.Equal(t, expectedPoolAssetsByDenom, tc.originalPoolAssets)
```
mattverse, line 121: Hrmmm wait i'm not sure if this part makes sense? How is the updated pool assets equal to original pool assets?

AlpinYukseloglu: Nice catch, comment should say "make sure actual pool assets are properly updated". I also changed `originalPoolAssets` to `poolAssets` to make it less confusing (`originalPoolAssets` is mutated during the main fx call)

mattverse (approved):
LGTM!

PR #1979, opened 2022-07-06
chore: Refactor and Add tests for `AddToExistingLock`

mattverse:
sub-component of: #1838 

`AddToExistingLock` had a bad design of returning array of locks, and then only using the first index out of the given array. 
There's no need to return an array here, but simply return the lock that we've added the token to. This PR changes the returning values of `AddToExistingLock` from an array to a single lock(not pointer referenced).

This PR also includes changing tests for `AddToExistingLock` to be TDD, adding some edge cases to the tests.

## Brief Changelog
- Change return values of `AddToExistingLock`
- Change tests for `AddToExistingLock` to be TDD

## Testing and Verifying

This change added tests 


x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,20 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
...
 
 // AddToExistingLock adds the given coin to the existing lock with the same owner and duration.
-// Returns an empty array of period lock when a lock with the given condition does not exist.
-func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) ([]types.PeriodLock, error) {
+// Returns an empty struct of period lock when a lock with the given condition does not exist.
+func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) (types.PeriodLock, error) {
 	locks := k.GetAccountLockedDurationNotUnlockingOnly(ctx, owner, coin.Denom, duration)
 	// if existing lock with same duration and denom exists, just add there
 	if len(locks) > 0 {
 		lock := locks[0]
```
stackman27, line 56: This has been a confusion for many PRs ([one example](https://github.com/osmosis-labs/osmosis/pull/1233#discussion_r866039858)). Can we add a simple comment on why we take the first lock?

mattverse: addressed!

x/lockup/keeper/msg_server.go
```diff
@@ -39,26 +39,26 @@ func (server msgServer) LockTokens(goCtx context.Context, msg *types.MsgLockToke
...
 				sdk.NewAttribute(types.AttributePeriodLockAmount, msg.Coins.String()),
 			),
 		})
-		return &types.MsgLockTokensResponse{ID: locks[0].ID}, nil
+		return &types.MsgLockTokensResponse{ID: lock.ID}, nil
 	}
 
 	// if the owner + duration combination is new, create a new lock.
-	lock, err := server.keeper.CreateLock(ctx, owner, msg.Coins, msg.Duration)
+	lock, err = server.keeper.CreateLock(ctx, owner, msg.Coins, msg.Duration)
```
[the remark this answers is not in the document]
mattverse: Hrmm im not sure if I follow, the events below are only emitted when a new lock is creating?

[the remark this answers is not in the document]
stackman27: I was moreso wondering whether the emmited events values like `lock.Owner` `lock.Coins` etc would be from this lock `lock, err := server.keeper.AddToExistingLock(ctx, owner, msg.Coins[0], msg.Duration)` or the other `lock, err = server.keeper.CreateLock(ctx, owner, msg.Coins, msg.Duration)`? 


I believe it has to be from the second i.e. CreateLock function but the lock value was deassigned there so all of those emit events values are coming from  AddToExistingLock function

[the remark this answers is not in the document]
stackman27: Like can't we just replace `lock, err = server.keeper.CreateLock(ctx, owner, msg.Coins, msg.Duration)` to 
`_, err = server.keeper.CreateLock(ctx, owner, msg.Coins, msg.Duration)` ?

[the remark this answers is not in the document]
stackman27: aaah! So `lock` has already been declared + assigned at `AddToExistingLock` but if it is empty we just assign lock to `CreateLock` output

I just got confused on assignment and declaration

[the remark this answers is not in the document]
mattverse: No problem!

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,21 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
...
 	// if existing lock with same duration and denom exists, just add there
 	if len(locks) > 0 {
+		// there should only be a sinlge lock with the same duration + token, thus we take the first lock
 		lock := locks[0]
 		_, err := k.AddTokensToLockByID(ctx, lock.ID, owner, coin)
 		if err != nil {
-			return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+			return types.PeriodLock{}, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
 		}
+		return locks[0], nil
```
p0mvn, line 62:
```suggestion
		return lock, nil
```

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,21 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
 }
 
 // AddToExistingLock adds the given coin to the existing lock with the same owner and duration.
-// Returns an empty array of period lock when a lock with the given condition does not exist.
-func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) ([]types.PeriodLock, error) {
+// Returns an empty struct of period lock when a lock with the given condition does not exist.
```
p0mvn, line 51:
```suggestion
// Returns an empty period lock struct when a lock with the given condition does not exist.
```

p0mvn, line 51: Also, what is returned when a lock does exist? Consider elaborating please

x/lockup/keeper/lock.go
p0mvn, line 62: I have a concern - so we add tokens to lock by id but then return an old lock. Is that expected? 

The lock must be updated with tokens, maybe we need to return a new lock returned by ` AddTokensToLockByID(...)`

p0mvn (changes requested):
Nice work. I worry that we might be returning an old lock from `AddToExistingLock`. Could you please confirm if that is the case? 

We should make assertions on the contents of the lock returned by `AddToExistingLock`

x/lockup/keeper/msg_server.go
```diff
@@ -39,26 +39,26 @@ func (server msgServer) LockTokens(goCtx context.Context, msg *types.MsgLockToke
...
 	// If so, simply add tokens to the existing lock.
-	locks, err := server.keeper.AddToExistingLock(ctx, owner, msg.Coins[0], msg.Duration)
+	lock, err := server.keeper.AddToExistingLock(ctx, owner, msg.Coins[0], msg.Duration)
 	if err != nil {
 		return nil, err
 	}
 
 	// return the lock id of the existing lock when successfully added to the existing lock.
-	if len(locks) > 0 {
+	if lock.ID != 0 {
```
AlpinYukseloglu, line 48: Is the idea here that if the lock exists (ID > 0), then we can assume that it was successfully added to?

x/lockup/keeper/lock.go
p0mvn: What if you try logging `updatedLock`? That's what I suspect is going to be outdated.

It seems to me that we update the db correctly so a future
```
lock1, err := suite.App.LockupKeeper.GetLockByID(suite.Ctx, updatedLock.ID)
```

should work as expected. However, the return from `AddToExistingLock` itself might be outdated

stackman27: Here's the output, interesting thing it is logging the same thing when i try it with return value you suggested. 
```
Matt 
Lock: {1 osmo1v9jxgu3395kj6tfd95kj6tfd95kj6tfd0hptum 1s 0001-01-01 00:00:00 +0000 UTC 10stake}
UpdatedLock:  {1 osmo1v9jxgu3395kj6tfd95kj6tfd95kj6tfd0hptum 1s 0001-01-01 00:00:00 +0000 UTC 10stake}

Roman
Lock: {1 osmo1v9jxgu3395kj6tfd95kj6tfd95kj6tfd0hptum 1s 0001-01-01 00:00:00 +0000 UTC 10stake}
UpdatedLock: {1 osmo1v9jxgu3395kj6tfd95kj6tfd95kj6tfd0hptum 1s 0001-01-01 00:00:00 +0000 UTC 10stake}
```

I think a solution might be to return the lock from `TestLog, err := k.AddTokensToLockByID(ctx, lock.ID, owner, coin)` which actually outputs, the correct result: `TestLog: ID:1 owner:"osmo1v9jxgu3395kj6tfd95kj6tfd95kj6tfd0hptum" duration:<seconds:1 > end_time:<seconds:-62135596800 > coins:<denom:"stake" amount:"20" > `

p0mvn: Yeah we need to be returning from `AddTokensToLockByID`

x/lockup/keeper/lock.go
mattverse: done!

x/lockup/keeper/lock.go
mattverse: good catch!

I took a more detailed look, currently changing this to simply return the lock ID, as that's whats only being used when this method is getting called. No reason to pass locks around!

x/lockup/keeper/lock_test.go
```diff
@@ -350,64 +349,110 @@ func (suite *KeeperTestSuite) TestCreateLock() {
...
-	_, err = suite.App.LockupKeeper.AddTokensToLockByID(cacheCtx, 1111, addr1, addCoins)
-	suite.Require().Error(err)
+		suite.FundAcc(addr1, sdk.Coins{initialLockCoin})
+		balanceBeforeLock := suite.App.BankKeeper.GetAllBalances(suite.Ctx, tc.lockingAddress)
 
-	// try to add tokens with lack balance
-	cacheCtx, _ = suite.Ctx.CacheContext()
-	_, err = suite.App.LockupKeeper.AddTokensToLockByID(cacheCtx, locks[0].ID, addr1, addCoins)
-	suite.Require().Error(err)
+		_, err = suite.App.LockupKeeper.AddToExistingLock(suite.Ctx, tc.lockingAddress, tc.tokenToAdd, tc.duration)
```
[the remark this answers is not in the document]
mattverse: Changed the API so that it returns a lock ID and we use that lock ID in the test cases.  lmk what you think!

x/lockup/keeper/lock_test.go
```diff
@@ -350,64 +349,112 @@ func (suite *KeeperTestSuite) TestCreateLock() {
...
+			duration:                     time.Second,
+			lockingAddress:               addr1,
+			expectAddTokensToLockSuccess: true,
+		},
+		{
+			name:                         "not the owner of the lock",
+			tokenToAdd:                   initialLockCoin,
+			duration:                     time.Second,
+			lockingAddress:               addr2,
+			expectAddTokensToLockSuccess: false,
```
p0mvn, line 376:
```suggestion
```

x/lockup/keeper/lock_test.go
```diff
@@ -350,64 +349,112 @@ func (suite *KeeperTestSuite) TestCreateLock() {
...
+		},
+		{
+			name:                         "lock with matching duration not existing",
+			tokenToAdd:                   initialLockCoin,
+			duration:                     time.Second * 2,
+			lockingAddress:               addr1,
+			expectAddTokensToLockSuccess: false,
```
p0mvn, line 383:
```suggestion
```

x/lockup/keeper/lock_test.go
```diff
@@ -350,64 +349,112 @@ func (suite *KeeperTestSuite) TestCreateLock() {
...
+		},
+		{
+			name:                         "lock invalid tokens",
+			tokenToAdd:                   sdk.NewCoin("unknown", sdk.NewInt(10)),
+			duration:                     time.Second,
+			lockingAddress:               addr1,
+			expectAddTokensToLockSuccess: false,
```
p0mvn, line 390:
```suggestion
```

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,22 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
...
 // AddToExistingLock adds the given coin to the existing lock with the same owner and duration.
-// Returns an empty array of period lock when a lock with the given condition does not exist.
-func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) ([]types.PeriodLock, error) {
+// Returns the updated lock ID if successfully added coin, returns 0 when a lock with
+// given condition does not exist.
+func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) (uint64, error) {
 	locks := k.GetAccountLockedDurationNotUnlockingOnly(ctx, owner, coin.Denom, duration)
 	// if existing lock with same duration and denom exists, just add there
 	if len(locks) > 0 {
+		// there should only be a single lock with the same duration + token, thus we take the first lock
```
p0mvn, line 57: Please consider adding a check for lock length not being 1, returning an error in such a case and a test.

I think it's useful to make sure that we catch errors as soon as possible. For example, if someone accidentally introduces a bug in `GetAccountLockedDurationNotUnlockingOnly`, returning no locks and there is no test for it, we would be able to catch it right here.

p0mvn (commented):
New return with lock id LGTM, nice work!

On another look, thought about adding an extra check. Let me know what you think

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,22 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
...
 		_, err := k.AddTokensToLockByID(ctx, lock.ID, owner, coin)
 		if err != nil {
-			return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+			return 0, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
 		}
+		return lock.ID, nil
 	}
-	return locks, nil
+
+	return 0, nil
```
p0mvn, line 66: Since there is no existing lock, why wouldn't this be an error?

x/lockup/keeper/msg_server.go
mattverse: yes correct!

x/lockup/keeper/lock.go
mattverse: I was also thinking about this as well, I think further investigation needs to be done before we change the logic that you've pointed out, since this was a logic that has been laying here for a long time.

I wanted to investigate more on this and confirm with certainty before we change it so that we assume that lock length is always 1. I can create an issue to track this! lmk what you think

x/lockup/keeper/lock.go
mattverse: it's mainly because of how this API is getting used. 

this API is getting called in the msg server, we attempt to add to existing lock before creating a new lock with the user + duration combination.

Theefore, in this case, it's not an error that we couldn't find a lock with the user + duration combination.

A idea that pops up in my mind rn is maybe changing this API to return a boolean value as well to indicate whether it has been added to existing lock or not?

lmk what you think

x/lockup/keeper/lock.go
AlpinYukseloglu: If that's the case, then wouldn't we have to come to a conclusion on whether lock length = 1 is a sound assumption before merging this?

Happy to help investigate if you get blocked/want a second pair of eyes

x/lockup/keeper/lock_test.go
```diff
@@ -369,47 +369,43 @@ func (suite *KeeperTestSuite) TestAddTokensToLock() {
 			expectAddTokensToLockSuccess: true,
 		},
 		{
-			name:                         "not the owner of the lock",
-			tokenToAdd:                   initialLockCoin,
-			duration:                     time.Second,
-			lockingAddress:               addr2,
-			expectAddTokensToLockSuccess: false,
```
AlpinYukseloglu: Am I correct in assuming you removed `expectAddTokensToLockSuccess` wherever it was false since it's not used in the test logic (because we use `else`)?

If so, I have personally found that having an expected pass or fail on every test case makes them much easier to read and understand, especially since we have no explicit standard saying "if `expectSuccess` is omitted then assume it's a failure case"

So I have a preference to keep an expected pass/fail in each test case just for clarity, although I recognize this might be a question about testing standards so this is a light preference

x/lockup/keeper/lock_test.go
```diff
@@ -369,47 +369,43 @@ func (suite *KeeperTestSuite) TestAddTokensToLock() {
...
+			expectedError:  true,
 		},
 	}
 
 	for _, tc := range testCases {
 		suite.SetupTest()
 		// lock with balance
 		suite.FundAcc(addr1, sdk.Coins{initialLockCoin})
-		lock, err := suite.App.LockupKeeper.CreateLock(suite.Ctx, addr1, sdk.Coins{initialLockCoin}, time.Second)
+		originalLock, err := suite.App.LockupKeeper.CreateLock(suite.Ctx, addr1, sdk.Coins{initialLockCoin}, time.Second)
```
AlpinYukseloglu, line 402: Out of curiosity, what's the point of clarifying that this is the "original" lock? Doesn't `AddToExistingLock` mutate it and make it non-original?

x/lockup/keeper/lock_test.go
```diff
@@ -369,47 +369,43 @@ func (suite *KeeperTestSuite) TestAddTokensToLock() {
...
 		suite.Require().NoError(err)
 
 		suite.FundAcc(addr1, sdk.Coins{initialLockCoin})
 		balanceBeforeLock := suite.App.BankKeeper.GetAllBalances(suite.Ctx, tc.lockingAddress)
 
-		_, err = suite.App.LockupKeeper.AddToExistingLock(suite.Ctx, tc.lockingAddress, tc.tokenToAdd, tc.duration)
+		lockID, err := suite.App.LockupKeeper.AddToExistingLock(suite.Ctx, tc.lockingAddress, tc.tokenToAdd, tc.duration)
```
AlpinYukseloglu, line 408: Just to clarify, this is because `AddToExistingLock` now returns the lock ID instead of the full lock, correct?

x/lockup/keeper/lock.go
mattverse: This was the exact logic gate(we had the same check before this PR) we had before changing so that part should be fine in terms of safteyness

x/lockup/keeper/lock_test.go
mattverse: Good question! This is not a pointer reference to the lock so `AddToExistingLock` would not make this non-original

x/lockup/keeper/lock_test.go
mattverse: correct!

x/lockup/keeper/lock.go
p0mvn: How do you think the boolean would help? I might not be seeing it but I'm afraid that it would only overcomplicate the logic of the caller.

In my opinion, independent of the solution chosen, we should decouple the implementation details of this method from its caller.

I'd like to suggest returning an error here and then let the caller decide how they want to handle it. That is, the caller can check what error it is specifically and continue normally if the error is deemed acceptable.

Another idea is to implement a method `HasLock()` that we call before calling this method. At the same time, we can make this method assume that the lock already exists. If not, an error would be an expected return here.

Let me know what you decide

x/lockup/keeper/lock_test.go
AlpinYukseloglu: Ah I see, thanks!

AlpinYukseloglu (approved):
All my comments/questions have been addressed – LGTM!

x/lockup/keeper/lock.go
mattverse: The original intention behind my suggestion of adding a boolean field was to separate real errors and simple cases without matching locks, but I think your solution works as well!

Just added `HasLock` check!

x/lockup/keeper/lock.go
mattverse: https://github.com/osmosis-labs/osmosis/issues/2142

x/lockup/keeper/lock.go
p0mvn: Awesome! So do you think it makes sense to return an error now here, making this method only be successful if the lock exists?

mattverse: Makes sense, this change makes it state breaking though right?

p0mvn: Yes, it does. We shouldn't backport this

mattverse: hrrmmm do you think this change is worth doing if its state breaking?

mattverse:
agreed with you on adding error when there's no lock druing `AddToExistingLock`,changing it + adding a changlog entry and requesting a final review

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,31 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
...
+	if err != nil {
+		return 0, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
 	}
-	return locks, nil
+
+	return lock.ID, nil
+}
+
+// HasLock returns true if lock with the given condition exists
+func (k Keeper) HasLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) bool {
```
p0mvn, line 73: Can we add a table-driven test for this method as well, please? Also, it might make sense to have it unexported since we, currently, only call it from the same package.

x/lockup/keeper/lock.go
```diff
@@ -48,18 +48,31 @@ func (k Keeper) BeginUnlockAllNotUnlockings(ctx sdk.Context, account sdk.AccAddr
 }
 
 // AddToExistingLock adds the given coin to the existing lock with the same owner and duration.
-// Returns an empty array of period lock when a lock with the given condition does not exist.
-func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sdk.Coin, duration time.Duration) ([]types.PeriodLock, error) {
+// Returns the updated lock ID if successfully added coin, returns 0 when a lock with
+// given condition does not exist.
```
p0mvn, line 52:
```suggestion
// Returns the updated lock ID if successfully added coin, returns 0 and error when a lock with
// given condition does not exist, or if fails to add to lock.
```

x/lockup/keeper/lock_test.go
```diff
@@ -333,64 +332,182 @@ func (suite *KeeperTestSuite) TestCreateLock() {
...
+		}
 
-	// try to add tokens to unavailable lock
-	cacheCtx, _ := suite.Ctx.CacheContext()
-	err = simapp.FundAccount(suite.App.BankKeeper, cacheCtx, addr1, sdk.Coins{addCoins})
-	suite.Require().NoError(err)
-	// curBalance := suite.App.BankKeeper.GetAllBalances(cacheCtx, addr1)
-	_, err = suite.App.LockupKeeper.AddTokensToLockByID(cacheCtx, 1111, addr1, addCoins)
-	suite.Require().Error(err)
+		if tc.expectAddTokensToLockSuccess {
```
p0mvn, line 401: What is the difference between `expectError` and `expectAddTokensToLockSuccess`? Can we use only `expectError`?

p0mvn (approved):
LGTM. Thanks for working out all edge cases here @mattverse

x/lockup/keeper/lock_test.go
mattverse: good catch, the original intent was to create a separate case for lock failing and adding to lock failing, but these can be grouped to one. Changed to removing the field!

alexanderbez (approved):

PR #1980, opened 2022-07-06
Add partial ord for begin block

ValarDragon:

Adds a partial ordering for begin block. Right now, this replicates the exact same ordering that was present before,
and just removes no-op begin blockers. Its left to future work to further improve this.

I'd like this to be backported, so we can easily sanity check state compatibility / not have to worry about that come release time. (Our WIP minor release workflow should catch this if problematic =)) )

## Brief Changelog
 
  - Use partial ordering for begin blockers
  - Adds a partial ord.Sequence method, for setting a sequence of modules.

## Testing and Verifying

- Verify the github comment, for every module listed as no-op in fact being a no-op
- Checking that every remaining module is constrained to be in the same order that was present before.
- Existing tests pass
- New test for partialOrd.Sequence

  - How is the feature or change documented? In-line comments / reducing surrounding noise.

### DIFF
--- a/app/modules.go
+++ b/app/modules.go
@@ -144,7 +144,21 @@ func appModules(
 
 // orderBeginBlockers Tell the app's module manager how to set the order of
 // BeginBlockers, which are run at the beginning of every block.
-func orderBeginBlockers() []string {
+func orderBeginBlockers(allModuleNames []string) []string {
+	ord := partialord.NewPartialOrdering(allModuleNames)
+	// Upgrades should be run VERY first
+	// Epochs is set to be next right now, this in principle could change to come later / be at the end.
+	// requires more thought into if it breaks existing logic.
+	ord.FirstElements(upgradetypes.ModuleName, epochstypes.ModuleName, capabilitytypes.ModuleName)
+	// mint must come before pool incentives & distribution
+	ord.Before(minttypes.ModuleName, poolincentivestypes.ModuleName)
+	ord.Before(minttypes.ModuleName, distrtypes.ModuleName)
+	// staking, slashing & superfluid should come after distribution
+	ord.Before(distrtypes.ModuleName, slashingtypes.ModuleName)
+	ord.Before(distrtypes.ModuleName, stakingtypes.ModuleName)
+	ord.Before(distrtypes.ModuleName, superfluidtypes.ModuleName)
+	// evidence should come after slashing
+	ord.Before(slashingtypes.ModuleName, evidencetypes.ModuleName)
 	return []string{
 		// Upgrades should be run VERY first
 		upgradetypes.ModuleName,
--- a/app/modules.go
+++ b/app/modules.go
@@ -142,58 +142,23 @@ func appModules(
 	}
 }
 
-// orderBeginBlockers Tell the app's module manager how to set the order of
-// BeginBlockers, which are run at the beginning of every block.
+// orderBeginBlockers returns the order of BeginBlockers, by module name.
 func orderBeginBlockers(allModuleNames []string) []string {
-	// ord := partialord.NewPartialOrdering(allModuleNames)
-	// // Upgrades should be run VERY first
-	// // Epochs is set to be next right now, this in principle could change to come later / be at the end.
-	// // requires more thought into if it breaks existing logic.
-	// ord.FirstElements(upgradetypes.ModuleName, epochstypes.ModuleName, capabilitytypes.ModuleName)
+	ord := partialord.NewPartialOrdering(allModuleNames)
+	// Upgrades should be run VERY first
+	// Epochs is set to be next right now, this in principle could change to come later / be at the end.
+	// But would have to be a holistic change with other pipelines taken into account.
+	ord.FirstElements(upgradetypes.ModuleName, epochstypes.ModuleName, capabilitytypes.ModuleName)
 
-	// // IBC ordering constraint
-	// ord.Sequence(ibchost.ModuleName)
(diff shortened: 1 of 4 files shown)


app/modules.go
```diff
@@ -145,53 +145,54 @@ func appModules(
...
+		tokenfactorytypes.ModuleName,   // no begin block logic
 		// superfluid must come after distribution and epochs
 		superfluidtypes.ModuleName,
-		bech32ibctypes.ModuleName,
-		txfeestypes.ModuleName,
-		wasm.ModuleName,
+		bech32ibctypes.ModuleName, // no begin block logic
+		txfeestypes.ModuleName,    // no begin block logic
+		wasm.ModuleName,           // no begin block logic
 	}
```
ValarDragon: Please verify this no-op list. 

Links to modules that arent in the SDK or Osmosis:
* ibchost.ModuleName - Actually does have a begin block! Added back in subsequent commit. Proof of name https://github.com/cosmos/ibc-go/blob/main/modules/core/module.go#L43-L46 (I made an issue in ibc-go for this being an odd location for the name), BeginBlock https://github.com/cosmos/ibc-go/blob/main/modules/core/module.go#L165-L168
* ibctransfertypes.ModuleName - https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L144-L146
* icatypes.ModuleName - https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/module.go#L191
* wasm: https://github.com/CosmWasm/wasmd/blob/main/x/wasm/module.go#L176

app/modules.go
```diff
@@ -142,43 +142,25 @@ func appModules(
...
+func orderBeginBlockers(allModuleNames []string) []string {
+	ord := partialord.NewPartialOrdering(allModuleNames)
+	// Upgrades should be run VERY first
+	// Epochs is set to be next right now, this in principle could change to come later / be at the end.
+	// But would have to be a holistic change with other pipelines taken into account.
+	ord.FirstElements(upgradetypes.ModuleName, epochstypes.ModuleName, capabilitytypes.ModuleName)
+
+	// Staking ordering
+	// TODO: Perhaps this can be relaxed, left to future work to analyze.
+	ord.Sequence(distrtypes.ModuleName, slashingtypes.ModuleName, evidencetypes.ModuleName, stakingtypes.ModuleName)
```
AlpinYukseloglu, line 155: Out of curiosity, what is the purpose of creating/using `Sequence` here? i.e. in what context would there be elements interspersed for `BeginBlock` logic? Is `BeginBlock` ordering being changed/affected anywhere else in the codebase?

AlpinYukseloglu: Is the part that's "just found" random, or is the ordering somehow externally controlled e.g. by validators?

ValarDragon: No, its deterministically found by solving for a total ordering that satisfies all constraints. If your interested in the details of the implemented algorithm, checkout the partialord package! (But TL;DR topological sort)

app/modules.go
```diff
@@ -139,42 +139,25 @@ func appModules(
...
+	// superfluid must come after distribution & epochs.
+	// TODO: we actually set it to come after staking, since thats what happened before, and want to minimize chance of break.
+	ord.After(superfluidtypes.ModuleName, stakingtypes.ModuleName)
+	// TODO: This can almost certainly be un-constrained, but we keep the constraint to match prior functionality.
+	// IBChost came after staking, before superfluid.
+	ord.After(superfluidtypes.ModuleName, ibchost.ModuleName)
+	// every remaining module's begin block is a no-op.
+	return ord.TotalOrdering()
```
p0mvn, line 160: Are all of the TODOs actionable here? I feel like they are all important notes to be kept but some do not require any extra action

app/modules.go
```diff
@@ -139,42 +139,25 @@ func appModules(
...
+
+	// Staking ordering
+	// TODO: Perhaps this can be relaxed, left to future work to analyze.
+	ord.Sequence(distrtypes.ModuleName, slashingtypes.ModuleName, evidencetypes.ModuleName, stakingtypes.ModuleName)
+	// superfluid must come after distribution & epochs.
+	// TODO: we actually set it to come after staking, since thats what happened before, and want to minimize chance of break.
+	ord.After(superfluidtypes.ModuleName, stakingtypes.ModuleName)
+	// TODO: This can almost certainly be un-constrained, but we keep the constraint to match prior functionality.
+	// IBChost came after staking, before superfluid.
+	ord.After(superfluidtypes.ModuleName, ibchost.ModuleName)
```
p0mvn, line 158: How is `ibchost` after `staking` guaranteed?

I might be missing something but I think we have:

- [x] `superfluid` after `staking`
- [x] `superfluid` after `ibchost`
- [ ] ibchost after staking (not in line with the comment)

app/modules.go
p0mvn: No-op list LGTM

app/modules.go
ValarDragon: ah your right, good catch

app/modules.go
ValarDragon: I think so, I'm viewing them as notes to come back and relax these constraints

p0mvn (approved):
LGTM! Nice work

I'm not sure what the timeline for addressing TODOs is. If not immediate, please consider moving them to issues to avoid having context missed over time

ValarDragon:
Good point, will make an issue for it

PR #1982, opened 2022-07-06
Add TWAP stub package

ValarDragon:

This PR adds a stub starting some of the TWAP logic. Has a couple high level notes around where more logic goes.
One of the goals is that it is relatively decoupled from gamm, only having read-only access to a pool, along with getting hooks.

This PR basically includes minimum scaffolding necessary to only have complexity once per pool per block, rather than on every swap.

I'd like to merge this if viewed as acceptable, so we can carry forward with more TWAP experimentation with this boilerplate out of the way in review.

## Brief Changelog

- Add Stub

## Testing and Verifying

No core logic / wiring present to test yet. Just stubs for future work / to make this not need to be reviewed in subsequent work.

  - How is the feature or change documented? Should be added to README of TWAP and gamm.

### DIFF
--- a/x/gamm/twap/keeper.go
+++ b/x/gamm/twap/keeper.go
@@ -3,6 +3,6 @@ package twap
 import sdk "github.com/cosmos/cosmos-sdk/types"
 
 type twapkeeper struct {
-	storeKey     sdk.StoreKey
+	// storeKey     sdk.StoreKey
 	transientKey sdk.TransientStoreKey
 }
--- a/x/gamm/twap/keeper.go
+++ b/x/gamm/twap/keeper.go
@@ -1,0 +1,8 @@
+package twap
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+type twapkeeper struct {
+	storeKey     sdk.StoreKey
+	transientKey sdk.TransientStoreKey
+}
--- a/x/gamm/twap/module.go
+++ b/x/gamm/twap/module.go
@@ -1,0 +1,135 @@
+package twap
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/gorilla/mux"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/spf13/cobra"
+	abci "github.com/tendermint/tendermint/abci/types"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/keeper"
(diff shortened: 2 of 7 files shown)


x/gamm/twap/module.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (AppModuleBasic) Name() string { return twaptypes.ModuleName }
+
+func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+}
+
+func (AppModuleBasic) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
+	return json.RawMessage{}
+	// return cdc.MustMarshalJSON(types.DefaultGenesis())
```
p0mvn, line 40:
```suggestion
```

x/gamm/twap/module.go
```diff
@@ -0,0 +1,137 @@
...
+func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
+}
+
+type AppModule struct {
+	AppModuleBasic
+
+	ak types.AccountKeeper
+	bk types.BankKeeper
+	gk keeper.Keeper
+	tk twapkeeper
```
p0mvn, line 80: Consider expanding abbreviations for readability

x/gamm/twap/module.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (AppModuleBasic) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
+	return json.RawMessage{}
+	// return cdc.MustMarshalJSON(types.DefaultGenesis())
+}
+
+// ValidateGenesis performs genesis state validation for the gamm module.
+func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, config client.TxEncodingConfig, bz json.RawMessage) error {
+	// var genState types.GenesisState
+	// if err := cdc.UnmarshalJSON(bz, &genState); err != nil {
```
p0mvn, line 46: deadcode

x/gamm/twap/keeper.go
```diff
@@ -0,0 +1,8 @@
+package twap
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+type twapkeeper struct {
+	// storeKey     sdk.StoreKey
```
p0mvn, line 6: deadcode

p0mvn (approved):

x/gamm/twap/module.go
ValarDragon: This is standard within the SDK

x/gamm/twap/module.go
ValarDragon: Lets keep this for now, basically have to add back that line once default genesis exists

x/gamm/twap/keeper.go
ValarDragon: Lets keep this for now, basically have to add back that line once more logic exists

x/gamm/twap/module.go
ValarDragon: Lets keep this for now, basically have to add back that line once default genesis exists

Issue #1942, opened 2022-07-01
x/gamm: test that swap and LP events are emitted


p0mvn:
## Background

Currently, our tests are missing assertions about events. Moreover, it seems that some of our events are missing based on #1939

Therefore, this issue is twofold:
1. Investigate if / which events are missing and restore them
3. Add assertions in tests that events are being emitted in the correct places

## Acceptance Criteria
- all events are in the right places
   - [x] swap events:
   - [x] pool joined
   - [x] pool exited
   - [ ] pool created
   - [ ] msg events
- consider emitting all events from the message server
- assertions in unit tests are added
- all unit tests continue to pass

p0mvn:
I have some progress started here: https://github.com/osmosis-labs/osmosis/tree/roman/gamm-events

p0mvn:
@mattverse following up on your comment in one of the gamm event PRs. After doing some investigation, I think we should not move the logic for emitting swap and liquidity events to the message server.

The reason is that some gamm keeper methods might be called from other modules. For example, superfluid:
https://github.com/osmosis-labs/osmosis/blob/4970f165f439ad43c8a188e056b4ad69f54aebb4/x/superfluid/keeper/unpool.go#L60

If we were emitting events in the message server, the remove liquidity event would not be emitted.  Therefore, I think it makes sense to keep the locations as is and thoroughly test.

mattverse:
good point! Agreed 😎

ValarDragon:
Pretty sure this got completed!

PR #1983, opened 2022-07-06
refactor(x/gamm): swap events

p0mvn:
Part of #1942

## Brief Changelog

Refactor and test gamm swap events. This does not change when the events are emitted. It only does some basic refactor to make events more testable.

Additionally, it unit tests and documents when swap events are emitted. All other events in gamm will be tested in incoming PRs

## Testing and Verifying

This change added tests for swap events at 3 levels of abstraction:
- internal events package
- keeper swap methods
- message server

### DIFF
--- a/x/gamm/keeper/internal/events/emit.go
+++ b/x/gamm/keeper/internal/events/emit.go
@@ -1,0 +1,70 @@
+package events
+
+import (
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+)
+
+func EmitSwapEvent(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, input sdk.Coins, output sdk.Coins) {
+	if ctx.EventManager() == nil {
+		return
+	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		newSwapEvent(sender, poolId, input, output),
+	})
+}
+
+func newSwapEvent(sender sdk.AccAddress, poolId uint64, input sdk.Coins, output sdk.Coins) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtTokenSwapped,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		sdk.NewAttribute(sdk.AttributeKeySender, sender.String()),
+		sdk.NewAttribute(types.AttributeKeyPoolId, strconv.FormatUint(poolId, 10)),
+		sdk.NewAttribute(types.AttributeKeyTokensIn, input.String()),
+		sdk.NewAttribute(types.AttributeKeyTokensOut, output.String()),
+	)
+}
+
+func EmitAddLiquidityEvent(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, liquidity sdk.Coins) {
+	if ctx.EventManager() == nil {
+		return
+	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		newAddLiquidityEvent(sender, poolId, liquidity),
+	})
+}
+
+func newAddLiquidityEvent(sender sdk.AccAddress, poolId uint64, liquidity sdk.Coins) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtPoolJoined,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		sdk.NewAttribute(sdk.AttributeKeySender, sender.String()),
+		sdk.NewAttribute(types.AttributeKeyPoolId, strconv.FormatUint(poolId, 10)),
+		sdk.NewAttribute(types.AttributeKeyTokensIn, liquidity.String()),
+	)
+}
+
+func EmitRemoveLiquidityEvent(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, liquidity sdk.Coins) {
+	if ctx.EventManager() == nil {
+		return
+	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		newRemoveLiquidityEvent(sender, poolId, liquidity),
+	})
+}
+
+func newRemoveLiquidityEvent(sender sdk.AccAddress, poolId uint64, liquidity sdk.Coins) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtPoolExited,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		sdk.NewAttribute(sdk.AttributeKeySender, sender.String()),
+		sdk.NewAttribute(types.AttributeKeyPoolId, strconv.FormatUint(poolId, 10)),
+		sdk.NewAttribute(types.AttributeKeyTokensOut, liquidity.String()),
+	)
+}
--- a/x/gamm/keeper/msg_server_test.go
+++ b/x/gamm/keeper/msg_server_test.go
@@ -1,0 +1,226 @@
+package keeper_test
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/keeper"
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+)
+
+const (
+	doesNotExistDenom = "nodenom"
+)
+
+var _ = suite.TestingSuite(nil)
+
+// TestSwapExactAmountIn_Events tests that events are correctly emitted
+// when calling SwapExactAmountIn.
+func (suite *KeeperTestSuite) TestSwapExactAmountIn_Events() {
+	const (
+		tokenInMinAmount = 1
+		tokenIn          = 5
+	)
+
+	testcases := map[string]struct {
+		routes                []types.SwapAmountInRoute
+		tokenIn               sdk.Coin
+		tokenOutMinAmount     sdk.Int
+		expectError           bool
+		expectedSwapEvents    int
+		expectedMessageEvents int
+	}{
+		"zero hops": {
+			routes:                []types.SwapAmountInRoute{},
+			tokenIn:               sdk.NewCoin("foo", sdk.NewInt(tokenIn)),
+			tokenOutMinAmount:     sdk.NewInt(tokenInMinAmount),
+			expectedMessageEvents: 1, // 1 gamm.
+		},
+		"one hop": {
+			routes: []types.SwapAmountInRoute{
+				{
+					PoolId:        1,
+					TokenOutDenom: "bar",
+				},
+			},
+			tokenIn:               sdk.NewCoin("foo", sdk.NewInt(tokenIn)),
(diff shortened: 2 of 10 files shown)


x/gamm/keeper/multihop.go
```diff
@@ -52,6 +52,9 @@ func (k Keeper) MultihopSwapExactAmountOut(
 	if err != nil {
 		return sdk.Int{}, err
 	}
+	if len(insExpected) == 0 {
+		return sdk.Int{}, nil
+	}
```
AlpinYukseloglu, line 57: Probably belongs in a separate PR but good addition (small enough to keep in this one imo)

x/gamm/keeper/msg_server_test.go
```diff
@@ -0,0 +1,226 @@
...
+	const (
+		tokenInMinAmount = 1
+		tokenIn          = 5
+	)
+
+	testcases := map[string]struct {
+		routes                []types.SwapAmountInRoute
+		tokenIn               sdk.Coin
+		tokenOutMinAmount     sdk.Int
+		expectError           bool
```
AlpinYukseloglu, line 29: This should probably be flipped to `expectPass` to be in line with the rest of the table driven tests in our codebase. If there are any other `expectError`s, we should probably flip them to expectPass (or vice versa) for consistency (I'll make a separate issue about making this switch for other files)

x/gamm/spec/README.md
```diff
@@ -968,6 +950,48 @@ osmosisd tx gamm swap-exact-amount-out 1000000ibc/27394FB092D2ECCD56123C74F36E4C
 ```
 
 ## Other resources
-* [Creating a liquidity bootstrapping pool](./client/docs/create-lbp-pool.md)
-* [Creating a pool with a pool file](./client/docs/create-pool.md)
 
+- [Creating a liquidity bootstrapping pool](./client/docs/create-lbp-pool.md)
+- [Creating a pool with a pool file](./client/docs/create-pool.md)
```
AlpinYukseloglu, line 955: Looks like your IDE has the asterisk -> dash auto formatting too – should probably keep them as asterisks for now to be consistent with the rest of our markdown files but this is something I'm dealing with too

AlpinYukseloglu, line 955:
```suggestion
* [Creating a liquidity bootstrapping pool](./client/docs/create-lbp-pool.md)
* [Creating a pool with a pool file](./client/docs/create-pool.md)
```

x/gamm/spec/README.md
```diff
@@ -968,6 +950,48 @@ osmosisd tx gamm swap-exact-amount-out 1000000ibc/27394FB092D2ECCD56123C74F36E4C
...
+
+## Events
+
+There are 4 types of events that exist in GAMM:
+
+- `sdk.EventTypeMessage` - "message"
+- `types.TypeEvtPoolJoined` - "pool_joined"
+- `types.TypeEvtPoolExited` - "pool_exited"
+- `types.TypeEvtPoolCreated` - "pool_created"
+- `types.TypeEvtTokenSwapped` - "token_swapped"
```
AlpinYukseloglu, line 965:
```suggestion
* `sdk.EventTypeMessage` - "message"
* `types.TypeEvtPoolJoined` - "pool_joined"
* `types.TypeEvtPoolExited` - "pool_exited"
* `types.TypeEvtPoolCreated` - "pool_created"
* `types.TypeEvtTokenSwapped` - "token_swapped"
```

x/gamm/spec/README.md
```diff
@@ -968,6 +950,48 @@ osmosisd tx gamm swap-exact-amount-out 1000000ibc/27394FB092D2ECCD56123C74F36E4C
...
+- `sdk.AttributeKeyModule` - "module"
+  - The value is the module's name - "gamm".
+- sdk.AttributeKeySender
+  - The value is the address of the sender who created the swap message.
+- types.AttributeKeyPoolId
+  - The value is the pool id of the pool where swap occurs.
+- types.AttributeKeyTokensIn
+  - The value is the string representation of the tokens being swapped in.
+- types.AttributeKeyTokensOut
+  - The value is the string representation of the tokens being swapped out.
```
AlpinYukseloglu, line 997:
```suggestion
* `sdk.AttributeKeyModule` - "module"
  * The value is the module's name - "gamm".
* sdk.AttributeKeySender
  * The value is the address of the sender who created the swap message.
* types.AttributeKeyPoolId
  * The value is the pool id of the pool where swap occurs.
* types.AttributeKeyTokensIn
  * The value is the string representation of the tokens being swapped in.
* types.AttributeKeyTokensOut
  * The value is the string representation of the tokens being swapped out.
```

AlpinYukseloglu (approved):
Great work! I left a few minor comments but overall LGTM

x/gamm/spec/README.md
ValarDragon: I mean we can switch, I don't care that much. Just don't want a continual back and forth between the two.

x/gamm/keeper/msg_server_test.go
p0mvn: The best practice and a convention is to have `expectError error` where we can assert the actual kind of the error. Using a boolean is a shortcut to the original rule.

I'm not sure why we would have `expectPass` as a rule. I think it might make sense where the majority of test cases are errors and only one is valid. However, I think this can be decided on the case-by-case basis

x/gamm/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,70 @@
+package events
```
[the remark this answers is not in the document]
p0mvn: Having it in `types` allows it to be imported from any other package in the project. However, we only should be able to emit these events from the gamm keeper or message server. Therefore, making this package internal to the gamm `keeper` package allows for avoiding accidental imports anywhere else.

Another option could be is to simply define it as an "unexported" function in a separate file inside gamm keeper. With that, there are 2 issues:
- We would have to use `export_test.go` pattern to be able to export and test these event functions
- In my opinion, our keeper package is already bloated with functionality. `events` felt like a separate component that could be extracted

At the same time, I liked the ability to call these functions as `events.Emit...()` (by prefixing with `events.`. This makes it more visible than `k.emit...()` that events are being emitted. 

I'm planning to incrementally move all remaining events from the gamm keeper into this internal package and unit test them in PRs to come.

Let me know what you think

x/gamm/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,70 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+)
+
+func EmitSwapEvent(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, input sdk.Coins, output sdk.Coins) {
+	if ctx.EventManager() == nil {
+		return
+	}
```
mattverse, line 14: asking to learn: what are the supposable scenario that we don't ctx doesn't have EventManager? We didn't have this check before right?

x/gamm/types/events.go
```diff
@@ -1,11 +1,5 @@
 package types
 
-import (
-	"strconv"
-
-	sdk "github.com/cosmos/cosmos-sdk/types"
-)
-
 const (
```
mattverse, line 3: Consider moving these and replacing these type calls with the newly added events emit. IMO could be confusing to maintain two different event packages for a single module.

mattverse (commented):
Moving events into an internal package does seem like something unconventional, but I dont feel conflicted to this, as long as we figure out a way to maintain the two different packages for events, one in types, one in `events` package, which imo could be a confusing architecture

x/gamm/types/events.go
p0mvn: The motivation behind leaving constants in types was to have them exposed / exported to clients who might want to use these constants to subscribe to events.

However, the emission logic should only be exposed to the gamm package, not the clients. That's why it was moved to an internal package inside gamm.

This is to preserve information hiding where we only export what clients should be able to see while unexport any package-private functionality.

x/gamm/keeper/internal/events/emit.go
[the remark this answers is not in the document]
alexanderbez: In general, I prefer to be consistent with other modules. Typically, events are defined in `types/events.go`. 

Now I don't have a strong opinion here, but I will state that having events "public" can be a good and desirable thing. E.g. I've wrote many tools that exported and scraped on-chain data, where I used events to query for txs/blocks. Being able to directly reference the events was immensely useful instead of having to redefine them.  

So for this reason, I say just keep them in `types/`.

x/gamm/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,70 @@
+package events
+
+import (
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v7/x/gamm/types"
+)
+
```
alexanderbez, line 10: Unless we have multiple places where we call `Emit*Event`, I'd say extracting these into functions is overkill IMO.

x/gamm/keeper/internal/events/emit.go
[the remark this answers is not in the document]
mattverse: agreed here!

x/gamm/types/events.go
p0mvn: Let's continue discussing this in this thread please:
https://github.com/osmosis-labs/osmosis/pull/1983#discussion_r915894294

x/gamm/keeper/internal/events/emit.go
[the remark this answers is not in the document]
p0mvn: Thanks for the feedback.

I agree about the need for having event names public by clients/tools, For this reason, they are still kept in types as per suggestion:
https://github.com/osmosis-labs/osmosis/blob/ad633616412c32d8918411e06120b65a9fd84a07/x/gamm/types/events.go#L3-L14

This is also consistent with other modules such as `epochs`:
https://github.com/osmosis-labs/osmosis/blob/716c6e4e0cb01bc089559bc3914fa17c93f73f8b/x/epochs/types/events.go#L1-L9

Contrary to event names, clients should never be able to emit the events. This is something that only the gamm module and internal logic should be able to do. That's why this logic was moved to an internal `events` package that is never exposed to outside packages or tools.

Having it in `types` would be information leakage because we are exposing functionality that is not meant to be used by any other module, and, as a result, opening up a vector to make a mistake by accidentally emitting events from somewhere else.

@alexanderbez @mattverse @ValarDragon  please let me know what you think

x/gamm/keeper/internal/events/emit.go
p0mvn: There are indeed multiple places where events are emitted. In addition, the events differ as well as their attributes.

Having each event encapsulated in a function allows for asserting the kind of attributes emitted by events in tests. Otherwise, they would be left completely untested

An example where attributes are tested:
https://github.com/osmosis-labs/osmosis/blob/ad633616412c32d8918411e06120b65a9fd84a07/x/gamm/keeper/internal/events/emit_test.go#L61-L70

ValarDragon (approved):
LGTM

I do share the hesitance around the `internal/events` package, versus making these methods public.

Not sure that I have a particularly thought through opinion here, versus just general hesitance perhaps due to unfamiliarity. The API difference here is in the concrete, just making the old keeper-only private methods (that I liked) now moved to an internal package.

Issue #1837, opened 2022-06-20
x/incentives - internal audit


ValarDragon:

czarcas7ic:
The following are the remaining TODOs from the first pass. Would like to decide which of these we want to tackle as part of this audit and which we should just create issues for:

- [ ] Fixed rewards estimation to work sometimes, but other times it panics [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r909126390)
- [ ] Optimize getToDistributeCoinsFromGauges [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910382317)
- [ ] Fix hack here getLocksToDistributionWithMaxDuration [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910382971)
- [ ] Have doDistributionSends have way more re-use with distribute internal [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910384227)
- [ ] Need further explanation of checkFinishDistribution previous TODO [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910384852)
- [ ] Make TestGetModuleToDistributeCoins TDD [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r911416017)
- [ ] Make TestGetModuleDistributedCoins TDD [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r911416142)
- [ ] Do we need to be concerned with case where this should be ActiveGauges? [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910385386)
- [ ] Figure out what this number should be [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r911417161)
- [ ] Make TestGRPCToDistributeCoins table driven, or simpler [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r911417335)
- [ ] Clean up AfterEpochEnd [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910385724)
- [ ] Switch more code to use userLocks and perpGaugeDesc [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r910385884)
- [ ] Delete all usages of this method (why?) [Discussion](https://github.com/osmosis-labs/osmosis/pull/1885/files#r911417528)
[...]

PR #1984, opened 2022-07-06
x/incentives: protobuf updates

czarcas7ic:
Part of: #1837 

Audit, clean up, and test modifications for the incentives module

This is part 1 of 4:

1. Proto changes
2. CLI fixes
3. Tests
4. All other changes

## Brief Changelog
 
- Better explain incentive protobufs with comments

## Testing and Verifying

This change is already covered by existing tests, such as check-proto CI job

### DIFF
--- a/proto/osmosis/incentives/gauge.proto
+++ b/proto/osmosis/incentives/gauge.proto
@@ -9,8 +9,8 @@ import "osmosis/lockup/lock.proto";
 
 option go_package = "github.com/osmosis-labs/osmosis/v7/x/incentives/types";
 
-// A Gauge is a location for yields to be stored in order to be distributed to
-// stakers who meet the specified conditions
+// A Gauge is an object that stores and distributes yields to recipients who satisfy certain conditions.
+// Currently gauges support conditions around the duration for which a given denom is locked.
 message Gauge {
   // Unique ID of a Gauge
   uint64 id = 1;
--- a/proto/osmosis/incentives/gauge.proto
+++ b/proto/osmosis/incentives/gauge.proto
@@ -18,7 +18,7 @@ message Gauge {
   // Non-perpetual gauges distribute their tokens equally per epoch while the
   // gauge is in the active period. Perpetual gauges distribute all their tokens
   // at a single time and only distribute their tokens again once the gauge is
-  // refilled
+  // refilled, Intended for use with incentives that get refilled daily.
   bool is_perpetual = 2;
   // Rewards are distributed to lockups that are are returned by this lockup
   // query. This is queried via lock duration or by timestamp
--- a/proto/osmosis/incentives/gauge.proto
+++ b/proto/osmosis/incentives/gauge.proto
@@ -9,8 +9,9 @@ import "osmosis/lockup/lock.proto";
 
 option go_package = "github.com/osmosis-labs/osmosis/v7/x/incentives/types";
 
-// A Gauge is an object that stores and distributes yields to recipients who satisfy certain conditions.
-// Currently gauges support conditions around the duration for which a given denom is locked.
+// A Gauge is an object that stores and distributes yields to recipients who
+// satisfy certain conditions. Currently gauges support conditions around the
+// duration for which a given denom is locked.
 message Gauge {
   // Unique ID of a Gauge
   uint64 id = 1;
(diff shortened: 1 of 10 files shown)


proto/osmosis/incentives/gauge.proto
```diff
@@ -9,48 +9,46 @@ import "osmosis/lockup/lock.proto";
 
 option go_package = "github.com/osmosis-labs/osmosis/v7/x/incentives/types";
 
-// Gauge is a struct that works as a single unit for each distribution of
-// incentives defined by x/incentives module.
+// A Gauge is a location for yields to be stored in order to be distributed to
+// stakers who meet the specified conditions
```
ValarDragon, line 13:
```suggestion
// A Gauge is an object that stores and distributes yields to recipients who satisfy certain conditions.
// Currently gauges support conditions around the duration for which a given denom is locked.
```

proto/osmosis/incentives/gauge.proto
```diff
@@ -9,48 +9,46 @@ import "osmosis/lockup/lock.proto";
...
-  // IsPerpetual defines whether the gauge is perpetual or not.
-  // A non-perpetual gauge distributes the incentive tokens equally per epoch
-  // while the gauge is in an active period.
-  // A perpetual gauge distributes all tokens at a single distribution, mainly
-  // used to distribute minted osmo to lp token stakers.
+  // Flag to show if it's a perpetual or non-perpetual gauge
+  // Non-perpetual gauges distribute their tokens equally per epoch while the
+  // gauge is in the active period. Perpetual gauges distribute all their tokens
+  // at a single time and only distribute their tokens again once the gauge is
+  // refilled
```
ValarDragon, line 21:
```suggestion
  // at a single time and only distribute their tokens again once the gauge is
  // refilled, Intended for use with incentives that get refilled daily.
```

proto/osmosis/incentives/gauge.proto
```diff
@@ -9,48 +9,46 @@ import "osmosis/lockup/lock.proto";
...
-  // The ID of the gauge is decided upon gauge creation, incrementing by 1 for
-  // every gauge.
+  // Unique ID of a Gauge
   uint64 id = 1;
-  // IsPerpetual defines whether the gauge is perpetual or not.
-  // A non-perpetual gauge distributes the incentive tokens equally per epoch
-  // while the gauge is in an active period.
-  // A perpetual gauge distributes all tokens at a single distribution, mainly
-  // used to distribute minted osmo to lp token stakers.
+  // Flag to show if it's a perpetual or non-perpetual gauge
```
p0mvn, line 17: I thought the convention was to prefix according to the name of the field like `is_perpetual`?

Example in epochs: https://github.com/osmosis-labs/osmosis/blob/a364494a3ac7f9d369b88ac0c3c29cb0f3542a16/proto/osmosis/epochs/genesis.proto#L10-L18

p0mvn (approved):

proto/osmosis/incentives/gauge.proto
czarcas7ic: If this is the case I will go back and add the prefixes!

czarcas7ic: I went back and did this for all the functions since Dev told me that was golang convention, I didn't realize that was also the protobuf convention (you are probably right)

czarcas7ic: Would be cool if we could add this in the linter too :)

p0mvn: What protobuf does is it generates go code from `proto` files, including comments. For example, for the epochs proto code I linked, the Go would be:
https://github.com/osmosis-labs/osmosis/blob/a364494a3ac7f9d369b88ac0c3c29cb0f3542a16/x/epochs/types/genesis.pb.go#L30-L34
- note that comments are ported over as well

Since we would like to maintain the Go standards, we also want to have the comments in proto files be in the same way as Go so that `pb.go` files are generated with the right style.

The only problem with proto -> Go comment conversion is that it does not cover the case according to Go conventions.
For example, `identifier` should be converted to `Identifier` in Go. However, it preserves the proto case. We might want to address this in the future but this is not a big deal at all for now.

p0mvn (approved):
:rocket:

PR #1885, opened 2022-06-27
chore: x/incentives audit

czarcas7ic:
Closes: #1837

## What is the purpose of the change

Audit, clean up, and test modifications for the incentives module

## Brief Changelog
- Better explain incentive protobufs with comments
- Remove root folder README in favor of spec folder
- Fixes reward estimation query to work as intended (pulls owners lockIDs if not provided)
- Improve / cleanup CLI tests
- Allow benchmark to work again
- Takes all TODOs that existed previously and either made a new issue for it, fixed it, or requires further discussion
- Adds comments to all functions with brief explanation of what it does
- Adds TestSyntheticDistribute

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

PR #1985, opened 2022-07-07
x/incentives: cli updates

czarcas7ic:
Part of: #1837 

Audit, clean up, and test modifications for the incentives module

This is part 2 of 4:

1. Proto changes
2. CLI fixes
3. Tests
4. All other changes

## Brief Changelog
 
- Fixes reward estimation query to work as intended (pulls owners lockIDs if not provided)
- Improve / cleanup CLI tests

## Testing and Verifying

This change is already covered by existing tests, such as cli_test.go within the incentives module

### DIFF
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -322,7 +322,7 @@ $ %s query incentives upcoming-gauges
 	return cmd
 }
 
-// GetCmdUpcomingGaugesPerDenom returns active gauges for a specified denom.
+// GetCmdActiveGaugesPerDenom returns active gauges for specified denom.
 func GetCmdUpcomingGaugesPerDenom() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "upcoming-gauges-per-denom [denom]",
@@ -438,12 +438,7 @@ $ %s query incentives rewards-estimation
 				}
 			}
 
-			// TODO: Figure out why some lock ids are good and some causes "Error: rpc error: code = Unknown desc = panic message redacted to hide potentially sensitive system info: panic"
-			// since owner checks have already been completed above, we switch the owner address to a random module account address since a blank owner panics.
-			// we should find a better way to circumvent this address validity check
-			if owner == "" {
-				owner = "osmo14kjcwdwcqsujkdt8n5qwpd8x8ty2rys5rjrdjj"
-			}
+			// TODO: Fix accumulation store bug. For now, we return a graceful error when attempting to query bugged gauges
 			rewardsEstimateResult, err := queryClient.RewardsEst(cmd.Context(), &types.RewardsEstRequest{
 				Owner:    owner, // owner is used only when lockIds are empty
 				LockIds:  lockIds,
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -394,8 +394,6 @@ $ %s query incentives rewards-estimation
 				if err != nil {
 					return err
 				}
-			} else {
-				owner = ""
 			}
 
 			ownerLocks := []uint64{}
@@ -447,16 +445,16 @@ $ %s query incentives rewards-estimation
 			if owner == "" {
 				owner = "osmo14kjcwdwcqsujkdt8n5qwpd8x8ty2rys5rjrdjj"
 			}
-			res1, err1 := queryClient.RewardsEst(cmd.Context(), &types.RewardsEstRequest{
+			rewardsEstimateResult, err := queryClient.RewardsEst(cmd.Context(), &types.RewardsEstRequest{
 				Owner:    owner, // owner is used only when lockIds are empty
 				LockIds:  lockIds,
 				EndEpoch: endEpoch,
 			})
-			if err1 != nil {
-				return err1
+			if err != nil {
+				return err
 			}
 
-			return clientCtx.PrintProto(res1)
+			return clientCtx.PrintProto(rewardsEstimateResult)
 		},
 	}
 
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -408,7 +408,7 @@ $ %s query incentives rewards-estimation
 
 			lockIdStrs := strings.Split(lockIdsCombined, ",")
 			lockIds := []uint64{}
-			// if user doesn't provide any lockIDs or a lock owner, we don't have enough information to proceed
+			// if user doesn't provide at least one of the lock ids or owner, we don't have enough information to proceed.
 			if lockIdsCombined == "" && owner == "" {
 				err = fmt.Errorf("if owner flag is not set, lock IDs must be provided")
 				return err
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -400,10 +400,8 @@ $ %s query incentives rewards-estimation
 
 			ownerLocks := []uint64{}
 
-			if res != nil {
-				for _, lockId := range res.Locks {
-					ownerLocks = append(ownerLocks, lockId.ID)
-				}
+			for _, lockId := range res.Locks {
+				ownerLocks = append(ownerLocks, lockId.ID)
 			}
 
 			lockIdStrs := strings.Split(lockIdsCombined, ",")
@@ -413,7 +411,7 @@ $ %s query incentives rewards-estimation
 				return fmt.Errorf("if owner flag is not set, lock IDs must be provided")
 
 				// if user provides lockIDs, use these lockIDs in our rewards estimation
-			} else if lockIdsCombined != "" {
+			} else if owner == "" {
 				for _, lockIdStr := range lockIdStrs {
 					lockId, err := strconv.ParseUint(lockIdStr, 10, 64)
 					if err != nil {
@@ -423,7 +421,7 @@ $ %s query incentives rewards-estimation
 				}
 
 				// if no lockIDs are provided but an owner is provided, we query the rewards for all of the locks the owner has
-			} else if lockIdsCombined == "" && owner != "" {
+			} else if lockIdsCombined == "" {
 				lockIds = append(lockIds, ownerLocks...)
 			}
 
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -408,7 +408,7 @@ $ %s query incentives rewards-estimation
 			lockIds := []uint64{}
 			// if user doesn't provide at least one of the lock ids or owner, we don't have enough information to proceed.
 			if lockIdsCombined == "" && owner == "" {
-				err = fmt.Errorf("either one of owner flag or lock IDs must be provided")
+				return fmt.Errorf("either one of owner flag or lock IDs must be provided")
 
 				// if user provides lockIDs, use these lockIDs in our rewards estimation
 			} else if owner == "" {
--- a/x/incentives/client/cli/query.go
+++ b/x/incentives/client/cli/query.go
@@ -370,11 +370,9 @@ $ %s query incentives rewards-estimation
 		),
 		Args: cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			clientCtx, err := client.GetClientQueryContext(cmd)
-			if err != nil {
-				return err
-			}
-			queryClient := types.NewQueryClient(clientCtx)
+			var res *lockuptypes.AccountLockedLongerDurationResponse
+			ownerLocks := []uint64{}
+			lockIds := []uint64{}
 
 			owner, err := cmd.Flags().GetString(FlagOwner)
 			if err != nil {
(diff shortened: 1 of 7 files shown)


czarcas7ic:
Still having the issue [mentioned here](https://github.com/osmosis-labs/osmosis/pull/1885/files#r909126390)

For some addresses the rewards estimation is fine and others it panics.

Here is a random one I found on mintscan as an example:
`osmosisd q incentives rewards-estimation --owner osmo1lqxwdzlxvvh700sxagcm5flw2tpmp890s76lsj --end-epoch 400`

Using this branch will give you a panic (using main wont work at all since this query has been broken, I modified the query in this PR to try to get it to work as intended. When presented with just an address and no explicit lock-ids this command was supposed to utilize all locks the user has that are active, but that is not the behavior on main)

You can use `osmosisd q lockup account-locked-longer-duration osmo1lqxwdzlxvvh700sxagcm5flw2tpmp890s76lsj 1s` to see all the lock IDs and then query the rewards estimation using the lock-ids flag for each individual lock to see which locks cause this panic, but I have yet to see a clear pattern to which locks cause panics and which dont.

This address though, with multiple locks, gives a valid response:
`osmosisd q incentives rewards-estimation --owner osmo13gx9q6q8cxhym9pzyvjtkgagjqrgwxe0gcleyh --end-epoch 400`

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
 				return err
 			}
 
-			lockIdStrs := strings.Split(lockIdsCombined, ",")
-			lockIds := []uint64{}
-			for _, lockIdStr := range lockIdStrs {
-				lockId, err := strconv.ParseUint(lockIdStr, 10, 64)
+			var res *lockuptypes.AccountLockedLongerDurationResponse
```
p0mvn, line 389: Consider  declaring and initializing on line 390 like:
```
res, err := ...
```

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+				queryClientLockup := lockuptypes.NewQueryClient(clientCtx)
+
+				res, err = queryClientLockup.AccountLockedLongerDuration(cmd.Context(), &lockuptypes.AccountLockedLongerDurationRequest{Owner: owner, Duration: time.Millisecond})
 				if err != nil {
 					return err
 				}
-				lockIds = append(lockIds, lockId)
+			} else {
+				owner = ""
+			}
```
p0mvn, line 399: Why would we do this? So if the error is not nil, we set the owner to an empty string?

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+
+			ownerLocks := []uint64{}
+
+			if res != nil {
```
p0mvn, line 403: Why would `res` be nil if we already checked that there is no error above?

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+				for _, lockId := range res.Locks {
+					ownerLocks = append(ownerLocks, lockId.ID)
+				}
+			}
+
+			lockIdStrs := strings.Split(lockIdsCombined, ",")
+			lockIds := []uint64{}
+			// if user doesn't provide any lockIDs or a lock owner, we don't have enough information to proceed
```
p0mvn, line 411:
```suggestion
			// if user doesn't provide at least one of the lock ids or owner, we don't have enough information to proceed.
```

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+			if lockIdsCombined == "" && owner == "" {
+				err = fmt.Errorf("if owner flag is not set, lock IDs must be provided")
```
p0mvn, line 413: Current wording implies the presence of lock ids is conditional on the lack of owner but not vice versa
```suggestion
				err = fmt.Errorf("either one of owner flag or lock IDs must be provided")
```

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+				return err
```
p0mvn, line 414:
```suggestion
				return fmt.Errorf("if owner flag is not set, lock IDs must be provided")
```

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
+
+				// if user provides lockIDs, use these lockIDs in our rewards estimation
+			} else if lockIdsCombined != "" {
```
p0mvn, line 417: Please consider to simplify:
```
if lockIdsCombined == "" && owner == "" {
} else if lockIdsCombined == "" {
} else if owner == "" {
}
```

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,80 @@ $ %s query incentives rewards-estimation
...
 			}
 
-			res, err := queryClient.RewardsEst(cmd.Context(), &types.RewardsEstRequest{
+			// TODO: Figure out why some lock ids are good and some causes "Error: rpc error: code = Unknown desc = panic message redacted to hide potentially sensitive system info: panic"
+			// since owner checks have already been completed above, we switch the owner address to a random module account address since a blank owner panics.
+			// we should find a better way to circumvent this address validity check
+			if owner == "" {
+				owner = "osmo14kjcwdwcqsujkdt8n5qwpd8x8ty2rys5rjrdjj"
+			}
+			res1, err1 := queryClient.RewardsEst(cmd.Context(), &types.RewardsEstRequest{
```
p0mvn, line 453: Let's reuse `res` and `err` please

p0mvn (commented):
I don't have a clear suggestion but I'm wondering if it is possible to decouple the command logic from the flags provided. In my opinion, having so many edge cases makes it likely for something wrong to happen

x/incentives/client/cli/query.go
czarcas7ic: Correct! If an owner flag is set, lock ids are optional. If an owner flag is not set, lock ids are required. But both owner and lock ids can also be provided

x/incentives/client/cli/query.go
czarcas7ic: I wasn't sure, if an owner flag is not provided, does it still initialize an empty owner string? The below logic requires this so thats why I was explicit with it.

x/incentives/client/cli/query.go
czarcas7ic: Ahh good call, my bad! No need to put this within an "if" statement since we already checked above

x/incentives/client/cli/query.go
czarcas7ic: Sorry, can you explain further? My intent was to only initialize if the owner flag is present. I'm sure the way you are explaining is better, I just don't quite get what you are suggesting

x/incentives/client/cli/query.go
p0mvn: Ahh I see, I misread this and thought that this applies to:
```
if err != nil {
    return err
}
```

Anyways, a string is "" by default so we can remove this code branch

x/incentives/client/cli/query.go
czarcas7ic: Can you please show me how to reuse res if the prior definition was of a different type?

x/incentives/client/cli/query.go
p0mvn: So the suggestion applies, right?

```
if owner flag is not set, lock IDs must be provided
```
means `if !owner -> lock ids` but we also want the text to convey `if !lock ids -> owner`

x/incentives/client/cli/query.go
p0mvn: If that was the intent, I think it should be fine. I misinterpreted it, sorry

x/incentives/client/cli/query.go
czarcas7ic: Yeah you are right, I had to read it like 30 times to make sure but you are right haha

czarcas7ic: It's not allowing me to commit the suggestion since its "out of date" and I still want you to get credit for the suggestion, can you resubmit? If you don't care I will happily commit it myself.

x/incentives/client/cli/query.go
p0mvn: If `res1` is of a different type, let's choose a more meaningful name instead such as `rewardsEstimateResult`? And for `res` something like `lockedAccountsResult`? That would help to reduce the similarities with names.

`err` might still be possible to reuse

czarcas7ic: Sweet, sounds good! Just was making sure I wasn't missing some special way to redeclare a variable 👍

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,77 @@ $ %s query incentives rewards-estimation
 				return err
 			}
 
-			lockIdStrs := strings.Split(lockIdsCombined, ",")
-			lockIds := []uint64{}
-			for _, lockIdStr := range lockIdStrs {
-				lockId, err := strconv.ParseUint(lockIdStr, 10, 64)
+			var res *lockuptypes.AccountLockedLongerDurationResponse
+			if owner != "" {
+				queryClientLockup := lockuptypes.NewQueryClient(clientCtx)
```
p0mvn, line 391: Consider moving all this logic inside the if branch + `ownerLocks` creation to this branch of the following `else if`:
```
else if lockIdsCombined == "" {
    ....
}
```

They are logically equivalent and `ownerLocks` are only needed there since we are copying them all to `lockIds`

x/incentives/client/cli/query.go
czarcas7ic: Hey, I made this suggestion and the cli_test started to fail because of it. Did I misinterpret your suggestion?

x/incentives/client/cli/query.go
```diff
@@ -384,31 +386,77 @@ $ %s query incentives rewards-estimation
...
+
+				// if no lockIDs are provided but an owner is provided, we query the rewards for all of the locks the owner has
+			} else if lockIdsCombined == "" {
+				lockIds = append(lockIds, ownerLocks...)
+			}
+
+			// if lockIDs are provided and an owner is provided, only query the lockIDs that are provided
+			// if a lockID was provided and it doesn't belong to the owner, return an error
+			if len(lockIds) != 0 && owner != "" {
+				for _, lockId := range lockIds {
```
p0mvn, line 431: How can this work when no owner is provided?

If no owner is provided, then `ownerLocks` are empty and we only use locks from `lockIdStrs`. However, since `ownerLocks` are empty, `validInputLockId` would be false for any lock supplied via a flag string

x/incentives/client/cli/query.go
p0mvn: I think this comment is related to: https://github.com/osmosis-labs/osmosis/pull/1985/files#r915441993

Something feels off about how and where we are using `ownerLocks`

mattverse (commented):
I suggest we not merge this until we fix the bug in `GetCmdUpcomingGaugesPerDenom`

mattverse:
> You can use osmosisd q lockup account-locked-longer-duration osmo1lqxwdzlxvvh700sxagcm5flw2tpmp890s76lsj 1s to see all the lock IDs and then query the rewards estimation using the lock-ids flag for each individual lock to see which locks cause this panic, but I have yet to see a clear pattern to which locks cause panics and which dont.

Can you try running binary with --debug to see the specific error / panic message?

czarcas7ic:
@mattverse running with debug logs still provides the same panic message:

`Error: rpc error: code = Unknown desc = panic message redacted to hide potentially sensitive system info: panic`

I wish we didn't have this panic message, it makes debugging so difficult -_-

czarcas7ic:
> GetCmdUpcomingGaugesPerDenom

Which bug are you referring to?

ValarDragon (approved):
comment updates LGTM, and the GetRewardsEst logic. (beyond the comment I left)

Lets capture the remaining mystery problem there into an issue that we add to a backlog?

Didn't review cli_test changes in detail, but I'm also generally not concerned with the cli_tests

x/incentives/client/cli/query.go
```diff
@@ -361,45 +363,88 @@ func GetCmdRewardsEst() *cobra.Command {
...
+					ownerLocks = append(ownerLocks, lockId.ID)
+				}
+			}
+
+			// TODO: Figure out why some lock ids are good and some causes "Error: rpc error: code = Unknown desc = panic message redacted to hide potentially sensitive system info: panic"
+			// since owner checks have already been completed above, we switch the owner address to a random module account address since a blank owner panics.
+			// we should find a better way to circumvent this address validity check
+			if owner == "" {
+				owner = "osmo14kjcwdwcqsujkdt8n5qwpd8x8ty2rys5rjrdjj"
+			}
```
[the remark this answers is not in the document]
czarcas7ic: We are more so worried about the case where owner = "" && len(lockIds != 0) right? Because the below RewardsEstRequest struct requires and owner even if its not used

czarcas7ic:
According to Dev, pools that existed prior to v4 or v5 have an incorrect/negative accumulation store value. After porting the redacted error fix from the sdk, I was able to see that the accumulation store was in fact returning a negative number which was causing the error. Looking into either fixing this or merging this as is and fixing this problem in a future PR

x/incentives/client/cli/query.go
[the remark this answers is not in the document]
czarcas7ic: Fixed by adding logic to grpc_query.go

czarcas7ic:
Made it so that the est rewards query will always return a valid result and will show an error like the following within the daemon if a bugged gauge is part of the request
Screen Shot 2022-07-09 at 12 11 43 PM
Also added logic to grpc_query.go to allow for blank owner address unlike before. Please let me know if this matches the intent, if so I will merge!

ValarDragon:
Awesome, thanks!

czarcas7ic:
Backport this

Issue #1988, opened 2022-07-07
Switch `expectError` to `expectPass` where applicable to keep tests consistent with each other


AlpinYukseloglu:
## Background

Most of our table driven tests check errors by using `expectPass`, but there are a few tests lingering around that seem to use `expectError`. Since we are starting to introduce testing standards for our codebase, a small but important test cleanup would be to align all of our tests to use the same convention to remove the inconsistency.

## Suggested Design

- Switch `expectError` to `expectPass` where applicable

## Acceptance Criteria

- All existing and new tests should pass

PR #1989, opened 2022-07-07
Switch `expectError` to `expectPass` where applicable to keep tests consistent with each other

AlpinYukseloglu:
Closes: #1988 

Most of our table driven tests check errors by using `expectPass`, but there are a few tests lingering around that seem to use `expectError`. Since we are starting to introduce testing standards for our codebase, a small but important test cleanup would be to align all of our tests to use the same convention to remove the inconsistency. This PR makes this change where applicable.

## Brief Changelog

- Switch `expectError` to `expectPass` in balancer tests

## Testing and Verifying

### DIFF
--- a/tests/e2e/initialization/init_test.go
+++ b/tests/e2e/initialization/init_test.go
@@ -16,11 +16,9 @@ import (
 
 const forkHeight = 10
 
-var (
-	expectedConfigFiles = []string{
-		"app.toml", "config.toml", "genesis.json", "node_key.json", "priv_validator_key.json",
-	}
-)
+var expectedConfigFiles = []string{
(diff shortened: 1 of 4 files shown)


mattverse (commented):
I'm not against this what's introduced in the PR, but I'm also not convinced if `expectPass` should be the standard that we "must" keep across test cases, since IMO it differs from tests to tests depending on if we want to focus on the failing cases or the sucess cases.

For example, the `expectError` in this PR were mostly false and in this case, we can simply remove this field since the tests defaults to false VS having to input all values with true values.

ValarDragon:
yeah, also not sure this is worth doing right now / here

Issue #1972, opened 2022-07-05
refactor: remove legacy `handler.go` files


p0mvn:
**Background**

From each module, we should remove legacy `handler.go` files.

A similar refactor was don in the SDK that can be referenced: https://github.com/cosmos/cosmos-sdk/pull/9650

**Acceptance Criteria** 
- all `handler.go` logic is removed and refactored according to https://github.com/cosmos/cosmos-sdk/pull/9650
- all new functionality is covered by unit tests

_Originally posted by @alexanderbez in https://github.com/osmosis-labs/osmosis/pull/1956#discussion_r913264144_

alexanderbez:
literally just code deletion haha. I can tackle this

PR #1996, opened 2022-07-07
chore: remove legacy msg handlers

alexanderbez:
Remove legacy sdk.Msg handlers.

Closes: #1972


ValarDragon:
Did you get a chance to double check the `RegisterServices` registering the message service correctly for all of these?
(Not 100% on Gamm)

Issue #1797, opened 2022-06-12
Log Message - ERR no delegation distribution info


chris-remus:
## Background

During the network restart using ``v10.0.0`` our logs displayed only the following message immediately after restarting ``osmosisd``.

```ERR no delegation distribution info```

Upon [reporting this](https://discord.com/channels/798583171548840026/842550346865180682/985578355648774174) to the Osmosis Validators discord channel, @ValarDragon [explained ](https://discord.com/channels/798583171548840026/842550346865180682/985578461596880947)the message was due to "no swaps in that pool" and asked me to create this issue so the log messaging could be improved.

## Suggested Design

TBD

## Acceptance Criteria

TBD

SlingshotCade:
I too have run into this problem after upgrading to Osmosisd 10.0.0.    Once this issue shows up in the logs then the API stops responding (port 26657).    A couple times I restarted cosmovisor and it started to work again.   But then the "ERR no delegation distribution info"  shows up and then the server stops working.  Is there any more information on this?

ValarDragon:
The API is down generally due to epoch time, which is a more a longstanding problem with queries aren't available during block execution.

mattverse:
After investigation, noticed that we had this error not because the pool didn't have any swaps in between epochs, it's because we don't delete intermediary accounts. Working on a PR with the fix

ValarDragon:
Hrmm, don't we need to be careful wrt when we need to delete intermediary accounts? (IIRC theres something in the spec about this)

effofxprime:
Regarding this issue, since I also face the same error, is it possible that it's causing a performance hit that I've described here in this ticket (https://github.com/osmosis-labs/osmosis/issues/2173) ?

mattverse:
Epoch has performance hit mainly because of distribution logic and internal logics that's happening in epoch time, shouldn't have relation with the error log we see here

ValarDragon:
I believe this has been fixed

PR #2010, opened 2022-07-10
x/superfluid: no delegation distribution info

czarcas7ic:
Closes: #1797 

Whenever a pool does not have swaps occur between last epoch and the current epoch, an error message is displayed during epoch. Since we have many pools that this is the case for, there are many error logs which does not help the end user. 

More specifically, we call WithdrawDelegationRewards here

https://github.com/osmosis-labs/osmosis/blob/079bdd496aefea6ad386263b349f2b60f0358738/x/superfluid/keeper/epoch.go#L60-L65

When we make this call to the sdk, we return an ErrEmptyDelegationDistInfo

https://github.com/osmosis-labs/cosmos-sdk/blob/d49b3e7549293d3bdf70c898c7afa58728438338/x/distribution/keeper/keeper.go#L84-L94

## Brief Changelog

This change simply handles the ErrEmptyDelegationDistInfo with a debug log stating that "no swaps occurred in this pool between last epoch and this epoch" instead of spamming the daemon with error logs every epoch.

## Testing and Verifying

I tested this by running a local state exported testnet and setting the epoch time to every 60 seconds.

### DIFF
--- a/x/superfluid/keeper/epoch.go
+++ b/x/superfluid/keeper/epoch.go
@@ -5,6 +5,8 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	distributiontypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
+
 	"github.com/osmosis-labs/osmosis/v7/osmoutils"
 	gammtypes "github.com/osmosis-labs/osmosis/v7/x/gamm/types"
 	incentivestypes "github.com/osmosis-labs/osmosis/v7/x/incentives/types"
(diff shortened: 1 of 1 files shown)


x/superfluid/keeper/epoch.go
```diff
@@ -61,6 +63,10 @@ func (k Keeper) MoveSuperfluidDelegationRewardToGauges(ctx sdk.Context) {
 		// we use cacheCtx and apply the changes later
 		_ = osmoutils.ApplyFuncIfNoError(ctx, func(cacheCtx sdk.Context) error {
 			_, err := k.dk.WithdrawDelegationRewards(cacheCtx, addr, valAddr)
+			if err == distributiontypes.ErrEmptyDelegationDistInfo {
```
alexanderbez, line 66: Can we use `errors.Is`?

czarcas7ic: Just added!

alexanderbez (approved):

PR #2011, opened 2022-07-10
Fix Initgenesis bug in tokenfactory, when the denom creation fee para…

ValarDragon:
…m is set

Fixes a bug Adam pointed out, in TokenFactory genesis handling. (Which led to panics in his testnet generation)

## Brief Changelog

- Refactor TokenFactory createdenom logic, to make a direct method post-fee-charging that genesis can call
- Switch genesis to no longer charge a fee
- Add test cases to cover this.

## Testing and Verifying

- Added a case for a denom with a different admin (previously untested in InitGenesis)
- Sets the token creation fee

### DIFF
--- a/x/tokenfactory/keeper/createdenom.go
+++ b/x/tokenfactory/keeper/createdenom.go
@@ -28,7 +28,6 @@ func (k Keeper) CreateDenom(ctx sdk.Context, creatorAddr string, subdenom string
 // Runs CreateDenom logic after the charge has been handled.
 // Made into a separate method for genesis handling.
 func (k Keeper) createDenomAfterValidation(ctx sdk.Context, creatorAddr string, denom string) (err error) {
-
 	denomMetaData := banktypes.Metadata{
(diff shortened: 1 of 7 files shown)


x/tokenfactory/keeper/createdenom.go
```diff
@@ -11,34 +11,22 @@ import (
...
-	accAddr, err := sdk.AccAddressFromBech32(creatorAddr)
+	err = k.chargeForCreateDenom(ctx, creatorAddr, subdenom)
 	if err != nil {
 		return "", err
 	}
-	if len(creationFee) > 0 {
-		if err := k.distrKeeper.FundCommunityPool(ctx, creationFee, accAddr); err != nil {
-			return "", err
-		}
-	}
```
ValarDragon, line 30: Moved to chargeForCreateDenom

alexanderbez (approved):

czarcas7ic (approved):
Tested this locally on a state export testnet and it works 👍

Issue #1903, opened 2022-06-30
x/gamm: Charge gas for computation in swaps instead of just I/O


AlpinYukseloglu:
Swaps currently only charge gas for I/O, regardless of the amount of computation they take up. One minor but useful change would be to make it so that gas is also charged for computation in swaps.
[
](https://github.com/osmosis-labs/osmosis/pull/1858)

AlpinYukseloglu:
> What exactly do you have in mind @AlpinYukseloglu? Some fixed cost per specific operation?

Yes, some fixed cost for now with a TODO to turn it into a param later (since making it a param now would require migration code and it's too small of a change to be worth it for that)

ValarDragon:
Agreed, lets add a fixed cost for now.

rrrliu:
@AlpinYukseloglu is anyone working on this right now? If not, happy to take this!

AlpinYukseloglu:
> @AlpinYukseloglu is anyone working on this right now? If not, happy to take this!

Nope, feel free to take it!

PR #2016, opened 2022-07-11
x/gamm: Add fixed gas cost for swaps

rrrliu:
Closes: #1903 

To further prevent spam, adds fixed amount of gas we consume per swap in/out. Previously we had only been consuming gas for built-in I/O (store reads and writes), and in the future we might make 

## Brief Changelog

- Added new `BalancerGasFeeForSwap` constant + TODO comment to turn into param
- Have `swapExactAmountIn` and `swapExactAmountOut` consume gas
- Test this in `swap_test.go`

## Testing and Verifying

- Added lines to test gas consumed in `swap_test.go`


x/gamm/types/constants.go
```diff
@@ -12,6 +12,8 @@ const (
 	// Raise 10 to the power of SigFigsExponent to determine number of significant figures.
 	// i.e. SigFigExponent = 8 is 10^8 which is 100000000. This gives 8 significant figures.
 	SigFigsExponent = 8
+	// TODO: Current fixed cost gas fee per swap -- turn this into a param in the future.
+	GasFeeForSwap = 10000
```
alexanderbez, line 16: Is there a reason to not directly just make this a param?

x/gamm/keeper/swap.go
```diff
@@ -49,6 +49,7 @@ func (k Keeper) swapExactAmountIn(
 	}
 	tokensIn := sdk.Coins{tokenIn}
 
+	ctx.GasMeter().ConsumeGas(types.GasFeeForSwap, "swap computation")
```
alexanderbez, line 52: It feels safer to move the gas consumption to the top-most level of abstraction, i.e the `msg` handler for swapping.

x/gamm/types/constants.go
AlpinYukseloglu: We decided to make it a fixed cost for now to avoid  needing to set up migration code for a change this small: https://github.com/osmosis-labs/osmosis/issues/1903#issuecomment-1173907587

x/gamm/types/constants.go
```diff
@@ -12,6 +12,8 @@ const (
 	// Raise 10 to the power of SigFigsExponent to determine number of significant figures.
 	// i.e. SigFigExponent = 8 is 10^8 which is 100000000. This gives 8 significant figures.
 	SigFigsExponent = 8
+	// TODO: Current fixed cost gas fee per swap -- turn this into a param in the future.
```
AlpinYukseloglu, line 15:
```suggestion
	// TODO: Turn this into a param in the future.
```

x/gamm/keeper/swap_test.go
```diff
@@ -188,11 +192,15 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleSwapExactAmountOut() {
...
 
+			prevGasConsumed := suite.Ctx.GasMeter().GasConsumed()
 			tokenInAmount, err := keeper.SwapExactAmountOut(suite.Ctx, suite.TestAccs[0], poolId, test.param.tokenInDenom, test.param.tokenInMaxAmount, test.param.tokenOut)
 			suite.NoError(err, "test: %v", test.name)
 			suite.True(tokenInAmount.Equal(test.param.expectedTokenInAmount),
 				"test: %v\n expect_eq actual: %s, expected: %s",
 				test.name, tokenInAmount, test.param.expectedTokenInAmount)
+			gasConsumedForSwap := suite.Ctx.GasMeter().GasConsumed() - prevGasConsumed
+			// We consume 10,000 gas directly, so the extra I/O operation mean we end up consuming more.
+			suite.Assert().Greater(gasConsumedForSwap, uint64(10000))
```
AlpinYukseloglu, line 203: Since we plan on turning the fixed cost into a param/will likely change it at some point, I wonder if it would make sense to reference `GasFeeForSwap` in `x/gamm/types/constants.go` directly here instead of hard-coding 10,000?

x/gamm/keeper/swap.go
ValarDragon: I guess really it should probably be defined per pool type

ValarDragon: That way we have it per Balancer swap, rather than per total swap. (And can then later charge a different gas amount for stableswap / next CFMM)

rrrliu: Yeah that makes sense, just updated! Should we update balancer pool docs with this or should we wait until it becomes a param?

mattverse (approved):
LGTM! Can we add a Changelog entry for this PR?

mattverse:
Also out of curiousity, how was the constant 10000 decided?

x/gamm/types/constants.go
```diff
@@ -12,6 +12,8 @@ const (
 	// Raise 10 to the power of SigFigsExponent to determine number of significant figures.
 	// i.e. SigFigExponent = 8 is 10^8 which is 100000000. This gives 8 significant figures.
 	SigFigsExponent = 8
+	// TODO: Turn this into a param in the future.
```
p0mvn, line 15: Great idea! Can we make an issue for this please and remove the comment?

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+* [#2016](https://github.com/osmosis-labs/osmosis/pull/2016) Add fixed 10000 gas cost for each Balancer swap
```
alexanderbez, line 44:
```suggestion

* [#2016](https://github.com/osmosis-labs/osmosis/pull/2016) Add fixed 10000 gas cost for each Balancer swap
```

x/gamm/types/constants.go
rrrliu: Created one! https://github.com/osmosis-labs/osmosis/issues/2058

rrrliu:
> Also out of curiousity, how was the constant 10000 decided?

@mattverse oh I just benchmarked the swap i/o gas and it was around ~40000, so decided to have it be something around the same order of magnitude. No strong opinion on this though so open to changing.

ValarDragon (approved):

ValarDragon:
Thanks for making this!

mattverse:
@rrrliu ah gotcha thanks for explaining! Nice work done here!

Issue #2039, opened 2022-07-12
fix(e2e): flakiness when upgrading containers for `chainB`


p0mvn:
## Background

e2e has recently had several refactors. The latest change #1999 has likely caused test flakiness.

Info from @czarcas7ic :
- Failing run: https://github.com/osmosis-labs/osmosis/actions/runs/2645267194/attempts/1
- Failing line:
https://github.com/osmosis-labs/osmosis/blob/938f9bdb4ce05e178340b63c25452505ff7c6a3d/tests/e2e/configurer/upgrade.go#L250
- It always fails when upgrading containers for `chainB`

Info from @p0mvn :
- Impossible to reproduce locally, tried re-running 10 times
- #1999 is the most likely cause

## Acceptance Criteria
- investigate and fix e2e test flakiness
- Re-running CI 10 times does not cause an issue (try making redundant changes to trigger)

p0mvn:
Added more logs here: https://github.com/p0mvn/osmosis/pull/14

Trying to manually trigger e2e in CI multiple times to repro this

p0mvn:
I have 2 updates on this:

1. I was not able to reproduce in CI with extra logs. Tried running 10 times on my fork: https://github.com/p0mvn/osmosis/runs/7311008352?check_suite_focus=true

2. The first update should not be a big problem because #2040 refactors the logic for waiting for a certain height. Instead of using CLI, it now uses Tendermint RPC which should be more reliable and easier to debug. We need this refactor for the next step in state-sync so it might address 2 problems at once.

p0mvn:
Recent e2e flakiness: https://github.com/osmosis-labs/osmosis/runs/7351687619?check_suite_focus=true

https://github.com/osmosis-labs/osmosis/pull/2078

czarcas7ic:
Another recent instance: https://github.com/osmosis-labs/osmosis/runs/7373637913?check_suite_focus=true

This was at initialization though

czarcas7ic:
Another instance: https://github.com/osmosis-labs/osmosis/runs/7458951335?check_suite_focus=true

p0mvn:
Had a chat with @nikever about this. The plan is to get something going with the self-hosted runners and retain container logs to improve the debugging experience

czarcas7ic:
Awesome, self hosted runners are for sure a game changer

czarcas7ic:
[...]

PR #2040, opened 2022-07-12
refactor(e2e): run individual nodes; rely on rpc client in setup

p0mvn:
Next step in: https://github.com/osmosis-labs/osmosis/issues/1879
Related to: https://github.com/osmosis-labs/osmosis/issues/2039

This change includes the following interconnected parts:

1. Ability to run individual nodes by calling `chainConfig.RunNode(...)`
We need the ability to run each node individually for testing state sync. This refactor implements it.

2. Use the Tendermint RPC client in setup
Switch from relying on CLI during setup to using Tendermint RPC client. It should be more reliable and easier to debug than the CLI. This might help to resolve or mitigate issues in #2039 

3. Reduce code duplication
We used to have code duplicated for waiting for a certain height:
- upgrade containers: https://github.com/osmosis-labs/osmosis/blob/5f17b1941d85e1bf906370327a635733bce2a65f/tests/e2e/configurer/upgrade.go#L239-L258
- halt at upgrade height: https://github.com/osmosis-labs/osmosis/blob/5f17b1941d85e1bf906370327a635733bce2a65f/tests/e2e/configurer/upgrade.go#L148-L178
- wait to reach fork height: https://github.com/osmosis-labs/osmosis/blob/5f17b1941d85e1bf906370327a635733bce2a65f/tests/e2e/configurer/upgrade.go#L207-L221

This is now reduced to [`WaitUntilHeight`](https://github.com/osmosis-labs/osmosis/blob/ad093356b21fac3dfbf3135e37efd5f7397b2861/tests/e2e/configurer/chain/chain.go#L124)

### DIFF
--- a/tests/e2e/configurer/chain/chain.go
+++ b/tests/e2e/configurer/chain/chain.go
@@ -40,9 +40,9 @@ type syncInfo struct {
 
 const (
 	// waitUntilRepeatPauseTime is the time to wait between each check of the node status.
-	waitUntilRepeatPauseTime = 500 * time.Millisecond
+	waitUntilRepeatPauseTime = 2 * time.Second
 	// waitUntilrepeatMax is the maximum number of times to repeat the wait until condition.
-	waitUntilrepeatMax = 20
+	waitUntilrepeatMax = 60
 )
 
 func New(t *testing.T, containerManager *containers.Manager, id string, initValidatorConfigs []*initialization.NodeConfig) *Config {
--- a/tests/e2e/configurer/chain/chain.go
+++ b/tests/e2e/configurer/chain/chain.go
@@ -57,7 +57,7 @@ func New(t *testing.T, containerManager *containers.Manager, id string, initVali
 }
 
 // RunNode runs a node container for the given nodeIndex.
-// The node configuration must be already added to chain prior to calling this
+// The node configuration must be already added to the chain config prior to calling this
 // method.
 func (c *Config) RunNode(nodeIndex int) error {
 	c.t.Logf("starting %s validator containers...", c.Id)
--- a/tests/e2e/configurer/chain/chain.go
+++ b/tests/e2e/configurer/chain/chain.go
@@ -102,9 +102,7 @@ func (c *Config) RunNode(nodeIndex int) error {
 func (c *Config) WaitUntil(nodeIndex int, doneCondition func(syncInfo coretypes.SyncInfo) bool) error {
 	var latestBlockHeight int64
 	for i := 0; i < waitUntilrepeatMax; i++ {
-		ctx, cancel := context.WithTimeout(context.Background(), waitUntilRepeatPauseTime)
-		defer cancel()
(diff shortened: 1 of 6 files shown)


tests/e2e/configurer/chain/chain.go
```diff
@@ -41,3 +55,96 @@ func New(t *testing.T, containerManager *containers.Manager, id string, initVali
 		containerManager:     containerManager,
 	}
 }
+
+// RunNode runs a node container for the given nodeIndex.
+// The node configuration must be already added to chain prior to calling this
```
p0mvn, line 60:
```suggestion
// The node configuration must be already added to the chain config prior to calling this
```

tests/e2e/configurer/chain/chain.go
```diff
@@ -31,6 +38,13 @@ type syncInfo struct {
 	SyncInfo status `json:"SyncInfo"`
 }
 
+const (
+	// waitUntilRepeatPauseTime is the time to wait between each check of the node status.
+	waitUntilRepeatPauseTime = 5 * time.Second
```
czarcas7ic, line 43: what is the purpose of waiting between each query? I feel like it would be better to just spam the node until valid query instead of waiting potentially five extra seconds for the valid response. not a big deal though

tests/e2e/configurer/chain/chain.go
```diff
@@ -41,3 +55,96 @@ func New(t *testing.T, containerManager *containers.Manager, id string, initVali
...
+		ctx, cancel := context.WithTimeout(context.Background(), waitUntilRepeatPauseTime)
+		defer cancel()
+		status, err := c.NodeConfigs[nodeIndex].rpcClient.Status(ctx)
+		if err != nil {
+			return err
+		}
+		latestBlockHeight = status.SyncInfo.LatestBlockHeight
+		// let the node produce a few blocks
+		if !doneCondition(status.SyncInfo) {
+			time.Sleep(waitUntilRepeatPauseTime)
```
czarcas7ic, line 114: yeah imo we could just remove this line and increase the repeat max. Probably wont shed a ton of time off but  seconds can add up.

czarcas7ic (approved):
Nice work 👍 Excited to see this soon culminate with the addition of state sync next 🚀

tests/e2e/configurer/chain/chain.go
p0mvn: Spamming is prone to causing undetermined behavior. Although Tednermint doesn't have it, there might be rate-limiting on GitHub runners.

5 seconds might be an overkill though will reduce it in a follow-up PR.

p0mvn: 500ms applied

p0mvn: Actually, seems that 500 ms is too fast. Times out before reaching success condition. I will try 2s and 40 times repeat

PR #2043, opened 2022-07-13
e2e: Reduce repetitive args for tx creation commands

ValarDragon:
Closes: #1623

Quick PR to reduce boilerplate in e2e, stemming from CLI.

## Testing and Verifying

### DIFF
--- a/tests/e2e/containers/containers.go
+++ b/tests/e2e/containers/containers.go
@@ -47,9 +47,10 @@ func NewManager(isUpgrade bool, isFork bool) (docker *Manager, err error) {
 }
 
 // ExecTxCmd Runs ExecCmd, with flags for txs added.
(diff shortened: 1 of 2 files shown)


tests/e2e/containers/containers.go
```diff
@@ -46,6 +46,15 @@ func NewManager(isUpgrade bool, isFork bool) (docker *Manager, err error) {
 	return docker, nil
 }
 
+// ExecTxCmd Runs ExecCmd, with flags for txs added.
+// namely `--chain-id={chain-id} -b=block --yes --keyring-backend=test`, and searching for `code: 0`
```
ValarDragon, line 50: Can we add --log-format=json to this? Its not added on all the CLI commands, but I assume this is not intentional

czarcas7ic (approved):

PR #2044, opened 2022-07-13
e2e: move endpoint complexity into config.QueryRPC

ValarDragon:

Minor commit, moves more complexity around endpoints into QueryRPC

Driveby fix of changing the addr arg of QueryPropTally to be the correct proposalNumber

### DIFF
--- a/tests/e2e/configurer/chain/queries.go
+++ b/tests/e2e/configurer/chain/queries.go
@@ -27,7 +27,6 @@ func (c *Config) QueryRPC(validatorIdx int, path string) ([]byte, error) {
 	endpoint := fmt.Sprintf("http://%s", hostPort)
 	fullQueryPath := fmt.Sprintf("%s/%s", endpoint, path)
 
-	// TODO: Document why we have this loop
 	var resp *http.Response
 	retriesLeft := 5
 	for {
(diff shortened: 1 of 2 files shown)


tests/e2e/configurer/chain/queries.go
```diff
@@ -20,12 +20,18 @@ import (
...
-func (c *Config) QueryRPC(path string) ([]byte, error) {
-	var err error
+func (c *Config) QueryRPC(validatorIdx int, path string) ([]byte, error) {
+	// add the URL for the given validator ID, and pre-pend to to path.
+	hostPort, err := c.containerManager.GetValidatorHostPort(c.Id, validatorIdx, "1317/tcp")
+	require.NoError(c.t, err)
+	endpoint := fmt.Sprintf("http://%s", hostPort)
+	fullQueryPath := fmt.Sprintf("%s/%s", endpoint, path)
+
+	// TODO: Document why we have this loop
```
p0mvn, line 30: We can have a `require.Eventually()` for this instead. Going to make an issue right away so we can remove
```suggestion
```

p0mvn (approved):
Nice cleanup!

tests/e2e/configurer/chain/queries.go
p0mvn: https://github.com/osmosis-labs/osmosis/issues/2051

Issue #1995, opened 2022-07-07
x/incentives: table driven tests


czarcas7ic:
## Background

distribute_test.go within the x/incentives/keeper folder had a lot of TODOs to re-write some of the tests to make them table driven. I have removed these TODOs in favor of this issue.

## Suggested Design

- Review each test within distribute_test.go
- If test would benefit from being changed to table driven, change that test

## Acceptance Criteria

- All existing and new tests should pass

PR #2046, opened 2022-07-13
Add table-driven distribute test

catShaark:
Closes: #1995

Change `TestGetModuleToDistributeCoins` and `TestGetModuleDistributedCoins` into table driven tests

`SetupNewGauge` and `SetupNewGaugeWithDenom` now takes in an additional param `numEpochsPaidOver`


x/incentives/keeper/distribute_test.go
```diff
@@ -177,80 +185,107 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
 
-	// check that the sum of coins yet to be distributed is nil
-	coins := suite.App.IncentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
-	suite.Require().Equal(coins, sdk.Coins(nil))
+	tests := []struct {
+		// each sdk.Coins in initialGaugeCoins will be used to create a new gauge
+		initialGaugeCoins []sdk.Coins
+		// sdk.Coins to add to gauge with gauge id
+		coinsToAddToGauges map[uint64]sdk.Coins
+	}{
```
ValarDragon, line 198: Should we add locks to make as a field here, and expected token amounts in addr by epoch num?

catShaark: this test only takes into account of total coins distributed and to be distributed, should we also check for reward per accounts. If so I think we should do that in a different test.

x/incentives/keeper/distribute_test.go
```diff
@@ -177,80 +185,107 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
+	lockOwner := sdk.AccAddress([]byte("addr1---------------"))
 
-	// check that the sum of coins yet to be distributed is nil
-	coins := suite.App.IncentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
-	suite.Require().Equal(coins, sdk.Coins(nil))
+	tests := []struct {
+		// each sdk.Coins in initialGaugeCoins will be used to create a new gauge
+		initialGaugeCoins []sdk.Coins
+		// sdk.Coins to add to gauge with gauge id
+		coinsToAddToGauges map[uint64]sdk.Coins
```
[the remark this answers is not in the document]
catShaark: it is small in the original test, I'll make it larger

x/incentives/keeper/distribute_test.go
```diff
@@ -177,80 +185,107 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
-	suite.SetupTest()
+// TestGetModuleToDistributeAndDistributedCoins tests the sum of coins yet to be distributed
+// and the distributed coins is correct.
+func (suite *KeeperTestSuite) TestGetModuleToDistributeAndDistributedCoins() {
+	lockOwner := sdk.AccAddress([]byte("addr1---------------"))
 
-	// check that the sum of coins yet to be distributed is nil
-	coins := suite.App.IncentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
-	suite.Require().Equal(coins, sdk.Coins(nil))
+	tests := []struct {
```
stackman27, line 193: nit: could you please add names to the tests? Just so it's clear to follow

catShaark: sure, I'll add some comment at each test cases, you think that's okay?

x/incentives/keeper/distribute_test.go
```diff
@@ -177,80 +185,107 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
 
 }
 
-// TestGetModuleToDistributeCoins tests the sum of coins yet to be distributed for all of the module is correct.
-func (suite *KeeperTestSuite) TestGetModuleToDistributeCoins() {
-	suite.SetupTest()
+// TestGetModuleToDistributeAndDistributedCoins tests the sum of coins yet to be distributed
+// and the distributed coins is correct.
+func (suite *KeeperTestSuite) TestGetModuleToDistributeAndDistributedCoins() {
```
p0mvn, line 190: Why are we merging tests for `GetModuleDistributedCoins(...)` and `GetModuleToDistributeCoins(...)`?

Although the setup logic / testcases might be similar, they should still have separate fixtures. A test needs to test one unit of computation and do it well.

Also, I'm not seeing where `GetModuleDistributedCoins(...)` is now tested? I think it might have been removed and untested in general.

catShaark: ohhh, that is indeed missing, I'll add it

x/incentives/keeper/distribute_test.go
```diff
@@ -177,80 +185,107 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
 
-	// setup a non perpetual lock and gauge
-	_, gaugeID, _, startTime := suite.SetupLockAndGauge(false)
+		for gaugeID, coins := range tc.coinsToAddToGauges {
+			suite.AddToGauge(coins, gaugeID)
 
-	// check that the sum of coins yet to be distributed is equal to the newly created gaugeCoins
-	coins = suite.App.IncentivesKeeper.GetModuleDistributedCoins(suite.Ctx)
-	suite.Require().Equal(coins, sdk.Coins(nil))
+			moduleToDistributeCoins = suite.App.IncentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
```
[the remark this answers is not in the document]
catShaark: thanks, I'll make the change accordingly

x/incentives/keeper/distribute_test.go
catShaark: I've separated the 2 tests

x/incentives/keeper/distribute_test.go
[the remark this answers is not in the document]
catShaark: enlargement done

x/incentives/keeper/distribute_test.go
```diff
@@ -171,98 +178,238 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
-	suite.Require().NoError(err)
-	suite.Require().Equal(distrCoins, sdk.Coins{sdk.NewInt64Coin("stake", 105)})
+			gauges = append(gauges, gauge)
+
+			moduleToDistributeCoins = incentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
+			expModuleToDistributeCoins = expModuleToDistributeCoins.Add(gaugeCoins...)
+			suite.Equal(expModuleToDistributeCoins, moduleToDistributeCoins, "test %v", tc.name)
+
+			// div by 2 since all gauge in this tests distribute over 2 epochs
+			expDistributedCoinsFirstEpoch[gauge.Id] = DivCoin(coins, 2)
```
mattverse, line 260: Curious, what happens to roundings when perfect division is not possible? Ex) three epochs with gauge of 1000 stake coins. 

Can we add test case field with epochs, add tests with cases that can't be perfectly divided or test cases with epoch 0?

catShaark: yeah, you're right, how about we have odd gauge coins and 2 epochs ?

czarcas7ic: Although it shouldn't matter, maybe we also add odd epochs and even gauge just to cover all our bases (in case of future refactor of distribution logic)

catShaark: test cases added, also I've changed `SetupNewGaugeWithDenom` and  `SetupNewGauge` so that they takes in additional param `numEpochsPaidOver`

czarcas7ic (approved):

alexanderbez (commented):
Is there a reason we removed the `Require` statements?

catShaark:
> Is there a reason we removed the `Require` statements?

to reduce verbosity, said p0mvn

alexanderbez:
> > Is there a reason we removed the `Require` statements?
> 
> to reduce verbosity, said p0mvn

I'm not following? How does `Require()` add verbosity? We want to be consistent in our test usage and have tests exit early as possible upon failures IMO.

p0mvn:
> > > Is there a reason we removed the `Require` statements?
> > 
> > 
> > to reduce verbosity, said p0mvn
> 
> I'm not following? How does `Require()` add verbosity? We want to be consistent in our test usage and have tests exit early as possible upon failures IMO.

I assumed that `suite.Error()` and `suite.Require().Error()` are equivalent. That was the motivation for the suggestion. However, after seeing @alexanderbez 's comment, I checked that the `suite.Require()` does indeed fail immediately contrary to the alternative.

Knowing this, I think my earlier suggestion isn't what we want. Apologies for the confusion @catShaark and thanks @alexanderbez

mattverse:
Let's merge this upon reverting excluding the `Require()` :) !

catShaark:
> Let's merge this upon reverting excluding the `Require()` :) !

changes reverted

x/incentives/keeper/distribute_test.go
```diff
@@ -177,92 +184,261 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
+			gauges = append(gauges, gauge)
+
+			moduleToDistributeCoins = incentivesKeeper.GetModuleToDistributeCoins(suite.Ctx)
+			expModuleToDistributeCoins = expModuleToDistributeCoins.Add(gaugeCoins...)
+			suite.Require().Equal(expModuleToDistributeCoins, moduleToDistributeCoins, "test %v", tc.name)
+
+			expDistributedCoinsFirstEpoch[gauge.Id] = coins
+		}
+
+		for gaugeID, coins := range tc.coinsToAddToGauges {
```
p0mvn, line 278: Although `GetModuleToDistributeCoins` is consistently used here, I feel like there are several goals packages into a single test case: 
- test GetModuleToDistributeCoins
- test  gauge creation
- test  adding coins to gauge
- test  moving gauge from upcoming to active

Does it make sense to have a separate fixture with its own setup for each goal listed above?

catShaark: I'm good with either options.

x/incentives/keeper/distribute_test.go
```diff
@@ -177,92 +184,261 @@ func (suite *KeeperTestSuite) TestSyntheticDistribute() {
...
+				1: sdk.NewCoins(sdk.NewInt64Coin("notstake", 1000384571)),
+				2: sdk.NewCoins(sdk.NewInt64Coin("stake", 500222222)),
+			},
+			numEpochsPaidOver: 3,
+		},
+	}
 
-	// setup a non perpetual lock and gauge
-	_, gaugeID, _, startTime := suite.SetupLockAndGauge(false)
+	for _, tc := range tests {
```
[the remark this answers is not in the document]
catShaark: thanks, I'll make the changes needed accrodingly

x/incentives/keeper/distribute_test.go
p0mvn: Could you help me understand why we need one large test? I'm totally okay with it as long as I understand the reasons behind this choice.

In my view of things, a unit test is called this way because it is meant to test a single unit of computation (a function or a method). Here, the test is called `TestGetModuleToDistributeCoins` implying that it tests `GetModuleToDistributeCoins`. However, there is a lot of other functionality being tested. As a result, I'm having a hard time evaluating what it is that is the goal of this test.

In particular, I don't understand why do we do both ` suite.SetupNewGauge()` and `suite.AddToGauge()` while testing `GetModuleToDistributeCoins`. If the goal is to test `GetModuleToDistributeCoins` only,  having `suite.SetupNewGauge()` is enough IMO.

Additionally, moving gauges to active and distributing feels like a completely separate test case from just testing  `GetModuleToDistributeCoins`.

It feels to me that each for loop in `TestGetModuleToDistributeCoins` deserves its own fixture like I described above:
- basic test for GetModuleToDistributeCoins
- test gauge creation
- test adding coins to gauge
- test moving gauge from upcoming to active and distributing

In my opinion, right now, a single large test is doing too many things.

Please let  me know what you think

p0mvn: You've been doing fantastic with this PR @catShaark ! Thanks for all the thoroughness in applying and discussing suggestions

catShaark:
So I want to use `BeginNewBlock` to execute epoch instead of distribute manually, but right now `BeginNewBlock` with execute epoch doesn't work as expected ( it won't trigger gauge distribution in incentives test). I explain the problem in this issue here, @p0mvn can you take a look https://github.com/osmosis-labs/osmosis/issues/2241

catShaark:
https://github.com/osmosis-labs/osmosis/issues/2327, I have some questions needed answer before I can actually finish this issue.

mattverse:
@catShaark answered! I think ur understanding there is correct!

x/incentives/keeper/distribute_test.go
```diff
@@ -14,6 +14,13 @@ import (
 
 var _ = suite.TestingSuite(nil)
 
+func DivCoin(coins sdk.Coins, divisor int64) sdk.Coins {
```
alexanderbez, line 17:
```suggestion
func QuoCoins(coins sdk.Coins, divisor int64) sdk.Coins {
```

x/incentives/keeper/suite_test.go
```diff
@@ -196,8 +194,8 @@ func (suite *KeeperTestSuite) setupNewGaugeWithDenom(isPerpetual bool, coins sdk
 }
 
 // SetupNewGaugeWithDenom creates a gauge with the specified duration and denom.
-func (suite *KeeperTestSuite) SetupNewGaugeWithDenom(isPerpetual bool, coins sdk.Coins, denom string) (uint64, *types.Gauge, sdk.Coins, time.Time) {
-	return suite.setupNewGaugeWithDenom(isPerpetual, coins, defaultLockDuration, denom)
+func (suite *KeeperTestSuite) SetupNewGaugeWithDenom(isPerpetual bool, coins sdk.Coins, denom string, numEpochsPaidOver uint64) (uint64, *types.Gauge, sdk.Coins, time.Time) {
+	return suite.setupNewGaugeWithDenom(isPerpetual, coins, defaultLockDuration, denom, numEpochsPaidOver)
```
ValarDragon, line 198: This feels odd to me, num epochs paid over only makes sense as an arg if `isPerpetual=false`

Perhaps make a new constructor for `SetupNewNonPerpetualGauge`?

catShaark: yes, I'll do it

catShaark: can you check this [issue](https://github.com/osmosis-labs/osmosis/issues/2327) and tell if this a problem or not

czarcas7ic:
Hey @catShaark! What specific questions / unknowns are present for this PR? I want to ensure there are no blockers on our side for this.

catShaark:
@czarcas7ic  , I have some [questions](https://github.com/osmosis-labs/osmosis/issues/2327) needed answer before I can actually finish this issue.

PR #2052, opened 2022-07-13
Move simulator test to tests package, temporarily disable from CI until determinism is fixed

ValarDragon:

After merging #1817, we found out there is simulator non-determinism, and some bug found with the lockup module. (Unclear if simulation bug, or real bug)

## Brief Changelog
 
  - Moves simulator to tests package
  - Disables (temporarily) simulator from CI


simulation/types/action.go
```diff
@@ -92,7 +92,8 @@ func (m msgBasedAction) Execute(sim *SimCtx, ctx sdk.Context) (
 	}
 	tx, err := sim.txbuilder(ctx, msg, m.name)
 	if err != nil {
-		return simulation.NoOpMsg(m.name, m.name, "unable to build tx"), nil, err
+		panic(err)
+		return simulation.NoOpMsg(m.name, m.name, fmt.Sprintf("unable to build tx due to: %v", err)), nil, err
```
czarcas7ic, line 96:
```suggestion
		return simulation.NoOpMsg(m.name, m.name, fmt.Sprintf("unable to build tx due to: %v", err)), nil, err
```

PR #2053, opened 2022-07-13
refactor(e2e):  improve abstractions for node, chain and containers

p0mvn:
Part of: https://github.com/osmosis-labs/osmosis/issues/1879

Some functionality was defined on `chain.Config` abstractions such as:
- `RunNode(< node index >)`
- `QueryRPC(< node index >, ...)`
- Many commands, for example: `VoteNoProposal(< nodeIndex >)`

This required providing redundant parameters for the node index. Majority of the time we don't care about which node to query.

Therefore, these methods were redefined on `NodeCondig` instead so that we can do:
```
node.Run()
node.QueryRPC(...)
node.VoteProposal()
```
instead of 
```
chain.Run(nodeIndex)
chain.QueryRPC(nodeIndex, ...)
chain.VoteProposal(nodeIndex, )
```

The new way is also helpful for the next step in state-sync as it simplifies the process of starting up a new node.

Additionally, in the container manager, we used to be indexing container resources by chain id + node name. For every node, the node name already includes the chain id. Therefore, there will never be overlap in names between the two chains. As a result, we can remove the chainId index to simplify the container manager.

Other minor changes include cleaning up code comments and variable names to make sure they still make sense.

The majority of the logic here is shifted around for changing abstractions. No new logic is implemented.

## Testing and Verifying


tests/e2e/configurer/chain/queries.go
```diff
@@ -110,33 +84,40 @@ func (c *Config) QueryPropTally(validatorIdx int, proposalNumber int) (sdk.Int,
...
+	block, err := n.rpcClient.Block(context.Background(), &height)
+	if err != nil {
+		return "", err
+	}
+	return block.BlockID.Hash.String(), nil
+}
+
+// QueryCurrentHeight returns current of the node or error.
+func (n *NodeConfig) QueryCurrentHeight() (int64, error) {
+	status, err := n.rpcClient.Status(context.Background())
```
czarcas7ic, line 118:
```suggestion
// QueryCurrentHeight returns the current block height of the node or error.
func (n *NodeConfig) QueryCurrentHeight() (int64, error) {
	status, err := n.rpcClient.Status(context.Background())
```

czarcas7ic (approved):
👍

PR #2057, opened 2022-07-14
Commit timeout

faddat:

Reduce osmosis block times to about three seconds. 

* #2009 compliments this and has been used by notional and others since evmos upgraded.  Those using it and not are pretty strikingly visible here:

https://ping.pub/evmos/uptime

The trouble with the tm db branch is it is enormous and it is now very different from the mainline. I don't know if it will ever become the mainline. Here is the branch:

https://github.com/tendermint/tm-db/pull/237

This is still very much draft work, and doesn't help with wall clock time at all yet. 

## Brief Changelog

* make init.go set the commit timeout to two seconds
* fix the commit timeout in root.go to two seconds

Please note that these are the only changes that have been made. I am treating additional changes that may be needed because of shorter block times is out of scope for this PR.  I also think that it's possible that there are not additional changes needed.

## Testing and Verifying

The best test of this code will be to run it on the osmosis test net.  It has not yet been tested like that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Features
 
+* [#2057](https://github.com/osmosis-labs/osmosis/pull/2057) Reduce block times to about three seconds
 * [#1312] Stableswap: Createpool logic 
 * [#1230] Stableswap CFMM equations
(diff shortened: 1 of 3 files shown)


ValarDragon (approved):
This LGTM! Have you been trying it out on mainnet, whats the precommit inclusion like on blocks you produce?

faddat:
Actually yes, and in fact it turned out to work so well that we didn't notice the difference

PR #2059, opened 2022-07-14
feat: Delete intermediary accounts

### LINKED ISSUE #1797: Log Message - ERR no delegation distribution info

mattverse:
Closes #1797 

We currently do not delete intermediary accounts without any delegations, instead keeping them all in store. 
The problem is that we even iterate them during epoch, which is the main cause of the log `ERR no delegation distribution info`, as we're trying to get rewards from delegations that doesn't exist.

This PR introduces logic to deleting intermediary accounts, where we delete an intermediary account when there's no intermediary account connection and delegations from the intermediary account.

Upon my scope of investigation, I do not find any points that can be troublesome when deleting an intermediary account, although there were potential concerns for deleting intermediary account. (Note that we're not deleting the account itself using accountKeeper, we're only deleting this from the superfluid module state)

This PR achieves the goal of deleting unused intermediary accounts by adding the deleting logic and check inside EndBlocker of the superfluid module. (This has to be done in superfluid module not the lockup module because of go lang circular dependency issue. We can get an get around with this via interfaces but tbd upon reveiws)

## Brief Changelog

- Add logic for checking and deleting unused intermediary accounts in `EndBlocker` of the superfluid module.

## Testing and Verifying

Added test cases


x/superfluid/abci.go
```diff
@@ -17,3 +19,10 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper, ek types.EpochKeeper) {
 		k.AfterEpochStartBeginBlock(ctx)
 	}
 }
+
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) []abci.ValidatorUpdate {
```
mattverse, line 23: Question to reviewers: How do we set orderings of Endblockers between modules?

app/modules.go
```diff
@@ -164,6 +164,10 @@ func orderBeginBlockers(allModuleNames []string) []string {
 
 func OrderEndBlockers(allModuleNames []string) []string {
 	ord := partialord.NewPartialOrdering(allModuleNames)
+
+	// superfluid should come after lockup module because matured synth locks
+	// has to be deleted before superfluid endblocker
```
p0mvn, line 169:
```suggestion
	// have to be deleted before superfluid EndBlocker.
```

x/superfluid/module.go
```diff
@@ -187,6 +187,7 @@ func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
 // EndBlock executes all ABCI EndBlock logic respective to the capability module. It
 // returns no validator updates.
 func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	EndBlocker(ctx, am.keeper)
```
p0mvn, line 190: What is the reason for having this `EndBlocker` function? Why can't we run that logic directly here?

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
 	prefixStore := prefix.NewStore(store, types.KeyPrefixIntermediaryAccount)
-	prefixStore.Delete(address)
+	prefixStore.Delete(intermedairyAcc.GetAccAddress())
+}
+
+// IntermediaryAccountDelegationsExists returns true if the gien intermediary account has any delegations.
+// We check this by
+// - using staking keeper to check if actual delegations exist,
+// - checking if there are any intermediary account connection remaining to the intermediaryAcc
+// - if there's no synthetic lock with the intermediaryAccount.Denom + intermediaryAccount.ValAddr combination
```
p0mvn, line 140: nit: some of the points were read as "check by checking"
```suggestion
// We determine this by
// - using staking keeper to check if actual delegations exist,
// - checking if there are any intermediary account connection remaining to the intermediaryAcc
// - checking if there's no synthetic lock with the intermediaryAccount.Denom + intermediaryAccount.ValAddr combination
```

p0mvn, line 140: Another nit: I don't think that the method spec should be concerned about such implementation details. The API users only need to know what the method does / returns given the inputs.

The caller should not need to know such implementation details, especially when there are comments directly in code that say the same thing.

Please consider removing method spec implementation details while keeping the comments in code.

I'm not feeling strongly about this so feel free to resolve it if you disagree

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+func (k Keeper) IntermediaryAccountDelegationsExists(ctx sdk.Context, intermedairyAcc types.SuperfluidIntermediaryAccount) (delegations bool) {
+	store := ctx.KVStore(k.storeKey)
+
+	// we first check if the intermediary account does not have any connections
+	// store for intermediary account connection
+	intermediaryAccConnectionPrefixStore := prefix.NewStore(store, types.KeyPrefixLockIntermediaryAccAddr)
+	iterator := intermediaryAccConnectionPrefixStore.Iterator(nil, nil)
```
p0mvn, line 147:
```suggestion
	iterator := intermediaryAccConnectionPrefixStore.Iterator(nil, nil)
	defer iterator.Close()
```

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+
+	for ; iterator.Valid(); iterator.Next() {
+		if sdk.AccAddress(iterator.Value()).Equals(intermedairyAcc.GetAccAddress()) {
+			intermediaryConnectionExists = true
+		}
+	}
+
+	if intermediaryConnectionExists {
+		return true
+	}
```
p0mvn, line 159: It feels to me that we can immediately return if we see an intermediate account existence.

```suggestion
	for ; iterator.Valid(); iterator.Next() {
		if sdk.AccAddress(iterator.Value()).Equals(intermedairyAcc.GetAccAddress()) {
			return true
		}
	}
```

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+
+	// now check and verify that we don't have any delegations
+	_, found := k.sk.GetDelegation(ctx, intermedairyAcc.GetAccAddress(), sdk.ValAddress(intermedairyAcc.ValAddr))
```
p0mvn, line 162: nit: noticing multiple calls to `intermedairyAcc.GetAccAddress()`. Does it make sense to have a variable for it?

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+	if found {
+		return found
+	}
+
+	// check that the there's no synth denom with the synthdenom from the intermediary account
+	locks := k.lk.GetLocksLongerThanDurationDenom(ctx, intermedairyAcc.Denom, time.Second)
+	for _, lock := range locks {
+		// check if there are bonded synth locks with the synthdenom
+		synthLock, _ := k.lk.GetSyntheticLockup(ctx, lock.ID, stakingSyntheticDenom(intermedairyAcc.Denom, intermedairyAcc.ValAddr))
```
p0mvn, line 171: I think err check might be missed here

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+		if synthLock != nil {
+			return true
+		}
+
+		// check if there are unbonding synth locks with the synthdenom
+		synthLock, _ = k.lk.GetSyntheticLockup(ctx, lock.ID, unstakingSyntheticDenom(intermedairyAcc.Denom, intermedairyAcc.ValAddr))
```
p0mvn, line 177: ditto for error check

x/superfluid/keeper/stake_test.go
```diff
@@ -519,6 +519,113 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+		timePassedBeforeEndblock    time.Duration
+		expInterAccNumAfterDeletion int
+	}{
+		{
+			"sinlge intermediary account deletion upon single superfluid undelegation finish",
+			[]stakingtypes.BondStatus{stakingtypes.Bonded},
+			[]superfluidDelegation{{0, 0, 0, 1000000}},
+			[]bool{true},
+			unbondingDuration,
+			0,
```
p0mvn, line 539: 0 by default
```suggestion
```

x/superfluid/keeper/stake_test.go
```diff
@@ -519,6 +519,113 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+			for i, lock := range locks {
+				if tc.undelegationStart[i] {
+					err := suite.App.SuperfluidKeeper.SuperfluidUndelegate(suite.Ctx, lock.Owner, lock.ID)
+					suite.Require().NoError(err)
+				}
+			}
+
+			// add unbonding duration time to current block time
+			// we set block height to value greater than MinBlockHeightToBeginAutoWithdrawing in lockup module
+			suite.Ctx = suite.Ctx.WithBlockHeight(7)
```
p0mvn, line 612: nit: `suite.Ctx` -> `ctx := suite.Ctx` and then reuse `ctx`

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
-	prefixStore.Delete(address)
+	prefixStore.Delete(intermedairyAcc.GetAccAddress())
+}
+
+// IntermediaryAccountDelegationsExists returns true if the gien intermediary account has any delegations.
+// We check this by
+// - using staking keeper to check if actual delegations exist,
+// - checking if there are any intermediary account connection remaining to the intermediaryAcc
+// - if there's no synthetic lock with the intermediaryAccount.Denom + intermediaryAccount.ValAddr combination
+func (k Keeper) IntermediaryAccountDelegationsExists(ctx sdk.Context, intermedairyAcc types.SuperfluidIntermediaryAccount) (delegations bool) {
```
p0mvn, line 141: Do we have a test specifically for this method?

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
 	prefixStore.Set(acc.GetAccAddress(), bz)
 }
 
-func (k Keeper) DeleteIntermediaryAccount(ctx sdk.Context, address sdk.AccAddress) {
+func (k Keeper) DeleteAllEmptyIntermediaryAccounts(ctx sdk.Context) {
```
p0mvn, line 113: Do we have a direct test for this? I think we are only testing it vicariously via `EndBlocker`

x/superfluid/keeper/intermediary_account.go
p0mvn, line 141: Does it make sense having this method exported?

x/superfluid/keeper/intermediary_account.go
```diff
@@ -107,10 +110,76 @@ func (k Keeper) SetIntermediaryAccount(ctx sdk.Context, acc types.SuperfluidInte
...
+	itermedairyAccounts := k.GetAllIntermediaryAccounts(ctx)
+	for _, intermediaryAccount := range itermedairyAccounts {
+		k.DeleteIntermediaryAccountIfNoDelegation(ctx, intermediaryAccount)
+	}
+}
+
+// DeleteIntermediaryAccount deletes given intermediary account from store
+// Note that intermediary account is highly related to staking and delgation, and this
+// method should be used with caution.
+func (k Keeper) DeleteIntermediaryAccountIfNoDelegation(ctx sdk.Context, intermedairyAcc types.SuperfluidIntermediaryAccount) {
```
p0mvn, line 123: Does this method need to be exported?

p0mvn (commented):
Matt's explanation + implementation makes sense. Nice work on this!

I would like to request testing some methods directly. Other than that, most of my comments are nits.

x/superfluid/module.go
mattverse: I don't know the exact reason behind this, but afaik, this is one of the conventions in settings endblockers, I'm guessing it's because we want to separate out logic for abci? 
cc @alexanderbez

x/superfluid/keeper/intermediary_account.go
mattverse: changed this to private!

x/superfluid/keeper/intermediary_account.go
mattverse: agreed with you that method spec does not need these information, but still thought having organized list of checks would be a useful comment to have. Moved these comments inside methods!

x/superfluid/keeper/intermediary_account.go
mattverse: This method is not changing state in any way, just checks if the given intermediary account has a check or not, I think it'd be nice to have this public as this method can come in handy in different places. lmk what you think

x/superfluid/keeper/intermediary_account.go
mattverse: Ah good catch on this!

x/superfluid/keeper/stake_test.go
mattverse: deciding to leave this as having this visible is better for test readability imo

x/superfluid/keeper/intermediary_account.go
mattverse: This is intentional, `GetSyntheticLockup` returns an error when it doesnt have the given lock, but in our case, this is not an error

We can also add a comment for this if this happens to be confusing.

I agree with you that this is bad API usage, an idea I would like to move forward to  after getting this PR merged is adding a method `HasSyntheticLockup` in the lockup keeper that returns a boolean for usage like this(we have other instances that uses `GetSyntheticLockup` like this whilst ignoring error, just to check if we have a synthetic lock)

x/superfluid/keeper/intermediary_account.go
mattverse: ditto for the reply above

x/superfluid/keeper/intermediary_account.go
mattverse: imo, we don't need direct test as the only thing `EndBlocker` is doing is calling `DeleteAllEmptyIntermediaryAccounts`. Even if we do have a test case, the test cases would look exactly the same.

I think the current test of `TestDeleteAllEmptyIntermediaryAccounts` should not be seen as a vicorous way of testing, but fully testing the original method + testing endblocker execution

lmk what you think

app/modules.go
alexanderbez, line 169:
```suggestion
	// Superfluid should come after lockup module because matured synth locks
	// have to be deleted before superfluid EndBlock.
```

x/superfluid/keeper/epoch.go
```diff
@@ -63,10 +61,6 @@ func (k Keeper) MoveSuperfluidDelegationRewardToGauges(ctx sdk.Context) {
 		// we use cacheCtx and apply the changes later
 		_ = osmoutils.ApplyFuncIfNoError(ctx, func(cacheCtx sdk.Context) error {
 			_, err := k.dk.WithdrawDelegationRewards(cacheCtx, addr, valAddr)
-			if errors.Is(err, distributiontypes.ErrEmptyDelegationDistInfo) {
```
alexanderbez, line 66: just for context, why was this removed?

x/superfluid/module.go
alexanderbez: It would be ugly to have that defined here IMO. Having it encapsulated in a function makes it easier to test.

Actually, now that I think of it, this really should be defined on the keeper itself, just like `InitGenesis`. I.e. `(k Keeper) EndBlock(...)`

x/superfluid/keeper/epoch.go
mattverse: This line wasn't here in the first place, this was recently added to better log errors. However, with the main problem that caused all these errors fixed(by deleting intermediary account), we could revert it back to the code where we didn't have this line. 

Could still leave logging errors if necessary

x/superfluid/keeper/intermediary_account.go
mattverse: Added test cases!

alexanderbez (approved):

alexanderbez:
I definitely feel like we should have a changelog entry for this.

mattverse:
Current status of PR is that we're merging this upon checking logs for the current logs we have every epoch to confirm this PR would fix the main cause for the error logs

mattverse:
Confirmed that the logs are actually happening because of the undeleted intermediary accounts.
Planning to merge this after getting the changelog entry and fixing the merge conflicts.

I don't think we should pressure getting it on chain asap imo as long as we have merged it.

mattverse:
TODO For this PR pre-merge: write down edge casese that we think needs to be ensured testing, ensure that those cases are well unit tested

PR #2070, opened 2022-07-14
fix(e2e): upgrade proposals

p0mvn:
Related to: https://github.com/osmosis-labs/osmosis/issues/2039

Discovered that upgrade proposals were broken after trying to upgrade our SDK.

I think the reason the CI was passing before is that despite switching v10 to v11 containers during the upgrade, the state machine was compatible so it was able to continue.

However, upon upgrading SDK, that wasn't the case anymore so the CI started failing.

The problem was that each node on a chain was submitting and voting on its own proposal. As a result, none of them would eventually be accepted.

This change makes only one node to submit a proposal and all other nodes vote on the unified proposal.

The tests passed with this change + SDK upgraded. The actual upgrade is coming in a separate PR.
## Passthrough Change

- Formatted command logs consistently to improve the UX

## Testing and Verifying

### DIFF
--- a/tests/e2e/configurer/chain/commands.go
+++ b/tests/e2e/configurer/chain/commands.go
@@ -10,7 +10,7 @@ import (
 )
 
 func (n *NodeConfig) CreatePool(poolFile, from string) {
-	n.t.Logf("creating pool for chain-id: %s", n.chainId)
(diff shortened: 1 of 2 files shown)


czarcas7ic (approved):
LGTM!

Issue #1805, opened 2022-06-14
Statesync broken on v10.0.0


mturkia:
#### System information

Osmosis version: v10.0.0
OS & Version: Linux (most likely all OSes)
Commit hash: 7ca72c5fc4547f62b9984e62cd83639727691b62

#### Expected behaviour

Chain successfully synced to last statesync snapshot.

#### Actual behaviour

Statesync fails with following error:

```
12:34PM INF Applied snapshot chunk to ABCI app chunk=67 format=1 height=4737000 module=statesync total=68
12:34PM ERR State sync failed err="snapshot restoration failed: app version mismatch. Expected: 9, got: 0" module=statesync
```

After setting expected app version and starting the chain again, I get following error:
```
Error: fromProto: validatorSet proposer error: nil validator
```

#### Steps to reproduce the behaviour

- Configure statesync 
- Run **osmosisd start** to sync
- Once all statesync snapshot chunk are applied and you get the first error message about app version mismatch, terminate osmosisd
- Set the expected app-version with **tendermint set-app-version 9 --home wherever** See https://github.com/cosmos/cosmos-sdk/issues/10791 for reference

PolkachuIntern:
We have the exactly same issue while state-syncing.

p0mvn:
Thanks for the issue.

Before we try anything to resolve this, I think we need to merge this first: https://github.com/osmosis-labs/osmosis/pull/1698 and test again. That PR adds some logic for wiring wasm snapshotter correctly. We haven't had the time to include or test it for v10.

This is likely to be unrelated to the main issue but I would like to note that the app version did not get increased to v10:
```
12:34PM ERR State sync failed err="snapshot restoration failed: app version mismatch. Expected: 9, got: 0" module=statesync
```

The expected should be 10. I suspect that it did not get increased because we did a fork instead of a regular upgrade

CC: @czarcas7ic once we add logic for e2e testing forks, we should test the app version to be increased by 1

p0mvn:
Got state sync to work with #1698 and https://github.com/osmosis-labs/cosmos-sdk/pull/264 (WIP)

I discovered that state sync snapshots do not share the param store. The app version was stored there. As a result, the app version would not be propagated to the state-synching node, resulting in the reported error.

I made 2 changes. First, moved the app version to be stored in the multistore instead of the param store. Second, I had to make a breaking change where state-sync snapshots now include the app version and share it. As a result, the snapshot format is now different.

[...]

PR #2071, opened 2022-07-14
chore(app/upgrades): set app version correctly for v11

p0mvn:
Related to: https://github.com/osmosis-labs/osmosis/issues/1805

Although the app version was already set during the v9 upgrade, our v10 was a fork.
As a result, the upgrade handler was not executed to increment the app version.
This change helps to correctly set the app version for v11.

## Testing and Verifying

This change can be verified by e2e upgrade tests

### DIFF
--- a/app/upgrades/v11/upgrades.go
+++ b/app/upgrades/v11/upgrades.go
@@ -9,6 +9,10 @@ import (
 	"github.com/osmosis-labs/osmosis/v7/app/upgrades"
 )
 
+// We set the app version to pre-upgrade because it will be incremented by one
+// after the upgrade is applied by the handler.
+const preUpgradeAppVersion = 10
+
 func CreateUpgradeHandler(
 	mm *module.Manager,
 	configurator module.Configurator,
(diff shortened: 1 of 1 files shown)


app/upgrades/v11/upgrades.go
```diff
@@ -9,13 +9,24 @@ import (
...
 	bpm upgrades.BaseAppParamManager,
 	keepers *keepers.AppKeepers,
 ) upgradetypes.UpgradeHandler {
 	return func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+		// Although the app version was already set during the v9 upgrade, our v10 was a fork.
+		// As a result, the upgrade handler was not executed to increment the app version.
+		// This change helps to correctly set the app version for v11.
+		if err := keepers.UpgradeKeeper.SetAppVersion(ctx, preUpgradeAppVersion); err != nil {
+			return nil, err
+		}
```
ValarDragon, line 28: Do we have to do this manually every upgrade? Or is this just one-off code that we only have to do once?

p0mvn: > Do we have to do this manually every upgrade?

No, we need to do this now because the last release was a fork.

> Also why does the API take in pre upgrade version, not post?

This is because this logic is run before the upgrade handler logic that increases this version by 1

ValarDragon (approved):

mattverse (approved):
utACK

Issue #1916, opened 2022-06-30
x/mint: refactor, document and unit test keeper's `DistributeMintedCoin`


p0mvn:
## Background

This is a follow-up to #1839

[DistributeMintedCoin](https://github.com/osmosis-labs/osmosis/blob/97e1bc60a441a5fbeaff5ff94c6f241973b3da28/x/mint/keeper/keeper.go#L194) does too many things and violates the separation of concerns. It should be refactored to focus on one thing and potentially call other methods.

## Suggested Design

- [Distributing developer rewards](https://github.com/osmosis-labs/osmosis/blob/97e1bc60a441a5fbeaff5ff94c6f241973b3da28/x/mint/keeper/keeper.go#L212-L255) has its own method that is unit tested
- Distributing [staking incentives](https://github.com/osmosis-labs/osmosis/blob/97e1bc60a441a5fbeaff5ff94c6f241973b3da28/x/mint/keeper/keeper.go#L198-L203), [pool incentives](https://github.com/osmosis-labs/osmosis/blob/97e1bc60a441a5fbeaff5ff94c6f241973b3da28/x/mint/keeper/keeper.go#L205-L210) and [community pool incentives](https://github.com/osmosis-labs/osmosis/blob/97e1bc60a441a5fbeaff5ff94c6f241973b3da28/x/mint/keeper/keeper.go#L261-L266) can be refactored into a single reusable method and unit tested

## Acceptance Criteria

- the suggested refactor is complete to extract 2 functions:
 1. For developer rewards distribution
 2. Reusable function for distributing staking, pool incentives and community pool rewards.
- the newly extracted methods are fully covered by unit tests
- other tests continue to pass
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

p0mvn:
Marking this as blocked by #1915.

This is better be done after the existing tests are understood and refactored

PR #2026, opened 2022-07-12
refactor(x/mint): introduce distributeDeveloperRewards methods; unit tests

p0mvn:
Part of: https://github.com/osmosis-labs/osmosis/issues/1916

Continue refactoring `DistributeMintedCoin` to make it more testable. This change extracts a separate function for distributing developer rewards (`distributeDeveloperRewards`) and covers it with tests.

This change has no state-breaking changes and does not fix rounding issues from #1917 yet. However, it should help to isolate them to make the fix easier to do.

## Brief Changelog

- move developer rewards distribution to `distributeDeveloperRewards` and unit test it
   * remove some obsolete validation such as:
       * [account address being an empty string](https://github.com/osmosis-labs/osmosis/blob/3be2640a842394e962e323e9ad1885019e4e4229/x/mint/keeper/keeper.go#L242)
           * This is not possible due to param validation here:
https://github.com/osmosis-labs/osmosis/blob/c0052bbdaebac0ec94085e7742184ae2141706b9/x/mint/types/params.go#L226-L229
           * It has also proved to be untestable. Instead, a `CONTRACT`/invariant was made for this method:
https://github.com/osmosis-labs/osmosis/blob/c0052bbdaebac0ec94085e7742184ae2141706b9/x/mint/keeper/keeper.go#L265-L267 
   * add more validations such as:
       * developer vesting module account not having enough balance to distribute the dev rewards:
https://github.com/osmosis-labs/osmosis/blob/c0052bbdaebac0ec94085e7742184ae2141706b9/x/mint/keeper/keeper.go#L265-L267
    * the rest of the logic is the same / with comments added
- TestDistributeDeveloperRewards to cover any possible behavior and assert on all account state changes
- Create issues for the logic to be refactored further and document them in code:
  * #1916  (next step)
  * #1917
  * #2025

## Testing and Verifying

This change added tests and can be verified as follows:
- `TestDistributeDeveloperRewards`

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

Issue #1917, opened 2022-06-30
x/mint: fix rounding issues


p0mvn:
#### Background

This is a follow-up to https://github.com/osmosis-labs/osmosis/issues/1839

https://github.com/osmosis-labs/osmosis/pull/1874 helped to uncover rounding issues with minting. Please follow the PR description to understand the context.

There are currently 2 discovered rounding issues:

1. Truncating epoch provisions
- Line that causes this: https://github.com/osmosis-labs/osmosis/blob/ff383fc1f640bd1c2f902ce147659b23b751285c/x/mint/types/minter.go#L53

Changing this from Dec to Into might not be as simple. Can potentially lead to a larger refactor and a db migration.

2. Truncating while calculating developer rewards
- This one causes a much larger delta
Line with the problem: https://github.com/osmosis-labs/osmosis/blob/ff383fc1f640bd1c2f902ce147659b23b751285c/x/mint/keeper/keeper.go#L205
where GetProportions truncates the result before returning.

Also, not sure about the impact yet but the test shows that this is the source.

Before addressing this issue, we should first complete `DistributeMintedCoin` refactor in: https://github.com/osmosis-labs/osmosis/issues/1916

**Acceptance Criteria**
- The 2 sources of rounding issues are mitigated
- Unit tests added to ensure supply correctness
- The test introduced in #1874 is either refactored or removed
- Calculate the exact value that we are off by the next upgrade
- Allocate the missed value to the community pool during the next upgrade: https://github.com/osmosis-labs/osmosis/tree/main/app/upgrades/v11
- remove references to this issue in the codebase. It is referenced multiple times.

p0mvn:
Blocked by #1916

p0mvn:
Fixing this isn't as simple due to the dependencies on the cosmos-sdk APIs.

The core problem is that some sdk modules such as `x/bank` and `x/distribution` operate on `sdk.Coins` as opposed to `sdk.DecCoins`. For example:
- [`BurnCoins`](https://github.com/cosmos/cosmos-sdk/blob/3672c981b5d03e9ace80458023b31a97476d38a4/x/bank/keeper/keeper.go#L437)
- [`FundCommunityPool`](https://github.com/cosmos/cosmos-sdk/blob/3672c981b5d03e9ace80458023b31a97476d38a4/x/distribution/keeper/keeper.go#L172)

[...]

PR #2073, opened 2022-07-14
refactor(x/mint): introduce distributeDeveloperRewards methods; unit tests (2/2) 

p0mvn:
Closes: https://github.com/osmosis-labs/osmosis/issues/1916

Second part of the original PR: https://github.com/osmosis-labs/osmosis/pull/2026

This change _should not have_ state-breaking changes and does not fix rounding issues from #1917 yet.

## Brief Changelog
-  add new validation for:
       * developer vesting module account not having enough balance to distribute the dev rewards:
https://github.com/osmosis-labs/osmosis/blob/fadd99a21dfcf93b198cb40950cea7716cc47e78/x/mint/keeper/keeper.go#L274-L278
- Add more comments and link follow-up issues

## Testing and Verifying

- `TestDistributeDeveloperRewards`


x/mint/keeper/keeper.go
```diff
@@ -286,29 +303,21 @@ func (k Keeper) distributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.
...
-				if err != nil {
-					return sdk.Int{}, err
-				}
-				// If recipient is vesting account, pay to account according to its vesting condition
-				err = k.bankKeeper.SendCoinsFromModuleToAccount(
-					ctx, types.DeveloperVestingModuleAcctName, devRewardsAddr, devRewardPortionCoins)
-				if err != nil {
-					return sdk.Int{}, err
-				}
+			devRewardsAddr, err := sdk.AccAddressFromBech32(w.Address)
```
p0mvn, line 306: Removed `w.Address == ""` condition. The motivation is in the PR description. Diff is large due to tab changes

AlpinYukseloglu: Doesn't the block of code link in the PR description specifically allow `w.Address` to be an empty string? i.e. the block of code only runs if it's not empty, so if it is then the check that prevents empty strings in `sdk.AccAddressFromBech32` is bypassed?

Zoomed out block from what you linked in PR desc:
https://github.com/osmosis-labs/osmosis/blob/c0052bbdaebac0ec94085e7742184ae2141706b9/x/mint/types/params.go#L224-L230

AlpinYukseloglu (changes requested):
Overall clean/straightforward refactor! Left one comment to clarify something you mentioned in the PR desc. Other than that, LGTM

x/mint/keeper/keeper.go
p0mvn: Oh you are right! Good catch! I totally misread that check and assumed the opposite. I think it's best to keep the original approach and add more tests for that behavior

p0mvn: fixed

ValarDragon (approved):
LGTM! This is state machine compatible, except in the edge case where the developer rewards doesn't have enough coins. 

Old behavior was to panic and halt the chain, this will just return an error So I guess it is technically compatible in that the old code would just never proceed, and not commit to a different state machine result. Thus I think this is good to backport as well

AlpinYukseloglu (approved):
Thanks for the fix – LGTM!

p0mvn:
FYI, force pushed due to an issue with the merge. Had to reset and re-resolve

PR #2082, opened 2022-07-15
Simulation executor cleanup

ValarDragon:

This PR continues cleanup of the simulation executor back-end.

The first commit does two things:
- Delete legacy structs that are unused
- Breaks up the large RandomBeginBlock into simpler functions

Second commit:
- Makes operation scheduling work. (It was actually broken in the SDK / hasn't been working for years!), see subseuent comment

Third commit: 
- Adds one bug fix to time op scheduling (turns out the future ops in the SDK are riddled with errors that will make the backend panic)
- Makes better debugging for weighted operations, under the action name interface

Fourth commit: linter fix

Fifth commit: Just entirely delete time queueing operations. The old implementation wasn't too far off what it would need to work, but its going to be easier to just rewrite this with a clear correct consumer of the API. Furthermore, gov future ops are riddled with bugs that seem not worth fixing here. As legacy was broken, so we lose nothing by abandoning it and just rewriting it under an API we want, with clear consumer.

Sixth commit: Move cursed initialization logic that needs to explicitly be rewritten into a standalone function that can be gradually iterated on.

Seventh commit: Deletes unused SimParam argument from simulator.

Eight commit: Move more simulator complexity into simState.SimulateAllBlocks

Ninth commit: Move action result logging to dedicated function, delete ~effectively duplicated code from before.

## Testing and Verifying

Make test-sim works, and still produces valid messages for both legacy and new. Furthermore the logs look basically the same (so its functionality is preserved)

  - How is the feature or change documented? Removes the amount of things we need to document :^)


ValarDragon:
Lol the failing commit fixed queue operations to work. Turns out they were silently broken in the SDK at some point after we initially wrote them.

See here: https://github.com/cosmos/cosmos-sdk/blob/main/x/simulation/operation.go#L76

So the gov operation queue actually just had bugs that would cause it to never work. (Returns 1 nil future operation, that will always cause a panic)

I'm thinking we can take the fact that it never worked, to just drop support for that all together, and build it correctly.

ValarDragon:
I view this PR as now achieving the goal of "break up every mega function in the simulator, to more iteratively fixable chunks". (And deleting lots of deadcode in the process)

Issue #1911, opened 2022-06-30
Minor release testing workflow


ValarDragon:
## Background

We should make a workflow for testing minor releases against mainnet for state compatibility. The goal would be that on a branch thats in a PR to v10.x, there is some manual thing I could easily run to check a node against it. In case of app hash failure, being given the last {say} thousand lines of logs would be helpful

## Suggested Design

* We sync a new node against mainnet, using the relevant branch / commit
  * Where do we start the sync from?
    - [x] Snapshot?
    - [ ] State sync?
* Sync {X} blocks
  * Ideal if it syncs through a full day /  an epoch, this also informs where the start of our sync is from

## Acceptance Criteria

* Theres a github workflow or simple command we can execute to test minor releases, with the guarantee that that code syncs & correctly executes the last {day} of mainnet

ValarDragon:
Per discussion: We have a snapshot service, maybe we set it to some time before epoch each day, and then sync some number of blocks from there.

We have to figure out how to trigger it exactly, and report on results (is it a slackbot? Comment on a PR? )

nikever:
Implemented in https://github.com/osmosis-labs/osmosis/pull/2083 and https://github.com/osmosis-labs/osmosis/pull/2178

PR #2083, opened 2022-07-15
Check state compatibility on `v10.x` changes via CI

nikever:
Closes: #1911 

This pull request adds a GitHub Action to check state compatibility of new changes made to the `v10.x` branch.

_(as specified in the workflow comments)_
This workflow runs on every new commit to the latest development branch. 
It checks that the new commit didn't break state compatibility by:

- building the new `osmosisd` binary with the latest changes
- replaying a configurable number of previous blocks from chain history

Currently, the node starts from a snapshot taken some blocks before the last epoch 
and waits `DELTA_HALT_HEIGHT` blocks after epoch before halting.

Important Caveat:

The fact that this workflow succeeds and the binary doesn't fail doesn't 
directly imply that the new binary is state-compatible.
It could be that the binary is not state-compatible, but the condition 
which would break state compatibility was not present in the chunk of block history used.

On the other hand, if the workflow fails, the binary is not state-compatible. 

### Next steps

Now that we have self-hosted runners, we should migrate the most time-consuming and resource-intensive workflows there. This should reduce a lot of the CI time.

While we're at it, we might re-structure our CI. 
There are some jobs that are not used (e.g. build)

Also, this check will included in the _minor release workflow_ that will be used to automate the release process on new minor osmosisd versions. 

## Brief Changelog

  - Add workflow to check state compatibility of new changes. We run osmosis nodes via CI. yay! 

## Testing and Verifying

I have tested this workflow on different repositories with different conditions.
A complete run: https://github.com/osmosis-labs/osmosis/runs/7361263908?check_suite_focus=true

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no (don't think we should add DevOps stuff in the CHANGELOG.md)
  - How is the feature or change documented? added documentation in the workflow itself


nikever:
In order to correctly test the CI I have branched from `v10.x`

Should I re-branch from `main` and open a new PR against `main`?

nikever:
Do not merge It yet pls, I need to add a cleaning step

nikever:
Ok, done. Should be good to go :)

.github/workflows/check-state-compatibility.yaml
```diff
@@ -0,0 +1,108 @@
...
+  pull_request:
+    branches:
+      - 'v10.x'
+  workflow_dispatch:
+    branches:
+      - 'v10.x'
+
+env:
+  SNAPSHOT_URL: https://osmosis-snapshot.sfo3.cdn.digitaloceanspaces.com/osmosis.json
+  RPC_ENDPOINT: https://rpc.osmosis.zone
```
czarcas7ic, line 44: We should add fallback RPCs, that way when we get a invalid response we can simple query the next one down the line. If that isn't feasible, we should add some kind of validation to the response and continue querying for x minutes

czarcas7ic (approved):

.github/workflows/check-state-compatibility.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+# - building the new `osmosisd` binary with the latest changes
+# - replaying a configurable number of previous blocks from chain history
+
+# Currently, the node starts from a snapshot taken some blocks before the last epoch 
+# and waits `DELTA_HALT_HEIGHT` blocks after epoch before finally halting.
+
+# Important Caveat:
+
+# The fact that this workflow succeeds and the binary doesn't fail doesn't
```
p0mvn, line 12: Thank you for including this here. This is important :+1:

p0mvn (approved):

p0mvn:
@ValarDragon noticed that you self-requested review here. Is this goog to be merged?

nikever:
What do you think about running this workflow every day a `00.00` UTC against the latest commit of `v10.x`, so we can increase the likelihood of catching state incompatibility.

ValarDragon:
Running once a day sounds great to me, as long as we have some way to bubble the error (i.e. automatically created github issue)

Issue #1888, opened 2022-06-28
Update-Pool-Incentives Description field error


JohnnyWyles:
#### System information

Osmosis version: `10.0.1`
OS & Version: Linux 
Commit hash: e2217ddc22646f1be93a23132fdbf130275984a7

#### Expected behaviour
Update-Pool-Incentives proposal submission to change Community Pool gauge (0) to 25000000.

#### Actual behaviour
Proposal description is lost, flags are no longer available on v10.0.1 so json proposal used instead.  Json proposal has the deposit info read as the description.
![image](https://user-images.githubusercontent.com/97029546/176215483-92e29248-bf92-4baa-99b9-986f05481ae0.png)

#### Steps to reproduce the behaviour
`osmosisd tx gov submit-proposal update-pool-incentives 0 25000000 --proposal=ReduceEmissions.json --from=osmo19mywfjzj324w5ukf7ss6jak0dg9hnljfp0rfx4 --gas=1000000 --fees="2500uosmo" --node="https://testnet-rpc.osmosis.zone:443" --chain-id="osmo-test-4"`

ReduceEmissions.Json
```
{
   "title": "Continued reduction of emissions to Liquidity Pools",
[...]

PR #2086, opened 2022-07-15
modify pool-incentives submit proposal cli

catShaark:
Closes: #1888

## Brief Changelog

- add proposal-specific flags to NewCmdSubmitUpdatePoolIncentivesProposal and NewCmdSubmitReplacePoolIncentivesProposal
- disable poolincetives tx commands since these should be only used with a gov prefix
- register NewCmdSubmitReplacePoolIncentivesProposal handler on the gov module

## Testing and Verifying

  - Manually verified


x/pool-incentives/types/gov.go
```diff
@@ -39,7 +39,7 @@ func (p *ReplacePoolIncentivesProposal) GetDescription() string { return p.Descr
 func (p *ReplacePoolIncentivesProposal) ProposalRoute() string { return RouterKey }
 
 func (p *ReplacePoolIncentivesProposal) ProposalType() string {
-	return ProposalTypeUpdatePoolIncentives
+	return ProposalTypeReplacePoolIncentives
 }
```
ValarDragon, line 43: ooof, would this be state breaking to backport? (I don't have a sense for where ProposalType gets used)

catShaark: I think they're used in routing

alexanderbez: This is used in x/gov routing when executing the proposal's handler.

p0mvn (approved):
LGTM - I think we should create a changelog entry for this

x/pool-incentives/types/gov.go
mattverse: If it's only used for routing, I assume it's not going to be state-breaking to backport?

x/pool-incentives/client/cli/tx.go
```diff
@@ -9,30 +9,13 @@ import (
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	"github.com/cosmos/cosmos-sdk/x/gov/client/cli"
+	govcli "github.com/cosmos/cosmos-sdk/x/gov/client/cli"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 
 	"github.com/osmosis-labs/osmosis/v7/osmoutils"
 	"github.com/osmosis-labs/osmosis/v7/x/pool-incentives/types"
 )
 
-func NewTxCmd() *cobra.Command {
```
mattverse, line 19: hrmm is it safe to delete this / is there a reason behind deleting this?

catShaark: It is decided in #1888 that submit proposal cmd should only have gov prefix, so these command will be registered with gov handler instead of tx cmd of incentives module.

x/pool-incentives/types/gov.go
catShaark: I think If we have a ProposalTypeReplacePoolIncentives submited, this will be state-breaking

x/pool-incentives/client/cli/tx.go
catShaark: > LGTM - I think we should create a changelog entry for this

Should I add this to state breaking entry or not

mattverse: imo should go under breaking changes

CHANGELOG.md
```diff
@@ -70,13 +71,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Features
 
+* [#2086] Pool-incentives: remove tx command, add proposal-specific flags to proposal-submit commands
```
p0mvn, line 74: Let's try to avoid duplication
```suggestion
```

p0mvn (approved):

CHANGELOG.md
p0mvn: @catShaark please ping me once this is addressed - I think this is ready for merge

catShaark: @catShaark, I've committed your suggestion

ValarDragon:
Sorry, I'm actually generally confused about whats going on in these governance proposal routing.

My understanding is that in the today code, replace incentives proposal works on mainnet. Looking into this, looks like the handler is registered here? https://github.com/osmosis-labs/osmosis/blob/main/app/keepers/keepers.go#L366

Which then splits execution based on type https://github.com/osmosis-labs/osmosis/blob/fd23552f94caf1df9dbeb6fb4784e9d0efded2a4/x/pool-incentives/handler.go#L12-L24

So the routes being equivalent may be correct?

ValarDragon:
(this is also needlessly complex, and we should simplify it)

p0mvn:
> Sorry, I'm actually generally confused about whats going on in these governance proposal routing.
> 
> My understanding is that in the today code, replace incentives proposal works on mainnet. Looking into this, looks like the handler is registered here? https://github.com/osmosis-labs/osmosis/blob/main/app/keepers/keepers.go#L366
> 
> Which then splits execution based on type
> 
> https://github.com/osmosis-labs/osmosis/blob/fd23552f94caf1df9dbeb6fb4784e9d0efded2a4/x/pool-incentives/handler.go#L12-L24
> 
> So the routes being equivalent may be correct?

I think that the linked handler registration is focused on routing. On the other hand, this PR is primarily about fixing the client bootstrapping / cli.

Currently, pool-incentives gov cli is broken due to incorrectly setting up the cli logic and registering some flags twice. I've verified that [this](https://github.com/osmosis-labs/osmosis/issues/1888#issuecomment-1169454617) should help fix the original problem in #1888. 

We expose 2 ways to submit pool-incentives proposals:
- `osmosisd tx poolincentives update-pool-incentives` 
- `osmosisd tx gov submit-proposal update-pool-incentives`

Only the latter is correct because it properly gets registered on the gov module. The original bug is stemming from attempting to support both ways and, as a result, registering some CLI flags twice which breaks both commands.

The change related to the [proposal type](https://github.com/osmosis-labs/osmosis/blob/37e4a0ca1ac2284f28be34244d2a059635c7d23c/x/pool-incentives/types/gov.go#L42) seems to be a pass-through change. However, It does seem like a valid change that we should keep because we were returning an invalid proposal type.

The claim that `ProposalType()` is used in routing does seem to be correct. Although this is [client-side (rest) routing](https://github.com/osmosis-labs/cosmos-sdk/blob/f09814c9e23fd5832b11666a871bf51f36f85e41/x/gov/client/rest/tx.go#L39-L50).

Based on my investigation, the changes in this PR do not affect [the state-breaking routing we are concerned about](https://github.com/osmosis-labs/cosmos-sdk/blob/f09814c9e23fd5832b11666a871bf51f36f85e41/x/gov/keeper/proposal.go#L22).

I think we need the changes in this PR to fix the CLI and client (REST) routing. However, it might make sense to spend more time manually testing to ensure that everything is sound

p0mvn:
Hi @catShaark . I just would like to confirm that you've gotten a chance to manually test that both pool incentives proposals are functioning as expected.

Please let me know

p0mvn:
I think it would be great to have an e2e test for submitting these pool-incentive proposals. Would you be interested in adding it?

CHANGELOG.md
```diff
@@ -89,6 +90,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+* [#2086](https://github.com/osmosis-labs/osmosis/pull/2086) `ReplacePoolIncentivesProposal` ProposalType() returns correct value of `ProposalTypeReplacePoolIncentives` instead of `ProposalTypeUpdatePoolIncentives`
```
p0mvn, line 93: Do we need a second entry for this?

p0mvn (approved):
utACK. Changes lgtm.

I would like to know whether a manual test has been done or if there is intention to add an e2e test for this?

Either is fine with me to merge this

czarcas7ic (approved):

CHANGELOG.md
catShaark: I've manual tested this before on a local one node testnet, I'll try to submit proposals on osmosis testnet just for sure

p0mvn: Thank you. If this is manually tested on the local net, we can merge.

Issue #2100, opened 2022-07-17
Investigate e2e test execution time


ValarDragon:
## Background

E2E execution time takes very long. Often takes 8+ minutes, adding notable delay to CI (and thus developer flow). Roman's made a series of very helpful issues for mitigating build times. This issue is for investigating whats happening in our test execution time. CI run copy/paste for reference

```
--- PASS: TestIntegrationTestSuite (431.08s)
    --- PASS: TestIntegrationTestSuite/TestCreatePoolPostUpgrade (5.42s)
    --- PASS: TestIntegrationTestSuite/TestIBCTokenTransfer (94.79s)
    --- PASS: TestIntegrationTestSuite/TestSuperfluidVoting (77.14s)
```

## Ideas

Ideas I have for why this could be taking so long (in sequence of what I'd try / check):
* What are block times in e2e? For e2e test, we should be setting `precommit_timeout_wait` to be negligible. I'd even suggest `5ms`, we gain nothing by having delay in these tests. We don't need every validator signing every block either. 
  * Lets just adjust this config parameter and see what happens?
* Whats the blocking behavior for subsequent txs and subsequent queries?
  * Are we getting blocked in this? Can we add a "time trace'" to our ExecCLI & ExecQuery commands to learn more?
* Are relayers getting stuck in getting things onto both chains?

## Acceptance Criteria

We have some understanding of where time is getting stuck, via "guess and check", metrics / hacky log file, or otherwise

ValarDragon:
Some things that strike out to me with adding some basic time logging in https://github.com/osmosis-labs/osmosis/pull/2103 is that:
* We are running all tests / actions in sequence. When we have more tests post-upgrade, they should be able to run in parallel if they use different accounts
* IBC test is really slow (looks like 1.5 minutes, and is ran twice)
* Superfluid voting - we should be doing votes across nodes in parallel. Maybe we can make a helper method for doing the same action, on all node configs, in parallel?

ValarDragon:
update from #2103, most of the time was in require.Eventually tick timers. Changing those brought execution time to 3min 30 sec! Once we can get the test config propogated to lower block times, we'll probs be fine for now in terms of execution time

p0mvn:
Most of the items in this issue are tracked separately so I'm going to close this. Open an issue for parallelizing tests

PR #2102, opened 2022-07-17
chore: lower pre-commit timeout in E2E tests

alexanderbez:
Lower's the `precommit-timeout` to 5ms in the e2e tests.

ref: #2100

cc @ValarDragon @p0mvn

### DIFF
--- a/tests/e2e/initialization/node.go
+++ b/tests/e2e/initialization/node.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	sdkcrypto "github.com/cosmos/cosmos-sdk/crypto"
 	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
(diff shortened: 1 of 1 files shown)


alexanderbez:
Looks like this didn't help @ValarDragon, unless I changed the wrong config value?

czarcas7ic:
@alexanderbez there are also hard coded wait times in order to allow all Validators to vote on and gov prop. Your change should have still at least reduced the time to wait between txs though..

ValarDragon:
This was the wrong param, should've been precommit!

tests/e2e/initialization/node.go
```diff
@@ -310,6 +311,7 @@ func (n *internalNode) initValidatorConfigs(c *internalChain, persistentPeers []
 	valConfig.StateSync.Enable = false
 	valConfig.LogLevel = "info"
 	valConfig.P2P.PersistentPeers = strings.Join(persistentPeers, ",")
+	valConfig.Consensus.TimeoutCommit = 5 * time.Millisecond
```
p0mvn, line 314: Note that for this change to take effect we need to backport this to `v10.x` and rebuild the chain-init image. Then, it has to be updated here:
https://github.com/osmosis-labs/osmosis/blob/8e6e4de0bc65d521185f2aa82ff5884c107bd8c5/tests/e2e/containers/config.go#L30

If the `initialization` package is updated, the `previousVersionInit` tag should be updated as well

This is a manual step to make sure that we are able to initialize genesis and configs of the previous Osmosis version correctly.

ValarDragon:
Closing in favor of #2103 

(I switched the config to TM's test Config inspired by your change, and then found source of a lot of the delay via logging)

alexanderbez:
Man, this e2e stuff has really evolved from what I had originally designed 😆

Issue #1624, opened 2022-05-30
E2E test, abstract logs' container descriptor formatting


ValarDragon:
## Background

The e2e test logs all very helpfully indicate which container they are being ran from. However, the code to do so is quite repeated. 

They all look roughly of the form:
```golang
			s.T().Logf("reached upgrade height on %s container: %s", s.valResources[curChain.meta.Id][i].Container.Name[1:], s.valResources[curChain.meta.Id][i].Container.ID)
```

where `i` is validatorIndex, and `curChain` is a `chainConfig`. 

## Suggested Design

We should abstract this argument to be `s.getLogDescriptor(chainConfig, validatorIndex)` which returns a string with some common formatting and this same data.

## Acceptance Criteria

- [ ] Logging lines have the common docker complexity abstracted.

ValarDragon:
Or we could even just include the container directly, and then make a custom `contianer.String()` method, that gets the `%s` formatted as we like here.

ValarDragon:
After refactors state of logging is:
```
	c.t.Logf("submitting text proposal on %s container: %s", validatorResource.Container.Name[1:], validatorResource.Container.ID)
```

which is way better than before! I suggest we still make this:
```
	c.t.Logf("submitting text proposal on container: %s", validatorResource.Container)
```

And then we add a custom .String() method on the Container to format as we want, in a consistent manner

PR #2103, opened 2022-07-17
E2E: speedup execution time, add basic time tracking

ValarDragon:
Closes: #1624 , progress on #2100

To help with finding out why e2e execution is so slow. This PR achieves the following:
- add some basic time tracking to logs.
- Makes a unified method, and abstracts the log formatting for the container. This PR does it for txs, its up to a future PR to apply it everywhere else
- Switch the consensus config to be TM's test config (lower wait times intended for CI)
- Lower polling intervals. (This is also shows that we need more unification, theres so many different requires in this code. Really there should only be 2, One for CLI submission, one for query submission)

## Testing and Verifying

### DIFF
--- a/tests/e2e/initialization/node.go
+++ b/tests/e2e/initialization/node.go
@@ -310,6 +310,7 @@ func (n *internalNode) initValidatorConfigs(c *internalChain, persistentPeers []
 	valConfig.StateSync.Enable = false
 	valConfig.LogLevel = "info"
 	valConfig.P2P.PersistentPeers = strings.Join(persistentPeers, ",")
+	valConfig.Consensus = tmconfig.TestConsensusConfig()
 
 	tmconfig.WriteConfigFile(tmCfgPath, valConfig)
 	return nil
--- a/tests/e2e/configurer/base.go
+++ b/tests/e2e/configurer/base.go
@@ -177,7 +177,9 @@ func (bc *baseConfigurer) connectIBCChains(chainA *chain.Config, chainB *chain.C
 func (bc *baseConfigurer) initializeChainConfigFromInitChain(initializedChain *initialization.Chain, chainConfig *chain.Config) {
 	chainConfig.ChainMeta = initializedChain.ChainMeta
 	chainConfig.NodeConfigs = make([]*chain.NodeConfig, 0, len(initializedChain.Nodes))
+	setupTime := time.Now()
 	for _, validator := range initializedChain.Nodes {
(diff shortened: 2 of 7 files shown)


tests/e2e/initialization/node.go
```diff
@@ -310,6 +310,7 @@ func (n *internalNode) initValidatorConfigs(c *internalChain, persistentPeers []
 	valConfig.StateSync.Enable = false
 	valConfig.LogLevel = "info"
 	valConfig.P2P.PersistentPeers = strings.Join(persistentPeers, ",")
+	valConfig.Consensus = tmconfig.TestConsensusConfig()
```
p0mvn, line 313: Note that for this change to take effect we need to backport this to `v10.x` and rebuild the chain-init image. Then, it has to be updated here:
https://github.com/osmosis-labs/osmosis/blob/8e6e4de0bc65d521185f2aa82ff5884c107bd8c5/tests/e2e/containers/config.go#L30

If the `initialization` package is updated, the `previousVersionInit` tag should be updated as well

This is a manual step to make sure that we are able to initialize genesis and configs of the previous Osmosis version correctly.

ValarDragon: ahh that makes sense, I was confused why this was having no impact.

ValarDragon:
Lowering the polling interval from last commit shaved off 40 seconds

tests/e2e/initialization/node.go
ValarDragon: Can you lead applying this, post this PR merge?

ValarDragon:
ahh fixing query polling intervals halved the time. (IBC was using many queries with a tick of 10 seconds lmao. We literally spent all this time in a time.Sleep())

Execution time now at 3min30 sec, down 4 minutes from 7 min 30 seconds! Once we apply the TestConfig, we'll probably be good enough on execution time! (Not needing to pursue parallelism) Then just need to kill the build times, and were set

tests/e2e/initialization/node.go
p0mvn: I already pushed an image with the tag `v10.0.0-e2e-v3` to the Docker hub. It contains this change. Feel free to try it here

Also, I suggest not backporting this to `v10.x` via tag. We have been manually applying only `initialization` package changes so the rest of the e2e logic on `v10.x` is incompatible. As a result, backporting this via label might cause a large diff.

I opened a manual PR #2106 instead. I created the `v10.0.0-e2e-v3` image tag from that branch too

p0mvn (approved):
LGTM

PR #2108, opened 2022-07-18
chore(e2e): test tm config init image

p0mvn:
Closes: #XXX

Follow-up to https://github.com/osmosis-labs/osmosis/pull/2103 and https://github.com/osmosis-labs/osmosis/pull/2106

Updating the init image with the test tm config change

## Testing and Verifying


p0mvn:
@ValarDragon this new config makes IBC tests fail.

Observation: the chain is making progress but slow, I can see a lot of timeout error messages in consensus.

The error consistently happens during IBC tests. I think it times out [here](https://github.com/osmosis-labs/osmosis/blob/363192573d9f7a513b950da9b9699d51e92dc326/tests/e2e/containers/containers.go#L127)  after one minute because the chain is making progress slower due to timing out.

After the test runner failed, I kept the containers running so I noticed that IBC eventually connects. That makes me more confident that the chain is making progress but slower.

Please let me know how I should go about this. Should we revert the config change?

PR #2115, opened 2022-07-18
auto: update Go import paths to v10

ValarDragon:
**Automated pull request**

Updating Go import paths to v10


p0mvn (approved):

p0mvn:
Would like for someone else to review this too, please.

Just FYI, the reason why ValarDragon is the PR author is because his `PAT` token being used by the GitHub Action:
https://github.com/osmosis-labs/osmosis/blob/363192573d9f7a513b950da9b9699d51e92dc326/.github/workflows/import_paths.yml#L63

ValarDragon:
uhhh, lets not use my PAT for this. I'm going to see if I can revoke permissions for it to do this

ValarDragon:
It actually shouldn't have any of these permissions to even open a PR. I'm confused, going to delete my PAT, and remake via osmo bot

p0mvn:
There are a few other actions that depend on PAT:
- https://github.com/osmosis-labs/osmosis/blob/363192573d9f7a513b950da9b9699d51e92dc326/.github/workflows/add-issue-to-project.yml#L15
- https://github.com/osmosis-labs/osmosis/blob/363192573d9f7a513b950da9b9699d51e92dc326/.github/workflows/add-pr-to-project.yml#L16

ValarDragon:
yeah, they're just going to broken for now. But add_to_project_pat should have just had permissions for adding to the project (in fact the only thing github said it had write access to was projects ... . Explicitly did not give it any other write permission so I;m confused)

PR opening deserves a second access token.

mattverse (approved):
LGTM

What do folks think about creating an issue for creating a bot for automating this process when needed?

mattverse:
Oh wow, didn't know we had this!

p0mvn:
> Oh wow, didn't know we had this!

yeah, we just added it on Friday. We need to figure out this problem with the token so that Dev is not the PR author for these changes. Other than that, it seems to be functioning correctly

ValarDragon:
(new tokens made)

p0mvn:
I'm going to merge this since there are 2 approvals and CI passes. Would like to avoid this to keep having merge conflicts with other PRs

p0mvn:
> (new tokens made from osmobot)

Awesome thanks! I will test if it's working shortly

p0mvn:
> (new tokens made from osmobot)

@ValarDragon what is the new token name? The old token does not work. I'm assuming that the old one is still associated with your account but has reduced permissions?

ValarDragon:
`COMMIT_TO_BRANCH`
`ADD_TO_PROJECT_PAT`

PR #2125, opened 2022-07-18
e2e: enable state-sync tests (final MVP)

p0mvn:
Closes: #1879

## What is the purpose of the change

This is a final state sync e2e test MVP. It adds a `TestStateSync` that works independently from any other test.

`TestStateSync` works in the following way:

1.  Initializes configs for a new state-synching node. 
2. It queries trust height and trust hash from a running node.
3. It waits for the running node to produce snapshots at a height greater than trust height.
4. It starts the state-synching node and waits for it to catch up to the running node.
5. Finally, the state synching node is stopped and removed.

Verified that state-sync indeed works by exploring the logs:
```
9:21PM INF Discovered new snapshot format=1 hash="\\W�簮�~P��zm�02z���y[�]\x12�4\x0fM��n" height=25 module=statesync
9:21PM INF VerifyHeader hash=EE38ADA1C660EE1E06E5A7804D782E8C11CD42DCB3AB091CF22C8ED36BABB41C height=126 module=light
9:21PM INF VerifyHeader hash=0040847FD66AA6E8929793626B25CDA90B48BECAC0DE640453986A87CAB924F0 height=127 module=light
9:21PM INF Offering snapshot to ABCI app format=2 hash="\x14� \x05O����~\x05�\x00փM�\x05\x02�i�3����\\��u�" height=125 module=statesync
9:21PM INF Snapshot accepted, restoring format=2 hash="\x14� \x05O����~\x05�\x00փM�\x05\x02�i�3����\\��u�" height=125 module=statesync
9:21PM INF Fetching snapshot chunk chunk=0 format=2 height=125 module=statesync total=1
9:21PM INF VerifyHeader hash=7917722B8F4A448DE73BEEED98048E3BFFF8CEA8E6E67D14CB84C680A2A8567D height=125 module=light
9:21PM INF Header has already been verified hash=EE38ADA1C660EE1E06E5A7804D782E8C11CD42DCB3AB091CF22C8ED36BABB41C height=126 module=light
9:21PM INF Header has already been verified hash=0040847FD66AA6E8929793626B25CDA90B48BECAC0DE640453986A87CAB924F0 height=127 module=light
9:21PM INF Header has already been verified hash=EE38ADA1C660EE1E06E5A7804D782E8C11CD42DCB3AB091CF22C8ED36BABB41C height=126 module=light
[...]

PR #2129, opened 2022-07-18
e2e: backport #2125 - state-sync initialization

p0mvn:
Closes: #XXX

This is a backport of e2e initialization related to #2125.

### DIFF
--- a/tests/e2e/initialization/init_test.go
+++ b/tests/e2e/initialization/init_test.go
@@ -112,7 +112,7 @@ func TestSingleNodeInit(t *testing.T) {
 	existingChain, err := initialization.InitChain(id, dataDir, existingChainNodeConfigs, time.Second*3, forkHeight)
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


tests/e2e/initialization/node/main.go
```diff
@@ -55,11 +58,16 @@ func main() {
 		panic("rpc-servers is required, separated by commas")
 	}
 
+	persistenrPeers := strings.Split(persistentPeersStr, ",")
+	if len(persistenrPeers) == 0 {
+		panic("persistent peers are required, separated by commas")
+	}
```
ValarDragon, line 64: Where are these set into the node configs rn?

p0mvn: Also, it seems to me that there is some initialization logic that we don't need anymore. For example, the file we are discussing right now. This file is meant to be used for initializing a single node via Docker by running this as a Go program.

Contrary to the full chain, we never ended up needing initialziing single nodes via containers. Instead, this is done by directly [calling Go functions](https://github.com/osmosis-labs/osmosis/blob/90891d6cb9462e51655ed851ef0cb85d6d5503dd/tests/e2e/e2e_test.go#L152).

We can do this because state sync always starts after the upgrade tests.

czarcas7ic (approved):
Reviewed previous PR, this contains the same logic 👍

Issue #2128, opened 2022-07-18
Refactor: Change Pool API to have separate methods for `JoinPool` and `JoinPoolNoSwap`


ValarDragon:
## Background

Currently we have a single `JoinPool` and `CalcJoinPoolShares` method on the pool API. Something thats become quite clear in #2126 , is that this is terrible for client side users right now, and we need to split the AMM interface to have `JoinPool` and `JoinPoolNoSwap` OR make `JoinPool` just implement the no swap side.

I'm in preference of separating the go API, seeing as we already separate the proto API for messages.

At the moment, theres no clean way to estimate the `JoinPoolNoSwap` share amount out. IDK what the CLI does here, and I think theres no real way for other users to get the right numbers on multi-asset joins.

## Suggested Design

- [ ] Make a new `JoinPoolNoSwap` and `CalcJoinPoolNoSwapShares`  methods on pool interface
- [ ] Delete the code we had to introduce into simulator logic to emulate this https://github.com/osmosis-labs/osmosis/pull/2091/files#diff-77253003f1d97ce0424f8d9a21f789b506c6e01c20f89ef0220dfd6d61fc5e95R57-R96
- [ ] Make state machine call this for JoinPoolNoSwap
- [ ] Make JoinPool code have code re-use with this
- [ ] Adjust tests

## Acceptance Criteria

- [ ] We've improved the JoinPoolNoSwap API so that we can in golang easily estimate it, and delete that function in the simulator.

PR #2133, opened 2022-07-19
Refactor: Change Pool API to have separate methods for `JoinPool` and `JoinPoolNoSwap`

AlpinYukseloglu:
Closes: #2128

Add `JoinPoolNoSwap` and `CalcJoinPoolNoSwapShares` to our GAMM pool interface to have a clear and accessible flow for no-swap joins.

## Brief Changelog

- Add `JoinPoolNoSwap` and `CalcJoinPoolNoSwapShares` to GAMM pool interface
- Implement `JoinPoolNoSwap` and `CalcJoinPoolNoSwapShares` in balancer package
- Route `JoinPoolNoSwap` in pool_service.go to the new `JoinPoolNoSwap` in our pool interface
- Add tests for new balancer methods

TODO:
- Create issue for stableswap implementation

## Testing and Verifying

- `TestCalcJoinPoolNoSwapShares` in balancer's pool_test.go


x/gamm/pool-models/balancer/pool.go
```diff
@@ -714,20 +727,22 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("an error has occurred, more coins joined than tokens passed in")
+	} else if remainingTokensIn.Empty() {
 		return numShares, tokensJoined, nil
 	}
 
 	// 4) Still more coins to join, so we update the effective pool state here to account for
 	// join that just happened.
 	// * We add the joined coins to our "current pool liquidity" object (poolAssetsByDenom)
 	// * We increment a variable for our "newTotalShares" to add in the shares that've been added.
-	tokensJoined = tokensIn.Sub(remainingTokensIn)
```
AlpinYukseloglu, line 730: Context: `CalcJoinPoolNoSwapShares` includes this calculation and returns the resulting `tokensJoined` above

x/gamm/pool-models/balancer/pool.go
```diff
@@ -714,20 +727,22 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+	numShares, tokensJoined, remainingTokensIn, err := p.CalcJoinPoolNoSwapShares(ctx, tokensIn, swapFee)
 	if err != nil {
 		return sdk.ZeroInt(), sdk.NewCoins(), err
 	}
-	if remainingTokensIn.Empty() {
-		tokensJoined = tokensIn
+
+	// safely ends the calculation if all input tokens are successfully LP'd
+	if tokensJoined.IsAnyGT(tokensIn) {
+		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("an error has occurred, more coins joined than tokens passed in")
```
AlpinYukseloglu, line 737: Context: this check was included for single-asset joins at the end but for some reason not for all-asset joins (even though the method can terminate here). Added for consistency.

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -1218,3 +1218,51 @@ func (suite *BalancerTestSuite) TestIsActive(t *testing.T) {
...
+			)
+			require.NoError(t, err)
+
+			numShare, tokensJoined, remCoins, err := balancerPool.CalcJoinPoolNoSwapShares(ctx, test.tokensIn, balancerPool.GetSwapFee(ctx))
+
+			require.NoError(t, err)
+			require.Equal(t, test.expNumShare, numShare)
+			require.Equal(t, test.expTokensJoined, tokensJoined)
+			require.Equal(t, test.expRemCoin, remCoins)
+		})
```
mattverse, line 1266: consider adding checks for the state of the pool itself below(ex. pool liquidity, pool assets etc)

x/gamm/pool-models/balancer/pool.go
```diff
@@ -749,6 +764,52 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+			return sdk.ZeroInt(), sdk.NewCoins(), sdk.NewCoins(), sdkerrors.Wrapf(types.ErrDenomAlreadyInPool, invalidInputDenomsErrFormat, coin.Denom)
+		}
+	}
+
+	// ensure that there aren't too many or too few assets in `tokensIn`
+	if tokensIn.Len() != p.NumAssets() {
+		return sdk.ZeroInt(), sdk.NewCoins(), sdk.NewCoins(), errors.New("no-swap joins require LP'ing with all assets in pool")
+	}
+
+	// execute a no-swap join with as many tokens as possible given a perfect ratio:
```
mattverse, line 796: Some of these methods are exactly the same with `CalcJoinPoolShares`. consider generalizing these!

mattverse (commented):
Feature LGTM! 
Left some comments, please take a look!

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: Agreed there is some overlap even though we use `CalcJoinPoolNoSwapShares` in `CalcJoinPoolShares` – I don't want to complicate this PR too much so what if we make a separate small issue/PR after merging this to refactor/generalize these methods?

x/gamm/pool-models/balancer/pool.go
```diff
@@ -662,6 +662,19 @@ func (p *Pool) JoinPool(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (n
 	return numShares, nil
 }
 
+// JoinPoolNoSwap calculates the number of shares needed for an all-asset join given tokensIn with swapFee applied.
+// It updates the liquidity if the pool is joined successfully. If not, returns error.
+func (p *Pool) JoinPoolNoSwap(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, err error) {
```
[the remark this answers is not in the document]
AlpinYukseloglu: I believe both are already tested for at the keeper lever – do you think we should add tests in addition to those for balancer specifically? 

I'm in support of this but probably worth taking it to a separate small issue/PR on adding tests for balancer implementations of pool interface methods. Will make this issue if you agree.

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: Addressed in recent commit. I also added a failure case test and restructured the test logic to make it easier to parse visually and add to later

x/gamm/pool-models/balancer/pool.go
mattverse: Sgtm!

x/gamm/pool-models/balancer/pool.go
```diff
@@ -749,6 +764,52 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+		return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, err
+	}
+
+	// check to make sure the input denoms exist in the pool
+	for _, coin := range tokensIn {
+		_, ok := poolAssetsByDenom[coin.Denom]
+		if !ok {
+			return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, sdkerrors.Wrapf(types.ErrDenomAlreadyInPool, invalidInputDenomsErrFormat, coin.Denom)
+		}
+	}
```
ValarDragon, line 789: Can we reuse validation logic with CalcJoinPoolShares?

x/gamm/pool-models/balancer/pool.go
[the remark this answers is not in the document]
p0mvn: I think we should have a direct test for every method or function, especially, if they are exported. This is helpful in catching bugs early.

It would be great to have tests for methods/functions once they are introduced to avoid having them missed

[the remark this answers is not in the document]
AlpinYukseloglu: I tried implementing this and it ended up becoming a massive chunk of copy-pasted tests since both `JoinPool` and `JoinPoolNoSwap` are pretty much identical with `CalcJoinPool` and `CalcJoinPoolNoSwap`. Since this is a broader topic outside of the scope of this PR, I made a new issue and will add those changes to a new PR there.

Issue: https://github.com/osmosis-labs/osmosis/issues/2343

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -469,6 +469,53 @@ func TestGetPoolAssetsByDenom(t *testing.T) {
...
+			tokensIn:    sdk.NewCoins(sdk.NewCoin("uatom", sdk.OneInt())),
+			expectPass:  true,
+			expectedErr: nil,
+		},
+		"none of tokensIn is in pool asset map": {
+			poolAssets:  []balancer.PoolAsset{defaultOsmoPoolAsset, defaultAtomPoolAsset},
+			tokensIn:    sdk.NewCoins(sdk.NewCoin("foo", sdk.OneInt())),
+			expectPass:  false,
+			expectedErr: types.ErrDenomNotFoundInPool,
+		},
```
czarcas7ic, line 490: Non-blocking, but maybe a test case where one of two coins within `tokensIn` matches the poolAsset to ensure a partial match doesn't trigger a false positive.

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -1218,3 +1265,91 @@ func TestIsActive(t *testing.T) {
...
+		{Token: sdk.NewInt64Coin("foo", 100), Weight: sdk.NewIntFromUint64(5)},
+		{Token: sdk.NewInt64Coin("bar", 100), Weight: sdk.NewIntFromUint64(5)},
+	}
+
+	tests := map[string]struct {
+		tokensIn        sdk.Coins
+		expNumShare     sdk.Int
+		expTokensJoined sdk.Coins
+		expPoolAssets   sdk.Coins
+		expRemCoin      sdk.Coins
```
czarcas7ic, line 1280: maybe switch to `expRemainingCoins`, at first I thought this meant "expected removed coins" until I read the test. Just would help with immediate clarity :)

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -1218,3 +1265,91 @@ func TestIsActive(t *testing.T) {
...
+			expTokensJoined: sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(10)), sdk.NewCoin("bar", sdk.NewInt(10))),
+			expPoolAssets:   sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(100)), sdk.NewCoin("bar", sdk.NewInt(100))),
+			expRemCoin:      sdk.NewCoins(sdk.NewCoin("bar", sdk.NewIntFromUint64(1))),
+			expectPass:      true,
+		},
+		"two asset pool, no-swap join attempt with one asset": {
+			tokensIn:        sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(10))),
+			expNumShare:     sdk.NewIntFromUint64(0),
+			expTokensJoined: sdk.Coins{},
+			expPoolAssets:   sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(100)), sdk.NewCoin("bar", sdk.NewInt(100))),
```
czarcas7ic, line 1303: for my learning, why is the expected pool assets the same quantity on both failed and successful joins? Shouldn't the successful join increase poolAsset quantity?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -258,11 +258,21 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
 	return pCopy.joinPoolSharesInternal(ctx, tokensIn, swapFee)
 }
 
+// TODO: implement this
```
czarcas7ic, line 261: Do we have an issue for this?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -258,11 +258,21 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+func (p *Pool) CalcJoinPoolNoSwapShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, newLiquidity sdk.Coins, remainingCoins sdk.Coins, err error) {
+	return sdk.ZeroInt(), nil, nil, err
+}
+
 func (p *Pool) JoinPool(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, err error) {
 	numShares, _, err = p.joinPoolSharesInternal(ctx, tokensIn, swapFee)
 	return numShares, err
 }
 
+// TODO: implement this
```
czarcas7ic, line 271: Same

czarcas7ic (approved):
Nice work on this! I actually implemented this branch when I did some testing on the simulator and it worked as expected (this is why I had merged in main a couple weeks ago).

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: https://github.com/osmosis-labs/osmosis/issues/2376

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: https://github.com/osmosis-labs/osmosis/issues/2376

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: because `CalcJoinPoolNoSwapShares` should not mutate pool assets :)

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: ah yeah this is a good point but `RemCoins` is pulled from the convention we use in the function itself so we'd have to make the change there. might be worth making a small issue/pr for readability there

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: added!

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: changed anyway for test readability!

x/gamm/pool-models/balancer/pool.go
```diff
@@ -663,6 +663,19 @@ func (p *Pool) JoinPool(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (n
...
+	if err != nil {
+		return sdk.Int{}, err
+	}
+
+	// update pool with the calculated share and liquidity needed to join pool
+	p.IncreaseLiquidity(numShares, newLiquidity)
+	return numShares, nil
+}
+
 // CalcJoinPoolShares calculates the number of shares created to join pool with the provided amount of `tokenIn`.
```
ValarDragon, line 679: So now this method is basically not called in the codebase anywhere, right?

AlpinYukseloglu: `CalcJoinPoolShares`? Do you mean calls outside of the module? Otherwise it's called in `JoinPool`

ValarDragon: But JoinPool is no longer called in the state machien right?

AlpinYukseloglu: I might be missing your point here but I believe `JoinPool` is still called in `JoinSwapExactAmountIn`: https://github.com/osmosis-labs/osmosis/blob/16e3e4bdaa29006387a250121e9939e5d1d013df/x/gamm/keeper/pool_service.go#L272

x/gamm/pool-models/balancer/pool.go
```diff
@@ -751,6 +764,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+// and the remaining tokens in `tokensIn` after joining. If an all-asset join is not possible, returns an error.
+//
+// Since CalcJoinPoolNoSwapShares is non-mutative, the steps for updating pool shares / liquidity are
+// more complex / don't just alter the state.
+// We should simplify this logic further in the future using multi-join equations.
+func (p *Pool) CalcJoinPoolNoSwapShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, tokensJoined sdk.Coins, remainingTokens sdk.Coins, err error) {
+	// get all 'pool assets' (aka current pool liquidity + balancer weight)
+	poolAssetsByDenom, err := getPoolAssetsByDenom(p.GetAllPoolAssets())
+	if err != nil {
+		return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, err
```
p0mvn, line 780: I'm just going to point out which branches are not covered by tests.

@AlpinYukseloglu do you mind confirming please if it is possible to introduce test cases to cover these

This is the first branch

x/gamm/pool-models/balancer/pool.go
```diff
@@ -751,6 +764,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+	if tokensIn.Len() != p.NumAssets() {
+		return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, errors.New("no-swap joins require LP'ing with all assets in pool")
+	}
+
+	// execute a no-swap join with as many tokens as possible given a perfect ratio:
+	// * numShares is how many shares are perfectly matched.
+	// * remainingTokensIn is how many coins we have left to join that have not already been used.
+	numShares, remainingTokensIn, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
+	if err != nil {
+		return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, err
```
p0mvn, line 798: This is the second branch pending confirmation

x/gamm/pool-models/balancer/pool.go
```diff
@@ -751,6 +764,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+	}
+
+	// ensure that no more tokens have been joined than is possible with the given `tokensIn`
+	tokensJoined = tokensIn.Sub(remainingTokensIn)
+	if tokensJoined.IsAnyGT(tokensIn) {
+		return sdk.ZeroInt(), sdk.NewCoins(), tokensIn, errors.New("an error has occurred, more coins joined than token In")
```
p0mvn, line 804: This is the third branch pending confirmation

x/gamm/pool-models/balancer/pool.go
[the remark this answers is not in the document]
p0mvn: I'm okay with tackling this in a separate PR. However, do you mind tackling that before v12, please? I already added it to the milestone.

I know it looks good visually but I would really prefer having maximum direct test coverage in gamm for v12

x/gamm/pool-models/balancer/pool.go
```diff
@@ -689,12 +702,10 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
-	// check to make sure the input denoms exist in the pool
-	for _, coin := range tokensIn {
-		_, ok := poolAssetsByDenom[coin.Denom]
-		if !ok {
-			return sdk.ZeroInt(), sdk.NewCoins(), sdkerrors.Wrapf(types.ErrDenomAlreadyInPool, invalidInputDenomsErrFormat, coin.Denom)
-		}
+	// check to make sure the input denom exists in the pool
+	err = ensureDenomInPool(poolAssetsByDenom, tokensIn)
+	if err != nil {
+		return sdk.ZeroInt(), sdk.NewCoins(), err
```
p0mvn, line 708: Is there a test case covering this branch?

x/gamm/pool-models/balancer/pool.go
p0mvn: This check already occurs inside `CalcJoinPoolNoSwapShares`. I think we might not need it here again

x/gamm/pool-models/balancer/pool.go
```diff
@@ -715,12 +726,15 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
 	// * numShares is how many shares are perfectly matched.
 	// * remainingTokensIn is how many coins we have left to join, that have not already been used.
 	// if remaining coins is empty, logic is done (we joined all tokensIn)
-	numShares, remainingTokensIn, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
+	numShares, tokensJoined, remainingTokensIn, err := p.CalcJoinPoolNoSwapShares(ctx, tokensIn, swapFee)
```
p0mvn, line 729: I don't really understand the need for introducing a 3-value return when in any valid case we should be able to compute:
 `tokensJoined = tokensIn - remainingTokensIn`

I think that an additional return that is computed from other return values and/or arguments, only introduces unnecessary complexity.

Please let me know what you think @AlpinYukseloglu

x/gamm/pool-models/balancer/pool.go
p0mvn: However, if you follow my other suggestion of removing a 3-value return from `CalcJoinPoolNoSwapShares`, I think it makes sense to have it here instead

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: The original issue here was that `CalcJoinPoolShares` reimplemented a significant portion of the logic that already existed in `CalcJoinPoolNoSwapShares`.

This line happened to be a convenient place to put `CalcJoinPoolNoSwapShares`, but it's doing more than adding `tokensJoined` as a return value (e.g. it's also running the validation logic that was removed above).

My intent was to minimize rewritten code for general code simplicity and clarity. Does this change your position on this, or are you still in favor of removing the use of `CalcJoinPoolNoSwapShares` and reimplementing some of the logic here directly?

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: This should not be possible to hit, as it would require `remainingTokensIn` to have a negative number of tokens by definition since we're setting `tokensJoined = tokensIn.Sub(remainingTokensIn)` and then checking if `tokensJoined > tokensIn`, which can only happen if we subtracted by a negative number

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -469,6 +469,59 @@ func TestGetPoolAssetsByDenom(t *testing.T) {
...
+			k := suite.App.GAMMKeeper
+
+			poolId := suite.PrepareBalancerPoolWithPoolAsset(tc.poolAssets)
+			poolI, err := k.GetPoolAndPoke(suite.Ctx, poolId)
+			suite.Require().NoError(err, "test: %s", name)
+			balancerPool, isPool := poolI.(*balancer.Pool)
+			suite.Require().True(isPool, "test: %s", name)
+
+			poolAssetsByDenom, err := balancer.GetPoolAssetsByDenom(balancerPool.GetAllPoolAssets())
+			suite.Require().NoError(err, "test: %s", name)
```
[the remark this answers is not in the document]
AlpinYukseloglu: This is a fantastic point. Removed pool initialization logic and implemented as you described

x/gamm/pool-models/balancer/pool.go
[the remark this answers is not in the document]
AlpinYukseloglu: Agreed, will assign myself to it right now. Thanks for adding it to the V12 blocker list

x/gamm/pool-models/balancer/pool.go
p0mvn: Is it due to using a balancer pool constructor in `TestCalcJoinPoolNoSwapShares` where the constructor performs this validation?

What if we avoid it using it to trigger an invalid state and achieve full coverage?

We should be able to:
```
balancerPool := balancer.Pool{
  Address:            types.NewPoolAddress(defaultPoolId).String(),
  Id:                 defaultPoolId,
  PoolParams:         balancer.PoolParams{SwapFee: defaultSwapFee, ExitFee: defaultExitFee},
  PoolAssets:         test.poolAssets,
  FuturePoolGovernor: defaultFutureGovernor,
  TotalShares:        sdk.NewCoin(types.GetPoolShareDenom(defaultPoolId), types.InitPoolSharesSupply),
}
``` 

Now, we can force duplicate pool assets

x/gamm/pool-models/balancer/pool.go
p0mvn: Awesome, thank you!

x/gamm/pool-models/balancer/pool.go
p0mvn: On another look, it seems to me that we should be able to remove the `remainingTokens` in return value and have:

```
	numShares, tokensJoined, _, err = p.CalcJoinPoolNoSwapShares(ctx, tokensIn, swapFee)
	if err != nil {
	}

	// safely ends the calculation if all input tokens are successfully LP'd
	if tokensJoined.IsAnyGT(tokensIn) {
		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("an error has occurred, more coins joined than tokens passed in")
	} else if tokensIn.IsEqual(tokensJoined) {
	}
	remainingTokensIn := tokensIn.Sub(tokensJoined)
```

The [other caller](https://github.com/osmosis-labs/osmosis/blob/4797f8f3fc01594c3ce9fdd9f3b4b07128c74b04/x/gamm/pool-models/balancer/pool.go#L669) of `CalcJoinPoolNoSwapShares` does not currently use the third return. Since it can be computed from the second return `tokensJoined` and the argument `tokensIn`, I think we can simply keep the 2 return values.

Wdyt?

x/gamm/pool-models/balancer/pool.go
```diff
@@ -663,6 +663,19 @@ func (p *Pool) JoinPool(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (n
 	return numShares, nil
 }
 
+// JoinPoolNoSwap calculates the number of shares needed for an all-asset join given tokensIn with swapFee applied.
+// It updates the liquidity if the pool is joined successfully. If not, returns error.
+func (p *Pool) JoinPoolNoSwap(ctx sdk.Context, tokensIn sdk.Coins, swapFee sdk.Dec) (numShares sdk.Int, err error) {
+	numShares, newLiquidity, _, err := p.CalcJoinPoolNoSwapShares(ctx, tokensIn, swapFee)
```
p0mvn, line 669: Should we rename this similar to the name of the other caller?

```suggestion
	numShares, tokensJoined, _, err := p.CalcJoinPoolNoSwapShares(ctx, tokensIn, swapFee)
```

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: Please see the most recent case I added. It is still technically a trivial test case because duplicate pool assets are caught in multiple places throughout this function and ultimately end up in a division by zero even if you comment out all of them, but it should be covered now.

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: I think removing `remainingTokens` makes sense here – let me know if my recent commit addresses your point!

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -469,6 +469,51 @@ func TestGetPoolAssetsByDenom(t *testing.T) {
...
+			expectPass:  true,
+			expectedErr: nil,
+		},
+		"one of tokensIn is in pool asset map": {
+			poolAssets:  []balancer.PoolAsset{defaultOsmoPoolAsset, defaultAtomPoolAsset},
+			tokensIn:    sdk.NewCoins(sdk.NewCoin("uatom", sdk.OneInt()), sdk.NewCoin("foo", sdk.OneInt())),
+			expectPass:  false,
+			expectedErr: types.ErrDenomNotFoundInPool,
+		},
+		"none of tokensIn is in pool asset map": {
```
p0mvn, line 491: Please consider a case where `poolAssets` are empty. I know it shouldn't be possible considering the system in macro. However, it would still be great to test this function with any possible combination of inputs

x/gamm/pool-models/balancer/amm.go
```diff
@@ -228,3 +228,15 @@ func calcPoolSharesInGivenSingleAssetOut(
 	sharesInFeeIncluded := sharesIn.Quo(sdk.OneDec().Sub(exitFee))
 	return sharesInFeeIncluded
 }
+
+// check to make sure the input denoms exist in the provided pool asset map
```
p0mvn, line 232: nit:
```suggestion
// ensureDenomInPool check to make sure the input denoms exist in the provided pool asset map
```

x/gamm/pool-models/balancer/pool.go
```diff
@@ -749,6 +763,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
 	return numShares, tokensJoined, nil
 }
 
+// CalcJoinPoolNoSwapShares calculates the number of shares created to execute an all-asset pool join with the provided amount of `tokensIn`.
+// The input tokens must contain exactly the same tokens as the pool contains
```
p0mvn, line 767:
```suggestion
// The input tokens must contain the same tokens as in the pool.
```

x/gamm/pool-models/balancer/pool.go
```diff
@@ -749,6 +763,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+	}
+
+	err = ensureDenomInPool(poolAssetsByDenom, tokensIn)
+	if err != nil {
+		return sdk.ZeroInt(), sdk.NewCoins(), err
+	}
+
+	// ensure that there aren't too many or too few assets in `tokensIn`
+	if tokensIn.Len() != p.NumAssets() {
+		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("no-swap joins require LP'ing with all assets in pool")
```
p0mvn, line 789: Just trying to understand:

Why do you emphasize "no-swap joins"?

`CalcJoinPoolNoSwapShares` is called from `CalcJoinPoolShares` which is called from regular `JoinPool` that does swaps. I also noticed a similar check n `CalcJoinPoolShares`.

My point is that it seems that this is relevant for joins with the swaps as well.

x/gamm/pool-models/balancer/pool.go
```diff
@@ -749,6 +763,49 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
...
+	}
+
+	// execute a no-swap join with as many tokens as possible given a perfect ratio:
+	// * numShares is how many shares are perfectly matched.
+	// * remainingTokensIn is how many coins we have left to join that have not already been used.
+	numShares, remainingTokensIn, err := cfmm_common.MaximalExactRatioJoin(p, sdk.Context{}, tokensIn)
```
p0mvn, line 795: any reason we are not passing through the `ctx` from the parameter here?

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -1217,3 +1262,114 @@ func TestIsActive(t *testing.T) {
...
+				{Token: sdk.NewCoin("foo", sdk.NewInt(1)), Weight: sdk.NewIntFromUint64(1)},
+			},
+			expNumShare:     sdk.NewIntFromUint64(0),
+			expTokensJoined: sdk.Coins{},
+			expPoolAssets: []balancer.PoolAsset{
+				{Token: sdk.NewCoin("foo", sdk.NewInt(1)), Weight: sdk.NewIntFromUint64(1)},
+			},
+			expectPass: false,
+		},
+		"duplicate asset pool, no-swap join attempt with duplicate assets": {
```
p0mvn, line 1332: Does it make sense to have a test case for 3 assets?

p0mvn (approved):
LGTM. Great work on this @AlpinYukseloglu . Thank you for maximizing test coverage.

The remaining comments are non-blocking. However, could you please double-check the following:
- the error message about "no swap joins requiring all assets in pool". I just want to ensure that this is not exclusive of "joins with swaps" because `CalcJoinPoolNoSwapShares` is called from `CalcJoinPoolShares`.
- the new empty `sdk.Context` being passed in to `MaximalExactRatioJoin` from `CalcJoinPoolNoSwapShares`

x/gamm/pool-models/balancer/pool_test.go
```diff
@@ -469,6 +469,51 @@ func TestGetPoolAssetsByDenom(t *testing.T) {
 	}
 }
 
+func (suite *KeeperTestSuite) TestEnsureDenomInPool() {
```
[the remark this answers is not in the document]
AlpinYukseloglu: Oh yeah good catch, moving it

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: I don't think it makes a difference here but good catch, updated

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: This specific function is for no-swap joins, so the errors returned by it specify that it is a no-swap pool join (this specific line wouldn't error if this was a join with swaps)

Even though `CalcJoinPoolNoSwapShares` is still called in `CalcJoinPoolShares`, it is called in the "no-swap"/"all-asset"/"exact ratio" join component of it, and all pool join logic related to swaps is handled elsewhere in that function

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: Added!

x/gamm/pool-models/balancer/pool_test.go
AlpinYukseloglu: Added three-asset versions of the core test cases!

x/gamm/pool-models/balancer/pool.go
p0mvn: Awesome, thank you! Just wanted to make sure that this check was intentional to have on the path for regular joins as well

p0mvn (approved):

Issue #2079, opened 2022-07-15
Simulator weight management


ValarDragon:
## Background

Weights in the simulator were historically extraordinarily screwed up. The current simulator inherits this. We've taken steps with the Action work, to mitigate the module defining weights specifically.

I think that it is nice for a module to sometimes be able to specify a default weight, but this shouldn't be mandatory. The old UX for weights was just using a magical "int", which was incredibly unclear to parameterize, and then could be overriden at the top-level by a config. (Which makes sense, but the old code was architected very badly, and had initialization per module read the config)

What should happen instead is a module can define a weight, from an enum. The enum gets some semantic meaning to and from ints from the simulation back-end. (And we make a legacy wrapper for the legacy int) By enforcing this separation, we can also improve the meaning of the enums. (E.g. if theres 100 rare actions, and 1 normal action, maybe we should be bumping the normal action frequency to "Very frequent")

## Suggested Design

Suggested design to get modules to be able to define a weight:
- [x] Add an `WithWeight(w Weight) Action` method to the action interface
  - Should really be `func (a Action) WithWeight(w Weight) Action {a.weight = w; return a}` for our existing code
- [x] Change the type of Weight to `string` or smth thats not an int. 
- [x] Make a weight from int function, that we can use for legacy operations. (Can be many to one, e.g. weights 0-10 map to "Rare")
- [x] Move the action selection out of executor into types, and make it convert the weights into ints / replicate old functionality for now.
- [x] Add a demo API to some action. I'd suggest something like:
  - `simtypes.NewMsgBasedAction("CreateUniV2Pool", am.keeper, gamm.CreateUniV2Pool).WithWeight(simtypes.Rare)`
  - Should not be added to every message, they should default to `Common`!
- [x] Rename Weight to Freq

(Can likely be done across PRs)

PR #2156, opened 2022-07-19
simulator: frequency management

czarcas7ic:
Closes: #2079

This PR improves upon the previous `Weights` provided to specific actions. We instead want to move away from numerically defining how often actions should occur across modules and instead categorize them with human readable definitions (i.e. "Rare", "Common", "Frequent")

## Brief Changelog

- [x] Add an `WithFrequency(w Frequency) Action` method to the action interface
- [x] Change the type of Frequency to `string`
- [x] Make a frequency from int function, that we can use for legacy operations.
- [x] Move the action selection out of executor into types, and make it convert the frequencies into ints / replicate old functionality for now.
- [x] Add a demo API to some action:
  - `simtypes.NewMsgBasedAction("CreateUniV2Pool", am.keeper, gamm.CreateUniV2Pool).WithWeight(simtypes.Rare)`
- [x] Rename Weight to Freq

## Testing and Verifying


simulation/simtypes/weight.go
```diff
@@ -1,11 +1,40 @@
...
+		return 5
+	case strFrequency == Infrequent:
+		return 15
+	case strFrequency == Common:
+		return 35
+	case strFrequency == Frequent:
+		return 65
+	default:
+		return 35
+	}
```
p0mvn, line 39:
```suggestion
	switch strFrequency {
	case Rare:
		return 5
	case Infrequent:
		return 15
	case Common:
		return 35
	case Frequent:
		return 65
	default:
		return 35
	}
```

simulation/simtypes/action.go
```diff
@@ -105,3 +104,32 @@ func (m msgBasedAction) Execute(sim *SimCtx, ctx sdk.Context) (
...
+func totalFrequency(actions []Action) int {
+	totalFrequency := 0
+	for _, action := range actions {
+		totalFrequency += mapFrequencyFromString(action.Frequency())
+	}
+
+	return totalFrequency
+}
+
+type selectActionFn func(r *rand.Rand) Action
```
p0mvn, line 117: let's define types up top

simulation/simtypes/action.go
```diff
@@ -105,3 +104,32 @@ func (m msgBasedAction) Execute(sim *SimCtx, ctx sdk.Context) (
...
+
+func GetSelectActionFn(actions []Action) selectActionFn {
+	totalOpFrequency := totalFrequency(actions)
+
+	return func(r *rand.Rand) Action {
+		x := r.Intn(totalOpFrequency)
+		// TODO: Change to an accum list approach
```
p0mvn, line 124: What is that?

czarcas7ic: This was a TODO Dev wrote previously and I moved it over, so I am not 100 percent sure

ValarDragon (approved):

p0mvn (approved):

PR #2161, opened 2022-07-20
chore(x/mint): use Require() for assertions in tests

p0mvn:
Closes: #XXX

Making assertions on `Require()` fails test suite immediately. If it is not used, the execution of the test continues.
We should be using `Require()` in all cases.

If this is accepted, will create a follow-up issue for other modules.

## Testing and Verifying


ValarDragon (approved):

ValarDragon:
This LGTM, but I don't think we need to be set on one behavior, Often assert is more helpful than require, e,g, with a table driven test you can see all cases that failed

p0mvn:
> This LGTM, but I don't think we need to be set on one behavior, Often assert is more helpful than require, e,g, with a table driven test you can see all cases that failed

Agreed about it being useful in some cases. With table-driven tests this should not be a problem though. `t.Run(...)` should create a separate execution environment and try running all tests despite earlier cases failing

PR #2167, opened 2022-07-20
Pool incentives grpc test

hieuvubk:
## Refactor some tests in pool-incentives grpc_test.go to table drive
 
  - *TestGaugeIds*
  - *TestDistrInfo*
  - *TestIncentivizedPools*


x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -150,140 +172,123 @@ func (suite *KeeperTestSuite) TestLockableDurations() {
...
+				} else {
+					suite.Require().NoError(err)
+				}
+			}
+			res, err := queryClient.IncentivizedPools(context.Background(), &types.QueryIncentivizedPoolsRequest{})
+			if !tc.expectedErr {
+				suite.NoError(err)
+				suite.Equal(tc.expectedRecordLength, len(res.IncentivizedPools))
+
+				for i := 0; i < 0; i++ {
```
stackman27, line 282: Is this loop necessary?

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -150,140 +172,123 @@ func (suite *KeeperTestSuite) TestLockableDurations() {
...
+					})
+				}
+				tc.distRecords = append(distRecords, tc.distRecords...)
+
+				// Sort in ascending order of gaugeId
+				distRecords = tc.distRecords
+				sort.Slice(distRecords[:], func(i, j int) bool {
+					return distRecords[i].GaugeId < distRecords[j].GaugeId
+				})
+				// Create records
```
stackman27, line 269:
```suggestion
				// update records and ensure that non-perpetuals pot cannot get rewards. 
```

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -150,140 +172,123 @@ func (suite *KeeperTestSuite) TestLockableDurations() {
...
+					suite.NoError(err)
+
+					distRecords = append(distRecords, types.DistrRecord{
+						GaugeId: gaugeId,
+						Weight:  sdk.NewInt(100),
+					})
+				}
+				tc.distRecords = append(distRecords, tc.distRecords...)
+
+				// Sort in ascending order of gaugeId
```
stackman27, line 264: Why do we have put distrRecords in ascending order?

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -35,89 +30,116 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+			desc:                 "Happy case",
+			poolCreated:          true,
+			weights:              []sdk.Int{sdk.NewInt(100), sdk.NewInt(200), sdk.NewInt(300)},
+			expectedTotalWeight:  sdk.NewInt(600),
+			expectedRecordLength: 3,
+		},
+	} {
+		tc := tc
+		suite.Run(tc.desc, func() {
+			if tc.poolCreated {
```
stackman27, line 114: is `poolCreated` = false the same as `len(weights).= 0` 

We can have a valid pool without any tokens locked right?

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -35,89 +30,116 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+				suite.NoError(err)
+				suite.Equal(3, len(res.GaugeIdsWithDuration))
+
+				poolLpDenom := gammtypes.GetPoolShareDenom(pool.GetId())
+				for i := 0; i < len(res.GaugeIdsWithDuration); i++ {
+					suite.Equal(lockableDurations[i], res.GaugeIdsWithDuration[i].Duration)
+
+					gauge, err := suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, res.GaugeIdsWithDuration[i].GaugeId)
+					suite.NoError(err)
+					suite.Equal(0, len(gauge.Coins))
```
stackman27, line 74: nit: I think the verdict was to use `suite.Require()` whenever necessary [link](https://github.com/osmosis-labs/osmosis/pull/2046#pullrequestreview-1041071475)

alexanderbez: yup!!!

x/pool-incentives/keeper/grpc_query_test.go
hieuvubk: thanks you

x/pool-incentives/keeper/grpc_query_test.go
hieuvubk: I believe the UpdateDistrRecords function only accepts sorted records

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+		request *types.QueryGaugeIdsRequest
+		err     bool
+	}{
+		{
+			desc:    "Empty request",
+			request: &types.QueryGaugeIdsRequest{},
+			err:     true,
+		},
+		{
+			desc: "Unexisted pool",
```
czarcas7ic, line 44:
```suggestion
			desc: "Nonexistent pool",
```

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+		},
+	} {
+		tc := tc
+		suite.Run(tc.desc, func() {
+			res, err := queryClient.GaugeIds(context.Background(), tc.request)
+			if tc.err {
+				suite.Require().Error(err)
+			} else {
+				suite.Require().NoError(err)
+				suite.Require().NoError(err)
```
czarcas7ic, line 65:
```suggestion
```

was this accidental duplication?

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
-	suite.Equal(3, len(res.DistrInfo.Records))
+	for _, tc := range []struct {
+		desc                 string
+		poolCreated          bool
+		weights              []sdk.Int
+		expectedTotalWeight  sdk.Int
+		expectedRecordLength int
+	}{
+		{
+			desc:                 "No pool exist",
```
czarcas7ic, line 98:
```suggestion
			desc:                 "No pool exists",
```

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+			desc:                 "Happy case",
+			poolCreated:          true,
+			weights:              []sdk.Int{sdk.NewInt(100), sdk.NewInt(200), sdk.NewInt(300)},
+			expectedTotalWeight:  sdk.NewInt(600),
+			expectedRecordLength: 3,
+		},
+	} {
+		tc := tc
+		suite.Run(tc.desc, func() {
+			if tc.poolCreated {
```
czarcas7ic, line 114: It appears as though we aren't even testing the poolCreated: false test right? And it seems like we are just using poolCreated as an "is an err expected" variable. If this is the case, can we keep this to similar semantics like "err","valid", or "errExpected".

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -138,152 +160,136 @@ func (suite *KeeperTestSuite) TestLockableDurations() {
...
+				// update records and ensure that non-perpetuals pot cannot get rewards.
+				err = keeper.UpdateDistrRecords(suite.Ctx, tc.distRecords...)
+				if tc.expectedErr {
+					suite.Require().Error(err)
+				} else {
+					suite.Require().NoError(err)
+				}
+			}
+			res, err := queryClient.IncentivizedPools(context.Background(), &types.QueryIncentivizedPoolsRequest{})
+			if !tc.expectedErr {
```
czarcas7ic, line 279: It looks like we aren't requiring errors for expectedErr = true cases right?

czarcas7ic (changes requested):

x/pool-incentives/keeper/grpc_query_test.go
mattverse: yeah it errors if not

x/pool-incentives/keeper/grpc_query_test.go
mattverse: yeah we should add the case fpr Require Error

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
-	// Unexisted pool
-	_, err := queryClient.GaugeIds(context.Background(), &types.QueryGaugeIdsRequest{
-		PoolId: 1,
-	})
-	suite.Error(err)
-
 	// LockableDurations should be 1, 3, 7 hours from the default genesis state.
 	lockableDurations := suite.App.PoolIncentivesKeeper.GetLockableDurations(suite.Ctx)
-	suite.Equal(3, len(lockableDurations))
+	suite.Require().Equal(3, len(lockableDurations))
```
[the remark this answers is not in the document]
alexanderbez: Yeeah we should be using `Require()` so the tests exit upon failure.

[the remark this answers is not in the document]
mattverse: ahhh I see thx for the explanation!

x/pool-incentives/keeper/grpc_query_test.go
hieuvubk: I use that variable to test the case that no pool is created. 
Actually DistrInfo, IncentivizedPools query always returns nil error so we will check through response length and total weight. In no pool case, length and weights both be 0.

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -21,103 +22,124 @@ func (suite *KeeperTestSuite) TestGaugeIds() {
...
+			} else {
+				suite.Require().NoError(err)
+				suite.Require().NoError(err)
+				suite.Require().Equal(3, len(res.GaugeIdsWithDuration))
+
+				poolLpDenom := gammtypes.GetPoolShareDenom(pool.GetId())
+				for i := 0; i < len(res.GaugeIdsWithDuration); i++ {
+					suite.Require().Equal(lockableDurations[i], res.GaugeIdsWithDuration[i].Duration)
+
+					gauge, err := suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, res.GaugeIdsWithDuration[i].GaugeId)
```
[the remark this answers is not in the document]
hieuvubk: This code seems unnecessary because it was tested in gauges_test. I will remove the detailed check of each gauge above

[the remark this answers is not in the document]
hieuvubk: In the old test, there was also a detailed check of each gauge, should we remove them? Will make tests look clearer

hieuvubk:
I just refactored my tests, removed unnecessary loops.
About expErr variables in DistrInfo, IncentivizedPools test, functions always return nil error. So I will check through record length and total weight
Could u pls take a look @czarcas7ic @mattverse

x/pool-incentives/keeper/grpc_query_test.go
czarcas7ic: I see now, thanks!

x/pool-incentives/keeper/grpc_query_test.go
```diff
@@ -170,81 +156,53 @@ func (suite *KeeperTestSuite) TestLockableDurations() {
...
-	suite.Require().NoError(err)
 
 	for _, tc := range []struct {
 		desc                 string
 		poolCreated          bool
-		distRecords          []types.DistrRecord
-		expectedErr          bool
+		weights              []sdk.Int
+		perpetual		     bool
+		nonPerpetual		 bool
```
czarcas7ic: if perpetual is false that means nonPerpetual is automatically true right?

hieuvubk: Ahh just optional variables to test cases with additional Perpetual or NonPerpetual gauges

czarcas7ic (approved):

czarcas7ic:
Thanks for the ping @hieuvubk sorry about how long it took me to get back to this. Would like one more approval and then we can merge!

ValarDragon (approved):
Nice PR, thanks for making it!

Not sure why it got stuck, sorry for that delay

Issue #2117, opened 2022-07-18
Simulator: Property checks


ValarDragon:
## Background

This issue discusses implementation details for "Property Checks", a suggested abstraction over the legacy SDK invariants.

The goal of a property check is to be able to verify properties of the state machine, during these complex simulation runs. There are a couple dimensions the property checks should abstract over:
* "Memoryless" (AKA legacy invariants) - Just look at the current state, and run some consistency checks. If consistency check fails, error. E.g. "No account has a negative balance"
* Memoried - The checker can maintain state from its prior invokations. (So it can tell that a quantity that is only increasing, is in fact only increasing)

Triggers:
* We want to be able to trigger an event before and after certain message types are executed
  * In a first pass, we maybe just architect around before/after an action, and pass the message blob through. TBD
* Trigger an event after _any_  action
* Trigger an event at beginning and end block.

## Suggested Design

I suggest a "pub sub" model of  for the property checks. We have the simualtor "publish" events to an "event bus" struct. We have the simulator take in property checks, that each register "subscriptions". The simualtor handles putting those subscriptions onto the "event bus" struct as well.

To be honest, I think this is the correct answer for hooks should be designed everywhere in the SDK, and I suggest we use this as a demo to showcase a much cleaner way to rewrite hooks.

## Acceptance Criteria

- [ ] Property check written into new simulator
- [ ] Property check written for CFMM k only increasing between swaps & join pools. (And handling exit pools)
- [ ] Property check from legacy invariant
  - [ ] We will want to amend to give more fine-grained subscription control as well, but can be later.
- [ ] Actions trigger pre and post events

PR #2168, opened 2022-07-20
TWAP logic

ValarDragon:

Core logic for TWAP implementation. Right now wanted to ask if we wanted me to split up PR'ing the `types` package, `osmoutils`, and `twap/store` changes first. I think the work in `osmoutils` is generally quite helpful for our codebase, happy to break that out sooner before finishing the remainder.

Remaining steps needed:
- [x] Wire into app
- [x] Arithmetic for TWAP
- [x] Add before swap hook (EDIT: Actually needs to be deleted)
- [ ] Tests - (very load bearing todo haha)
  - [x] Actual API tests (Mostly done)
  - [x] CreatePool tests
  - [x] Swap triggering update
  - [x] EndBlock writing/reading tests
  - [ ] CreatePool + swap + end block
  - [ ] Overflow tests
  - [ ] Simulator integration for twap (blocked on #2117 )
  - [x] CreatePool -> Makes Twap Record
  - [x] KeyFormat & parsing consistency
  - [x] Unit tests for types package
  - [x] Unit test for update and interpolate record
  - [x] Unit test for arithmetic of twap calculation
- [x] Golang Docs
- [x] Issues for doing prune parameters in another PR
- [x] Issue for doing bindings in another PR
- [ ] Add-on features relegated for another PR (ArithmeticTwap from self-stored accumulator, simulator integration, mutation testing)

## Brief Changelog

- Implement TWAPs

## Testing and Verifying

Tests are added in this PR, will be breaking up components to be more easily reviewable. First broken out component is #2175 

  - How is the feature or change documented? x/gamm/twap/README


faddat:
Dev, this PR is aimed at giving users the ability to buy / sell using twap?

ValarDragon:
Nah, thats DCA'ing. This PR exposes on-chain TWAPs that people can then use to build that though!

Exposes an on-chain TWAP oracle for every pool

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+
+To illustrate with an example, given the sequence: `(0s, $1), (2s, $5), (5s, $1)`, the arithmetic mean TWAP is: $\frac{\$1 * (2s - 0s) + \$5 * (5s - 3s)}{5s - 0s} = \frac{\$10}{5} = \$2$.
+
+## Computation via accumulators method
+
+The prior example for how to compute the TWAP takes linear time in the number of time entries in a range, which is too inefficient. We require TWAP operations to have constant time complexity (in the number of records).
+
+This is achieved by using an accumulator. In the case of an arithmetic Twap, we can maintain an accumulator from `a_n`, representing the numerator of the Twap expression for the interval `t_0...t_n`, namely $a_n = \sum_{i=0}^{n-1} p_i (t_{i+1} - t_i)$. If we maintain such an accumulator for every pool, with `t_0 = pool_creation_time` to `t_n = current_block_time`, we can easily compute the TWAP for any interval. The twap for the time interval of price points `t_i` to `t_j` is then $twap = \frac{a_j - a_i}{t_j - t_i}$, which is constant time given the accumulator values.
```
p0mvn, line 17: For consistency:
```suggestion
This is achieved by using an accumulator. In the case of an arithmetic TWAP, we can maintain an accumulator from `a_n`, representing the numerator of the TWAP expression for the interval `t_0...t_n`, namely $a_n = \sum_{i=0}^{n-1} p_i (t_{i+1} - t_i)$. If we maintain such an accumulator for every pool, with `t_0 = pool_creation_time` to `t_n = current_block_time`, we can easily compute the TWAP for any interval. The TWAP for the time interval of price points `t_i` to `t_j` is then $twap = \frac{a_j - a_i}{t_j - t_i}$, which is constant time given the accumulator values.
```

p0mvn, line 17: > Notice that the latest price `p_n` isn't used, as it has lasted for a time interval of `0` seconds in this range

> If we maintain such an accumulator for every pool, with `t_0 = pool_creation_time` to `t_n = current_block_time`, we can easily compute the TWAP for any interval

Would accumulator at `t_n` be equal to `t_{n-1}` given the notice?

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+          - getInterpolatedRecord
+          - updateRecord
+          - computeArithmeticTwap
+        - Test overflow handling in all relevant arithmetic
+        - Complete testing code coverage (up to return err lines) for logic.go file
+    - API
+        - Unit tests for the public API, under foreseeable setup conditions
+- [ ] Integration into the Osmosis simulator
+    - The osmosis simulator, simulates building up complex state machine states, in random ways not seen before. We plan on, in a property check, maintaining expected TWAPs for short time ranges, and seeing that the keeper query will return the same value as what we get off of the raw price history for short history intervals.
+    - Not currently in scope for release blocking, but planned: Integration for gas tracking, to ensure gas of reads/writes does not grow with time.
```
p0mvn, line 99:
```suggestion
    - Not currently in scope for the release, but planned: Integration for gas tracking, to ensure gas of reads/writes does not grow with time.
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+- [ ] Mutation testing usage
+    - integration of the TWAP module into go mutation testing: https://github.com/osmosis-labs/go-mutesting
+        - The success we've seen with the tokenfactory module, is it succeeds at surfacing behavior for untested behavior.
```
p0mvn, line 102:
```suggestion
        - The success we've seen with the `tokenfactory` module. It succeeds at surfacing behavior for untested logic.
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+          e.g. if you delete a line, or change the direction of a conditional, does a test catch it.
```
p0mvn, line 103:
```suggestion
          e.g. if you delete a line, or change the direction of a conditional, mutation tests show if regular Go tests catch it.
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+    - We expect to get this to a state, where after mutation testing is ran, the only items it mutates, that is not caught in a test, is: Deleting `return err`, or `panic` lines, in the situation where that error return or panic isn't reachable.
```
p0mvn, line 104: - While this is awesome, do folks need to know such detail about mutation testing?
- Suggest paraphrasing to make it clearer. A lot of context for one sentence

p0mvn (commented):
README review. Overall, great write-up, pointed out some minor things or questions I had while reading

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
+A time weighted average price is a function that takes a sequence of `(time, price)` pairs, and returns a price representing an 'average' over the entire time period. The method of averaging can vary from the classic arithmetic mean, (such as geometric mean, harmonic mean), however we currently only implement arithmetic mean.
 
-## Basic architecture notes
+## Arithmetic mean TWAP
 
-We maintain the list of pools altered within in a block
\ No newline at end of file
+Using the arithmetic mean, the TWAP of a sequence `(t_i, p_i)`, from `t_0` to `t_n`, indexed by time in ascending order, is: $\frac{1}{t_n - t_0}\sum_{i=0}^{n-1} p_i (t_{i+1} - t_i)$. (Notice that the latest price `p_n` isn't used, as it has lasted for a time interval of `0` seconds in this range!)
+
+To illustrate with an example, given the sequence: `(0s, $1), (2s, $5), (5s, $1)`, the arithmetic mean TWAP is: $\frac{\$1 * (2s - 0s) + \$5 * (5s - 3s)}{5s - 0s} = \frac{\$10}{5} = \$2$.
```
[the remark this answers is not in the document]
stackman27: Also i think calculation is  lil off 
```
(1 * 2 + 5 * 2 ) / 5  = 12/5
```

Also if we follow the formula wouldn't it be 
```
(1 * 2) + 5 * (5 - 2) / 5 = 15/5 
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,102 @@
...
 
-NOTE: Not yet integrated into state machine, this package is a stub.
+A time weighted average price is a function that takes a sequence of `(time, price)` pairs, and returns a price representing an 'average' over the entire time period. The method of averaging can vary from the classic arithmetic mean, (such as geometric mean, harmonic mean), however we currently only implement arithmetic mean.
 
-## Basic architecture notes
+## Arithmetic mean TWAP
 
-We maintain the list of pools altered within in a block
\ No newline at end of file
+Using the arithmetic mean, the TWAP of a sequence `(t_i, p_i)`, from `t_0` to `t_n`, indexed by time in ascending order, is: $\frac{1}{t_n - t_0}\sum_{i=0}^{n-1} p_i (t_{i+1} - t_i)$. (Notice that the latest price `p_n` isn't used, as it has lasted for a time interval of `0` seconds in this range!)
```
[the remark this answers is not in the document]
ValarDragon: hrmm, yeah this is weird. This is definitely valid latex syntax :confused:

x/gamm/twap/README.md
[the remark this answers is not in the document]
ValarDragon: ah yeah woops, thanks. Fixed the numbers.

I'm generally confused why github latex renders differently than most latex renderers :/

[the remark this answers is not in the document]
stackman27: all good 👍

x/gamm/twap/README.md
ValarDragon: No, because the accumulator at `t_n` (`a_n`) contains the result of `p_{n-1}`, wheres the accumulator at `t_{n-1}`, (`a_{n-1}`) does not have the result of multiplying by `p_{n-1}`

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+- api.go - Public API, that other users / modules can/should depend on
+- hook_listener.go - Defines hooks & calls to logic.go, for triggering actions on 
+- keeper.go - generic SDK boilerplate (defining a wrapper for store keys + params)
+- logic.go - Implements all TWAP module 'logic'. (Arithmetic, defining what to get/set where, etc.)
+- module.go - SDK AppModule interface implementation.
+- store.go - Managing logic for getting and setting things to underlying stores
+
+## Store layout
+
+We maintain twap accumulation records for every AMM pool on Osmosis.
```
p0mvn, line 82:
```suggestion
We maintain TWAP accumulation records for every AMM pool on Osmosis. 
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+- store.go - Managing logic for getting and setting things to underlying stores
+
+## Store layout
+
+We maintain twap accumulation records for every AMM pool on Osmosis. 
+
+Because Osmosis supports multi-asset pools, a complicating factor is that we have to store a record for every asset pair in the pool.
+For every pool, at a given point in time, we make one twap record entry per unique pair of denoms in the pool. If a pool has `k` denoms, the number of unique pairs is `k * (k - 1) / 2`.
+
+Each twap record stores [(source)](https://github.com/osmosis-labs/osmosis/tree/main/proto/osmosis/gamm/twap):
```
p0mvn, line 87: page not found for the link

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+* last spot price of base asset A in terms of quote asset B
+* last spot price of base asset B in terms of quote asset A
+* Accumulation value of base asset A in terms of quote asset B
+* Accumulation value of base asset B in terms of quote asset A
+
+All TWAP records are indexed in state by the time of write.
+
+A new TWAP record is created in two situations:
+* When a pool is created
+* In the `EndBlock`, if the block contains any potentially price changing event for the pool. (Swap, LP, Exit)
```
p0mvn, line 97: Are (Swap, LP, Exit) exhaustive? Suggest specifying if this is an example or an exhaustive list

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+
+When a pool is created, records are created with the current spot price of the pool.
+
+During EndBlock, new records are created, with:
```
p0mvn, line 101:
```suggestion
During `EndBlock`, new records are created, with:
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+* The accumulator's updated based upon the most recent prior accumulator's stored last spot price
```
p0mvn, line 102:
```suggestion
* The accumulator's value is updated based upon the most recent prior accumulator's stored last spot price
```

x/gamm/twap/README.md
```diff
@@ -1,9 +1,137 @@
...
+* The LastSpotPrice's equal to the EndBlock spot price.
```
p0mvn, line 103: I'm guessing this is what was intended:
```suggestion
* The `LastSpotPrice` value is equal to the EndBlock spot price.
```

p0mvn (commented):
README "Store layout" review

x/gamm/twap/README.md
ValarDragon: Is now!

p0mvn (commented):
I finished reviewing the core logic but haven't reviewed the tests in detail yet.

I would like to spend more time tomorrow trying to run the tests.

LGTM overall, great job on this! Also, it is fine with me if this PR is not split further

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+func (k Keeper) afterCreatePool(ctx sdk.Context, poolId uint64) error {
+	denoms, err := k.ammkeeper.GetPoolDenoms(ctx, poolId)
+	denomPairs0, denomPairs1 := types.GetAllUniqueDenomPairs(denoms)
+	for i := 0; i < len(denomPairs0); i++ {
+		record, err := types.NewTwapRecord(k.ammkeeper, ctx, poolId, denomPairs0[i], denomPairs1[i])
+		// err should be impossible given GetAllUniqueDenomPairs guarantees
+		if err != nil {
+			return err
+		}
+		k.storeNewRecord(ctx, record)
```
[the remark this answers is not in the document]
ValarDragon: you need a record to exist for the case of "I create a pool and I swap against it" in the same block

Its also possible to create a pool in end block, good point that we should add a comment for this

x/gamm/twap/hook_listener.go
```diff
@@ -1,14 +1,22 @@
...
 import (
-	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	epochtypes "github.com/osmosis-labs/osmosis/v10/x/epochs/types"
 	"github.com/osmosis-labs/osmosis/v10/x/gamm/types"
 )
 
+func (k Keeper) MigrateExistingPools(ctx sdk.Context, poolIds []uint64) error {
```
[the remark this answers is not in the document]
ValarDragon: I wanted this file to be `listeners` for things that could trigger a state migration. Maybe I just change the file name to listeners.go?

I'm imagining we try to to change the entire Hook system into a publisher subscriber system

x/gamm/twap/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+			baseAssetDenom, quoteAssetDenom = quoteAssetDenom, baseAssetDenom
+		}
+		return getTwapInput{1, quoteAssetDenom, baseAssetDenom, startTime, endTime}
+	}
+
+	quoteAssetA := true
+	quoteAssetB := false
+	tPlusOne := baseTime.Add(time.Second)
+	// base record is a record with t=baseTime, sp0=10, sp1=.1, accumulators set to 0
+	baseRecord := newTwapRecordWithDefaults(baseTime, sdk.NewDec(10), sdk.ZeroDec(), sdk.ZeroDec())
```
p0mvn, line 92: I've seen `baseRecord` and `tPlusOne` being used in other tests. Can we reuse these in the test-global scope?

x/gamm/twap/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+	// record with t=baseTime+10, sp0=5, sp1=.2, accumulators updated from baseRecord
+	tPlus10sp5Record := newTwapRecordWithDefaults(
+		baseTime.Add(10*time.Second), sdk.NewDec(5), OneSec.MulInt64(10*10), OneSec)
+	// record with t=baseTime+20, sp0=2, sp1=.5, accumulators updated from tPlus10sp5Record
+	// tPlus20sp2Record := newTwapRecordWithDefaults(
```
p0mvn, line 97: Can deadcode be removed?

x/gamm/twap/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+			input:        makeSimpleTwapInput(baseTime.Add(-time.Hour), baseTime, quoteAssetA),
+			expErrorStr:  "too old",
+		},
+		"start time too old": {
+			recordsToSet: []types.TwapRecord{baseRecord},
+			ctxTime:      baseTime.Add(time.Second),
+			input:        makeSimpleTwapInput(baseTime.Add(-time.Hour), baseTime, quoteAssetA),
+			expErrorStr:  "too old",
+		},
+		// TODO: overflow tests, multi-asset pool handling, make more record interpolation cases
```
p0mvn, line 183: Agree that these would be useful. Do we have an issue for adding these?

x/gamm/twap/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	quoteAssetA := true
+	quoteAssetB := false
+	tPlusOne := baseTime.Add(time.Second)
+	// base record is a record with t=baseTime, sp0=10, sp1=.1, accumulators set to 0
+	baseRecord := newTwapRecordWithDefaults(baseTime, sdk.NewDec(10), sdk.ZeroDec(), sdk.ZeroDec())
+	// record with t=baseTime+10, sp0=5, sp1=.2, accumulators updated from baseRecord
+	tPlus10sp5Record := newTwapRecordWithDefaults(
+		baseTime.Add(10*time.Second), sdk.NewDec(5), OneSec.MulInt64(10*10), OneSec)
```
p0mvn, line 95: I know `OneSec` was probably chosen with the goal of having any value here but I think, in terms of abstraction, it might be better to avoid using only `OnseSec` because the accumulator also includes the price

x/gamm/twap/hook_listener.go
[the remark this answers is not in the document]
p0mvn: > Maybe I just change the file name to listeners.go?

I like `listeners.go` since hook and listener feel to be synonymous. Also, I think `listener` is frequently used when speaking of pub-sub systems.

> I wanted this file to be listeners for things that could trigger a state change

But this specific function is a one-time migration that is done during the upgrade. Why would we ever use pub-sub for this logic?

Also, I was referring to existing examples such as:
https://github.com/osmosis-labs/osmosis/blob/3b9e72c7e243eabb9c2bd5c91c1f3ced044cb3dd/x/lockup/keeper/migration.go#L40

where upgrade migrations logic exists in the keeper's `migrations.go` file

p0mvn (commented):
I'm part way through tests. Still have a couple of files to review

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+		return types.TwapRecord{}, err
+	}
+	record = interpolateRecord(record, ctx.BlockTime())
+	return record, nil
+}
+
+// interpolate record updates the record's accumulator values and time to the interpolate time.
+//
+// pre-condition: interpolateTime >= record.Time
+func interpolateRecord(record types.TwapRecord, interpolateTime time.Time) types.TwapRecord {
```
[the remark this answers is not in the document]
ValarDragon: interpolation logic should be defined here though imo, being internal

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (k Keeper) endBlock(ctx sdk.Context) {
+	changedPoolIds := k.getChangedPools(ctx)
+	for _, id := range changedPoolIds {
+		err := k.updateRecords(ctx, id)
+		if err != nil {
+			panic(err)
+		}
+	}
+	// 'altered pool ids' gets automatically cleared by being a transient store
```
alexanderbez, line 34: nit: I would just add this comment as part of the godoc.

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// mutates record argument, but not with all the changes.
+// Use the return value, and drop usage of the argument.
+func (k Keeper) updateRecord(ctx sdk.Context, record types.TwapRecord) types.TwapRecord {
+	newRecord := recordWithUpdatedAccumulators(record, ctx.BlockTime())
+
+	newRecord.Height = ctx.BlockHeight()
+
+	// TODO: Ensure order is correct
```
alexanderbez, line 58: is this still relevant?

x/gamm/twap/export_test.go
```diff
@@ -32,14 +32,14 @@ func (k Keeper) GetChangedPools(ctx sdk.Context) []uint64 {
...
-// 	return computeArithmeticTwap(startRecord, endRecord, quoteAsset)
-// }
+func ComputeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quoteAsset string) sdk.Dec {
+	return computeArithmeticTwap(startRecord, endRecord, quoteAsset)
+}
 
-// func InterpolateRecord(record types.TwapRecord, t time.Time) types.TwapRecord {
-// 	return interpolateRecord(record, t)
-// }
+func InterpolateRecord(record types.TwapRecord, t time.Time) types.TwapRecord {
```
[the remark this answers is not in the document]
ValarDragon: This is how export_test works? (All keeper tests are in `twap_test` package, due to testing integrations with the real osmosis app)

We want an internal function in the module, but it to get unit tested with app logic

x/gamm/twap/api.go
```diff
@@ -0,0 +1,89 @@
...
+// N.B. If there is a notable use case, the state machine could maintain more historical records, e.g. at one per hour.
+func (k Keeper) GetArithmeticTwap(
+	ctx sdk.Context,
+	poolId uint64,
+	baseAssetDenom string,
+	quoteAssetDenom string,
+	startTime time.Time,
+	endTime time.Time) (sdk.Dec, error) {
+	if startTime.After(endTime) {
+		return sdk.Dec{}, errors.New("called GetArithmeticTwap with a start time that is after the end time")
```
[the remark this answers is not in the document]
ValarDragon: Good call, just ran into something similar in debugging #2212

x/gamm/twap/hook_test.go
```diff
@@ -0,0 +1,81 @@
+package twap_test
+
+import (
+	"time"
+
+	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
+)
+
+// TestCreatePoolFlow tests that upon a pool being created,
```
p0mvn, line 9: Most tests in this file seem to be more functional than unit in nature. While it is useful to have both, I don't think we have any unit tests for hooks.

Do we have an issue for unit testing hooks?

x/gamm/twap/hook_test.go
```diff
@@ -0,0 +1,81 @@
...
+// we have made the correct store entries.
```
p0mvn, line 10: I think we should also be more specific about what exactly we are testing in this file. At first, I was looking in the `hook_listener.go` due to the name of this file. However, then, I released that a lot of the functionality is actually related to `endBlock` which is defined in `logic.go`

So I suggest renaming this to `hook_functional_test.go` and adding a comment that all these tests are functional and testing the relationship between twap with other system components.

x/gamm/twap/hook_listener.go
```diff
@@ -35,11 +42,11 @@ func (k Keeper) GammHooks() types.GammHooks {
...
 func (hook *gammhook) AfterPoolCreated(ctx sdk.Context, sender sdk.AccAddress, poolId uint64) {
-	// err := hook.k.afterCreatePool(ctx, poolId)
-	// // Will halt pool creation
-	// if err != nil {
-	// 	panic(err)
-	// }
+	err := hook.k.afterCreatePool(ctx, poolId)
+	// Will halt pool creation
+	if err != nil {
+		panic(err)
```
p0mvn, line 48: q: why are we okay with panicking here? Does this halt the chain?

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// interpolate record updates the record's accumulator values and time to the interpolate time.
+//
+// pre-condition: interpolateTime >= record.Time
+func interpolateRecord(record types.TwapRecord, interpolateTime time.Time) types.TwapRecord {
+	return recordWithUpdatedAccumulators(record, interpolateTime)
+}
+
+// TODO: Test math, test p0 vs p1 correctness
```
p0mvn, line 131: I see a large test for this function (`TestComputeArithmeticTwap`). Does it cover it?

If not, would be great to keep track of this in an issue

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+			test.expRecord.P0LastSpotPrice = test.record.P0LastSpotPrice
+			test.expRecord.P1LastSpotPrice = test.record.P1LastSpotPrice
+
+			gotRecord := twap.InterpolateRecord(test.record, test.interpolateTime)
+			require.Equal(t, test.expRecord, gotRecord)
+		})
+	}
+}
+
+func (s *TestSuite) TestUpdateTwap() {
```
p0mvn, line 86: Some tests in this file utilize `TestSuite` while others don't. Let's make the use of test suite consistent please

x/gamm/twap/logic.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		return endRecord.P1LastSpotPrice
+	}
+	var accumDiff sdk.Dec
+	if quoteAsset == startRecord.Asset0Denom {
+		accumDiff = endRecord.P0ArithmeticTwapAccumulator.Sub(startRecord.P0ArithmeticTwapAccumulator)
+	} else {
+		accumDiff = endRecord.P1ArithmeticTwapAccumulator.Sub(startRecord.P1ArithmeticTwapAccumulator)
+	}
+	return types.AccumDiffDivDuration(accumDiff, timeDelta)
```
p0mvn, line 148: I don't think we have a test for `AccumDiffDivDuration`. I think every exported function deserves a dedicated test. Please consider adding it

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+			denom1,
+			expectedTwap,
+		}
+	}
+	tenSecAccum := OneSec.MulInt64(10)
+	pointOneAccum := OneSec.QuoInt64(10)
+	plusOneSec := baseTime.Add(time.Second)
+	tests := map[string]testCase{
+		"basic: spot price = 1 for one second, 0 init accumulator": {
+			newOneSidedRecord(baseTime, sdk.ZeroDec(), true),
```
p0mvn, line 164: I think having named fields would improve readability here. Especially, since the test case definition is slightly above

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+			newOneSidedRecord(plusOneSec, OneSec, true),
```
p0mvn, line 165: Why are we providing `OneSec` for the accumulator?

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+		})
+	}
+}
+
+// TestComputeArithmeticTwap tests ComputeArithmeticTwap on various inputs.
+// The test vectors are structured by setting up different start and records,
+// based on time interval, and their accumulator values.
+// Then an expected TWAP is provided in each test case, to compare against computed.
+func TestComputeArithmeticTwap(t *testing.T) {
+	denom0, denom1 := "token/B", "token/A"
```
p0mvn, line 123: Can we make these global and reuse them across all twap tests, please? I see these being hardcoded in other tests

x/gamm/twap/hook_listener.go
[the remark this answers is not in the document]
ValarDragon: hrmm yeah your right the migraiton being here is weird.

I just realized that we can do this in InitGenesis, which is likely the correct entry point. WIll move it there.

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+			record:          newRecord(time.Unix(11, 0), sdk.NewDec(10), oneDec, twoDec),
+			interpolateTime: time.Unix(55, 0),
+			expRecord:       newExpRecord(oneDec.Add(OneSec.MulInt64(44*10)), twoDec.Add(OneSec.MulInt64(44).QuoInt64(10))),
+		},
+		"same time": {
+			record:          newRecord(time.Unix(1, 0), sdk.NewDec(10), oneDec, twoDec),
+			interpolateTime: time.Unix(1, 0),
+			expRecord:       newExpRecord(oneDec, twoDec),
+		},
+		// TODO: Overflow tests
```
p0mvn, line 70: Bump for TODO - is this being kept track of?

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+		"accumulator = 10*OneSec, t=5s. 10 base accum": testCaseFromDeltas(
+			sdk.NewDec(10), tenSecAccum, 5*time.Second, sdk.NewDec(2)),
+		"accumulator = 10*OneSec, t=3s. 10*second base accum": testCaseFromDeltas(
+			tenSecAccum, tenSecAccum, 3*time.Second, osmoutils.ThreePlusOneThird),
+		"accumulator = 10*OneSec, t=100s. .1*second base accum": testCaseFromDeltas(
+			pointOneAccum, tenSecAccum, 100*time.Second, sdk.NewDecWithPrec(1, 1)),
+
+		"accumulator = 10*OneSec, t=100s. 0 base accum (asset 1)": testCaseFromDeltasAsset1(sdk.ZeroDec(), OneSec.MulInt64(10), 100*time.Second, sdk.NewDecWithPrec(1, 1)),
+
+		// TODO: Overflow, rounding
```
p0mvn, line 198: Bump for TODO - is this being kept track of?

x/gamm/twap/logic_test.go
```diff
@@ -0,0 +1,206 @@
...
+			record:          newRecord(time.Unix(1, 0), sdk.NewDec(10), oneDec, twoDec),
+			interpolateTime: time.Unix(2, 0),
+			expRecord:       newExpRecord(oneDec.Add(OneSec.MulInt64(10)), twoDec.Add(OneSec.QuoInt64(10))),
+		},
+		"larger time interval": {
+			record:          newRecord(time.Unix(11, 0), sdk.NewDec(10), oneDec, twoDec),
+			interpolateTime: time.Unix(55, 0),
+			expRecord:       newExpRecord(oneDec.Add(OneSec.MulInt64(44*10)), twoDec.Add(OneSec.MulInt64(44).QuoInt64(10))),
+		},
+		"same time": {
```
p0mvn, line 65: What happens if `newTime` is before record time by mistake so the `timeDelta` is negative? How is this handled by the core logic? Is there a test covering this?

p0mvn (commented):
Tests LGTM. My first full pass on this PR is now complete.

Left some minor comments. I agree for the need for more tests outlined in TODOs. However, the test coverage in the core functions seems pretty good already, great work on that.

x/gamm/twap/hook_listener.go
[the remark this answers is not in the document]
ValarDragon: ah rip we can't do it in InitGenesis without adding to the gamm keeper interface, which seems worse.

I feel like keeping it in listeners.go, as what an external module can trigger logic for, is ok personally. I'd prefer not blocking on where this function lives, as it will ultimately be deleted later.

Perhaps we can call it InitGenesisAtMigration?

x/gamm/twap/logic_test.go
p0mvn: I understand why now. Just got confused initially because the accumulator is the sum of the product of time and price

x/gamm/twap/logic.go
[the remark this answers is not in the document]
ValarDragon: added comment

x/gamm/twap/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+			// setup time
+			s.Ctx = s.Ctx.WithBlockTime(tc.time)
+			tc.expRecord.Time = tc.time
+
+			// setup record
+			initSetRecord := tc.startRecord
+			if (tc.startRecord == types.TwapRecord{}) {
+				initSetRecord = initStartRecord
+			}
+			s.twapkeeper.StoreNewRecord(s.Ctx, initSetRecord)
```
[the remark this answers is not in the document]
ValarDragon: Good catch that was code smell. Fixed!

x/gamm/twap/api_test.go
ValarDragon: Done!

x/gamm/twap/api_test.go
ValarDragon: Wait no, this is `accumulator = (one second as number) * (10 seconds) * (10 spot price)`

x/gamm/twap/hook_test.go
ValarDragon: I don't think theres any unit tests that makes sense - they should all be functionality tests

app/upgrades/v12/upgrades.go
```diff
@@ -42,6 +42,14 @@ func CreateUpgradeHandler(
 			bpm.StoreConsensusParams(ctx, cp)
 		}
 
+		// Initialize TWAP state
+		// TODO: Get allPoolIds from gamm keeper, and write test for migration.
```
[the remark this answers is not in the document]
ValarDragon: https://github.com/osmosis-labs/osmosis/issues/2198

x/gamm/twap/hook_listener.go
ValarDragon: It wouldn't halt the chain, all panics are caught for txs, and should be caught in end block scopes.

x/gamm/twap/api_test.go
ValarDragon: We do for multi-asset pool handling. Not yet for overflow, but handling overflow is one of the big remaining twap items

x/gamm/twap/logic.go
ValarDragon: oh yeah lol, that is an awkward dangling comment haha

x/gamm/twap/hook_test.go
ValarDragon: I guess we could make these more table driven! Definitely very open to figuring out how to make better tests for what we want here!

x/gamm/twap/logic_test.go
ValarDragon: Hrmm, there isn't a test handling it because its incorrect behavior. But you should still get the same result. This is because TWAP is:
```
(accumEnd - accumStart) / (timeEnd - timeStart)
```

So if you reverse the arguments, you compute 
```
(accumStart - accumEnd) / (timeStart - timeEnd)
= -(accumEnd - accumStart) / - (timeEnd - timeStart)
= (accumEnd - accumStart) / (timeEnd - timeStart)
```
Thus the function is argument order independent.

This is for the same reason that you can don't care about the order of points when trying to compute the slope of a line.

x/gamm/twap/logic_test.go
ValarDragon: Why? Its fine for things that don't depend on app integration to just use testing.T ?

x/gamm/twap/logic.go
ValarDragon: ah yeah it does, beyond comments in that function for rounding behavior and overflow

TY for catching

x/gamm/twap/logic.go
ValarDragon: True, alright if I add this in an upcoming PR as we add overflow and alternate duration handling?

For now correctness is tested by the TWAP numbers

x/gamm/twap/logic.go
ValarDragon: Its not, thanks!

x/gamm/twap/logic.go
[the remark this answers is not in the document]
ValarDragon: fixed! Origin is I implemented it twice initially, and then realized they did the same thing lol

ValarDragon:
`TODO: overflow tests` are mentally kept track of, as items that need to be addressed in a subsequent overflow handling issue & PR.

x/gamm/twap/api_test.go
p0mvn: This is true for `accum0` but there is also `accum1` which only gets gets `OneSec` I think

x/gamm/twap/logic_test.go
p0mvn: Another purpose of the test suite is to group related spec together. I would assume that all the specs in the same file are logically grouped. 

In addition, just helps with consistency. I really like how we changed many of our test files post-audit to either use or not use the test suite

x/gamm/twap/api_test.go
ValarDragon: Ah yeah, I should probably add a comment here: that one is `accumulator = (one second as number) * (10 seconds) * (.1 spot price)`

These are explicitly chosen to be correct, but incredibly not obvious. Let me define these in their own line w/ a comment

x/gamm/twap/logic_test.go
p0mvn: That makes sense, thanks for explaining. I think it still makes to add a test even if the behavior is incorrect. We should aim to have test cases relative to all possible inputs even though some might be invalid.

x/gamm/twap/logic_test.go
ValarDragon: These are logically grouped though? Unit testing logic files.

I think its _weirder_ to break them away by what interacts with state here.

In the gamm logic, I contrarily think that makes sense. I feel like this depends on file structure. These files aren't that big, making more files imo hurts readability.

x/gamm/twap/logic.go
```diff
@@ -117,19 +121,14 @@ func (k Keeper) getMostRecentRecord(ctx sdk.Context, poolId uint64, assetA, asse
...
-func interpolateRecord(record types.TwapRecord, interpolateTime time.Time) types.TwapRecord {
-	return recordWithUpdatedAccumulators(record, interpolateTime)
-}
-
-// TODO: Test math, test p0 vs p1 correctness
-// precondition: endRecord.Time > startRecord.Time
+// precondition: endRecord.Time >= startRecord.Time
+// if (endRecord.Time == startRecord.Time) returns endRecord.LastSpotPrice
+// else returns
+// (endRecord.Accumulator - startRecord.Accumulator) / (endRecord.Time - startRecord.Time)
```
p0mvn, line 131: Suggesting for the sake of convention, at first I thought this was dead code

```suggestion
// computeArithmeticTwap computes and returns an arithmetic TWAP between
// two records given the quote asset.
// precondition: endRecord.Time >= startRecord.Time
// if (endRecord.Time == startRecord.Time) returns endRecord.LastSpotPrice
// else returns
// (endRecord.Accumulator - startRecord.Accumulator) / (endRecord.Time - startRecord.Time)
```

x/gamm/twap/hook_listener.go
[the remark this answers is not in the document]
p0mvn: We agreed to leaves as is offline - I guess we can discuss this in a separate issue later

x/gamm/twap/hook_test.go
p0mvn: I will make a separate issue for this to discuss further

p0mvn (approved):
LGTM. Great work!

One last thing I would like to bump is discussed here:
https://github.com/osmosis-labs/osmosis/pull/2168#discussion_r939039459

ValarDragon:
Added that test! Planning to merge upon test pass, to unblock other fronts

PR #2178, opened 2022-07-21
Improve CI for state-compatibility-check 

nikever:
Continues PR: [#2083 ](https://github.com/osmosis-labs/osmosis/pull/2083)

This PR improve the state-compatiblity check by:

- Forcing the workflow to fail in case of an `osmosisd` panic (Thanks @mattverse for the help spotting this)
- Add caching of `snapshots` to avoid re-downloading them every run if already available in the runner
- Schedule this workflow to run every day at `00.00` UTC
- Add slack notification in case of failure 

## Brief Changelog

- Improve `check-state-compability` CI (caching, scheduling, notification in case of failures)

## Testing and Verifying

Now workflow fails on errors: 
https://github.com/osmosis-labs/osmosis-ci/runs/7434236108?check_suite_focus=true

Regular run:
https://github.com/osmosis-labs/osmosis-ci/runs/7448557374?check_suite_focus=true

### DIFF
--- a/.github/workflows/check-state-compatibility.yaml
+++ b/.github/workflows/check-state-compatibility.yaml
@@ -92,6 +92,7 @@ jobs:
           SNAPSHOT_URL=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).url')
           SNAPSHOT=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).filename' | cut -f 1 -d '.')
 
+          # Download snapshot if not already present
           mkdir -p $HOME/snapshots/
           if [ ! -d "$HOME/snapshots/$SNAPSHOT" ]; then
               rm -rf $HOME/snapshots/*
--- a/.github/workflows/check-state-compatibility.yaml
+++ b/.github/workflows/check-state-compatibility.yaml
@@ -90,16 +90,17 @@ jobs:
         run:  |
           SNAPSHOT_INFO=$(curl -s --retry 5 --retry-delay 5 --connect-timeout 30 -H "Accept: application/json" ${{ env.SNAPSHOT_URL }})
           SNAPSHOT_URL=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).url')
-          SNAPSHOT_FILENAME=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).filename')
+          SNAPSHOT=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).filename' | cut -f 1 -d '.')
 
           mkdir -p $HOME/snapshots/
-          if [ ! -f "$HOME/snapshots/$SNAPSHOT_FILENAME" ]; then
-              wget -q -O $HOME/snapshots/$SNAPSHOT_FILENAME $SNAPSHOT_URL
+          if [ ! -d "$HOME/snapshots/$SNAPSHOT" ]; then
+              rm -rf $HOME/snapshots/*
+              mkdir $HOME/snapshots/$SNAPSHOT
+              wget -q -O - $SNAPSHOT_URL | lz4 -d | tar -C $HOME/snapshots/$SNAPSHOT -xvf -
           fi
-          lz4 -c -d $HOME/snapshots/$SNAPSHOT_FILENAME | tar -C $HOME/.osmosisd/ -xvf -
 
-          # Remove old snapshots
-          find $HOME/snapshots -type f -not -name '$SNAPSHOT_FILENAME' -delete
+          # Copy snapshot in Data folder
+          cp -R $HOME/snapshots/$SNAPSHOT/data $HOME/.osmosisd/
       -
         name: 🧪 Configure Osmosis Node
         run:  |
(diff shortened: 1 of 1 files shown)


nikever:
Caching the snapshots make the workflow pretty fast:
https://github.com/osmosis-labs/osmosis-ci/runs/7453664427?check_suite_focus=true

It's only the time needed to actually process the blocks (~ 8 mins) + build time (~ 2mins)

p0mvn:
Should this be merged to `main` instead of `v10.x`?

Otherwise, how is this work going to be preserved when we have v11?

I think we can have this merged to `main` and then simply bump up versions in triggers, once new Osmosis is released.

What do you think @nikever ?

nikever:
The previous PR https://github.com/osmosis-labs/osmosis/pull/2083 was merged in `v10.x`. I think, we can merge this one  to `v10.x` too.

 I had to branch from `v10.x` to correctly test the workflow in CI. 

I know it's not the standard approach but since this workflows needs to pinned exactly to correct branch I don't mind not having it on main. We can backport it easily to `main` though.

p0mvn:
> 

Makes sense. Let's eventually backport both #2083 and this PR to main, please. I'm just worried that all this great work gets lost because we don't have it on `main`

.github/workflows/check-state-compatibility.yaml
```diff
@@ -49,39 +52,62 @@ jobs:
...
+          wget -O $HOME/.osmosisd/config/genesis.json ${{ env.GENESIS_URL }}
+      -
+        name: ⏬ Download pre-epoch snapshot
+        run:  |
+          SNAPSHOT_INFO=$(curl -s --retry 5 --retry-delay 5 --connect-timeout 30 -H "Accept: application/json" ${{ env.SNAPSHOT_URL }})
+          SNAPSHOT_URL=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).url')
+          SNAPSHOT=$(echo $SNAPSHOT_INFO | dasel --plain -r json  '(file=osmosis-1-pre-epoch).filename' | cut -f 1 -d '.')
+
+          mkdir -p $HOME/snapshots/
+          if [ ! -d "$HOME/snapshots/$SNAPSHOT" ]; then
```
p0mvn, line 96: I'm assuming that this `mkdir` and checking if the directory exists logic is due to caching snapshots. Could you please add a comment that we do this because of caching?

.github/workflows/check-state-compatibility.yaml
```diff
@@ -95,14 +121,38 @@ jobs:
...
       -
-        name: Upload logs
-        if: always()
-        uses: actions/upload-artifact@v2
+        name: 🧹 Clean up Osmosis Home
+        run:  rm -rf $HOME/.osmosisd/
+      -
+        name: Send alert via Slack message
+        if: failure()
+        uses: slackapi/slack-github-action@v1.19.0
```
p0mvn, line 133: what channel does it send to?

p0mvn (approved):

.github/workflows/check-state-compatibility.yaml
nikever: #alerts

.github/workflows/check-state-compatibility.yaml
```diff
@@ -49,39 +52,62 @@ jobs:
 
   check_state_compatibility:
     # DO NOT CHANGE THIS: please read the note above
```
czarcas7ic, line 54: "please read the note above"

why not and what is the note above?

.github/workflows/check-state-compatibility.yaml
```diff
@@ -49,39 +52,62 @@ jobs:
...
-        name: Download pre-epoch snapshot
+      -
+        name: 🧪 Initialize Osmosis Node
         run:  |
-          rm -rf $HOME/.osmosisd/data
-          SNAPSHOT_URL=$(curl -s ${{ env.SNAPSHOT_URL }}  | dasel --plain -r json  '(file=osmosis-1-pre-epoch).url')
-          wget -q -O - $SNAPSHOT_URL | lz4 -d | tar -C $HOME/.osmosisd/ -xvf - 
-      - 
+          rm -rf $HOME/.osmosisd/ || true
+          osmosisd init osmo-runner -o
```
czarcas7ic, line 86: running osmosisd init already downloads the genesis, is there a reason we are downloading it again below?

czarcas7ic (approved):
Just two notes, good work 👍

.github/workflows/check-state-compatibility.yaml
nikever: Added

.github/workflows/check-state-compatibility.yaml
nikever: This is the note above that explains the reason why:
https://github.com/osmosis-labs/osmosis/blob/8a816f4c9c642a0fca493a79d19f7a72e0a03848/.github/workflows/check-state-compatibility.yaml#L21-L29

.github/workflows/check-state-compatibility.yaml
nikever: Does `osmosisd init` download genesis?

Running the command generates a different genesis from https://github.com/osmosis-labs/networks/raw/main/osmosis-1/genesis.json 

Would it work?

czarcas7ic: Hm does it? What if you specify --chain-id osmosis-1? I thought it was automated but you are probably right

czarcas7ic: This is non blocking anyway so you should be good to merge!

PR #2189, opened 2022-07-22
Change depandabot timer from daily to weekly

ValarDragon:

Unless folks particularly like dependabot being on daily, I'd like to suggest we keep it at weekly. I don't think theres much benefit to immediate updates (e.g. did we really need 3 golangci updates in 4 days?) - I typically go through the changelog and release notes / sometimes source for dep updates.

I'd only want instant PRs for security updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
 - package-ecosystem: gomod
   directory: "/"
   schedule:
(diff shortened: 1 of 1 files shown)


mattverse (approved):
Makes sense to me to change this to weekly basis LGTM

p0mvn (approved):

PR #2190, opened 2022-07-22
Enable deterministic Cosmwasm stargate queries

mattverse:
cref: https://github.com/osmosis-labs/wasmd/pull/7

The original design was to have basic initialization for the whitelisted quries in the wasmd side, but upon leaving the inititialized whitelist empty, this is the attempt to move the whitelisted quries and bindings layer into osmosis side. 

## Brief Changelog

- Add whitelisted stargate queries and ensure deterministic query.

## Testing and Verifying
Further plans to test this is TBD(upon concept ACK of this approach), although I do not have exact plans to test this in a local machine as of rn.


wasmbinding/stargate_bindings.go
```diff
@@ -0,0 +1,76 @@
...
+// StargateLayerBindings keeps whitelist and its deterministic
+// response binding for stargate queries.
+//
+// The query can be multi-thread, so we have to use
+// thread safe sync.Map instead map[string]bool.
+var StargateLayerBindings sync.Map
+
+func init() {
+	// auth
+	StargateLayerBindings.Store("/cosmos.auth.v1beta1.Query/Account", &auth.QueryAccountResponse{})
```
mattverse, line 24: I'm wondering if we can use these protos directly from `github.com/cosmos/cosmos-sdk` packages instead of defining them within our repo

wasmbinding/stargate_bindings.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/osmosis-labs/osmosis/v10/wasmbinding/stargate/mint"
+	"github.com/osmosis-labs/osmosis/v10/wasmbinding/stargate/slashing"
+	"github.com/osmosis-labs/osmosis/v10/wasmbinding/stargate/staking"
+)
+
+// StargateLayerBindings keeps whitelist and its deterministic
+// response binding for stargate queries.
+//
+// The query can be multi-thread, so we have to use
+// thread safe sync.Map instead map[string]bool.
```
p0mvn, line 19: It seems that the value in this map is not `bool` but the binding itself. If that's the case, can we avoid mentioning `map[string]bool` or change the value, please? I got stumbled by this for a sec

wasmbinding/stargate_bindings.go
```diff
@@ -0,0 +1,76 @@
...
+var StargateLayerBindings sync.Map
```
p0mvn, line 20: Suggest the name to contain "whitelist" in it.

wasmbinding/stargate_bindings.go
p0mvn: I think we should be using them directly from the sdk. 

In https://github.com/osmosis-labs/wasmd/pull/7/files, I noticed that there is a function called "NormalizeResponses".

IMO, there is no need for it if we are importing these from the SDK

wasmbinding/wasm.go
```diff
@@ -28,3 +31,13 @@ func RegisterCustomPlugins(
 		messengerDecoratorOpt,
 	}
 }
+
+func RegisterStargateQueries(queryRouter baseapp.GRPCQueryRouter) []wasmkeeper.Option {
```
p0mvn, line 35: How are you planning to propagate `StargateLayerBindings` to the logic in `wasmd`

I'm assuming the plan is to remove the `wasmd` copy of `StargateLayerBindings` and use this instead?

wasmbinding/stargate_bindings.go
mattverse: Ah my bad, fixed!

wasmbinding/wasm.go
mattverse: We are basically adding the newly added stargatelayerbindings rather than overriding it. This is possible because the stargate whitelist is initialized empty

https://github.com/osmosis-labs/osmosis/pull/2190/files#diff-1ac782d2fab1d48ad7f7ef06cf2d58b37c99b6b4c98879a5ad86a985b73fad58R321

alexanderbez (commented):
What's the conclusion we came to on today's call re stargate queries? I'd like to get clarification prior to reviewing this :)

mattverse:
@alexanderbez For twap, we're planning to use bindings. However, we also do want to work towards enabling whitelisted stargate queries, this PR should be good for a round of review!

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+	"github.com/cosmos/cosmos-sdk/baseapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	proto "github.com/gogo/protobuf/proto"
+	abci "github.com/tendermint/tendermint/abci/types"
 
 	"github.com/osmosis-labs/osmosis/v10/wasmbinding/bindings"
 )
 
+func StarGateQuerier(queryRouter baseapp.GRPCQueryRouter) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
```
alexanderbez, line 17: nit: also add a godoc :)

```suggestion
func StargateQuerier(queryRouter baseapp.GRPCQueryRouter) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
```

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+		route := queryRouter.Route(request.Path)
+		if route == nil {
+			return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("No route to query '%s'", request.Path)}
+		}
+
+		res, err := route(ctx, abci.RequestQuery{
+			Data: request.Data,
+			Path: request.Path,
+		})
+
```
alexanderbez, line 33:
```suggestion
```

wasmbinding/query_plugin.go
```diff
@@ -110,6 +144,31 @@ func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessag
 	}
 }
 
+func NormalizeReponses(binding interface{}, bz []byte) ([]byte, error) {
```
alexanderbez, line 147: nit: godoc

wasmbinding/query_plugin.go
```diff
@@ -110,6 +144,31 @@ func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessag
...
+	}
+
+	// build new deterministic response
+	bz, err = proto.Marshal(message)
+	if err != nil {
+		return nil, wasmvmtypes.Unknown{}
+	}
+
+	// clear proto message
+	message.Reset()
```
alexanderbez, line 167: why is this necessary?

alexanderbez (commented):
LGTM

wasmbinding/query_plugin.go
mattverse: @alexanderbez  tbh, I do not have full understanding why this is here left it here for now as I wanted to be careful of removing what I don't know it does. It was one of the chunks that I've forked from other repository that was using stargate query after enabling it. From my brief understanding, it seems that we are doing this because it was an Any value, and it would cached the old value. 

Looking at it again, I do feel that it might be safe to remove this line since we're not using the `message` varaible any more. Can you give opinions for this?

alexanderbez: mehh, honestly it's fine to leave it. I was just curious. I asked because `message` is created in the function and never leaves it.

alexanderbez (approved):

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+	proto "github.com/gogo/protobuf/proto"
+	abci "github.com/tendermint/tendermint/abci/types"
 
 	"github.com/osmosis-labs/osmosis/v10/wasmbinding/bindings"
 )
 
+// StargateQuerier dispatches whitelisted stargate queries
+func StargateQuerier(queryRouter baseapp.GRPCQueryRouter) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
+	return func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
+		binding, whitelisted := StargateWhitelist.Load(request.Path)
```
ValarDragon, line 20: Just want to be extra-sure, this path is the `/cosmos.auth.v1beta1.Query/Account` path, NOT the http path in the proto files right?

mattverse: yes, they are defined in the whitelist file

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+		res, err := route(ctx, abci.RequestQuery{
+			Data: request.Data,
+			Path: request.Path,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		// normalize response to ensure backward compatibility
+		bz, err := NormalizeReponse(binding, res.Value)
```
p0mvn, line 39:
```suggestion
		bz, err := NormalizeResponse(binding, res.Value)
```

wasmbinding/query_plugin.go
```diff
@@ -110,6 +144,33 @@ func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessag
 	}
 }
 
+// NormalizeReponses normalizes the responses by unmarshalling the response then marshalling them again.
+// Normalizing the response is specifically important for responses that contain type of Any.
+func NormalizeReponse(binding interface{}, bz []byte) ([]byte, error) {
```
p0mvn, line 149: Why is this normalization needed? Since we are not defining custom query protos in our repo and, instead, are directly using what is in the sdk and/or Osmosis, what are we ensuring backward compatibility against?

mattverse:
Found out that we need to Json Marshal the responses because cosmwasm stargate querier json unmarshals the responses by default. Closing this PR and opening a new PR replacing this one

Issue #2198, opened 2022-07-22
TWAP: Integrate & test migration logic


ValarDragon:
## Background

Add migration logic for creating initial TWAP records for every existing pool, at upgrade time. The logic for 'migrating' an existing pool is the same for creating a new pool.

The TWAP keeper exposes a function called `MigrateExistingPools` which just runs the afterCreatePool logic on each input pool. (This should be correct)

We need to integrate passing in all pools from gamm, inside of the upgrades logic. We furthermore need to add an upgrade test for v11, that ensures were still producing blocks with swaps post-upgrade.

## Suggested Design & Acceptance Criteria

- Add a method to get all pool IDs from gamm, and integrate in `app/upgrades/v11`
- Copy upgrade test format from v9
- Ensure that post-upgrade, we can execute blocks with sends in them.

PR #2137, opened 2022-07-19
Speedup mint keeper tests (20x locally)

ValarDragon:
## What is the purpose of the change

The mint keeper tests had become egregiously slow - 2 minutes in CI, 45 seconds locally.

It now takes 2 seconds locally

This PR does some pretty basic speedups for it:
- Remove every test being ran 4 times (4x speedup)
- Make the default setup parameters have the `reduction period` be 10 epochs, rather than the default (365), lowering amount of iteration. (Estimated rest of speedup)
- Delete unneeded LP incentive setup. (This was doing absolutely nothing, I don't know why it was here)

From this, it also strikes out to me that we have a ways to go in improving these tests further. (A lot of these legacy tests are hard to follow / egregiously slow / not using our suite.app tooling)

## Brief Changelog
 
  - 20x speedup mint keeper tests

## Testing and Verifying

This change is already covered by existing tests, such as *(please describe tests)*.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #2199, opened 2022-07-22
TWAP store.go & store_test.go files

ValarDragon:

Store management from #2168 . Every method in here should be covered by tests, with exception of deletion, which is not yet integrated. #2198 

## Brief Changelog

Adds files for TWAP store management

## Testing and Verifying

This change added tests for every function. I believe that all 2 asset pool functionality is covered, but please verify. We may need to add more tests for 3 asset pools for `getRecordAtOrBeforeTime`, I suggest we do that in multi-asset pool integration tests after more logic is through.

  - How is the feature or change documented? TWAP spec, and code comments

### DIFF
--- a/x/gamm/twap/store.go
+++ b/x/gamm/twap/store.go
@@ -76,7 +76,7 @@ func (k Keeper) getMostRecentRecordStoreRepresentation(ctx sdk.Context, poolId u
 	return types.ParseTwapFromBz(bz)
 }
 
-// getMostRecentRecordStoreRepresentation returns all most recent twap records
+// getAllMostRecentRecordsForPool returns all most recent twap records
 // (in state representation) for the provided pool id.
 func (k Keeper) getAllMostRecentRecordsForPool(ctx sdk.Context, poolId uint64) ([]types.TwapRecord, error) {
 	store := ctx.KVStore(k.storeKey)
--- a/x/gamm/twap/store.go
+++ b/x/gamm/twap/store.go
@@ -1,3 +1,4 @@
+//nolint:unused,deadcode
 package twap
 
 import (
--- a/x/gamm/twap/store.go
+++ b/x/gamm/twap/store.go
@@ -11,6 +11,9 @@ import (
 	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
 )
 
+// trackChangedPool places an entry into a transient store,
+// to track that this pool changed this block.
+// This tracking is for use in EndBlock, to create new TWAP records.
 func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
 	store := ctx.TransientStore(k.transientKey)
 	poolIdBz := make([]byte, 8)
@@ -20,6 +23,9 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
 	store.Set(poolIdBz, sentinelExistsValue)
 }
 
+// getChangedPools returns all poolIDs that changed this block.
+// This is to be guaranteed by trackChangedPool being called on every
+// price-affecting pool action.
 func (k Keeper) getChangedPools(ctx sdk.Context) []uint64 {
 	store := ctx.TransientStore(k.transientKey)
 	iter := store.Iterator(nil, nil)
@@ -34,6 +40,7 @@ func (k Keeper) getChangedPools(ctx sdk.Context) []uint64 {
 	return alteredPoolIds
 }
 
+// storeHistoricalTWAP writes a twap to the store, in all needed indexing.
 func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
 	store := ctx.KVStore(k.storeKey)
 	key1 := types.FormatHistoricalTimeIndexTWAPKey(twap.Time, twap.PoolId, twap.Asset0Denom, twap.Asset1Denom)
@@ -55,6 +62,11 @@ func (k Keeper) deleteHistoricalRecord(ctx sdk.Context, twap types.TwapRecord) {
 	store.Delete(key2)
 }
 
+// getMostRecentRecordStoreRepresentation returns the most recent twap record in the store
+// for the provided (pool, asset0, asset1) triplet.
+// Its called store representation, because most recent record can refer to it being
+// interpolated to the current block time, or after events in this block.
+// Neither of which apply to the record this returns.
 func (k Keeper) getMostRecentRecordStoreRepresentation(ctx sdk.Context, poolId uint64, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
 	store := ctx.KVStore(k.storeKey)
 	key := types.FormatMostRecentTWAPKey(poolId, asset0Denom, asset1Denom)
@@ -62,11 +74,14 @@ func (k Keeper) getMostRecentRecordStoreRepresentation(ctx sdk.Context, poolId u
 	return types.ParseTwapFromBz(bz)
 }
 
+// getMostRecentRecordStoreRepresentation returns all most recent twap records
+// (in state representation) for the provided pool id.
 func (k Keeper) getAllMostRecentRecordsForPool(ctx sdk.Context, poolId uint64) ([]types.TwapRecord, error) {
 	store := ctx.KVStore(k.storeKey)
 	return types.GetAllMostRecentTwapsForPool(store, poolId)
 }
 
+// storeNewRecord stores a record, in both the most recent record store and historical stores.
 func (k Keeper) storeNewRecord(ctx sdk.Context, twap types.TwapRecord) {
 	store := ctx.KVStore(k.storeKey)
 	key := types.FormatMostRecentTWAPKey(twap.PoolId, twap.Asset0Denom, twap.Asset1Denom)
@@ -74,8 +89,17 @@ func (k Keeper) storeNewRecord(ctx sdk.Context, twap types.TwapRecord) {
 	k.storeHistoricalTWAP(ctx, twap)
 }
 
-// returns an error if theres no historical record at or before time.
-// (Asking for a time too far back)
+// getRecordAtOrBeforeTime on a given input (id, t, asset0, asset1)
+// returns the TWAP record from state for (id, t', asset0, asset1),
+// where t' is such that:
+// * t' <= t
+// * there exists no t'' in state, where t'' < t'
+//
+// This returns an error if:
+// * there is no historical record in state at or before t
+//   - Occurs if t is older than pruning period, or pool creation date.
+// * there is no record for the asset pair (asset0, asset1) in particular
+//   - e.g. asset not in pool, or provided in wrong order.
 func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.Time, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
 	store := ctx.KVStore(k.storeKey)
 	// We make an iteration from time=t + 1ns, to time=0 for this pool.
--- a/x/gamm/twap/store.go
+++ b/x/gamm/twap/store.go
@@ -4,6 +4,7 @@ package twap
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -139,6 +140,6 @@ func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.T
 			return twap, nil
 		}
 	}
-	return types.TwapRecord{}, errors.New("something went wrong - TWAP not found, but there are twaps available for this time." +
-		" Were provided asset0denom and asset1denom correct, and in order (asset0 > asset1)?")
+	return types.TwapRecord{}, fmt.Errorf("TWAP not found, but there are twaps available for this time."+
+		" Were provided asset0denom and asset1denom (%s, %s) correct, and in order (asset0 > asset1)?", asset0Denom, asset1Denom)
 }
--- a/x/gamm/twap/store.go
+++ b/x/gamm/twap/store.go
@@ -2,8 +2,13 @@ package twap
 
 import (
 	"encoding/binary"
+	"errors"
+	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v10/osmoutils"
+	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
 )
 
 func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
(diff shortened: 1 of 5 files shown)


x/gamm/twap/types/keys.go
```diff
@@ -83,8 +83,8 @@ func ParseTimeFromHistoricalPoolIndexKey(key []byte) (time.Time, error) {
 }
 
 func GetAllMostRecentTwapsForPool(store sdk.KVStore, poolId uint64) ([]TwapRecord, error) {
-	startPrefix := fmt.Sprintf("%s%s%d%s", mostRecentTWAPsPrefix, KeySeparator, poolId, KeySeparator)
-	endPrefix := fmt.Sprintf("%s%s%d%s", mostRecentTWAPsPrefix, KeySeparator, poolId+1, KeySeparator)
+	startPrefix := fmt.Sprintf("%s%d%s", mostRecentTWAPsPrefix, poolId, KeySeparator)
+	endPrefix := fmt.Sprintf("%s%d%s", mostRecentTWAPsPrefix, poolId+1, KeySeparator)
```
ValarDragon, line 87: Bug fix found by tests over old twap_types PR

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
+)
+
+// TestTrackChangedPool takes a list of poolIds as test cases, and runs one list per block.
+// Every simulated block, checks that there no changed pools.
+// Then runs k.trackChangedPool on every item in the test case list.
+// Then checks that changed pools returns the list, deduplicated.
+//
+// This achieves testing the functionality that we depend on, that this clears every end block.
```
p0mvn, line 14: Nor sure if this is already documented elsewhere but I think we should highlight in README that transient store is what enables these to be cleared every block. Might not be an obvious thing to inexperienced sdk devs

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+	tPlusOne := baseTime.Add(time.Second)
+	baseRecord := newEmptyPriceRecord(1, baseTime, "tokenB", "tokenA")
+	tPlusOneRecord := newEmptyPriceRecord(1, tPlusOne, "tokenB", "tokenA")
+	tests := map[string]struct {
+		recordsToSet    []types.TwapRecord
+		poolId          uint64
+		expectedRecords []types.TwapRecord
+	}{
+		"set single record": {
+			[]types.TwapRecord{baseRecord},
```
p0mvn, line 58: nit: suggest making these named for readability

e.g.
```
"set single record": {
	recordsToSet: []types.TwapRecord{baseRecord},
	poolId: 1,
	expectedRecords: []types.TwapRecord{baseRecord},
},
```

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+	}
+
+	for name, test := range tests {
+		s.Run(name, func() {
+			s.SetupTest()
+			for _, record := range test.recordsToSet {
+				s.twapkeeper.StoreNewRecord(s.Ctx, record)
+			}
+			actualRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, test.poolId)
```
p0mvn, line 97: q: is `types` package a good location for `GetAllMostRecentTwapsForPool`? I think the only place where it is called right now is from keeper method that is unexported

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+		recordsToSet    []types.TwapRecord
+		poolId          uint64
+		expectedRecords []types.TwapRecord
+	}{
+		"set single record": {
+			[]types.TwapRecord{baseRecord},
+			1,
+			[]types.TwapRecord{baseRecord},
+		},
+		"query non-existent pool": {
```
p0mvn, line 62: Why doesn't this case produce an error?

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		"set two records": {
+			[]types.TwapRecord{baseRecord, tPlusOneRecord},
+			1,
+			[]types.TwapRecord{tPlusOneRecord},
+		},
+		"set two records, reverse order": {
+			// The last record, independent of time, takes precedence for most recent.
+			[]types.TwapRecord{tPlusOneRecord, baseRecord},
+			1,
```
p0mvn, line 75: consider making a record with a poolId other than 1 and having 1 test case for that

The motivation is to have a sanity check that we don't have a dependency on any specific pool id (1 in this case)

p0mvn (commented):
Partially through, only have `TestGetRecordAtOrBeforeTime` left. Everything looks great so far, clean implementation + tests and the comments are very helpful!

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+				s.twapkeeper.StoreNewRecord(s.Ctx, record)
+			}
+			actualRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, test.poolId)
+			s.Require().NoError(err)
+			s.Require().Equal(test.expectedRecords, actualRecords)
+		})
+	}
+}
+
+// TestGetAllMostRecentRecordsForPool takes a list of records as test cases,
```
p0mvn, line 104:
```suggestion
// TestGetRecordAtOrBeforeTime takes a list of records as test cases,
```

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+	tPlus1 := baseTime.Add(time.Second)
+	tPlus1Record := newEmptyPriceRecord(1, tPlus1, "tokenB", "tokenA")
+
+	tests := map[string]struct {
+		recordsToSet   []types.TwapRecord
+		input          getRecordInput
+		expectedRecord types.TwapRecord
+		expErr         bool
+	}{
+		"no entries":            {[]types.TwapRecord{}, defaultInputAt(baseTime), baseRecord, true},
```
p0mvn, line 129: please consider making these named and each test case field on a new line for readability

x/gamm/twap/store_test.go
p0mvn: Similar suggestion for `TestGetRecordAtOrBeforeTime`

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+		s.Run(name, func() {
+			s.SetupTest()
+			for _, record := range test.recordsToSet {
+				s.twapkeeper.StoreNewRecord(s.Ctx, record)
+			}
+			record, err := s.twapkeeper.GetRecordAtOrBeforeTime(
+				s.Ctx,
+				test.input.poolId, test.input.t, test.input.asset0Denom, test.input.asset1Denom)
+			if test.expErr {
+				s.Require().Error(err)
```
p0mvn, line 174: consider asserting on what error specifically is returned

x/gamm/twap/store.go
```diff
@@ -15,17 +24,121 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
...
+// * there is no record for the asset pair (asset0, asset1) in particular
+//   - e.g. asset not in pool, or provided in wrong order.
+func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.Time, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
+	store := ctx.KVStore(k.storeKey)
+	// We make an iteration from time=t + 1ns, to time=0 for this pool.
+	// Note that we cannot get any time entries from t + 1ns, as the key would be `prefix|t+1ns`
+	// and the end for a reverse iterator is exclusive. Thus the largest key that can be returned
+	// begins with a prefix of `prefix|t`
+	startKey := types.FormatHistoricalPoolIndexTimePrefix(poolId, time.Unix(0, 0))
+	endKey := types.FormatHistoricalPoolIndexTimePrefix(poolId, t.Add(time.Nanosecond))
```
p0mvn, line 111: Example from one of the test cases:
```
"historical_pool_index|1|2009-11-10T23:00:00.000000001|"
```

Is key separator needed at the end of the key if time format is standardized? Seems like a redundant byte in db

x/gamm/twap/store.go
```diff
@@ -15,17 +24,121 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
...
+	lastParsedTime := time.Time{}
+	stopFn := func(key []byte) bool {
+		// halt iteration if we can't parse the time, or we've successfully parsed
+		// a time, and have iterated beyond records for that time.
+		parsedTime, err := types.ParseTimeFromHistoricalPoolIndexKey(key)
+		if err != nil {
+			return true
```
p0mvn, line 118: Is returning true correct here? Wdyt about making `stopFn` return an error?

x/gamm/twap/store.go
```diff
@@ -15,17 +24,121 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
...
+	if err != nil {
+		return types.TwapRecord{}, err
+	}
+	if len(twaps) == 0 {
+		return types.TwapRecord{}, errors.New("looking for a time thats too old, not in the historical index. " +
+			" Try storing the accumulator value.")
+	}
+
+	for _, twap := range twaps {
+		if twap.Asset0Denom == asset0Denom && twap.Asset1Denom == asset1Denom {
```
p0mvn, line 138: Since denoms exist in a key, do we need this loop? 

We can probably make the `stopFn` stop once the right denoms are reached. Then, grab the last value

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
+package twap_test
+
+import (
+	"time"
+
+	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
+)
+
```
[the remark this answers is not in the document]
ValarDragon: This isn't right though, It should only appear once per module, otherwise all tests run twice, this was an issue with mint as well, that was fixed in #2137

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,181 @@
...
+				cumulativeIds[v] = true
+				s.twapkeeper.TrackChangedPool(s.Ctx, v)
+			}
+
+			changedPools = s.twapkeeper.GetChangedPools(s.Ctx)
+			s.Require().Equal(len(cumulativeIds), len(changedPools))
+			for _, v := range changedPools {
+				s.Require().True(cumulativeIds[v])
+			}
+			s.Commit()
```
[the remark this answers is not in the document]
ValarDragon: I feel like in this case its good - every unit test here should work regardless of prior state though? Were testing more configurations by doing this, no?

I agree for most unit tests, but this is specifically about store operations and clearing behavior

x/gamm/twap/store_test.go
ValarDragon: it returns a blank record list

x/gamm/twap/store_test.go
ValarDragon: I don't think thats important personally

x/gamm/twap/store.go
```diff
@@ -1,11 +1,20 @@
...
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v10/osmoutils"
+	"github.com/osmosis-labs/osmosis/v10/x/gamm/twap/types"
 )
 
+// trackChangedPool places an entry into a transient store,
+// to track that this pool changed this block.
+// This tracking is for use in EndBlock, to create new TWAP records.
```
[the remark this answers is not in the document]
ValarDragon: I don't think this is going to show up in any system benchmarks as a measurable bottleneck though

x/gamm/twap/store.go
ValarDragon: This is the prefix though? We iterate based off entries with the prefix, we need that separator to get all assets.

We could claim that its not needed due to fixed size times, but I'd rather not rely on that too explicitly for simplicity.

A bigger saving would be reconsidering using a fixed size unix time serialization than the key separator.

x/gamm/twap/store.go
ValarDragon: This is correct imo, the behavior of the iterator is to stop here, it accidentally iterated too far

x/gamm/twap/store.go
ValarDragon: This feels far simpler though

ValarDragon: Happy for us to explore the option your describing in a subsequent pass if you'd like

x/gamm/twap/store_test.go
ValarDragon: agreed here, since its already split into newlines!

x/gamm/twap/store_test.go
ValarDragon: I found this more readable personally

x/gamm/twap/store_test.go
ValarDragon: Agreed its a bit awkward there, but it has the advantage that all key formatting is encapsulated into that file.

ValarDragon:
I believe that all comments are now either addressed or replied to!

Thanks for the great review!

x/gamm/twap/store_test.go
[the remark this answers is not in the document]
czarcas7ic: The way you do this without running tests multiple times is by adding the following instead:

`var _ = suite.TestingSuite(nil)`

x/gamm/twap/store_test.go
ValarDragon: Adding to README in the main PR

ValarDragon: Added!

x/gamm/twap/store_test.go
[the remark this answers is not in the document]
p0mvn: ahh didn't realize that

@czarcas7ic this doesn't work for me. How did you get it to work?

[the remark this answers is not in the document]
ValarDragon: What I do is I have vscode run all tests in a package on save

x/gamm/twap/store.go
p0mvn: https://github.com/osmosis-labs/osmosis/issues/2251

x/gamm/twap/store.go
p0mvn: Your explanation makes sense. As long as there is a good reason for this choice (and there is), I'm good with the approach :+1:

p0mvn (approved):
LGTM. I resolved all addressed earlier addressed comments.

There are some remaining discussions but they are more nice-to-haves and should not be blocking merge IMO

x/gamm/twap/store.go
```diff
@@ -15,17 +25,121 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
...
+// interpolated to the current block time, or after events in this block.
+// Neither of which apply to the record this returns.
+func (k Keeper) getMostRecentRecordStoreRepresentation(ctx sdk.Context, poolId uint64, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
+	store := ctx.KVStore(k.storeKey)
+	key := types.FormatMostRecentTWAPKey(poolId, asset0Denom, asset1Denom)
+	bz := store.Get(key)
+	return types.ParseTwapFromBz(bz)
+}
+
+// getMostRecentRecordStoreRepresentation returns all most recent twap records
```
czarcas7ic, line 79:
```suggestion
// getAllMostRecentRecordsForPool returns all most recent twap records
```

x/gamm/twap/store.go
```diff
@@ -15,17 +25,121 @@ func (k Keeper) trackChangedPool(ctx sdk.Context, poolId uint64) {
...
+//
+// This returns an error if:
+// * there is no historical record in state at or before t
+//   - Occurs if t is older than pruning period, or pool creation date.
+// * there is no record for the asset pair (asset0, asset1) in particular
+//   - e.g. asset not in pool, or provided in wrong order.
+func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.Time, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
+	store := ctx.KVStore(k.storeKey)
+	// We make an iteration from time=t + 1ns, to time=0 for this pool.
+	// Note that we cannot get any time entries from t + 1ns, as the key would be `prefix|t+1ns`
```
czarcas7ic, line 108: Does this logic change if we decide to go through with https://github.com/osmosis-labs/osmosis/issues/2251?

x/gamm/twap/store_test.go
```diff
@@ -0,0 +1,191 @@
...
+		"set multi-asset pool record": {
+			recordsToSet: []types.TwapRecord{
+				newEmptyPriceRecord(1, baseTime, "tokenB", "tokenA"),
+				newEmptyPriceRecord(1, baseTime, "tokenC", "tokenB"),
+				newEmptyPriceRecord(1, baseTime, "tokenC", "tokenA")},
+			poolId: 1,
+			expectedRecords: []types.TwapRecord{
+				newEmptyPriceRecord(1, baseTime, "tokenB", "tokenA"),
+				newEmptyPriceRecord(1, baseTime, "tokenC", "tokenA"),
+				newEmptyPriceRecord(1, baseTime, "tokenC", "tokenB")},
```
czarcas7ic, line 92: Was the order of records here switched on purpose or does it just not matter? Just asking to ensure I fully understand.

czarcas7ic (approved):

x/gamm/twap/store_test.go
ValarDragon: getAllMostRecentTwapsForPool is a private method, that is only exported in export_test. If your looking at it, your supposed to be looking at the implementation details - its not for outside consumers!

x/gamm/twap/store.go
ValarDragon: Not really, unix time in golang is still nano seconds based. But we could just safe guard against the +1ns case

x/gamm/twap/store_test.go
ValarDragon: The returned order is the order that our helper for getting all pairs returns

x/gamm/twap/store_test.go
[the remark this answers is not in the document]
ValarDragon: This is explicitly documented that this is intentionally the case though?

Its quite different imo then ones that have non-sensical dependencies without documentation?

Issue #2183, opened 2022-07-21
Complete msgs_test.go for the lockup module


czarcas7ic:
## Background

As part of the simulator work, we found that the lockup module did not have any tests for its messages. I created a msgs_test.go file there and completed it for `MsgLockTokens`, but the remaining messages must be completed.

## Suggested Design

See initial implementation here: https://github.com/osmosis-labs/osmosis/blob/72477ccb738fdba3c70c15420245980660e89e44/x/lockup/types/msgs_test.go

Also see another module's implementation here: https://github.com/osmosis-labs/osmosis/blob/1a9d0bcf8de96b1d6cc382ee0f72ec32928c2860/x/gamm/types/msgs_test.go

## Acceptance Criteria

Tests are created and pass for the remaining messages
- [ ] MsgBeginUnlockingAll
- [ ] MsgBeginUnlocking
- [ ] MsgExtendLockup

PR #2205, opened 2022-07-23
Loockup msg server test

hieuvubk:
Closes: #2183

> Add tests for lockup module msg_server.test. Add some missing ValidateBasic.

- [x] MsgBeginUnlockingAll

- [x] MsgBeginUnlocking

- [x] MsgExtendLockup


x/lockup/types/msgs.go
```diff
@@ -98,6 +102,24 @@ func NewMsgBeginUnlocking(owner sdk.AccAddress, id uint64, coins sdk.Coins) *Msg
...
+	_, err := sdk.AccAddressFromBech32(m.Owner)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid owner address (%s)", err)
+	}
+
+	if m.ID == 0 {
+		return fmt.Errorf("invalid lockup ID, got %v", m.ID)
+	}
+
+	// allow unlocks with one denom same with lock tokens
```
czarcas7ic, line 114:
```suggestion
	// only allow unlocks with a single denom
```

x/lockup/types/msgs_test.go
```diff
@@ -95,10 +95,224 @@ func TestMsgLockTokens(t *testing.T) {
 	}
 }
 
-// TODO: Complete table driven tests for the remaining messages
+func TestMsgBeginUnlockingAll(t *testing.T) {
+	appParams.SetAddressPrefixes()
+	pk1 := ed25519.GenPrivKey().PubKey()
+	addr1 := sdk.AccAddress(pk1.Address()).String()
+	invalidAddr := sdk.AccAddress("invalid")
```
czarcas7ic, line 102: Seems like we need to generate these three vars for every test. Maybe we can create a function to do this to reduce boilerplate?

x/lockup/types/msgs_test.go
```diff
@@ -95,10 +95,224 @@ func TestMsgLockTokens(t *testing.T) {
...
+			}),
+			expectPass: true,
+		},
+		{
+			name: "invalid owner",
+			msg: createMsg(func(msg MsgBeginUnlockingAll) MsgBeginUnlockingAll {
+				msg.Owner = invalidAddr.String()
+				return msg
+			}),
+			expectPass: false,
```
czarcas7ic, line 142:
```suggestion
```

expectPass defaults to false, so every occurrence of false can be removed from all the tests to make the tests shorter.

czarcas7ic (approved):
LGTM! 👍

x/lockup/types/msgs_test.go
hieuvubk: Yes thanks for the suggestion. I created a function to gen valid & invalid addresses for all tests

czarcas7ic:
This all LGTM, much cleaner! Going to wait on one more review in case I missed anything then should be good to merge!

x/lockup/types/msgs_test.go
```diff
@@ -95,10 +89,217 @@ func TestMsgLockTokens(t *testing.T) {
...
+	msg := createMsg(func(msg MsgBeginUnlockingAll) MsgBeginUnlockingAll {
+		// Do nothing
+		return msg
+	})
+
+	require.Equal(t, msg.Route(), RouterKey)
+	require.Equal(t, msg.Type(), "begin_unlocking_all")
+	signers := msg.GetSigners()
+	require.Equal(t, len(signers), 1)
+	require.Equal(t, signers[0].String(), addr1)
```
p0mvn, line 113: Can we please instead make the assertions inside the for loop for each test case?

Also, let's try to avoid the complexity with `createMsg` callbacks, please. I think we can directly initialize:
```
msg: MsgBeginUnlockingAll{
			Owner: addr1,
}
```
instead of having the callback overhead

x/lockup/types/msgs_test.go
```diff
@@ -95,10 +89,217 @@ func TestMsgLockTokens(t *testing.T) {
...
+			name: "invalid owner",
+			msg: createMsg(func(msg MsgBeginUnlockingAll) MsgBeginUnlockingAll {
+				msg.Owner = invalidAddr
+				return msg
+			}),
+		},
+	}
+
+	for _, test := range tests {
+		if test.expectPass {
```
p0mvn, line 138: Let's make sure each test case has its own context pelase

`t.Run(test.name, ...)`

p0mvn (changes requested):

x/lockup/types/msgs_test.go
hieuvubk: Yes thank you. I think it makes the code look better.

x/lockup/types/msgs_test.go
```diff
@@ -14,91 +14,246 @@ import (
...
+				require.Equal(t, signers[0].String(), addr1)
+			} else {
+				require.Error(t, test.msg.ValidateBasic(), "test: %v", test.name)
+			}
+		})
+	}
+}
 
-// MsgExtendLockup
+func generateTestAddrs() (string, string) {
```
p0mvn, line 254: Can we move this helper to test utils?

x/lockup/types/msgs_test.go
```diff
@@ -14,91 +14,246 @@ import (
...
+			} else {
+				require.Error(t, test.msg.ValidateBasic(), "test: %v", test.name)
+			}
+		})
+		
 	}
 }
 
-// TODO: Complete table driven tests for the remaining messages
+func TestMsgExtendLockup(t *testing.T) {
```
p0mvn, line 194: Does it make sense to define these tests on the test suite?

p0mvn (approved):
LGTM, great work @hieuvubk 

I have 2 follow-up questions. If questions make sense, can you please either address them or create issues for them to be addressed? Thank you

x/lockup/types/msgs_test.go
hieuvubk: Yes I'll do it now

x/lockup/types/msgs_test.go
hieuvubk: Do u mean to use the test suite instead of testing.T?

p0mvn: Sorry for not being clear enough.

What I meant is to define something like this:
https://github.com/osmosis-labs/osmosis/blob/fe98f6e4453cd32035d277c21ef2f3669b677bb2/x/lockup/keeper/keeper_test.go#L15

Instead, we would have `TypesTestSuite`

x/lockup/types/msgs_test.go
p0mvn: What I meant is to define it on the app testing suite:
https://github.com/osmosis-labs/osmosis/blob/450e837158f4ae4cb908ec496da47fc169007218/app/apptesting/test_suite.go#L273

If you go with my other suggestion of [utilizing test suite](https://github.com/osmosis-labs/osmosis/pull/2205#discussion_r933548244), it would be convenient to reuse

x/lockup/types/msgs_test.go
hieuvubk: I think we should create an issue for that. I just took a look through all modules and tests in `types` all use testing.T
Should refactor all those tests in a pr right?

p0mvn: If that's the case, let's keep this as is. I think consistency is more important here. Thanks for double-checking that

p0mvn (approved):
:rocket:

p0mvn:
Merging since @czarcas7ic previously approved this and only my comments have been applied since

Issue #2187, opened 2022-07-22
Cosmwasm bindings for TWAP


ValarDragon:
## Background

Make a cosmwasm binding for the TWAP work. 

## Suggested Design

Should be a binding around the exposed functions in the `x/gamm/twap/api.go` file, namely `GetArithmeticTwap` and `GetArithmeticTwapToNow`.

We should keep an eye out towards how can we store our own records cosmwasm side, and then use the record based API once it exists.

## Acceptance Criteria

- [x] Cosmwasm binding for TWAP
- [x] Example contract invoking it

Issue #2209, opened 2022-07-23
Twap: Make timeDelta for calculation in miliseconds


ValarDragon:
## Background

Currently all of our time deltas in the TWAP code cast time durations to their integer representations, which is in nanoseconds.
Having nanoseconds precision is very unnecessary as Tendermint forces every block time to be at least 1ms since the last, and because this costs us in precision for the Arithmetic accumulator, which we have to think about.

I suggest we make a helper method in types, for taking the difference of two times, and then rounding to the nearest milisecond, and then dividing by 10**6.

This will help save us on precision, as we start working towards overflow handling in the TWAP logic, and thinking about max spot price precisions.

## Acceptance Criteria

- [ ] Implementation of previously mentioned

stackman27:
Hi @ValarDragon can we achieve this by doing something simple like this;


```
x/gamm/twap/types/utils.go

func SpotPriceTimesDuration(sp sdk.Dec, timeDelta time.Duration) sdk.Dec {
	deltaMilliSecond := timeDelta.Milliseconds()
	return sp.MulInt64(int64(deltaMilliSecond))
}

func AccumDiffDivDuration(accumDiff sdk.Dec, timeDelta time.Duration) sdk.Dec {
	deltaMilliSecond := timeDelta.Milliseconds()
	return accumDiff.QuoInt64(int64(deltaMilliSecond))
}
```

PR #2212, opened 2022-07-23
Twap Wasm Bindings

mattverse:
Sub component of : #[2187](https://github.com/osmosis-labs/osmosis/issues/2187)

This PR adds wasm bindings for query on two APIs `GetArithmeticTWAP` and `GetArithmeticTwapToNow`. 

This PR invlolves adding  twap keeper to the wasm query plugin to use the API provided by the twap keeper.

## Brief Changelog

- Add wasm bindings for twap APIs(`GetArithmeticTWAP`, `GetArithmeticTWAPToNow`)

## Testing and Verifying

Testing is TBD and to be sorted out, as of right now, do not have exact test cases for testing. 


p0mvn (approved):
utACK

mattverse:
Figuring plans for testing w/ exact numbers for TWAP

mattverse:
Converting to draft.

Current status for this PR is that I've added test codes for new queries in custom bindings test, whilst getting error `Error parsing into type osmo_reflect::msg::QueryMsg: unknown variant `arithmetic_twap`, expected one of `full_denom`, `pool_state`, `spot_price`, `estimate_swap`: query wasm contract failed`. Further investigation needed in order to figure out where this is coming from / why / where this is happening

ValarDragon:
This LGTM!

nicolaslara:
This looks good to me.

My only comment is that it's not clear which queries should have their own bindings and which should go through `StargateQuery` with convenience wrappers in https://github.com/osmosis-labs/osmosis-rust. I don't see manipulation of the query result to avoid non-determinism, in which case adding it to the stargate query whitelist should have the same effect.

wasmbinding/bindings/query.go
```diff
@@ -52,6 +56,21 @@ type EstimateSwap struct {
...
+	BaseAssetDenom  string `json:"base_asset_denom"`
+	StartTime       int64  `json:"start_time"`
+	EndTime         int64  `json:"end_time"`
+}
+
+type ArithmeticTwapToNow struct {
+	PoolId          uint64 `json:"id"`
+	QuoteAssetDenom string `json:"quote_asset_denom"`
+	BaseAssetDenom  string `json:"base_asset_denom"`
+	StartTime       int64  `json:"start_time"`
```
ValarDragon, line 71: Can you add a comment specifying its UnixTime?

wasmbinding/test/custom_query_test.go
```diff
@@ -172,6 +174,97 @@ func TestQuerySpotPrice(t *testing.T) {
...
+		sdk.NewInt64Coin("uosmo", 100000000),
+		sdk.NewInt64Coin("ustar", 200000000),
+	}
+	// 20 star to 1 osmo
+	starPoolID := preparePool(t, ctx, osmosis, actor, poolFunds)
+
+	reflect := instantiateReflectContract(t, ctx, osmosis, actor)
+	require.NotEmpty(t, reflect)
+
+	// TODO: figure out why it errors when we use start time without second
```
ValarDragon, line 193: wdym

ValarDragon: Oh I see, the query errors in this case, interesting

ValarDragon: here's the error message: `looking for a time thats too old, not in the historical index.  Try storing the accumulator value.`

ValarDragon: ah the error is correct, this is a test error here =p

Committing fix here.

ValarDragon: Hrmm actually its not so clean, this is actually a deeper issue with the binding using Unix time for the time passing here.

The reason its not working at the start time is because its being passed in unix time in miliseocnds. However the state record has greater precision.

Trying to see more about how to fix this

ValarDragon: So I think what we need to do is make the TWAP binding parse from UnixMiliseconds, and then in the state machine we write at UnixMilisecond precision.

For this PR, can we change the binding to be from Unix milisecond and then call it good from that? Then in the test if we set ctx.BlockTime to be at a milisecond boundary, this should be fixed. Once we do https://github.com/osmosis-labs/osmosis/issues/2209 should also get fixed state machine side as well.

ValarDragon:
@mattverse please check my latest commit to fix the time threading.

if you agree, please make docs for the binding, for start time being in Unix time miliseconds. Also communicate to Mars that we can change the time input API to be whatever they'd prefer. 

Right now there will be an edge case with StartTime for a TWAP, that will get fixed in a subsequent commit pre-release. Namely that if start time = pool creation time, and the block time is not milisecond aligned, it may not work. Setting it to be 1 milisecond after pool creation time would fix this for now as workaround, until next week we change it state machine to not be nanosecond precise in state machine storage.

(The state machine computes the correct result, its that this binding API doesn't offer sufficient precision)

mattverse:
Agreed with using Unix time millisecond for now for precision, updated documentation

ValarDragon:
(I rebased this after I merged the twap branch)

wasmbinding/queries.go
```diff
@@ -103,3 +107,40 @@ func (qp QueryPlugin) EstimateSwap(ctx sdk.Context, estimateSwap *bindings.Estim
...
+
+func (qp QueryPlugin) ArithmeticTwapToNow(ctx sdk.Context, arithmeticTwap *bindings.ArithmeticTwapToNow) (*sdk.Dec, error) {
+	if arithmeticTwap == nil {
+		return nil, wasmvmtypes.InvalidRequest{Err: "gamm arithmetic twap null"}
+	}
+
+	poolId := arithmeticTwap.PoolId
+	quoteAssetDenom := arithmeticTwap.QuoteAssetDenom
+	baseAssetDenom := arithmeticTwap.BaseAssetDenom
+	startTime := time.Unix(arithmeticTwap.StartTime, 0)
```
ValarDragon, line 138: ah we forgot to make this milli in the version we sent to mars

czarcas7ic: I can spin one up fairly quickly if this needs to be changed

PR #2223, opened 2022-07-24
feat:  Tokenfactory SetMetadata

sunnya97:

This PR allows the admin of a tokenfactory denom to set its bank.Metadata, thereby changing its display units, name, and symbol.

## Brief Changelog

## Testing and Verifying

This change added unit tests in x/tokenfactory/keeper/admins_test.go.

  - How is the feature or change documented? x/tokenfactory/README


x/tokenfactory/keeper/msg_server.go
```diff
@@ -158,3 +159,39 @@ func (server msgServer) ChangeAdmin(goCtx context.Context, msg *types.MsgChangeA
...
+func (server msgServer) SetDenomMetadata(goCtx context.Context, msg *types.MsgSetDenomMetadata) (*types.MsgSetDenomMetadataResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Defense in depth validation of metadata
+	err := msg.Metadata.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println(msg.Metadata.Base)
```
sunnya97, line 172:
```suggestion
```

x/tokenfactory/keeper/msg_server.go
```diff
@@ -158,3 +159,37 @@ func (server msgServer) ChangeAdmin(goCtx context.Context, msg *types.MsgChangeA
...
+		return nil, err
+	}
+
+	authorityMetadata, err := server.Keeper.GetAuthorityMetadata(ctx, msg.Metadata.Base)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println(msg.Sender)
+	fmt.Println(authorityMetadata)
```
sunnya97, line 178:
```suggestion
```

x/tokenfactory/README.md
```diff
@@ -99,6 +99,18 @@ message MsgChangeAdmin {
 }
 ```
 
+### SetDenomMetadata
+
+Setting of metadata for a specific denom is only allowed for the admin of the denom.
+It allows the overwriting of the denom metadata in the bank module.
+
+```go
+message MsgChangeAdmin {
```
mattverse, line 108:
```suggestion
message MsgSetDenomMetadata {
```

x/tokenfactory/keeper/msg_server.go
```diff
@@ -158,3 +158,34 @@ func (server msgServer) ChangeAdmin(goCtx context.Context, msg *types.MsgChangeA
...
+
+	if msg.Sender != authorityMetadata.GetAdmin() {
+		return nil, types.ErrUnauthorized
+	}
+
+	server.Keeper.bankKeeper.SetDenomMetaData(ctx, msg.Metadata)
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.TypeMsgChangeAdmin,
```
mattverse, line 184:
```suggestion
			types.TypeMsgSetDenomMetadata,
```

mattverse (approved):
LGTM! 

Can we register codec for this as well before merging this?

ValarDragon:
Thus conflicts with the goal of the state machine having a reverse map from human readable name to bank denom

sunnya97:
I think the metadata stored in the bank module should a token's self-declared name/metadata (similar to how all ERC20s can self-declare their metadata).  Then, the state machine can have its own on-chain curated assetlist, separate from the self-declared info).

x/tokenfactory/keeper/admins_test.go
```diff
@@ -267,3 +268,131 @@ func (suite *KeeperTestSuite) TestChangeAdminDenom() {
...
+						Denom:    "uosmo",
+						Exponent: 6,
+					},
+				},
+				Base:    fmt.Sprintf("factory/%s/litecoin", suite.TestAccs[0].String()),
+				Display: "uosmo",
+				Name:    "OSMO",
+				Symbol:  "OSMO",
+			}),
+			expectedPass: false,
```
p0mvn, line 323: it's false by default so can be omitted
```suggestion
```

x/tokenfactory/keeper/admins_test.go
```diff
@@ -267,3 +268,131 @@ func (suite *KeeperTestSuite) TestChangeAdminDenom() {
...
+						Denom:    "uosmoo",
+						Exponent: 6,
+					},
+				},
+				Base:    "uosmo",
+				Display: "uosmoo",
+				Name:    "OSMO",
+				Symbol:  "OSMO",
+			}),
+			expectedPass: false,
```
p0mvn, line 344:
```suggestion
```

x/tokenfactory/keeper/admins_test.go
```diff
@@ -267,3 +268,131 @@ func (suite *KeeperTestSuite) TestChangeAdminDenom() {
...
+				Base:    suite.defaultDenom,
+				Display: "uosmo",
+				Name:    "OSMO",
+				Symbol:  "OSMO",
+			}),
+			expectedPass: false,
+		},
+	} {
+		suite.Run(fmt.Sprintf("Case %s", tc.desc), func() {
+			_, err := suite.msgServer.SetDenomMetadata(sdk.WrapSDKContext(suite.Ctx), &tc.msgSetDenomMetadata)
```
p0mvn, line 386: nit: we can also make an assertion that the response is not nil

x/tokenfactory/types/msgs.go
```diff
@@ -202,3 +204,43 @@ func (m MsgChangeAdmin) GetSigners() []sdk.AccAddress {
...
+
+	return nil
+}
+
+func (m MsgSetDenomMetadata) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&m))
+}
+
+func (m MsgSetDenomMetadata) GetSigners() []sdk.AccAddress {
+	sender, _ := sdk.AccAddressFromBech32(m.Sender)
```
p0mvn, line 244: Can error be propagated to the caller?

p0mvn (approved):

x/tokenfactory/types/msgs.go
sunnya97: I don't think so.  This is part of the sdk.Msg interface, so we can't change the function signature

x/tokenfactory/keeper/admins_test.go
sunnya97: I think I prefer to still see visually when reading the code that this test case is expected to fail.

x/tokenfactory/keeper/admins_test.go
sunnya97: Done

ValarDragon:
I'm pretty concerned about this, as metadata already has users with people expecting safety guarantees.

E.g. integration into ledger

sunnya97:
integration into ledger doesn't use metadata right now?

ValarDragon:
Oh I thought it read from bank metadata, its actually just hardcoded into ledger: https://github.com/cosmos/ledger-cosmos/blob/78a1a260ee66702cc21c81dbcbffef8ebe9647f5/app/src/coin.h

Issue #2191, opened 2022-07-22
Create a fee for Gauge Creation


ValarDragon:
## Background

We should make a fee for creating a gauge for a new locked denom (e.g. 50 osmo).

This fee should be based off of the msg_server message, rather than the keeper. We want the pool-incentives hook to not make a cost.

## Suggested Design

Make a new param for gauge creation fee, add it as a cost to the msg_server.

Ensure that pool creation still works with the hooks

## Acceptance Criteria

- [ ] Create gauge requires a cost from the sender
- [ ] pool incentives does not add a cost

czarcas7ic:
@vuong177 hey, I am just ensuring that you are aware @p0mvn and I are working on this right now. If you are looking to also do this to double check our work that is fine, I just wanted to make sure that you knew!

vuong177:
@czarcas7ic Oh,  I don't know you're working on this. ! Thank you so much for letting me know, and please continue your work.

PR #2227, opened 2022-07-24
feat(incentives)!: create gauge and add to gauge fee charge

p0mvn:
Closes: #2191

This change introduces a message fee for creating and adding to gauge.

Create gauge has a fee of 50 OSMO while adding to gauge 25. The feel is always subtracted in OSMO.

If the user does not have enough OSMO to create/add to gauge + pay the fee, no action is done and the message server returns an error.

## Brief Changelog

- `chargeFeeIfSufficientFeeDenomBalance` method create on the keeper
- test for `chargeFeeIfSufficientFeeDenomBalance`
- fee tests for `CreateGauge` in the `incentives` message server
- `chargeFeeIfSufficientFeeDenomBalance` is called from `CreateGauge` in `incentives` message server
- fee tests for `AddToGauge` in the `incentives` message server
- `chargeFeeIfSufficientFeeDenomBalance` is called from `AddToGauge` in `incentives` message server

## Testing and Verifying

- `TestCreateGauge_Fee`
- `TestAddToGauge_Fee`
- `TestChargeFeeIfSufficientFeeDenomBalance`

  - How is the feature or change documented? godocs


x/incentives/keeper/gauge.go
```diff
@@ -278,3 +286,27 @@ func (k Keeper) GetEpochInfo(ctx sdk.Context) epochtypes.EpochInfo {
...
+// chargeFeeIfSufficientFeeDenomBalance charges fee in the base tx fee denom on the address if the address has
+// balance that is less than fee + amount of the coin from gaugeCoins that is of base tx fee denom.
+// gaugeCoins might not have a coin of tx base fee denom. In that case, only fee is compared to balance.
+// Returns nil on success, error otherwise.
+func (k Keeper) chargeFeeIfSufficientFeeDenomBalance(ctx sdk.Context, address sdk.AccAddress, fee int64, gaugeCoins sdk.Coins) (err error) {
+	// Send creation fee to community pool
+	feeDenom, err := k.tk.GetBaseDenom(ctx)
+	if err != nil {
+		return err
+	}
```
sunnya97, line 299: I think we can just hardcode this to "uosmo" for now, instead of needing to query feeDenom.

Or if you want, you can use `appparams.BaseCoinUnit`

czarcas7ic:
Simulator was failing because we are essentially explicitly requiring the "uosmo" denom for the fees and we use the "stake" denom for the simulator. Normally I would just change the parameters from uosmo to stake, but because we are defining this directly in the msg_server, I am unable to override it. Regardless, I added a check in the simulator to only allow account with enough tokens in fees to create/add to gauges. Because no account has uosmo all attempts fail but the simulator will pass

czarcas7ic (approved):
As discussed offline, we should eventually fix the fact that a user will still pay the add to gauge fee if they submit an add to gauge message with a gaugeID that does not exist.

mattverse (approved):
LGTM

PR #2234, opened 2022-07-25
move gauge fees to types 

p0mvn:
- move fees to types and export
- reuse fees in the simulator

### DIFF
--- a/x/incentives/keeper/export_test.go
+++ b/x/incentives/keeper/export_test.go
@@ -6,13 +6,6 @@ import (
 	"github.com/osmosis-labs/osmosis/v10/x/incentives/types"
 )
 
-var (
-	// CreateGaugeFee is the fee required to create a new gauge.
-	CreateGaugeFee = createGaugeFee
-	// AddToGagugeFee is the fee required to add to gauge.
-	AddToGaugeFee = addToGaugeFee
-)
-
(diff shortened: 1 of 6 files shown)


czarcas7ic:
RandomSimAccountWithMinCoins should already do this for you no?

p0mvn:
> RandomSimAccountWithMinCoins should already do this for you no?

I was just trying things. The main problem here is that balances are set with "stake" denom in the SDK bank module's `GenerateGenesisState(...)`. So all accounts are guaranteed to not have the "uosmo" denom coins.

I think the cleanest solution would be to propagate txfees keeper to incentives and extract the denom. This can be done in a follow-up PR but not backported to `v11.x`.

Also, there is no access to `RandomSimAccountWithMinCoins` from the current simulation setup since it is defined on [`SimCxt`](https://github.com/osmosis-labs/osmosis/blob/3bfa6ba02481022bce59a26af8e03d69c92c773d/simulation/simtypes/simctx.go#L18). We would have to refactor simulation via `Actions()` like token factory to be able to access the `SimCtx`:
https://github.com/osmosis-labs/osmosis/blob/3bfa6ba02481022bce59a26af8e03d69c92c773d/x/tokenfactory/module.go#L198-L203

PR #2239, opened 2022-07-27
simulator: tokenfactory sim updates

czarcas7ic:
Closes: #XXX

Further simulator improvements (see changelog)

## Brief Changelog

- Adds tokenfactory keeper to the simulator
- Fetches creation fee amount instead of hardcoding
- Adds BurnDenom and ChangeAdmin messages

## Testing and Verifying

### DIFF
--- a/x/tokenfactory/module.go
+++ b/x/tokenfactory/module.go
@@ -23,6 +23,8 @@ import (
 	"github.com/spf13/cobra"
 	abci "github.com/tendermint/tendermint/abci/types"
 
+	appparams "github.com/osmosis-labs/osmosis/v10/app/params"
+
 	"github.com/osmosis-labs/osmosis/v10/simulation/simtypes"
 	simulation "github.com/osmosis-labs/osmosis/v10/x/tokenfactory/simulation"
 
@@ -189,7 +191,12 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 // GenerateGenesisState creates a randomized GenState of the tokenfactory module.
 func (am AppModule) SimulatorGenesisState(simState *module.SimulationState, s *simtypes.SimCtx) {
 	tfDefaultGen := types.DefaultGenesis()
-	tfDefaultGen.Params.DenomCreationFee = sdk.NewCoins(sdk.NewInt64Coin(sdk.DefaultBondDenom, 10_000_000))
+	denomCreationFeeCoins := tfDefaultGen.Params.DenomCreationFee
+	for _, coin := range denomCreationFeeCoins {
+		if coin.Denom == appparams.BaseCoinUnit {
+			coin.Denom = sdk.DefaultBondDenom
+		}
+	}
 	tfDefaultGenJson := simState.Cdc.MustMarshalJSON(tfDefaultGen)
 	simState.GenState[types.ModuleName] = tfDefaultGenJson
 }
--- a/x/tokenfactory/module.go
+++ b/x/tokenfactory/module.go
@@ -189,7 +189,7 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 // GenerateGenesisState creates a randomized GenState of the tokenfactory module.
 func (am AppModule) SimulatorGenesisState(simState *module.SimulationState, s *simtypes.SimCtx) {
 	tfDefaultGen := types.DefaultGenesis()
-	tfDefaultGen.Params.DenomCreationFee = simulation.TokenFactoryCreationFee
+	tfDefaultGen.Params.DenomCreationFee = sdk.NewCoins(sdk.NewInt64Coin("stake", 10_000_000))
 	tfDefaultGenJson := simState.Cdc.MustMarshalJSON(tfDefaultGen)
 	simState.GenState[types.ModuleName] = tfDefaultGenJson
 }
@@ -199,5 +199,7 @@ func (am AppModule) Actions() []simtypes.Action {
 	return []simtypes.Action{
 		simtypes.NewMsgBasedAction("create token factory token", am.keeper, simulation.RandomMsgCreateDenom),
 		simtypes.NewMsgBasedAction("mint token factory token", am.keeper, simulation.RandomMsgMintDenom),
+		simtypes.NewMsgBasedAction("burn token factory token", am.keeper, simulation.RandomMsgBurnDenom),
+		simtypes.NewMsgBasedAction("change admin token factory token", am.keeper, simulation.RandomMsgChangeAdmin),
 	}
 }
--- a/x/tokenfactory/module.go
+++ b/x/tokenfactory/module.go
@@ -189,12 +189,7 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 // GenerateGenesisState creates a randomized GenState of the tokenfactory module.
 func (am AppModule) SimulatorGenesisState(simState *module.SimulationState, s *simtypes.SimCtx) {
 	tfDefaultGen := types.DefaultGenesis()
-	denomCreationFeeCoins := tfDefaultGen.Params.DenomCreationFee
-	for _, coin := range denomCreationFeeCoins {
-		if coin.Denom == "uosmo" {
-			coin.Denom = sdk.DefaultBondDenom
-		}
-	}
+	tfDefaultGen.Params.DenomCreationFee = sdk.NewCoins(sdk.NewCoin("stake", sdk.NewInt(10000000)))
 	tfDefaultGenJson := simState.Cdc.MustMarshalJSON(tfDefaultGen)
 	simState.GenState[types.ModuleName] = tfDefaultGenJson
 }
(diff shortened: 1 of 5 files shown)


x/tokenfactory/module.go
```diff
@@ -189,7 +189,7 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 // GenerateGenesisState creates a randomized GenState of the tokenfactory module.
 func (am AppModule) SimulatorGenesisState(simState *module.SimulationState, s *simtypes.SimCtx) {
 	tfDefaultGen := types.DefaultGenesis()
-	tfDefaultGen.Params.DenomCreationFee = simulation.TokenFactoryCreationFee
+	tfDefaultGen.Params.DenomCreationFee = sdk.NewCoins(sdk.NewInt64Coin("stake", 10_000_000))
```
p0mvn, line 192: Can we use `sdk.DefaultBondDenom?

Also, how this `10_000_000` chosen? does it make sense to have it in a constant?

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -13,43 +13,98 @@ import (
...
-)
-
 // RandomMsgCreateDenom creates a random tokenfactory denom that is no greater than 44 alphanumeric characters
 func RandomMsgCreateDenom(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (*types.MsgCreateDenom, error) {
-	minCoins := TokenFactoryCreationFee
+	minCoins := sim.TokenFactoryKeeper().GetParams(ctx).DenomCreationFee
 	acc, err := sim.RandomSimAccountWithMinCoins(ctx, minCoins)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("no account has enough coins to pay the denom creation fee")
```
p0mvn, line 21: suggest also wrapping the actual `err`

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -13,43 +13,98 @@ import (
...
 		return nil, errors.New("no addr has created a tokenfactory coin")
 	}
-	// Pick denom
-	store := k.GetCreatorPrefixStore(ctx, acc.Address.String())
-	denoms := osmoutils.GatherAllKeysFromStore(store)
-	denom := simtypes.RandSelect(sim, denoms...)
+
+	denom, addr, err := getTokenFactoryDenomAndItsAdmin(k, sim, ctx, acc)
+	if err != nil {
+		return nil, errors.New("cannot find factory denom or admin")
```
p0mvn, line 39: let's also wrap the actual err. It helps with debugging a lot

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -13,43 +13,98 @@ import (
...
+// RandomMsgBurnDenom takes a random denom that has been created and uses the denom's admin to burn a random amount
+func RandomMsgBurnDenom(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (*types.MsgBurn, error) {
+	acc, senderExists := sim.RandomSimAccountWithConstraint(accountCreatedTokenFactoryDenom(k, ctx))
+	if !senderExists {
+		return nil, errors.New("no addr has created a tokenfactory coin")
+	}
+
+	denom, addr, err := getTokenFactoryDenomAndItsAdmin(k, sim, ctx, acc)
+	if err != nil {
+		return nil, errors.New("cannot find factory denom or admin")
```
p0mvn, line 62: wrap `err` please

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -13,43 +13,98 @@ import (
...
+// RandomMsgChangeAdmin takes a random denom that has been created and changes the admin to another random account
+func RandomMsgChangeAdmin(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (*types.MsgChangeAdmin, error) {
+	acc, senderExists := sim.RandomSimAccountWithConstraint(accountCreatedTokenFactoryDenom(k, ctx))
+	if !senderExists {
+		return nil, errors.New("no addr has created a tokenfactory coin")
+	}
+
+	denom, addr, err := getTokenFactoryDenomAndItsAdmin(k, sim, ctx, acc)
+	if err != nil {
+		return nil, errors.New("cannot find factory denom or admin")
```
p0mvn, line 90: wrap `err` please

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -58,3 +113,14 @@ func accountCreatedTokenFactoryDenom(k keeper.Keeper, ctx sdk.Context) simtypes.
 		return iterator.Valid()
 	}
 }
+
+func getTokenFactoryDenomAndItsAdmin(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context, acc legacysimulationtype.Account) (string, sdk.AccAddress, error) {
+	store := k.GetCreatorPrefixStore(ctx, acc.Address.String())
+	denoms := osmoutils.GatherAllKeysFromStore(store)
+	denom := simtypes.RandSelect(sim, denoms...)
+
+	authData, _ := k.GetAuthorityMetadata(ctx, denom)
```
p0mvn, line 122: Is this an unchecked error or something else?

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -58,3 +113,14 @@ func accountCreatedTokenFactoryDenom(k keeper.Keeper, ctx sdk.Context) simtypes.
...
+	admin := authData.Admin
+	addr, err := sdk.AccAddressFromBech32(admin)
+	return denom, addr, err
```
p0mvn, line 125: This is useful for catching errors early
```suggestion
	addr, err := sdk.AccAddressFromBech32(admin)
	if err != nil {
	     return "", sdk.AccAddress{}, err
	}
	return denom, addr, nil
```

x/tokenfactory/module.go
czarcas7ic: Agreed on using sdk.DefaultBondDenom

As for how that number is chosen, I really only want to change the denom to stake, the actual value can be random for all intents and purposes. With that being said because it is an sdk.Coins struct I would need to specify an index of 0 to only change the denom, and that would assume we arent trying to set the fee as a multi coin struct in future tests. What do you think?

x/tokenfactory/simulation/sim_msgs.go
czarcas7ic: lol I have no idea why I did this for some of these, thank you!

x/tokenfactory/module.go
p0mvn: What if you iterate over all coins in a loop, only changing the denom?

p0mvn: If that doesn't work, let's just leave a comment that the value is arbitrary

czarcas7ic: That does work, implementing now!

x/tokenfactory/simulation/sim_msgs.go
```diff
@@ -13,43 +13,98 @@ import (
...
+	if addr == nil {
+		return nil, errors.New("denom has no admin")
+	}
+
+	denomBal := sim.BankKeeper().GetBalance(ctx, addr, denom)
+	if denomBal.IsZero() {
+		return nil, errors.New("addr does not have enough balance to burn")
+	}
+
+	// TODO: Replace with an improved rand exponential coin
```
p0mvn, line 73: Let's make an issue for this please if not planning to be immediately addressed

p0mvn (approved):

ValarDragon (changes requested):
Nice job on adding more sim updates!

Lets revert the changes to app/modules.go & the simulation app interface.

The simulator should only get explicit dependencies on auth and bank. Any other keeper should be explicitly be passed in to the app module. Also its not even needed for this module.

x/tokenfactory/module.go
```diff
@@ -23,6 +23,8 @@ import (
 	"github.com/spf13/cobra"
 	abci "github.com/tendermint/tendermint/abci/types"
 
+	appparams "github.com/osmosis-labs/osmosis/v10/app/params"
+
```
ValarDragon, line 27: Sorry I didn't notice this in my first pass - we shouldn't have modules depend on the app params. The simulator genesis state should just hardcode a denom / amount for now, rather than replace what the app params base denom is

czarcas7ic: Default genesis defaults to uosmo, so I just wanted to replace uosmo (BaseCoinUnit) with stake (DefaultBondDenom) without hardcoding either. Should I remove this and instead just hardcode "uosmo" in place of appparams.BaseCoinUnit?

czarcas7ic: Ref this right here https://github.com/osmosis-labs/osmosis/blob/88125a37830035ef7c9222ce36f8bfc6aed2aedc/x/tokenfactory/module.go#L195-L199

ValarDragon: Just hardcode the new DenomCreationFee

ValarDragon: Theres not really any point in caring about the default setting. The space that makes sense is:
- Random coin / coins
- hardcoded constant

Generically adapting the default doesn't really matter.

ValarDragon (approved):

ValarDragon:
Nice work!

Issue #2241, opened 2022-07-27
Set up epochs start time in `KeeperTestHelper.App`


catShaark:
## Background

So the epoch time in `KeeperTestHelper.App`  is set to be empty. It should not be empty since in modules' test suite we often set values related to time to be Time.Now(). For example, in incentives' test suite,
we set all the gauges start time to be Time.Now(). This in turn causes unexpected behavior in our tests, like `BeginNewBlock` (with `executeNextEpoch` equals true) won't trigger gauge distribution in incentives' test suite. Specifically, `BeginNewBlock` (with `executeNextEpoch` equals true) will set the suite.Ctx.BlockTime() to be = epoch start time + 5 sec ( = time.Time{} + 5 sec) , because block time (time.Time{} + 5 sec) is way before all the gauges start time (time.Now()) in the tests suite, gauge distribution won't happen

## Suggested Design

- Set epoch time in  `KeeperTestHelper.App`  to be time.Now()

## Acceptance Criteria

- all existing and new tests should pass

catShaark:
@ValarDragon, plz tell me if this is good or not

ValarDragon:
Oh interesting, this is a good point. Yes lets do this!

PR #2242, opened 2022-07-27
set epoch time KeeperTestHelper

catShaark:
Closes: #2241

- set epoch time KeeperTestHelper 

## Brief Changelog

- set epoch time in `KeeperTestHelper` to be time.Now()
- fix `BeginNewBlock` issue with executeNextEpoch = true

## Testing and Verifying

  - Does this pull request introduce a new feature or user-facing behavior changes?  no
  - How is the feature or change documented?   not documented

### DIFF
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -48,7 +48,19 @@ func (s *KeeperTestHelper) Setup() {
 		Ctx:             s.Ctx,
 	}
 
+	s.SetEpochStartTime()
 	s.TestAccs = CreateRandomAccounts(3)
+}
+
+func (s *KeeperTestHelper) SetEpochStartTime() {
+	epochsKeeper := s.App.EpochsKeeper
+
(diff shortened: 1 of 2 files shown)


p0mvn (approved):
LGTM

I've looked into the reasons why this is the case. So it seems that the default genesis `StartTime` is set to the following:
https://github.com/osmosis-labs/osmosis/blob/fe98f6e4453cd32035d277c21ef2f3669b677bb2/x/epochs/types/genesis.go#L61

I don't think that we should be making changes to the default genesis. As a result, I think the approach in this PR is preferable.

I'm going to merge this to unblock @catShaark . Also, thank you for reporting this and finding a solution!

p0mvn:
@catShaark can you undraft this please if you think it's ready and we can merge

catShaark:
> guess

I don't know why yet, I'm investigating it

catShaark:
@p0mvn, I've fixed the test, please check

ValarDragon (approved):

alexanderbez (approved):

PR #2243, opened 2022-07-27
chore(deps): upgrade sdk with min proposer deposit and update e2e

p0mvn:
Closes: #XXX

Upgrading sdk with min proposer deposit and adjusting e2e tests accordingly

## Testing and Verifying

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -285,8 +285,8 @@ require (
 replace (
 	// branch: v0.27.0.rc3-osmo, current tag: v0.27.0.rc3-osmo
 	github.com/CosmWasm/wasmd => github.com/osmosis-labs/wasmd v0.27.0-rc2.0.20220517191021-59051aa18d58
-	// Our cosmos-sdk branch is:  https://github.com/osmosis-labs/cosmos-sdk, current tag: v0.45.0x-osmo-v11-alpha.3.2, current branch: osmosis-main
-	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.45.1-0.20220727205601-c034facbbb23
+	// Our cosmos-sdk branch is:  https://github.com/osmosis-labs/cosmos-sdk, current tag: 0.45.0x-osmo-v11-alpha.4 current branch: osmosis-main
+	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.45.1-0.20220727204202-c61a24d1ef30
 	// Use Osmosis fast iavl
 	github.com/cosmos/iavl => github.com/osmosis-labs/iavl v0.17.3-osmo-v7
 	// use cosmos-compatible protobufs
--- a/go.mod
+++ b/go.mod
@@ -285,8 +285,8 @@ require (
 replace (
 	// branch: v0.27.0.rc3-osmo, current tag: v0.27.0.rc3-osmo
 	github.com/CosmWasm/wasmd => github.com/osmosis-labs/wasmd v0.27.0-rc2.0.20220517191021-59051aa18d58
-	// Our cosmos-sdk branch is:  https://github.com/osmosis-labs/cosmos-sdk, current tag: 0.45.0x-osmo-v11-alpha.4 current branch: osmosis-main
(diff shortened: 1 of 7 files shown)


p0mvn:
@czarcas7ic this is a change similar to what we merged to v11 but using `osmosis-main`. Could you give this a look when you have time please?

go.mod
```diff
@@ -285,8 +285,8 @@ require (
 replace (
 	// branch: v0.27.0.rc3-osmo, current tag: v0.27.0.rc3-osmo
 	github.com/CosmWasm/wasmd => github.com/osmosis-labs/wasmd v0.27.0-rc2.0.20220517191021-59051aa18d58
-	// Our cosmos-sdk branch is:  https://github.com/osmosis-labs/cosmos-sdk, current tag: v0.45.0x-osmo-v11-alpha.2, current branch: osmosis-main
-	github.com/cosmos/cosmos-sdk => github.com/osmosis-labs/cosmos-sdk v0.45.1-0.20220721082708-0aca1df58871
+	// Our cosmos-sdk branch is:  https://github.com/osmosis-labs/cosmos-sdk, current tag: 0.45.0x-osmo-v11-alpha.5 current branch: osmosis-main
```
czarcas7ic, line 288: Should we start tagging sdk branch as v11 proper now that we have our v11 cut? If not, should we be calling this v12-alpha?

tests/e2e/configurer/config/constants.go
```diff
@@ -13,4 +13,8 @@ const (
 	PropBufferBlocks float32 = 5
 	// max retries for json unmarshalling
 	MaxRetries = 60
+	// Minimum deposit value for a proposal to enter a voting period.
+	MinDepositValue = 500000000
```
czarcas7ic, line 17:
```suggestion
	MinDepositValue = 10000000
```

If the min deposit param defaults to 10000000 like before this needs to get reduced, otherwise every time the `InitialMinDeposit` is actually depositing >100% so it wont properly test that the 25% prop deposit requirement is actually working or not

czarcas7ic (approved):

go.mod
p0mvn: This is made against `osmosis-main` so, in my opinion, this shouldn't be named `v11` in general since main tends to be independent of versioning. Next time we update this, I think we should rename it to something like `v0.45.0x-osmo-main.alpha.1`. Wdyt?

tests/e2e/configurer/config/constants.go
p0mvn: I don't understand - `MinDepositValue` is the minimum deposit necessary for the proposal to move to the voting period so this is expected. For determining an acceptable value for the initial deposit, we divide this by 4

p0mvn: https://github.com/osmosis-labs/osmosis/blob/31b524cadb16f0f79ff13ae5020716eff3e5cbd3/tests/e2e/configurer/config/constants.go#L19

p0mvn:
@czarcas7ic comments addressed. Waiting for follow-up replies

tests/e2e/configurer/config/constants.go
czarcas7ic: Right what I am saying is I think the minimum deposit requirement when starting a chain up defaults to 10000000, not 500000000 (what mainnet is)

czarcas7ic: So you are hardcoding 500000000 here but it is still utilizing 10000000 as the min deposit param

czarcas7ic: Yeah I just checked and that is the case

```
czar@Adams-MacBook-Pro osmosis % docker exec -it 017fdf84fbdc bash
root@017fdf84fbdc:~# osmosisd q gov params
deposit_params:
  max_deposit_period: "172800000000000"
  min_deposit:
  - amount: "10000000"
    denom: uosmo
tally_params:
  quorum: "0.334000000000000000"
  threshold: "0.500000000000000000"
  veto_threshold: "0.334000000000000000"
voting_params:
  voting_period: "19000000000"
```

p0mvn: This is now addressed

PR #2244, opened 2022-07-27
simulator: gamm sim updates

czarcas7ic:
Closes: #XXX

Improved upon/adds messages to the gamm simulator

## Brief Changelog

- Adds the following messages:
  - RandomSwapExactAmountIn
  - RandomSwapExactAmountOut
  - RandomJoinSwapExternAmountIn
  - RandomJoinSwapShareAmountOut
  - RandomExitSwapExternAmountOut
  - RandomExitSwapShareAmountIn
- Changes the seed from 10 to 11 (seed 10 has all the validators get jailed...we should look into preventing this)

## Testing and Verifying


ValarDragon:
Uhh, I'm concerned about continuing the approach of adding more methods with API's that are mismatched / way too long

czarcas7ic:
@ValarDragon The only difference would be implementing this API that hasn't been PRd yet right? https://github.com/osmosis-labs/osmosis/pull/2133

ValarDragon:
nah, take for example:
```
func RandomSwapExactAmountIn(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (*types.MsgSwapExactAmountIn, error) {
	// select a pseudo-random pool ID, max bound by the upcoming pool ID
	pool_id := simtypes.RandLTBound(sim, k.GetNextPoolNumber(ctx))
	pool, err := k.GetPoolAndPoke(ctx, pool_id)
	if err != nil {
		return nil, err
	}
	poolCoins := pool.GetTotalPoolLiquidity(ctx)

	// TODO: Improve this, don't just assume two asset pools
	// randomly select one of the pool denoms to be the coinIn and on to be the coinOut
	r := sim.GetSeededRand("select random seed")
	index := r.Intn(len(poolCoins) - 1)
	coinIn := poolCoins[index]
	poolCoins = simtypes.RemoveIndex(poolCoins, index)
	coinOut := poolCoins[0]

	// set the swap route to use this pool
	route := []types.SwapAmountInRoute{}
	route = append(route, types.SwapAmountInRoute{
		PoolId:        pool_id,
		TokenOutDenom: coinOut.Denom,
	})

	// find an address that has a balance of the coinIn
	sender, accCoinIn, senderExists := sim.SelAddrWithDenom(ctx, coinIn.Denom)
	if !senderExists {
		return nil, fmt.Errorf("no sender with denom %s exists", coinIn.Denom)
	}

	// select a random amount that is upper-bound by the address's balance of coinIn
	randomSubset := sim.RandomAmount(accCoinIn.Amount)
	if randomSubset.IsZero() {
		return nil, errors.New("cannot make a swap with zero amount")
	}
	randomCoinSubset := sdk.NewCoins(sdk.NewCoin(accCoinIn.Denom, randomSubset))

	// calculate the minimum number of tokens received from input of tokenIn
	tokenOutMin, err := pool.CalcOutAmtGivenIn(ctx, randomCoinSubset, coinOut.Denom, pool.GetSwapFee(ctx))
	if err != nil {
		return nil, err
	}

	return &types.MsgSwapExactAmountIn{
		Sender:            sender.Address.String(),
		Routes:            route,
		TokenIn:           randomCoinSubset[0],
		TokenOutMinAmount: tokenOutMin.Amount,
	}, nil
```

This is 40 lines long, with lots of reuse, but why? It could instead be:
```
func RandomSwapExactAmountIn(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (*types.MsgSwapExactAmountIn, error) {
  pool, err := getRandPool(k, sim)
  if err != nil { return nil, err }
  route := getRandRoute(k, sim, pool)
  sender, accCoinIn, senderExists := sim.SelAddrWithDenom(ctx, route.TokenInDenom)
  if !senderExists { 
     return errors.New(...) 
  }
  TokenInAmt := sim.RandAmtLT(accCoinIn)
  tokenOutMin, err := pool.CalcOutAmtGivenIn(ctx, randomCoinSubset, coinOut.Denom, pool.GetSwapFee(ctx))
  if err != nil {
	return nil, err
  }
return &types.MsgSwapExactAmountIn{
		Sender:            sender.Address.String(),
		Routes:            route,
		TokenIn:           randomCoinSubset[0],
		TokenOutMinAmount: tokenOutMin.Amount,
	}, nil
}

File: x/gamm/simulation/sim_msgs.go [hunk 1]
```diff
@@ -472,3 +412,24 @@ func createPoolRestriction(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Contex
...
+	}
+	poolCoins := pool.GetTotalPoolLiquidity(ctx)
+
+	// TODO: Improve this, don't just assume two asset pools
+	// randomly select one of the pool denoms to be the coinIn and one to be the coinOut
+	r := sim.GetSeededRand("select random seed")
+	index := r.Intn(len(poolCoins) - 1)
+	coinIn := poolCoins[index]
+	poolCoins = simtypes.RemoveIndex(poolCoins, index)
+	coinOut := poolCoins[0]
```
ValarDragon, line 429: Why not use route directly here?

czarcas7ic:
Just attempted to reduce it a bit with the added `getRandPool` function as that where most of the overlap was

x/gamm/simulation/sim_msgs.go
czarcas7ic: The route various based on the type right? SwapAmountInRoute vs SwapAmountOutRoute, they are different types so couldn't use the same output AFAIU

ValarDragon: ah right :/

czarcas7ic: And because there is only two occurrences of this, creating two separate functions for each type wouldn't make sense right?

czarcas7ic: I'm just making sure I am not missing something! 😄

x/gamm/simulation/sim_msgs.go
```diff
@@ -140,18 +106,302 @@ func RandomCreateUniV2Msg(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context
...
+	sender, accCoinIn, senderExists := sim.SelAddrWithDenom(ctx, coinIn.Denom)
+	if !senderExists {
+		return nil, fmt.Errorf("no sender with denom %s exists", coinIn.Denom)
+	}
+
+	// select a random amount that is upper-bound by the address's balance of coinIn
+	randomCoinSubset := sdk.NewCoins(sdk.NewCoin(accCoinIn.Denom, sim.RandomAmount(accCoinIn.Amount)))
+	if randomCoinSubset.IsZero() {
+		return nil, errors.New("cannot make a swap with zero amount")
+	}
```
ValarDragon, line 147: Can we make this condition impossible to occur? (By using a different amount selection function?)

ValarDragon (approved):
I think we should do a second pass, to make a lot of this more succinct

still seems wrong to me that this simulate file is 400 lines long

Issue #2194, opened 2022-07-22
x/superfluid: ensure events are emitted and tests added


p0mvn:
## Background

There are a few issues with how we emit superfluid events. 

First, some of them are not being emitted on success:
https://github.com/osmosis-labs/osmosis/blob/979cbd449274c660d8ed31b095664b8625b506bf/x/superfluid/keeper/msg_server.go#L49-L55

The conditional should be the opposite of what it is now.

Second, there are no tests for any superfluid events so we cannot be confident if the rest function as expected.

Third, there is no documentation about what events are for and which attributes they should contain.

## Suggested Design

Follow a similar change in `x/gamm`: https://github.com/osmosis-labs/osmosis/issues/1942

To limit the scope and ease the review, I suggest associating each PR with the following tasks:

- [ ] Task 1 - Core logic + Test `TypeEvtSetSuperfluidAsset `
1. Create an internal `x/superfluid/internal/events` package that contains core logic for emitting events
2. Keep the event names in `types` since they might be used by integrators
3. Follow [this PR](https://github.com/osmosis-labs/osmosis/pull/1983) for an example in GAMM
4. Test [`TypeEvtSetSuperfluidAsset`](https://github.com/osmosis-labs/osmosis/blob/cf72a781100c5a413b72f35d61a0a5ca25c87176/x/superfluid/types/events.go#L5)
    * test attributes in `x/superfluid/internal/events`
    * test that the event is emitted in the right location
    * Add docs to the README about what the event is for and which attributes it contains

- [ ] Task 2 - Test `TypeEvtRemoveSuperfluidAsset`
Test [`TypeEvtRemoveSuperfluidAsset`](https://github.com/osmosis-labs/osmosis/blob/cf72a781100c5a413b72f35d61a0a5ca25c87176/x/superfluid/types/events.go#L6)
    * test attributes in `x/superfluid/internal/events`
    * test that the event is emitted in the right location
    * Add docs to the README about what the event is for and which attributes it contains
- [ ] Task 3 - Test `TypeEvtSuperfluidDelegate`
Test [`TypeEvtSuperfluidDelegate`](https://github.com/osmosis-labs/osmosis/blob/cf72a781100c5a413b72f35d61a0a5ca25c87176/x/superfluid/types/events.go#L7)
    * test attributes in `x/superfluid/internal/events`
    * test that the event is emitted in the right location
    * Add docs to the README about what the event is for and which attributes it contains
[...]

PR #2186, opened 2022-07-21
test(x/gamm): remove liquidity events

p0mvn:
Part of: https://github.com/osmosis-labs/osmosis/issues/1942

## What is the purpose of the change

Tests remove liquidity events that should be emitted when a pool is exited.

Similarly to #2141, I noticed that they were emitted twice:
1. at the end of the message server
2. right after a pool state is updated.

I removed 1. while keeping 2. 

No breaking changes.

## Passthrough changes

- moved `AssertEventEmitted` test helper to be defined on the suite. It is also useful for testing events in other modules
- cleaned up some tests by creating variables for keepers and `ctx`
- added `suite.Run()` to some table-driven tests

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

PR #2255, opened 2022-07-29
x/superfluid: ensure events are emitted and tests added

hieuvubk:
Closes: #2194

## Brief Changelog
 
  - *Refactor and unit test superfluid module events*
  - *Write test to make sure events emitted in right place*
  - *Update readme*

## Testing and Verifying

  - *Internal events unit tests*
  - *msg server tests*


x/superfluid/README.md
```diff
@@ -455,7 +455,88 @@ Disable multiple assets from being used for superfluid staking.
...
+This event is emitted in the proposal which remove superfluid asset
+
+It consists of the following attributes:
+
+* `types.AttributeDenom`
+  * The value is the asset denom.
+
+### `types.TypeEvtSuperfluidDelegate`
+
+This event is emitted in the message server after succesfully create a delegation for the given lock ID and the validator to delegate to.
```
p0mvn, line 490:
```suggestion
This event is emitted in the message server after successfully creating a delegation for the given lock ID and the validator to delegate to.
```

x/superfluid/README.md
```diff
@@ -455,7 +455,88 @@ Disable multiple assets from being used for superfluid staking.
...
+It consists of the following attributes:
+
+* `types.AttributeLockId`
+  * The value is the given lock ID.
+* `types.AttributeAmount`
+  * The value is the token amount added to the lock.
+
+### `types.TypeEvtSuperfluidUndelegate`
+
+This event is emitted in the message server after undelegating currently superfluid delegated position given by lock ID.
```
p0mvn, line 512:
```suggestion
This event is emitted in the message server after undelegating the currently superfluid delegated position given by lock ID.
```

x/superfluid/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func newRemoveSuperfluidAssetEvent(denom string) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtRemoveSuperfluidAsset,
+		sdk.NewAttribute(types.AttributeDenom, denom),
+	)
+}
+
+func EmitSuperfluidDelegateEvent(ctx sdk.Context, lockId uint64, valAddress string) {
```
p0mvn, line 46: Making this comment to gather more thoughts. This does not need to be addressed.

Seems to be a common pattern. Maybe we can extract a util function that utilizes Go generics where the message being emitted is the generic type.

p0mvn (approved):
LGTM.

I verified that all attributes are preserved and the locations of where events are emitted are not changed.

Phenomenal work on this @hieuvubk ! and thank you for documenting it all!

x/superfluid/keeper/gov/suite_test.go
```diff
@@ -30,3 +30,15 @@ func (suite *KeeperTestSuite) SetupTest() {
 func TestKeeperTestSuite(t *testing.T) {
 	suite.Run(t, new(KeeperTestSuite))
 }
+
+func assertEventEmitted(suite *KeeperTestSuite, ctx sdk.Context, eventTypeExpected string, numEventsExpected int) {
```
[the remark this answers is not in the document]
hieuvubk: should I use this function from gamm module instead create a new function for superfluid as I did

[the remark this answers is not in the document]
p0mvn: Eventually, we should make sure that we aren't duplicating code and using a shared function from #2186

However, for now, it is fine to have this function in superfluid. Once #2186 is merged, we can refactor

x/superfluid/keeper/internal/events/emit.go
p0mvn: I think we can decide on this later, so going to resolve

p0mvn:
I would like to have one more reviewer give this a look before merging since the diff is large. Overall, the change LGTM. Also, I left some a comment for follow-up work that would be good to decide on

x/superfluid/keeper/gov/gov_test.go
```diff
@@ -59,6 +59,7 @@ func (suite *KeeperTestSuite) TestHandleSetSuperfluidAssetsProposal() {
 		assets         []types.SuperfluidAsset
 		expectedAssets []types.SuperfluidAsset
 		expectErr      bool
+		expectedEvent  string
 	}
 	testCases := []struct {
```
czarcas7ic, line 64: Are we able to / should we break down this actions struct by item? Maybe its not a big deal but just seeing a lump of five items together doesn't make the table driven test super easy to figure out what is getting tested.

x/superfluid/keeper/internal/events/emit.go
czarcas7ic: Agreed on this, but also not blocking as you said

x/superfluid/keeper/internal/events/emit_test.go
```diff
@@ -0,0 +1,339 @@
...
+		assetType types.SuperfluidAssetType
+	}{
+		"basic valid": {
+			ctx:       suite.CreateTestContext(),
+			denom:     testDenomA,
+			assetType: types.SuperfluidAssetTypeNative,
+		},
+		"context with no event manager": {
+			ctx: sdk.Context{},
+		},
```
czarcas7ic, line 45: @p0mvn is there no need to test an invalid case, such as an invalid denom?

x/superfluid/keeper/msg_server_test.go
```diff
@@ -192,3 +193,191 @@ func (suite *KeeperTestSuite) TestMsgLockAndSuperfluidDelegate() {
...
+		{
+			name: "basic valid",
+			param: param{
+				lockOwner: sdk.AccAddress([]byte("addr1---------------")), // setup wallet
+				duration:  time.Hour * 504,
+			},
+			expectPass: true,
+		},
+		{
+			name: "invalid",
```
czarcas7ic, line 219: can we be more specific just for future debugging, something like "invalid due to duration below minimum" or something along those lines?

x/superfluid/keeper/internal/events/emit_test.go
p0mvn: I think this function already assumes that a denom is pre-validated somewhere else. As far as it is concerned, any received denom string is valid. So I think it should be fine without extra tests unless you would like to suggest anything specific that I'm not seeing.

x/superfluid/keeper/gov/gov_test.go
p0mvn: Good idea, naming fields is preferable for readability IMO

x/superfluid/keeper/internal/events/emit_test.go
czarcas7ic: No that all sounds good to me, was just checking!

x/superfluid/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,133 @@
...
+		newSetSuperfluidAssetEvent(denom, assetType),
+	})
+}
+
+func newSetSuperfluidAssetEvent(denom string, assetType types.SuperfluidAssetType) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtSetSuperfluidAsset,
+		sdk.NewAttribute(types.AttributeDenom, denom),
+		sdk.NewAttribute(types.AttributeSuperfluidAssetType, assetType.String()),
+	)
```
stackman27, line 26: Since we are going over all the module events, what are the thoughts on using `EmitTypedEvent` instead of `EmitEvents`

It'll definitely reduce the code size and gives a better clarity. For instance the above code could be rewritten as with an event.proto file; 
```
ctx.EventManager().EmitTypedEvent(&superfluid.AssetEvent{
		denom:  denom,
	         assetType: assetType.String(), 
	})
```

ps: `EmitEvents` is labelled deprecated too

cc @alexanderbez @ValarDragon @p0mvn

p0mvn: Yes, I think that's a great idea. We would have to make the event constructors exported.

For example, `newSetSuperfluidAssetEvent` -> `NewSetSuperfluidAssetEvent`

x/superfluid/keeper/msg_server.go
```diff
@@ -92,11 +85,8 @@ func (server msgServer) SuperfluidUnbondLock(goCtx context.Context, msg *types.M
...
 	err := server.keeper.SuperfluidUnbondLock(ctx, msg.LockId, msg.Sender)
-	if err != nil {
-		ctx.EventManager().EmitEvent(sdk.NewEvent(
-			types.TypeEvtSuperfluidUnbondLock,
-			sdk.NewAttribute(types.AttributeLockId, fmt.Sprintf("%d", msg.LockId)),
-		))
+	if err == nil {
+		events.EmitSuperfluidUnbondLockEvent(ctx, msg.LockId)
 	}
 	return &types.MsgSuperfluidUnbondLockResponse{}, err
```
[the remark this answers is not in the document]
alexanderbez: Yeah we should use typed events for each message execution. We should also consider populating `MsgSuperfluidUnbondLockResponse` with relevant fields.

x/superfluid/keeper/internal/events/emit.go
alexanderbez: Are we reusing events? I would only use constructors and separate out the event logic if we use the same ones in multiple places, otherwise it would be overkill.

x/superfluid/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,133 @@
...
+
+func newSuperfluidUndelegateEvent(lockId uint64) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtSuperfluidUndelegate,
+		sdk.NewAttribute(types.AttributeLockId, fmt.Sprintf("%d", lockId)),
+	)
+}
+
+func EmitSuperfluidUnbondLockEvent(ctx sdk.Context, lockId uint64) {
+	if ctx.EventManager() == nil {
```
alexanderbez, line 100: `EventManager` should never be nil, but I guess it can't hurt. Just so many LOC :-/

x/superfluid/README.md
```diff
@@ -455,7 +455,88 @@ Disable multiple assets from being used for superfluid staking.
 
 ## Events
```
alexanderbez, line 456: I love the motivation for this, but I can see this going out of date and being difficult to maintain.

x/superfluid/keeper/internal/events/emit.go
hieuvubk: Ahh thank you. If so, I'll remove them.

x/superfluid/keeper/internal/events/emit.go
ValarDragon: We are in 0 rush to switch to typed events. 

I also agree that constructors seem overkill

x/superfluid/keeper/msg_server_test.go
```diff
@@ -192,3 +193,191 @@ func (suite *KeeperTestSuite) TestMsgLockAndSuperfluidDelegate() {
 		}
 	}
 }
+
+// TestMsgSuperfluidDelegate_Event tests that events are correctly emitted
+// when calling SuperfluidDelegate.
+func (suite *KeeperTestSuite) TestMsgSuperfluidDelegate_Event() {
```
stackman27, line 199: I noticed that events tests are very identical to their actual tests and there is repetitive code like (`TestMsgSuperfluidDelegate` and `TestMsgSuperfluidDelegate_Event`) 

is there any way we can just use one?

x/superfluid/keeper/msg_server_test.go
```diff
@@ -192,3 +193,191 @@ func (suite *KeeperTestSuite) TestMsgLockAndSuperfluidDelegate() {
...
+	for index, poolId := range poolIds {
+		sender, _ := sdk.AccAddressFromBech32(locks[index].Owner)
+		suite.Ctx = suite.Ctx.WithBlockHeight(v8constants.UpgradeHeight)
+		_, err := msgServer.UnPoolWhitelistedPool(sdk.WrapSDKContext(suite.Ctx), types.NewMsgUnPoolWhitelistedPool(sender, poolId))
+		suite.Require().NoError(err)
+		assertEventEmitted(suite, suite.Ctx, types.TypeEvtUnpoolId, 1)
+	}
+}
+
+func assertEventEmitted(suite *KeeperTestSuite, ctx sdk.Context, eventTypeExpected string, numEventsExpected int) {
```
stackman27, line 373: this function has been defined in `app/apptesting/events.go` pretty sure we can just use that

x/superfluid/keeper/msg_server_test.go
```diff
@@ -192,3 +193,191 @@ func (suite *KeeperTestSuite) TestMsgLockAndSuperfluidDelegate() {
...
+		resp, err := lockupMsgServer.LockTokens(c, lockuptypes.NewMsgLockTokens(test.param.lockOwner, test.param.duration, coinsToLock))
+
+		valAddrs := suite.SetupValidators([]stakingtypes.BondStatus{stakingtypes.Bonded})
+
+		msgServer := keeper.NewMsgServerImpl(suite.App.SuperfluidKeeper)
+		_, err = msgServer.SuperfluidDelegate(c, types.NewMsgSuperfluidDelegate(test.param.lockOwner, resp.ID, valAddrs[0]))
+
+		if test.expectPass {
+			suite.Require().NoError(err)
+			assertEventEmitted(suite, suite.Ctx, types.TypeEvtSuperfluidDelegate, 1)
```
stackman27, line 246: why do you need to pass `Suite` and `Suite.Ctx`? You can just get the context if you pass `Suite`

x/superfluid/keeper/msg_server_test.go
hieuvubk: Thanks for reminding

x/superfluid/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,133 @@
...
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		newSuperfluidDelegateEvent(lockId, valAddress),
+	})
+}
+
+func newSuperfluidDelegateEvent(lockId uint64, valAddress string) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtSuperfluidDelegate,
+		sdk.NewAttribute(types.AttributeLockId, fmt.Sprintf("%d", lockId)),
```
stackman27, line 59:
```suggestion
		sdk.NewAttribute(types.AttributeLockId, utils.Uint64ToString(lockId)),
```

lets change this for tests too

x/superfluid/keeper/internal/events/emit_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/osmosis-labs/osmosis/v10/app/apptesting"
+	"github.com/osmosis-labs/osmosis/v10/x/superfluid/keeper/internal/events"
+	"github.com/osmosis-labs/osmosis/v10/x/superfluid/types"
+)
+
+type SuperfluidEventsTestSuite struct {
```
stackman27, line 16: Just wondering, what was the motivation behind create a new test suite? Can't we just use the `KeeperTestSuite`?

cc @p0mvn

x/superfluid/keeper/internal/events/emit_test.go
```diff
@@ -0,0 +1,339 @@
...
+	apptesting.KeeperTestHelper
+}
+
+const (
+	addressString = "addr1---------------"
+	testDenomA    = "denoma"
+	testDenomB    = "denomb"
+	testDenomC    = "denomc"
```
stackman27, line 24: this seems to be unused

x/superfluid/keeper/internal/events/emit_test.go
```diff
@@ -0,0 +1,339 @@
...
+	testDenomD    = "denomd"
```
stackman27, line 25: this seems to be unused

hieuvubk: thank you

x/superfluid/README.md
p0mvn: I think we might still want to have it since these docs get auto-uploaded to the docs website. Currently, there is no info about events but it could be useful for integrators. WDYT?

ValarDragon (approved):
LGTM

I do have the same general point of I think this should be managed by the framework / we shouldn't do more of this, but I think this is good to merge in!

Thanks for the fix, and sorry this got blocked for so long

PR #2257, opened 2022-07-29
Hackatom seoul

Thunnini:
Change the stargate querier for cosmwasm to use json


mattverse (approved):
Reviewed Off line together, this PR marshals responses into JSON Format so that the contract doesnt have to handle proto encoded data

ValarDragon:
@mattverse please don't merge things like this without more review -- this looks quite senstiive, and definitely underdocumented.

PR #2264, opened 2022-07-31
Try to speedup dockerfile build time

ValarDragon:

* Reorder cosmwasm steps to come earlier in dockerfile
* Run go mod download to cache the go dependency downloads. (Shaves ~30 seconds as far as I can tell)
* Use RUN --mount=type=cache to speedup repeat docker builds locally. (See below for why this doesn't work in CI, I tried a few times, at some point github cache stopped working, so I quit trying)
* Rename github action steps to be more concise
* Combine a few of the prior layers

## Brief Changelog

Add go cache to dockerfile

## Testing and Verifying

This change is already covered by E2E tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Tests & Code Coverage
 
 on:
   pull_request:
@@ -59,9 +59,9 @@ jobs:
             ~/.cache/go-build
             ~/Library/Caches/go-build
             ~\AppData\Local\go-build
-          key: ${{ runner.os }}-${{ hashFiles('**/go.sum') }}
+          key: ${{ runner.os }}-go-${{ matrix.go-version }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
-            ${{ runner.os }}-
+            ${{ runner.os }}-go-${{ matrix.go-version }}-
       -
         name: Run all tests
         run: make test-unit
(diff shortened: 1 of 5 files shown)


ValarDragon:
This is sad :(

You need to use `RUN --mount=type=cache` to get golang cache used in docker build steps. This works locally. But its been an open issue for years, to get it working in CI's: 
* https://github.com/moby/buildkit/issues/1512
* https://github.com/docker/setup-buildx-action/issues/122
* https://github.com/docker/buildx/issues/156
* https://github.com/moby/buildkit/issues/1474
* https://stackoverflow.com/questions/60448114/how-do-i-reuse-the-cache-from-a-run-mount-type-cache-docker-build/60485339?noredirect=1#comment120062963_60485339


Two repo workarounds:
* https://github.com/Mahoney-playground/docker-cache-action (suffers the entire size of copying `var/lib/.docker`)
* https://github.com/dashevo/gh-action-cache-buildkit-state (idk if it has issues)


The first repo copies the entire internal dockerkit local directory via CI into github action cache, but this is overly expensive. (Your adding a minute due to its size of excess data per github issue reports. Still would net save us 1.5 minute tho) This is what the first repo does.

Ideally we could just copy the relevant part of the internal buildkit cache state, unsure if theres documentation on where this all goes. (Maybe thats what the second repo does)

ValarDragon:
Wait maybe we can just copy out the relevant buildkit cache:
```
$ ls /var/lib/docker/buildkit
cache.db  containerdmeta.db  content  executor	metadata_v2.db	net  snapshots.db
```

niccoloraspa (approved):
LGTM. 🚀

alexanderbez (approved):
LGTM!

PR #2265, opened 2022-08-01
Add codec init to tokenfactory

mattverse:
Closes: #XXX

Currently we cannot use codec for token factory module as it misses initialization for amino codec registration. 
This PR adds codec init to the tokenfactory module

## Brief Changelog

- Add codec initialization to token factory module.

### DIFF
--- a/x/tokenfactory/types/codec.go
+++ b/x/tokenfactory/types/codec.go
@@ -29,4 +29,12 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
 }
 
-var ModuleCdc = codec.NewProtoCodec(cdctypes.NewInterfaceRegistry())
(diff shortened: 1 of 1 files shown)


ValarDragon:
I don;t understand, whats an example of something thats failing before, that this fixes?

mattverse:
So if we don't have these initialized, get sign bytes would not return the registered codec(missing path), it would return incompatible(or at least not what we normally return(https://github.com/osmosis-labs/osmosis/blob/main/x/lockup/types/codec.go#L31-L34)  codec values when calling
`GetSignBytes() []byte {
	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&msg))`.

alexanderbez (approved):
Given this follows other module's codec initialization/usage -- ACK

ValarDragon (approved):
Gotcha, I see. I suspect that theres something wrong here with the, but can investigate in a followup

PR #2268, opened 2022-08-01
Fix v10.x state incompatability

ValarDragon:

There was a State machine incompatability in v10.x, due to the ordering of gas measuring methods. cref:
* https://github.com/cosmos/cosmos-sdk/issues/12788
* https://twitter.com/valardragon/status/1554117537230618624?s=20&t=tixHGobU2WBZ6zlU8ov6Gw

These two PRs caused several gas-behavior changing logics, it just happens that the subtle case is what we hit on mainnet:
* https://github.com/osmosis-labs/osmosis/pull/2030
* https://github.com/osmosis-labs/osmosis/pull/2029

They changed the order of operations in the normal case (which is what caused the out of gas bug. They also cause wholly different behavior on error situations.

## Brief Changelog

- Revert state machine incompatible changes

## Testing and Verifying

I reviewed all the commits in https://github.com/osmosis-labs/osmosis/tree/v10.x to look for any other state machine incompatabilities due to ordering. I didn't see any. We changed ordering of things in mint, but that is begin block logic, which doesn't have gas measurement concerns.

I've tested that this syncs through the offending block in v10.1.x that caused app hash mismatches.

  - Does this pull request introduce a new feature or user-facing behavior changes? yes - v10.x will work

### DIFF
--- a/x/lockup/keeper/lock.go
+++ b/x/lockup/keeper/lock.go
@@ -353,16 +353,7 @@ func (k Keeper) unlockMaturedLockInternalLogic(ctx sdk.Context, lock types.Perio
 // 2. Locks that are unlokcing are not allowed to change duration.
 // 3. Locks that have synthetic lockup are not allowed to change.
 // 4. Provided duration should be greater than the original duration.
-func (k Keeper) ExtendLockup(ctx sdk.Context, lockID uint64, owner sdk.AccAddress, newDuration time.Duration) error {
-	lock, err := k.GetLockByID(ctx, lockID)
-	if err != nil {
-		return err
-	}
-
-	if lock.GetOwner() != owner.String() {
-		return types.ErrNotLockOwner
-	}
-
+func (k Keeper) ExtendLockup(ctx sdk.Context, lock types.PeriodLock, newDuration time.Duration) error {
 	if lock.IsUnlocking() {
 		return fmt.Errorf("cannot edit unlocking lockup for lock %d", lock.ID)
 	}
@@ -372,15 +363,10 @@ func (k Keeper) ExtendLockup(ctx sdk.Context, lockID uint64, owner sdk.AccAddres
 		return fmt.Errorf("cannot edit lockup with synthetic lock %d", lock.ID)
 	}
 
-	// completely delete existing lock refs
-	err = k.deleteLockRefs(ctx, unlockingPrefix(lock.IsUnlocking()), *lock)
-	if err != nil {
-		return err
-	}
+	oldLock := lock
 
-	oldDuration := lock.GetDuration()
 	if newDuration != 0 {
-		if newDuration <= oldDuration {
+		if newDuration <= lock.Duration {
 			return fmt.Errorf("new duration should be greater than the original")
 		}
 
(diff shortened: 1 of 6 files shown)


p0mvn (approved):

mattverse (approved):
LGTM, confirmed that this PR reverts the aforementioned Prs

x/lockup/keeper/lock.go
```diff
@@ -66,79 +66,100 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
+		return nil, err
 	}
 
 	if k.hooks == nil {
 		return lock, nil
 	}
 
-	k.hooks.AfterAddTokensToLock(ctx, lock.OwnerAddress(), lock.GetID(), sdk.NewCoins(tokensToAdd))
-
+	k.hooks.OnTokenLocked(ctx, lock.OwnerAddress(), lock.ID, sdk.Coins{coin}, lock.Duration, lock.EndTime)
```
stackman27, line 92: Why is `OnTokenLocked` hook called when `AfterAddTokensToLock` is already called above? Are both the hook calls necessary?

x/lockup/keeper/lock_test.go
```diff
@@ -144,17 +144,47 @@ func (suite *KeeperTestSuite) TestUnlockPeriodLockByID() {
...
+	err = suite.App.LockupKeeper.Lock(suite.Ctx, lock)
+	suite.Require().Error(err)
+
+	// lock with balance with different id
+	lock = types.NewPeriodLock(2, addr1, time.Second, suite.Ctx.BlockTime().Add(time.Second), coins)
+	suite.FundAcc(addr1, coins)
+	err = suite.App.LockupKeeper.Lock(suite.Ctx, lock)
+	suite.Require().NoError(err)
+}
+
```
stackman27, line 175: should we start writing test that checks gas at every intermediate steps of lock functions? Is that even possible?

x/lockup/keeper/lock.go
```diff
@@ -66,79 +66,100 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
 
 	// store lock object into the store
-	err = k.setLock(ctx, lock)
+	store := ctx.KVStore(k.storeKey)
+	bz, err := proto.Marshal(&lock)
+	if err != nil {
+		return err
+	}
+	store.Set(lockStoreKey(lock.ID), bz)
+
```
stackman27, line 154: doesnot line 148-153 do the exact thing as line 135 which was removed?

stackman27, line 154: doesnot line 148-153 do the exact thing as line 135 which was removed?

x/lockup/keeper/lock_test.go
mattverse: I really like this idea, want to use this for e2e testings or state compatibility testing :)

x/lockup/keeper/lock.go
mattverse: we're basically changing(reverting) the order pre-merge of the two refactor PRs

x/lockup/keeper/lock_test.go
ValarDragon: No this seems excessive. Let's just fix the reason that this was an issue

x/lockup/keeper/lock.go
```diff
@@ -66,79 +66,100 @@ func (k Keeper) AddToExistingLock(ctx sdk.Context, owner sdk.AccAddress, coin sd
...
+	err := k.setLock(ctx, *lock)
+	if err != nil {
+		return err
+	}
+
+	// modifications to accumulation store
+	k.accumulationStore(ctx, coin.Denom).Increase(accumulationKey(lock.Duration), coin.Amount)
+
+	// CONTRACT: lock will have synthetic lock only if it has a single coin
```
alexanderbez, line 108:
```suggestion
	// CONTRACT: Lock will have synthetic lock only if it has a single coin
```

alexanderbez (approved):
🍍

x/lockup/keeper/lock_test.go
ValarDragon: Depends on if we have agreement for this approach: https://github.com/cosmos/cosmos-sdk/issues/12788

ValarDragon: agreeing on that & finishing tests in https://github.com/osmosis-labs/cosmos-sdk/pull/304 << writing tooling for this lol.

PR #2269, opened 2022-08-02
update documentation for admin role and burning

Jelena647:

Update documentation on role of admin: Since the admin can change, and therefore is not always the creator of the denom, the Burning, Minting, ChangeAdmin documentation should be updated to account for this possibility. 

Incorrect documentation for Burning: Section specified instructions for Minting, not Burning.

### DIFF
--- a/x/tokenfactory/README.md
+++ b/x/tokenfactory/README.md
@@ -44,8 +44,8 @@ message MsgCreateDenom {
 
 ### Mint
 
-Minting of a specific denom is only allowed for the creator of the denom
-registered during `CreateDenom`, or the current admin.
+Minting of a specific denom is only allowed for the current admin.
+Note, the current admin is defaulted to the creator of the denom.
 
 ```go
 message MsgMint {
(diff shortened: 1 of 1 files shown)


x/tokenfactory/README.md
```diff
@@ -45,7 +45,7 @@ message MsgCreateDenom {
 ### Mint
 
 Minting of a specific denom is only allowed for the creator of the denom
-registered during `CreateDenom`.
+registered during `CreateDenom`, or the current admin.
```
ValarDragon, line 48: These updates should all probably be:
```
Minting of a specific denom is only allowed for the current admin (default creator of the denom)
```

(Since its not the case that its denom creator AND current admin)

ValarDragon:
Great catch!

ValarDragon:
bump on updating those comments

x/tokenfactory/README.md
```diff
@@ -66,8 +66,9 @@ message MsgMint {
 
 ### Burn
 
-Burning of a specific denom is only allowed for the creator of the denom
-registered during `CreateDenom`.
+
```
ValarDragon, line 69:
```suggestion
```

ValarDragon (approved):
Thank you for the PR and for catching this!

Issue #2236, opened 2022-07-26
x/incentives: refactor create gauge and add to gauge fees to use txfees denom


p0mvn:
## Background

In #2227, we hard coded the fee denom for create gauge and add to gauge fees. This was done to minimize the changes for a smooth and expedited v11 release.

However, the hard-coded denom causes issues with the simulator. Simulator assumes "stake" denom for majority of its tests by default. As a result, no account has the hard coded "uosmo" denom to pay the fee.

We had to disable simulator in the [incentives tests](https://github.com/osmosis-labs/osmosis/pull/2227/commits/87dd1fe8423b4b680826375138ddfd4d2337489d) for now.

## Suggested Design

We should revert the following 3 commits separately and avoid backporting them to `v11.x`:
- https://github.com/osmosis-labs/osmosis/pull/2227/commits/100b693c51cb28e2da6e23954f7960b881485546
- https://github.com/osmosis-labs/osmosis/pull/2227/commits/915e9ff99a4a615ee4a5de2c74a06bf6368dc94a
- https://github.com/osmosis-labs/osmosis/pull/2227/commits/87dd1fe8423b4b680826375138ddfd4d2337489d

## Acceptance Criteria
- create gauge and add to gauge fee uses denom from the `txfees` keeper
- simulator incentives tests function correctly

stackman27:
Hi @p0mvn is this being worked on?

p0mvn:
Hi @stackman27 . No, please feel free to take this if you'd like

PR #2283, opened 2022-08-02
x/incentives: refactor create gauge and add to gauge fees to use txfees denom

stackman27:
Closes: #2236

In  [`#2227`](https://github.com/osmosis-labs/osmosis/pull/2227) we hard coded the fee denom for create gauge and add to gauge fees. This was done to minimize the changes for a smooth and expedited v11 release.

However, the hard-coded denom causes issues with the simulator. Simulator assumes "stake" denom for majority of its tests by default. As a result, no account has the hard coded "uosmo" denom to pay the fee. The simulator doesn't run because no account has "uosmo", only "stake"

## Brief Changelog
We should avoid backporting these changes to v11.x

## Testing and Verifying
Use dynamic denom calculation instead of hardcoding it in these tests ;
1. TestCreateGauge_Fee
2. TestAddToGauge_Fee
3. TestChargeFeeIfSufficientFeeDenomBalance

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -307,7 +307,6 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 		appKeepers.EpochsKeeper,
 		appKeepers.keys[txfeestypes.StoreKey],
 		appKeepers.GAMMKeeper,
-		appKeepers.GAMMKeeper,
 		txfeestypes.FeeCollectorName,
 		txfeestypes.NonNativeFeeCollectorName,
 	)
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -260,6 +260,19 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 
 	appKeepers.EpochsKeeper = epochskeeper.NewKeeper(appCodec, appKeepers.keys[epochstypes.StoreKey])
 
+	txFeesKeeper := txfeeskeeper.NewKeeper(
+		appCodec,
+		appKeepers.AccountKeeper,
+		appKeepers.BankKeeper,
+		appKeepers.EpochsKeeper,
+		appKeepers.keys[txfeestypes.StoreKey],
+		appKeepers.GAMMKeeper,
+		appKeepers.GAMMKeeper,
+		txfeestypes.FeeCollectorName,
+		txfeestypes.NonNativeFeeCollectorName,
+	)
+	appKeepers.TxFeesKeeper = &txFeesKeeper
+
 	appKeepers.IncentivesKeeper = incentiveskeeper.NewKeeper(
 		appCodec,
 		appKeepers.keys[incentivestypes.StoreKey],
@@ -268,6 +281,7 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 		appKeepers.LockupKeeper,
 		appKeepers.EpochsKeeper,
 		appKeepers.DistrKeeper,
+		appKeepers.TxFeesKeeper,
 	)
 
 	appKeepers.SuperfluidKeeper = superfluidkeeper.NewKeeper(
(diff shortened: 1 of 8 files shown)


x/incentives/simulation/operations.go
```diff
@@ -117,7 +115,7 @@ func SimulateMsgCreateGauge(ak stakingTypes.AccountKeeper, bk stakingTypes.BankK
 	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
 		simAccount, _ := simtypes.RandomAcc(r, accs)
 		simCoins := bk.SpendableCoins(ctx, simAccount.Address)
-		if simCoins.Len() <= 0 || simCoins.AmountOf(appparams.BaseCoinUnit).LT(types.CreateGaugeFee) {
+		if simCoins.Len() <= 0 || simCoins.AmountOf(simCoins[0].Denom).LT(types.CreateGaugeFee) {
```
p0mvn, line 118: nit: suggest `sdk.DefaultBondDenom` since we specifically expect this denom here

x/incentives/simulation/operations.go
```diff
@@ -156,7 +154,7 @@ func SimulateMsgAddToGauge(ak stakingTypes.AccountKeeper, bk stakingTypes.BankKe
 	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
 		simAccount, _ := simtypes.RandomAcc(r, accs)
 		simCoins := bk.SpendableCoins(ctx, simAccount.Address)
-		if simCoins.Len() <= 0 || simCoins.AmountOf(appparams.BaseCoinUnit).LT(types.AddToGaugeFee) {
+		if simCoins.Len() <= 0 || simCoins.AmountOf(simCoins[0].Denom).LT(types.AddToGaugeFee) {
```
p0mvn, line 157: ditto: `sdk.DefaultBondDenom`

p0mvn (approved):
LGTM, nice work

app/keepers/keepers.go
```diff
@@ -260,6 +260,19 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
...
 	appKeepers.EpochsKeeper = epochskeeper.NewKeeper(appCodec, appKeepers.keys[epochstypes.StoreKey])
 
+	txFeesKeeper := txfeeskeeper.NewKeeper(
+		appCodec,
+		appKeepers.AccountKeeper,
+		appKeepers.BankKeeper,
+		appKeepers.EpochsKeeper,
+		appKeepers.keys[txfeestypes.StoreKey],
+		appKeepers.GAMMKeeper,
+		appKeepers.GAMMKeeper,
```
stackman27, line 270: @p0mvn why do we need to do use `appKeepers.GAMMKeeper` twice here?

p0mvn:
Could you please add a changelog entry for this? 

This would be API breaking due to the constructor update. However, I think we should still backport this to v11 since it's not state-breaking

app/keepers/keepers.go
p0mvn: I'm not sure - I think we should have grouped this into a single parameter and consolidate the two interfaces (`SpotPriceCalculator` and `GammKeeper`) into a single one

If you you'd like, please make an issue to fix this. If no objections from others, we can work on this

stackman27:
@p0mvn addressed your feedback here [`720f3bd`](https://github.com/osmosis-labs/osmosis/pull/2283/commits/720f3bde95486c53acbdb66f110deecd7ee5e061)

app/keepers/keepers.go
stackman27: Gotchu i'll look into it and try to fix it!

x/incentives/keeper/gauge.go
```diff
@@ -287,12 +286,19 @@ func (k Keeper) GetEpochInfo(ctx sdk.Context) epochtypes.EpochInfo {
 // The fee is sent to the community pool.
 // Returns nil on success, error otherwise.
 func (k Keeper) chargeFeeIfSufficientFeeDenomBalance(ctx sdk.Context, address sdk.AccAddress, fee sdk.Int, gaugeCoins sdk.Coins) (err error) {
-	totalCost := gaugeCoins.AmountOf(appparams.BaseCoinUnit).Add(fee)
-	accountBalance := k.bk.GetBalance(ctx, address, appparams.BaseCoinUnit).Amount
+	feeDenom, err := k.tk.GetBaseDenom(ctx)
+	if err != nil {
+		return err
+	}
```
p0mvn, line 292: This is technically a state break but the likelihood of this non-deterministically happening is low so I think we can still backport this

czarcas7ic (approved):
Thanks for doing this, now the simulator will be happy again 😄

stackman27:
@czarcas7ic good to go [`f245297`](https://github.com/osmosis-labs/osmosis/pull/2283/commits/f24529778af49d5f7da6bff8e6e4504acdce9d52)

Can't upset the simulator for too long :D

x/incentives/keeper/gauge.go
p0mvn: Actually, since this is on the message path, it is possible to run out of gas. So I think we should avoid backporting

Issue #2181, opened 2022-07-21
refactor(osmoutils): make ReverseSlice mutative 


p0mvn:
**Background**

`ReverseSlice` can be made mutative and reverse in place, achieving O(n/2) time and no extra space by maintaining a pointer to the beginning and to the end of the slice

Example: https://stackoverflow.com/questions/44452668/how-to-reverse-string-in-place-in-java

Original Discussion: https://github.com/osmosis-labs/osmosis/pull/2175#discussion_r926150035

**Acceptance Criteria**
- the suggested algorithm is implemented
- table-driven tests added

sehyunc:
Hey @p0mvn! Is anyone working on this, and if not could I be assigned it?

p0mvn:
Hey @sehyunc . Seeing that you're already assigned but yes, please go for it! Thank you

PR #2286, opened 2022-08-03
refactor: make ReverseSlice mutative

sehyunc:
Closes: #2181 

Refactor `ReverseSlice` utility function to be mutative and reverse in place.

## Brief Changelog

* refactor existing `ReverseSlice` function
* add tests in new test file specifically for this function

## Testing and Verifying

* `TestReverseSlice` in `slice_helper_test.go`

### DIFF
--- a/osmoutils/slice_helper.go
+++ b/osmoutils/slice_helper.go
@@ -24,7 +24,7 @@ func Filter[T interface{}](filter func(T) bool, s []T) []T {
 	return filteredSlice
 }
 
-// Reverse slice reverses the input slice in place.
+// ReverseSlice reverses the input slice in place.
 // Does mutate argument.
 func ReverseSlice[T any](s []T) []T {
 	maxIndex := len(s)
--- a/osmoutils/slice_helper.go
+++ b/osmoutils/slice_helper.go
@@ -24,13 +24,24 @@ func Filter[T interface{}](filter func(T) bool, s []T) []T {
 	return filteredSlice
 }
 
-// ReverseSlice returns a reversed copy of the input slice.
-// Does not mutate argument.
-func ReverseSlice[T any](s []T) []T {
-	newSlice := make([]T, len(s))
-	maxIndex := len(s) - 1
-	for i := 0; i < len(s); i++ {
-		newSlice[maxIndex-i] = s[i]
+// Reverse slice reverses the input slice in place.
+// Does mutate argument.
+func ReverseSlice[T any](s []T) {
+	maxIndex := len(s)
+	for i := 0; i < maxIndex/2; i++ {
+		temp := s[i]
+		s[i] = s[maxIndex-i-1]
+		s[maxIndex-1-i] = temp
 	}
-	return newSlice
 }
+
+// // ReverseSlice returns a reversed copy of the input slice.
+// // Does not mutate argument.
+// func ReverseSlice[T any](s []T) []T {
+// 	newSlice := make([]T, len(s))
+// 	maxIndex := len(s) - 1
+// 	for i := 0; i < len(s); i++ {
+// 		newSlice[maxIndex-i] = s[i]
+// 	}
+// 	return newSlice
+// }
--- a/osmoutils/slice_helper.go
+++ b/osmoutils/slice_helper.go
@@ -26,22 +26,12 @@ func Filter[T interface{}](filter func(T) bool, s []T) []T {
 
 // Reverse slice reverses the input slice in place.
 // Does mutate argument.
-func ReverseSlice[T any](s []T) {
+func ReverseSlice[T any](s []T) []T {
 	maxIndex := len(s)
 	for i := 0; i < maxIndex/2; i++ {
 		temp := s[i]
 		s[i] = s[maxIndex-i-1]
 		s[maxIndex-1-i] = temp
 	}
+	return s
 }
-
-// // ReverseSlice returns a reversed copy of the input slice.
-// // Does not mutate argument.
-// func ReverseSlice[T any](s []T) []T {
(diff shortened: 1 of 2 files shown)


stackman27 (commented):
Left a bit of the comments. Congratulations on the PR!! 😎

osmoutils/slice_helper_test.go
```diff
@@ -0,0 +1,28 @@
...
+		"Empty array":             {[]string{}, []string{}},
+		"Odd length array":        {[]string{"a", "b", "c"}, []string{"c", "b", "a"}},
+		"Single element array":    {[]string{"a"}, []string{"a"}},
+		"Array with empty string": {[]string{"a", "b", "c", "", "d"}, []string{"d", "", "c", "b", "a"}},
+		"Array with numbers":      {[]string{"a", "b", "c", "1", "2", "3"}, []string{"3", "2", "1", "c", "b", "a"}},
+	}
+
+	for _, tc := range tests {
+		actualSolvedInput := ReverseSlice(tc.s)
+		require.True(t, reflect.DeepEqual(actualSolvedInput, tc.expectedSolvedInput))
```
stackman27, line 26: Not sure if `require.True` is necessary while using DeepEqual 

```suggestion
		reflect.DeepEqual(actualSolvedInput, tc.expectedSolvedInput)
```

osmoutils/slice_helper.go
```diff
@@ -24,13 +24,14 @@ func Filter[T interface{}](filter func(T) bool, s []T) []T {
 	return filteredSlice
 }
 
-// ReverseSlice returns a reversed copy of the input slice.
-// Does not mutate argument.
+// Reverse slice reverses the input slice in place.
```
stackman27, line 27:
```suggestion
// ReverseSlice reverses the input slice in place.
```

osmoutils/slice_helper_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}{
+		"Even length array":       {[]string{"a", "b", "c", "d"}, []string{"d", "c", "b", "a"}},
+		"Empty array":             {[]string{}, []string{}},
+		"Odd length array":        {[]string{"a", "b", "c"}, []string{"c", "b", "a"}},
+		"Single element array":    {[]string{"a"}, []string{"a"}},
+		"Array with empty string": {[]string{"a", "b", "c", "", "d"}, []string{"d", "", "c", "b", "a"}},
+		"Array with numbers":      {[]string{"a", "b", "c", "1", "2", "3"}, []string{"3", "2", "1", "c", "b", "a"}},
+	}
+
+	for _, tc := range tests {
```
stackman27, line 24: We can probably wrap this up with something like; 

```
for name, tc := range tests {
t.Run(name, func(t *testing.T) {
 ...
}
}
```

osmoutils/slice_helper_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestReverseSlice(t *testing.T) {
+	tests := map[string]struct {
+		s []string
+
+		expectedSolvedInput []string
+	}{
+		"Even length array":       {[]string{"a", "b", "c", "d"}, []string{"d", "c", "b", "a"}},
```
AlpinYukseloglu, line 16:
```suggestion
		"Even length array":       {
		    s: []string{"a", "b", "c", "d"}, 
		    expectedSolvedInput: []string{"d", "c", "b", "a"}
		},
```

AlpinYukseloglu (commented):
Thanks for this PR! Other than the comment I left about test structure/readability, the core logic LGTM! It would be great if you added error case(s) just to be comprehensive though imo

osmoutils/slice_helper_test.go
p0mvn: `DeepEqual` returns a boolean but does not assert. We still would like to make the assertion on the result

p0mvn: Suggest trying the following if the goal is to simplify:
```
require.EqualValues(t, tc.expectedSolvedInput, actualSolvedInput)
```

osmoutils/slice_helper_test.go
```diff
@@ -0,0 +1,28 @@
+package osmoutils
```
p0mvn, line 1:
```suggestion
package osmoutils_test
```

p0mvn (commented):
Nice work!

osmoutils/slice_helper_test.go
ValarDragon: You can also just do `require.Equal` for slices,. https://play.golang.com/p/HKavaKV9bd0

ValarDragon (approved):
LGTM after minor notes added by others

osmoutils/slice_helper_test.go
sehyunc: When I make this change, I get the `undeclared name: ReverseSlice` error. Looking at `binary_search_test.go`, it also has package set to `osmoutils`. Have I configured something incorrectly?

stackman27: You can fix this by importing osmoutils like below. I'm not sure why we dont use package osmoutils_test in `binary_search_test.go`. Maybe it's worth changing both or keeping this package name the same as `binary_search_test.go` 

```
import (
	"github.com/osmosis-labs/osmosis/v10/osmoutils"
...
)

osmoutils.ReverseSlice(...)
```

sehyunc: Thanks for the tip sir. Still getting used to Go 😅

ValarDragon:
Thanks for the PR!

PR #2290, opened 2022-08-03
refactor(x/mint): simplify TestDistributeMintedCoin

p0mvn:
Closes: #XXX

This change simplifies `DistributeMintedCoin` by doing the following:
-  removing `TestDistributeMintedCoin_ToDeveloperRewardsAddr`. Instead, making this a table-driven test case in `TestDistributeMintedCoin`
- Removing all gauge logic from `TestDistributeMintedCoin`. 
   * By having gauge logic in that test, as it was previously,  we were overcomplicating that test and making it a functional test as opposed to a unit test. 
   * What we want to assert in the table-driven unit test at this level of abstraction is simply the fact that the `AfterDistributeMintedCoin` hook is called. It should be irrelevant what the `AfterDistributeMintedCoin` hook does. Instead, this should be verified by the tests for the  `AfterDistributeMintedCoin` or functional / integration tests.
   * The fact that the hook is called can be verified with the newly added mock.

## Testing and Verifying

This change is already covered by existing tests, such as:
- `TestDistributeMintedCoin`

### DIFF
--- a/x/mint/keeper/export_test.go
+++ b/x/mint/keeper/export_test.go
@@ -27,3 +27,14 @@ func (k Keeper) DistributeToModule(ctx sdk.Context, recipientModule string, mint
 func (k Keeper) DistributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.Coin, developerRewardsProportion sdk.Dec, developerRewardsReceivers []types.WeightedAddress) (sdk.Int, error) {
 	return k.distributeDeveloperRewards(ctx, totalMintedCoin, developerRewardsProportion, developerRewardsReceivers)
 }
+
+// Set the mint hooks. This is used for testing purposes only.
+func (k *Keeper) SetMintHooksUnsafe(h types.MintHooks) *Keeper {
+	k.hooks = h
+	return k
+}
+
+// Get the mint hooks. This is used for testing purposes only.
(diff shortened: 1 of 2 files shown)


x/mint/keeper/export_test.go
```diff
@@ -27,3 +27,14 @@ func (k Keeper) DistributeToModule(ctx sdk.Context, recipientModule string, mint
 func (k Keeper) DistributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.Coin, developerRewardsProportion sdk.Dec, developerRewardsReceivers []types.WeightedAddress) (sdk.Int, error) {
 	return k.distributeDeveloperRewards(ctx, totalMintedCoin, developerRewardsProportion, developerRewardsReceivers)
 }
+
+// Set the mint hooks. This is used for testing purposes only.
+func (k *Keeper) SetMintHooksUnsafe(h types.MintHooks) *Keeper {
```
p0mvn, line 32: Note: since thse methods are in `export_test.go`, they do not get compiled into the final binary and are not accessible by the core application logic

alexanderbez (approved):

p0mvn:
Merging this since this is an improvement to the tests and does not change the core logic

PR #2293, opened 2022-08-03
simulator: include gauge fees and cap jailed validators

czarcas7ic:

A couple of issues have been found on long simulator runs. This PR fixes both of the issues I have found so far

## Brief Changelog

- Reduces rate at which validators double sign
  - Caps double signs to allow a minimum of 10 validators at any time
  - (we need to address this further by including some mech to add validators to the sim set)
- Now takes into consideration the respective gauge fees when calculating random Create/Add to gauge messages

## Testing and Verifying

### DIFF
--- a/simulation/executor/mock_tendermint.go
+++ b/simulation/executor/mock_tendermint.go
@@ -193,13 +193,17 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
 	pastVoteInfos [][]abci.VoteInfo,
 	event func(route, op, evResult string), header tmproto.Header, voteInfos []abci.VoteInfo) []abci.Evidence {
 	evidence := []abci.Evidence{}
-	// return if no past times
-	if len(pastTimes) == 0 {
+	// return if no past times or if only 10 validators remaining in the active set
+	if len(pastTimes) == 0 || len(voteInfos) <= 10 {
 		return evidence
 	}
-
+	var n float64 = 1
 	// TODO: Change this to be markov based & clean this up
-	for r.Float64() < params.EvidenceFraction() {
+	// Right now we incrementally lower the evidence fraction to make
+	// it less likely to jail many validators in one run
+	// We should also add some method of including new validators into the set
+	// instead of being stuck with the ones we start with during initialization
+	for r.Float64() < (params.EvidenceFraction() / n) {
 		height := header.Height
 		time := header.Time
 		vals := voteInfos
@@ -229,6 +233,7 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
 		)
 
 		event("begin_block", "evidence", "ok")
+		n++
 	}
 	return evidence
 }
--- a/simulation/executor/mock_tendermint.go
+++ b/simulation/executor/mock_tendermint.go
@@ -200,12 +200,12 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
 	var n float64 = 1
 	// TODO: Change this to be markov based & clean this up
 	// Right now we incrementally lower the evidence fraction to make
-	// it less likely to jail many validators in one run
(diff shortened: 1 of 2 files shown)


simulation/executor/mock_tendermint.go
```diff
@@ -193,13 +193,17 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
...
 	}
-
+	var n float64 = 1
 	// TODO: Change this to be markov based & clean this up
-	for r.Float64() < params.EvidenceFraction() {
+	// Right now we incrementally lower the evidence fraction to make
+	// it less likely to jail many validators in one run
+	// We should also add some method of including new validators into the set
+	// instead of being stuck with the ones we start with during initialization
+	for r.Float64() < (params.EvidenceFraction() / n) {
```
ValarDragon, line 206: Why don't we just say that if num validators remaining = 1, continue? (Don't test double signing the last validator?)

czarcas7ic: Sounds good, just implemented that! Still needed to keep the counter for the following reasons:

1. The validator length doesn't update after appending to the evidence slice, so this leads to edge cases of still jailing too many
2. Dividing the evidence fraction by the iteration number sizeably slows down the jailing

simulation/executor/mock_tendermint.go
```diff
@@ -193,13 +193,21 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
...
 	if len(pastTimes) == 0 {
 		return evidence
 	}
-
+	var n float64 = 1
 	// TODO: Change this to be markov based & clean this up
-	for r.Float64() < params.EvidenceFraction() {
+	// Right now we incrementally lower the evidence fraction to make
+	// it less likely to jail many validators in one run
+	// We should also add some method of including new validators into the set
```
alexanderbez, line 204:
```suggestion
	// it less likely to jail many validators in one run.
	// We should also add some method of including new validators into the set
```

simulation/executor/mock_tendermint.go
```diff
@@ -193,13 +193,21 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
...
+	// instead of being stuck with the ones we start with during initialization
```
alexanderbez, line 205:
```suggestion
	// instead of being stuck with the ones we start with during initialization.
```

simulation/executor/mock_tendermint.go
```diff
@@ -193,13 +193,21 @@ func randomDoubleSignEvidence(r *rand.Rand, params Params,
...
+	for r.Float64() < (params.EvidenceFraction() / n) {
+		// if only one validator remaining, don't jail any more validators
+		if len(voteInfos)-int(n) == 0 {
```
alexanderbez, line 208: might be safer
```suggestion
		if len(voteInfos)-int(n) <= 0 {
```

x/incentives/simulation/operations.go
```diff
@@ -63,14 +63,23 @@ func WeightedOperations(
 }
 
 // genRewardCoins generates a random number of coin denoms with a respective random value for each coin.
-func genRewardCoins(r *rand.Rand, coins sdk.Coins) (res sdk.Coins) {
+func genRewardCoins(r *rand.Rand, coins sdk.Coins, fee sdk.Int) (res sdk.Coins) {
 	numCoins := 1 + r.Intn(Min(coins.Len(), 1))
 	denomIndices := r.Perm(numCoins)
 	for i := 0; i < numCoins; i++ {
+		var amt sdk.Int
+		var err error
```
alexanderbez, line 71: nit: group the vars

alexanderbez (approved):

ValarDragon (approved):

Issue #2197, opened 2022-07-22
Twap: Add historic pruning


ValarDragon:
## Background

A follow-up from #2168 thats needed is adding pruning of old TWAP records.

We need to basically implement it right here: https://github.com/osmosis-labs/osmosis/pull/2168/files#diff-4afe8ba5fca88fef0034d6b2e8f9803a4be33d4735a41163564c4c7495934985R47-R49

and then write a test for it,

## Suggested Design

Use the HistoricalTimeIndex (https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/twap/types/keys.go#L33) to get all locks defined before a time, and delete them all.

Write a test for deletion working as expected (and not deleting things in last 48 hours)

## Acceptance Criteria

- [ ] Records older than 48 hours are deleted at epoch.
- [ ] Records younger than 48 hours are not deleted at epoch
- [ ] We computationally only iterate over epochs to be deleted.
- [ ] Test exists guaranteeing above properties

PR #2294, opened 2022-08-03
feat(twap): historic pruning

p0mvn:
Closes: #2197

Adds TWAP historic pruning and tests.

## Brief Changelog
- [x] implement pruning in `pruneRecordsBeforeTime`
- [x] implement getters for all records and unit test them (`helpers_test.go`)
- [x] use getters to test the pruning logic
- [x] reuse denom constants in tests
- [x] Create follow-up issue for [`AfterEpochEnd`](https://github.com/osmosis-labs/osmosis/blob/5e74b564c796519cae7d528b08cd3af95ccfb71f/x/gamm/twap/hook_listener.go#L29) tests: https://github.com/osmosis-labs/osmosis/issues/2372
- [x] Create follow-up issue for setting pruning-keep period via parameter: https://github.com/osmosis-labs/osmosis/issues/2371

## Testing and Verifying

- `TestGetAllHistoricalPoolIndexedTWAPs`
- `TestGetAllHistoricalTimeIndexedTWAPs`
- `TestPruneRecordsBeforeTime`


x/gamm/twap/store.go
```diff
@@ -51,8 +51,18 @@ func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
...
+	store := ctx.KVStore(k.storeKey)
+	iter := store.Iterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
+	defer iter.Close()
+	for ; iter.Valid(); iter.Next() {
+		twapToRemove, err := types.ParseTwapFromBz(iter.Value())
+		if err != nil {
+			return err
+		}
+		k.deleteHistoricalRecord(ctx, twapToRemove)
+	}
```
ValarDragon, line 64: Can we push this to a helper in osmo_utils/store ?

Also I'm not entirely sure if your allowed to delete keys while iterating over them or if you have to queue all the keys and delete in a second pass over it. I mostly have just forgotten / mixed up a lot of the oddiites of the cache kv store

p0mvn: > not entirely sure if your allowed to delete keys while iterating over them or if you have to queue all the keys and delete in a second pass over it

I looked into the underlying implementation. So, what seems to be happening on `Delete()` is that we:
- mark a record as deleted in cache: https://github.com/osmosis-labs/cosmos-sdk/blob/f09814c9e23fd5832b11666a871bf51f36f85e41/store/cachekv/store.go#L99
- then, when calling `iter.Next()`, it skips until a record exists or iterator becomes invalid:https://github.com/osmosis-labs/cosmos-sdk/blob/f09814c9e23fd5832b11666a871bf51f36f85e41/store/cachekv/mergeiterator.go#L63

We do not write into IAVL until commit so the deletions are handled in cache.
Additionally, all tests passing makes me more confident that there is no problem with deleting while iterating.

p0mvn: > Can we push this to a helper in osmo_utils/store ?

I'm not sure if I can generalize this well enough to justify the move to `osmo_utils/store`. 

I tried this:
```
func iterate(store sdk.KVStore, from []byte, to []byte, fn func() error) error {
	iter := store.Iterator(from, to)
	defer iter.Close()
	for ; iter.Valid(); iter.Next() {
		if err := fn(); err != nil {
			return err
		}
	}
	return nil
}
```

So that the method in question would look like:
```
func (k Keeper) pruneRecordsBeforeTime(ctx sdk.Context, lastTime time.Time) error {

	err := iterate(store, []byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""), func() error {
		return nil
	})

	if err != nil {
		return err
	}
```

However, I think this abstraction only adds unnecessary complexity without reducing that much boilerplate. Please let me know what you think

x/twap/helpers_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v10/osmoutils"
+	"github.com/osmosis-labs/osmosis/v10/x/twap/types"
+)
+
+// GetAllHistoricalTimeIndexedTWAPs returns all historical TWAPs indexed by time.
+func (k Keeper) GetAllHistoricalTimeIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
```
p0mvn, line 11: N.B: these methods are only used in tests so I moved them into a file suffixed by `_test.go` to avoid having them exposed in the rest of the system.

x/gamm/twap/store.go
ValarDragon: Ah agreed your right - I was thinking that it was deleting all keys in the range. Didn't realize it was actually calling another methoid for delete, that deleted two things.

ValarDragon:
We should remove 
```
//nolint:unused,deadcode
```
right above
```
//nolint:unused,deadcode
func (k Keeper) deleteHistoricalRecord(ctx sdk.Context, twap types.TwapRecord) {
```

x/twap/api_test.go
```diff
@@ -69,7 +69,7 @@ func (s *TestSuite) TestGetArithmeticTwap() {
 	}
 
 	makeSimpleTwapInput := func(startTime time.Time, endTime time.Time, isQuoteTokenA bool) getTwapInput {
-		quoteAssetDenom, baseAssetDenom := denom0, denom1
```
ValarDragon, line 72: Why isn't this breaking test cases here? it breaks test cases for me locally?

x/twap/helpers_test.go
```diff
@@ -0,0 +1,18 @@
...
+	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPTimeIndexPrefix), []byte(types.HistoricalTWAPTimeIndexPrefix+"a"), types.ParseTwapFromBz)
+}
+
+// GetAllHistoricalPoolIndexedTWAPs returns all historical TWAPs indexed by pool.
+func (k Keeper) GetAllHistoricalPoolIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
+	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPPoolIndexPrefix), []byte(types.HistoricalTWAPPoolIndexPrefix+"a"), types.ParseTwapFromBz)
```
ValarDragon, line 17: Can we move these helper methods to `keeper_test.go`

x/twap/keeper_test.go
```diff
@@ -14,9 +14,10 @@ import (
 
 // TODO: Consider switching this everywhere
 var (
-	denom0            = "token/B"
-	denom1            = "token/A"
-	defaultUniV2Coins = sdk.NewCoins(sdk.NewInt64Coin(denom0, 1_000_000_000), sdk.NewInt64Coin(denom1, 1_000_000_000))
+	denom0            = "token/A"
+	denom1            = "token/B"
```
ValarDragon, line 18: why are we switching / changing denom order?

ValarDragon: Actually I think this is fine

x/twap/listeners.go
```diff
@@ -26,7 +28,11 @@ type epochhook struct {
 
 func (hook *epochhook) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) {
 	if epochIdentifier == hook.k.PruneEpochIdentifier(ctx) {
-		hook.k.pruneOldTwaps(ctx)
+		// TODO: configure lastKeptTime via parameter.
+		lastKeptTime := ctx.BlockTime().Add(-48 * time.Hour)
+		if err := hook.k.pruneRecordsBeforeTime(ctx, lastKeptTime); err != nil {
+			ctx.Logger().Error("Error pruning old twaps at the epoch end", err)
+		}
```
ValarDragon, line 35: Can we put this pruning logic in logic.go

x/twap/helpers_test.go
p0mvn: They need to be in `twap` package to be defined on keeper correctly. I ended up moving them unexported to `store.go` and then exposing via `export_test.go`. Let me know what you think

x/twap/keeper_test.go
p0mvn: Yes, I did not change the order in tests. Only renamed here so that I can add another `denom2` == `token/C` while having the identifier ordering to still make sense

x/twap/listeners.go
p0mvn: I think that the log and final error handling should belong at the highest level of abstraction. However, I encapsulated the pruning time logic: [04ca37d](https://github.com/osmosis-labs/osmosis/pull/2294/commits/04ca37df6615a666f569a15c351a7283d490893c). Let me know what you think

x/twap/logic.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/osmosis-labs/osmosis/v10/x/twap/types"
 )
 
+// TODO: configure lastKeptTime via parameter.
```
p0mvn, line 11:
```suggestion
// TODO: configure recordHistoryKeepPeriod via parameter.
```

x/twap/types/keys.go
```diff
@@ -29,9 +29,16 @@ var (
 	historicalTWAPTimeIndexNoSeparator = "historical_time_index"
 	historicalTWAPPoolIndexNoSeparator = "historical_pool_index"
 
-	mostRecentTWAPsPrefix         = mostRecentTWAPsNoSeparator + KeySeparator
-	historicalTWAPTimeIndexPrefix = historicalTWAPTimeIndexNoSeparator + KeySeparator
-	historicalTWAPPoolIndexPrefix = historicalTWAPPoolIndexNoSeparator + KeySeparator
+	mostRecentTWAPsPrefix = mostRecentTWAPsNoSeparator + KeySeparator
+	// keySeparatorPlusOne is used for creating prefixes for the key end in iterators
```
mattverse, line 33: What's the reason we can't use the conventional way of iterating through all stores using prefix stores?

x/twap/listeners_test.go
```diff
@@ -11,14 +11,14 @@ import (
 func (s *TestSuite) TestCreateTwoAssetPoolFlow() {
 	poolId := s.PrepareUni2PoolWithAssets(defaultUniV2Coins[0], defaultUniV2Coins[1])
 
-	expectedTwap, err := types.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denom0, denom1)
+	expectedTwap, err := types.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denom1, denom0)
```
mattverse, line 14: What's the context behind all these denom order changes?

x/twap/types/keys.go
```diff
@@ -29,9 +29,16 @@ var (
 	historicalTWAPTimeIndexNoSeparator = "historical_time_index"
 	historicalTWAPPoolIndexNoSeparator = "historical_pool_index"
 
-	mostRecentTWAPsPrefix         = mostRecentTWAPsNoSeparator + KeySeparator
-	historicalTWAPTimeIndexPrefix = historicalTWAPTimeIndexNoSeparator + KeySeparator
-	historicalTWAPPoolIndexPrefix = historicalTWAPPoolIndexNoSeparator + KeySeparator
+	mostRecentTWAPsPrefix = mostRecentTWAPsNoSeparator + KeySeparator
```
mattverse, line 32: What's the reason we can't use the conventional way of iterating through all stores using prefix stores?

x/twap/keeper_test.go
```diff
@@ -43,12 +45,59 @@ func (s *TestSuite) setupDefaultPool() (poolId uint64, denomA, denomB string) {
...
+// preSetRecords pre sets records on the twap keeper to the
+// given records.
+func (s *TestSuite) preSetRecords(records []types.TwapRecord) {
+	for _, record := range records {
+		s.twapkeeper.StoreNewRecord(s.Ctx, record)
+	}
+}
+
+// validateExpectedRecords validates that the twap keeper has the expected records.
+func (s *TestSuite) validateExpectedRecords(expectedRecords []types.TwapRecord) {
```
mattverse, line 57: love this <3

x/twap/logic.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/osmosis-labs/osmosis/v10/x/twap/types"
 )
 
+// TODO: configure recordHistoryKeepPeriod via parameter.
```
mattverse, line 11: by parameter, does it indicate gov params?

x/twap/keeper_test.go
```diff
@@ -43,12 +45,59 @@ func (s *TestSuite) setupDefaultPool() (poolId uint64, denomA, denomB string) {
...
+	// validate that the pool indexed TWAPs are cleared.
+	poolIndexedTwaps, err := twapKeeper.GetAllHistoricalPoolIndexedTWAPs(s.Ctx)
+	s.Require().NoError(err)
+	s.Require().Len(poolIndexedTwaps, len(expectedRecords))
+	// N.B.: ElementsMatch is used here because order might differ from expected due to
+	// diverging indexing structure.
+	s.Require().ElementsMatch(poolIndexedTwaps, expectedRecords)
+}
+
+// createTestRecordsFromTime creates and returns 2 test records in the following order:
```
mattverse, line 75:
```suggestion
// createTestRecordsFromTime creates and returns 4 test records in the following order:
```

x/twap/store_test.go
```diff
@@ -189,3 +188,173 @@ func (s *TestSuite) TestGetRecordAtOrBeforeTime() {
...
+				return
+			}
+			s.Require().NoError(err)
+
+			s.validateExpectedRecords(tc.expectedKeptRecords)
+		})
+	}
+}
+
+func (s *TestSuite) TestGetAllHistoricalTimeIndexedTWAPs() {
```
mattverse, line 286: I think we can combine what we're testing with the test above by calling `GetAllHistoricalPoolIndexedTWAPs` before we prune in the test above

x/twap/store_test.go
```diff
@@ -189,3 +188,173 @@ func (s *TestSuite) TestGetRecordAtOrBeforeTime() {
...
+	for name, tc := range tests {
+		s.Run(name, func() {
+			// Setup.
+			s.SetupTest()
+			ctx := s.Ctx
+			twapKeeper := s.twapkeeper
+			s.preSetRecords(tc.expectedRecords)
+
+			// System under test.
+			actualRecords, err := twapKeeper.GetAllHistoricalPoolIndexedTWAPs(ctx)
```
mattverse, line 353: ditto here for combining the test cases

x/twap/keeper_test.go
```diff
@@ -43,12 +45,59 @@ func (s *TestSuite) setupDefaultPool() (poolId uint64, denomA, denomB string) {
...
+// - at time t - 2
+// - at time t - 1
+// - at time t
+// - at time t + 1
```
mattverse, line 79:
```suggestion
// - at time t - 2 second
// - at time t - 1 second
// - at time t
// - at time t + 1 second
```

mattverse (approved):
Left some minor comments, please take a look! Otherwise LGTM!

x/twap/keeper_test.go
alexanderbez: or just `*s`

alexanderbez (approved):

x/twap/listeners_test.go
czarcas7ic: Second this. Wish there was a way we could make `Asset0Denom: denom0` to prevent mental gymnastics that may need to occur when working on this in the future.

x/twap/logic.go
```diff
@@ -53,7 +56,7 @@ func (k Keeper) updateRecords(ctx sdk.Context, poolId uint64) error {
 	return nil
 }
 
-// mutates record argument, but not with all the changes.
+// updateRecord mutates record argument, but not with all the changes.
 // Use the return value, and drop usage of the argument.
 func (k Keeper) updateRecord(ctx sdk.Context, record types.TwapRecord) types.TwapRecord {
```
czarcas7ic, line 61: I read this a couple times and can't quite figure out what we mean by "but not with all the changes". Can we be more specific?

x/twap/logic.go
```diff
@@ -69,6 +72,13 @@ func (k Keeper) updateRecord(ctx sdk.Context, record types.TwapRecord) types.Twa
 	return newRecord
 }
 
+// pruneRecords prunes twap records that hapenned earlier than recordHistoryKeepPeriod
```
czarcas7ic, line 75:
```suggestion
// pruneRecords prunes twap records that happened earlier than recordHistoryKeepPeriod
```

x/twap/logic_test.go
```diff
@@ -209,3 +209,26 @@ func TestComputeArithmeticTwap(t *testing.T) {
 		})
 	}
 }
+
+// TestPruneRecordsBeforeTime tests that all twap records earlier than
```
czarcas7ic, line 213:
```suggestion
// TestPruneRecords tests that all twap records earlier than
```

x/twap/logic_test.go
```diff
@@ -209,3 +209,26 @@ func TestComputeArithmeticTwap(t *testing.T) {
...
+// current block time - RecordHistoryKeepPeriod are pruned from the store.
+func (s *TestSuite) TestPruneRecords() {
+	tMin2Record, tMin1Record, baseRecord, tPlus1Record := s.createTestRecordsFromTime(baseTime.Add(-twap.RecordHistoryKeepPeriod))
+
+	// non-ascending inserton order.
```
czarcas7ic, line 218:
```suggestion
	// non-ascending insertion order.
```

x/twap/store.go
```diff
@@ -82,6 +91,16 @@ func (k Keeper) getAllMostRecentRecordsForPool(ctx sdk.Context, poolId uint64) (
 	return types.GetAllMostRecentTwapsForPool(store, poolId)
 }
 
+// getAllHistoricalTimeIndexedTWAPs returns all historical TWAPs indexed by time.
+func (k Keeper) getAllHistoricalTimeIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
+	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPTimeIndexPrefix), []byte(types.HistoricalTWAPTimeIndexPrefixEnd), types.ParseTwapFromBz)
+}
+
+// getAllHistoricalPoolIndexedTWAPs returns all historical TWAPs indexed by pool.
```
czarcas7ic, line 99: Indexed by Pool ID right?

x/twap/store_test.go
```diff
@@ -82,14 +81,14 @@ func (s *TestSuite) TestGetAllMostRecentRecordsForPool() {
...
+				newEmptyPriceRecord(1, baseTime, denom2, denom0)},
 			poolId: 1,
 			expectedRecords: []types.TwapRecord{
-				newEmptyPriceRecord(1, baseTime, denomB, denomA),
-				newEmptyPriceRecord(1, baseTime, denomC, denomA),
-				newEmptyPriceRecord(1, baseTime, denomC, denomB)},
+				newEmptyPriceRecord(1, baseTime, denom1, denom0),
+				newEmptyPriceRecord(1, baseTime, denom2, denom0),
+				newEmptyPriceRecord(1, baseTime, denom2, denom1)},
 		},
```
czarcas7ic, line 92: Might be overkill, but should we also do two multi-asset pool records, then set them in reverse order to unsure the last record, independent of time, still takes precedence?

x/twap/store_test.go
```diff
@@ -189,3 +188,173 @@ func (s *TestSuite) TestGetRecordAtOrBeforeTime() {
 		})
 	}
 }
+
+// TestPruneRecordsBeforeTime tests that all twap records earlier than
+// current block time - given time are pruned from the store.
+func (s *TestSuite) TestPruneRecordsBeforeTime() {
+	tMin2Record, tMin1Record, baseRecord, tPlus1Record := s.createTestRecordsFromTime(baseTime)
+
+	// non-ascending inserton order.
```
czarcas7ic, line 197:
```suggestion
	// non-ascending insertion order.
```

x/twap/types/keys.go
```diff
@@ -29,9 +29,16 @@ var (
...
+	// keySeparatorPlusOne is used for creating prefixes for the key end in iterators
+	// when we want to get all of the keys in a prefix. Since it is one byte larger
+	// than the original key separator and the end prefix is exclusive, it is a valid
```
czarcas7ic, line 35:
```suggestion
	// than the original key separator and the end prefix is exclusive, it is valid
```

czarcas7ic (approved):
LGTM, only minor non-blocking comments!

x/twap/listeners_test.go
mattverse: I think its because of how the twap assets are alphabetically ordered (maybe)

x/twap/store.go
```diff
@@ -82,6 +91,16 @@ func (k Keeper) getAllMostRecentRecordsForPool(ctx sdk.Context, poolId uint64) (
...
+func (k Keeper) getAllHistoricalPoolIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
+	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPPoolIndexPrefix), []byte(types.HistoricalTWAPPoolIndexPrefixEnd), types.ParseTwapFromBz)
```
AlpinYukseloglu, line 101: This version is much more intuitive than the original `+"a"` for getting prefixed entries in range, great change!

AlpinYukseloglu (approved):
LGTM! Left a couple minor comments but overall looks good

p0mvn:
Thanks for the reviews everyone!

x/twap/logic.go
```diff
@@ -69,6 +72,13 @@ func (k Keeper) updateRecord(ctx sdk.Context, record types.TwapRecord) types.Twa
...
+// before current block time.
+func (k Keeper) pruneRecords(ctx sdk.Context) error {
+	lastKeptTime := ctx.BlockTime().Add(-recordHistoryKeepPeriod)
```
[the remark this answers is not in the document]
p0mvn: Same here

x/twap/store_test.go
p0mvn: I see it would. However, each test should be focused on testing one thing only, we shouldn't be testing multiple units of computation in one fixture

x/twap/keeper_test.go
p0mvn: Changed locally

x/twap/types/keys.go
p0mvn: What do you mean by the conventional way? Happy to change this if there is something more straight-forward

x/twap/logic.go
p0mvn: Exactly: https://github.com/osmosis-labs/osmosis/issues/2371

x/twap/listeners_test.go
p0mvn: reverted

x/twap/logic.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/osmosis-labs/osmosis/v10/x/twap/types"
 )
 
+// TODO: configure recordHistoryKeepPeriod via parameter.
+const recordHistoryKeepPeriod = 48 * time.Hour
```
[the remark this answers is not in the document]
p0mvn: yes, records younger than 48 hours before current time are deleted at epoch

PR #2297, opened 2022-08-04
Update chain registry for Osmosis upgrade

ValarDragon:

Updates chain registry, also fixes incorrect cosmwasm version entry in both go.mod and chain registry. (Note that nothing about the cosmwasm version changed, despite the go.mod update, We have a replace directive further below that moves it to v0.27.0 thats been on mainnet since v9)

## Brief Changelog

Update chain registry for upgrade

## Testing and Verifying

End to end should cover go/mod sanity check.

### DIFF
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -2,15 +2,15 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "codebase":{
     "git_repo": "https://github.com/osmosis-labs/osmosis",
-    "recommended_version": "v10.0.1",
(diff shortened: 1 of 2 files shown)


sunnya97 (approved):

sunnya97:
Why do we have a chain.json in the osmosis repo?

ValarDragon:
We just PR here, and it auto-updates whats in chain registry. Much easier to manage with binary releases, etc.

czarcas7ic:
@niccoloraspa is this something we want to add in the release CI action?

Issue #2177, opened 2022-07-21
Add test cases for `Unbonded`, `Unbonding` Validator stats in Superfluid module


mattverse:
## Background

We have a field named `validatorStats` all over the place within the superfluid module ( e.g https://github.com/osmosis-labs/osmosis/blob/main/x/superfluid/keeper/stake_test.go#L26). This was orignally designed  to test test cases with different validator status possible, currently three different validator Stats available in the comsos SDK as of now( bonding, unbonded, unbonding). We lack test cases utilizing these fields right now, we need to add different test cases of using these fields all over the superfluid module and test cases where validators have different stats.

## Suggested Design and Acceptance Criteria

Add test cases of `[]stakingtypes.BondStatus{stakingtypes.Unbonded}` and  `[]stakingtypes.BondStatus{stakingtypes.Unbonding}` for the test cases that has the field with `validatorStats`. Doesn't necessarily have to add these test cases in all tests though, mainly the test cases where having different validator stats would have different impact on the tests

PR #2310, opened 2022-08-04
Add test cases Validator stats in Superfluid module

hieuvubk:
Closes: #2177

> Add test cases for `Unbonded`, `Unbonding` validator stat


x/superfluid/keeper/distribution_test.go
```diff
@@ -55,6 +55,14 @@ func (suite *KeeperTestSuite) TestMoveSuperfluidDelegationRewardToGauges() {
 			[]int64{0},
 			[]gaugeChecker{{0, 0, 0, true}, {1, 1, 0, false}},
 		},
+		{
```
mattverse, line 58: can we add more detailed comments for what we're expecting from this test case?
e.g) unbonding validators are not expected to be rewarded

x/superfluid/keeper/hooks_test.go
```diff
@@ -283,6 +303,13 @@ func (suite *KeeperTestSuite) TestBeforeSlashingUnbondingDelegationHook() {
 				suite.Require().NoError(err)
 				// slash by slash factor
 				power := sdk.TokensToConsensusPower(validator.Tokens, sdk.DefaultPowerReduction)
+
+				// should not be slashing unbonded validator
```
mattverse, line 307: asking to learn : does it not matter for the unbonding case?

x/superfluid/keeper/intermediary_account_test.go
```diff
@@ -26,6 +26,18 @@ func (suite *KeeperTestSuite) TestIntermediaryAccountCreation() {
 			2,
 			[]superfluidDelegation{{0, 0, 0, 1000000}, {1, 1, 0, 1000000}},
 		},
+		{
```
mattverse, line 29: nit: can we add details on what we're testing in the name or comments for the expected behavior when the validator is unbonded or unbonding?

x/superfluid/keeper/slash_test.go
```diff
@@ -25,6 +25,22 @@ func (suite *KeeperTestSuite) TestBeforeValidatorSlashed() {
...
+		{
+			"with an unbonding validator and single superfluid delegation",
+			[]stakingtypes.BondStatus{stakingtypes.Unbonding},
+			1,
+			[]superfluidDelegation{{0, 0, 0, 1000000}},
+			[]int64{0},
+			[]int64{0},
+		},
+		{
+			"with single validator and single superfluid delegation",
```
mattverse, line 37: Should be changed to unbonded!

mattverse (commented):
LGTM overall! Nice work figuring out all the cases for unbonding and unbonded cases must have been gnarly. Would you mind adding comments for what we're expecting for when it is unbonding or unbonded? I think also adding documentation on the method itself on how stuff work when validators are unbonded & unbonding would be also super helpful!

x/superfluid/keeper/hooks_test.go
hieuvubk: Yes i tested it does not matter for the unbonding case?

x/superfluid/keeper/distribution_test.go
hieuvubk: It was actually not allocating the reward to the Unbonded validator. Therefore gauges are not distributed.
I added the case of allocating to both Unbonded and Unbonding vals and gauges are distributed normally.
Also in these tests, except for the case that slash does not work with Unbonded validator. I think there is no difference between the 3 stats together

x/superfluid/keeper/distribution_test.go
```diff
@@ -47,9 +47,20 @@ func (suite *KeeperTestSuite) TestMoveSuperfluidDelegationRewardToGauges() {
...
+		// There is no difference between Bonded, Unbonding, Unbonded
 		{
-			"add unbonded validator case",
-			[]stakingtypes.BondStatus{stakingtypes.Bonded, stakingtypes.Unbonded},
+			"add unbonded, unbonding validator case",
+			[]stakingtypes.BondStatus{stakingtypes.Bonded, stakingtypes.Unbonded, stakingtypes.Unbonding},
+			3,
+			[]superfluidDelegation{{0, 0, 0, 1000000}, {1, 1, 0, 1000000}, {2, 2, 0, 1000000}},
+			[]int64{0,1,2},
+			[]gaugeChecker{{0, 0, 0, true}, {1, 1, 0, true}, {2, 2, 0, true}},
```
czarcas7ic, line 58: Sorry I don't quite understand this. Why are we expecting a reward in the unbonding case here, but in the next test we do not expect a reward for the same unbonding case? I'm not saying you are wrong, it's just this is unclear to me.

x/superfluid/keeper/intermediary_account_test.go
```diff
@@ -26,6 +26,19 @@ func (suite *KeeperTestSuite) TestIntermediaryAccountCreation() {
 			2,
 			[]superfluidDelegation{{0, 0, 0, 1000000}, {1, 1, 0, 1000000}},
 		},
+		// Can create intermediary account with unbonded, unbonding validators
+		{
+			"test intermediary account with unbonded validator",
+			[]stakingtypes.BondStatus{stakingtypes.Bonded, stakingtypes.Unbonded},
+			2,
+			[]superfluidDelegation{{0, 0, 0, 1000000}, {1, 1, 0, 1000000}},
```
czarcas7ic, line 34: Is there a reason we allow intermediary account creation with unbonded validators? Seems odd to me, but I might be missing something.

x/superfluid/keeper/slash_test.go
```diff
@@ -25,6 +25,22 @@ func (suite *KeeperTestSuite) TestBeforeValidatorSlashed() {
...
+		},
+		{
+			"with an unbonded validator and single superfluid delegation",
+			[]stakingtypes.BondStatus{stakingtypes.Unbonded},
+			1,
+			[]superfluidDelegation{{0, 0, 0, 1000000}},
+			[]int64{0},
+			[]int64{0},
+		},
 		// {
```
czarcas7ic, line 44: Should we remove this dead code?

x/superfluid/keeper/intermediary_account_test.go
czarcas7ic: This is just an aside, no change required on this PR in regards to this!

x/superfluid/keeper/distribution_test.go
hieuvubk: In the next test, I want to show that If there is no reward to the validator, the gauge will not be distributed.

czarcas7ic: Ohh, the `rewardVals` field confused me, I now understand

czarcas7ic (approved):

mattverse (approved):
🌮

PR #2322, opened 2022-08-05
refactor/test: improve DecApproxEq, fix misuse in mint hooks, create osmoassert package

p0mvn:
Closes: #XXX

This change focuses on refactoring `DecApproxEq` helper functions.

Previously, it had to be called in the following way:
```
require.True(osmoutils.DecApproxEq(t, ...)
```

There is no reason for the verbosity when we pass in `*testing.T` as the first argument anyways.

Additionally, it is prone to misuse. Since this function takes `*testing.T` as the first argument, I assumed without reading the spec that this function must be asserting and misused it in mint tests. As a result, this change also fixes the mint hook tests.

Moreover, our assertions functions existed in `osmoutils` package that is used in core logic. We should not be mixing core logic with test helpers. As a result, `osmoassert` package was introduced and all relevant helpers moved there.

Lastly, `DecApproxEq` did not return the actual decimals if the assertion failed. So the error message was improved to help with debugging.

Now, `DecAppoxEq` can be called in the following way and handle assertions under the hood:
```
osmoassert.DecApproxEq(t, ...)
```

## Brief Changelog

- refactor `DecApproxEq`  to assert as part of its implementation
- refactor the failing / previously misused test cases in `x.mint`
- consolidate all assertions in the new `osmoassert` package

## Testing and Verifying


ValarDragon (approved):
Good call!

PR #2336, opened 2022-08-08
chore: Have our own proto build env image

faddat:
This will allow us to have a proto build env image of our own, automatically built when it needs to be. 

### DIFF
--- a/.github/workflows/proto-env.yml
+++ b/.github/workflows/proto-env.yml
@@ -1,0 +1,80 @@
+name: ghcr.io/osmosis-labs/proto-builder
+# Build & Push rebuilds the cosmos sdk development environment image every time the master branch is updated,
+# and pushes the image to https://ghcr.io/cosmos/cosmos-sdk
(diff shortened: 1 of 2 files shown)


ValarDragon:
can you explain in a github issue, more of the overarching reason of why we should do this? I don't really understand why we shouldn't use the process we do rn with the SDK image? (I don't have a great reason for it either - but this is a degree of context thats needed before this can be reviewed)

faddat:
I prefer to use the SDK image... just need to get it working.  

How about, if I don't get auto builds there working, then I make the issue.

If I do, I close this

mattverse:
just curious, does the current SDK image have problems or bugs?

faddat:
Not exactly.  I just haven't gotten it to work with ghcr.io just yet. 

Also, I'm pretty sure that they all have bugs, including the sdk one.  I have yet to see protobuf things work fully smoothly on any repo.

faddat:
I am going to close this, because the sdk reacted really oddly to go v1.19

Issue #2307, opened 2022-08-04
Updated outdated GenesisSupplyOffsets to match the sdk fork's bank genesis state constructor


nicolaslara:
## Background

The sdk fork is using `GenesisSupplyOffsets` in the `banktypes.NewGenesisState()` constructor. This is incompatible with the ibc-go testing package which expects the new signature (https://github.com/cosmos/ibc-go/tree/main/testing). 

## Suggested Design

~~Remove outdated GenesisSupplyOffsets from the sdk fork and make the necessary changes to both this repo and the sdk fork for the tests to pass~~

Add a new constructor to x/bank's `GenesisState` that uses offsets and remove the legacy parameter from the default constructor

## Acceptance Criteria

- [ ] Tests pass on both repos
- [ ] Someone that understands the genesis state has reviewed the changes (on both repos) to make sure they're ok

nicolaslara:
This branch has my changes to the osmosis repo for this: https://github.com/osmosis-labs/osmosis/tree/supply-offset-changes-on-sdk-fork

And this one has the changes on the sdk fork: https://github.com/osmosis-labs/cosmos-sdk/tree/remove-genesis-supply-offsets

Thanks to @xBalbinus for fixing one of the tests (https://github.com/osmosis-labs/osmosis/pull/2274) and for @p0mvn for pointing it out to me.

Issue #2305, opened 2022-08-04
IBC Rate Limiting


nicolaslara:
## Background

In the event of a chain exploit on osmosis or another chain connected via IBC, an attacker may want to quickly move the funds off osmosis (or into osmosis in the case of an infinite mint). To mitigate this, IBC channels should have a rate limiting capability so that no more than a pre-established amount can be moved through the channel

## Suggested Design

Create a cosmwasm contract that keeps track of the balance moved through the bridge during a period of time and allows or denies IBC messages. Add an IBC middleware that calls the contract and forwards the messages in the case of success or fails when rate limiting. 

## Acceptance Criteria
 - [ ] Test for the middleware are comprehensive and pass
 - [ ] Test for the contract are comprehensive and pass

hiromaily:
I have a question about why this contract is created by cosmwasm, not basic golang code.

ValarDragon:
So it can be upgraded without a chain upgrade!

Closing this as IBC rate limiting has gotten into v13

PR #2339, opened 2022-08-09
IBC Rate Limiting

nicolaslara:
Closes: ##2305

We want to add the feature of rate limiting sends on IBC channels to Osmosis. The intended goal, is to be able to cap the amount of certain assets that can flow into and out of Osmosis per unit time, to mitigate bridge risk. It should be the case that if there is some unforeseen bridge bug or infinite token mint, the max amount extractable out of Osmosis is bounded.

See issue #2305. This PR tracks the development of the IBC Rate Limiting functionality. 

## Brief Changelog

  - An IBC middleware is now provided and registered with the IBC transfer module. 
  - The level of rate limiting is controlled by a rate limiting contract that can be controlled by an address specified by governance (this could be the gov module itself)
 - The "governance address" of the rate limiting contract can reset the quota when a rate limit is hit  (this could be the gov module itself)

## Testing and Verifying

  - Added integration tests for the middleware (IBC Message <-> IBC transfer module <-> RateLimitingMiddlware <-> Relayer <-> Counterparty chain)
  - TBD: Added E2E tests for the middleware

 - This PR depends on https://github.com/osmosis-labs/osmosis/issues/2307, since that is necessary to use the ibc testing framework on Osmosis

TBD


app/app.go
```diff
@@ -453,3 +459,20 @@ func GetMaccPerms() map[string][]string {
...
+	return *app.AppKeepers.StakingKeeper
+}
+
+func (app *OsmosisApp) GetIBCKeeper() *ibckeeper.Keeper {
+	return app.AppKeepers.IBCKeeper
+}
+
+func (app *OsmosisApp) GetScopedIBCKeeper() capabilitykeeper.ScopedKeeper {
+	return app.AppKeepers.ScopedIBCKeeper
+}
```
ValarDragon, line 474: Lets move these to modules.go, we have to do something similar for account keeper

app/app.go
```diff
@@ -453,3 +459,20 @@ func GetMaccPerms() map[string][]string {
 
 	return dupMaccPerms
 }
+
+// Required for ibctesting
+func (app *OsmosisApp) GetStakingKeeper() stakingkeeper.Keeper {
+	return *app.AppKeepers.StakingKeeper
+}
```
ValarDragon, line 466: Yikes, this actually looks to me like a bug in ibctesting, that it depends on a keeper struct return arg, instead of an interface

ValarDragon: issue made https://github.com/cosmos/ibc-go/issues/1971

app/apptesting/events.go
```diff
@@ -15,3 +18,16 @@ func (s *KeeperTestHelper) AssertEventEmitted(ctx sdk.Context, eventTypeExpected
...
+	return events[index]
+}
+
+func (s *KeeperTestHelper) ExtractAttributes(event sdk.Event) map[string]string {
+	attrs := make(map[string]string)
+	for _, a := range event.Attributes {
+		attrs[string(a.Key)] = string(a.Value)
+	}
+	return attrs
+}
```
ValarDragon, line 33: :O nice

app/app.go
nicolaslara: done!

app/app.go
nicolaslara: yeah, it doesn't seem like reuse was the main priority in ibctesting, but it's easy enough to override stuff once you know where to look :)

nicolaslara:
Moved this out of `Draft` status. It's not `Ready` (still missing E2E tests, and there's one odd thing with the handling of the params that I still want to figure out), but already in a good place to get a review and some feedback.

tests/e2e/e2e_test.go
```diff
@@ -44,6 +44,32 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
 	chainB.SendIBC(chainA, chainA.NodeConfigs[0].PublicAddress, initialization.StakeToken)
 }
 
+func (s *IntegrationTestSuite) TestRateLimitingTestsSetupCorrectly() {
+	// Checking the rate limiting tests are setup correctly
+	//f1, err := ioutil.ReadFile("../../x/ibc-rate-limit/testdata/rate-limiter.wasm")
+	//s.NoError(err)
```
p0mvn, line 50: Suggest trying `s.Require().NoError(err)`

Without it, the test does not fail and continues.

It seems that there might be a path/permission issue at this step with:
```
open ../../x/ibc-rate-limit/testdata/rate-limiter.wasm: no such file or directory
```

nicolaslara: yes, sorry! that test was just me experimenting with making sure the file we're using for e2e tests (the one copied to the docket containers) is the same as the one in the integration tests. The path should be `/osmosis/rate-limiter.wasm`.

nicolaslara: Fixed. Moved these tests to the integration tests (since they don't depend on the e2e containerization). 

I also want to add a test to check the rust code compiles to the wasm file we're using for testing, but I'm not sure if the go tests is the right place to do this. Seems like something that should be its own task in the workflow (especially if more contracts are going to start being part of the source)

nicolaslara:
Splitting this into multiple PRs for review as requested by @ValarDragon 

Contract: https://github.com/osmosis-labs/osmosis/pull/2408
Middlware: https://github.com/osmosis-labs/osmosis/pull/2409 
Chain integration and tests: https://github.com/osmosis-labs/osmosis/pull/2410 

The last PR depends on the other two having been merged. The easiest way to review those changes is to compare it to a branch where the contract and middleware have been merged

tests/e2e/e2e_test.go
p0mvn: Awesome! I think a separate task in the workflow makes sense

x/ibc-rate-limit/testutil/chain.go
```diff
@@ -0,0 +1,96 @@
...
+	app.BeginBlock(abci.RequestBeginBlock{Header: header})
+	gInfo, res, err := app.Deliver(txCfg.TxEncoder(), tx)
+
+	app.EndBlock(abci.RequestEndBlock{})
+	app.Commit()
+
+	return gInfo, res, err
+}
+
+// Move epochs to the future to avoid issues with minting
```
mattverse, line 81: what's the issue when tokens get minted? Also curious if we don't need epoch within these testings at all

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,350 @@
...
+		suite.path.EndpointB.ChannelID,
+		sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(1)),
+		suite.chainB.SenderAccount.GetAddress().String(),
+		strings.Repeat("x", 4097),
+		clienttypes.NewHeight(0, 100),
+		0,
+	)
+	ack, _ := suite.ExecuteReceive(msg)
+	suite.Require().Contains(string(ack), "error",
+		"acknoledgment is not an error")
```
mattverse, line 123:
```suggestion
		"acknowledgment is not an error")
```

mattverse (commented):
Left some comments! Havent looked at the e2e tests, but other changes lgtm!

app/modules.go
```diff
@@ -245,3 +249,20 @@ func (app *OsmosisApp) GetAccountKeeper() simtypes.AccountKeeper {
 func (app *OsmosisApp) GetBankKeeper() simtypes.BankKeeper {
 	return app.AppKeepers.BankKeeper
 }
+
+// Required for ibctesting
+func (app *OsmosisApp) GetStakingKeeper() stakingkeeper.Keeper {
+	return *app.AppKeepers.StakingKeeper
```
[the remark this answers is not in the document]
nicolaslara: I think this is wrong. I'll fix it

app/keepers/keepers.go
```diff
@@ -440,6 +464,7 @@ func (appKeepers *AppKeepers) initParamsKeeper(appCodec codec.BinaryCodec, legac
 	paramsKeeper.Subspace(wasm.ModuleName)
 	paramsKeeper.Subspace(tokenfactorytypes.ModuleName)
 	paramsKeeper.Subspace(twaptypes.ModuleName)
+	paramsKeeper.Subspace(ibcratelimittypes.ModuleName)
```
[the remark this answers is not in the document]
nicolaslara: Added an issue for this: https://github.com/osmosis-labs/osmosis/issues/2869

x/ibc-rate-limit/contracts/rate-limiter/src/helpers.rs
```diff
@@ -46,8 +46,8 @@ pub mod tests {
         quota_name: &str,
         send_recv: (u32, u32),
         duration: u64,
-        inflow: u128,
-        outflow: u128,
+        inflow: Uint256,
```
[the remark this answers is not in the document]
nicolaslara: This is to deal with the case of a token having a very high precision: see this [comment](https://github.com/osmosis-labs/osmosis/pull/2408#discussion_r947581473)

x/ibc-rate-limit/ibc_module.go
```diff
@@ -103,12 +103,27 @@ func (im *IBCModule) OnChanCloseConfirm(
 	return im.app.OnChanCloseConfirm(ctx, portID, channelID)
 }
 
+func ValidateReceiverAddress(packet channeltypes.Packet) error {
+	var packetData transfertypes.FungibleTokenPacketData
+	if err := json.Unmarshal(packet.GetData(), &packetData); err != nil {
```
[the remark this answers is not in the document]
nicolaslara: yes

x/ibc-rate-limit/testutil/chain.go
nicolaslara: See this: https://github.com/osmosis-labs/osmosis/pull/2409#discussion_r952771827.

I  considered removing the epochs, but I think it was more work that this workaround. Since it's only for this test, I figured this was enough.

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,350 @@
...
+	var port, channel, accountFrom, accountTo string
+
+	if forward {
+		coins = sdk.NewCoin(sdk.DefaultBondDenom, amount)
+		port = suite.path.EndpointA.ChannelConfig.PortID
+		channel = suite.path.EndpointA.ChannelID
+		accountFrom = suite.chainA.SenderAccount.GetAddress().String()
+		accountTo = suite.chainB.SenderAccount.GetAddress().String()
+	} else {
+		coins = transfertypes.GetTransferCoin(
```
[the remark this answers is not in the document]
nicolaslara: I think we need this to test the receive case, but I may be wrong. I'll double check

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: ahh, I can never write acknowledgement right!  I'll update that in the error too so that the test doesn't fail

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,350 @@
...
+		channel,
+		coins,
+		accountFrom,
+		accountTo,
+		timeoutHeight,
+		0,
+	)
+}
+
+// Tests that a receiver address longer than 4096 is not accepted
```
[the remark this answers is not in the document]
nicolaslara: I wrote it this way cause I like the helpers to be close to where they're being used, but not a strong opinion, so I can move them

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,350 @@
...
+// Test rate limiting on receives
+func (suite *MiddlewareTestSuite) TestRecvTransferWithRateLimiting() {
+	// Setup contract
+	suite.chainA.StoreContractCode(&suite.Suite)
+	quotas := suite.BuildChannelQuota("weekly", 604800, 5, 5)
+	addr := suite.chainA.InstantiateContract(&suite.Suite, quotas)
+	suite.chainA.RegisterRateLimitingContract(addr)
+
+	// Setup receiver chain's quota
+	osmosisApp := suite.chainA.GetOsmosisApp()
```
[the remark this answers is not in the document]
nicolaslara: It doesn't. That comment refers to the block bellow. I'll move it

app/modules.go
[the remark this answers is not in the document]
nicolaslara: It wasn't. The underlying `app.AppKeepers.StakingKeeper` is a pointer, so it needs to be dereferenced. The signature is what's expected by ibctesting

Issue #2087, opened 2022-07-15
test(gov): manually / e2e test expedited proposals


p0mvn:
## Background

We need to manually and/or e2e test expedited proposals. This issue is blocked on me providing a description about the plan
## Suggested Design

TBD

## Acceptance Criteria
- verified that expedited proposals work as expected and can be released

p0mvn:
This issue is blocked by the parameter migrations in: https://github.com/osmosis-labs/osmosis/issues/2249

czarcas7ic:
Issue unblocked, I am now working on this

PR #2341, opened 2022-08-09
e2e: expedited proposal test

czarcas7ic:
Closes: #2087 

This pull request adds the ability to test if the newly introduced expedited proposals work as intended

## Brief Changelog

- Fixes small Makefile bug I introduced in a previous PR
- Updates SDK fork to 0.45.0x-osmo-v12-alpha.1 which includes [this PR](https://github.com/osmosis-labs/cosmos-sdk/pull/303)
- Adds `ExpeditedVotingPeriod` variable
- Adds `isExpedited` arg to TextProps and Deposits
- Adds `QueryPropStatus` function to query a proposal's status (passed, rejected, etc)
  - Adds a helper function `QueryPropStatusTimed` that time's how long it takes to get to a desired proposal status
- Fixes bug in [configurer/current.go](https://github.com/osmosis-labs/osmosis/pull/2341/files#diff-fc3948505a8cdd85079639e3a9d59393dce6c00a48e2a040522bd70028bcf607) that wouldn't take votingPeriod properly
- Comments out further changes [tests/e2e/containers/containers.go](https://github.com/osmosis-labs/osmosis/pull/2341/files#diff-5f119aa77c3b92f785aace7eab66b696056f43373dec9857f0ad466cdd58324e) that will be needed once v12 is introduced
- Adds `TestExpeditedProposals` that holds the ultimate expedited proposal logic

## Testing and Verifying

CI will NOT properly test this. As discussed with @p0mvn, we only run expedited proposal tests if upgrade tests are disabled until v12 is released due to compatibility issues. This can still be run locally though with upgrades disabled to show that expedited proposals work as intended!


czarcas7ic:
This should be R4R. It's unclear to me why gotest are failing. The only sdk diff between main and this are three commits, none of them having to do with mint 

https://github.com/osmosis-labs/cosmos-sdk/compare/0.45.0x-osmo-v11-alpha.5...v0.45.0x-osmo-v12-alpha.1

czarcas7ic:
Specifically it's this line here https://github.com/osmosis-labs/osmosis/blob/50b81179cb87dab5161c2e83e4ae67e5fb7e7a1d/x/mint/keeper/keeper_test.go#L354

czarcas7ic:
Ah I see what it is, we now return 0 for `GetSupplyWithOffset` instead of a negative number. I will change this test to reflect

tests/e2e/configurer/chain/commands.go
```diff
@@ -105,3 +114,26 @@ func (n *NodeConfig) CreateWallet(walletName string) string {
...
+		n.t,
+		func() bool {
+			status, err := n.QueryPropStatus(proposalNumber)
+			if err != nil {
+				return false
+			}
+			if status != desiredStatus {
+				return false
+			}
+			return true
```
alexanderbez, line 131:
```suggestion

			return status == desiredStatus
```

alexanderbez (approved):
Excellent work 👏

czarcas7ic:
This is blocked on a consensus in regards to this sdk PR: https://github.com/osmosis-labs/cosmos-sdk/pull/308

czarcas7ic:
@alexanderbez thank you!

@p0mvn addressed your comment from the sdk PR, agreed that setting the test offset to 1 is the best way. thanks for talking through that with me!

Now unblocked 😄

tests/e2e/configurer/current.go
```diff
@@ -44,7 +45,16 @@ func (cb *CurrentBranchConfigurer) ConfigureChain(chainConfig *chain.Config) err
 		return err
 	}
 	cb.t.Logf("temp directory for chain-id %v: %v", chainConfig.Id, tmpDir)
-	initializedChain, err := initialization.InitChain(chainConfig.Id, tmpDir, chainConfig.ValidatorInitConfigs, time.Duration(chainConfig.VotingPeriod), 0)
+
+	numVal := float32(len(chainConfig.ValidatorInitConfigs))
+
+	chainConfig.VotingPeriod = config.PropDepositBlocks + numVal*config.PropVoteBlocks + config.PropBufferBlocks
+	chainConfig.ExpeditedVotingPeriod = config.PropDepositBlocks + numVal*config.PropVoteBlocks + config.PropBufferBlocks - 1
+
```
p0mvn, line 53: Seems that we do a similar calculation in `upgrade.go`. Can we encapsulate this into a shared function or a method defined on the chain config?

tests/e2e/containers/containers.go
```diff
@@ -212,8 +212,10 @@ func (m *Manager) RunNodeResource(chainId string, containerName, valCondifDir st
 // The genesis and configs are to be mounted on the init container as volume on mountDir path.
 // Returns the container resource and error if any. This method does not Purge the container. The caller
 // must deal with removing the resource.
-func (m *Manager) RunChainInitResource(chainId string, chainVotingPeriod int, validatorConfigBytes []byte, mountDir string, forkHeight int) (*dockertest.Resource, error) {
+func (m *Manager) RunChainInitResource(chainId string, chainVotingPeriod, chainExpeditedVotingPeriod int, validatorConfigBytes []byte, mountDir string, forkHeight int) (*dockertest.Resource, error) {
 	votingPeriodDuration := time.Duration(chainVotingPeriod * 1000000000)
+	// TODO: Uncomment this after v12 release
```
p0mvn, line 217: Let's make an issue for this, please

tests/e2e/e2e_test.go
```diff
@@ -211,3 +212,33 @@ func (s *IntegrationTestSuite) TestStateSync() {
 	err = chain.RemoveNode(stateSynchingNode.Name)
 	s.NoError(err)
 }
+
+func (s *IntegrationTestSuite) TestExpeditedProposals() {
+	if !s.skipUpgrade {
+		s.T().Skip("this can be re-enabled post v12")
```
p0mvn, line 218: Let's make an issue and link it here with a comment please

x/mint/keeper/keeper_test.go
```diff
@@ -336,6 +336,12 @@ func (suite *KeeperTestSuite) TestSetInitialSupplyOffsetDuringMigration() {
 			bankKeeper := suite.App.BankKeeper
 			mintKeeper := suite.App.MintKeeper
 
+			// in order to ensure the offset is correctly calculated, we need to mint the supply + 1
+			// this is because a negative supply offset will always return zero
+			// by setting this to the supply + 1, we ensure we are correctly calculating the offset by keeping it delta positive
+			supply := bankKeeper.GetSupply(ctx, sdk.DefaultBondDenom)
+			mintKeeper.MintCoins(ctx, sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, supply.Amount.Add(sdk.NewInt(1)))))
```
p0mvn, line 343: I think we should instead mint the amount that is equal to `keeper.DeveloperVestingAmount`. While these are equal with supply, it might be less apparent why we are using supply here. 

The ultimate reason for having the offset in the first place is having developer vesting rewards so using the amount related to it might be more obvious in terms of why this is done

tests/e2e/e2e_test.go
```diff
@@ -211,3 +212,33 @@ func (s *IntegrationTestSuite) TestStateSync() {
...
+	}
+
+	var wg sync.WaitGroup
```
p0mvn, line 221: What is the purpose of using both a channel and a wait group? Waitgroup seems redundant since we would block on a channel until it gives us a value anyways

tests/e2e/e2e_test.go
```diff
@@ -211,3 +212,33 @@ func (s *IntegrationTestSuite) TestStateSync() {
...
+	chain := s.configurer.GetChainConfig(0)
+	node, err := chain.GetDefaultNode()
+	s.NoError(err)
+
+	node.SubmitTextProposal("expedited text proposal", sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(config.InitialMinExpeditedDeposit)), true)
+	chain.LatestProposalNumber += 1
+	node.DepositProposal(chain.LatestProposalNumber, true)
+	wg.Add(1)
+	totalTime := make(chan time.Duration, 1)
+	go node.QueryPropStatusTimed(&wg, chain.LatestProposalNumber, "PROPOSAL_STATUS_PASSED", totalTime)
```
p0mvn, line 231: nit: to make it clearer
```suggestion
	go node.QueryPropStatusTimed(&wg, chain.LatestProposalNumber, "PROPOSAL_STATUS_PASSED", totalTimeChan)
```

tests/e2e/configurer/chain/commands.go
```diff
@@ -105,3 +114,24 @@ func (n *NodeConfig) CreateWallet(walletName string) string {
 	n.LogActionF("created wallet %s, waller address - %s", walletName, walletAddr)
 	return walletAddr
 }
+
+func (n *NodeConfig) QueryPropStatusTimed(wg *sync.WaitGroup, proposalNumber int, desiredStatus string, totalTime chan time.Duration) {
+	defer wg.Done()
```
p0mvn, line 119: I think we can remove the waitgroup and instead send `1*time.Minute` on the `totalTime` channel since that's how long it takes at most for this query to wait. I think that would allow for the caller to get unblocked and continue. On the other hand, currently if `require.Eventually(...)` fails, the caller is blocked forever wiaitng on the result from `totalTime` channel that never arrives

tests/e2e/e2e_test.go
```diff
@@ -211,3 +212,33 @@ func (s *IntegrationTestSuite) TestStateSync() {
 	err = chain.RemoveNode(stateSynchingNode.Name)
 	s.NoError(err)
 }
+
+func (s *IntegrationTestSuite) TestExpeditedProposals() {
```
p0mvn, line 216: Is it possible to set the timeout specifically for this test case? Just in case it gets blocked due to multi-goroutine handling here

tests/e2e/e2e_test.go
```diff
@@ -211,3 +212,33 @@ func (s *IntegrationTestSuite) TestStateSync() {
...
+		node.VoteYesProposal(initialization.ValidatorWalletName, chain.LatestProposalNumber)
+	}
+	// wait till prop status reaches desired pass status
+	wg.Wait()
+	elapsed := <-totalTime
+	// compare the time it took to reach pass status to expected expedited voting period
+	expeditedVotingPeriodDuration := time.Duration(chain.ExpeditedVotingPeriod * 1000000000)
+	timeDelta := elapsed - expeditedVotingPeriodDuration
+	// ensure delta is within one second of expected time
+	s.Require().Less(timeDelta, time.Second)
```
p0mvn, line 242: Can we have `time.Second` as a constant that we also use to set the parameter on chain? Just to make the link clear between how we configure it and the assertion here

p0mvn (commented):
Great work @czarcas7ic . I had some minor comments/thoughts but the solution looks great overall. Please let me know what you think

tests/e2e/containers/containers.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/2386, I can likely take this post v12

tests/e2e/e2e_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/2386, included this in the same issue as above, will likely take this post v12 since I have context

x/mint/keeper/keeper_test.go
czarcas7ic: This is a really good catch, thank you!

tests/e2e/e2e_test.go
czarcas7ic: I actually wasn't aware that channels block until a value is returned. I tested it just to make sure and that is the case, thanks!

tests/e2e/e2e_test.go
czarcas7ic: I don't think we would want this to be a set parameter right? I am basically just saying that the expeditedVotingPeriodDuration that this go routine calculates should realistically be within a second of the actual on chain param. I might be misunderstanding your recommendation here though

tests/e2e/e2e_test.go
czarcas7ic: There is a one minute timeout within the function the go routine is calling. Should I have a second timeout here as well? Would that be redundant or is it a good practice?

czarcas7ic (commented):
Thanks so much for the in depth review Roman! Only had a couple questions, everything else was clear!

tests/e2e/configurer/chain/commands.go
```diff
@@ -105,3 +113,23 @@ func (n *NodeConfig) CreateWallet(walletName string) string {
...
+			}
+
+			return status == desiredStatus
+		},
+		1*time.Minute,
+		10*time.Millisecond,
+		"Osmosis node failed to retrieve prop tally",
+	)
+	elapsed := time.Since(start)
+	totalTime <- elapsed
```
p0mvn, line 134: This is related to an earlier comment:
https://github.com/osmosis-labs/osmosis/pull/2341/files#r944726827

Currently, the problem, is that we can fail after trying for 1 minute. However, nothin is sent on the `totalTime` channel if the failure does occur.

As a result, the test can get blocked for a long period of time (I think 30 minutes max)

I think we have 2 options to mitigate that and fail immediately:
1. Create a second timeout specifically for that test case (e.g. 2 minute timeout). So that if we block on the channel, the test case fails early

2. have a `defer` statement that either sends `elapsed` on the channel or a dummy value (i.e. -1)signifying failure.

tests/e2e/e2e_test.go
p0mvn: Sorry my comment ended up being confusing.

What I meant was let's have a global constant defined for `expeditedVotingPeriodDuration`, `votingPeriodDuration`.

And also reuse those constants here:
https://github.com/osmosis-labs/osmosis/blob/6c65ea106b402245788f1390dbdfdbc9229fdfa6/tests/e2e/configurer/current.go#L48

Basically, the goal is to avoid hardcoding these values in different places. If we decide to change them in the future, we might update in one place but forget about another place and start having weird issues.

Having shared constants defined for such things helps mitigate these problems

p0mvn: Actually, nvm I think that's already the case since we are using `chainConfig.ExpeditedVotingPeriod`

p0mvn: The `*1000000000` part misled me. I'm going to resolve this, sorry for the confusion

tests/e2e/e2e_test.go
p0mvn: Replied here: https://github.com/osmosis-labs/osmosis/pull/2341/files#r944758494

tests/e2e/configurer/chain/commands.go
czarcas7ic: I went back and forth between your two suggestion and decided to implement it through a case timeout (similar to the recommendation from Go by Example [here](https://gobyexample.com/timeouts))

I added a sleep to the timed go routine to ensure it stopped at the desired time and it did! I think this was the simplest implementation possible. Lmk what you think

tests/e2e/e2e_test.go
```diff
@@ -214,3 +215,39 @@ func (s *IntegrationTestSuite) TestStateSync() {
...
+		node.VoteYesProposal(initialization.ValidatorWalletName, chain.LatestProposalNumber)
+	}
+	// if querying proposal takes longer than timeoutPeriod, stop the goroutine and error
+	var elapsed time.Duration
+	timeoutPeriod := time.Duration(2 * time.Minute)
+	select {
+	case elapsed = <-totalTimeChan:
+	case <-time.After(timeoutPeriod):
+		err := fmt.Errorf("go routine took longer than %s", timeoutPeriod)
+		s.NoError(err)
```
p0mvn, line 243:
```suggestion
		s.Require().NoError(err)
```

p0mvn (approved):
Nice! Great job

Issue #2098, opened 2022-07-17
Fork cosmwasm to re-enable queries


ValarDragon:
## Background

For our next release, I believe that we want to fork wasmd to re-enable stargate queries. We eventually want to move to a whitelist approach, so we can control which queries were making state machine compatability guarantees about. Until they are added back, it must be part of contract review to ensure that all approved contract queries are fine. 

To be more precise:
* A query that is exposed should not have any fields deleted, otherwise it will be state machine breaking
* When gas tracking is added, it should not be changed within a minor version.

And we should have a beaker test doing this.

As part of this, we should also scope how hard would it be to build a whitelist. (hopefully not much, seeing as we already have a string description of the query type)

## Suggested Design

We need to investigate how stargate queries were removed, so that we can add them back.

## Acceptance Criteria

- [ ] It is possible to use stargate queries in forked cosmwasm.

alexanderbez:
The queries should be deterministic too, i.e. no map encoding or anything like that. And if they do, ensure that the output is deterministic.

mattverse:
https://github.com/osmosis-labs/osmosis/pull/2190

mattverse:
https://github.com/osmosis-labs/osmosis/pull/2353

PR #2353, opened 2022-08-10
RE: Enable Whitelist Stargate Query

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/pull/2190
Closes: https://github.com/osmosis-labs/osmosis/pull/2098

Add whitelisted stargate queries and ensure deterministic query. The difference of this PR with the previous PR(#2190) is that we json marshal the responses, not proto marshalling them. Test cases have been also been updated for this accordingly, tested and confirmed that this way of enabling stargate queries work via sample contract (https://github.com/osmosis-labs/swaprouter)

## Brief Changelog

- Enable Stargate Query

## Testing and Verifying

Tested locally and confrimed that this way of whitelisting works contract side by deploying contract with this commit.  
Also added go test to ensure we're safe with json marshalling.


wasmbinding/query_plugin.go
```diff
@@ -138,6 +173,32 @@ func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessag
...
+		return nil, wasmvmtypes.Unknown{}
+	}
+
+	// unmarshal binary into stargate response data structure
+	err := proto.Unmarshal(bz, message)
+	if err != nil {
+		return nil, wasmvmtypes.Unknown{}
+	}
+
+	bz, err = codec.MarshalJSON(message)
```
mattverse, line 191: We json marshal the response here for the contracts because stargate querier json unmarshals the response by default

nicolaslara: Just to see if I understand this correctly: this method converts the encoding from protobuf (received from the `route` call above) to json, which is what wasmd expects. Is that correct? If so, I'd like to suggest renaming this method from `normalize` to something like `protoToJson`

mattverse: makes sense! Thanks for the recommendation :) fixed!

wasmbinding/stargate_whitelist_test.go
```diff
@@ -0,0 +1,114 @@
...
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		suite.Run(fmt.Sprintf("Case %s", tc.name), func() {
+			suite.SetupTest()
+
+			binding, ok := wasmbinding.StargateWhitelist.Load(tc.queryPath)
```
p0mvn, line 92: I think we can have a separate fixture (`TestStargateWhitelistLoad`) just for this where we can validate that all of the queries we expect are loaded correctly + have a test case where a query path fails because it is not whitelisted

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,47 @@ import (
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	proto "github.com/gogo/protobuf/proto"
+	abci "github.com/tendermint/tendermint/abci/types"
 
 	"github.com/osmosis-labs/osmosis/v11/wasmbinding/bindings"
 )
 
+// StargateQuerier dispatches whitelisted stargate queries
+func StargateQuerier(queryRouter baseapp.GRPCQueryRouter, codec codec.Codec) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
```
p0mvn, line 19: Is it possible to have test cases for this function directly?

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,47 @@ import (
...
+	return func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
+		binding, whitelisted := StargateWhitelist.Load(request.Path)
```
p0mvn, line 21: Is it correct to call it binding? From my understanding, bindings are distinct from stargate queries. Does it make sense to change this to something like `queryProtoResponse`?

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,47 @@ import (
...
+
+		res, err := route(ctx, abci.RequestQuery{
+			Data: request.Data,
+			Path: request.Path,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		// normalize response to ensure backward compatibility
```
p0mvn, line 39: I think the comment might be obsolete

wasmbinding/query_plugin.go
```diff
@@ -138,6 +173,32 @@ func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessag
 	}
 }
 
+// NormalizeReponses normalizes the responses by unmarshalling the response then marshalling them again.
```
p0mvn, line 176: The comment seems to be outdated. Could you please explain why we need this conversion in the spec?

wasmbinding/stargate_whitelist_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			originVersionBz, err := hex.DecodeString(tc.originalResponse)
+			suite.Require().NoError(err)
+			jsonMarshalledOriginalBz, err := wasmbinding.ConvertProtoToJsonMarshal(binding, originVersionBz, suite.app.AppCodec())
+			suite.Require().NoError(err)
+
+			newVersionBz, err := hex.DecodeString(tc.updatedResponse)
+			suite.Require().NoError(err)
+			jsonMarshalledUpdatedBz, err := wasmbinding.ConvertProtoToJsonMarshal(binding, newVersionBz, suite.app.AppCodec())
+			suite.Require().NoError(err)
```
p0mvn, line 103: Do you think it makes sense to split "origin response" and "new version response" into 2 separate test cases?

I feel like we are testing several things in one test case

p0mvn (commented):
Overall looks good, nice work!

Could you please elaborate in comments/specs on the following things:
- why do we need to proto unmarshal and then json marshal?
- why do we need to `message.Reset()`?

I also think that we might consider splitting up some test cases and testing `StargateQuerier` directly

p0mvn:
Why do we redefine `StargateQuerier` in https://github.com/osmosis-labs/wasmd/pull/7?

Shouldn't we encourage clients to define their own?

nicolaslara:
> Why do we redefine StargateQuerier in https://github.com/osmosis-labs/wasmd/pull/7?
> Shouldn't we encourage clients to define their own?

Having clients define their own was the previews solution. Upstreaming the whitelist means that very few clients will need to define their won (they can just whitelist endpoints instead)

mattverse:
> > Why do we redefine StargateQuerier in [osmosis-labs/wasmd#7](https://github.com/osmosis-labs/wasmd/pull/7)?
> > Shouldn't we encourage clients to define their own?
> 
> Having clients define their own was the previews solution. Upstreaming the whitelist means that very few clients will need to define their won (they can just whitelist endpoints instead)

Yeah also having clients define their own was too much overhead, almost impossible for clients to use.

wasmbinding/query_plugin.go
mattverse: yeah agreed, changed to protoResponse!

wasmbinding/query_plugin.go
mattverse: Good catch! Updated the comment with reasons why this has to change

wasmbinding/stargate_whitelist_test.go
mattverse: Sure! I can add a test case for testing cases that hasnt been whitelisted.

mattverse: wdym by 

> where we can validate that all of the queries we expect are loaded correctly

Looks kind of hard to me to validate through all queries by directly calling them

wasmbinding/stargate_whitelist_test.go
mattverse: sorry if I missed something, why / what do you think is getting tested in this single test case?
It seems to me that we're only testing if updated proto still returns same json marshalled bytes

wasmbinding/query_plugin.go
mattverse: Added test case! Please take a look!

p0mvn:
Sounds good, makes sense @mattverse @nicolaslara . Then, why are we defining our own instead of merging the generic `StargateQuerier` in the `wasmd` fork, updating the `go.mod` and reusing it here given that our goal is to eventually upstream it?

Should we then completely move all the `StargateQuerier` logic to the wasmd PR: https://github.com/osmosis-labs/wasmd/pull/7

wasmbinding/query_plugin_test.go
```diff
@@ -0,0 +1,90 @@
...
+			requestData: func() []byte {
+				epochrequest := epochtypes.QueryEpochsInfoRequest{}
+				bz, err := proto.Marshal(&epochrequest)
+				suite.Require().NoError(err)
+				return bz
+			},
+			responseProtoStruct: &epochtypes.QueryEpochsInfoResponse{},
+		},
+		{
+			name: "unregsitered path(not whitelisted)",
```
p0mvn, line 36:
```suggestion
			name: "unregistered path(not whitelisted)",
```

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+func StargateQuerier(queryRouter baseapp.GRPCQueryRouter, codec codec.Codec) func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
+	return func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
+		protoResponse, whitelisted := StargateWhitelist.Load(request.Path)
+		if !whitelisted {
+			return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("'%s' path is not allowed from the contract", request.Path)}
+		}
+
+		route := queryRouter.Route(request.Path)
+		if route == nil {
+			return nil, wasmvmtypes.UnsupportedRequest{Kind: fmt.Sprintf("No route to query '%s'", request.Path)}
```
p0mvn, line 28: Is it possible to add a test case for this branch?

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+		}
+
+		res, err := route(ctx, abci.RequestQuery{
+			Data: request.Data,
+			Path: request.Path,
+		})
+		if err != nil {
+			return nil, err
```
p0mvn, line 36: Ditto for test case for this branch

wasmbinding/query_plugin.go
```diff
@@ -5,12 +5,46 @@ import (
...
+		}
+
+		bz, err := ConvertProtoToJsonMarshal(protoResponse, res.Value, codec)
+		if err != nil {
+			return nil, err
```
p0mvn, line 41: ditto for test case

Don't mean to be annoying here with such attention to detail but I feel like we should aim to maximize test coverage for the core route in cosmwasm queries

wasmbinding/stargate_whitelist_test.go
p0mvn: What I was thinking is to have a test case for each of the queries we have whitelisted right now:
https://github.com/osmosis-labs/osmosis/blob/204f20ad1ee487bdd723121264ce71783ddf43c6/wasmbinding/stargate_whitelist.go#L20-L23

+ 1 that is not whitelisted and expect that after `SetupTest()`,
```
binding, ok := wasmbinding.StargateWhitelist.Load(tc.queryPath)
```
ok is false

I'm not sure of all the limitations. Is there anything specific blocking us from doing this in a separate new test fixture `TestStargateWhitelistLoad`?

wasmbinding/stargate_whitelist_test.go
p0mvn: No probs at all, I could have been more clear. What I was thinking is to change:
```
originVersionBz, err := hex.DecodeString(tc.originalResponse)
suite.Require().NoError(err)
jsonMarshalledOriginalBz, err := wasmbinding.ConvertProtoToJsonMarshal(binding, originVersionBz, suite.app.AppCodec())
suite.Require().NoError(err)

newVersionBz, err := hex.DecodeString(tc.updatedResponse)
suite.Require().NoError(err)
jsonMarshalledUpdatedBz, err := wasmbinding.ConvertProtoToJsonMarshal(binding, newVersionBz, suite.app.AppCodec())
suite.Require().NoError(err)
```

to

```
decodedBuf, err := hex.DecodeString(tc.response)
suite.Require().NoError(err)
jsonMarshalledUpdatedBz, err := wasmbinding.ConvertProtoToJsonMarshal(binding, decodedBuf, suite.app.AppCodec())
suite.Require().NoError(err)
```

So that we have 2 separate test cases:
```
{
	"Query All Balances - Old Version",
	"0a090a036261721202333012050a03666f6f",
	"/cosmos.bank.v1beta1.Query/AllBalances",
	&banktypes.QueryAllBalancesResponse{
		Balances: sdk.NewCoins(sdk.NewCoin("bar", sdk.NewInt(30))),
		Pagination: &query.PageResponse{
			NextKey: []byte("foo"),
		},
	},
},
{
	"Query All Balances - New Version",
"0a090a036261721202333012050a03666f6f1a2d636f736d6f73316a366a357473717571326a6c77326166376c3378656b796171377a67346c386a737566753738",
	"/cosmos.bank.v1beta1.Query/AllBalances",
	&banktypes.QueryAllBalancesResponse{
		Balances: sdk.NewCoins(sdk.NewCoin("bar", sdk.NewInt(30))),
		Pagination: &query.PageResponse{
			NextKey: []byte("foo"),
		},
	},
},
```

Just to be clear, this is an optional comment. It just felt that we can split this one case into two to keep it more focused on the system under test (`wasmbinding.ConvertProtoToJsonMarshal`)

wasmbinding/stargate_whitelist_test.go
```diff
@@ -0,0 +1,114 @@
...
+func (suite *StargateTestSuite) SetupTest() {
+	suite.app = app.Setup(false)
+	suite.ctx = suite.app.BaseApp.NewContext(false, tmproto.Header{Height: 1, ChainID: "osmosis-1", Time: time.Now().UTC()})
+}
+
+func TestStargateTestSuite(t *testing.T) {
+	suite.Run(t, new(StargateTestSuite))
+}
+
+func (suite *StargateTestSuite) TestDeterministicJsonMarshal() {
```
p0mvn, line 37: Looking at this again - I think we should also consider testing invalid inputs, not only happy path

wasmbinding/stargate_whitelist_test.go
mattverse: For the test case that is not whitelisted, it is tested in stargate queier. 
The reason I think testing the whitelists individually is not a sustainable test case is because we now only have 3 whitelist to test, which is not something hard to hard code, but future plans are to add 50+ whitelists, which would then be meaningless to test individually by hard coding the paths!

wasmbinding/query_plugin.go
mattverse: Added!

iboss-ptk:
On testing, do we have list of known causes for non-determinism? Notice that there seems to be only schema upgrade test that address non-determinism issue.

If we have the list, then we might be able to create a generic test scenario to test against all enabled stargate query and construct test from `func (*Map) Range` to avoid human error of forgetting to add test when there is new whitelist.

Other point is what I pointed out to @mattverse few weeks back about the potential of making whitelist a config file of some sort. Is that possible? Will need that to filter out CosmWasm Querier generation, but if it's too tricky, I could make code gen call `StargateWhitelist` directly through ffi as well.

nicolaslara:
> On testing, do we have list of known causes for non-determinism? 

I don't think there is a comprehensive list, and there could always be new, previously unknown causes of non-determinism that pop-up. The main culprits at the moment are returning values based on: explicit randomness, any sort of unordered lists, system time, floats, and maps. 

> generic test scenario against all enabled stargate query

I think doing these generic tests for the causes of non-determinism would be quite difficult (they're all quite unique), but great if achievable! maybe some sort of simulation with different underlying architectures/setups for the nodes? Could be done with e2e testing, but I'd say it's outside the scope of this PR.

wasmbinding/query_plugin.go
mattverse: this is the only test case branch I disagree on adding tests for now, we error catch here just because, it's hard to  manipulate so that we break queryRouter from SDK within this test case

wasmbinding/stargate_whitelist_test.go
mattverse: Oh, I see what you mean. The main purpose of this test is not to `ConvertProtoToJsonMaral` but to test and ensure that stargate queries return the same, deterministic response, even though we have an update for the proto struct used in the state machine.

I added a separate test case for ConvertProtoToJsonMarshal while adding more comment on this test function to avoid future confusion

wasmbinding/stargate_whitelist_test.go
mattverse: Added test case!

mattverse:
Added way more test cases upon @p0mvn 's review!

wasmbinding/query_plugin.go
p0mvn: Is it not possible to break it by providing an invalid `request.Data` for a route?

p0mvn:
> Sounds good, makes sense @mattverse @nicolaslara . Then, why are we defining our own instead of merging the generic `StargateQuerier` in the `wasmd` fork, updating the `go.mod` and reusing it here given that our goal is to eventually upstream it?
> 
> Should we then completely move all the `StargateQuerier` logic to the wasmd PR: [osmosis-labs/wasmd#7](https://github.com/osmosis-labs/wasmd/pull/7)

Synched with Matt. The goal is to eventually remove `StargateQuerier` from here, once it is upstreamed to wasmd. For now, we want to have it for the speed of development since upstreaming to wasmd might take some time

p0mvn (approved):
LGTM! Great work @mattverse . Thanks for testing this thoroughly

wasmbinding/query_plugin.go
mattverse: providing invalid request.Data route would break the next error catching case, not this one :(

mattverse: I did find a way to break this, added in the last commit

iboss-ptk (approved):
LGTM! Great work @mattverse

ValarDragon (approved):

PR #2355, opened 2022-08-10
Move twap to its own top level module

ValarDragon:

As pointed about by @p0mvn somewhere, I forget where, its inconsistent that TWAP isn't its own module, as it implements the module interface and has its own keeper for state storage.

## Brief Changelog

- Move TWAP to its own top level keeper

## Testing and Verifying


p0mvn (approved):

mattverse:
in this case, what do folks think about following the same struct for the twap module?

Issue #2263, opened 2022-07-31
GRPC query refactor


ValarDragon:
## Background

Right now, the grpc query interface we have seems untenable to me:

- Mixes generic query logic and grpc-specific query logic
- Grpc query test logic is super ad hoc / mostly unnecessary looking
- Too much direct store handling
- Mostly boilerplate that could just get generated

This manifests in our codebase as problems with:
- Bad golang client interfaces (see every simulation PR, and all CLI code)
- Mixing of responsibilities at the grpc_query file logic, and therefore bad tests
- Increased complexity of Keeper packages

## Suggested Design

- [ ] Move all grpc_query logic into `client/grpc` in every module, and these should only use generic keeper methods
- [ ] Have stable, simple methods for queries in the keeper package
  - A measurable outcome of this is that every packages simulation package should have **0** `store` package references.
- [ ] Make a YML descriptor file for query logic that we codegen the entire `client/grpc` folder from
  - All the logic should really be boilerplate, of "run Keeper function, run through a pagination function, package into proto response"\
  - This is maybe excessive, should be explored more. 

## Acceptance Criteria

- [ ] No grpc_query files in keeper package
- [ ] Simulation message creation no longer needs to directly access the store
- [ ] Some understanding of how to reduce overhead of our grpc client creation.

vuong177:
Hi @ValarDragon is this being worked on? Can I try doing this task ?

ValarDragon:
Good idea, though I think we may want to do this via AUtoCLI. Going to close for now

PR #2357, opened 2022-08-10
Add first query for twap, try alternate code structure

ValarDragon:
Cref: #2263 

This PR adds a query for TWAP, but more importantly tries out a new code structure for queries:
* All GRPC logic in `x/{module}/client/grpc` package
* Trying out a `query.yml` that we can use to codegen the grpc_query logic & in the future CLI code off the same logic.
  * grpc_logic now code gen'd!

Also because of the decision to make keeper logic top level (which is imo much more readable), once you add queries you have to move module.go to its own package, so this PR does that.

What do folks think of this? I feel like:
- All GRPC logic in `x/{module}/client/grpc` package` is a win
- I'm unsure about the current query.yml form. How I imagine this working eventually is:
  1) We have a unified set of proto structs for all clients (right now CLI & grpc).
  2) Every client maps their input, into this proto input (also deriving some of how they get input from the proto itself)
  3) We have a unified function that goes from proto input -> direct keeper method -> proto output that every client calls
  4) Each client then displays / returns the proto output however they see fit.
   
So step (2) is totally code-gennable for grpc. This PR currently codegens (2) for GRPC, and mocks (3). We could imagine (3) exists as hand written functions in `x/client/` along with proto types, or being code generated. 

Step (4) is entirely automated out of grpc already, we should ideally do the same for CLI

## Brief Changelog

- Add GRPC query
- Add tooling to codegenerate query code

## Testing and Verifying

Does not add tests for queries, the logic is tested at other layers. Just needs integration tests, but to what degree we do that will depend on how we want the code structure.

  - How is the feature or change documented? not documented atm


x/twap/twapmodule/module.go
```diff
@@ -15,6 +16,10 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 
+	"github.com/osmosis-labs/osmosis/v10/x/twap"
+	twapclient "github.com/osmosis-labs/osmosis/v10/x/twap/client"
```
p0mvn, line 20: nit: why would we want to rename the default 'client` to `twapclient` when there is no clash in names?

proto/osmosis/twap/v1beta1/query.yml
```diff
@@ -0,0 +1,8 @@
+keeper: 
+- path: "github.com/osmosis-labs/osmosis/v10/x/twap"
+- struct: "Keeper"
+GetArithmeticTwap:
+- default_values:
+  - Req.end_time: "time.Now()"
+- func: "twap, err := k.GetArithmeticTwap(Req.pool_id, Req.base_asset, Req.quote_asset, Req.start_time, Req.end_time)"
+- response: "GetArithmeticTwapResponse{twap}"
```
p0mvn, line 8: I'm assuming that this is just an idea template and is not actually used for code gen yet.

If so, just a thought - we seem to repeat `GetArithmeticTwap` several times. For requests, responses, and the function name we could reuse the `GetArithmeticTwap` from a single place on line 4.

I think it could be useful to keep the configuration as simple as possible and minimize repetitions

p0mvn (approved):
LGTM. Left a thought on the code gen template

proto/osmosis/twap/v1beta1/query.yml
ValarDragon: Yeah just an idea!

So in line 7, thats the name of the internal keeper function

On line 8, I think your right, we could make it accept both `proto struct name` and `registered proto type` (and in the latter case allow just putting `Response{twap}`

x/twap/twapmodule/module.go
ValarDragon: Theres a client import further up `	"github.com/cosmos/cosmos-sdk/client"`

ValarDragon:
Latest commit codegens the grpc package

proto/osmosis/twap/v1beta1/query.yml
```diff
@@ -0,0 +1,12 @@
+keeper: 
+  path: "github.com/osmosis-labs/osmosis/v10/x/twap"
```
p0mvn, line 2: Can we also update our import paths job to replace the version here please:
https://github.com/osmosis-labs/osmosis/blob/main/scripts/replace_import_paths.sh

p0mvn (approved):

cmd/querygen/main.go
```diff
@@ -0,0 +1,95 @@
...
+		if !errors.Is(err, os.ErrExist) {
+			return err
+		}
+	}
+	// generate file
+	f, err := os.Create(fsClientPath + "/grpc/grpc_query.go")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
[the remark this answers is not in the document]
ValarDragon: Its actually golang standard to have this after the error check. Staticcheck enforces this.

proto/osmosis/twap/v1beta1/query.yml
ValarDragon: Why is that not just a sed over every single file?

Issue #2337, opened 2022-08-09
Change `EstimateSwap` to not require assets


mattverse:
## Background

Current `EstimateSwap` queries, namely `EstimateSwapExactAmountIn` and `EstimateSwapExactAmountOut` and the `EstimateSwap` API in the bindings query does not only estimate swaps but includes the actual process of swapping to calculate and estimate. IMO, an API for estimating should not include balance check for the account trying to do the query. Take the gas estimation query in the Cosmos SDK for example, it does not require the sender to have the designated amount of fee and balance as well. 

## Suggested Design

I suggest that we move towards supporting the `EstimateSwap` API so that swaps can be estimated without performing the actual performance of swaps. This can be possibly done by separating out some of the logic within `MultihopSwapExactAmountOut` and `MultihopSwapExactAmountIn`

ValarDragon:
Closed by #3424 which got into v13!

We should update the queries accordingly as well

PR #2378, opened 2022-08-12
Estimate swap

vuong177:
Closes: #2337

Change `EstimateSwap` to not require assets.

## Brief Changelog
- Remove field `Sender` in `EstimateSwap` queries.

## Testing and Verifying

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no) no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no) no 
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented) not applicable


proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -146,9 +146,8 @@ message QuerySpotPriceResponse {
 
 //=============================== EstimateSwapExactAmountIn
 message QuerySwapExactAmountInRequest {
-  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
-  uint64 pool_id = 2 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
-  string token_in = 3 [ (gogoproto.moretags) = "yaml:\"token_in\"" ];
+  uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string token_in = 2 [ (gogoproto.moretags) = "yaml:\"token_in\"" ];
```
ValarDragon, line 150: Thanks for catching this, pretty funny mistake lol!

However, please don't change the field numbers in proto files like this, lets keep them at the old field numbers, this makes it so legacy clients don't have to update, to 'just work' with the changed proto.

vuong177: Yep, I'll return old field number.

vuong177:
I've created a PR  https://github.com/osmosis-labs/bindings/pull/37 to remove sender field.

wasmbinding/queries.go
```diff
@@ -95,16 +95,12 @@ func (qp QueryPlugin) EstimateSwap(ctx sdk.Context, estimateSwap *bindings.Estim
...
-	if err != nil {
-		return nil, sdkerrors.Wrap(err, "gamm estimate swap sender address")
-	}
 
 	if estimateSwap.Amount == (bindings.SwapAmount{}) {
 		return nil, wasmvmtypes.InvalidRequest{Err: "gamm estimate swap empty swap"}
 	}
 
-	estimate, err := PerformSwap(qp.gammKeeper, ctx, senderAddr, estimateSwap.ToSwapMsg())
+	estimate, err := PerformSwap(qp.gammKeeper, ctx, nil, estimateSwap.ToSwapMsg())
```
mattverse, line 103: Can we change this API as well? We can erase the sender Addr here as well

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -165,7 +164,6 @@ message QuerySwapExactAmountInResponse {
 
 //=============================== EstimateSwapExactAmountOut
 message QuerySwapExactAmountOutRequest {
-  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
   uint64 pool_id = 2 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
```
mattverse, line 167: This change is going to be a breaking change anyhow, Can we update the field numbers? (e.g pool_id -> 1)

x/gamm/keeper/swap.go
```diff
@@ -154,22 +154,24 @@ func (k Keeper) updatePoolForSwap(
 		return err
```
mattverse, line 154: I don't think calling a setter method(`setPool`) inside a query method is a safe way of doing this IMO.
We should either completely separate out the querying methods from the existing method or create a new separate method for this estimate call

mattverse (changes requested):
Thanks for tackling this! I personally don't think it's a safe way to have a setter inside a query method nor do I think using the same route for swaps and queries is a safe way of doing this. 

I would like to suggest further separation of the setters inside the existing method of swapping, or create a new method for this query

proto/osmosis/gamm/v1beta1/query.proto
mattverse: Ah just checked the convo above resolving this

x/gamm/keeper/swap.go
ValarDragon: Yeah, this is concerning, great catch.

So turns out some queries in the SDK already do this, so for now this is grudgingly ok, we should cache the context in our query layer to prevent this for now.

x/gamm/keeper/swap.go
```diff
@@ -154,22 +154,24 @@ func (k Keeper) updatePoolForSwap(
...
-	if err != nil {
-		return err
-	}
-
-	err = k.bankKeeper.SendCoins(ctx, pool.GetAddress(), sender, sdk.Coins{
-		tokenOut,
-	})
-	if err != nil {
-		return err
+	if sender != nil {
```
ValarDragon, line 157: I don't think we should do this

vuong177:
Thanks @ValarDragon @mattverse. I'll make your suggestion.

x/gamm/keeper/swap.go
vuong177: Yes, I'll revert old logic.

x/gamm/keeper/swap.go
ValarDragon: I suggest we don't block this pr on this

mattverse (approved):

ValarDragon:
This absolutely needs a changelog update. @mattverse can you PR one?

ValarDragon:
Wait this code completely changed since last review

vuong177:
hey duplicate code beacuse esimate swap have similar logic as real swap. Should we add a boolean variable `estimate` in `UpdatePoolForSwap`

vuong177:
Also add `estimate` in `SwapExactAmountIn` and `MultihopSwapExactAmountIn`

mattverse:
Yeah, with the scope of this PR getting big, the approach I would want to take on re-merging this PR is me shooting a secondary PR on top of this branch for further refactoring and reducing code duplication, after we merge that second PR into this one, we can give it a last review and then re merge this PR.

mattverse:
@vuong177 https://github.com/osmosis-labs/osmosis/pull/2501 would appreciate it if you could also review this :)

PR #2384, opened 2022-08-12
fix(ci): import paths script to work with query gen

p0mvn:
Closes: #XXX

The ci script is not working for the new query generation because it selects specific files types to update import paths for. Instead, we should relax the constraint of what files should be updated (all but in `vendor/*`, `git/*` or `*.md` files)

## Testing and Verifying

Tested that script works locally

### DIFF
--- a/scripts/replace_import_paths.sh
+++ b/scripts/replace_import_paths.sh
@@ -24,9 +24,7 @@ files=$(find ./ -type f -and -not \( -path './vendor*' -or -path './.git*' -or -
 
 echo "Updating all files"
 for file in $files; do
-    if [ -f "${file}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/replace_import_paths.sh
```diff
@@ -17,30 +17,25 @@ replace_paths() {
...
-for mod in $modules;
-do
-    for file in $mod/*; do
-        if [ -f "${file}" ]; then
-            replace_paths $file
-        fi
-    done
-done
+echo "Replacing import paths in all files"
+pwd
```
p0mvn, line 21:
```suggestion
```

ValarDragon (approved):

Issue #2042, opened 2022-07-13
proposal: implement the ability to print out container logs when e2e fails waiting on node to reach a certain height


p0mvn:
## Background

Currently, it is challenging to debug e2e upgrades and setup both in CI and locally because, when a node container fails, its logs are not printed to stdout.

We need to be able to print the logs from a failing container to the host's stdout.

## Story
>As an Osmosis engineer, I would like to be able to easily identify the source of e2e node container failures so that I can avoid debugging our tooling and focus at being efficient at fixing the actual failures

## Example

This is an example of what we see in the CI logs when node fails to produce blocks:
https://github.com/osmosis-labs/osmosis/runs/7312134568?check_suite_focus=true

Not much useful information is available.

## Suggested Design

We should be able to achieve this by adding logic to [the following branch](https://github.com/osmosis-labs/osmosis/blob/3e449625e06fe8d86ad320d1a2989c19075fea3b/tests/e2e/configurer/chain/chain.go#L117) of `WaitUntil` method of `chain.Config` module.

`chain.Config` has `containerManager` embedded as [its field](https://github.com/osmosis-labs/osmosis/blob/3e449625e06fe8d86ad320d1a2989c19075fea3b/tests/e2e/configurer/chain/chain.go#L30). Therefore, it should be possible to get access to the node's container by calling [`GetValidatorResource(...)`](https://github.com/osmosis-labs/osmosis/blob/3e449625e06fe8d86ad320d1a2989c19075fea3b/tests/e2e/containers/containers.go#L245)

Once we have the container resource, we can access its `LogPath`. Next, read the last 100-200 files of the [`LogPath`](https://github.com/ory/dockertest/blob/4d608dbb348c52b3f719f3631aae6008cafefcb5/docker/container.go#L462) file directly and try to output it to the host's stdout.

I propose implementing a method on `containers.Manager`:
 ```
// ExtractLogs extracts logs from a container dedicated to a node running on chainId and identified by nodeIndex. It 
// returns the last 100 lines of the container logs, if available. Returns error if not available or if a container does not exist.
func (m *Manager)  ExtractLogs(chainId string, nodeIndex) ([]byte, error)
```
This method should be called from [this branch](https://github.com/osmosis-labs/osmosis/blob/3e449625e06fe8d86ad320d1a2989c19075fea3b/tests/e2e/configurer/chain/chain.go#L117) in `WaitUntil`

Then, the result ican be used to log the bytes to the host's stdout.

[...]

PR #2385, opened 2022-08-12
feat(e2e): container logs printed by default and option to disable

p0mvn:
Closes: #2042

Based on my experience with e2e, we should be printing container logs from executing CLI commands by default to assist with debugging.

It does add verbosity. However, I think having more logs is better than having none.

Therefore, this PR adds a change to enable logs by default and adds an option to disable it via an environment variable `OSMOSIS_E2E_DEBUG_LOG`.

The output is printed at most 3 times per CLI command. Multiple prints may happen if the container is unresponsive/slow on startup. However, from my tests locally, it prints once per command most of the time.

## Testing and Verifying

Tested locally.

Example new logs: https://github.com/osmosis-labs/osmosis/runs/7811392841?check_suite_focus=true

### DIFF
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -84,12 +84,19 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		s.T().Log("skipping state sync testing")
 	}
 
+	isDebugLogEnabled := false
+	if str := os.Getenv("OSMOSIS_E2E_DEBUG_LOG"); len(str) > 0 {
+		isDebugLogEnabled, err = strconv.ParseBool(str)
+		s.Require().NoError(err)
+		s.T().Log("debug logging is enabled. container logs from running cli commands will be printed to stdout")
+	}
+
 	if str := os.Getenv(upgradeVersionEnv); len(str) > 0 {
 		upgradeSettings.Version = str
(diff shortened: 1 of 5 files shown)


tests/e2e/e2e_setup_test.go
```diff
@@ -84,12 +84,19 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		s.T().Log("skipping state sync testing")
 	}
 
+	isDebugLogEnabled := false
+	if str := os.Getenv("OSMOSIS_E2E_DEBUG_LOG"); len(str) > 0 {
+		isDebugLogEnabled, err = strconv.ParseBool(str)
+		s.Require().NoError(err)
+		s.T().Log("debug logging is enabled. container logs from running cli commands will be printed to stdout")
```
czarcas7ic, line 91: Confused about this log. I know its just in the test file but wouldn't this be triggered on both True and False?

czarcas7ic (approved):
This is great, slowly making debugging in e2e less and less painful 😅 Nice work 🚀

tests/e2e/e2e_setup_test.go
p0mvn: good point, will fix

ValarDragon (approved):

ValarDragon:
I somewhat think this may eventually evolve into we have a caller-provided logger, but I think thats something we do once we need more fine grained logs

PR #2389, opened 2022-08-12
Delete some deadcode

ValarDragon:

Driveby deadcode elimination, that I separated out from another PR to not distract there.


czarcas7ic (approved):
ACK

Issue #2270, opened 2022-08-02
x/lockup: Ensure events are emitted and tests added


stackman27:
## Background

Currently, our tests are missing assertions about events. Moreover, it seems that some of our events are missing based on [`#1939`](https://github.com/osmosis-labs/osmosis/pull/1939)

The lockup module emits the following events:

There are 4 types of events that exist in lockup module:
```
* `types.TypeEvtLockTokens` - "lock_tokens"
* `types.TypeEvtAddTokensToLock` - "add_tokens_to_lock"
* `types.TypeEvtBeginUnlockAll` - "begin_unlock_all"
* `types.TypeEvtBeginUnlock` - "begin_unlock"

### `types.TypeEvtLockTokens`

* `types.AttributePeriodLockID`
* `types.AttributePeriodLockOwner`
* `types.AttributePeriodLockAmount`
* `types.AttributePeriodLockDuration`
* `types.AttributePeriodLockUnlockTime`

### `types.TypeEvtAddTokensToLock`

* `types.AttributePeriodLockID` 
* `types.AttributePeriodLockOwner`
* `types.AttributePeriodLockAmount`

### `types.TypeEvtBeginUnlockAll`

* `types.AttributePeriodLockOwner`
* `types.AttributeUnlockedCoins`

### `types.TypeEvtBeginUnlock`

* `types.AttributePeriodLockID`
* `types.AttributePeriodLockOwner`
* `types.AttributePeriodLockDuration`
* `types.AttributePeriodLockUnlockTime`
```

Therefore, this issue is twofold:

- Investigate if / which events are missing and restore them

- Add assertions in tests that events are being emitted in the correct places

## Suggested Design
Very similar to what @p0mvn suggested for superfluid events! [`#2194`](https://github.com/osmosis-labs/osmosis/issues/2194)

## Acceptance Criteria
- lockup events are emitted correctly in the right locations
- tests added for attributes
- tests added for verifying that events are emitted in the correct locations
- every event is documented in lockup README
   * what the events is for / when it is emitted
   * what attributes it contains
- final design is consistent with x/gamm and x/superfluid events

stackman27:
Epic #2195

ValarDragon:
See PR discussion

PR #1939, opened 2022-07-01
lint: enable unused in golangci-lint

p0mvn:
Follow-up to: https://github.com/osmosis-labs/osmosis/issues/1897

## What is the purpose of the change

We would like to improve the quality of our codebase by trying out components of the [staticcheck](https://staticcheck.io/) linter.

This change introduced `unused` - 1st of the 4 components

Although, we added `deadcode` linter in https://github.com/osmosis-labs/osmosis/pull/1934, `unused` seems to be detecting more unused logic.

In addition to removing the unused code, the following was done:
- disable `unused` and `deadcode` for `x/gamm/pool-models/stableswap` since it is currently in development

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

Issue #2195, opened 2022-07-22
[Epic] Quality assurance for events


p0mvn:
## Background

During the internal audit, we discovered that the tests and documentation for events across all modules are limited. This is an epic to track work across modules

## Stories
- https://github.com/osmosis-labs/osmosis/issues/1942
- https://github.com/osmosis-labs/osmosis/issues/2194
- TBD: `epochs`
- https://github.com/osmosis-labs/osmosis/issues/2270
- TBD: `mint`
- TBD: `pool-incentives`
- TBD: `tokenfactory`
- TBD: `txfees` 

## Acceptance Criteria
- events are tested and documented across all modules
- the design is consistent across modules

p0mvn:
We should investigate if events are emitted correctly in other modules and create issues for them, if applicable

stackman27:
I can try looking into x/lockup as i'm pretty familiar with that module

p0mvn:
> I can try looking into x/lockup as i'm pretty familiar with that module

That would be awesome! Would you like to create an issue to track this? Similar to #1942 and #2194

p0mvn:
Synched offline - will not pursue this further unless asked for guarantees by integrators. Will wait for more event improvements from the SDK and/or code gen what's possible

PR #2391, opened 2022-08-12
x/lockup: ensure events are emitted and tests added 

stackman27:
Closes: #2270

Currently, our tests are missing assertions about events. Moreover, it seems that some of our events are missing based on 
[`1939`](https://github.com/osmosis-labs/osmosis/pull/1939)

The lockup module emits the following events:

There are 4 types of events that exist in lockup module:

* `types.TypeEvtLockTokens` - "lock_tokens"
* `types.TypeEvtAddTokensToLock` - "add_tokens_to_lock"
* `types.TypeEvtBeginUnlockAll` - "begin_unlock_all"
* `types.TypeEvtBeginUnlock` - "begin_unlock"

## Testing and Verifying
1. Added events test in msg_server_test 
2. Added emit.test to test out if events emit correct events


x/lockup/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,110 @@
...
+		return
+	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		BeginUnlockEvent(lock),
+	})
+}
+
+// Question: Any way we can make this private method?
+func BeginUnlockEvent(lock *types.PeriodLock) sdk.Event {
```
stackman27, line 102: We use this Event to append [here](https://github.com/osmosis-labs/osmosis/pull/2391/files#diff-2b31c7623003924d48a9340898c86e65c99d2790bd5f60b04af99a5d60e91bd9L136)

It would be ideal to make this a private method, but i dont want to repeat code

x/lockup/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,110 @@
...
+		types.TypeEvtLockTokens,
+		sdk.NewAttribute(types.AttributePeriodLockID, utils.Uint64ToString(lock.ID)),
+		sdk.NewAttribute(types.AttributePeriodLockOwner, lock.Owner),
+		sdk.NewAttribute(types.AttributePeriodLockAmount, lock.Coins.String()),
+		sdk.NewAttribute(types.AttributePeriodLockDuration, lock.Duration.String()),
+		sdk.NewAttribute(types.AttributePeriodLockUnlockTime, lock.EndTime.String()),
+	)
+}
+
+// EmitLockToken returns a new event when user lock tokens.
```
p0mvn, line 31:
```suggestion
// EmitExtendLockToken returns a new event when user locks tokens.
```

x/lockup/keeper/msg_server.go
```diff
@@ -46,14 +46,7 @@ func (server msgServer) LockTokens(goCtx context.Context, msg *types.MsgLockToke
...
 
-		ctx.EventManager().EmitEvents(sdk.Events{
-			sdk.NewEvent(
-				types.TypeEvtAddTokensToLock,
-				sdk.NewAttribute(types.AttributePeriodLockID, utils.Uint64ToString(lockID)),
-				sdk.NewAttribute(types.AttributePeriodLockOwner, msg.Owner),
-				sdk.NewAttribute(types.AttributePeriodLockAmount, msg.Coins.String()),
-			),
-		})
+		events.EmitAddTokenToLock(ctx, lockID, msg.Owner, msg.Coins.String())
```
p0mvn, line 49: I think we should move this inside keeper so that if another module ends up calling `AddToExistingLock`, the event does not get missed

x/lockup/keeper/msg_server.go
```diff
@@ -63,17 +56,7 @@ func (server msgServer) LockTokens(goCtx context.Context, msg *types.MsgLockToke
...
-			types.TypeEvtLockTokens,
-			sdk.NewAttribute(types.AttributePeriodLockID, utils.Uint64ToString(lock.ID)),
-			sdk.NewAttribute(types.AttributePeriodLockOwner, lock.Owner),
-			sdk.NewAttribute(types.AttributePeriodLockAmount, lock.Coins.String()),
-			sdk.NewAttribute(types.AttributePeriodLockDuration, lock.Duration.String()),
-			sdk.NewAttribute(types.AttributePeriodLockUnlockTime, lock.EndTime.String()),
-		),
-	})
-
+	events.EmitLockToken(ctx, &lock)
```
p0mvn, line 59: ditto for moving this inside the keeper's `CreateLock`

p0mvn:
Can we also add the event info to docs, please?

stackman27:
> Can we also add the event info to docs (README), please?

I didn't add README.md following @alexanderbez  comment [here](https://github.com/osmosis-labs/osmosis/pull/2255/files#r936843446)

p0mvn:
> > Can we also add the event info to docs (README), please?
> 
> I didn't add README.md following @alexanderbez comment [here](https://github.com/osmosis-labs/osmosis/pull/2255/files#r936843446)
> 
> Didn't know what the final verdict was

I think Bez is right that these tend to get out of sync and are difficult to maintain. However, these docs get auto-uploaded to our docs.osmosis.zone website. Currently, there is no information about events on the docs site.

I think that this information could be useful for integrators. So, in my opinion, we should still have it

stackman27:
> > > Can we also add the event info to docs (README), please?
> > 
> > 
> > I didn't add README.md following @alexanderbez comment [here](https://github.com/osmosis-labs/osmosis/pull/2255/files#r936843446)
> > Didn't know what the final verdict was
> 
> I think Bez is right that these tend to get out of sync and are difficult to maintain. However, these docs get auto-uploaded to our docs.osmosis.zone website. Currently, there is no information about events on the docs site.
> 
> I think that this information could be useful for integrators. So, in my opinion, we should still have it

Gotchu i'll definitely add it then

x/lockup/keeper/internal/events/emit_test.go
```diff
@@ -0,0 +1,264 @@
+package events_test
```
ValarDragon, line 1: This entire test file seems kinda useless

ValarDragon:
I don't really like this entire direction. 

Every event here seems like we could derive it automatically from the message response, and these tests aren't actually helping check that we are emitting things correctly post message execution

x/lockup/keeper/internal/events/emit.go
```diff
@@ -0,0 +1,110 @@
...
+func EmitExtendLockToken(ctx sdk.Context, lock *types.PeriodLock) {
+	if ctx.EventManager() == nil {
+		return
+	}
```
ValarDragon, line 35: Why are we doing all these nil guards? Seems fine to me if it panics with NPE, this is misconfiguration that should get caught in testing, not be graceful no-op

x/lockup/keeper/internal/events/emit_test.go
stackman27: This makes sense, the entire boilerplate tests definitely seems like an overkill. I will remove this so that we can test messages that emit these events directly in `msg_server_test.go`

stackman27:
> I don't really like this entire direction.
> 
> Every event here seems like we could derive it automatically from the message response, and these tests aren't actually helping check that we are emitting things correctly post message execution. At most extracting building an event to its own function makes sense to me. `ctx.EventManager().EmitEvent(constructAddTokensToLockEvent(lock))`, but again seems like short term status
> 
> The `message_server_test` changes are great, we should amend to test the real constructed event as wekk

We should definitely make a decision on how we should structure events + test events as we've followed this structure in x/gamm and x/superfluid [EPIC 2195](https://github.com/osmosis-labs/osmosis/issues/2195)

@p0mvn what do you think?

x/lockup/keeper/msg_server_test.go
```diff
@@ -322,8 +325,107 @@ func (suite *KeeperTestSuite) TestMsgEditLockup() {
...
 		if test.expectPass {
 			suite.Require().NoError(err, test.name)
+			suite.AssertEventEmitted(suite.Ctx, types.TypeEvtLockTokens, 2)
 		} else {
 			suite.Require().Error(err, test.name)
 		}
 	}
 }
+
+func (suite *KeeperTestSuite) TestMsgLockToken_EmitEvent() {
```
stackman27, line 335: @ValarDragon What are your thoughts on testing events like this?

x/lockup/keeper/msg_server_test.go
```diff
@@ -322,8 +325,107 @@ func (suite *KeeperTestSuite) TestMsgEditLockup() {
...
+
+			allEvents := suite.Ctx.EventManager().Events()
+			// filter out other events
+			actualEvents := make([]sdk.Event, 0)
+			for _, event := range allEvents {
+				if event.Type == types.TypeEvtLockTokens {
+					actualEvents = append(actualEvents, event)
+				}
+			}
+
```
stackman27, line 422: line [413-421] will be moved to `app/apptesting/events.go` once i receive approach ack

x/lockup/keeper/msg_server_test.go
```diff
@@ -322,8 +325,107 @@ func (suite *KeeperTestSuite) TestMsgEditLockup() {
...
+			if tc.expectPass {
+				suite.Require().Equal(expectedEvents[0], actualEvents[0])
```
stackman27, line 424: Tests the events type as well as its attributes

x/lockup/keeper/internal/events/emit.go
p0mvn: While I agree that no-op is probably not the way to go, I think it is worth having a check for nil event manager somewhere since `ctx` is injected into Osmosis's message server by its dependency - `cosmos-sdk`. I don't think we should be spending time on this right now though so it should be fine if we just remove the check

x/lockup/keeper/internal/events/emit_test.go
p0mvn: In my eyes, the goal at this level of abstraction is to test that attributes are set correctly. This is achieved by the tests here

Then, it is possible to simply focus that a certain event kind is emitted at the keeper/message server layer. Given that we have already tested the attributes here in `emit_test.go`, we can reduce the boilerplate in tests for keeper and simply call for testing the fact that an event has been emitted:
https://github.com/osmosis-labs/osmosis/blob/a1dc69413fd908810dbafe150967e814b796897b/app/apptesting/events.go#L7

Otherwise, we would have to validate the attributes in keeper and message server tests which goes against the separation of concerns

p0mvn:
> I don't really like this entire direction.
> 
> Every event here seems like we could derive it automatically from the message response, and these tests aren't actually helping check that we are emitting things correctly post message execution. At most extracting building an event to its own function makes sense to me. `ctx.EventManager().EmitEvent(constructAddTokensToLockEvent(lock))`, but again seems like short term status
> 
> The `message_server_test` changes are great, we should amend to test the real constructed event as week

> Every event here seems like we could derive it automatically from the message response

How can events be derived from the message response?

I don't think it is possible in the example here.

First, some messages emit events based on conditions. For example, [`LockTokens`](https://github.com/osmosis-labs/osmosis/blob/a1dc69413fd908810dbafe150967e814b796897b/x/lockup/keeper/msg_server.go#L32) emits `TypeEvtAddTokensToLock` if lock exists, `TypeEvtLockTokens` if it doesn't.

Second, these should really not be in the message server but in keeper. Other modules can call `CreateLock` bypassing `lockup` message server. As a result, if we have messages emitted here, they would be missed when `CreateLock` is called from another module.

> At most extracting building an event to its own function makes sense to me

That's the goal of the `internal/events` package inside the module to standardize the event creation for consistency. I agree that duplicating `Emit...` function for every event is unnecessary and was planning to refactor that after these PRs are merged in. However, I think that we should keep the constructors.

> again seems like short term status

That's fine because we can do incremental refactorings. I think changing the direction now when there is a relevant PR merged and 2 more are in progress, leading to inconsistent designs would be difficult to manage.

I also don't see how most of these events can be generated due to the reasons mentioned above.

> The message_server_test changes are great, we should amend to test the real constructed event as week

I mentioned the original thinking for the `emit_test.go` to be used for validating attributes in [this comment](https://github.com/osmosis-labs/osmosis/pull/2391#discussion_r946223130). Basically, if we test event construction at a lower level of abstraction, here we can only test the fact of a specific kind of event being emitted, without validating the attributes. The construction and attributes are already tested at a lower level of abstraction.

To summarize my thinking:
- I still think that we should construct the events from the `internal/events` package and test the attributes there
   * This construction pattern across modules makes it more visible and consistent where and when the events are emitted.
- We should test the event construction / attributes in the `events` package
- We can remove the duplicate `Emit...` functions and the event manager check
- I think validating the event construction at the message server layer might be useful but not necessary _given that we test the attributes / construction in the internal package_
- Some of these events should be moved to the keeper methods as per my earlier comments:
   * https://github.com/osmosis-labs/osmosis/pull/2391#discussion_r944995460
   * https://github.com/osmosis-labs/osmosis/pull/2391#discussion_r944995571

@ValarDragon @stackman27 please let me know what you think

p0mvn:
Hi @stackman27 

Just wanted to follow up here as well. Dev and I discussed this offline.

Let's move the events to keeper please but not use the internal/events logic. Please feel free to add tests like you were doing before.

We’ll not pursue this event QA further unless asked for by integrators and wait for improvements from the SDK

Sorry for having this blocked for quite some time

stackman27:
> Hi @stackman27
> 
> Just wanted to follow up here as well. Dev and I discussed this offline.
> 
> Let's move the events to keeper please but not use the internal/events logic. Please feel free to add tests like you were doing before.
> 
> We’ll not pursue the events QA further unless asked for by integrators and wait for improvements from the SDK
> 
> Sorry for having this blocked for quite some time

Gotchu! Thank you for sorting this out. I will continue adding new tests

PR #2404, opened 2022-08-14
sim: transfer sdk sim functions

czarcas7ic:

Making this PR to reduce the diff on future simulator work.

As discussed with @ValarDragon, we want to reduce our reliance on sdk simulation parameters/functions to only the ones that are absolutely necessary. This PR removes sdkSimapp from sim_test and replaces it with configs and utils that live within the executor folder.

## Brief Changelog

- Removes sdkSimapp from sim_test.go
  - Implements required flags/functions in executor folder in osmosis simulation

## Testing and Verifying


simulation/executor/util.go
```diff
@@ -28,7 +37,7 @@ func getTestingMode(tb testing.TB) (testingMode bool, t *testing.T, b *testing.B
 //  - "over stuffed" blocks with average size of 2 * avgblocksize,
 //  - normal sized blocks, hitting avgBlocksize on average,
 //  - and empty blocks, with no txs / only txs scheduled from the past.
-func getBlockSize(simCtx *simtypes.SimCtx, params Params, lastBlockSizeState, avgBlockSize int) (state, blockSize int) {
+func getBlockSize(simCtx *osmosim.SimCtx, params Params, lastBlockSizeState, avgBlockSize int) (state, blockSize int) {
```
ValarDragon, line 40: Lets keep this as simtypes and move the new code to its own file

simulation/executor/config.go
```diff
@@ -0,0 +1,83 @@
+package simulation
```
ValarDragon, line 1: I'd rec, lets move the other copied functions into this file, and call it `legacyconfig.go`, as a note to ourselves to come back and better abstract later

ValarDragon (approved):
Awesome!

simulation/executor/util.go
czarcas7ic: I think a prematurely opened this PR as I made quite a few more changes in which I reference SimCtx via osmosim instead of through sdk simulator. If it is okay with you I will make one more large commit, and from there could you let me know if this point still stands?

simulation/executor/types.go
```diff
@@ -0,0 +1,175 @@
+package simulation
```
czarcas7ic, line 1: unsure how you feel about bringing this into osmosis @ValarDragon . I felt like it would potentially be helpful to have direct control over the OperationMsg struct specifically

czarcas7ic:
These additional changes may not be welcome / might be going in the wrong direction with it. I went down this long rabbit hole of types that are strung together between the sdk and osmosis in terms of the simulator (was really just trying to bring a couple things over when I was working on the SQL stuff and they got here lol)

czarcas7ic:
Okay, reverted the unnecessary changes! Will let it sit here for just a bit in case anyone else wants to give it some eyes. Otherwise will merge when my database PR is ready 😄

simulation/executor/init_functions.go
```diff
@@ -1,4 +1,4 @@
-package simtypes
```
ValarDragon, line 1: Why did we move this file?

Recommend not moving it, since these are type definitions

simulation/executor/types.go
```diff
@@ -0,0 +1,17 @@
...
+	simtypes "github.com/cosmos/cosmos-sdk/types/simulation"
+)
+
+// AppStateFn returns the app state json bytes and the genesis accounts
+type AppStateFn func(r *rand.Rand, accs []simtypes.Account, config Config) (
+	appState json.RawMessage, accounts []simtypes.Account, chainId string, genesisTimestamp time.Time,
+)
+
+// RandomAccountFn returns a slice of n random simulation accounts
+type RandomAccountFn func(r *rand.Rand, n int) []simtypes.Account
```
ValarDragon, line 17: I'm confused why were adding these types?

AppStateFn is a leagcy type, that new things shouldn't be using. We just have references to it, for the legacy actions / genesis handling

czarcas7ic: Without calling these from osmosis, we are forced to call them from simulation (sdk side) which uses a simulation.Config type rather than a Config type. We want to use our local definition of Config since this is what holds all the flags etc. Please lmk if I am misunderstanding

simulation/executor/init_functions.go
czarcas7ic: In conjunction with the comment below, if we want to use our local definition of Config, we need to use our local definition of both RandomAccountFn and AppStateFn. Because executor folder imports simtypes, we cant have simtypes import executor (as far as I have learned) due to a cyclical import error. This is why I moved this file to executor so I can still call our local implementation of both RandomAccountFn and AppStateFn. If the below comment I made about Config is moot then this can easily be switched back.

simulation/executor/types.go
ValarDragon: hrmm, not sure I understand still, but this isn't a big deal / we can just change it later

Issue #2207, opened 2022-07-23
gamm: Create tests to find the smallest and largest spot prices


ValarDragon:
## Background

As part of implementing overflow handling for TWAP, we need to have a clear understanding of what is the full range of SpotPrices in the gamm module, and if we need to limit these. (e.g. by requiring the pool to always have at least 1000 base units of each coin)

## Suggested Design

- Using the Balancer equation try to get the smallest and large spot prices we can
- See if theres limits on weights / token supplies we can do, to bound the maximum if its too precise.

## Acceptance Criteria

- [ ] Clear understanding of smallest and largest spot price
- [ ] If largest = sdk.MaxDec, thinking of ways to reduce it. (e.g. min liquidity guaranteed after every action)

PR #2405, opened 2022-08-14
x/gamm: Catch overflow panics on `CalculateSpotPrice`

AlpinYukseloglu:
Closes: #2207, #2445

This PR prevents `SpotPrice` from panicing and makes it return an error for invalid or overflowing spot prices instead. It also adds tests to ensure spot prices stay within the set bounds.

## Brief Changelog

- `SpotPrice` now recovers panics and returns an error. It also sets a max spot price of 2^160

## Testing and Verifying

- New tests can be found in x/gamm/pool-models/balancer/pool_suite_test.go

### DIFF
--- a/x/gamm/pool-models/balancer/pool.go
+++ b/x/gamm/pool-models/balancer/pool.go
@@ -609,7 +609,7 @@ func (p *Pool) applySwap(ctx sdk.Context, tokensIn sdk.Coins, tokensOut sdk.Coin
 //
 // panics if the pool in state is incorrect, and has any weight that is 0.
 // TODO: Come back and improve docs for this.
-func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec, error) {
+func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (spotPrice sdk.Dec, err error) {
 	quote, base, err := p.parsePoolAssetsByDenoms(quoteAsset, baseAsset)
 	if err != nil {
 		return sdk.Dec{}, err
@@ -618,14 +618,23 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
+	defer func() {
+		// defer function to escape the panic when spot price overflows
+		if r := recover(); r != nil {
+			spotPrice = sdk.Dec{}
+			err = types.ErrSpotPriceOverflow
+		}
+	}()
+
 	// spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
 	// spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
 	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
 	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
 	fullRatio := supplyRatio.Mul(invWeightRatio)
 	// we want to round this to `SigFigs` of precision
-	ratio := osmomath.SigFigRound(fullRatio, types.SigFigs)
-	return ratio, nil
+	spotPrice = osmomath.SigFigRound(fullRatio, types.SigFigs)
+
+	return spotPrice, err
 }
 
 // calcPoolOutGivenSingleIn - balance pAo.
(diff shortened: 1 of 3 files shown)


x/gamm/pool-models/balancer/pool.go
```diff
@@ -618,14 +618,23 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
+	defer func() {
+		// defer function to escape the panic when spot price overflows
+		if r := recover(); r != nil {
+			spotPrice = sdk.Dec{}
+			err = types.ErrSpotPriceOverflow
```
ValarDragon, line 625: Lets mark this as a spot price internal error

x/gamm/pool-models/balancer/pool.go
```diff
@@ -618,14 +618,23 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
...
 	// spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
 	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
 	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
 	fullRatio := supplyRatio.Mul(invWeightRatio)
 	// we want to round this to `SigFigs` of precision
-	ratio := osmomath.SigFigRound(fullRatio, types.SigFigs)
-	return ratio, nil
+	spotPrice = osmomath.SigFigRound(fullRatio, types.SigFigs)
+
+	return spotPrice, err
```
ValarDragon, line 637: Needs to add an overflow check here right? (E.g. a maximum of 2**128 or 2**160)?

AlpinYukseloglu: Right! I just added a check that capped at 2^160, although we could technically increase it to 2^196 without overflowing on equal weight pools. I personally think the buffer is useful regardless but would be curious to hear your thoughts

AlpinYukseloglu:
Comments should all be addressed!

x/gamm/pool-models/balancer/pool.go
```diff
@@ -618,14 +618,27 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (sdk.Dec,
...
 	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
 	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
 	fullRatio := supplyRatio.Mul(invWeightRatio)
 	// we want to round this to `SigFigs` of precision
-	ratio := osmomath.SigFigRound(fullRatio, types.SigFigs)
-	return ratio, nil
+	spotPrice = osmomath.SigFigRound(fullRatio, types.SigFigs)
+	if spotPrice.GT(sdk.NewDec(2).Power(160)) {
+		spotPrice = sdk.Dec{}
+		err = errors.New("spot price overflow")
```
ValarDragon, line 638: err, the prior error type was good! We just need two errors:
`ErrSpotPriceInternal` and `ErrSpotPriceOverflow`

ValarDragon (approved):
LGTM once we re-instate the error types!

x/gamm/pool-models/balancer/pool.go
AlpinYukseloglu: Ah I see what you meant now – should be fixed now!

ValarDragon (approved):
Nice work on this, thanks!

We should also update the spec + code comment, but we can do that in a follow up PR

ValarDragon:
Oh I didn't catch this, this was done at the balancer level, which is the wrong place to do it.

This should be done at the keeper level, will move in a follow-up PR.

ValarDragon:
Yikes, spot price rounding is also at the wrong layer of abstraction

AlpinYukseloglu:
By spot price rounding do you mean the `2^160` spot price cap? In any case it should be a quick move to [here](https://github.com/osmosis-labs/osmosis/blob/bd683a39f362eca659a73c2e8844e5fe62d3bdcd/x/gamm/keeper/pool_service.go#L29) for both the overflow check and rounding if that's what you mean – the issue had originally started as balancer tests for finding min/max prices (which was easiest to implement at the balancer level) so it seems as though the rest of the changes got stuck at that abstraction layer as well. Happy to open a new PR with changes moved to `CalculateSpotPrice` at the keeper level if you haven't started on it yet

PR #2408, opened 2022-08-15
IBC Rate limiting contract

nicolaslara:
Partial work for: #2305

Add the contract to be used for IBC rate limiting

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

 * There are unit tests and integration tests in the contract


x/ibc-rate-limit/.beaker/state.json
```diff
@@ -0,0 +1 @@
+{}
```
ValarDragon, line 1: should this file be in the gitignore?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,501 @@
...
+use crate::error::ContractError;
+use crate::management::{
+    add_new_channels, try_add_channel, try_remove_channel, try_reset_channel_quota,
+};
+use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
+use crate::state::{ChannelFlow, FlowType, CHANNEL_FLOWS, GOVMODULE, IBCMODULE};
+
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:rate-limiter";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
```
ValarDragon, line 17: oh huh, is this standard things you have to do for cosmwasm?

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: >
+          rustup target add wasm32-unknown-unknown;
+          cargo build --release --target wasm32-unknown-unknown
+
+      - name: Test
+        working-directory: ${{ matrix.workdir }}
+        run: >
+          cargo test
+
+#      - name: Check Test Data
```
p0mvn, line 37: Is this going to be re-enabled in a future PR?

x/ibc-rate-limit/.beaker/state.json
p0mvn, line 1: Is this file needed for something?

x/ibc-rate-limit/.gitignore
```diff
@@ -0,0 +1,21 @@
+# Generated by Cargo
+# will have compiled files and executables
+debug/
```
p0mvn, line 3: What do you think about having this in the root git ignore to have a single source of truth for exclusions?

x/ibc-rate-limit/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+Cargo.lock
+
+# These are backup files generated by rustfmt
+**/*.rs.bk
+
+# MSVC Windows builds of rustc generate these, which store debugging information
+*.pdb
+
+
+# Ignores local beaker state
```
p0mvn, line 20: There seems to be an empty `.gitkeep` file. Do we need to keep it?

x/ibc-rate-limit/.gitignore
p0mvn: Noticed that there is another `.gitignore` file. Is there a reason for that?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,501 @@
...
+    direction: FlowType,
+    now: Timestamp,
+) -> Result<Response, ContractError> {
+    // Only the IBCMODULE can execute transfers
+    let ibc_module = IBCMODULE.load(deps.storage)?;
+    if sender != ibc_module {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let channels = CHANNEL_FLOWS.may_load(deps.storage, &channel_id)?;
```
p0mvn, line 106: asking to learn - is `CHANNEL_FLOWS` a whitelist of some kind?

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, Copy)]
+pub struct Flow {
+    pub inflow: u128,
+    pub outflow: u128,
+    pub period_end: Timestamp,
+}
+
+impl Flow {
```
p0mvn, line 27: Can we document what this is exactly please? I'm not very familiar with IBC concepts so it would be amazing having docs for some of these abstractions

p0mvn (commented):
This is really interesting, great work @nicolaslara . I'm still ramping up on the core concepts but had some initial set of questions

ValarDragon:
Should we have the IBC interface functions be under Cosmwasm Sudo message? https://docs.cosmwasm.com/docs/1.0/smart-contracts/sudo/

x/ibc-rate-limit/.gitignore
ValarDragon: I think it makes sense for per-contract .gitignore's

x/ibc-rate-limit/.beaker/state.json
nicolaslara: Not sure. I just used the defaults from beaker which have `state.local.json` in the gitignore, but not state.json. @iboss-ptk is there any reason to commit `.beaker/state.json`?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: yeah, this is part of [cw2](https://docs.cosmwasm.com/cw-plus/0.9.0/cw2/spec) and is used for migrating the contract. At the moment, we don't have migrations, so we could remove it; just left it there since it's standard in most contracts.

nicolaslara: Re: Not having a migration entrypoint: My thought was that since any changes would have to be approved by governance anyway, it would be easier to just deploy a new contract and update the param to point to it. 

Now that I think about it again, though, I'm noticing this would require two transactions (instantiate a new contract to get its address, update the param) instead of one (migrate the existing contract). So I should probably add a migration entrypoint

.github/workflows/contracts.yml
nicolaslara: yes, this doesn't make sense in the current PR since the testdata isn't there (that's part of the go tests)

One issue I still need to figure out is what the best way to compile the contract in the GH action is. Right now we're testing the build in dev mode, which is faster and has no optimizations. To build the wasm code that can be uploaded to the chain (the one in `testdata/rate_limiter.wasm`) we need to optimize by either using https://github.com/CosmWasm/rust-optimizer/ or https://github.com/mandrean/cw-optimizoor. The first one adds complexity to the task cause we need to setup docker inside the image, and the second one takes a long time to install (and I didn't want to add a workflow that takes 10+ minutes).

Right now, I'm leaning towards using cw-optimizoor and caching the install step, but I'm not very familiar with how caching works on GH workfows, so I have to look into that.

x/ibc-rate-limit/.beaker/state.json
nicolaslara: TBH, I'm not sure. It's part of beaker: https://github.com/osmosis-labs/osmosis/pull/2408#discussion_r946427708

x/ibc-rate-limit/.gitignore
nicolaslara: The reason for all the gitignore files is that we use a template to generate the boilerplate on each contract, and the template comes with the gitignore. 

It makes sense to have only one source of truth. I'll merge those.

x/ibc-rate-limit/.gitignore
nicolaslara: Also part of beaker. Not sure what beaker does if the directory isn't there. @iboss-ptk is it safe to remove the empty dir?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: `CHANNEL_FLOWS` is the main state of this contract (https://github.com/osmosis-labs/osmosis/blob/a6cf294173fd49ec6ed7b879b941c0694475b932/x/ibc-rate-limit/contracts/rate-limiter/src/state.rs#L114).

It maps the channel id to its "flows" which are: how much value has moved through the channel during the currently active time period (this week, for example), and a "quota" which represents what percentage of the channel's value (balance+locked value on osmosis) we are allowing to flow through that channel in a specific duration (1w, for example)

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
nicolaslara: yes! a flow isn't really an IBC concept but something specific to rate limiting, but I'll add comments (and maybe a readme?) to document this.

nicolaslara:
> Should we have the IBC interface functions be under Cosmwasm Sudo message? https://docs.cosmwasm.com/docs/1.0/smart-contracts/sudo/
> 
> (SendPacket, RecvPacket, AddChannel, CloseChannel)

The reason to add it as regular execute messages instead of sudo is that they can be executed by someone other than the chain. Right now we define who can execute them when we instantiate the contract (https://github.com/osmosis-labs/osmosis/blob/a6cf294173fd49ec6ed7b879b941c0694475b932/x/ibc-rate-limit/contracts/rate-limiter/src/state.rs#L104-L107).

Permissions are then checked so that only the gov module is allowed to make changes and only the transfer module is allowed to send/recv packets. 

My thought process is that even if these values will most likely be set to the module addresses, we *could* set them to something else (say, a multisig) if necessary.

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: Added a migration entrypoint in case we want to add migrations in the future

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
nicolaslara: Added some docs to `state.rs`

ValarDragon:
That makes sense. Wdyt about send packet and recv packet being sudo though? I feel like we have an api leak if its called by something not in the state machine, right?

nicolaslara:
> That makes sense. Wdyt about send packet and recv packet being sudo though? I feel like we have an api leak if its called by something not in the state machine, right?

yeah, that makes sense. I'll make the change.

x/ibc-rate-limit/contracts/rate-limiter/src/msg.rs
```diff
@@ -0,0 +1,73 @@
+use cosmwasm_std::Addr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Channel {
+    pub name: String,
+    pub quotas: Vec<QuotaMsg>,
+}
```
ValarDragon, line 9:
```suggestion
// The channel struct contains a name representing a unique identifier within ibc-go, and a list of rate limit quotas
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
pub struct Channel {
    pub name: String,
    pub quotas: Vec<QuotaMsg>,
}
```

ValarDragon: A bit unclear to me how this is used from the msg file, still working through the entire contract

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,506 @@
...
```
ValarDragon, line 106:
```suggestion
    let channel_flow_trackers = CHANNEL_FLOWS.may_load(deps.storage, &channel_id)?;
```

I was a bit confused about that this was iterating over IBC channels on first pass

x/ibc-rate-limit/.beaker/state.json
iboss-ptk: intentionally keep tracks of public chain so for example, if another contributor wants to interact with / migrate deployed contract, they have all the refs in place.

x/ibc-rate-limit/.gitignore
iboss-ptk: if it's `.gitkeep` under `contracts` dir, you can delete them, it is there for project template to generate empty `contracts` dir along.

x/ibc-rate-limit/.beaker/state.json
ValarDragon: This contract should be getting deployed in a future Osmosis upgrade

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: this makes sense. Had a hard time naming that one, but I like "trackers". I'll update that function

x/ibc-rate-limit/.gitignore
nicolaslara: removed

x/ibc-rate-limit/.beaker/state.json
nicolaslara: Right, so IIUC this will only have information if the contract is deployed via beaker and this file is committed afterwards. On the one hand, we would have information such as the code_id/contract_addr there; on the other, no one will probably use this directly since all interactions with the contract happen via governance (except maybe getting the current state, but I'm guessing people easily can craft that query manually). 

If that's right, I think it makes sense to delete this.

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -17,16 +17,19 @@ pub enum FlowType {
...
 /// It tracks inflows (transfers into osmosis) and outflows (transfers out of
 /// osmosis).
 ///
 /// The period_end represents the last point in time that for which this Flow is
 /// tracking the value transfer.
+/// TODO: Document that time windows are not rolling windows, but instead discrete repeating windows.
+/// This is a design decision chosen for gas reasons.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, Copy)]
 pub struct Flow {
+    // Q: Do we have edge case issues with inflow/outflow being u128, e.g. what if a token has super high precision.
```
ValarDragon, line 32: Highlighting question

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -119,6 +122,8 @@ impl From<&QuotaMsg> for Quota {
     }
 }
 
+// TODO: Add docs that this is the main tracker, we should be setting multiple of these per contract
+// Q: Should we rename to "ChannelRateLimiter", and rename flow to "FlowTracker"?
```
ValarDragon, line 126: Highlighting question

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -119,6 +122,8 @@ impl From<&QuotaMsg> for Quota {
...
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct ChannelFlow {
```
ValarDragon, line 128: What do you think about moving `Channel_flows` then `channelFlow` definition to the top of this file?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -88,6 +88,9 @@ pub struct ChannelFlowResponse {
     pub max: u128,
 }
 
+// TODO: Lets add some more docs for this, namely that channel_value is the total supply of the denom
+// Q: Is an ICS 20 transfer only 1 denom at a time, or does the caller have to split into several
+// calls if its a multi-denom ICS-20 transfer
```
ValarDragon, line 93: highlighting question

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,520 @@
...
+        let Attribute { key, value } = &res.attributes[2];
+        assert_eq!(key, "weekly_used");
+        assert_eq!(value, "90");
+
+        // Receiving 1% should fail. 3% already executed through the channel
+        let recv_msg = ExecuteMsg::RecvPacket {
+            channel_id: "channel".to_string(),
+            channel_value: 3_000,
+            funds: 30,
+        };
```
ValarDragon, line 413: This feels incorrect right? (Relative to what we want)

E.g. the 1% recv here should be going to `-1%` of what we had the prior week, rather than total net change

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,520 @@
...
+            channel_id: "channel".to_string(),
+            channel_value: 3_000,
+            funds: 300,
+        };
+        let err = execute(deps.as_mut(), mock_env(), info.clone(), msg).unwrap_err();
+        assert!(matches!(err, ContractError::Unauthorized { .. }));
+    }
+
+    #[test]
+    fn consume_allowance() {
```
ValarDragon, line 280: I think we should add a test ensuring that quotas roll over as expected

ValarDragon:
Left some comments with questions (some design questions, some minor code ones)

I'm happy to merge this as long as we track these and write a spec, to unblock the other work going on for getting rate limits to happen!

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: This is done implicitly here: https://github.com/osmosis-labs/osmosis/blob/1f237e8fb0f04a8fec2e17812be0295cbdcdffc3/x/ibc-rate-limit/contracts/rate-limiter/src/integration_tests.rs#L64, but I could check the internal state to make sure the reset actually sets the values to zero. 

The reason to have that in integration test is because we need to modify the time, so we need a mock chain

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
nicolaslara: That's a really good question. Internally, cosmwasm uses `Uint128` ([bank queries](https://github.com/CosmWasm/cosmwasm/blob/9bf920b00b76f31f3c73ed3632cd07cbd6a94b7b/packages/std/src/results/cosmos_msg.rs#L61), for example): https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/coins.rs#L10. So that's why I went with 128. 

Their implementation is basically `u128` with some [helpers](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/math/uint128.rs) attached.

It should be easy to change to larger representations though. Cosmwasm has `Uint256` and `Uint512`.

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
nicolaslara: Usually cosmwasm contracts have the state definition at the bottom of the state file, but we could move this around or maybe split the data structures into a separate file

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: We're operating at the [SendPacket](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L151) and [RecvPacket](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L174) level on the ibc transfer app (highlighted in the links). Those functions expect the packet to be of type [FungibleTokenPacketData](https://github.com/cosmos/ibc-go/blob/0a887e47318c21dc4668c1056ceae7a7e795a28c/modules/apps/transfer/types/packet.go#L29), which deals with only one denom at a time. 

I'm not sure where in the call-chain we split these for a multi-denom data, but might be worth figuring out how to test those at the go level (maybe [here](https://github.com/osmosis-labs/osmosis/blob/8128b0ca5bdf0a9c1a4ee021c6ed3323c39e26b1/x/ibc-rate-limit/ibc_middleware_test.go), though I don't see helpers for multi-denom transfers in ibctesting or docs for it in the [spec](https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#data-structures))

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: The way it's currently implemented it's tracking only one *balance*.  In that case, 3% has moved out of the chain, so if 1% were to move in, we're be at a 2% *balance*, so that's not allowed. The opposite would work: with `quotas={send: 3%, recv: 2%)`, even if you've already sent 3%, you could receive 1% cause that would put the balance at 2%. 

We could change the implementation to what you're describing though. This probably makes more sense if the quotas aren't going to be symetric. I think the "balance" implementation is a leftover from when I had the same quota for send and receive.

nicolaslara: >  going to -1% of what we had the prior week

Right now we're not tracking "what we had the previous week". Instead, the percentages are calculated based on the value of the channel at the time of the request. The current implementation actually allows more than the specified percentage (as the value grows/shrinks with every send) but the actual max converges pretty quickly for small percentages. The advantage is that we avoid tracking the the network value for each channel.

Happy to change the implementation here. I think both make sense

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    let configured = match channels {
+        None => false,
+        Some(ref x) if x.is_empty() => false,
+        _ => true,
+    };
+
+    if !configured {
+        // No Quota configured for the current channel. Allowing all messages.
+        // TODO: Should there be an attribute for it being allowed vs denied?
```
nicolaslara, line 120: @ValarDragon If we return an `Ok(response...)` it's always allowed. Otherwise we return a `Err(ContractError::...)`. Most of the times this will be `ContractError::RateLimitExceeded {}` but there could be other errors (unauthorized, channel doesn't exist, etc).

This is good because we want to stop the transaction (on the go side) when *any* error happens. Otherwise we'd have to check the response type.

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    let response = Response::new()
+        .add_attribute("method", "try_transfer")
+        .add_attribute("channel_id", channel_id);
+
+    // Adding the attributes from each quota to the response
+    // Code style Q: Should we move attribute setting to a function on response?
+    // Rust question: How does this work with one response being an error, I'm not sure how the flow works here
+    results.iter().fold(Ok(response), |acc, result| {
+        Ok(acc?
```
nicolaslara, line 178: @ValarDragon for the rust question: the `?` here on the accumulator catches and returns any potential error

x/ibc-rate-limit/contracts/rate-limiter/src/management.rs
```diff
@@ -88,6 +90,7 @@ pub fn try_reset_channel_quota(
                 channel_id: channel_id.clone(),
             }),
             Some(mut flows) => {
+                // Q: What happens here if quote_id not found? seems like we return ok?
```
nicolaslara, line 93: We shouldn't end up there if the quota is not found (that would be caught by the `None` arm above), but what could happen is that the vector is empty.

In that case, yeah, we return `Ok` with no changes applied. Update takes this and updates the stored data (which is the same as before). A bit unnecessary to update it here, but it allows is to return the error directly from that function (see the `?` at the end) instead of having to catch and handle it. Storing an empty vec should be a very cheap operation anyway

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
nicolaslara: Added docs and renamed

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+            .add_attribute("channel_id", channel_id)
+            .add_attribute("quota", "none"));
+    }
+
+    let mut rate_limits = trackers.unwrap();
+
+    let results: Result<Vec<RateLimitResponse>, _> = rate_limits
+        .iter_mut()
+        .map(|limit| {
+            // TODO: Should we move this to more methods on ChannelFlow?
```
mattverse, line 140: Is Flow == ChannelFlow? Are they the same thing?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+            // (Or at least update for time + rename for track send. the last one is a bit of a code style nit)
+            let max = limit.quota.capacity_at(&channel_value, &direction);
+            if limit.flow.is_expired(now) {
+                limit.flow.expire(now, limit.quota.duration)
+            }
+            limit.flow.add_flow(direction.clone(), funds);
+
+            let balance = limit.flow.balance();
+            if balance > max {
+                return Err(ContractError::RateLimitExceded {
```
mattverse, line 154: question: what happens to all the other computations and mutations happened prior to this err if errored?

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+    use cosmwasm_std::{from_binary, Addr, Attribute};
+
+    use crate::helpers::tests::verify_query_response;
+    use crate::msg::{Channel, QuotaMsg};
+    use crate::state::RESET_TIME_WEEKLY;
+
+    const IBC_ADDR: &str = "IBC_MODULE";
+    const GOV_ADDR: &str = "GOV_MODULE";
+
+    #[test]
```
mattverse, line 231: I wonder if we have something like the test suite in rust, maybe utilize helpers.rs, we should revisit to have test setup methods that we can reuse!

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,527 @@
...
+            channel_id: "channel".to_string(),
+            channel_value: 3_000,
+            funds: 300,
+        };
+        let err = execute(deps.as_mut(), mock_env(), info, msg).unwrap_err();
+        assert!(matches!(err, ContractError::RateLimitExceded { .. }));
+    }
+
+    #[test]
+    fn symetric_flows_dont_consume_allowance() {
```
mattverse, line 324: nit: Can we have one liner comments for some of the test cases on what we're trying to test?

mattverse (commented):
Super excited to see this PR! 
Left some comments and questions, contract looks good to me!

x/ibc-rate-limit/contracts/rate-limiter/src/management.rs
```diff
@@ -0,0 +1,239 @@
...
+                    quota: q.into(),
+                    flow: Flow::new(0_u128, 0_u128, now, q.duration),
+                })
+                .collect(),
+        )?
+    }
+    Ok(())
+}
+
+pub fn try_add_channel(
```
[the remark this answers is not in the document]
nicolaslara: No unit tests, but there's `management_add_and_remove_channel()` which tests add and remove in one go

x/ibc-rate-limit/contracts/rate-limiter/src/management.rs
```diff
@@ -0,0 +1,239 @@
...
+        GOVMODULE
+            .save(deps.as_mut().storage, &Addr::unchecked(GOV_ADDR))
+            .unwrap();
+
+        let msg = ExecuteMsg::AddChannel {
+            channel_id: "channel".to_string(),
+            quotas: vec![QuotaMsg {
+                name: "daily".to_string(),
+                duration: 1600,
+                send_recv: (3, 5),
```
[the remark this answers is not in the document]
nicolaslara: yes, those represent 3% and 5%. I went with integers to keep the math simple. CosmWasm doesn't support floats, so I would've had to use a custom [decimal](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/math/decimal.rs). Seemed like an overkill for something that can be represented as [0,100]

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: they're not! ChannelFlow is the old name for RateLimit. Just updated some of the names based on the feedback, but didn't change them in the review comments (so I could know who made which comment via git blame)

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: if we return an error, the transaction fails and everything gets reverted

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: yeah, I find cosmwasm tests to be quite repetitive. They're quick to write, but for larger suites it might be useful to abstract some of it.

x/ibc-rate-limit/contracts/rate-limiter/src/management.rs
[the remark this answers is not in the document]
mattverse: oh interesting, what happens if we want to indicate send recv in decimal point percentage (e.g 3.5%) in certain situations?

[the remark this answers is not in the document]
nicolaslara: right now we can't. It's either 3 or 4. We could change it if needed, though

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: I've just pushed a change so that this behaves as you expected. It's a bit less clean, as we need to consider both in and out flows, but closer to what a user would expect.

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: added

.github/workflows/contracts.yml
nicolaslara: Added the optimizations and data testing is uncommented in the main branch (will be uncommented in a future PR). Marking this as resolved for now.

x/ibc-rate-limit/.gitignore
nicolaslara: merged. marking as resolved.

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: Changed a few of the names. Marking this as resolved for now

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Self {
+        Self {
+            inflow: inflow.into(),
+            outflow: outflow.into(),
+            period_end: now.plus_seconds(duration),
+        }
+    }
+
+    /// The balance of a flow is how much absolute value has moved through the
+    /// channel before period_end.
```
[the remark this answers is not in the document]
nicolaslara: the code changed here, but I've improved the comments. Marking as resolved.

x/ibc-rate-limit/.beaker/state.json
nicolaslara: removed and added to the gitignore.

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: Added the change for "what we had the previous week" (caching the value in the quota on the first call or when it expires) in https://github.com/osmosis-labs/osmosis/pull/2408/commits/c2a72add8d1accaf188b7a61b296c885170dc3cc. 

I find it a bit confusing that the differences in the quotas could be quite large when you have multiple quotas for a path (denom+channel). Consider, for example, if there is the supply was 100 last week, but someone trades 1M and wants to ibc 1%. Even if the daily 1% limit would allow you, the weekly still would prevent you from sending this through IBC.

Maybe this is what we want. We can have a quick discussion about the tradeoffs and pick one

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
ValarDragon: Oh cool, then maybe IBC-go doesn't support multi-denom sends

ValarDragon: ah your definitely right; https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/tx.proto#L28-L29

multi-denom transfers aren't possible

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
ValarDragon: oh thats a good point, that its super weird. Lets def discuss what makes most sense

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
ValarDragon: ah gotcha, thats a great design choice!

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
nicolaslara: Did some reorg here. We have a function to set the attributes, and I'm considering making those attributes only available in testing. They are, after all, only seen by the state machine, not the end-user (and the response is being discarded).

nicolaslara:
Moved SendPacket and RecvPacket to sudo. There's a caveat worth noticing here though: when calling the contract via sudo, we don't specify a sender, which means that *any* module in the chain can call the contract. It's the responsibility of the go side to make sure this is not called in the wrong place.

ValarDragon (approved):
Awesome job!

ValarDragon:
Lets merge once tests pass!

asalzmann:
@nicolaslara we've been thinking about this problem as well, something I'm curious about - why key flows on (channel, denom)? Could a malicious actor create a new client-connection-channel and bridge assets off of the chain on a channel that's not rate-limited?

This is probably less relevant for assets that have been bridged onto the chain, but might be relevant for native assets that are minted and bridged off of the source chain.

asalzmann:
Thanks for talking through your reasoning here, it's very helpful!

> If the assets get sent through a suspicious channel (and thus wrapped accordingly) they just wouldn't be recognized by the other chains.

I think the question here is how users of other chains perceive the trust properties of vouchers originating from different channels. My mental model is that the trust properties between channels are equivalent and that trust properties are derived from clients, although it's very possible that mental model is wrong. For bridged assets, the application can choose canonical channels, so while the trust properties might be the same, most of the value is locked in a few channels.

A practical case where this could matter would be a DEX that allows deposits of any IBC voucher from another chain that's one hop away, and routes the asset to the associated pool by querying the denomTrace (this might be a bad design for a DEX!) Global rate limiting is tough because creating more clients / connections / channels is permissionless, so if you add a default rate limit of 5%, you could open 20 channels to get around it.

> The bigger danger (and what rate limiting is trying to prevent) is those assets flowing through "proper" channels, where they would be indistinguishable* from others of the same denom.

Agreed! I like the approach.

asalzmann:
One other question @nicolaslara, what's the rationale for rate-limiting inflows?

nicolaslara:
> The main question I have is how users of other chains perceive the trust properties of vouchers originating from different channels. My mental model is that the trust properties between channels are equivalent and that the trust properties of vouchers are derived from clients, although it's very possible that mental model is wrong.
> 
> A practical case where this could matter would be a DEX that allows deposits of any IBC voucher from another chain that's one hop away, and routes the asset to the associated pool by querying the denomTrace (this might be a bad design for a DEX!) Global rate limiting could definitely work!

I'd be interested to know if there are chains out there that do this. By default, IBC-go [sets](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L284-L305) the denom to the trace hash, which effectively makes them different coins. A chain with this design would need to either override this, or have some mechanism for mixing them (a stableswap, for example). 

I think this is a larger question of IBC routing/wrapping that is still unresolved. Not sure if there's an accepted design for this, but I'd expect unwrapping IBC denoms would require asking the issuing chains to validate them (haven't thought about this enough, though)

> One other question @nicolaslara, what's the rationale for rate-limiting inflows?

In the case of an infinite-mint on another chain (think terra collapse) someone could bring unlimited amounts of a token and tank the price of assets pooled against it; effectively draining the pools.

PR #2308, opened 2022-08-04
Supply offset changes on sdk fork

nicolaslara:
Closes: #2307

## What is the purpose of the change

The sdk fork is using GenesisSupplyOffsets in the banktypes.NewGenesisState() constructor. This is incompatible with the ibc-go testing package which expects the new signature (https://github.com/cosmos/ibc-go/tree/main/testing).

## Brief Changelog

 - Internal updates to match changes in https://github.com/osmosis-labs/cosmos-sdk/pull/312

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage. The existing tests pass. 

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (not applicable  )

PR #2409, opened 2022-08-15
IBC Rate-Limit Middleware

nicolaslara:
Partial work for: https://github.com/osmosis-labs/osmosis/issues/2305

Add the IBC middleware to be used for rate limiting

## Brief Changelog

  - Added the middleware to be used for rate limiting

## Testing and Verifying

No tests here. The tests will be part of the integration PR


tests/e2e/e2e_test.go
```diff
@@ -44,6 +44,23 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+	if s.skipIBC {
+		s.T().Skip("Skipping IBC tests")
+	}
+	chainA := s.configurer.GetChainConfig(0)
+	chainB := s.configurer.GetChainConfig(1)
+
+	//node, err := chainA.GetDefaultNode()
+	//s.NoError(err)
+	// This doesn't work. Why?
+	//node.StoreWasmCode("rate_limiter.wasm", initialization.ValidatorWalletName)
```
ValarDragon, line 58: I'd guess that its an issue with the rate_limiter.wasm filepath. Golang measures filepaths relative to where the binary is being executed

tests/e2e/initialization/config.go
```diff
@@ -295,9 +295,6 @@ func updateBankGenesis(bankGenState *banktypes.GenesisState) {
 			},
 		},
 	})
-	if len(bankGenState.SupplyOffsets) == 0 {
-		bankGenState.SupplyOffsets = []banktypes.GenesisSupplyOffset{}
-	}
```
ValarDragon, line 300: Is this something fixed with upstream changes?

tests/e2e/e2e_test.go
ValarDragon: Oh also wouldn't work in this PR since IBC rate limit isn't wired up

x/ibc-rate-limit/testutil/chain.go
```diff
@@ -0,0 +1,96 @@
...
+// Move epochs to the future to avoid issues with minting
+func (chain *TestChain) MoveEpochsToTheFuture() {
+	epochsKeeper := chain.GetOsmosisApp().EpochsKeeper
+	ctx := chain.GetContext()
+	for _, epoch := range epochsKeeper.AllEpochInfos(ctx) {
+		epoch.StartTime = ctx.BlockTime().Add(time.Hour * 24 * 30)
+		epochsKeeper.DeleteEpochInfo(chain.GetContext(), epoch.Identifier)
+		_ = epochsKeeper.AddEpochInfo(ctx, epoch)
+	}
+}
```
ValarDragon, line 90: Hrmm why'd we have to do this?

tests/e2e/e2e_test.go
nicolaslara: yeah, I'm getting these to run on the other branch. Slowly moving forward cause I still have to run the full suite every time to get e2e to work 😂

tests/e2e/initialization/config.go
nicolaslara: That was needed before https://github.com/osmosis-labs/osmosis/pull/2308 got merged, for some reason

x/ibc-rate-limit/testutil/chain.go
nicolaslara: yes! so, on epoch, rewards are distributed, which means new tokens are created, which means the total value of the channel increases. This was messing with my calculations of percentage of channel value cause I couldn't estimate what the channel was going to be worth on the next block.

An alternative was to make my test use way more than was being minted so the rate limit was hit, but I thought it was cleaner to have the numbers stay fixed.

This may not actually be needed now that the channel value is cached ("limit to a % of what the value was last week") but it may still be useful if we change it back.

ValarDragon: Ah thats an interesting consequence! This makes sense!

This makes me think that since so much Osmo is natively locked up (e.g. staking, community pool, lockup) we should perhaps in a future iteration removing those from supply count. (Not at all important for the v1 since the primary goal is externally sourced assets)

x/ibc-rate-limit/ibc_middleware.go
```diff
@@ -0,0 +1,251 @@
...
+func (i *ICS4Middleware) WriteAcknowledgement(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet exported.PacketI, ack exported.Acknowledgement) error {
+	return i.channel.WriteAcknowledgement(ctx, chanCap, packet, ack)
+}
+
+// CalculateChannelValue The value of an IBC channel. This is calculated using the denom supplied by the sender.
+//  if the denom is not correct, the transfer should fail somewhere else on the call chain
+func (i *ICS4Middleware) CalculateChannelValue(ctx sdk.Context, denom string) sdk.Int {
+	supply := i.BankKeeper.GetSupply(ctx, denom)
+	locked := i.LockupKeeper.GetModuleLockedCoins(ctx)
+	return supply.Amount.Add(locked.AmountOf(denom))
```
ValarDragon, line 87: Err should this be a sub? I think this is double counting the locked coin amount?

x/ibc-rate-limit/rate_limit.go
```diff
@@ -0,0 +1,96 @@
...
+
+	sendPacketMsg, _ := BuildWasmExecMsg(
+		msgType,
+		sourceChannel,
+		denom,
+		channelValue,
+		amount,
+	)
+
+	contractKeeper := wasmkeeper.NewDefaultPermissionKeeper(wasmKeeper)
```
ValarDragon, line 30: I think we should be passing in this contract keeper into the module, at the app wiring layer

x/ibc-rate-limit/ibc_middleware.go
nicolaslara: I may be misinterpreting the result of `GetSupply`. The way I understood it is that when locking up, the funds would move from the bank module's addr to the lockup module's addr, so they would no longer be reported by `GetSupply`. 

If that's not the case, then yeah, we can either remove this or substract (depending on what we want)

x/ibc-rate-limit/rate_limit.go
nicolaslara: good call! that way we only run this once.

nicolaslara:
One general thing I've been thinking about here is that we assume IBC transfers succeed (i.e.: once the packet is sent, we count it). We could add a hook on the ack so to subtract it if the transfer fails on the other end (and vice-versa for recv).

I can look into this tomorrow.

x/ibc-rate-limit/rate_limit.go
```diff
@@ -0,0 +1,96 @@
...
+	_, err = contractKeeper.Sudo(ctx, contractAddr, []byte(sendPacketMsg))
+
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrRateLimitExceeded, err.Error())
+	}
+	return nil
+}
+
+type SendPacketMsg struct {
```
p0mvn, line 39: Can we define types at the top?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	if success {
+		suite.Require().NoError(err, "IBC send failed. Expected success. %s", err)
+	} else {
+		suite.Require().Error(err, "IBC send succeeded. Expected failure")
+		suite.ErrorContains(err, types.RateLimitExceededMsg, "Bad error type")
+	}
+	return r, err
+}
+
+func (suite *MiddlewareTestSuite) BuildChannelQuota(name string, duration, send_precentage, recv_percentage uint32) string {
```
p0mvn, line 152: Asking to learn - where is the value that we apply `send_percentage` and `recv_percentage` to?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	// Each user has 10% of the supply
+	supply := osmosisApp.BankKeeper.GetSupply(suite.chainA.GetContext(), sdk.DefaultBondDenom)
+	quota := supply.Amount.QuoRaw(20)
+	half := quota.QuoRaw(2)
+
+	// send 2.5% (quota is 5%)
+	suite.AssertSendSuccess(true, suite.NewValidMessage(true, half))
+
+	// send 2.5% (quota is 5%)
+	r, _ := suite.AssertSendSuccess(true, suite.NewValidMessage(true, half))
```
p0mvn, line 192: err check

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	suite.chainA.MoveEpochsToTheFuture()
+	suite.chainB = &osmosisibctesting.TestChain{
+		TestChain: suite.coordinator.GetChain(ibctesting.GetChainID(2)),
+	}
+	suite.path = NewTransferPath(suite.chainA, suite.chainB)
+	suite.coordinator.Setup(suite.path)
+}
+
+// Helpers
+func (suite *MiddlewareTestSuite) NewValidMessage(forward bool, amount sdk.Int) sdk.Msg {
```
p0mvn, line 73: Can we have a godoc please with respect to its parameters? 

It wasn't immediately apparent to me what is the meaning of forward

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	// send 2.5% (quota is 5%)
+	suite.AssertSendSuccess(true, suite.NewValidMessage(true, half))
+
+	// send 2.5% (quota is 5%)
+	r, _ := suite.AssertSendSuccess(true, suite.NewValidMessage(true, half))
+
+	// Calculate remaining allowance in the quota
+	attrs := suite.ExtractAttributes(suite.FindEvent(r.GetEvents(), "wasm"))
+	used, _ := sdk.NewIntFromString(attrs["weekly_used_out"])
+	suite.Require().Equal(used, half.MulRaw(2))
```
p0mvn, line 197: `quota` would work here too, right?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	return attrs
+}
+
+// Test rate limits are reset when the specified time period has passed
+func (suite *MiddlewareTestSuite) TestSendTransferReset() {
+	// Same test as above, but the quotas get reset after time passes
+	attrs := suite.TestSendTransferWithRateLimiting()
+	parts := strings.Split(attrs["weekly_period_end"], ".") // Splitting timestamp into secs and nanos
+	secs, _ := strconv.ParseInt(parts[0], 10, 64)
+	nanos, _ := strconv.ParseInt(parts[1], 10, 64)
```
p0mvn, line 210: let' err check these

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Sending above the quota should fail.
+	suite.AssertSendSuccess(false, suite.NewValidMessage(true, sdk.NewInt(1)))
+	return attrs
+}
+
+// Test rate limits are reset when the specified time period has passed
+func (suite *MiddlewareTestSuite) TestSendTransferReset() {
+	// Same test as above, but the quotas get reset after time passes
+	attrs := suite.TestSendTransferWithRateLimiting()
```
p0mvn, line 207: Maybe I'm just unused to this but I've never seen returning results from a test spec. Is it possible to avoid this for consistency's sake?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	suite.chainA.SenderAccount.SetSequence(suite.chainA.SenderAccount.GetSequence() + 1)
+	suite.chainB.NextBlock()
+	suite.chainB.SenderAccount.SetSequence(suite.chainB.SenderAccount.GetSequence() + 1)
+
+	// Reset time + one second
+	oneSecAfterReset := resetTime.Add(time.Second)
+	suite.coordinator.IncrementTimeBy(oneSecAfterReset.Sub(suite.coordinator.CurrentTime))
+
+	// Sending should succeed again
+	suite.AssertSendSuccess(true, suite.NewValidMessage(true, sdk.NewInt(1)))
```
p0mvn, line 224: Should we error-check here?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+	addr := suite.chainA.InstantiateContract(&suite.Suite, ``)
+	suite.chainA.RegisterRateLimitingContract(addr)
+
+	// send 1 token.
+	// If the contract doesn't have a quota for the current channel, all transfers are allowed
+	suite.AssertSendSuccess(true, suite.NewValidMessage(true, sdk.NewInt(1)))
+}
+
+// Test the contract used for these tests is the same contract used for E2E testing
+func (s *MiddlewareTestSuite) TestRateLimitingE2ETestsSetupCorrectly() {
```
p0mvn, line 266: Wdyt about importing the contract from a single location from both e2e and here?

p0mvn (commented):
Awesome work @nicolaslara . Excited to see rate limiting in prod soon. Left some comments. Please take a look

I really like the middleware tests. In addition to that, can we also add table-driven tests for each function please? While middleware tests seem exhaustive, they are more functional in nature. Having table-driven tests for each function would help to test each unit and edge case separately to let us be more confident about the change.

Also, is there a reason for exporting many methods and struct fields? Should we hide information where possible?

x/ibc-rate-limit/ibc_middleware.go
nicolaslara: that simplifies things! I'll make the change. I'm wondering if it also makes sense to move this to the contract by doing a stargate query (not available yet, but soon!)

x/ibc-rate-limit/ibc_middleware.go
```diff
@@ -0,0 +1,251 @@
...
+		WasmKeeper:    wasmKeeper,
+		BankKeeper:    bankKeeper,
+		LockupKeeper:  lockupKeeper,
+		ParamSpace:    paramSpace,
+	}
+}
+
+func (i *ICS4Middleware) SendPacket(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet exported.PacketI) error {
+	var params types.Params
+	i.ParamSpace.GetIfExists(ctx, []byte("contract"), &params)
```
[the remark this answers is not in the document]
nicolaslara: I didn't add it cause we don't have a keeper, so I wasn't sure what a good place to add it would be. Maybe the ICS4Middleware?

x/ibc-rate-limit/rate_limit.go
```diff
@@ -0,0 +1,96 @@
...
+		asJson, err = json.Marshal(msg)
+	default:
+		return "", types.BadMessage
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	return string(asJson), nil
```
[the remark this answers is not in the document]
nicolaslara: good call. This is probably a leftover from when I was printing it to debug

x/ibc-rate-limit/types/errors.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	RateLimitExceededMsg = "rate limit exceeded"
```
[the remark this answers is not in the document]
nicolaslara: yes! this is used in `channeltypes.NewErrorAcknowledgement()` which expects a string, and in tests when checking the error message. Made it a var for reuse.

x/ibc-rate-limit/types/params.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Implements params.ParamSet.
+func (p Params) ParamSetPairs() paramtypes.ParamSetPairs {
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyContractAddress, p, validateContractAddress),
+	}
+}
+
+func validateContractAddress(i interface{}) error {
```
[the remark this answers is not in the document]
nicolaslara: None of the other of the `validate` methods on params I have seen in other modules have tests. Is there a reason for having them specifically here?

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,275 @@
...
+		suite.Require().NoError(err, "IBC send failed. Expected success. %s", err)
+	} else {
+		suite.Require().Error(err, "IBC send succeeded. Expected failure")
+		suite.ErrorContains(err, types.RateLimitExceededMsg, "Bad error type")
+	}
+	return r, err
+}
+
+func (suite *MiddlewareTestSuite) BuildChannelQuota(name string, duration, send_precentage, recv_percentage uint32) string {
+	return fmt.Sprintf(`
```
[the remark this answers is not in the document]
nicolaslara: yeah, we could create a struct and marshal it to json. Mostly left it as a string here for simplicity since (1) this is only used within a test, and (2) when we call the contract from the cli we use the text json format, so it's easy to copy/paste them to test

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: in the contract. The message adds the quota to the contract which checks `send_percentage` and `recv_percentage` when calculating the capacity for the denom in the channel: https://github.com/osmosis-labs/osmosis/blob/main/x/ibc-rate-limit/contracts/rate-limiter/src/state.rs#L165-L178

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: haha, yeah 🤷‍♂️

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: yeah, it's a bit weird. Just did it to avoid either repeating the code or abstracting it into a function that's only used twice. If we abstract it, the helper function would basically be exactly what `TestSendTransferWithRateLimiting` is now, and the test just a call to that function. 

Not sure which is better, but happy to change it. Do you have a better alternative?

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: `AssertSendSuccess` does error checking internally

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: I was discussing that with @czarcas7ic. I'll probably change that so that the e2e tests copy the contract from the ibc-rate-limit module before they run. 

I don't like the idea of e2e needing to modify the fs, but I don't see many alternatives. Right now I'm also adding code that writes a param change json spec to the fs when running e2e tests, which I wish there was a better solution for

x/ibc-rate-limit/rate_limit.go
nicolaslara: changed in the `ibc-rate-limit` branch

x/ibc-rate-limit/rate_limit.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	var (
+		asJson []byte
+		err    error
+	)
+	switch {
+	case msgType == "send_packet":
+		msg := SendPacketMsg{SendPacket: content}
+		asJson, err = json.Marshal(msg)
```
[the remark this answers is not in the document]
nicolaslara: I don't think we can, cause then we'd have to define msg to be of a type that can be either SendPacket or RecvPacket.  Is there an easy way to do this in go?

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: This is already checked inside the call so it matches the first param. I can check it again, but we'd be duplicating logic. 

The reason for returning it is that in some tests we may want to do more checks (the error type, for example). Here it's useful to get the response so we can check the attrs, but in most cases we can just call it and ignore the results.

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: Added some docs. Maybe "forward" is not the clearest name for the param.

x/ibc-rate-limit/ibc_middleware_test.go
p0mvn: I might not be having the full context here but why is e2e modifying the fs a concern for using the same contract path?

x/ibc-rate-limit/ibc_middleware_test.go
p0mvn: I see, makes sense. I still think handling errors any time they are returned is better in case the callee logic changes. Not feeling strongly about this though so feel free to resolve if you disagree

x/ibc-rate-limit/ibc_middleware_test.go
p0mvn: That makes sense, ty!

x/ibc-rate-limit/ibc_middleware_test.go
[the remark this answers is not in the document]
p0mvn: I generally found the strings to be hard to manage in tests in the SDK, especially when the interfaces keep getting updated and when there are many of them. However, not feeling strongly about this, and the 2) point makes sense so let's try the strings

x/ibc-rate-limit/types/params.go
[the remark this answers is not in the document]
p0mvn: In my opinion, It isn't great that the some of the modules are missing these tests. We've run into issues numerous times where some value in parameters or genesis state would be invalid during init / export genesis, causing hard-to-debug problems on testnets. So covering these with tests would be helpful.

It seems that for some of the newer modules, we have been trying to maximize test coverage even for params validation:
https://github.com/osmosis-labs/osmosis/blob/c676404b6569f7d4547761a19fb7539973facf1f/x/twap/types/export_test.go#L35

x/ibc-rate-limit/rate_limit.go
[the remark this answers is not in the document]
p0mvn: Makes sense!

A follow-up question: is there a reason for defining these messages as Go structs instead of generating them via proto files?

x/ibc-rate-limit/ibc_middleware.go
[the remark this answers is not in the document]
p0mvn: I think `ICS4Middleware` makes sense here

x/ibc-rate-limit/ibc_middleware.go
```diff
@@ -0,0 +1,251 @@
...
+	_ porttypes.ICS4Wrapper = &ICS4Middleware{}
+)
+
+type ICS4Middleware struct {
+	channel       porttypes.ICS4Wrapper
+	accountKeeper *authkeeper.AccountKeeper
+	BankKeeper    *bankkeeper.BaseKeeper
+	WasmKeeper    *wasmkeeper.Keeper
+	LockupKeeper  *lockupkeeper.Keeper
+	ParamSpace    paramtypes.Subspace
```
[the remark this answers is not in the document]
p0mvn: I tend to be a proponent of the other camp where we hide as much information as possible.

In my opinion, this helps with encapsulating all internal components and leads to an easier experience with reasoning about an abstraction.

Also, I understand the "we'are all adults" idea if it is in the scope of a limited group of team members working on the project. Since we are in the open source where there are many contributors with varying commitments as well as reviewers with limited capacity, I would prefer to err on the side of caution and avoid exposing interfaces that can be private.

Also, this middleware seems to be a similar abstraction to a `keeper`. As far as I know, most of our keepers aim to maximize information hiding by unexporting fields that are unused in other packages. So it would be great to aim to keep middleware consistent with keepers.

x/ibc-rate-limit/rate_limit.go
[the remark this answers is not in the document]
nicolaslara: These messages are defined in rust in the contract; the struct here is just a go representation of the message. For tests, we are probably going to be able to write those directly in rust and have the test use the chain as a packege. But I don't think there's a solution for dealing with this when calling the contract from the chain. 

An options is to try to auto generate the go types from the rust definitions using the json schema; I don't think this is necessary here since the messages are so simple (and so few), but we may want to do something like that for more complex contracts.

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: I generally would agree, but since we want to check different type of errors it was easier to encapsulate it.

To be fair, I've had to override some of the ibctesting functions for this exact reason: they assert NoError inside the function and I want to ignore failures (cause failures is what I'm testing). 

I'll see if there's a clean way to handle these outside and otherwise resolve this

x/ibc-rate-limit/ibc_middleware.go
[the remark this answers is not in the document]
nicolaslara: yeah, that makes sense. I think if osmosis was a library that others will import (the sdk, wasmd, etc) I'd feel more strongly about allowing implementors to change internals, but for this it makes sense to clearly signal which parts of the internals are stable and which aren't. I'll see how much I can hide.

ValarDragon:
qq, is uploading the .wasm file intentional?

ValarDragon (approved):
I've reviewed the go code, LGTM! Haven't reviewed latest tests yet, but I can review post-merge

nicolaslara:
> qq, is uploading the .wasm file intentional? 

@ValarDragon Yes! the wasm file is used in the tests. To be fair, we shouldn't really have tests in this PR since they will fail for lack of integration; it probably makes sense to either remove them from this branch or just merge https://github.com/osmosis-labs/osmosis/pull/2410 instead of this.

x/ibc-rate-limit/ibc_middleware.go
```diff
@@ -0,0 +1,320 @@
...
+	return contract
+}
+
+// CalculateChannelValue The value of an IBC channel. This is calculated using the denom supplied by the sender.
+// if the denom is not correct, the transfer should fail somewhere else on the call chain
+func (i *ICS4Middleware) CalculateChannelValue(ctx sdk.Context, denom string) sdk.Int {
+	return i.BankKeeper.GetSupplyWithOffset(ctx, denom).Amount
+}
+
+type IBCModule struct {
```
p0mvn, line 100: Wdyt about defining this in a separate file `ibc_module.go`?

Noticed the suggestion here:
https://github.com/cosmos/ibc-go/blob/f57170b1d4dd202a3c6c1c61dcf302b6a9546405/docs/ibc/middleware/develop.md?plain=1#L58

which seems to help split the abstractions per file

x/ibc-rate-limit/rate_limit.go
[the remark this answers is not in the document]
p0mvn: Gotcha. Thanks

x/ibc-rate-limit/ibc_middleware.go
nicolaslara: sure! I find it more clear to have the whole middleware in the same file, but no strong preference, so I'll separate them

x/ibc-rate-limit/ibc_middleware_test.go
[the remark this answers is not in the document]
nicolaslara: tried this but needed a lot of boilerplate for type conversions, so leaving as strings for now

x/ibc-rate-limit/types/params.go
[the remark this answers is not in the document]
nicolaslara: @p0mvn got any idea why the tests (https://github.com/osmosis-labs/osmosis/blob/20b40deacc9143cc0b9393def410e8b84e4bd759/x/ibc-rate-limit/types/params_test.go) would expect the cosmos bech32 prefix instead of osmo? I still think it makes sense to have the test, but it's driving me crazy that the setup is wrong

ValarDragon:
(I'm happy with merging this once lint passes)

x/ibc-rate-limit/ibc_middleware.go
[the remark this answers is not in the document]
nicolaslara: Moved there. Marking as resolved.

x/ibc-rate-limit/ibc_middleware.go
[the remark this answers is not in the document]
nicolaslara: changed.

x/ibc-rate-limit/types/params.go
```diff
@@ -0,0 +1,65 @@
...
+
+// default gamm module parameters.
+func DefaultParams() Params {
+	return Params{
+		ContractAddress: "",
+	}
+}
+
+// validate params.
+func (p Params) Validate() error {
```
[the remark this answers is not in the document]
nicolaslara: Added a simple test, though I'm having the same issue of the prefix being cosmos instead of osmo. I've spent a while debugging this and can't find the cause, but I'm not too worried about it

nicolaslara:
Cleaned up this branch so it's ready to merge. The tests will be part of the integration PR, as it doesn't make sense to have broken tests here.

PR #2410, opened 2022-08-15
IBC Rate Limit - integration and tests

nicolaslara:
Closes: #2305 

Integrates IBC rate limiting with the chain.  

## Brief Changelog

 * An IBC middleware is now provided and registered with the IBC transfer module.
 * The level of rate limiting is controlled by a rate limiting contract that can be controlled by an address specified by governance (this could be the gov module itself)
 * The "governance address" of the rate limiting contract can reset the quota when a rate limit is hit (this could be the gov module itself)

## Testing and Verifying

 * Added integration tests for the middleware (IBC Message <-> IBC transfer module <-> RateLimitingMiddlware <-> Relayer <-> Counterparty chain) under  `x/ibc-rate-limit`
 * Added tests to the contract (see https://github.com/osmosis-labs/osmosis/pull/2408)

This PR depends on @2408 and @2409. It also requires a change to the osmosis fork of the cosmos-sdk (https://github.com/osmosis-labs/osmosis/issues/2307) 

     - If the rate limit is reached, IBC transactions will be blocked for users. 
     - Should I add this?
      -  Currently undocummented. Should I add docs for this? what is the right place for that?


app/keepers/keepers.go
```diff
@@ -226,7 +247,7 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 	// Create static IBC router, add transfer route, then set and seal it
 	ibcRouter := porttypes.NewRouter()
 	ibcRouter.AddRoute(icahosttypes.SubModuleName, icaHostIBCModule).
-		AddRoute(ibctransfertypes.ModuleName, transferIBCModule)
+		AddRoute(ibctransfertypes.ModuleName, &rateLimitingTransferModule)
```
ValarDragon, line 250: I'm confused about this wiring. We make our middleware the transfer module, but we pass the middleware into the transfer keeper as well?

x/ibc-rate-limit/types/errors.go
```diff
@@ -5,8 +5,7 @@ import (
 )
 
 var (
-	RateLimitExceededMsg = "rate limit exceeded"
-	ErrRateLimitExceeded = sdkerrors.Register(ModuleName, 2, RateLimitExceededMsg)
+	ErrRateLimitExceeded = sdkerrors.Register(ModuleName, 2, "rate limit exceeded")
 	ErrBadMessage        = sdkerrors.Register(ModuleName, 3, "bad message")
 	ErrContractError     = sdkerrors.Register(ModuleName, 4, "contract error")
```
ValarDragon, line 10: Lets not register any of these codes imo, and just do normal errors. I think we should move away from code spaced errors for now

ValarDragon:
wdyt about splitting out:
- `tests/e2e/configurer/`
- `tests/e2e/container/`

+ some of the other small changes into another PR that we can merge pretty quickly?

app/keepers/keepers.go
nicolaslara: yeah, this is confusing, but that's how you wire up the middlewares. What we pass into the keeper is not the middleware, but the ICS4Wrapper which would be the middleware's keeper if we had one (Notice that the transfer keeper has two params that originally pointed to the channel: ics4Wrapper and channelKeeper; we override the first one)

So the ibc module's methods (the OnSomething methods) all go to the part implementing the IBCModule interface, but the ICS4 methods (SendPacket and WriteAck) go to the ICS4Wrapper (which ends up in the channel). This is because things go in one direction on send and the other on receive. So we get:

SendPacket -> Transfer -> RateLimit -> channel

OnRecv -> RateLimit -> transfer -> channel


But I'll double check this to make sure I'm wrapping it correctly

ValarDragon: hrmm, I somewhat think I get it? So middleware from IBC is currently insufficient to do everything we need, since we want rate limit code before other code in the receive code path?

(Not sure why it couldn't be in transfer first though, since as long as we error anywhere, it should get atomically reverted)

nicolaslara: No, this is not a workaround to get our use case to work; this is unfortunately the way the middleware wiring us designed to be used. The transfer IBCModule doesn't implement SendPacket or WriteAcknowledgement, as it only [implements](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L29) porttypes.IBCModule and not porttypes.ICS4Wrapper ([definitions](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/module.go)). It instead relies on the transfer module's keeper to have an [ICS4Wrapper](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/keeper.go#L23-L24) that is [initialized](https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go#L423) to either the channelKeeper (which implements ICS4Wrapper) or the ICS4Wrapper implementation of the middleware.

The middleware [interface](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/module.go#L132) from IBC requires users to implement both porttypes.IBCModule and porttypes.ICS4Wrapper, but the later can't really wrap SendPacket or WriteAcknowledgement because they don't exist in the transfer ibc module. 

I think the reason for designing it this way is so that a middleware would behave the same way as if it was a function wrapping another: what originates from the outside (other chain) must flow inwards and what originates in the last function called (our chain) must flow outwards. 

It's probably easier if I go over the code in a call though, so let me know so we can screenshare

nicolaslara:
Created https://github.com/osmosis-labs/osmosis/pull/2941 to add the testing helpers needed by this branch. That one should be a quick merge and make this PR smaller

app/app.go
```diff
@@ -130,6 +130,7 @@ type OsmosisApp struct {
 	mm           *module.Manager
 	sm           *simtypes.Manager
 	configurator module.Configurator
+	txConfig     client.TxConfig
```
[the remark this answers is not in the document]
nicolaslara: done!

x/ibc-rate-limit/types/errors.go
nicolaslara: You mean just use the strings? Eventually we have to convert it to a `NewError` right? which requires the code

ancazamfir:
Hi, i was looking at the osmosis rate limiter code and have some questions. Maybe this PR is not the best to ask but will do anyway and please redirect. On https://github.com/osmosis-labs/osmosis/blob/4fed7db03b4d41aef68af3669f99e2cab0e32ace/x/ibc-rate-limit/ibc_module.go#L117-L121
`GetFundsFromPacket` gets the denom as specified by the sender. On receive side (in the code above) shouldn't there be un-prefexing or prefexing performed depending on wether the receiving chain was the one originally sending the packet or not? Similar thing is done in the transfer module https://github.com/cosmos/ibc-go/blob/f693f6c11053192f369b5908c9186bd494c3a33d/modules/apps/transfer/keeper/relay.go#L165

The way it is the channel value is computed for a denom of the counterparty/ sender chain. Or maybe I am missing something important here.

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,351 @@
...
+		accountFrom = suite.chainA.SenderAccount.GetAddress().String()
+		accountTo = suite.chainB.SenderAccount.GetAddress().String()
+	} else {
+		coins = transfertypes.GetTransferCoin(
+			suite.path.EndpointB.ChannelConfig.PortID,
+			suite.path.EndpointB.ChannelID,
+			sdk.DefaultBondDenom,
+			sdk.NewInt(1),
+		)
+		coins = sdk.NewCoin(sdk.DefaultBondDenom, amount)
```
ancazamfir, line 92: why the overwrite and not:
```suggestion
		coins = transfertypes.GetTransferCoin(
			suite.path.EndpointB.ChannelConfig.PortID,
			suite.path.EndpointB.ChannelID,
			sdk.DefaultBondDenom,
			amount,
		)
```

nicolaslara: This got overlooked. Thanks for the catch!

nicolaslara: The correct code here is `coins = sdk.NewCoin(sdk.DefaultBondDenom, amount)` because that message is sent from chainB to chainA. What you said about getting the funds still applies though: I was not wrapping the token before getting the value and this was only working because both chains in my tests were the same chain.

I've fixed that in my latest commit (032f33b15feae8f40d3b088bd6d2bdb90d28a2bb)

nicolaslara:
Modified this implementation to use the escrowed amount as the channel value for native tokens. This introduces a bootstrapping problem: channels need to have tokens on escrow before we initialize their quotas (otherwise the quota will always be lower than the sent amount). This can be seen in the tests.

I don't think this is a problem for our use case as we will only use rate limiting for existing channels, but it may become important if someone wants all channels to have rate limiting, or to auto-initialize rate-limiting for channels on creation

ancazamfir:
> Modified this implementation to use the escrowed amount as the channel value for native tokens. This introduces a bootstrapping problem: channels need to have tokens on escrow before we initialize their quotas (otherwise the quota will always be lower than the sent amount). This can be seen in the tests.
> 
> I don't think this is a problem for our use case as we will only use rate limiting for existing channels, but it may become important if someone wants all channels to have rate limiting, or to auto-initialize rate-limiting for channels on creation

Right. Another possibility is to set some threshold for which the ratelimiter activates/ deactivates.

x/ibc-rate-limit/ibc_middleware_test.go
nicolaslara: Yeah, this is testing native B tokens being sent to a. But maybe I should also add a test for native A tokens being sent "back" from B to A

x/ibc-rate-limit/ics4_wrapper.go
```diff
@@ -84,6 +91,12 @@ func (i *ICS4Wrapper) GetParams(ctx sdk.Context) (contract string) {
...
-func (i *ICS4Wrapper) CalculateChannelValue(ctx sdk.Context, denom string) sdk.Int {
-	return i.bankKeeper.GetSupplyWithOffset(ctx, denom).Amount
+func (i *ICS4Wrapper) CalculateChannelValue(ctx sdk.Context, denom string, packet exported.PacketI) sdk.Int {
+	// If the source is the counterparty chain, this should be
+	if strings.HasPrefix(denom, "ibc/") {
+		return i.bankKeeper.GetSupplyWithOffset(ctx, denom).Amount
+	}
+
+	escrowAddress := transfertypes.GetEscrowAddress(packet.GetSourcePort(), packet.GetSourceChannel())
+	return i.bankKeeper.GetBalance(ctx, escrowAddress, denom).Amount
```
ancazamfir, line 101: At first look this doesn't seem correct. If we have something like:
```
 A            B              C
ch0  --- ch1    ch2 --- ch3
```
And `a` is chain A token. In the following cases:
1. B sends to C a packet with denom `ch1/a` over `ch2`. denom here is `ch1/a`. 
2. B receives (back) from C a packet with denom `ch3/ch1/a` over `ch2`. denom here is `ch1/a` (unprefixed)

Here I think we get the `ch1/a` supply instead of the balance of `escrow_address(ch2)` for denom `ch1/a`.

Maybe something along these lines:
```suggestion
func (i *ICS4Wrapper) CalculateChannelValue(ctx sdk.Context, denom string, packet exported.PacketI, send bool) sdk.Int {
	prefix_for_source:= transfertypes.GetDenomPrefix(packet.GetSourcePort(), packet.GetSourceChannel())
	if send { // send packet
		if !strings.HasPrefix(denom, prefix_for_source) {
			// Sender chain is the source
			escrowAddress := transfertypes.GetEscrowAddress(packet.GetSourcePort(), packet.GetSourceChannel())
			return i.bankKeeper.GetBalance(ctx, escrowAddress, denom).Amount
		} else {
			return i.bankKeeper.GetSupplyWithOffset(ctx, denom).Amount
		}
	} else { // receive packet
		if strings.HasPrefix(denom, prefix_for_source) {
			// Receiver chain is the source
			escrowAddress := transfertypes.GetEscrowAddress(packet.GetDestPort(), packet.GetDestChannel())
			return i.bankKeeper.GetBalance(ctx, escrowAddress, denom).Amount
		} else {
			return i.bankKeeper.GetSupplyWithOffset(ctx, denom).Amount
		}
	}
```

It would be great to have tests to cover scenarios like above.

nicolaslara: The current behaviour (what you described above) is what I was trying to achieve, where escrow value is *only* used for native tokens (osmo/ion/tokenfactory tokens). ch0/a, as an ibc token, should be rate limited based on the current  amount of ch0/a on osmosis. 

I'm open to changing it to what you suggest, but I think the volume of some token/channel combinations may be too low and make the limit to be hit too often

ancazamfir: > The current behaviour (what you described above) is what I was trying to achieve, where escrow value is _only_ used for native tokens (osmo/ion/tokenfactory tokens). ch0/a, as an ibc token, should be rate limited based on the current amount of ch0/a on osmosis.
> 
I see, thanks for the clarification.

> I'm open to changing it to what you suggest, but I think the volume of some token/channel combinations may be too low and make the limit to be hit too often

But in this case ratelimiting should not be configured on those channels as they have a low value.
Are there osmosis uscases like the ones I mention above (the multi hop/channel transfers)?

PR #2417, opened 2022-08-15
refactor(x/mint): unexport keeper methods

p0mvn:
Closes: #XXX

Currently, a lot of the keeper methods are exported when they are only used inside the keeper package or in tests.

For the purposes of information hiding, we should unexport as many methods as possible to avoid misuse from incorrect locations.

There are no changes other than unexporting some methods and none of the methods were removed.

## Brief Changelog

- unexport `setLastReductionEpochNum`
- unexport `getLastReductionEpochNum`
- unexport `createDeveloperVestingModuleAccount`
- unexport `mintCoins`
- create a helper in `apptesting` for minting tokens in tests
- reshuffle order in mint `keeper.go` so that exported methods go first, then unexported

## Testing and Verifying

### DIFF
--- a/x/mint/keeper/keeper.go
+++ b/x/mint/keeper/keeper.go
@@ -108,7 +108,7 @@ func (k *Keeper) SetHooks(h types.MintHooks) *Keeper {
 	return k
 }
 
-// GetMinter get the minter.
+// GetMinter gets the minter.
 func (k Keeper) GetMinter(ctx sdk.Context) (minter types.Minter) {
 	store := ctx.KVStore(k.storeKey)
 	b := store.Get(types.MinterKey)
--- a/x/mint/keeper/keeper.go
+++ b/x/mint/keeper/keeper.go
@@ -92,34 +92,6 @@ func (k Keeper) SetInitialSupplyOffsetDuringMigration(ctx sdk.Context) error {
 	return nil
 }
 
-// CreateDeveloperVestingModuleAccount creates the developer vesting module account
-// and mints amount of tokens to it.
-// Should only be called during the initial genesis creation, never again. Returns nil on success.
-// Returns error in the following cases:
-// - amount is nil or zero.
-// - if ctx has block height greater than 0.
-// - developer vesting module account is already created prior to calling this method.
-func (k Keeper) CreateDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.Coin) error {
-	if amount.IsNil() || amount.Amount.IsZero() {
-		return sdkerrors.Wrap(types.ErrAmountNilOrZero, "amount cannot be nil or zero")
-	}
-	if k.accountKeeper.HasAccount(ctx, k.accountKeeper.GetModuleAddress(types.DeveloperVestingModuleAcctName)) {
-		return sdkerrors.Wrapf(types.ErrModuleAccountAlreadyExist, "%s vesting module account already exist", types.DeveloperVestingModuleAcctName)
-	}
-
-	moduleAcc := authtypes.NewEmptyModuleAccount(
-		types.DeveloperVestingModuleAcctName, authtypes.Minter)
-	k.accountKeeper.SetModuleAccount(ctx, moduleAcc)
-
-	err := k.bankKeeper.MintCoins(ctx, types.DeveloperVestingModuleAcctName, sdk.NewCoins(amount))
-	if err != nil {
-		return err
-	}
-	return nil
-}
-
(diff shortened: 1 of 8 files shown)


x/mint/keeper/keeper.go
```diff
@@ -92,34 +92,6 @@ func (k Keeper) SetInitialSupplyOffsetDuringMigration(ctx sdk.Context) error {
...
 }
 
-// CreateDeveloperVestingModuleAccount creates the developer vesting module account
-// and mints amount of tokens to it.
-// Should only be called during the initial genesis creation, never again. Returns nil on success.
-// Returns error in the following cases:
-// - amount is nil or zero.
-// - if ctx has block height greater than 0.
-// - developer vesting module account is already created prior to calling this method.
-func (k Keeper) CreateDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.Coin) error {
```
p0mvn, line 102: Note to reviewers: some functions in this file were shuffled around so that exported methods go first, and, then, unexported follow

x/mint/keeper/keeper.go
```diff
@@ -185,19 +138,6 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
...
-func (k Keeper) MintCoins(ctx sdk.Context, newCoins sdk.Coins) error {
-	if newCoins.Empty() {
-		// skip as no coins need to be minted
-		return nil
-	}
-
-	return k.bankKeeper.MintCoins(ctx, types.ModuleName, newCoins)
-}
-
 // DistributeMintedCoins implements distribution of minted coins from mint to external modules.
```
czarcas7ic, line 141:
```suggestion
// DistributeMintedCoin implements distribution of a minted coin from mint to external modules.
```

x/mint/keeper/keeper.go
```diff
@@ -338,3 +306,29 @@ func getProportions(mintedCoin sdk.Coin, ratio sdk.Dec) (sdk.Coin, error) {
 	}
 	return sdk.NewCoin(mintedCoin.Denom, mintedCoin.Amount.ToDec().Mul(ratio).TruncateInt()), nil
 }
+
+// CreateDeveloperVestingModuleAccount creates the developer vesting module account
```
czarcas7ic, line 310:
```suggestion
// createDeveloperVestingModuleAccount creates the developer vesting module account
```

x/mint/keeper/keeper.go
```diff
@@ -165,15 +120,13 @@ func (k Keeper) GetMinter(ctx sdk.Context) (minter types.Minter) {
 	return
 }
 
-// set the minter.
+// SetMinter set the minter.
```
czarcas7ic, line 123:
```suggestion
// SetMinter sets the minter.
```

x/mint/keeper/keeper.go
```diff
@@ -136,24 +108,7 @@ func (k *Keeper) SetHooks(h types.MintHooks) *Keeper {
...
-}
-
-// SetLastReductionEpochNum set last reduction epoch number.
-func (k Keeper) SetLastReductionEpochNum(ctx sdk.Context, epochNum int64) {
-	store := ctx.KVStore(k.storeKey)
-	store.Set(types.LastReductionEpochKey, sdk.Uint64ToBigEndian(uint64(epochNum)))
-}
-
-// get the minter.
+// GetMinter get the minter.
```
czarcas7ic, line 111:
```suggestion
// GetMinter gets the minter.
```

czarcas7ic (approved):
LGTM. Is there any specific methodology you go through to determine which keeper methods should stay unexported? I think it may be helpful to document this somewhere so we can easily determine in the future (and possibly automate warnings!)

p0mvn:
> LGTM. Is there any specific methodology you go through to determine which keeper methods should stay unexported? I think it may be helpful to document this somewhere so we can easily determine in the future (and possibly automate warnings!)

As a general rule, we should be hiding as many methods as possible without a specific use case outside of the internal logic. More on [the principle](https://en.wikipedia.org/wiki/Information_hiding).

For example, with `x/mint` no other module should be able to `MintTokens` or `CreateDeveloperVestingModuleAccount` so it makes sense to hide it. 

We should be encapsulating logic into stable interfaces and exposing them to the clients. In the context of `x/mint` keeper, the core of these stable interfaces is `AfterEpochEnd` hook + some other boilerplate methods that all keepers implement. Everything else is internal logic and should be encapsulated by the keeper.

Issue #2208, opened 2022-07-23
TWAP: Make integration tests for relevant methods working on 3 asset pools


ValarDragon:
## Background

Most of the TWAP tests, are centered around 2 asset pools. All of the store management and code is written to explicitly work with more pairs, and _should_ work, with a few unit tests for this.

We should make multi-asset pool tests at the api_test and hook_test layers.

## Acceptance Criteria

- [x] Multi asset pool test for Endblock flow
- [x] Multi asset pool test for `GetArithmeticTwap`

mattverse:
Blocked on https://github.com/osmosis-labs/osmosis/pull/2340

mattverse:
- [ ] https://github.com/osmosis-labs/osmosis/pull/2364

czarcas7ic:
Closing in favor of #2538

PR #2420, opened 2022-08-16
twap: Add Multi-Pool hooks test for `AfterPoolCreated`

mattverse:
Sub component of : https://github.com/osmosis-labs/osmosis/issues/2208

Adds multi-pool(balancer 3 asset pool) test for `AfterPoolCreated` hooks.
Most of the file differences come from renaming `PrepareUni2PoolWithAssets` to `PrepareUni2PoolWithCoins`, as I thought we had inconsistency with the naming conventions.

## Brief Changelog
- Adds tests for three asset pool  for the `AfterPoolCreated` hook


x/twap/listeners_test.go
```diff
@@ -3,35 +3,99 @@ package twap_test
...
+
+			// check that creating a pool saved a new state of changed pools
+			s.Require().Equal([]uint64{poolId}, s.twapkeeper.GetChangedPools(s.Ctx))
+
+			// check that all twap records have been created for all denom pairs in pool
+			twapRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, poolId)
+			s.Require().NoError(err)
+			denoms, err := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, poolId)
+			s.Require().NoError(err)
+			denomPairs0, _ := types.GetAllUniqueDenomPairs(denoms)
```
czarcas7ic, line 75: What is the significance of the 0 in `denomPairs0`?

x/twap/listeners_test.go
```diff
@@ -42,7 +106,7 @@ func (s *TestSuite) TestSwapTriggeringTrackPoolId() {
 // TODO: Abstract this to be more table driven, and test more pool / block setups.
 func (s *TestSuite) TestSwapAndEndBlockTriggeringSave() {
```
czarcas7ic, line 107: Do we benefit from testing both UniV2 and multi-asset pools here as well? (asking to learn, unsure if it would provide value)

czarcas7ic (approved):
Was able to follow the tests fairly easily, well done 🙂

app/apptesting/gamm.go
```diff
@@ -14,22 +14,37 @@ var DefaultAcctFunds sdk.Coins = sdk.NewCoins(
 	sdk.NewCoin("baz", sdk.NewInt(10000000)),
 )
 
-// PrepareUni2PoolWithAssets returns a Univ2 pool with the initial liquidity being the provided balances.
-func (s *KeeperTestHelper) PrepareUni2PoolWithAssets(asset1, asset2 sdk.Coin) uint64 {
+// PrepareUni2PoolWithCoins returns a Univ2 pool with the initial liquidity being the provided balances.
+func (s *KeeperTestHelper) PrepareUni2PoolWithCoins(coin1, coin2 sdk.Coin) uint64 {
```
p0mvn, line 18: Can we use `PrepareBalancerPoolWithCoins` with 2 coins instead?

mattverse: Didn't know folks would be down for this change as well :) Im totally down to changing this

x/twap/listeners_test.go
mattverse: `GetAllUniqueDenomPairs` would return two slices of string, the first slice indicating the first asset of the pair, the second slice indicating the second asset of the pair. We would use both in normal cases, denomPairs0[1], denomParis1[1] or something similar

x/twap/listeners_test.go
mattverse: yes! These are WIP, I wanted to have them in separate PR so that the PR doesn't get too big

app/apptesting/gamm.go
mattverse: done!

x/twap/keeper_test.go
```diff
@@ -14,11 +14,13 @@ import (
...
 var (
-	denom0            = "token/B"
-	denom1            = "token/A"
-	defaultUniV2Coins = sdk.NewCoins(sdk.NewInt64Coin(denom0, 1_000_000_000), sdk.NewInt64Coin(denom1, 1_000_000_000))
-	baseTime          = time.Unix(1257894000, 0).UTC()
-	tPlusOne          = baseTime.Add(time.Second)
+	denom0                 = "token/B"
+	denom1                 = "token/A"
+	denom2                 = "token/C"
+	defaultUniV2Coins      = sdk.NewCoins(sdk.NewInt64Coin(denom0, 1_000_000_000), sdk.NewInt64Coin(denom1, 1_000_000_000))
```
p0mvn, line 20: nit: Consider renaming this to `defaultTwoAssetCoins` for a clearer distinction and relationship to  `defaultThreeAssetCoins`

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+	three_asset_coins := defaultThreeAssetCoins
+
+	tests := []struct {
+		name      string
+		poolCoins sdk.Coins
+		// if this field is set true, we swap in the same block with pool creation
+		runSwap bool
+	}{
+		{
+			"Uni2 Pool, no swap on pool creation block",
```
p0mvn, line 25: nit: for anyone unfamiliar with what univ2 means
```suggestion
			"two asset pool, no swap on pool creation block",
```

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+			two_asset_coin,
+			false,
+		},
+		{
+			"Uni2 Pool, swap on pool creation block",
```
p0mvn, line 30:
```suggestion
			"two asset pool, swap on pool creation block",
```

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
-func (s *TestSuite) TestSwapTriggeringTrackPoolId() {
-	poolId := s.PrepareUni2PoolWithAssets(defaultUniV2Coins[0], defaultUniV2Coins[1])
-	s.BeginNewBlock(false)
-	s.RunBasicSwap(poolId)
-
-	s.Require().Equal([]uint64{poolId}, s.twapkeeper.GetChangedPools(s.Ctx))
+// TestAfterPoolCreatedHook tests if internal tracking logic has been triggered correctly,
+// and the corerct state entries have been created upon pool creation.
+// This test includes test cases for swapping on the same block with pool creation.
+func (s *TestSuite) TestAfterPoolCreatedHook() {
```
p0mvn, line 14: This is named `TestAfterPoolCreatedHook` but I don't see a call to `AfterPoolCreated`

What exactly is being tested here?

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
-
-// Tests that after a swap, we are triggering internal tracking logic for a pool.
-func (s *TestSuite) TestSwapTriggeringTrackPoolId() {
-	poolId := s.PrepareUni2PoolWithAssets(defaultUniV2Coins[0], defaultUniV2Coins[1])
-	s.BeginNewBlock(false)
-	s.RunBasicSwap(poolId)
-
-	s.Require().Equal([]uint64{poolId}, s.twapkeeper.GetChangedPools(s.Ctx))
+// TestAfterPoolCreatedHook tests if internal tracking logic has been triggered correctly,
+// and the corerct state entries have been created upon pool creation.
```
p0mvn, line 12:
```suggestion
// and the correct state entries have been created upon pool creation.
```

x/twap/listeners_test.go
p0mvn: I see - this is done inside `PrepareBalancerPoolWithCoins` that calls `CreatePool`.

Can we add a comment explaining that please?

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+// This test includes test cases for swapping on the same block with pool creation.
+func (s *TestSuite) TestAfterPoolCreatedHook() {
+	two_asset_coin := defaultUniV2Coins
+	three_asset_coins := defaultThreeAssetCoins
```
p0mvn, line 16: On the earlier note, do we have a test that calls `afterCreatePool` directly and primarily focuses on testing just that? If not, can we please add it?

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+			// check that creating a pool saved a new state of changed pools
+			s.Require().Equal([]uint64{poolId}, s.twapkeeper.GetChangedPools(s.Ctx))
+
+			// check that all twap records have been created for all denom pairs in pool
+			twapRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, poolId)
+			s.Require().NoError(err)
+			denoms, err := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, poolId)
+			s.Require().NoError(err)
+			denomPairs0, _ := types.GetAllUniqueDenomPairs(denoms)
+			s.Require().Equal(len(denomPairs0), len(twapRecords))
```
p0mvn, line 70: Can we make the same check on `denomPairs1` just in case?

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+// This test includes test cases for swapping on the same block with pool creation.
+func (s *TestSuite) TestAfterPoolCreatedHook() {
+	two_asset_coin := defaultUniV2Coins
+	three_asset_coins := defaultThreeAssetCoins
+
+	tests := []struct {
+		name      string
+		poolCoins sdk.Coins
+		// if this field is set true, we swap in the same block with pool creation
+		runSwap bool
```
p0mvn, line 22: Does it make sense to have `expectedTwapRecordCouns` or `expectedTwapRecords` here so that we can have more specific assertions? 

Currently, I think that most assertions we make depend on another method in the internal logic, and if both break, we might get a false positive

app/apptesting/gamm.go
```diff
@@ -14,20 +14,19 @@ var DefaultAcctFunds sdk.Coins = sdk.NewCoins(
...
-				Token:  asset2,
-			},
-		},
-	)
+// PrepareBalancerPoolWithCoins returns a balancer pool
+// consisted of given coins with equal weight.
+func (s *KeeperTestHelper) PrepareBalancerPoolWithCoins(coins ...sdk.Coin) uint64 {
+	var poolAssets []balancer.PoolAsset
+	for _, coin := range coins {
+		poolAsset := balancer.PoolAsset{
```
alexanderbez, line 22: nit: no need to assign to a variable

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
-
-	twap, err = s.twapkeeper.GetRecordAtOrBeforeTime(s.Ctx, poolId, s.Ctx.BlockTime(), denom0, denom1)
-	s.Require().NoError(err)
-	s.Require().Equal(expectedTwap, twap)
-}
-
-// TODO: Write test for create pool, swap, and EndBlock in same block, we use post-swap spot price in record.
-
-// Tests that after a swap, we are triggering internal tracking logic for a pool.
-func (s *TestSuite) TestSwapTriggeringTrackPoolId() {
```
ValarDragon, line 29: Why was this deleted?

ValarDragon: we need tests for swaps against already existing pools

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+			var poolId uint64
+			// prepare pool according to test case
+			poolId = s.PrepareBalancerPoolWithCoins(tc.poolCoins...)
+
+			if tc.runSwap {
+				s.RunBasicSwap(poolId)
+			}
+
+			timeBeforeBeginBlock := s.Ctx.BlockTime()
+			s.twapkeeper.EndBlock(s.Ctx)
```
ValarDragon, line 58: why don't we run s.EndBlock() ? (or does no such method exist and we should write an issue?)

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+				s.Require().Equal(asset0sp, twapRecord.P0LastSpotPrice)
+				s.Require().Equal(timeBeforeBeginBlock, twapRecord.Time)
+
+				// check that we have same state entry
+				storeRepresentationRecord, err := s.twapkeeper.GetMostRecentRecordStoreRepresentation(s.Ctx, poolId, twapRecord.Asset0Denom, twapRecord.Asset1Denom)
+				s.Require().Equal(twapRecord, storeRepresentationRecord)
+
+				twapRecordBeforeTime, err := s.twapkeeper.GetRecordAtOrBeforeTime(s.Ctx, poolId, s.Ctx.BlockTime(), twapRecord.Asset0Denom, twapRecord.Asset1Denom)
+				s.Require().NoError(err)
+				s.Require().Equal(twapRecord, twapRecordBeforeTime)
```
ValarDragon, line 89: Can we just construct the expected TWAP record, and then check equality with one representation?

I don't think we need to check all representations, or if we do that should be a separate method - its a guarantee elsewhere in the code that all methods update together.

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+			if tc.runSwap {
+				s.RunBasicSwap(poolId)
+			}
+
+			timeBeforeBeginBlock := s.Ctx.BlockTime()
+			s.twapkeeper.EndBlock(s.Ctx)
+			s.BeginNewBlock(false)
+
+			// check that creating a pool saved a new state of changed pools
+			s.Require().Equal([]uint64{poolId}, s.twapkeeper.GetChangedPools(s.Ctx))
```
ValarDragon, line 62: err why didnt GetChangedPools get wiped on EndBlock?

x/twap/listeners_test.go
```diff
@@ -3,35 +3,93 @@ package twap_test
...
+// TestAfterPoolCreatedHook tests if internal tracking logic has been triggered correctly,
+// and the corerct state entries have been created upon pool creation.
+// This test includes test cases for swapping on the same block with pool creation.
+func (s *TestSuite) TestAfterPoolCreatedHook() {
+	two_asset_coin := defaultUniV2Coins
+	three_asset_coins := defaultThreeAssetCoins
+
+	tests := []struct {
+		name      string
+		poolCoins sdk.Coins
```
ValarDragon, line 20: We should be able to test multiple pool creations in one block

x/twap/listeners_test.go
ValarDragon: ahh were testing the flow wrong. Needs to have commit as done in `EndBlockTriggeringSave`.

This line should be moved up before EndBlock.

x/twap/listeners_test.go
ValarDragon: idk, I feel like were mixing what were testing. The goal is to just sanity check that we have the right number of records.

x/twap/listeners_test.go
ValarDragon: Good point. In the latest commit I made it get the correct expected record as `types.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])`, which we then test equality against. (As that is our definition of correct here)

Only thing I think that would be slipping is if:
* `types.NewTwapRecord` returned something wrong for an integration test failure,
* 0 records were created, and our expected logic missed this due to a bug in
  *  `types.GetAllUniqueDenomPairs(denoms)`failing. 
  *  denoms, err := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, poolId)

x/twap/listeners_test.go
ValarDragon: reintroduced

x/twap/listeners_test.go
ValarDragon: I think trusting `types.NewTwapRecord` is fine, since thats our definition of correct for a new record. 

I also feel like `types.GetAllUniqueDenomPairs(denoms)` should be well tested / have no edge cases, but we could if we wanted add an 'expected num records' just to sanity check

ValarDragon (approved):
Committed a slight restructuring, to use the function we have to test if the expected record is correct.

Otherwise LGTM.

Only thing I think we don't check is what if multiple pools are created in a single block, do you think that should be added to this test, or be a separate one?

mattverse:
Ty for the refactoring!

I think it'd be rather confusing to go around another iteration of reviews on this same PR, I'll be creating a new PR for that case

Issue #2362, opened 2022-08-11
Add make command for building binary w/ osmobuilder


ValarDragon:
We use osmo-builder directly in a couple spots to build binaries. We should probably make this into a single make command right?

cref @niccoloraspa 

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/2299#discussion_r942725835_

niccoloraspa:
Yes, I think it can be used to replace the various `build-reproducibile-{arm,amd}` in the Makefile.

I also propose that we dismiss: https://github.com/osmosis-labs/osmosis/tree/main/contrib/images/rbuilder

niccoloraspa:
We should also backport the building process to previous `v*.X` branches.

State compatibility check on previous `v*.X` branches is failing because it can't find the `osmo-builder`

ValarDragon:
We can directly PR the latest osmobuilder code to those branches (up to fixing cosmwasm version), since backporting all PRs will probably be a pain

Issue #1679, opened 2022-06-06
[Story][Dockerfile Improvements] Generalize and reduce the number of Dockerfiles


p0mvn:
## Background

> As an Osmosis engineer, I would like our Dockerfiles to be reusable and modular so that I don't have to make changes in many files and avoid the images going out of sync

Our Dockerfiles share a lot of the same logic. On one side of the issue, some folks update all Dockerfiles, wasting time for no reason. On the other side, other folks only update some, letting the images go out of sync. 

First, we should remove images that are not used. Second, we should investigate the rest.  Some can be generalized to be used as a base, and the rest import that base.

- #1684 
- #1685

## Status
- June 6 - WIP: Defining Scope

niccoloraspa:
At the time of the issue we had the following Dockerimages:

```bash
❯ find . -iname "*dockerfile*"

./Dockerfile
./tests/localosmosis/Dockerfile
./tests/localosmosis/mainnet_state/Dockerfile-stateExport
./tests/e2e/initialization/init.Dockerfile
./contrib/images/osmobuilder/Dockerfile
./contrib/images/rbuilder/Dockerfile
```

Fixing https://github.com/osmosis-labs/osmosis/issues/2482, we removed the `contrib` folder and the following images:

```bash
./contrib/images/osmobuilder/Dockerfile
./contrib/images/rbuilder/Dockerfile
```

niccoloraspa:
Removed also:

```bash
./tests/localosmosis/Dockerfile
./tests/localosmosis/mainnet_state/Dockerfile-stateExport
```

ValarDragon:
Seeing that theres only two dockerfiles remaining, can we close this issue?

niccoloraspa:
Yes. Closing

PR #2427, opened 2022-08-17
Fix reproducible linux builds 

niccoloraspa:
Closes: https://github.com/osmosis-labs/osmosis/issues/2362

This PR:

- Improves the main Dockerfile 
- Fixes reproducible builds in the Makefile

**🐳 General Dockerfile improvements:**

- Move up the layer to download `go` dependencies
- Automatically detect CosmWasm version and architecture to download the correct `libwasm`
- Explicitly use the `go build...` command
- Generalize runner images

**🛠 Reproducible build** idea:

- Compile the binary (statically linked with CosmWasm) while building the container image
- `cp` the binary out of the container and place it in the build folder

These simple changes drastically simplify the build process, and they allow very nice improvements:

1. `osmobuilder` is now deprecated as the functionality is now integrated in the main `Dockerfile`

> Removing `osmobuilder` would requires:
> - [ ] Changing the state-compatibility-check to build the binary using `make build-reproducible-amd64`
> - [ ] Update release docs

2. Deprecating `osmobuilder` allows to deprecated the whole `contrib/` folder, that can now be removed. 

> We would need to move the `scripts` 

4. We can now backport this build process to all `vX` branches to have consistent way to build the `osmosisd` binary across all versions (only difference would be the CosmWasm line)

5. The generalization of the runner image allows to create different flavour of the images starting from the same Dockerfile. This can be used to:
- Create different flavours of the `osmosis` Docker image (`distroless`, `nonroot`, `alpine`)
- Reduce the number of Dockerfile used internally (ref: https://github.com/osmosis-labs/osmosis/issues/1679)

I will create issues for every point.

## Brief Changelog

- Improve `Dockerfile`
- Fix commands in `Makefile` to create reproducible builds for linux

## Testing and Verifying

You can build the binaries with:

```
make build-reproducible
# Inspected build folder 
ls build
# osmosisd-linux-amd64 osmosisd-linux-arm64
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,61 @@
 # syntax=docker/dockerfile:1
 
-ARG BASE_IMG_TAG=nonroot
+ARG GO_VERSION="1.18"
+ARG RUNNER_IMAGE="gcr.io/distroless/static:nonroot"
 
 # --------------------------------------------------------
-# Build 
+# Builder
 # --------------------------------------------------------
 
-FROM golang:1.18.2-alpine3.15 as build
+FROM golang:${GO_VERSION}-alpine as builder
 
-# linux-headers needed by github.com/zondax/hid
 RUN set -eux; apk add --no-cache ca-certificates build-base; apk add git linux-headers
 
-# CosmWasm: see https://github.com/CosmWasm/wasmvm/releases
-# 1) Add the releases into /lib/
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/libwasmvm_muslc.aarch64.a \ 
-  https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/libwasmvm_muslc.x86_64.a \ 
-  /lib/
-# 2) Verify their hashes
-RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep 7d2239e9f25e96d0d4daba982ce92367aacf0cbd95d2facb8442268f2b1cc1fc; \
-  sha256sum /lib/libwasmvm_muslc.x86_64.a | grep f6282df732a13dec836cda1f399dd874b1e3163504dbd9607c6af915b2740479
-# Copy the right library according to architecture. The final location will be found by the linker flag `-lwasmvm_muslc`
-RUN cp /lib/libwasmvm_muslc.$(uname -m).a /lib/libwasmvm_muslc.a
-
-# Get go dependencies first
-COPY go.mod go.sum /osmosis/
+# Download go dependencies
 WORKDIR /osmosis
+COPY go.mod go.mod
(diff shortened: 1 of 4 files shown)


czarcas7ic:
Looks like we will need to modify e2e tests for this change, will look into it

niccoloraspa:
Thanks @czarcas7ic

czarcas7ic (approved):
Seems like its good to go now, just had to specify start command in e2e now that we took it out form the Dockerfile (I think that was the right move FWIW)

Dockerfile
```diff
@@ -1,50 +1,61 @@
...
-  sha256sum /lib/libwasmvm_muslc.x86_64.a | grep f6282df732a13dec836cda1f399dd874b1e3163504dbd9607c6af915b2740479
-# Copy the right library according to architecture. The final location will be found by the linker flag `-lwasmvm_muslc`
-RUN cp /lib/libwasmvm_muslc.$(uname -m).a /lib/libwasmvm_muslc.a
-
-# Get go dependencies first
-COPY go.mod go.sum /osmosis/
+# Download go dependencies
 WORKDIR /osmosis
+COPY go.mod go.mod
+COPY go.sum go.sum
```
p0mvn, line 17: Consider doing this in one command to minimize the number of layers. As per:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

p0mvn (approved):

Dockerfile
niccoloraspa: Thanks! I'add in the next PR I'm making so we don't have to review again the changes

Dockerfile
```diff
@@ -1,50 +1,61 @@
...
+      -tags "netgo,ledger,muslc" \
+      -ldflags "-X github.com/cosmos/cosmos-sdk/version.Name="osmosis" \
+              -X github.com/cosmos/cosmos-sdk/version.AppName="osmosisd" \
+              -X github.com/cosmos/cosmos-sdk/version.Version=$VERSION \
+              -X github.com/cosmos/cosmos-sdk/version.Commit=$COMMIT \
+              -X github.com/cosmos/cosmos-sdk/version.BuildTags='netgo,ledger,muslc' \
+              -w -s -linkmode=external -extldflags '-Wl,-z,muldefs -static'" \
+      -trimpath \
+      -o /osmosis/build/ \
+      ./...
```
ValarDragon, line 50: Is there a make command we should be running, instead of copying this here?

niccoloraspa: Not sure if there is already a `make` target that will generate the `go build` command with all these flags.

We could for sure write one but avoiding using the Makefile in the Dockerfile was intentional. It makes the image creation more transparent and immediate.

PR #2429, opened 2022-08-17
Twap: Integrate and Add migration tests

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/2198

Integrates and adds migration test. The migration logic is implemented using the last pool id, not by creating a method that iterates through all pools and gets all pool IDs.

This PR also introduces test cases for the upgrade itself, and the migration method. 

## Brief Changelog

Add migration test for existing pools on the twap module.


app/upgrades/v12/upgrade_test.go
```diff
@@ -0,0 +1,90 @@
...
+	suite.Run(t, new(UpgradeTestSuite))
+}
+
+const dummyUpgradeHeight = 5
+
+// note that this test does not perfectly test state migration, as pre_upgrade function
+// which includes setting up tests by creating new pools would also create twap records
+// automatically with this binary. The goal of this test is to test that the upgrade handler
+// does not panic upon upgrade.
+// Detailed state migration tests are placed within the twap keeper.
```
czarcas7ic, line 33: This is an important point. I think this is one of those things that just gets manually tested, because implementing it in say, e2e test suite, wouldn't make sense since it will only ever happen once. Since we have significant state on our testnet, if migration works there I would say we are good to go.

x/twap/migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	err := s.twapkeeper.MigrateExistingPools(s.Ctx, latestPoolId)
+	s.Require().NoError(err)
+
+	upgradeTime := s.Ctx.BlockTime()
+
+	// iterate through all pools, check that all state entries has been correctly updated
+	for poolId := 1; poolId <= int(latestPoolId); poolId++ {
+		recentTwapRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, uint64(poolId))
+		poolDenoms, _ := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, uint64(poolId))
+		denomPairs, _ := types.GetAllUniqueDenomPairs(poolDenoms)
```
czarcas7ic, line 29: Should we check errors here?

czarcas7ic (approved):

app/upgrades/v12/upgrade_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	for _, tc := range testCases {
+		suite.Run(fmt.Sprintf("Case %s", tc.name), func() {
+			suite.SetupTest() // reset
+
+			// creting pools before upgrade
+			poolId := tc.pre_upgrade()
+
+			// run upgrade
+			tc.upgrade()
```
p0mvn, line 84: Is it possible to avoid using callbacks? I think it would make sense if there were multiple cases with different `pre_upgrade`, `upgrade` or `post_upgrade` logic but here it seems that we have only 1

Also, are we expecting more test cases? Does it make sense for this test to not be table-driven?

x/twap/migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Second * 10))
+
+	// run migration logic
+	latestPoolId := s.App.GAMMKeeper.GetNextPoolId(s.Ctx) - 1
+	err := s.twapkeeper.MigrateExistingPools(s.Ctx, latestPoolId)
+	s.Require().NoError(err)
+
+	upgradeTime := s.Ctx.BlockTime()
+
+	// iterate through all pools, check that all state entries has been correctly updated
```
p0mvn, line 25:
```suggestion
	// iterate through all pools, check that all state entries have been correctly updated
```

x/twap/migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// create two pools before migration
+	s.PrepareUni2PoolWithAssets(defaultUniV2Coins[0], defaultUniV2Coins[1])
+	s.PrepareBalancerPool()
+
+	// suppose upgrade happened and increment block height and block time
+	s.Ctx = s.Ctx.WithBlockHeight(s.Ctx.BlockHeight() + 1)
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Second * 10))
+
+	// run migration logic
+	latestPoolId := s.App.GAMMKeeper.GetNextPoolId(s.Ctx) - 1
```
p0mvn, line 19: I think it would be valuable to also have a test with `s.App.GAMMKeeper.GetNextPoolId(s.Ctx)` (no -1) and expect that to error

x/twap/migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	upgradeTime := s.Ctx.BlockTime()
+
+	// iterate through all pools, check that all state entries has been correctly updated
+	for poolId := 1; poolId <= int(latestPoolId); poolId++ {
+		recentTwapRecords, err := s.twapkeeper.GetAllMostRecentRecordsForPool(s.Ctx, uint64(poolId))
+		poolDenoms, _ := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, uint64(poolId))
+		denomPairs, _ := types.GetAllUniqueDenomPairs(poolDenoms)
+		s.Require().NoError(err)
+		s.Require().Equal(len(denomPairs), len(recentTwapRecords))
```
p0mvn, line 31: Is there anything else that makes sense to check such that for every denom pair, there are twap records with matching denoms?

x/twap/migrate_test.go
```diff
@@ -0,0 +1,45 @@
...
+		s.Require().Equal(upgradeTime, recentTwapRecords[0].Time)
+
+		twapRecord, err := s.twapkeeper.GetMostRecentRecordStoreRepresentation(s.Ctx, uint64(poolId), recentTwapRecords[0].Asset0Denom, recentTwapRecords[0].Asset1Denom)
+		s.Require().NoError(err)
+		s.Require().Equal(upgradeTime, twapRecord.Time)
+
+		twapRecordBeforeTime, err := s.twapkeeper.GetRecordAtOrBeforeTime(s.Ctx, uint64(poolId), s.Ctx.BlockTime(), twapRecord.Asset0Denom, twapRecord.Asset1Denom)
+		s.Require().NoError(err)
+		s.Require().Equal(upgradeTime, twapRecordBeforeTime.Time)
+	}
```
p0mvn, line 44: Can we add a check for the transient store being updated?

p0mvn (commented):
Great work @mattverse 

I skimmed the original issue and I think there might be one point that needs to be addressed:

> add an upgrade test for v11, that ensures were still producing blocks with swaps post-upgrade.

I think that implies adding an e2e test. We currently don't have any doins swaps, only creating pools. 

Can we either add it here or create an issue, attaching it to the v12 milestone blocker, please?

I'm also happy to help with adding this check to e2e

app/upgrades/v12/upgrade_test.go
```diff
@@ -0,0 +1,90 @@
+package v12_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	abci "github.com/tendermint/tendermint/abci/types"
+
```
alexanderbez, line 9:
```suggestion
```

alexanderbez (approved):
ACK (pending review comments being addressed 👍 )

app/upgrades/v12/upgrade_test.go
ValarDragon: Wait I thought e2e test is for helping us with upgrade creation / things across versions

app/upgrades/v12/upgrade_test.go
ValarDragon: I think we should come back in a followup, and cleanup the entire go upgrade test framework. This is copied from another upgrade test, we should make a unified setup for all of them / decide what we generally want.

I think we should capture this into an issue

x/twap/migrate.go
```diff
@@ -0,0 +1,16 @@
...
+// MigrateExistingPools iterates through all pools and creates state entry for the twap module.
+func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error {
+	for i := 1; i <= int(latestPoolId); i++ {
+		err := k.afterCreatePool(ctx, uint64(i))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
ValarDragon, line 16: Thoughts on moving to listeners.go, keepers.go or api.go

Maybe i'm overly stubborn, with feeling bad about an extra file for this lol

app/upgrades/v12/upgrades.go
```diff
@@ -43,9 +43,8 @@ func CreateUpgradeHandler(
 		}
 
 		// Initialize TWAP state
-		// TODO: Get allPoolIds from gamm keeper, and write test for migration.
-		allPoolIds := []uint64{}
-		err := keepers.TwapKeeper.MigrateExistingPools(ctx, allPoolIds)
+		latestPoolId := keepers.GAMMKeeper.GetNextPoolId(ctx) - 1
+		err := keepers.TwapKeeper.MigrateExistingPools(ctx, latestPoolId)
```
ValarDragon, line 47: Nice, this is a better design

x/twap/migrate.go
ValarDragon: (this doesn't block PR merge)

app/upgrades/v12/upgrade_test.go
czarcas7ic: So e2e test will test this in the sense that, the state migration will happen and we just assume that if it doesn't panic then the state migration happened correctly. What I meant was, we wouldn't implement some e2e test that checks that the state migration ACTUALLY occurred correctly, but (on testnet say) we should manually inspect prior existing pools to ensure that post migration there are no oddities. I would also feel better seeing testnet, which has many pools and lots of existing state, go through this migration to feel 100%

czarcas7ic: I guess what I am saying is, e2e will test this, but maybe its an item we highlight as needing another look QA wise when testnet gets upgraded!

x/twap/migrate.go
mattverse: that I agree, but we're planning to delete this after upgrade in main no?
Feel like easier to manage them separately lmk what u think

ValarDragon: oh thats a good point, hrmm ok

ValarDragon (approved):

x/twap/migrate_test.go
ValarDragon: Nothing comes to mind for me for other important things to test in migration

ValarDragon:
I think we should merge this after @p0mvn 's comment for transient store being tested is addressed

x/twap/migrate_test.go
mattverse: Added err check for GetPoolDenoms, GetAllUniqueDenomPairs does not return err for its second value, it returns a second array of string as the second array of strings )

x/twap/migrate_test.go
mattverse: done!

mattverse:
Addressed! Merging :)

x/twap/migrate_test.go
mattverse: Addressed!

Issue #2426, opened 2022-08-16
TWAP: Mutation testing


czarcas7ic:
## Results

1 @@ -84,7 +84,7 @@
```
 // This accumulator can be stored, to compute wider ranged twaps.
 func (k Keeper) GetBeginBlockAccumulatorRecord(ctx sdk.Context, poolId uint64, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
 	// correct ordering of args for db
-	if asset1Denom > asset0Denom {
+	if asset1Denom >= asset0Denom {
 		asset0Denom, asset1Denom = asset1Denom, asset0Denom
 	}
 	return k.getMostRecentRecord(ctx, poolId, asset0Denom, asset1Denom)
```
FAIL "/var/folders/9x/tjdmlrtj6tjgf_42rdzpv6rw0000gn/T/go-mutesting-4154319244/./x/twap/api.go.0" with checksum ed25623be1b9b1b3495414abdaedc476
***
2 @@ -85,7 +85,8 @@
```
 func (k Keeper) GetBeginBlockAccumulatorRecord(ctx sdk.Context, poolId uint64, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
 	// correct ordering of args for db
 	if asset1Denom > asset0Denom {
-		asset0Denom, asset1Denom = asset1Denom, asset0Denom
+		_, _, _, _ = asset0Denom, asset1Denom, asset1Denom, asset0Denom
+
 	}
 	return k.getMostRecentRecord(ctx, poolId, asset0Denom, asset1Denom)
 }
```
FAIL "/var/folders/9x/tjdmlrtj6tjgf_42rdzpv6rw0000gn/T/go-mutesting-4154319244/./x/twap/api.go.1" with checksum 1233f31b5561b25202c70c3a3acecdca
***
3 @@ -23,7 +23,7 @@
```
 // GetParams returns the total set of twap parameters.
 func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
-	k.paramSpace.GetParamSet(ctx, &params)
+	_, _, _ = k.paramSpace.GetParamSet, ctx, params
 	return params
 }
```

FAIL "/var/folders/9x/tjdmlrtj6tjgf_42rdzpv6rw0000gn/T/go-mutesting-4154319244/./x/twap/keeper.go.0" with checksum 19d0e8e261272b23c6b93f4a03b87a38
***
4 @@ -29,7 +29,7 @@
[...]

PR #2436, opened 2022-08-17
feat: mutation test script improvements

czarcas7ic:
Related To: https://github.com/osmosis-labs/osmosis/issues/2426

We were previously hardcoding which module to run mutation tests on. This PR now allows for a comma separated `MODULES` argument to be given to the mutation test script and allows it to run all desired modules.

Example: 
`make test-mutation MODULES=twap`
OR
`make test-mutation MODULES=twap,superfluid`

Leaving `MODULES` blank assumes all modules are desired.

## Brief Changelog

- Adds comma separated module arguments to be passed into script
- Runs mutation tests on root directory of module (twap especially needs this)
- Sets current weekly CI to only run for tokenfactory

## Testing and Verifying

### DIFF
--- a/.github/workflows/mutest-issue-generate.yml
+++ b/.github/workflows/mutest-issue-generate.yml
@@ -1,4 +1,4 @@
-name: Generate Mutation Test Errors 
+name: Generate Mutation Test Errors
 
 on:
   schedule:
@@ -14,14 +14,14 @@ jobs:
       -
         name: Run mutation test
         continue-on-error: true
-        run: make test-mutation
+        run: make test-mutation MODULES=tokenfactory
       -
         name: Execute mutation test format script
         id: mutest-formatted
         run: |
           cat mutation_test_result.txt | grep -Ev "PASS" | grep -Ev "SKIP" | tee mutation_test_result.txt
       -
-        name: Generate code blocks 
+        name: Generate code blocks
         id: gen-code-blocks
         run: |
           cat mutation_test_result.txt  | sed "s# @@# @@\n\`\`\`go\n#g " | sed "s#FAIL#\`\`\`\nFAIL\n\n\n#g " > mutation_test_result.txt 
@@ -29,8 +29,8 @@ jobs:
         name: Get today's date
         id: date
         run: |
-          echo "::set-output name=today::$(date "+%Y/%m/%d")" 
-      - 
+          echo "::set-output name=today::$(date "+%Y/%m/%d")"
+      -
         name: Read mutation_test_result file
         id: result
         uses: juliangruber/read-file-action@v1
--- a/.github/workflows/mutest-issue-generate.yml
+++ b/.github/workflows/mutest-issue-generate.yml
@@ -14,7 +14,9 @@ jobs:
       -
         name: Run mutation test
         continue-on-error: true
-        run: make test-mutation MODULES=tokenfactory
+        run: make test-mutation
+        env:
+          MODULES: tokenfactory
       -
         name: Execute mutation test format script
         id: mutest-formatted
--- a/scripts/mutation-test.sh
+++ b/scripts/mutation-test.sh
@@ -15,12 +15,12 @@ arg_len=$#
 
 for i in "$@"; do
   if [ $arg_len -gt 1 ]; then
-  MODULE_FORMAT+="./x/$i\|"
-  MODULE_NAMES+="${i} "
-  let "arg_len--"
+    MODULE_FORMAT+="./x/$i\|"
+    MODULE_NAMES+="${i} "
+    let "arg_len--"
   else
-  MODULE_FORMAT+="./x/$i"
-  MODULE_NAMES+="${i}"
+    MODULE_FORMAT+="./x/$i"
+    MODULE_NAMES+="${i}"
   fi
 done
 
--- a/scripts/mutation-test.sh
+++ b/scripts/mutation-test.sh
@@ -1,22 +1,35 @@
 #!/usr/bin/env bash
 
 set -eo pipefail
+oIFS="$IFS"; IFS=, ; set -- $1 ; IFS="$oIFS"
 
 DISABLED_MUTATORS='branch/*'
 
 # Only consider the following:
-# * go files in types or keeper packages
+# * go files in types, keeper, or module root directories
 # * ignore test and Protobuf files
-MUTATION_SOURCES=$(find ./x -type f \( -path '*/keeper/*' -or -path '*/types/*' \) \( -name '*.go' -and -not -name '*_test.go' -and -not -name '*pb*' \))
+MUTATION_SOURCES=$(find ./x -type f \( -path '*/keeper/*' -or -path '*/types/*' -or -path '*' \) \( -name '*.go' -and -not -name '*_test.go' -and -not -name '*pb*' \))
 
-# XXX: Filter on a module-by-module basis and expand when we think other modules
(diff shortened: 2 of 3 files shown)


scripts/mutation-test.sh
```diff
@@ -1,22 +1,35 @@
...
-# are ready. Once all modules are considered stable enough to be tested, remove
-# this filter entirely.
-MUTATION_SOURCES=$(echo "$MUTATION_SOURCES" | grep './x/tokenfactory')
+# Filter on a module-by-module basis as provided by input
+arg_len=$#
+
+for i in "$@"; do
+  if [ $arg_len -gt 1 ]; then
+  MODULE_FORMAT+="./x/$i\|"
+  MODULE_NAMES+="${i} "
```
p0mvn, line 19: Should these be indented?

scripts/mutation-test.sh
```diff
@@ -1,22 +1,35 @@
...
+  let "arg_len--"
+  else
+  MODULE_FORMAT+="./x/$i"
+  MODULE_NAMES+="${i}"
```
p0mvn, line 23: and these

scripts/mutation-test.sh
```diff
@@ -1,22 +1,35 @@
...
+oIFS="$IFS"; IFS=, ; set -- $1 ; IFS="$oIFS"
 
 DISABLED_MUTATORS='branch/*'
 
 # Only consider the following:
-# * go files in types or keeper packages
+# * go files in types, keeper, or module root directories
 # * ignore test and Protobuf files
-MUTATION_SOURCES=$(find ./x -type f \( -path '*/keeper/*' -or -path '*/types/*' \) \( -name '*.go' -and -not -name '*_test.go' -and -not -name '*pb*' \))
+MUTATION_SOURCES=$(find ./x -type f \( -path '*/keeper/*' -or -path '*/types/*' -or -path '*' \) \( -name '*.go' -and -not -name '*_test.go' -and -not -name '*pb*' \))
```
p0mvn, line 11: So we are allowing any path according to `-or -path '*'`? Does it make sense to not apply any path filtering then?

czarcas7ic: Oof I meant root directory files but I didn't think that included the folders themselves, let me look into this, good catch

czarcas7ic: Great catch! Most recent commit should solve this problem by specifying a max depth after pulling what we want from keeper and types

p0mvn: Why are we excluding paths deeper than 2? Is the goal to exclude files in the `client` folder?

scripts/mutation-test.sh
```diff
@@ -1,35 +1,51 @@
...
+done
+
+MUTATION_SOURCES=$(echo "$MUTATION_SOURCES" | grep "$MODULE_FORMAT")
 
 # Collect multiple lines into a single line to be fed into go-mutesting
 MUTATION_SOURCES=$(echo $MUTATION_SOURCES | tr '\n' ' ')
+echo $MUTATION_SOURCES
 
-OUTPUT=$(go run github.com/osmosis-labs/go-mutesting/cmd/go-mutesting --disable=$DISABLED_MUTATORS $MUTATION_SOURCES)
+# echo "running mutation tests for the following module(s): $MODULE_NAMES"
```
p0mvn, line 34: This was for testing right?

czarcas7ic: No, I wanted it to print so that way when running you can see which modules it's running for. Should I remove?

scripts/mutation-test.sh
czarcas7ic: We want keeper folder and types folder which is the first cmd. Then the second command starts in the x folder, and two up would be the root module folder (because x counts as one) so this gives us any root folder files for the selected modules (I used prints to ensure this works as intended)

p0mvn: That makes sense, why do we only want the root files of each module? What about pool-module packages and alike??

scripts/mutation-test.sh
```diff
@@ -1,35 +1,51 @@
...
-# Fetch the final result output and the overall mutation testing score
-RESULT=$(echo "$OUTPUT" | grep 'The mutation score')
-SCORE=$(echo "$RESULT" | grep -Eo '[[:digit:]]\.[[:digit:]]+')
+# # Fetch the final result output and the overall mutation testing score
+# RESULT=$(echo "$OUTPUT" | grep 'The mutation score')
+# SCORE=$(echo "$RESULT" | grep -Eo '[[:digit:]]\.[[:digit:]]+')
 
-echo "writing mutation test result to mutation_test_result.txt"
-echo "$OUTPUT" > mutation_test_result.txt
+# echo "writing mutation test result to mutation_test_result.txt"
```
stackman27, line 41: are we not writing the results to .txt file anymore?

stackman27 (commented):
Just curious, What does the output of make look like now? 
`make test-mutation MODULES=tokenfactory`

scripts/mutation-test.sh
czarcas7ic: ah you were referring to the commented out stuff, yes that was for testing my bad

scripts/mutation-test.sh
czarcas7ic: Was no meant to be commented out, thanks for pointing out!

czarcas7ic:
@stackman27 now that I have uncommented the lines this should make more sense, sorry about that, was debugging

scripts/mutation-test.sh
```diff
@@ -1,22 +1,36 @@
 #!/usr/bin/env bash
 
 set -eo pipefail
+oIFS="$IFS"; IFS=, ; set -- $1 ; IFS="$oIFS"
```
alexanderbez, line 4: what does this do?

.github/workflows/mutest-issue-generate.yml
```diff
@@ -14,23 +14,23 @@ jobs:
       -
         name: Run mutation test
         continue-on-error: true
-        run: make test-mutation
+        run: make test-mutation MODULES=tokenfactory
```
alexanderbez, line 17: nit: set the env directly instead

alexanderbez (approved):
utACK

scripts/mutation-test.sh
czarcas7ic: This line allows for parsing a comma separated list of a single argument (i.e takes `superfluid,twap` and makes it two separate inputs)

.github/workflows/mutest-issue-generate.yml
czarcas7ic: Can you explain further what you mean here? I though I would need to be explicit with what I want tested here since it is a github action

alexanderbez: Github supports an `env` field.

```yml
env:
  MODULES: tokenfactory
```

https://docs.github.com/en/actions/learn-github-actions/environment-variables

czarcas7ic: Learn something new everyday, thanks!

scripts/mutation-test.sh
czarcas7ic: Hmm, I didn't want to drift too far away from the original implementation since I lacked context on mutation testing. In fact I would have completely left it alone if twap had the same structure as the other modules. You might be right though, we should probably test this as well

Issue #2432, opened 2022-08-17
V12 ICA message updates


ValarDragon:
As noted in a couple places, we should update the supported messages for ICA.

Lets:
- [ ] Remove all three Authz messages, due to bad security design of Authz
- [ ] Add Unbond

And review if theres any others that are missing

nicolaslara:
I was in the neighbourhood and curious about how this was done in upgrades, so I went ahead and added a PR for this.

PR #2439, opened 2022-08-18
feat: Removed authz and added unbound to ics host params

nicolaslara:
Closes: #2432

Removes authz messages from ICA and adds superfluid unbound

## Brief Changelog

 - Authz messages can no longer be executed via ICA on v12
 - Superfluid unbound messages can now be executed via ICA on v12

## Testing and Verifying

TBD. Not sure how to test upgrades

### DIFF
--- a/app/upgrades/v12/upgrades.go
+++ b/app/upgrades/v12/upgrades.go
@@ -57,6 +57,8 @@ func CreateUpgradeHandler(
 			sdk.MsgTypeURL(&stakingtypes.MsgBeginRedelegate{}),
 			sdk.MsgTypeURL(&stakingtypes.MsgCreateValidator{}),
 			sdk.MsgTypeURL(&stakingtypes.MsgEditValidator{}),
+			// Change: Added MsgUndelegate
+			sdk.MsgTypeURL(&stakingtypes.MsgUndelegate{}),
 			sdk.MsgTypeURL(&distrtypes.MsgWithdrawDelegatorReward{}),
 			sdk.MsgTypeURL(&distrtypes.MsgSetWithdrawAddress{}),
 			sdk.MsgTypeURL(&distrtypes.MsgWithdrawValidatorCommission{}),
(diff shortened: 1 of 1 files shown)


app/upgrades/v12/upgrades.go
```diff
@@ -42,6 +48,34 @@ func CreateUpgradeHandler(
...
 
+		// Update ICA allowedMessages param
+		params := keepers.ICAHostKeeper.GetParams(ctx)
+		// Specifying the whole list instead of adding and removing. Less fragile.
+		params.AllowMessages = []string{
+			sdk.MsgTypeURL(&banktypes.MsgSend{}),
+			sdk.MsgTypeURL(&stakingtypes.MsgDelegate{}),
+			sdk.MsgTypeURL(&stakingtypes.MsgBeginRedelegate{}),
+			sdk.MsgTypeURL(&stakingtypes.MsgCreateValidator{}),
+			sdk.MsgTypeURL(&stakingtypes.MsgEditValidator{}),
```
ValarDragon, line 59: We also need to add staking unbond / undelegate (I forget what its called)

ValarDragon:
Thanks for making this! Not sure why e2e is erroring

nicolaslara:
>  Not sure why e2e is erroring

I also saw it erroring on the rate limiting branch, but re running fixed it 🤷

app/upgrades/v12/upgrades.go
nicolaslara: added undelegate.

nicolaslara:
The e2e tests consistently fail. Ran them without the changes and they ran fine. 

Trying to figure out what's wrong with the changes now. 

Shot in the dark, but maybe this?

https://github.com/osmosis-labs/osmosis/blob/main/app/upgrades/v9/upgrades.go#L44-L46

alexanderbez (approved):

ValarDragon (approved):
I'm so confused why get params and set had an error but just set doesn't but the change LGTM!! (Sanity checked against v9)

How'd you find that?

nicolaslara:
yeah, it's still unclear to me. Something about how references are handled in `ParamSetPairs` and `SetParamSet`. 

I find that whole internal param implementation very confusing. I *think* it may have something to do with `SetParamSet` passing the params as a reference to the object we get via the `keeper.GetParams` (which has a reference to the params in the keeper), and then calling `(p *Params) ParamSetPairs()` on the dereferenced object, but it's all a bit blurry.

Issue #2406, opened 2022-08-14
osmo routed half swap fees


hleb-albau:
## Background

There is passed [proposal](https://www.mintscan.io/osmosis/proposals/187) and [its first implementation attempt](https://github.com/osmosis-labs/osmosis/pull/1069). Should we implement it, or it is abandoned idea?

## Suggested Design

The same as in PR, but implemented on top of current main branch.

## Other

I am ready to impl that.

hleb-albau:
Hey @ValarDragon could you please just drop a line - yes or no :)

ValarDragon:
We should implement this!

hleb-albau:
took it

PR #2454, opened 2022-08-19
x/gamm: Reduce swap fee for Osmo-routed trades (#2406)

hleb-albau:
Closes: #2406

This pull request reduce swap fee for a half during calculation for each pool for routes with length of 2 and osmo intermediate, ex `ATOM -> OSMO -> USDC`. 

In general it is re-implementation of https://github.com/osmosis-labs/osmosis/pull/1069 with addition tests added.

## Brief Changelog

  - x/gamm Change `SwapExactAmountIn` and `SwapExactAmountOut`to take fee multiplier as addition input param.
  - Update multihop swaps test, to insure that half fee applies only for osmo routed multihop swaps.

## Testing and Verifying

  - *Added cases for existing multihop unit tests.*

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?  yes
  - How is the feature or change documented? specification (`x/<module>/spec/`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#2418](https://github.com/osmosis-labs/osmosis/pull/2418) x/mint remove SetInitialSupplyOffsetDuringMigration from keeper
 * [#2417](https://github.com/osmosis-labs/osmosis/pull/2417) x/mint unexport keeper `SetLastReductionEpochNum`, `getLastReductionEpochNum`, `CreateDeveloperVestingModuleAccount`, and `MintCoins`
 * [#2587](https://github.com/osmosis-labs/osmosis/pull/2587) remove encoding config argument from NewOsmosisApp
+* [#2454](https://github.com/osmosis-labs/osmosis/pull/2454) x/gamm Change `SwapExactAmountIn` and `SwapExactAmountOut` to take fee multiplier as addition input param.
 
 ### Features
 
--- a/x/gamm/README.md
+++ b/x/gamm/README.md
@@ -86,6 +86,11 @@ All tokens are swapped using a multi-hop mechanism. That is, all swaps
 are routed via the most cost-efficient way, swapping in and out from
 multiple pools in the process.
 
+When a trade consists of just two OSMO-included routes during a single transaction,
+the swap fees on each hop would be automatically halved. 
+Example: for converting `ATOM -> OSMO -> LUNA` using two pools with swap fees `0.3% + 0.2%`,
+instead `0.15% + 0.1%` fees will be aplied. 
+
 [Multi-Hop](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/multihop.go)
 
 ## Weights
--- a/x/gamm/keeper/multihop.go
+++ b/x/gamm/keeper/multihop.go
@@ -17,6 +17,11 @@ func (k Keeper) MultihopSwapExactAmountIn(
 	tokenOutMinAmount sdk.Int,
 ) (tokenOutAmount sdk.Int, err error) {
 	for i, route := range routes {
+		swapFeeMultiplier := sdk.OneDec()
+		if types.SwapAmountInRoutes(routes).IsOsmoRoutedMultihop() {
+			swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
+		}
+
 		// To prevent the multihop swap from being interrupted prematurely, we keep
 		// the minimum expected output at a very low number until the last pool
 		_outMinAmount := sdk.NewInt(1)
@@ -25,7 +30,13 @@ func (k Keeper) MultihopSwapExactAmountIn(
 		}
 
 		// Execute the expected swap on the current routed pool
-		tokenOutAmount, err = k.SwapExactAmountIn(ctx, sender, route.PoolId, tokenIn, route.TokenOutDenom, _outMinAmount)
+		pool, poolErr := k.getPoolForSwap(ctx, route.PoolId)
+		if poolErr != nil {
+			return sdk.Int{}, poolErr
+		}
+
+		swapFee := pool.GetSwapFee(ctx).Mul(swapFeeMultiplier)
+		tokenOutAmount, err = k.swapExactAmountIn(ctx, sender, pool, tokenIn, route.TokenOutDenom, _outMinAmount, swapFee)
 		if err != nil {
 			return sdk.Int{}, err
 		}
@@ -33,7 +44,7 @@ func (k Keeper) MultihopSwapExactAmountIn(
 		// Chain output of current pool as the input for the next routed pool
 		tokenIn = sdk.NewCoin(route.TokenOutDenom, tokenOutAmount)
 	}
-	return
+	return tokenOutAmount, err
 }
 
 // MultihopSwapExactAmountOut defines the output denom and output amount for the last pool.
@@ -47,8 +58,14 @@ func (k Keeper) MultihopSwapExactAmountOut(
 	tokenInMaxAmount sdk.Int,
 	tokenOut sdk.Coin,
 ) (tokenInAmount sdk.Int, err error) {
+	swapFeeMultiplier := sdk.OneDec()
+
+	if types.SwapAmountOutRoutes(routes).IsOsmoRoutedMultihop() {
+		swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
+	}
+
 	// Determine what the estimated input would be for each pool along the multihop route
-	insExpected, err := k.createMultihopExpectedSwapOuts(ctx, routes, tokenOut)
+	insExpected, err := k.createMultihopExpectedSwapOuts(ctx, routes, tokenOut, swapFeeMultiplier)
 	if err != nil {
 		return sdk.Int{}, err
 	}
@@ -71,9 +88,14 @@ func (k Keeper) MultihopSwapExactAmountOut(
 		}
 
 		// Execute the expected swap on the current routed pool
-		_tokenInAmount, err := k.SwapExactAmountOut(ctx, sender, route.PoolId, route.TokenInDenom, insExpected[i], _tokenOut)
-		if err != nil {
-			return sdk.Int{}, err
+		pool, poolErr := k.getPoolForSwap(ctx, route.PoolId)
+		if poolErr != nil {
+			return sdk.Int{}, poolErr
+		}
+		swapFee := pool.GetSwapFee(ctx).Mul(swapFeeMultiplier)
+		_tokenInAmount, swapErr := k.swapExactAmountOut(ctx, sender, pool, route.TokenInDenom, insExpected[i], _tokenOut, swapFee)
+		if swapErr != nil {
+			return sdk.Int{}, swapErr
 		}
 
 		// Sets the final amount of tokens that need to be input into the first pool. Even though this is the final return value for the
@@ -92,7 +114,11 @@ func (k Keeper) MultihopSwapExactAmountOut(
 // amount for this last pool and then chains that input as the output of the previous pool in the route, repeating
 // until the first pool is reached. It returns an array of inputs, each of which correspond to a pool ID in the
 // route of pools for the original multihop transaction.
-func (k Keeper) createMultihopExpectedSwapOuts(ctx sdk.Context, routes []types.SwapAmountOutRoute, tokenOut sdk.Coin) ([]sdk.Int, error) {
+func (k Keeper) createMultihopExpectedSwapOuts(
+	ctx sdk.Context,
+	routes []types.SwapAmountOutRoute,
+	tokenOut sdk.Coin, swapFeeMultiplier sdk.Dec,
+) ([]sdk.Int, error) {
 	insExpected := make([]sdk.Int, len(routes))
 	for i := len(routes) - 1; i >= 0; i-- {
 		route := routes[i]
@@ -102,7 +128,7 @@ func (k Keeper) createMultihopExpectedSwapOuts(ctx sdk.Context, routes []types.S
 			return nil, err
 		}
 
-		tokenIn, err := pool.CalcInAmtGivenOut(ctx, sdk.NewCoins(tokenOut), route.TokenInDenom, pool.GetSwapFee(ctx))
+		tokenIn, err := pool.CalcInAmtGivenOut(ctx, sdk.NewCoins(tokenOut), route.TokenInDenom, pool.GetSwapFee(ctx).Mul(swapFeeMultiplier))
 		if err != nil {
 			return nil, err
 		}
--- a/x/gamm/keeper/multihop.go
+++ b/x/gamm/keeper/multihop.go
@@ -19,7 +19,7 @@ func (k Keeper) MultihopSwapExactAmountIn(
 	for i, route := range routes {
 		swapFeeMultiplier := sdk.OneDec()
 		if types.SwapAmountInRoutes(routes).IsOsmoRoutedMultihop() {
-			swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
+			swapFeeMultiplier = types.MultihopSwapFeeMultiplierForOsmoPools.Clone()
 		}
 
 		// To prevent the multihop swap from being interrupted prematurely, we keep
@@ -61,7 +61,7 @@ func (k Keeper) MultihopSwapExactAmountOut(
 	swapFeeMultiplier := sdk.OneDec()
 
 	if types.SwapAmountOutRoutes(routes).IsOsmoRoutedMultihop() {
-		swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
+		swapFeeMultiplier = types.MultihopSwapFeeMultiplierForOsmoPools.Clone()
 	}
 
 	// Determine what the estimated input would be for each pool along the multihop route
--- a/x/gamm/keeper/multihop_test.go
+++ b/x/gamm/keeper/multihop_test.go
@@ -1,9 +1,9 @@
 package keeper_test
 
 import (
-	"github.com/osmosis-labs/osmosis/v11/x/gamm/pool-models/balancer"
-
+	"errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/osmosis-labs/osmosis/v11/x/gamm/pool-models/balancer"
 
 	"github.com/osmosis-labs/osmosis/v12/x/gamm/types"
 )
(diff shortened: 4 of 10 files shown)


x/gamm/README.md
```diff
@@ -86,8 +86,12 @@ All tokens are swapped using a multi-hop mechanism. That is, all swaps
 are routed via the most cost-efficient way, swapping in and out from
 multiple pools in the process.
 
+When a trade consists of just two OSMO-included routes during a single transaction,
+the swap fees on each hop would be automatically halved.
+
```
stackman27, line 91: Thoughts on adding an example here?

Also, does it matter if we half the `swapFee` at the end (totalfee / 2) vs halving it per pool swap with usmo?

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#2227](https://github.com/osmosis-labs/osmosis/pull/2227) Enable charging fee in base denom for `CreateGauge` and `AddToGauge`.
 * [#2283](https://github.com/osmosis-labs/osmosis/pull/2283) x/incentives: refactor `CreateGauge` and `AddToGauge` fees to use txfees denom
 * [#2206](https://github.com/osmosis-labs/osmosis/pull/2283) Register all Amino interfaces and concrete types on the authz Amino codec. This will allow the authz module to properly serialize and de-serializes instances using Amino.
+* [#2206](https://github.com/osmosis-labs/osmosis/pull/2283) Register all Amino interfaces and concrete types on the authz Amino codec. This will allow the authz module to properly serialize and de-serializes instances using Amino.
```
stackman27, line 61: Duplicate changelog

```suggestion
```

x/gamm/keeper/multihop_test.go
```diff
@@ -193,6 +283,13 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
 				sp := tokenInAmount.ToDec().Quo(test.param.tokenOut.Amount.ToDec())
 				fmt.Printf("spBefore %s, spAfter %s, sp actual %s\n", spotPriceBefore, spotPriceAfter, sp)
 				suite.True(sp.GT(spotPriceBefore) && sp.LT(spotPriceAfter), "multi-hop spot price wrong, test: %v", test.name)
+
+				if test.reducedFeeApplied {
+					suite.Require().True(tokenInAmount.LT(tokenInCalculatedAsSeparateSwaps.Amount))
+					suite.Require().True(tokenInAmount.GT(tokenInCalculatedAsSeparateSwapsWithoutFee.Amount))
```
stackman27, line 289: Does it matter if `reducedFee` is applied or not if the swap fee is 0?

stackman27: Cause i think `suite.Require().True(tokenInAmount.GT(tokenInCalculatedAsSeparateSwapsWithoutFee.Amount))` doesnot have to be in the if statement

hleb-albau: if swap fee is 0, than fee will not reduced. 
this line is just one more defensive check, that we will not lost user assets due some miscode with fee reduction (ex: after fee reduction another user could get more assets, that raw swap without fee at all)

hleb-albau: we can change it to `GTE`

stackman27: > (ex: after fee reduction another user could get more assets, that raw swap without fee at all)

How is this possible? If we have fee reduction, there is still a `swapFee` so, how can a user get more assets than `tokenInAmount`?

x/gamm/README.md
hleb-albau: Yes, in fact `totalfee` will be reduced by  less than by 50%, depending on various pool states. On test introduced in that PR it will be around 25%, cuz we create a pool with low liquidity.  In general final percentage will grow up to 50% the higher liquidity is.
I just made it as in proposal description(see forum) and sunny first implementation. I think it more elegant and stable solution, rather than trying to make it half for all cases.

x/gamm/keeper/multihop_test.go
hleb-albau: > > (ex: after fee reduction another user could get more assets, that raw swap without fee at all)
> 
> How is this possible? If we have fee reduction, there is still a `swapFee` so, how can a user get more assets than `tokenInAmount`?

just remember a june case, when users was able to get 50% from pools :) By design it is not possible, but one math code mistake, that was not covered my simple unit test make it real.

stackman27: aah makes sense! You got me there 😁

x/gamm/README.md
hleb-albau: added example in re-base commit

hleb-albau:
Also feel free to edit pr on your own, if it can safe your time for merge.

hleb-albau:
@stackman27 also we should create issue for frontend as well. But i don't know where to do it.

stackman27:
> @stackman27 also we should create issue for frontend as well. But i don't know where to do it.

You can report frontend issues here (https://github.com/osmosis-labs/osmosis-frontend/issues)

hleb-albau:
> > @stackman27 also we should create issue for frontend as well. But i don't know where to do it.
> 
> You can report frontend issues here (https://github.com/osmosis-labs/osmosis-frontend/issues)

Done
https://github.com/osmosis-labs/osmosis-frontend/issues/745

x/gamm/README.md
```diff
@@ -86,8 +86,15 @@ All tokens are swapped using a multi-hop mechanism. That is, all swaps
 are routed via the most cost-efficient way, swapping in and out from
 multiple pools in the process.
 
+When a trade consists of just two OSMO-included routes during a single transaction,
+the swap fees on each hop would be automatically halved. 
+Example: for converting `ATOM -> OSMO -> LUNA` using two pools with swap fees `0.3% + 0.2%`,
+instead `0.15% + 0.1%` fees will be aplied. 
+Note: For pools with low liqudity(or for swaps with high slippage) final discount will be lower than 50%.
```
mattverse, line 93: asking to learn: why is this so?

x/gamm/keeper/multihop.go
```diff
@@ -58,6 +63,12 @@ func (k Keeper) MultihopSwapExactAmountOut(
 
 	insExpected[0] = tokenInMaxAmount
 
+	swapFeeMultiplier := sdk.OneDec()
+
+	if types.SwapAmountOutRoutes(routes).IsOsmoRoutedMultihop() {
+		swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
```
mattverse, line 69: should it not be (5,2)?

x/gamm/types/route.go
```diff
@@ -1,9 +1,16 @@
...
-import sdk "github.com/cosmos/cosmos-sdk/types"
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	appparams "github.com/osmosis-labs/osmosis/v11/app/params"
+)
 
 type SwapAmountInRoutes []SwapAmountInRoute
 
+func (routes SwapAmountInRoutes) IsOsmoRoutedMultihop() bool {
+	return len(routes) == 2 && (routes[0].TokenOutDenom == appparams.BaseCoinUnit)
```
mattverse, line 11: Two questions here:
1. Why do we have len == 2 here? What if route has length of three which all includes Osmo in the routes?
2. Why are we always taking 0 index of routes and expect it to be basecoin unit here?

x/gamm/keeper/multihop_test.go
```diff
@@ -95,6 +136,14 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
 				// Ratio of the token out should be between the before spot price and after spot price.
 				sp := test.param.tokenIn.Amount.ToDec().Quo(tokenOutAmount.ToDec())
 				suite.True(sp.GT(spotPriceBefore) && sp.LT(spotPriceAfter), "test: %v", test.name)
+
```
mattverse, line 139: Can we have test cases that tests with exact numbers on reduce fees?

mattverse (commented):
Nice implementation on this! Left some initial comments and reviews!

x/gamm/types/route.go
hleb-albau: In general the idea that osmo will have a lot of pools X-OSMO
Then to Swap X->Y  all swaps will looks like X -> OSMO -> Y
So to detect such swaps we take the first hop and check that it has output OSMO(that mean that input for the second hop will be osmo too)

What if route has length of three which all includes Osmo in the routes 
- That possible only if we have second hot OSMO->OSMO, or i miss something?

x/gamm/keeper/multihop.go
hleb-albau: (5,2) means 0.05
if we want to half fee, we need to use 0.5 multiplier

x/gamm/keeper/swap.go
```diff
@@ -14,21 +14,22 @@ import (
...
+	swapFeeMultiplier sdk.Dec,
 ) (sdk.Int, error) {
 	pool, err := k.getPoolForSwap(ctx, poolId)
 	if err != nil {
 		return sdk.Int{}, err
 	}
 
-	swapFee := pool.GetSwapFee(ctx)
+	swapFee := pool.GetSwapFee(ctx).Mul(swapFeeMultiplier)
 	return k.swapExactAmountIn(ctx, sender, pool, tokenIn, tokenOutDenom, tokenOutMinAmount, swapFee)
```
czarcas7ic, line 33: Is there a specific reason you decided to have the swapFeeMultiplier as an input to SwapExactAmountIn/Out? It seems like having the logic directly in the SwapExactAmountIn/Out function itself would be potentially safer/cleaner (I could also just totally be missing something!). Lmk what you think

x/gamm/keeper/multihop_test.go
czarcas7ic: Agreed, I don't think GT LT is sufficient for this

x/gamm/keeper/multihop_test.go
```diff
@@ -193,6 +283,13 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
 				sp := tokenInAmount.ToDec().Quo(test.param.tokenOut.Amount.ToDec())
 				fmt.Printf("spBefore %s, spAfter %s, sp actual %s\n", spotPriceBefore, spotPriceAfter, sp)
 				suite.True(sp.GT(spotPriceBefore) && sp.LT(spotPriceAfter), "multi-hop spot price wrong, test: %v", test.name)
+
+				if test.reducedFeeApplied {
+					suite.Require().True(tokenInAmount.LT(tokenInCalculatedAsSeparateSwaps.Amount))
+					suite.Require().True(tokenInAmount.GTE(tokenInCalculatedAsSeparateSwapsWithoutFee.Amount))
+				} else {
+					suite.Require().True(tokenInAmount.Equal(tokenInCalculatedAsSeparateSwaps.Amount))
+				}
```
czarcas7ic, line 292: Same here (changing to specific values rather than GT LT)

x/gamm/keeper/swap.go
```diff
@@ -83,12 +84,13 @@ func (k Keeper) SwapExactAmountOut(
...
 	tokenInMaxAmount sdk.Int,
 	tokenOut sdk.Coin,
+	swapFeeMultiplier sdk.Dec,
 ) (tokenInAmount sdk.Int, err error) {
 	pool, err := k.getPoolForSwap(ctx, poolId)
 	if err != nil {
 		return sdk.Int{}, err
 	}
-	swapFee := pool.GetSwapFee(ctx)
+	swapFee := pool.GetSwapFee(ctx).Mul(swapFeeMultiplier)
```
czarcas7ic, line 93: I think we should protect against negative numbers here (ensuring it is not possible to set a negative swap fee since we are now essentially modifying the pools swap fee). What do you think?

czarcas7ic (commented):
Well done on this! 

I would like to have a discussion on both having the swapFee reduction logic within SwapExactAmountIn/Out and protecting against negative swap fees. After that, IMO its good to go 😄

x/gamm/keeper/swap.go
hleb-albau: I think this place is not where we should force and check for negative swaps fee. If we have here negative swap fee, than something is broken prior this code.

Or you mean, that `swapFeeMultiplier` can't be negative. Is so, that make sense, will add one more check here.

x/gamm/README.md
hleb-albau: ops, seems i just invent it. I try to debug and see real values -> it is always around 50%

hleb-albau: hm, i found where a got 25% - in test of TestBalancerPoolSimpleMultihopSwapExactAmountOut

So for out i always got around 25%. Can't understand now, fill try to figure it out.

hleb-albau: heh, more exact test gives you sometimes nice catches
found a bug in out fee reduction calculation, now it is the same as input - around 50%

x/gamm/keeper/swap.go
hleb-albau: This functions performs only single swaps. That mean, them don't know context executions - whether it is multihop swaps or just single swap. But i feel something as you feel - that it is a bit strange to have that multiplier there. Wii try to create another version where multihop logic will use private keeper functions.

x/gamm/keeper/swap.go
hleb-albau: Check out new version, where i remove this input param. So it is not actual anymore.

x/gamm/keeper/multihop_test.go
```diff
@@ -68,33 +102,48 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
+						suite.Require().NoError(err)
+						nextTokenIn = sdk.NewCoin(hop.TokenOutDenom, tokenOut)
+					}
+					return nextTokenIn
+				}
 
-						pool, err := keeper.GetPoolAndPoke(suite.Ctx, route.PoolId)
-						suite.NoError(err, "test: %v", test.name)
+				tokenOutCalculatedAsSeparateSwaps := calcOutAmountAsSeparateSwaps(0)
+				tokenOutCalculatedAsSeparateSwapsWithoutFee := calcOutAmountAsSeparateSwaps(2)
```
czarcas7ic, line 122: I think what you are doing here is choosing 0 as the poolIdShift so you utilize the actual poolId which then utilizes the real fee. You then use 2 as the poolIdShift to select a non existent pool which has no fee because it doesn't exist.

What do you think about instead exporting the internal swapExactAmountIn (this https://github.com/osmosis-labs/osmosis/blob/c1808cd052160f3d187693b042cfd09892fc4f11/x/gamm/keeper/swap.go#L39-L40) via the export_test.go file (you would need to export it as some other name since the exported SwapExactAmountIn already exists). Idk, I am just thinking of ways to make this more robust / less hacky. Lmk what you think!

If you disagree with this method, I think at the very least we need to explain the hacky method we are doing in the comments (I know you said "use +2 to calc using second pair (w/o swap fee)" but without context this is confusing and took me a bit of time to figure out what was actually going on here)

x/gamm/keeper/multihop_test.go
```diff
@@ -68,33 +102,48 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
+				suite.NoError(err, "test: %v", test.name)
+
+				spotPriceAfter := calcSpotPrice()
 
 				// Ratio of the token out should be between the before spot price and after spot price.
 				sp := test.param.tokenIn.Amount.ToDec().Quo(tokenOutAmount.ToDec())
 				suite.True(sp.GT(spotPriceBefore) && sp.LT(spotPriceAfter), "test: %v", test.name)
+
+				if test.reducedFeeApplied {
+					// here we do not have exact 50% reduce due to amm math rounding and other staff
```
czarcas7ic, line 136: I think you meant "other stuff" here, can we either be more specific about what this other stuff is, or just remove this part at the end?

x/gamm/keeper/multihop_test.go
```diff
@@ -68,33 +102,48 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
+					// playing with input values for this test can result in different discount %
+					// so lets check, that we have around 50% +-1% reduction
+					diffA := tokenOutCalculatedAsSeparateSwapsWithoutFee.Amount.Sub(tokenOutAmount)
+					diffB := tokenOutAmount.Sub(tokenOutCalculatedAsSeparateSwaps.Amount)
+					diffDistinctionPercent := diffA.Sub(diffB).Abs().ToDec().Quo(diffA.Add(diffB).ToDec())
+					suite.Require().True(diffDistinctionPercent.LT(sdk.MustNewDecFromStr("0.01")))
```
czarcas7ic, line 142: Would like to do some work here to make this slightly clearer, such as:

1) for diffB, subtract in the same order as we did for diffA (then add the differences rather than subtract)
2) add comments explaining what exactly diffA and diffB represent

x/gamm/keeper/multihop_test.go
```diff
@@ -38,22 +39,55 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
 
 	for _, test := range tests {
 		// Init suite for each test.
 		suite.SetupTest()
 
 		suite.Run(test.name, func() {
-			// Prepare 2 pools
-			suite.PrepareBalancerPool()
-			suite.PrepareBalancerPool()
+			// Prepare 2 pools pairs
```
czarcas7ic, line 68:
```suggestion
			// Prepare pools
```

x/gamm/keeper/multihop_test.go
```diff
@@ -133,22 +183,55 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
```
czarcas7ic, line 212:
```suggestion
			// Prepare pools
```

x/gamm/keeper/multihop_test.go
```diff
@@ -164,35 +247,46 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
+						suite.Require().NoError(err)
+						nextTokenOut = sdk.NewCoin(hop.TokenInDenom, tokenOut)
+					}
+					return nextTokenOut
+				}
 
-						pool, err := keeper.GetPoolAndPoke(suite.Ctx, route.PoolId)
-						suite.NoError(err, "test: %v", test.name)
+				tokenInCalculatedAsSeparateSwaps := calcInAmountAsSeparateSwaps(0)
+				tokenInCalculatedAsSeparateSwapsWithoutFee := calcInAmountAsSeparateSwaps(2)
```
czarcas7ic, line 267: Same comment on this as above

x/gamm/keeper/multihop_test.go
```diff
@@ -164,35 +247,46 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
 				suite.True(sp.GT(spotPriceBefore) && sp.LT(spotPriceAfter), "multi-hop spot price wrong, test: %v", test.name)
+
+				if test.reducedFeeApplied {
+					// here we do not have exact 50% reduce due to amm math rounding and other staff
+					// playing with input values for this test can result in different discount %
+					// so lets check, that we have around 50% +-1% reduction
+					diffA := tokenInCalculatedAsSeparateSwaps.Amount.Sub(tokenInAmount)
+					diffB := tokenInAmount.Sub(tokenInCalculatedAsSeparateSwapsWithoutFee.Amount)
+					diffDistinctionPercent := diffA.Sub(diffB).Abs().ToDec().Quo(diffA.Add(diffB).ToDec())
+					suite.Require().True(diffDistinctionPercent.LT(sdk.MustNewDecFromStr("0.01")))
```
czarcas7ic, line 286: Same comment on this as above

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#2390](https://github.com/osmosis-labs/osmosis/pull/2390) x/mint remove unused mintCoins parameter from AfterDistributeMintedCoin
 * [#2418](https://github.com/osmosis-labs/osmosis/pull/2418) x/mint remove SetInitialSupplyOffsetDuringMigration from keeper
 * [#2417](https://github.com/osmosis-labs/osmosis/pull/2417) x/mint unexport keeper `SetLastReductionEpochNum`, `getLastReductionEpochNum`, `CreateDeveloperVestingModuleAccount`, and `MintCoins`
+* [#2454](https://github.com/osmosis-labs/osmosis/pull/2454) x/gamm Change `SwapExactAmountIn` and `SwapExactAmountOut` to take fee multiplier as addition input param.
```
czarcas7ic, line 80: Its now the multihop functions that are modified

czarcas7ic (commented):
I think this is a much cleaner / better implementation, nicely done! I just have a few more questions/points. After these addressed and we get one more set of eyes I think this will be good to go

x/gamm/keeper/multihop_test.go
hleb-albau: removed

x/gamm/keeper/multihop_test.go
hleb-albau: remade it with new logic of changing pool swap fee in cached context
previously it uses two pool pairs(what is not oblivious, i agree here.

CHANGELOG.md
hleb-albau: ops

hleb-albau:
@czarcas7ic rework test one more time again

tried different versions with `diffA` `diffB`, seems that variants with comments the most readable. 
Speaking about spot price asserts - that test was here before. I don't know how to make it better, without repeating whole AMM math in test or hard-coding final values. To ensure test is safe, i added assert that multihop output the same as separated swaps. Maybe it is enough(or even further we can remove spot price asserts)?

x/gamm/keeper/multihop_test.go
```diff
@@ -133,70 +189,135 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
+				}
 			} else {
 				_, err := keeper.MultihopSwapExactAmountOut(suite.Ctx, suite.TestAccs[0], test.param.routes, test.param.tokenInMaxAmount, test.param.tokenOut)
 				suite.Error(err, "test: %v", test.name)
 			}
 		})
 	}
 }
+
+func (s *KeeperTestSuite) UpdatePoolSwapFee(ctx sdk.Context, poolId uint64, adjustedPoolSwapFee sdk.Dec) error {
```
czarcas7ic, line 311: I don't think we need to make this function public right?

czarcas7ic (approved):
I think this is in a good place now, well done! Approving and can merge once we get another set of eyes on this.

x/gamm/types/route.go
```diff
@@ -21,6 +28,10 @@ func (routes SwapAmountInRoutes) Validate() error {
 
 type SwapAmountOutRoutes []SwapAmountOutRoute
 
+func (routes SwapAmountOutRoutes) IsOsmoRoutedMultihop() bool {
+	return len(routes) == 2 && (routes[1].TokenInDenom == appparams.BaseCoinUnit)
```
nicolaslara, line 32: Should this support routes longer than 2? We could check that every odd route passes `r.TokenInDenom == appparams.BaseCoinUnit`

hleb-albau: n general the idea that osmo will have a lot of pools X-OSMO
Then to Swap X->Y all swaps will looks like X -> OSMO -> Y
So to detect such swaps we take the first hop and check that it has output OSMO(that mean that input for the second hop will be osmo too)

    What if route has length of three which all includes Osmo in the routes

That possible only if we have second hop OSMO->OSMO, or i miss something?

x/gamm/keeper/multihop_test.go
hleb-albau: fixed

x/gamm/types/route.go
nicolaslara: I was thinking of X -> OSMO -> Y -> OSMO -> Z.

Those can probably be simplified to less hops if there are x/osmo and z/osmo pools, but it isn't guaranteed that they exist.

I don't have a very strong opinion here, just thought it would make sense to future proof the iteration

sunnya97: `X -> OSMO -> Y -> OSMO -> Z` would imply that there are two `Y/OSMO` pools?  Which I think is a case we don't really have to worry about rn

hleb-albau: Also, knowing that there is such mechanic in place, tokens will be listed to OSMO and immediately will be available to all other assets for swaps with discount.

hleb-albau:
@nicolaslara @sunnya97 @stackman27, so any new comments, or can we merge it?

stackman27 (approved):
@hleb-albau no more comments, I think everything looks solid!! 

Thank you for this feature :))

nicolaslara:
No further comments on my end. The PR looks good to me.

hleb-albau:
So...

mattverse (approved):

nicolaslara (approved):

x/gamm/keeper/multihop.go
```diff
@@ -47,8 +58,14 @@ func (k Keeper) MultihopSwapExactAmountOut(
 	tokenInMaxAmount sdk.Int,
 	tokenOut sdk.Coin,
 ) (tokenInAmount sdk.Int, err error) {
+	swapFeeMultiplier := sdk.OneDec()
+
+	if types.SwapAmountOutRoutes(routes).IsOsmoRoutedMultihop() {
+		swapFeeMultiplier = sdk.NewDecWithPrec(5, 1)
```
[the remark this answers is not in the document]
hleb-albau: done

x/gamm/keeper/multihop_test.go
```diff
@@ -38,63 +39,117 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
-					dec := sdk.NewDec(1)
-					tokenInDenom := test.param.tokenIn.Denom
-					for i, route := range test.param.routes {
-						if i != 0 {
-							tokenInDenom = test.param.routes[i-1].TokenOutDenom
-						}
-						pool, err := keeper.GetPoolAndPoke(suite.Ctx, route.PoolId)
-						suite.NoError(err, "test: %v", test.name)
+			// Prepare pools
+			suite.PrepareBalancerPoolWithPoolParams(balancer.PoolParams{
```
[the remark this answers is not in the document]
hleb-albau: Sorry, I don't understand what you suggest here :(

hleb-albau:
rebased for new v12 main

czarcas7ic:
Merging this! This however is not going to make it to v12, simply to reduce the sheer number of changes that need to get reviewed / tested before release.

hleb-albau:
> Merging this! This however is not going to make it to v12, simply to reduce the sheer number of changes that need to get reviewed / tested before release.

It is so small, plus 100% covered by unit tests, plus viewed by at least 6 eyes. Also, manual test here is simple and requires to check only a few cases. Maybe it could be included?

sunnya97:
We need to add support for this in the frontend as well

jonator:
@sunnya97 https://github.com/osmosis-labs/osmosis-frontend/issues/745

Issue #2179, opened 2022-07-21
Make tests for osmoutils/store_helper


ValarDragon:
## Background

In #2175  we added a utils package for `osmoutils/store_helper.go`, These don't have tests associated with them, we should add tests. (They have tests of the functionality in the TWAP logic PR #2168 )

We need to add unit tests for the store_helper file, and then look into if we can more generally use these

## Suggested Design / Acceptance criteria

- [ ] Write unit tests for functions in `store_helper.go`
- [ ] See if they can be reused for other iterators in the codebase

p0mvn:
I will also cover `osmoutils.MustSet`. The rest of the methods are still available

PR #2459, opened 2022-08-19
feat/test(osmoutils): prefix iteration for osmoutils gather; begin testing store

p0mvn:
Part of #2179

This change spearheads store testing of `osmoutils`. Additionally, it adds a convinenet way to iterate over prefix.

In future PRs, the rest of the methods will be tested.

## Testing and Verifying

This change added tests

### DIFF
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -181,3 +181,8 @@ func (s *TestSuite) TestGatherValuesFromIteratorWithStop() {
 		})
 	}
 }
+
+func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
+	s.Require().False(osmoutils.NoStopFn([]byte(keyA)))
+	s.Require().False(osmoutils.NoStopFn([]byte(keyB)))
+}
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -1,0 +1,183 @@
+package osmoutils_test
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/osmosis-labs/osmosis/v11/app/apptesting"
+	"github.com/osmosis-labs/osmosis/v11/osmoutils"
+)
+
+type TestSuite struct {
+	apptesting.KeeperTestHelper
+	store sdk.KVStore
+}
+
+const (
+	keyA               = "a"
+	keyB               = "b"
+	keyC               = "c"
+	mockStopValue      = "stop"
+	afterMockStopValue = mockStopValue + keyA
+	basePrefix         = "base"
+	prefixOne          = "one"
+	prefixTwo          = "two"
+)
+
+func TestOsmoUtilsTestSuite(t *testing.T) {
+	suite.Run(t, new(TestSuite))
+}
+
+func (s *TestSuite) SetupStoreWithBasePrefix() {
+	_, ms := s.CreateTestContextWithMultiStore()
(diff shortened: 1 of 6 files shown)


x/twap/store.go
```diff
@@ -93,12 +93,12 @@ func (k Keeper) getAllMostRecentRecordsForPool(ctx sdk.Context, poolId uint64) (
...
 // getAllHistoricalTimeIndexedTWAPs returns all historical TWAPs indexed by time.
 func (k Keeper) getAllHistoricalTimeIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
-	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPTimeIndexPrefix), []byte(types.HistoricalTWAPTimeIndexPrefixEnd), types.ParseTwapFromBz)
+	return osmoutils.GatherValuesFromStorePrefix(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPTimeIndexPrefix), types.ParseTwapFromBz)
 }
 
 // getAllHistoricalPoolIndexedTWAPs returns all historical TWAPs indexed by pool id.
 func (k Keeper) getAllHistoricalPoolIndexedTWAPs(ctx sdk.Context) ([]types.TwapRecord, error) {
-	return osmoutils.GatherValuesFromStore(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPPoolIndexPrefix), []byte(types.HistoricalTWAPPoolIndexPrefixEnd), types.ParseTwapFromBz)
+	return osmoutils.GatherValuesFromStorePrefix(ctx.KVStore(k.storeKey), []byte(types.HistoricalTWAPPoolIndexPrefix), types.ParseTwapFromBz)
```
ValarDragon, line 101: ah good call with the prefix store iterator!

ValarDragon (approved):
LGTM, nice job!

Leaving unmerged until second reviewer takes a look

x/twap/store.go
p0mvn: Matt's find!

osmoutils/store_helper_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+		expectedValues []string
+		expectedErr    bool
+	}{
+		"prefix iterator, no stop": {
+			preSetKeys: []string{prefixOne + keyA, prefixOne + keyB, prefixOne + keyC},
+
+			prefix: prefixOne,
+
+			expectedValues: []string{"0", "1", "2"},
```
stackman27, line 75: Hi @p0mvn I'm trying to understand how all the iterators work, how are figure out these expected values? Would it the expected value be different if this test case was a `ReverseIterator` or just a regular `Iterator` ?

p0mvn: Hi @stackman27. For every key in `preSetKeys`, we set a value corresponding to its index:
https://github.com/osmosis-labs/osmosis/blob/e5c75ce586dd7861c89c865fee58f013eab88ca6/osmoutils/store_helper_test.go#L142-L144

So, in this case `prefixOne + keyA` -> `0`, `prefixOne + keyB` -> `1`, `prefixOne + keyC` -> `2`

Yes, if it were a reverse iterator, the order would go 2, 1, 0

However, say we had the following insertion order:
- `prefixOne + keyB`
- `prefixOne + keyC`
- `prefixOne + keyA`

Then, the mapping would be
- B -> 0
- C -> 1
- A -> 2

and the expected values are:
```
[]string{"2", "0", "1"}
```

Because iterators use lexicographic order of keys where A < B < C

Let me know if that makes sense

stackman27: aah i see that makes a lot of sense! Thank you :))

PR #2468, opened 2022-08-19
chore: twap mutation fixes part 1

czarcas7ic:
Part Of: #2426 

This PR brings the mutation test score for the twap module from ~33% to ~46%

PRing this now to prevent a large diff for the next / upcoming mutation fixes

## Brief Changelog

- Adds `RunBasicJoin` as a GAMM keeper test helper
- Corrects lexicographical order of twap denoms
  - `getMostRecentRecord` switches order
  - also checks that the denoms are not equal
- Removes ordering from `GetBeginBlockAccumulatorRecord` (this is done in `getMostRecentRecord` which this function calls/returns. Better to have the logic at the higher level)
- Improves `TestGetBeginBlockAccumulatorRecord`

## Testing and Verifying

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -114,3 +114,25 @@ func (s *KeeperTestHelper) RunBasicSwap(poolId uint64) {
 	_, err = s.App.GAMMKeeper.SwapExactAmountIn(s.Ctx, s.TestAccs[0], poolId, msg.TokenIn, denoms[1], msg.TokenOutMinAmount)
 	s.Require().NoError(err)
 }
+
+func (s *KeeperTestHelper) RunBasicJoin(poolId uint64) {
+	pool, _ := s.App.GAMMKeeper.GetPoolAndPoke(s.Ctx, poolId)
+	denoms, err := s.App.GAMMKeeper.GetPoolDenoms(s.Ctx, poolId)
+	s.Require().NoError(err)
+
+	tokenIn := sdk.NewCoins(sdk.NewCoin(denoms[0], sdk.NewInt(1000)), sdk.NewCoin(denoms[1], sdk.NewInt(1000)))
+	s.FundAcc(s.TestAccs[0], tokenIn)
+
+	minShareOutAmt, _, err := pool.CalcJoinPoolShares(s.Ctx, tokenIn, pool.GetSwapFee(s.Ctx))
+	s.Require().NoError(err)
+
+	msg := gammtypes.MsgJoinPool{
+		Sender:         string(s.TestAccs[0]),
+		PoolId:         poolId,
+		ShareOutAmount: minShareOutAmt,
+		TokenInMaxs:    tokenIn,
+	}
+	// TODO: switch to message
+	_, _, err = s.App.GAMMKeeper.JoinPoolNoSwap(s.Ctx, s.TestAccs[0], poolId, msg.ShareOutAmount, msg.TokenInMaxs)
+	s.Require().NoError(err)
+}
--- a/x/twap/README.md
+++ b/x/twap/README.md
@@ -63,7 +63,8 @@ func (k Keeper) GetArithmeticTwap(ctx sdk.Context,
 There are convenience methods for `GetArithmeticTwapToNow` which sets `endTime = ctx.BlockTime()`, and has minor gas reduction.
 For users who need TWAPs outside the 48 hours stored in the state machine, you can get the latest accumulation store record from `GetBeginBlockAccumulatorRecord`.
 
-All functions that call `LexicographicalOrderDenoms` (such as `getMostRecentRecordStoreRepresentation`, `getRecordAtOrBeforeTime`, and `NewTwapRecord`), both directly and indirectly, will swap the provided denoms to be in ascending lexicographical order. This means that regardless of input, asset0Denom is guaranteed to be lexicographically smaller than asset1Denom. If this does not occur, an error will be returned, likely indicating that the two denoms provided are equivalent (which is not allowed).
+All functions that call `LexicographicalOrderDenoms` (such as `getMostRecentRecordStoreRepresentation`, `getRecordAtOrBeforeTime`, and `NewTwapRecord`), both directly and indirectly, will swap the provided denoms to be in ascending lexicographical order. This means that regardless of input, asset0Denom is guaranteed to be lexicographically smaller than asset1Denom.
+If this does not occur, an error will be returned, likely indicating that the two denoms provided are equivalent (which is not allowed).
 
 ## Code layout
 
--- a/x/twap/README.md
+++ b/x/twap/README.md
@@ -63,6 +63,8 @@ func (k Keeper) GetArithmeticTwap(ctx sdk.Context,
 There are convenience methods for `GetArithmeticTwapToNow` which sets `endTime = ctx.BlockTime()`, and has minor gas reduction.
 For users who need TWAPs outside the 48 hours stored in the state machine, you can get the latest accumulation store record from `GetBeginBlockAccumulatorRecord`.
 
+All functions that call `LexicographicalOrderDenoms` (such as `getMostRecentRecordStoreRepresentation`, `getRecordAtOrBeforeTime`, and `NewTwapRecord`), both directly and indirectly, will swap the provided denoms to be in ascending lexicographical order. This means that regardless of input, asset0Denom is guaranteed to be lexicographically smaller than asset1Denom. If this does not occur, an error will be returned, likely indicating that the two denoms provided are equivalent (which is not allowed).
+
 ## Code layout
 
 **api.go** is the main file you should look at as a user of this module.
--- a/x/twap/README.md
+++ b/x/twap/README.md
@@ -63,9 +63,6 @@ func (k Keeper) GetArithmeticTwap(ctx sdk.Context,
 There are convenience methods for `GetArithmeticTwapToNow` which sets `endTime = ctx.BlockTime()`, and has minor gas reduction.
 For users who need TWAPs outside the 48 hours stored in the state machine, you can get the latest accumulation store record from `GetBeginBlockAccumulatorRecord`.
 
-All functions that call `LexicographicalOrderDenoms` (such as `getMostRecentRecordStoreRepresentation`, `getRecordAtOrBeforeTime`, and `NewTwapRecord`), both directly and indirectly, will swap the provided denoms to be in ascending lexicographical order. This means that regardless of input, asset0Denom is guaranteed to be lexicographically smaller than asset1Denom.
-If this does not occur, an error will be returned, likely indicating that the two denoms provided are equivalent (which is not allowed).
-
 ## Code layout
 
 **api.go** is the main file you should look at as a user of this module.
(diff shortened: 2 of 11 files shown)


x/twap/api.go
```diff
@@ -83,9 +83,5 @@ func (k Keeper) GetArithmeticTwapToNow(
 // these swaps have had no time to be arbitraged back.
 // This accumulator can be stored, to compute wider ranged twaps.
 func (k Keeper) GetBeginBlockAccumulatorRecord(ctx sdk.Context, poolId uint64, asset0Denom string, asset1Denom string) (types.TwapRecord, error) {
-	// correct ordering of args for db
-	if asset1Denom > asset0Denom {
-		asset0Denom, asset1Denom = asset1Denom, asset0Denom
-	}
 	return k.getMostRecentRecord(ctx, poolId, asset0Denom, asset1Denom)
```
czarcas7ic, line 86: Note to reviewer: This gets checked/modified in `getMostRecentRecord` which is called in the return statement which is why I removed it here.

x/twap/types/utils.go
```diff
@@ -75,3 +76,13 @@ func SpotPriceTimesDuration(sp sdk.Dec, timeDelta time.Duration) sdk.Dec {
 func AccumDiffDivDuration(accumDiff sdk.Dec, timeDelta time.Duration) sdk.Dec {
 	return accumDiff.QuoInt64(int64(timeDelta))
 }
+
+func LexicographicalOrderDenoms(denom0, denom1 string) (string, string, error) {
```
p0mvn, line 80: Nice! This totally makes sense. Please consider covering with a test

x/twap/api_test.go
```diff
@@ -27,16 +28,18 @@ func (s *TestSuite) TestGetBeginBlockAccumulatorRecord() {
 		poolId     uint64
 		quoteDenom string
 		baseDenom  string
-		expError   bool
+		expError   error
```
czarcas7ic, line 31: Note to reviewer: changed this to specific error type since I realized things were passing with an error, but not the right error we desired. Might want to eventually have an errors type file in twap so we can specify specific error codes instead of manually specifying expected errors here

x/twap/types/utils.go
p0mvn, line 80: Also, since we don't seem to be using this anywhere outside of twap, does it make sense to have it directly in the `x/twap` to avoid leaking logic to the exported scope?

x/twap/keeper_test.go
```diff
@@ -41,7 +41,7 @@ func (s *TestSuite) SetupTest() {
 // sets up a new two asset pool, with spot price 1
 func (s *TestSuite) setupDefaultPool() (poolId uint64, denomA, denomB string) {
 	poolId = s.PrepareUni2PoolWithAssets(defaultUniV2Coins[0], defaultUniV2Coins[1])
-	denomA, denomB = defaultUniV2Coins[1].Denom, defaultUniV2Coins[0].Denom
```
czarcas7ic, line 44: Note to reviewer: This is where a majority of the issue was coming from, we were setting denomA as if it were lexicographically smaller when in reality it was the larger.

x/twap/listeners_test.go
```diff
@@ -50,6 +50,7 @@ func (s *TestSuite) TestSwapAndEndBlockTriggeringSave() {
 	s.Commit() // clear transient store
 	// Now on a clean state after a create pool
 	s.Require().Equal(baseTime.Add(time.Second), s.Ctx.BlockTime())
+	s.RunBasicJoin(poolId)
```
czarcas7ic, line 53: Note to reviewer: This is needed to test the JoinPool twap hook (mutation tests caught this!)

x/twap/types/utils.go
czarcas7ic: Agreed on creating a test for it. This is used in CreateTwapRecord as well though which is in utils.go. Should I keep it in utils because of this?

czarcas7ic: Done!

x/twap/api_test.go
ValarDragon: Yeah, we should make error types. We don't need a code though, golang `errors.Is` will suffice for our usecase here

x/twap/listeners_test.go
ValarDragon: Err, this should be its own test. We need to test on a clean instance, a join in a block will trigger the end tracking.

ValarDragon (approved):
Great job! Using the right order  + this helper really cleans things up!

One change needed for the `RunBasicJoin` test, then I think its good to merge :)

x/twap/types/utils.go
ValarDragon: I do think we should probably move CreateTwapRecord to the main keeper in logic.go, its basically only used there.

I am in favor of keeping lexicographic ordering of denoms here personally tho, its a helper that someone working with TWAPs externally may want

czarcas7ic: Moved CreateTwapRecord to logic.go!

x/twap/listeners_test.go
czarcas7ic: Done!

x/twap/listeners_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/osmosis-labs/osmosis/v11/osmoutils"
+	keeper "github.com/osmosis-labs/osmosis/v11/x/twap"
```
ValarDragon, line 9:
```suggestion
	"github.com/osmosis-labs/osmosis/v11/x/twap"
```

x/twap/listeners_test.go
```diff
@@ -48,8 +49,8 @@ func (s *TestSuite) TestAfterPoolCreatedHook() {
 			denoms := osmoutils.CoinsDenoms(tc.poolCoins)
 			denomPairs0, denomPairs1 := types.GetAllUniqueDenomPairs(denoms)
 			expectedRecords := []types.TwapRecord{}
-			for i := 0; i < len(denomPairs0); i++ {
-				expectedRecord, err := types.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])
+			for i := len(denomPairs0); i < 0; i-- {
+				expectedRecord, err := keeper.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])
```
ValarDragon, line 53:
```suggestion
				expectedRecord, err := twap.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])
```

x/twap/listeners_test.go
```diff
@@ -80,3 +79,46 @@ func (s *TestSuite) TestJoinSwapAndEndBlockTriggeringSave() {
...
+	s.Require().Equal(latestTwap.P1LastSpotPrice, expectedLatestTwapUpToAccum.P1LastSpotPrice)
+
+	latestTwap2, err := s.twapkeeper.GetRecordAtOrBeforeTime(s.Ctx, poolId, s.Ctx.BlockTime(), denom0, denom1)
+	s.Require().NoError(err)
+	s.Require().Equal(latestTwap, latestTwap2)
+
+	// check accumulators incremented - we test details of correct increment in logic
+	s.Require().True(latestTwap.P0ArithmeticTwapAccumulator.GT(historicalOldTwap.P0ArithmeticTwapAccumulator))
+	s.Require().True(latestTwap.P1ArithmeticTwapAccumulator.GT(historicalOldTwap.P1ArithmeticTwapAccumulator))
+}
```
ValarDragon, line 178: err, I think this should just be the 4 line test of: TestSwapTriggeringTrackPoolId

czarcas7ic: yeah you are correct, didn't need all this boiler plate, fixed!

app/apptesting/gamm.go
```diff
@@ -111,3 +111,25 @@ func (s *KeeperTestHelper) RunBasicSwap(poolId uint64) {
...
+	minShareOutAmt, _, err := pool.CalcJoinPoolShares(s.Ctx, tokenIn, pool.GetSwapFee(s.Ctx))
+	s.Require().NoError(err)
+
+	msg := gammtypes.MsgJoinPool{
+		Sender:         string(s.TestAccs[0]),
+		PoolId:         poolId,
+		ShareOutAmount: minShareOutAmt,
+		TokenInMaxs:    tokenIn,
+	}
+	// TODO: switch to message
```
p0mvn, line 132: Are we tracking this in an issue?

p0mvn (approved):

x/twap/listeners_test.go
```diff
@@ -48,8 +49,8 @@ func (s *TestSuite) TestAfterPoolCreatedHook() {
 			denoms := osmoutils.CoinsDenoms(tc.poolCoins)
 			denomPairs0, denomPairs1 := types.GetAllUniqueDenomPairs(denoms)
 			expectedRecords := []types.TwapRecord{}
-			for i := 0; i < len(denomPairs0); i++ {
-				expectedRecord, err := types.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])
+			for i := len(denomPairs0); i < 0; i-- {
+				expectedRecord, err := twap.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, poolId, denomPairs0[i], denomPairs1[i])
```
ValarDragon, line 53: Why are we iterating in reverse order?

x/twap/types/utils.go
```diff
@@ -10,25 +10,6 @@ import (
 	"github.com/osmosis-labs/osmosis/v11/osmoutils"
 )
 
-func NewTwapRecord(k AmmInterface, ctx sdk.Context, poolId uint64, denom0 string, denom1 string) (TwapRecord, error) {
-	if !(denom0 > denom1) {
-		return TwapRecord{}, fmt.Errorf("precondition denom0 > denom1 not satisfied. denom0 %s | denom1 %s", denom0, denom1)
-	}
```
ValarDragon, line 16: If we remove this check here, I think we need to be clear in the spec about guarantees we expect. (Which methods expect correct order, which ones will swap order)

x/twap/listeners_test.go
czarcas7ic: okay, fixed in most recent commit. Will need to write in spec about these guarantees still though

x/twap/types/utils.go
czarcas7ic: Done!

app/apptesting/gamm.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/2484 We are now!

x/twap/README.md
```diff
@@ -63,6 +63,9 @@ func (k Keeper) GetArithmeticTwap(ctx sdk.Context,
 There are convenience methods for `GetArithmeticTwapToNow` which sets `endTime = ctx.BlockTime()`, and has minor gas reduction.
 For users who need TWAPs outside the 48 hours stored in the state machine, you can get the latest accumulation store record from `GetBeginBlockAccumulatorRecord`.
 
+All functions that call `LexicographicalOrderDenoms` (such as `getMostRecentRecordStoreRepresentation`, `getRecordAtOrBeforeTime`, and `NewTwapRecord`), both directly and indirectly, will swap the provided denoms to be in ascending lexicographical order. This means that regardless of input, asset0Denom is guaranteed to be lexicographically smaller than asset1Denom.
+If this does not occur, an error will be returned, likely indicating that the two denoms provided are equivalent (which is not allowed).
```
ValarDragon, line 67: err, this doesn;t really make sense at the spec level / at least not the API layer: Its an abstraction leak for the caller!  It would only apply for the store layer.

PR #2471, opened 2022-08-20
x/mint - rounding; change accumulator semantics to persists truncationd delta across epochs instead of expected distributions

p0mvn:
Closes: #XXX

This PR changes the accumulator semantics so that the rounding fix is not coupled with superfluid minting and burning. As a result, we can now distribute truncations across epochs while not relying on superfluid logic.

This works as follows:

Let epoch provisions be 100.93. Then, we can distribute 100 at the current epoch and have 0.93 left.

Since 0.93 is not a whole integer, we cannot distribute that immediately due to the limitations of the interfaces in x/bank and x/distribution.

Therefore, we persist 0.93 in store until next epoch.

At the next epoch we get 0.93 * 2 = 1.86 truncation delta. We distribute the whole number part (1) and persist 0.86 until the next epoch.

The algorithm repeats every epoch.

As a result, eventually all truncations are distributed as expected.

The above example is simplified. In reality, we have 2 accumulators for distributing from the mint module account and from the developer vesting module account. At the per module account level of abstraction, the idea is the same.


p0mvn:
Merging into a private branch

PR #2472, opened 2022-08-20
improve abstractions for the mint rounding solution

p0mvn:
Closes: #XXX

- create common methods for handling truncations
   - used by mint distributions
   - used by developer rewards distributions
- make distribution functions deeper
- remove obsolete methods


p0mvn:
Merged into private branch

PR #2474, opened 2022-08-21
chore: twap mutation fixes part 2

czarcas7ic:
Closes: #2426 

Brings twap mutation tests to 70%. The remaining failing tests appear to be noise, so this should be the final twap mutation test fix PR.

## Brief Changelog

- specifies error types for store and util tests
- adds zero, negative, and nil cases for genesis validation
- other smaller changes more specific to making mutation tests happy

## Testing and Verifying


stackman27:
Should we find a way to keep track of all mutation test issues PRs? I'm trying to resolve few but dont want to overlap with your work

czarcas7ic:
@stackman27 I was tracking that I was going to take care of both tokenfactory and twap from chain dev call discussions. Would you like to choose another module, make an issue, and then assign to that? That should prevent any overlap!

stackman27:
> @stackman27 I was tracking that I was going to take care of both tokenfactory and twap from chain dev call discussions. Would you like to choose another module, make an issue, and then assign to that? That should prevent any overlap!

Yup i'll do that

x/twap/store.go
```diff
@@ -162,7 +161,11 @@ func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.T
 	}
 
 	for _, twap := range twaps {
-		if twap.Asset0Denom == asset0Denom && twap.Asset1Denom == asset1Denom {
+		if twap.Asset0Denom != asset0Denom {
+			return types.TwapRecord{}, fmt.Errorf(" Provided asset0denom (%s) does not match twap.Asset0Denom (%s)", asset0Denom, twap.Asset0Denom)
+		} else if twap.Asset1Denom != asset1Denom {
+			return types.TwapRecord{}, fmt.Errorf(" Provided asset1denom (%s) does not match twap.Asset1Denom (%s)", asset1Denom, twap.Asset1Denom)
+		} else {
```
czarcas7ic, line 168: While this achieves the same result as what it replaces, it:
1) now passes mutation tests
2) outputs specific errors instead of grouping it into one

x/twap/types/keys.go
```diff
@@ -79,8 +85,11 @@ func ParseTimeFromHistoricalTimeIndexKey(key []byte) time.Time {
 func ParseTimeFromHistoricalPoolIndexKey(key []byte) (time.Time, error) {
 	keyS := string(key)
 	s := strings.Split(keyS, KeySeparator)
-	if len(s) != 5 || s[0] != historicalTWAPPoolIndexNoSeparator {
-		return time.Time{}, fmt.Errorf("Called ParseTimeFromHistoricalPoolIndexKey on incorrectly formatted key: %v", s)
+	if len(s) != expectedLenOfKeySeparators {
+		return time.Time{}, fmt.Errorf("key incorrect length: expected length (%d), actual length (%d)", expectedLenOfKeySeparators, len(s))
+	}
+	if s[0] != historicalTWAPPoolIndexNoSeparator {
+		return time.Time{}, fmt.Errorf("historicalTWAPTimeIndexNoSeparator not formatted correctly: expected separator (%s), actual separator (%v)", historicalTWAPPoolIndexNoSeparator, s[0])
```
czarcas7ic, line 92: Same as previous comment, this achieves the same as what it replaces but now passes mutation tests and gives off specific errors (which can then be specifically tested for)

x/twap/store.go
```diff
@@ -147,7 +147,6 @@ func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.T
 		if lastParsedTime.After(parsedTime) {
 			return true
 		}
-		lastParsedTime = parsedTime
```
ValarDragon, line 150: This is needed, lastParsedTime is defined out of the scope

czarcas7ic: I added it back, but still don't quite understand its purpose. It updates lastParsedTime which is defined out of scope, but as far as I see, lastParsedTime is not used anywhere else but in that scope

x/twap/types/utils.go
```diff
@@ -61,11 +61,12 @@ func AccumDiffDivDuration(accumDiff sdk.Dec, timeDelta time.Duration) sdk.Dec {
 // In other words, the first returned denom string will be the lexicographically smaller of the two denoms.
 // If the denoms are equal, an error will be returned.
 func LexicographicalOrderDenoms(denom0, denom1 string) (string, string, error) {
-	if denom0 == denom1 {
-		return "", "", fmt.Errorf("both assets cannot be of the same denom: assetA: %s, assetB: %s", denom0, denom1)
-	}
-	if denom0 > denom1 {
-		denom0, denom1 = denom1, denom0
+	if denom0 >= denom1 {
```
p0mvn, line 64: I think this change only makes it more confusing by increasing the if/else branches. Do you think we can ignore it?

x/twap/types/keys.go
```diff
@@ -79,8 +85,11 @@ func ParseTimeFromHistoricalTimeIndexKey(key []byte) time.Time {
 func ParseTimeFromHistoricalPoolIndexKey(key []byte) (time.Time, error) {
 	keyS := string(key)
 	s := strings.Split(keyS, KeySeparator)
-	if len(s) != 5 || s[0] != historicalTWAPPoolIndexNoSeparator {
-		return time.Time{}, fmt.Errorf("Called ParseTimeFromHistoricalPoolIndexKey on incorrectly formatted key: %v", s)
+	if len(s) != expectedLenOfKeySeparators {
+		return time.Time{}, fmt.Errorf("key incorrect length: expected length (%d), actual length (%d)", expectedLenOfKeySeparators, len(s))
```
p0mvn, line 89: nit: we can probably define global errors to reuse, including in tests

x/twap/types/utils.go
czarcas7ic: Some of these really are silly, but just thought I would "fix" them all and then we could decide which of these were just simply too ridiculous to implement. Agree this one is confusing, will revert

x/twap/types/keys.go
czarcas7ic: I added an errors.go file and added these errors to it to get the ball rolling! Probably should do a separate PR though to flesh out all errors that should be abstracted to this new errors file

p0mvn: I was thinking of something like this (Go errors): https://github.com/osmosis-labs/osmosis/blob/8e63d6250eccffb6e060da5afb83a669fd287254/x/twap/store.go#L15-L29

Based on previous discussions, I don't think there is much value in registered errors, especially having recently discovered that changing the registered error codes can cause state breakage. Let me know what you think

czarcas7ic: Should be good to go now, thanks!

x/twap/types/errors.go
```diff
@@ -24,3 +24,32 @@ func (e StartTimeAfterEndTimeError) Error() string {
 	return fmt.Sprintf("called GetArithmeticTwap with a start time that is after the end time."+
 		" (start time %s, end time %s)", e.StartTime, e.EndTime)
 }
+
+type KeySeparatorLengthError struct {
```
p0mvn, line 28: Nice!

x/twap/types/keys.go
```diff
@@ -28,6 +28,7 @@ var (
 	mostRecentTWAPsNoSeparator         = "recent_twap"
 	historicalTWAPTimeIndexNoSeparator = "historical_time_index"
 	historicalTWAPPoolIndexNoSeparator = "historical_pool_index"
+	expectedLenOfKeySeparators         = 5
```
p0mvn, line 31: nit: at first I thought that this refers to the number of characters key separators take
```suggestion
	expectedKeySeparatedParts         = 5
```

p0mvn (approved):

x/twap/types/errors.go
ValarDragon, line 28: I feel like these typed errors in most cases isn't a great use of focus

Not opposed to it, but would not prioritize continuing it. (Lots of code / extra thought, for edge cases we likely won't hit after initial writing)

I feel like the approach for improving debugging should be "hit a bad debugging situation, then make a generalized solution to a large class of problems"

I feel like I mostly care about:
* stacktrace / line error was emitted at
* 'wrapped' context for how I got here.

Focus on error types at the bottom isn't how we achieve either.

x/twap/types/genesis_test.go
```diff
@@ -218,12 +253,57 @@ func TestTWAPRecord_Validate(t *testing.T) {
...
-	}
+		"invalid p1 arithmetic accum: negative": {
+			twapRecord: func() TwapRecord {
+				r := baseRecord
+				r.P1ArithmeticTwapAccumulator = sdk.OneDec().Neg()
+				return r
+			}(),
+
+			expectedErr: true,
+		},
```
ValarDragon, line 264: Hrmm, this should be handled by the case for making symmetric test cases?

x/twap/types/genesis_test.go
```diff
@@ -218,12 +253,57 @@ func TestTWAPRecord_Validate(t *testing.T) {
...
+					Time:                        tPlusOne.Add(time.Second),
+					P0LastSpotPrice:             sdk.OneDec(),
+					P1LastSpotPrice:             sdk.OneDec(),
+					P0ArithmeticTwapAccumulator: sdk.OneDec(),
+				}
+				return r
+			}(),
+			expectedErr: true,
+		},
+	}
```
ValarDragon, line 298: Should be handled by the for loop before to make it symmetric

x/twap/types/keys.go
```diff
@@ -63,25 +64,31 @@ func FormatHistoricalPoolIndexTimePrefix(poolId uint64, accumulatorWriteTime tim
 func ParseTimeFromHistoricalTimeIndexKey(key []byte) time.Time {
 	keyS := string(key)
 	s := strings.Split(keyS, KeySeparator)
-	if len(s) != 5 || s[0] != historicalTWAPTimeIndexNoSeparator {
-		panic("Called ParseTimeFromHistoricalTimeIndexKey on incorrectly formatted key")
+	if len(s) != expectedKeySeparatedParts {
+		panic(KeySeparatorLengthError{ExpectedLength: expectedKeySeparatedParts, ActualLength: len(s)})
```
ValarDragon, line 68: I'm so confused, why did we make error types for this

x/twap/types/keys.go
ValarDragon: I am not a fan of this decision, way more cognitive overhead in code, for unclear reasons to me

x/twap/types/genesis_test.go
czarcas7ic: The symmetric loop doesn't work for nil values, so that is why I duplicated them for both sides.

ValarDragon (approved):

Issue #2105, opened 2022-07-17
Tokenfactory: Mutation test issue (July 17th)


ValarDragon:
cref #2069 

The mutation score is 0.400000 (22 passed, 27 failed, 0 duplicated, 6 skipped, total is 55)
Mutation testing score below desired level (0.400000 < 0.75)

 1:
```
--- ./x/tokenfactory/keeper/createdenom.go	2022-07-17 12:29:15.308755888 -0500
+++ /tmp/go-mutesting-332867003/./x/tokenfactory/keeper/createdenom.go.0	2022-07-17 12:37:07.011614397 -0500
@@ -77,7 +77,7 @@
-	if len(creationFee) > 0 {
+	if len(creationFee) >= 0 {
 		if err := k.distrKeeper.FundCommunityPool(ctx, creationFee, accAddr); err != nil {
 			return err
 		}
```

2:  Deleting Create Module Account in InitGenesis
```
+++ /tmp/go-mutesting-332867003/./x/tokenfactory/keeper/genesis.go.0	2022-07-17 12:37:23.927441249 -0500
@@ -9,7 +9,7 @@
 // InitGenesis initializes the tokenfactory module's state from a provided genesis
 // state.
 func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
-	k.CreateModuleAccount(ctx)
+	_, _ = k.CreateModuleAccount, ctx
```

3 - 7: - Probably noise, deleting returning on err
```
+++ /tmp/go-mutesting-332867003/./x/tokenfactory/keeper/genesis.go.5	2022-07-17 12:37:29.591383397 -0500
@@ -19,7 +19,8 @@
 	for _, genDenom := range genState.GetFactoryDenoms() {
 		creator, _, err := types.DeconstructDenom(genDenom.GetDenom())
 		if err != nil {
-			panic(err)
+			_ = err
+
 		}
```

8: - Not sure if noise, deleting set module account. Why isn't a test failing?
```
+++ /tmp/go-mutesting-332867003/./x/tokenfactory/keeper/keeper.go.1	2022-07-17 12:37:36.423313694 -0500
@@ -81,5 +81,5 @@
 // and sends to the relevant address.
 func (k Keeper) CreateModuleAccount(ctx sdk.Context) {
 	moduleAcc := authtypes.NewEmptyModuleAccount(types.ModuleName, authtypes.Minter, authtypes.Burner)
[...]

PR #2477, opened 2022-08-22
tokenfactory: address mutation tests part 1

czarcas7ic:
Addresses: #2105 

Brings mutation tests for tokenfactory from 37% to 48% passing.

## Brief Changelog

- Remove unused function
- No longer charge for invalid create denom attempts
- Expand CreateDenom tests
- Properly test init genesis
- Test CreateModuleAccount for tokenfactory
- Test all tokenfactory msg server messages

## Testing and Verifying


x/tokenfactory/keeper/createdenom.go
```diff
@@ -4,19 +4,20 @@ import (
...
+	denom, err := k.validateCreateDenom(ctx, creatorAddr, subdenom)
 	if err != nil {
 		return "", err
 	}
 
-	denom, err := k.validateCreateDenom(ctx, creatorAddr, subdenom)
+	err = k.chargeForCreateDenom(ctx, creatorAddr, subdenom)
 	if err != nil {
 		return "", err
 	}
```
czarcas7ic, line 23: the order here was changed because we were charging to create denoms with invalid inputs

x/tokenfactory/keeper/createdenom.go
```diff
@@ -77,7 +78,12 @@ func (k Keeper) chargeForCreateDenom(ctx sdk.Context, creatorAddr string, subden
 	if err != nil {
 		return err
 	}
-	if len(creationFee) > 0 {
+	if creationFee != nil {
+		for _, coin := range creationFee {
+			if !k.bankKeeper.HasBalance(ctx, accAddr, coin) {
+				return sdkerrors.ErrInsufficientFunds
+			}
+		}
```
czarcas7ic, line 86: this was added after I found that if the creation fee was multiple denoms, it was possible for a user to be charged if they had one of the two required denoms and still not get to create a tokenfactory token

x/tokenfactory/keeper/genesis.go
```diff
@@ -9,13 +9,14 @@ import (
...
-	k.CreateModuleAccount(ctx)
-
 	if genState.Params.DenomCreationFee == nil {
 		genState.Params.DenomCreationFee = sdk.NewCoins()
 	}
 	k.SetParams(ctx, genState.Params)
 
+	// The call to GetModuleAccount creates a module account if it does not exist.
+	k.accountKeeper.GetModuleAccount(ctx, types.ModuleName)
+
```
czarcas7ic, line 19: I removed the `k.CreateModuleAccount(ctx)` and replaced it with this (this is what we do in other modules, unsure why we changed it here)

x/tokenfactory/keeper/genesis_test.go
```diff
@@ -32,6 +32,9 @@ func (suite *KeeperTestSuite) TestGenesis() {
 		},
 	}
 	app := suite.App
+	// remove module account to ensure initGenesis initializes it on its own
+	tokenfactoryModuleAccount := app.AccountKeeper.GetAccount(suite.Ctx, app.AccountKeeper.GetModuleAddress(types.ModuleName))
+	app.AccountKeeper.RemoveAccount(suite.Ctx, tokenfactoryModuleAccount)
```
czarcas7ic, line 37: If we don't remove the module account here, we are never actually testing if initGenesis creates the module account properly

app/apptesting/test_suite.go
```diff
@@ -142,11 +142,6 @@ func (s *KeeperTestHelper) SetupValidator(bondStatus stakingtypes.BondStatus) sd
 	return valAddr
 }
 
-// SetupTokenFactory sets up a token module account for the TokenFactoryKeeper.
-func (s *KeeperTestHelper) SetupTokenFactory() {
-	s.App.TokenFactoryKeeper.CreateModuleAccount(s.Ctx)
-}
-
```
czarcas7ic, line 149: This isn't used anywhere

x/tokenfactory/keeper/createdenom.go
```diff
@@ -4,19 +4,20 @@ import (
...
 // ConvertToBaseToken converts a fee amount in a whitelisted fee token to the base fee token amount
 func (k Keeper) CreateDenom(ctx sdk.Context, creatorAddr string, subdenom string) (newTokenDenom string, err error) {
-	err = k.chargeForCreateDenom(ctx, creatorAddr, subdenom)
+	denom, err := k.validateCreateDenom(ctx, creatorAddr, subdenom)
 	if err != nil {
 		return "", err
 	}
 
-	denom, err := k.validateCreateDenom(ctx, creatorAddr, subdenom)
+	err = k.chargeForCreateDenom(ctx, creatorAddr, subdenom)
```
czarcas7ic, line 20: this needed to be swapped or else we were charging for invalid denom creation attempts

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -57,19 +58,26 @@ func (suite *KeeperTestSuite) TestMsgCreateDenom() {
 }
 
 func (suite *KeeperTestSuite) TestCreateDenom() {
+	defaultDenomCreationFee := types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(50000000)))}
+	twoDenomCreationFee := types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(50000000)), sdk.NewCoin("uion", sdk.NewInt(50000000)))}
+	nilCreationFee := types.Params{DenomCreationFee: nil}
+	largeCreationFee := types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(5000000000)), sdk.NewCoin("uion", sdk.NewInt(50000000)))}
```
p0mvn, line 64: Can we group these?

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -78,24 +86,53 @@ func (suite *KeeperTestSuite) TestCreateDenom() {
...
 		},
 	} {
+		suite.SetupTest()
 		suite.Run(fmt.Sprintf("Case %s", tc.desc), func() {
 			if tc.setup != nil {
 				tc.setup()
 			}
-			// Create a denom
+			// Set denom creation fee in params
+			keeper.Keeper.SetParams(*suite.App.TokenFactoryKeeper, suite.Ctx, tc.denomCreationFee)
```
p0mvn, line 125: Are we setting params on the right thing? `keeper.Keeper` seems to be a type defintion in the `keeper` package.

Should this be `suite.App.TokenFactoryKeeper`?

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -78,24 +86,53 @@ func (suite *KeeperTestSuite) TestCreateDenom() {
...
+			denomCreationFee := suite.App.TokenFactoryKeeper.GetParams(suite.Ctx).DenomCreationFee
```
p0mvn, line 126: Wdyt about consolitdating all used keepers at the top like:
```
tokenFactoryKeeper := suite.App.TokenFactoryKeeper
...
tokenFactoryKeeper.GetParams(...)
```

It might help with alleviating some of my confusion about what keepers we are setting the parameters on

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -28,13 +29,11 @@ func (suite *KeeperTestSuite) SetupTest() {
 	suite.Setup()
 
 	// Fund every TestAcc with 100 denom creation fees.
-	fundAccsAmount := sdk.NewCoins(sdk.NewCoin(types.DefaultParams().DenomCreationFee[0].Denom, types.DefaultParams().DenomCreationFee[0].Amount.MulRaw(100)))
+	fundAccsAmount := sdk.NewCoins(sdk.NewCoin(types.DefaultParams().DenomCreationFee[0].Denom, types.DefaultParams().DenomCreationFee[0].Amount.MulRaw(100)), sdk.NewCoin("uion", sdk.NewInt(100000000)))
```
p0mvn, line 32: Is it possible to avoid hardcoding "uion" since we are planning to use this denom across the test suite?

I find that in some modules there are various hardcoded denoms in the suite setup. As a result, it is difficult to search and determine which denom is used in which test.

If we move it to a shared variable, the search experience would be much easier. Also, other tests would be more likely to reuse it.

x/tokenfactory/keeper/msg_server_test.go
```diff
@@ -0,0 +1,249 @@
...
+			burnDenom: "factory/osmo1t7egva48prqmzl59x5ngv4zx0dtrwewc9m7z44/evmos",
+			admin:     suite.TestAccs[0].String(),
+			valid:     false,
+		},
+		{
+			desc:                  "success case",
+			burnDenom:             suite.defaultDenom,
+			admin:                 suite.TestAccs[0].String(),
+			valid:                 true,
+			expectedMessageEvents: 2, // TODO: why is this two
```
p0mvn, line 81: Is it possible that these are emitted in both the message server and in the keeper?

I found that to be common in other modules

x/tokenfactory/types/expected_keepers.go
```diff
@@ -19,10 +19,12 @@ type BankKeeper interface {
 	BurnCoins(ctx sdk.Context, moduleName string, amt sdk.Coins) error
 
 	SendCoins(ctx sdk.Context, fromAddr sdk.AccAddress, toAddr sdk.AccAddress, amt sdk.Coins) error
+	HasBalance(ctx sdk.Context, addr sdk.AccAddress, amt sdk.Coin) bool
 }
 
 type AccountKeeper interface {
 	SetModuleAccount(ctx sdk.Context, macc authtypes.ModuleAccountI)
+	GetModuleAccount(ctx sdk.Context, moduleName string) authtypes.ModuleAccountI
```
p0mvn, line 27: Where are we using this? Is this needed?

p0mvn (commented):
Great work! Left some comments, please let me know what you think

x/tokenfactory/keeper/keeper_test.go
czarcas7ic: Done, added a SecondaryDenom and SecondaryAmount in apptesting that we can start using in all other tests. Lmk if this addresses your concern!

x/tokenfactory/keeper/msg_server_test.go
p0mvn: I was searching where that specific type of event is emitted and found that it was erroneously duplicated on the execution path. 

The quickest way to verify that those paths were emitting the events was to set breakpoints in all places where that event is emitted and confirm that they were being hit in different locations.
 
I haven't checked but suspect that something similar might be here too.

app/apptesting/test_suite.go
```diff
@@ -46,6 +46,11 @@ type KeeperTestHelper struct {
 	TestAccs    []sdk.AccAddress
 }
 
+var (
+	SecondaryDenom  = "uion"
```
p0mvn, line 50: Nice

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -57,19 +61,29 @@ func (suite *KeeperTestSuite) TestMsgCreateDenom() {
 }
 
 func (suite *KeeperTestSuite) TestCreateDenom() {
+	var (
+		defaultDenomCreationFee = types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(50000000)))}
+		twoDenomCreationFee     = types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(50000000)), sdk.NewCoin("uion", sdk.NewInt(50000000)))}
+		nilCreationFee          = types.Params{DenomCreationFee: nil}
+		largeCreationFee        = types.Params{DenomCreationFee: sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(5000000000)), sdk.NewCoin("uion", sdk.NewInt(50000000)))}
```
p0mvn, line 68: Does it make sense to use `apptesting.SecondaryDenom` here?

p0mvn (approved):
LGTM once we understand the duplicate event

x/tokenfactory/keeper/msg_server_test.go
czarcas7ic: Hmm so it turns out the bankKeeper is emitting a MsgBurn here https://github.com/osmosis-labs/osmosis/blob/83bd44e30067d5c5a147361696a413f45134b708/x/tokenfactory/keeper/bankactions.go#L51-L52

As well as we emit an event here https://github.com/osmosis-labs/osmosis/blob/83bd44e30067d5c5a147361696a413f45134b708/x/tokenfactory/keeper/msg_server.go#L95-L101

Unsure if I should just remove the manually emitted event or just accept two events being emitted for a single event. Wdyt?

p0mvn: Ahh I see so it seems that the event names for two different components are the same. I think we might consider renaming the tokenfactory burn event since it clashes with bank/

p0mvn: It might make sense to rename most of these by prefixing them with something else such as `tf_mint`, `tf_burn`
https://github.com/osmosis-labs/osmosis/blob/eacd4e21d4b994fb67e7087aa2134d7bdc397d5c/x/tokenfactory/types/msgs.go#L11-L16

The reason is that many of these event names clash with other sdk events so they can't really be useful IMO

p0mvn: If we go with this approach, we need to make sure to have a changelog entry

czarcas7ic: ahhh I wasn't sure where the name clash was happening between sdk and here but that makes sense! Agreed on changing with the tf prefix, will do that and add changelog, thanks!

p0mvn: I double-checked and, actually, "burn" is the only clash.

I still think that we should be prefixing events by module names to avoid such issues. However, we can create a separate issue to discuss this.

For now, I think the approach would be to only rename the tokenfactory "burn" event to mitigate the immediate problem.

p0mvn: Makes sense to me, thank you

p0mvn (approved):

x/tokenfactory/keeper/createdenom.go
ValarDragon: Wait what? FundCommunityPool should call SendCoins which does this check?

Did we have a test for this showing the problem, this seems like something very concerning if true.

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -78,24 +95,55 @@ func (suite *KeeperTestSuite) TestCreateDenom() {
...
 		},
 	} {
+		suite.SetupTest()
 		suite.Run(fmt.Sprintf("Case %s", tc.desc), func() {
 			if tc.setup != nil {
 				tc.setup()
 			}
-			// Create a denom
+			tokenFactoryKeeper := suite.App.TokenFactoryKeeper
+			bankKeeper := suite.App.BankKeeper
```
ValarDragon, line 134: I think this is getting excessive with aliasing. Don't need to revert or anything, but I'd suggest not feeling a strong need to interlace this pattern everywhere imo

x/tokenfactory/keeper/genesis_test.go
ValarDragon: We should make a new suite.SetupGenesisTest() constructor in the future, to not have this be a concern in the future

ValarDragon (approved):
Nice job! Only question is clarity around that FundCommunityPool comment

x/tokenfactory/keeper/createdenom.go
czarcas7ic: We didn't have a test for this, and it only showed its head when I added the `account doesn't have enough to pay for denom creation fee` test case. If you revert just this change and run the test case with prints you will see the following:

```
preCreateBalance 100000000uion,1000000000uosmo 
postCreateBalance 50000000uion,1000000000uosmo 
```

We had enough uion so it deducted that, but we didnt have enough uosmo and stops there.

x/tokenfactory/keeper/genesis_test.go
czarcas7ic: Issue created! https://github.com/osmosis-labs/osmosis/issues/2502

x/tokenfactory/keeper/createdenom.go
ValarDragon: This is an error in the test. 

The way reversion is supposed to work, is that all tx state changes outside of the ante handler get reverted if the message returns an error. So the test is wrong here, the code was right.

The test was for the keeper function, which is the incorrect layer to test reversion logic.

czarcas7ic: Ah I see, my apologies for raising a false flag! Will revert this

PR #2486, opened 2022-08-22
x/mint(docs): ADR for handling truncations

p0mvn:
Closes: #XXX

This is an ADR to propose changes for mitigating `x/mint` truncation issues ref #1917. 
The proof of concept implementation is ready in: https://github.com/osmosis-labs/osmosis/pull/2342

The goal of this ADR is to document the large changes in `x/mint` so that anyone can onboard on
the motivations behind this refactor quickly.

## Testing and Verifying

This change is a trivial rework/code cleanup without any test coverage.

### DIFF
--- a/docs/architecture/adr-001-mint-truncation-refactor.md
+++ b/docs/architecture/adr-001-mint-truncation-refactor.md
@@ -1,4 +1,4 @@
-# ADR 001: Mint Truncation Refactor
+# ADR 001: Mint Truncations Handling Refactor
 
 ## Changelog
 
--- a/docs/architecture/adr-001-mint-truncation-refactor.md
+++ b/docs/architecture/adr-001-mint-truncation-refactor.md
@@ -13,10 +13,10 @@ Draft: https://github.com/osmosis-labs/osmosis/pull/2342
 This ADR focuses on refactoring `x/mint` module to mitigate the discrepancies between the actual and 
 the projected inflation amounts.
 
-Currently, we under mint due to truncation. In the first year of operations, this happens to be approximately 2600 OSMO. As a result, we cannot reach the projected supply one to one.
+Currently, we under-mint due to truncations. In the first year of operations, this happens to be approximately 2600 OSMO. As a result, we cannot reach the projected OSMO supply one-to-one.
 
-Additionally, some of the constraints have made it difficult to refactor the module. Specifically, the developer vesting provisions 
-follow a distinct distribution logic from the rest of the provisions. While that is the case, these providiond are tightly coupled together, causing to utilize unsafe workarounds such as over-minting and later burning developer vesting provisions. This ADR addresses these issues by decoupling the two kinds of provisions and letting them use separate distribution logic.
+Additionally, some of the constraints have made it difficult to refactor the `x/mint` module. Specifically, the developer vesting provisions 
+follow a distinct distribution logic from the rest of the provisions. While that is the case, these provisions are tightly coupled together, causing the utilization of unsafe workarounds such as over-minting and later burning developer vesting provisions. This ADR addresses these issues by decoupling the two kinds of provisions and letting them use separate distribution logic.
 
 ## Context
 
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-001-mint-truncation-refactor.md
```diff
@@ -0,0 +1,284 @@
...
+#### Consequences
+
+##### Positive
+
+This change allows for better encapsulation of all distribution logic and for
+the ability to more thoroughly unit test it.
+
+The enhanced encapsulation in turn leads to a more modular mint keeper where
+each method strives to focus on a single concern. 
+
```
czarcas7ic, line 226: Just going off the format you have been using, does this imply there are no negatives to this new encapsulation? (Not disagreeing, just ensuring that was purposeful)

docs/architecture/adr-001-mint-truncation-refactor.md
```diff
@@ -0,0 +1,284 @@
...
+### Distributing Truncation Delta
+
+For any truncation deltas occurring at epochs before the proposed
+implementation is live, it will have to be manually estimated and minted/burned
+in the upgrade handler.
+
+The implementation proposed in this ADR is independent of the need for
+distributing the old truncation deltas. It only ensures that there are
+no more discrepancies after the proposed proof-of-concept is deployed.
```
czarcas7ic, line 242: I think this is an important point. 

Do you forsee potentially decoupling the implementation since its such a large change? First, implement this ADR. Then when we see that the truncation works as intended, we can manually estimate the truncations that should have happened prior and implement them in the next upgrade. 

Not married to the idea, just seems like it is something punt-able that might make this simpler to implement in stages.

czarcas7ic (approved):
Really nice job on this! I feel like this is a standard we should begin to follow on all large architecture changes

docs/architecture/adr-001-mint-truncation-refactor.md
p0mvn: Yes, the ADR is already implemented in #2342. The remaining work is just splitting it into smaller chunks / PRs.

With regards to distributing the old delta, there is already logic for estimating it. We can just add the correct number in the upgrade handler, once the upgrade height / last epoch number before the upgrade is known. So it would be a separate / decoupled change from #2342.

We can even do it in the next major upgrade to make sure that we get the last epoch number before the upgrade with the fix right.

docs/architecture/adr-001-mint-truncation-refactor.md
p0mvn: Yes, I just could not think of any negatives :smile:

Feels like encapsulating this would only make it cleaner and more testable. 

If I'm missing any negatives, please let me know

p0mvn:
> Really nice job on this! I feel like this is a standard we should begin to follow on all large architecture changes

Thank you. The format is borrowed 1:1 from the sdk: https://github.com/osmosis-labs/cosmos-sdk/tree/osmosis-main/docs/architecture

I found it helpful to understand some historic decisions in that repository so hoping this would serve a similar purpose here

docs/architecture/adr-001-mint-truncation-refactor.md
```diff
@@ -0,0 +1,336 @@
...
+
+Draft: <https://github.com/osmosis-labs/osmosis/pull/2342>
+
+## Abstract
+
+This ADR focuses on refactoring `x/mint` module to mitigate the discrepancies
+between the actual and the projected inflation amounts.
+
+Currently, we under-mint due to truncations. In the first year of operations,
+this happens to be approximately 2600 uosmo. As a result, we cannot reach the
```
czarcas7ic, line 17: the number is that small? That is surprising

p0mvn: Yes, I've explored the idea of removing the indexes completely to only keep the refactor for decoupling module accounts here:
https://github.com/osmosis-labs/osmosis/tree/roman/mint-refactor

However, I think the complexity overhead is not large enough relative to the precision benefit. Let me know what you think

PR #2498, opened 2022-08-23
feat(osmoutils/mint): mint truncations pt1 - store helpers

p0mvn:
Closes: #XXX

This is the first PR for merging the mint truncations solution.

ADR is available in: https://github.com/osmosis-labs/osmosis/pull/2486
This change is extracted from: https://github.com/osmosis-labs/osmosis/pull/2342

This change only includes store helpers used from the original mint truncations solution as well as their tests.

In addition, it covers an already existing function `MustSet` with tests.

## Testing and Verifying

This change added tests

### DIFF
--- a/osmoutils/store_helper.go
+++ b/osmoutils/store_helper.go
@@ -2,6 +2,7 @@ package osmoutils
 
 import (
 	"errors"
+	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	db "github.com/tendermint/tm-db"
@@ -102,4 +103,30 @@ func gatherValuesFromIteratorWithStop[T any](iterator db.Iterator, parseValue fu
 
 func noStopFn([]byte) bool {
 	return false
+}
+
+// MustGet gets key from store by mutating result
+// Panics on any error.
+func MustGet(store store.KVStore, key []byte, result proto.Message) {
+	b := store.Get(key)
+	if b == nil {
+		panic(fmt.Errorf("getting at key (%v) should not have been nil", key))
+	}
+	if err := proto.Unmarshal(b, result); err != nil {
+		panic(err)
+	}
+}
+
+// MustSetDec sets dec value to store at key. Panics on any error.
+func MustSetDec(store store.KVStore, key []byte, value sdk.Dec) {
+	MustSet(store, key, &sdk.DecProto{
+		Dec: value,
+	})
+}
+
+// MustGetDec gets dec value from store at key. Panics on any error.
+func MustGetDec(store store.KVStore, key []byte) sdk.Dec {
+	result := &sdk.DecProto{}
+	MustGet(store, key, result)
+	return result.Dec
 }
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -274,7 +274,7 @@ func (s *TestSuite) TestMustGet() {
 // and panics if an error is encountered.
 func (s *TestSuite) TestMustSet() {
 	tests := map[string]struct {
-		// keyys and values to preset
+		// keys and values to preset
 		setKey   string
 		setValue proto.Message
 
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -363,7 +363,8 @@ func (s *TestSuite) TestMustGetDec() {
 			},
 
 			expectedGetKeyValues: map[string]sdk.Dec{
-				keyB: sdk.OneDec().Add(sdk.OneDec()),
+				keyA: sdk.OneDec(),
+				keyB: sdk.Dec{}, // this one panics
 			},
 
 			expectPanic: true,
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -394,6 +394,11 @@ func (s *TestSuite) TestMustGetDec() {
 
 // TestMustSetDec tests that MustSetDec updates the store correctly
 // with the right decimal value.
+// N.B.: It is non-trivial to cause a panic
+// by calling `MustSetDec` because it provides
+// a valid proto argument to `MustSet` which will
+// only panic if the proto argument is invalid.
+// Therefore, we only test a success case here.
 func (s *TestSuite) TestMustSetDec() {
 	// Setup.
 	s.SetupStoreWithBasePrefix()
--- a/osmoutils/store_helper_test.go
+++ b/osmoutils/store_helper_test.go
@@ -339,7 +339,7 @@ func (s *TestSuite) TestMustGetDec() {
 		preSetKeyValues map[string]sdk.Dec
 
 		// keys and values to attempt to get and validate
-		getKeyValues map[string]sdk.Dec
+		expectedGetKeyValues map[string]sdk.Dec
 
 		expectPanic bool
 	}{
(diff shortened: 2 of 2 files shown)


osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
 	s.Require().False(osmoutils.NoStopFn([]byte(keyA)))
 	s.Require().False(osmoutils.NoStopFn([]byte(keyB)))
 }
+
+// TestMustGet tests that MustGet retrieves the correct
+// values from the store and panics if an error is encountered.
+func (s *TestSuite) TestMustGet() {
+
+	tests := map[string]struct {
+		// keyys and values to preset
```
czarcas7ic, line 198:
```suggestion
		// keys and values to preset
```

osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
...
+			for key, value := range tc.preSetKeyValues {
+				osmoutils.MustSet(s.store, []byte(key), value)
+			}
+
+			osmoassert.ConditionalPanic(s.T(), tc.expectPanic, func() {
+				for key, expectedValue := range tc.getKeyValues {
+					// System under test.
+					osmoutils.MustGet(s.store, []byte(key), tc.actualResultProto)
+					// Assertions.
+					s.Require().Equal(expectedValue.String(), tc.actualResultProto.String())
```
czarcas7ic, line 266: All the logic here makes sense, just had a design question. 

Normally within the table driven tests we have done, we specify the expected value in the table and then within the test logic itself we get the actual value and compare the two. It seems here that those events were flipped. Was this intentional / does it not really matter? Or should we stick to the same way for all table driven tests?

osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
...
+			})
+		})
+	}
+}
+
+// TestMustSet tests that MustSet updates the store correctly
+// and panics if an error is encountered.
+func (s *TestSuite) TestMustSet() {
+	tests := map[string]struct {
+		// keyys and values to preset
```
czarcas7ic, line 277:
```suggestion
		// keys and values to preset
```

osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
...
+		})
+	}
+}
+
+// TestMustGetDec tests that MustGetDec retrieves the correct
+// decimal values from the store and panics if an error is encountered.
+func (s *TestSuite) TestMustGetDec() {
+
+	tests := map[string]struct {
+		// keyys and values to preset
```
czarcas7ic, line 338:
```suggestion
		// keys and values to preset
```

osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
...
+		},
+		"attempt to get non-existent key - panic": {
+			preSetKeyValues: map[string]sdk.Dec{
+				keyA: sdk.OneDec(),
+				keyC: sdk.OneDec().Add(sdk.OneDec()).Add(sdk.OneDec()),
+			},
+
+			getKeyValues: map[string]sdk.Dec{
+				keyB: sdk.OneDec().Add(sdk.OneDec()),
+			},
```
czarcas7ic, line 367: Can we modify this so that the getKeyValues contains both an existing and non existing value (i.e. just add keyA to getKeyValues?) That or leave this be and add another test?

osmoutils/store_helper_test.go
```diff
@@ -186,3 +189,220 @@ func (s *TestSuite) TestNoStopFn_AlwaysFalse() {
...
+
+	originalDecValue := sdk.OneDec()
+
+	// System under test.
+	osmoutils.MustSetDec(s.store, []byte(keyA), originalDecValue)
+
+	// Assertions.
+	retrievedDecVaue := osmoutils.MustGetDec(s.store, []byte(keyA))
+	s.Require().Equal(originalDecValue.String(), retrievedDecVaue.String())
+}
```
czarcas7ic, line 408: I assume we are not testing the fail case because MustSet is already covered right?

czarcas7ic (approved):
Well done! Super easy/clear to review 🚀

osmoutils/store_helper_test.go
p0mvn: The intention here was to test multiple gets in sequence, that's why these are surrounded by a loop and the expected is map containing multiple expected values.

I renamed `tc.GetKeyValues` to `tc.expectedGetKeyValues` to make it clearer. Let me know what you think

osmoutils/store_helper_test.go
p0mvn: No, I don't think it is possible to get a panic in this case because the only reason `MustSet` could panic if we provide an invalid proto type:
https://github.com/osmosis-labs/osmosis/blob/3e334d042f298112a5704be410b77e9f1432d0d9/osmoutils/store_helper.go#L81-L84

However, in `MustSetDec` we pretty much hardcoded a valid proto:
https://github.com/osmosis-labs/osmosis/blob/3e334d042f298112a5704be410b77e9f1432d0d9/osmoutils/store_helper.go#L122-L124

I will add a comment about this

osmoutils/store_helper_test.go
czarcas7ic: yeah that is clearer now, thanks!

osmoutils/store_helper_test.go
czarcas7ic: Ahh that makes sense, looks good then!

p0mvn:
Planning on merging this tomorrow to unblock further work. If anyone has capacity, please take a look

osmoutils/store_helper.go
```diff
@@ -103,3 +104,29 @@ func gatherValuesFromIteratorWithStop[T any](iterator db.Iterator, parseValue fu
...
+		panic(err)
+	}
+}
+
+// MustSetDec sets dec value to store at key. Panics on any error.
+func MustSetDec(store store.KVStore, key []byte, value sdk.Dec) {
+	MustSet(store, key, &sdk.DecProto{
+		Dec: value,
+	})
+}
```
ValarDragon, line 125: Do we just not write Dec's as raw state entries anywhere in our codebase?

osmoutils/store_helper.go
```diff
@@ -103,3 +104,29 @@ func gatherValuesFromIteratorWithStop[T any](iterator db.Iterator, parseValue fu
...
+
+// MustGetDec gets dec value from store at key. Panics on any error.
+func MustGetDec(store store.KVStore, key []byte) sdk.Dec {
+	result := &sdk.DecProto{}
+	MustGet(store, key, result)
+	return result.Dec
+}
```
ValarDragon, line 132: Do we just not write Dec's as raw state entries anywhere in our codebase?

ValarDragon (approved):

osmoutils/store_helper.go
p0mvn: Technically, it is possible to `json.Marshal` `sdk.Dec` but I was aiming to reuse `MustGet` since the error checking seems pretty much identical, as a result, increasing boilerplate.

using a `proto.Message` wrapper for `Dec` allows to reust the `MustGet` and `MustSet`, reducing boilerplate.

Let me know what you think. I can change this in a future PR if you think json marshaling is preferable

PR #2503, opened 2022-08-24
feat(x/mint): add truncation delta store indexes

p0mvn:
Closes: #XXX

This is the third PR for merging the mint truncations solution.

ADR is available in: https://github.com/osmosis-labs/osmosis/pull/2486
This change is extracted from: https://github.com/osmosis-labs/osmosis/pull/2342

This change focuses on adding 2 new store indexes in the mint module. One for storing truncation delta across epochs for mint module account. Another one is for storing truncation delta for the developer vesting module account.

The idea is that after distributing both inflation and developer vesting provisions, we will call the `handleTruncationDelta(...)` method that will take care of the distribution of any truncations from this and previous epoch as well as persist any remaining truncations for the next.

For example, assume that for some number of epochs our expected provisions are `100.6` and the actual amount distributed is `100` due to truncations. Then, at epoch `1`, we have a delta of `0.6`. `0.6` cannot be distributed because it is not an integer. So we persist it until the next epoch. Then, at epoch `2`, we get a delta of `1.2` (`0.6` from epoch `1` and `0.6` from epoch `2`). Now, `1` can be distributed and `0.2` persisted until the next epoch (`3`).

**In this PR, only the new methods have been added that are not hooked up to the core logic yet.**

## Testing and Verifying

This change added tests


x/mint/keeper/genesis_test.go
```diff
@@ -135,6 +135,15 @@ func (suite *KeeperTestSuite) TestMintInitGenesis() {
 			expectedDeveloperVestingAmount := tc.expectedDeveloperVestingAmountDelta.Add(originalVestingCoins.Amount)
 			suite.Require().Equal(expectedDeveloperVestingAmount.Int64(), actualVestingCoins.Amount.Int64())
 
+			// Validate that truncation deltas are inirialized to zero.
```
p0mvn, line 138:
```suggestion
			// Validate that truncation deltas are initialized to zero.
```

x/mint/keeper/keeper.go
```diff
@@ -121,6 +115,30 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
 
+// GetInflationTruncationDelta returns the truncation delta.
+// Panics if key is invalid.
```
p0mvn, line 119:
```suggestion
// Only developer vesting and mint module accounts are allowed.
// Returns error if invalid module account. Panics if key is invalid.
```

x/mint/keeper/keeper.go
```diff
@@ -121,6 +115,30 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
...
+func (k Keeper) GetTruncationDelta(ctx sdk.Context, moduleAccountName string) (sdk.Dec, error) {
+	storeKey, err := getTruncationStoreKeyFromModuleAccount(moduleAccountName)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+	return osmoutils.MustGetDec(ctx.KVStore(k.storeKey), storeKey), nil
+}
+
+// SetTruncationDelta sets the truncation delta, returning an error if any. nil otherwise.
```
p0mvn, line 128:
```suggestion
// SetTruncationDelta sets the truncation delta, returning an error if any. nil otherwise.
// Only developer vesting and mint module accounts are allowed.
// Returns error if invalid module account.
```

x/mint/keeper/keeper.go
```diff
@@ -279,15 +297,23 @@ func (k Keeper) distributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.
...
-// getProportions gets the balance of the `MintedDenom` from minted coins and returns coins according to the
-// allocation ratio. Returns error if ratio is greater than 1.
-// TODO: this currently rounds down and is the cause of rounding discrepancies.
-// To be fixed in: https://github.com/osmosis-labs/osmosis/issues/1917
-func getProportions(mintedCoin sdk.Coin, ratio sdk.Dec) (sdk.Coin, error) {
-	if ratio.GT(sdk.OneDec()) {
-		return sdk.Coin{}, invalidRatioError{ratio}
+// calculateTotalTruncationDelta returns the total truncation delta that has not been distributed yet
+// for the given key given provisions and amount distributed. Both of the given values are epoch specific.
+// The returned delta might include the delta from previous epochs if they have not been distributed due to truncations.
```
p0mvn, line 302:
```suggestion
// for the given module account, provisions, and actual amount distributed in the current epoch.
// The returned delta might include the delta from previous epochs if they have not been distributed due to truncations.
```

x/mint/keeper/keeper.go
```diff
@@ -315,3 +341,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
 	}
 	return nil
 }
+
+// handleTruncationDelta estimates and distributes truncation delta from either mint module account or
+// developer vesting module account. Returns the total amount distributed from truncations.
+// If truncations are estimated to be less than one, persists them in store until the next epoch without
+// any distributions during the current epoch.
+// More on why this handling truncations is necessary: due to limitations of some SDK interfaces that operate on integers,
```
p0mvn, line 349:
```suggestion
// More on why this truncations handling is necessary: due to limitations of some SDK interfaces that operate on integers,
```

x/mint/keeper/keeper.go
```diff
@@ -315,3 +341,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+// there are known truncation differences from the expected total epoch mint provisions.
+// To use these interfaces, we always round down to the nearest integer by truncating decimals.
+// As a result, it is possible to undermint. To mitigate that, we distribute any delta to the community pool.
+// The delta is calculated by subtracting the actual distributions from the given expected total distributions
+// and adding it to any left overs from the previous epoch. The left overs might be stemming from the inability
+// to distribute decimal truncations less than 1. As a result, we store them in the store until the next epoch.
+// These truncation distributions have eventual guarantees. That is, they are guaranteed to be distributed
+// eventually but not necessarily during the same epoch.
+// Returns error if module account name is other than mint or developer vesting is given.
```
p0mvn, line 358:
```suggestion
// Returns an error if the module account name is any other than mint or developer vesting is given.
```

x/mint/keeper/keeper.go
```diff
@@ -315,3 +341,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+// getProportions gets the balance of the `MintedDenom` from minted coins and returns coins according to the
+// allocation ratio. Returns error if ratio is greater than 1.
+// TODO: this currently rounds down and is the cause of rounding discrepancies.
+// To be fixed in: https://github.com/osmosis-labs/osmosis/issues/1917
+func getProportions(mintedCoin sdk.Coin, ratio sdk.Dec) (sdk.Coin, error) {
+	if ratio.GT(sdk.OneDec()) {
+		return sdk.Coin{}, invalidRatioError{ratio}
+	}
+	return sdk.NewCoin(mintedCoin.Denom, mintedCoin.Amount.ToDec().Mul(ratio).TruncateInt()), nil
+}
```
p0mvn, line 422: Note to reviewers: this method is unchanged, just moved down

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
 		})
 	}
 }
+
+// TestCalculateTruncationsDelta tests that the per-epoch truncation delta is calculate
```
p0mvn, line 723:
```suggestion
// TestCalculateTruncationsDelta tests that the per-epoch truncation delta is calculated
```

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+// correctly. That includes checking that the pre-existing delta is correctly
```
p0mvn, line 724:
```suggestion
// correctly. That includes checking that the pre-existing delta is
```

x/mint/types/keys.go
```diff
@@ -1,12 +1,35 @@
...
+	// MinterKey is the key to use for the keeper store at which
+	// the Minter and its EpochProvisions are stored.
+	MinterKey = []byte{0x00}
 
-// LastReductionEpochKey is the key to use for the keeper store
-// for storing the last epoch at which reduction occurred.
-var LastReductionEpochKey = []byte{0x03}
+	// LastReductionEpochKey is the key to use for the keeper store
+	// for storing the last epoch at which reduction occurred.
+	LastReductionEpochKey = []byte{0x03}
```
p0mvn, line 10: Note to reviewers: these are unchanged,

x/mint/keeper/genesis.go
```diff
@@ -18,6 +18,13 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
 	k.SetMinter(ctx, data.Minter)
 	k.SetParams(ctx, data.Params)
 
+	if err := k.SetTruncationDelta(ctx, types.DeveloperVestingModuleAcctName, sdk.ZeroDec()); err != nil {
+		panic(err)
+	}
+	if err := k.SetTruncationDelta(ctx, types.ModuleName, sdk.ZeroDec()); err != nil {
+		panic(err)
+	}
+
```
czarcas7ic, line 27: Can we just add a comment above each of these saying that the top one is initializing developer module account and the next one is initializing the mint module account? It just threw me off initially because the top one is explicitly calling the module account while the second one is calling for the entire module name

x/mint/keeper/keeper.go
```diff
@@ -121,6 +115,33 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
 
+// GetInflationTruncationDelta returns the truncation delta.
```
czarcas7ic, line 118:
```suggestion
// GetTruncationDelta returns the truncation delta.
```

x/mint/keeper/keeper.go
```diff
@@ -279,15 +300,23 @@ func (k Keeper) distributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.
...
-		return sdk.Coin{}, invalidRatioError{ratio}
+// calculateTotalTruncationDelta returns the total truncation delta that has not been distributed yet
+// for the given module account, provisions, and actual amount distributed in the current epoch.
+// The returned delta might include the delta from previous epochs if they have not been distributed due to truncations.
+// For example, assume that for some number of epochs our expected provisions
+// are 100.6 and the actual amount distributed is 100 every epoch due to truncations.
+// Then, at epoch 1, we have a delta of 0.6. 0.6 cannot be distributed because it is not an integer.
+// So we persist it until the next epoch. Then, at at epoch 2, we get a delta of 1.2 (0.6 from epoch 1 and 0.6 from epoch 2).
+// Now, 1 can be distributed and 0.2 gets persisted until the next epoch.
+// CONTRACT: provisions are greater than or equal to amountDistributed.
```
czarcas7ic, line 311: What does contract here mean?

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+// any distributions during the current epoch.
+// More on why this truncations handling is necessary: due to limitations of some SDK interfaces that operate on integers,
+// there are known truncation differences from the expected total epoch mint provisions.
+// To use these interfaces, we always round down to the nearest integer by truncating decimals.
+// As a result, it is possible to undermint. To mitigate that, we distribute any delta to the community pool.
+// The delta is calculated by subtracting the actual distributions from the given expected total distributions
+// and adding it to any left overs from the previous epoch. The left overs might be stemming from the inability
+// to distribute decimal truncations less than 1. As a result, we store them in the store until the next epoch.
+// These truncation distributions have eventual guarantees. That is, they are guaranteed to be distributed
+// eventually but not necessarily during the same epoch.
```
czarcas7ic, line 360: nit: This technically isn't a guarantee right? Eventually emissions will become so low that the accumulator will never get past 1 right?

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+// Returns an error if the module account name is any other than mint or developer vesting is given.
+// The truncation delta is calculated by subtracting amountDistributed from probisions and adding to
```
czarcas7ic, line 362:
```suggestion
// The truncation delta is calculated by subtracting amountDistributed from provisions and adding to
```

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+// any leftover truncations from the previous epoch.
+// Therefore, provisions must be greater than or equal to the amount distributed. Errors if not.
+// For any amount to be distributed from the mint module account, it mints the estimated truncation amount
+// before distributing it to the community pool.
+// Additionally, it errors in the following cases:
+// - unable to mint tokens
+// - unable to fund the community pool
+// nolint: unused
+func (k Keeper) handleTruncationDelta(ctx sdk.Context, moduleAccountName string, provisions sdk.DecCoin, amountDistributed sdk.Int) (sdk.Int, error) {
+	if moduleAccountName != types.DeveloperVestingModuleAcctName && moduleAccountName != types.ModuleName {
```
czarcas7ic, line 372:
```suggestion
	if moduleAccountName != types.DeveloperVestingModuleAcctName || moduleAccountName != types.ModuleName {
```
right?

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+	// N.B: Truncation is acceptable because we check delta at the end of every epoch.
+	// As a result, actual minted distributions always approach the expected value.
+	truncationDeltaToDistribute := deltaAmount.TruncateInt()
+	// For funding from mint module account, we must pre-mint first.
+	if moduleAccountName == types.ModuleName {
+		if err := k.mintCoins(ctx, sdk.NewCoins(sdk.NewCoin(provisions.Denom, truncationDeltaToDistribute))); err != nil {
+			return sdk.Int{}, err
+		}
+	}
+	if err := k.communityPoolKeeper.FundCommunityPool(ctx, sdk.NewCoins(sdk.NewCoin(provisions.Denom, truncationDeltaToDistribute)), k.accountKeeper.GetModuleAddress(moduleAccountName)); err != nil {
```
czarcas7ic, line 399: I was not clear on this before...so mint truncations get distributed to the community pool, whereas dev truncations get distributed to dev rewards correct? If we weren't explicit about this in the writeup we should add it in there (its possible you did, I will have to go back and check)

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+
+	newDelta := deltaAmount.Sub(truncationDeltaToDistribute.ToDec())
+
+	if err := k.SetTruncationDelta(ctx, moduleAccountName, newDelta); err != nil {
+		return sdk.Int{}, err
+	}
+
+	if truncationDeltaToDistribute.IsInt64() {
+		defer telemetry.ModuleSetGauge(types.ModuleName, float32(truncationDeltaToDistribute.Int64()), fmt.Sprintf("mint_truncation_distributed_%s_delta", moduleAccountName))
+	}
```
czarcas7ic, line 411: Can you explain this further?

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+		"100.6 - 100  + 0.3 = 0.9; pre-existing delta of 0.3; inflation delta module account": {
+			preExistingStoreDelta: sdk.NewDecWithPrec(3, 1),
+
+			moduleAccountName: types.ModuleName,
+			provisions:        sdk.NewDecWithPrec(1006, 1),
+			amountDistributed: sdk.NewInt(100),
+
+			expectedTruncationDelta: sdk.NewDecWithPrec(9, 1),
+		},
+		"100.6 - 100  + 0.5 = 1.1; pre-existing delta of 1.1; inflation delta module account": {
```
czarcas7ic, line 756:
```suggestion
		"100.6 - 100  + 0.5 = 1.1; pre-existing delta of 0.5; inflation delta module account": {
```
right?

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+			preExistingStoreDelta: sdk.NewDecWithPrec(5, 1),
+
+			moduleAccountName: types.ModuleName,
+			provisions:        sdk.NewDecWithPrec(1006, 1),
+			amountDistributed: sdk.NewInt(100),
+
+			expectedTruncationDelta: sdk.NewDecWithPrec(11, 1),
+		},
+		"82191.178 - 17808  + 33.33 = 64416508; pre-existing delta of 33.33; developer vesting module account": {
```
czarcas7ic, line 765: I though the delta shouldn't exceed 1?

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+		},
+		"invalid module account - error": {
+			moduleAccountName: poolincentivestypes.ModuleName,
+			provisions:        sdk.NewDecWithPrec(1006, 1),
+			amountDistributed: sdk.NewInt(100),
+
+			expectError: true,
+		},
+	}
+	for name, tc := range tests {
```
czarcas7ic, line 789: I think we absolutely need to test a case where the overflow of the accumulator happens over greater than one epoch. In other words: pre-existing delta of 0.1, and the next two epochs have a delta of 0.4

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+			amountDistributed: sdk.NewInt(100),
+
+			expectError: true,
+		},
+		"invalid module account - error": {
+			moduleAccountName: poolincentivestypes.ModuleName,
+			provisions:        sdk.NewDecWithPrec(1006, 1),
+			amountDistributed: sdk.NewInt(100),
+
+			expectError: true,
```
czarcas7ic, line 786: Maybe not for this PR, but in general I feel like we should start moving away from bool errors and instead specify the error. Lmk what you think

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+		"attempted to distribute more than developer vesting module account balance - error": {
+			moduleAccountName: types.DeveloperVestingModuleAcctName,
+			preExistingDelta:  sdk.NewDecFromInt(sdk.NewInt(keeper.DeveloperVestingAmount)).Add(sdk.OneDec()),
+			provisions:        sdk.NewDecCoinFromDec(sdk.DefaultBondDenom, sdk.NewDecWithPrec(1006, 1)),
+			amountDistributed: sdk.NewInt(100),
+
+			expectErr: true,
+		},
+	}
+	for name, tc := range tests {
```
czarcas7ic, line 931: Same as above, I think we need a test case where we test the accumulation of the truncation delta over multiple epochs to ensure it works as expected (0.1 > 0.6 > 1.1 or something like that)

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+			suite.Require().NoError(err)
+			suite.Require().Equal(tc.expectedDistributedAmount.String(), actualAmount.String())
+			actualPersistedInstore, err := mintKeeper.GetTruncationDelta(ctx, tc.moduleAccountName)
+			suite.Require().NoError(err)
+			suite.Require().Equal(tc.expectedPersistedInStore, actualPersistedInstore)
+		})
+	}
+}
+
+func (suite *KeeperTestSuite) TestSetTruncationDelta() {
```
czarcas7ic, line 960: Am I right in saying the truncation delta persisted in store should never be greater than 1? If so, can/should we guarantee that through code/tests?

czarcas7ic (commented):
Great job on this! This was just a first pass, so once we address some of these comments I would like to do another pass so I can feel 100 percent on it

x/mint/keeper/keeper.go
alexanderbez: it pretty much means an invariant that must hold

x/mint/keeper/export_test.go
```diff
@@ -17,9 +17,14 @@ const (
 )
 
 var (
-	GetProportions = getProportions
+	GetProportions                         = getProportions
+	GetTruncationStoreKeyFromModuleAccount = getTruncationStoreKeyFromModuleAccount
 )
 
+func (k Keeper) CalculateTotalTruncationDelta(ctx sdk.Context, moduleAccountName string, provisions sdk.Dec, amountDistributed sdk.Int) (sdk.Dec, error) {
```
alexanderbez, line 24: nit: add godocs for new methods

x/mint/keeper/keeper.go
p0mvn: Good point. Keeping unresolved until I check

x/mint/keeper/keeper.go
p0mvn: What exactly? Telemetry? These are the metrics useful for debugging

This is how we pipeline debugging data to things like Prometheus and Grafana for observability

x/mint/keeper/keeper_test.go
p0mvn: The goal is to test all possible combinations a value can take, not only what we expect

p0mvn (commented):
Still working through comments

x/mint/keeper/keeper.go
czarcas7ic: Yeah that is fine, it was more me just thinking out loud haha, I'm cool with this guarantee

x/mint/keeper/keeper.go
czarcas7ic: Oof yeah I read this wrong, this is correct, my bad

x/mint/keeper/keeper.go
czarcas7ic: Ah okay, I have never seen this used before so that's why I was confused

x/mint/keeper/keeper_test.go
czarcas7ic: Should we make a comment somewhere explaining that this state is not expected to be reached however we are still testing for it? I only say this because it threw me off when I was reviewing.

Side note: Is there a reason we don't force the value to be less than 1 (i.e. if the delta were to be returned as anything less than 1 then we panic?)

p0mvn: Technically, it is still allowed to have > 1 truncation delta. For example, we might want to manually set it in the upgrade handler to be auto-distributed at the next epoch. On the other hand, we can also directly distribute it to the community pool instead too.

I will think about this suggestion.

x/mint/keeper/keeper_test.go
p0mvn: I'm not sure why we need that? What we are testing here is that the math is correct based on the input parameters and the value of the store index. I think the existing tests cover this pretty well.

Testing this over several epoch in such as way is the same as judging some of the existing test cases separately

x/mint/keeper/keeper_test.go
p0mvn: I don't understand the value of running this several times in a row in a unit test because, instead, we can make sure that a single run unit test covers all combinations of input parameters and relevant store values.

There are functional and integrations test things like this. We do have a functional test that helps to ensure that multi-epoch values are correct:
https://github.com/osmosis-labs/osmosis/blob/d7caed8acfe3b756134c42c839bfca1d6225cb1b/x/mint/keeper/hooks_test.go#L444

It's in the original large PR still

x/mint/keeper/keeper_test.go
p0mvn: Will resolve this comment because we were discussing this earlier here:
https://github.com/osmosis-labs/osmosis/pull/2503#discussion_r954465923

x/mint/keeper/keeper_test.go
czarcas7ic: I guess my concern is the store getting prematurely overridden before reaching > 1 if the store takes multiple epochs to reach this value. I felt like having the check there would be nice to have, especially if any of this logic gets touched in the future. On my second pass I will check how necessary I feel this should actually be

p0mvn: Sorry, I'm not sure I follow how this concern specifically relates to a unit test for the `calculateTotalTruncationDelta` function.

This is something that is covered by a functional test that runs the `AfterEpochEnd` hook several times and ensures that the amounts are valid:
https://github.com/osmosis-labs/osmosis/blob/d7caed8acfe3b756134c42c839bfca1d6225cb1b/x/mint/keeper/hooks_test.go#L444

czarcas7ic: Let's say there is a bug in `calculateTotalTruncationDelta` which causes the retained value to only be retained for one epoch, then, if the epoch after that still doesn't push it over the "1" delta to be distributed, it gets reset to zero. This case would not be caught by higher abstraction levels right?

This isn't a huge deal at all, I just thought it was something that should be added to these test cases but it isn't a blocker

x/mint/keeper/keeper.go
```diff
@@ -121,6 +115,33 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
...
+// Returns error if invalid module account.
+func (k Keeper) SetTruncationDelta(ctx sdk.Context, moduleAccountName string, truncationDelta sdk.Dec) error {
+	storeKey, err := getTruncationStoreKeyFromModuleAccount(moduleAccountName)
+	if err != nil {
+		return err
+	}
+
+	if truncationDelta.LT(sdk.ZeroDec()) {
+		return sdkerrors.Wrapf(types.ErrInvalidAmount, "truncation delta must be positive, was (%s)", truncationDelta)
+	}
```
czarcas7ic, line 140: nit: we should probably check trucationDelta first before fetching the store key to prevent unnecessary logic from executing.

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+	if moduleAccountName == types.ModuleName {
+		if err := k.mintCoins(ctx, sdk.NewCoins(sdk.NewCoin(provisions.Denom, truncationDeltaToDistribute))); err != nil {
+			return sdk.Int{}, err
+		}
+	}
+	if err := k.communityPoolKeeper.FundCommunityPool(ctx, sdk.NewCoins(sdk.NewCoin(provisions.Denom, truncationDeltaToDistribute)), k.accountKeeper.GetModuleAddress(moduleAccountName)); err != nil {
+		return sdk.Int{}, err
+	}
+
+	newDelta := deltaAmount.Sub(truncationDeltaToDistribute.ToDec())
```
czarcas7ic, line 403: I think this would be a good place to guarantee a delta that is less that 1 but greater than 0 right? That or within the `SetTruncationDelta` function itself.

x/mint/keeper/keeper_test.go
czarcas7ic: I think a guarantee of having a truncation delta less than 1 is the safer / more preferred method IMO. Especially since we are minting whatever this value is. Lmk what you think

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+		preExistingDelta  sdk.Dec
+		provisions        sdk.DecCoin
+		amountDistributed sdk.Int
+
+		// Expected outputs
+		expectedDistributedAmount sdk.Int
+		expectedPersistedInStore  sdk.Dec
+		expectErr                 bool
+	}{
+		"valid case: 100.6 - 100 + 0.2 = 0.8; none-distrubuted and persisted until next epoch": {
```
czarcas7ic, line 870:
```suggestion
		"valid case: 100.6 - 100 + 0.2 = 0.8; none distributed and 0.8 persisted in store until next epoch": {
```

czarcas7ic (approved):
After looking it over a second time I think this is good to go, well done! Just had a few more comments but none blocking

x/mint/keeper/keeper_test.go
p0mvn: It should be caught in the linked test, and I don't see how this could be tested here since `calculateTotalTruncationDelta` doesn't deal with any epoch logic.

x/mint/keeper/keeper.go
p0mvn: Added this point to the ADR

x/mint/keeper/keeper_test.go
czarcas7ic: That's fair! Just wanted to make sure we had test coverage somewhere for this scenario!

x/mint/keeper/genesis.go
```diff
@@ -18,6 +18,15 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
 	k.SetMinter(ctx, data.Minter)
 	k.SetParams(ctx, data.Params)
 
+	// developer vesting module account delta initialization.
+	if err := k.SetTruncationDelta(ctx, types.DeveloperVestingModuleAcctName, sdk.ZeroDec()); err != nil {
```
mattverse, line 22: do we not need to add these in export genesis?

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+
+	if err := k.SetTruncationDelta(ctx, moduleAccountName, newDelta); err != nil {
+		return sdk.Int{}, err
+	}
+
+	if truncationDeltaToDistribute.IsInt64() {
```
mattverse, line 409: Can we add a one line comment on what's happening here?

x/mint/keeper/keeper_test.go
```diff
@@ -719,3 +719,348 @@ func (suite *KeeperTestSuite) TestDistributeDeveloperRewards() {
...
+			actualAmount, err := mintKeeper.HandleTruncationDelta(ctx, tc.moduleAccountName, tc.provisions, tc.amountDistributed)
+
+			// Assertions.
+			if tc.expectErr {
+				suite.Require().Error(err)
+				return
+			}
+
+			suite.Require().NoError(err)
+			suite.Require().Equal(tc.expectedDistributedAmount.String(), actualAmount.String())
```
mattverse, line 952: Since handletruncatedelta handles more logic than calculatetruncate delta, I think it would be better if we test extra logics that are happening in handle truncate delta. Would it make sense to have extra cases for community pool funding or module set gauge?

mattverse (commented):
Looks nice! Left some comments, please take a look

osmoutils/store_helper_test.go
```diff
@@ -364,7 +364,7 @@ func (s *TestSuite) TestMustGetDec() {
 
 			expectedGetKeyValues: map[string]sdk.Dec{
 				keyA: sdk.OneDec(),
-				keyB: sdk.Dec{}, // this one panics
+				keyB: {}, // this one panics
```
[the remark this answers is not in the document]
ValarDragon: This is just syntax sugar, does the same thing

x/mint/types/keys.go
```diff
@@ -1,12 +1,35 @@
...
+
+	// TruncatedInflationDeltaKey represents a key for the the delta of minted
+	// inflation provisions that have not been distributed yet due to truncations.
+	// Truncations are stemming from the interfaces of the core modules such as
```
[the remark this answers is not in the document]
ValarDragon: Suppose mint says I should send 1.2 coins. I can only send 1 coin or 2 coins in bank, so this is tracking the rounding error I believe.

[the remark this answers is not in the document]
ValarDragon: Would be nice to clarify here if this is the case, and if so then that all values in here should be positive (or equivalently representing the decimal amount that ideally would have been distributed but is not)

x/mint/keeper/keeper.go
```diff
@@ -279,15 +300,23 @@ func (k Keeper) distributeDeveloperRewards(ctx sdk.Context, totalMintedCoin sdk.
...
+// calculateTotalTruncationDelta returns the total truncation delta that has not been distributed yet
+// for the given module account, provisions, and actual amount distributed in the current epoch.
+// The returned delta might include the delta from previous epochs if they have not been distributed due to truncations.
+// For example, assume that for some number of epochs our expected provisions
+// are 100.6 and the actual amount distributed is 100 every epoch due to truncations.
+// Then, at epoch 1, we have a delta of 0.6. 0.6 cannot be distributed because it is not an integer.
+// So we persist it until the next epoch. Then, at at epoch 2, we get a delta of 1.2 (0.6 from epoch 1 and 0.6 from epoch 2).
+// Now, 1 can be distributed and 0.2 gets persisted until the next epoch.
+// CONTRACT: provisions are greater than or equal to amountDistributed.
+// nolint: unused
```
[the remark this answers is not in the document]
ValarDragon: I imagine its leftover from some initial code I assume
```suggestion
```

x/mint/keeper/keeper.go
ValarDragon: eh, I would not be concerned about performance of this. This is not the bottleneck, its once per epoch behavior.

x/mint/keeper/keeper.go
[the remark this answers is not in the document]
p0mvn: This is a partial change. I'm splitting up into smaller chunks.

The remaining part that calls this method is here: https://github.com/osmosis-labs/osmosis/pull/2342

x/mint/keeper/keeper.go
```diff
@@ -93,21 +95,13 @@ func (k *Keeper) SetHooks(h types.MintHooks) *Keeper {
...
 // GetMinter gets the minter.
 func (k Keeper) GetMinter(ctx sdk.Context) (minter types.Minter) {
-	store := ctx.KVStore(k.storeKey)
-	b := store.Get(types.MinterKey)
-	if b == nil {
-		panic("stored minter should not have been nil")
-	}
-
-	k.cdc.MustUnmarshal(b, &minter)
+	osmoutils.MustGet(ctx.KVStore(k.storeKey), types.MinterKey, &minter)
```
ValarDragon, line 98: Happy that all of this is overhead is going down :)

x/mint/keeper/keeper.go
p0mvn: Already added in `SetTruncationDelta`

p0mvn: Nvm, that was for 0, adding now

x/mint/keeper/keeper_test.go
p0mvn: Added a check per suggestion

x/mint/keeper/keeper.go
p0mvn: added

x/mint/keeper/keeper.go
```diff
@@ -315,3 +344,96 @@ func (k Keeper) createDeveloperVestingModuleAccount(ctx sdk.Context, amount sdk.
...
+	if ratio.GT(sdk.OneDec()) {
+		return sdk.Coin{}, invalidRatioError{ratio}
+	}
+	return sdk.NewCoin(mintedCoin.Denom, mintedCoin.Amount.ToDec().Mul(ratio).TruncateInt()), nil
+}
+
+// getTruncationStoreKeyFromModuleAccount returns a truncation store key for the given module account name.
+// moduleAccountName can either be developer vesting or mint module account.
+// returns error if moduleAccountName is not either of the above.
+func getTruncationStoreKeyFromModuleAccount(moduleAccountName string) ([]byte, error) {
```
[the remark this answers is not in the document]
p0mvn: Yes, we pre-minted 225m to the developer vesting module account for compliance reasons. As a result, the distribution logic for developer vesting rewards is different from other (referred to as "inflation" in the PR) provisions.

Inflation provisions must be minted to the mint module account and then transferred over to the receiver module accounts. On the other hand, developer vesting rewards do not need to be pre-minted and just get transferred over.

Another difference is that we use supply offsets for the unvested developer rewards to make sure that they do not pre-maturely inflate the supply

x/mint/keeper/genesis.go
p0mvn: Added

Issue #2507, opened 2022-08-25
Move spot price error handling, panic catching, and rounding to keeper layer


ValarDragon:
## Background

Currently many spot price guarantees are implemented at the per-pool layer, which is the incorrect layer of abstraction. Namely spot price is now promising the following guarantees:
* Will not panic, returns an error in such situations
* Maximum returned spot price is 2**160, otherwise returns an error
* Returned spot price is rounded to types.SpotPriceSigFig units of precision

## Suggested Design

Move this safety check logic + tests into keeper layer

## Acceptance Criteria

- [ ] Tests + logic for this is in the keeper layer
- [ ] Queries use logic in the keeper layer

PR #2508, opened 2022-08-25
Move spot price safety checks to keeper layer

ValarDragon:
Closes: #2507 

- Move safety checks for spot price against internal logic issues, sig fig rounding, and overflow to the keeper layer
- Fix bug where grpc query uses pool spot price instead of keeper method
- Document these safety guarantees

~This is currently blocked on #2405 , as create pool panics right now~

## Testing and Verifying

This change should be moving + adding new test cases in total, each more modularly guaranteeing a component.

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? yes (TODO)
  - How is the feature or change documented? specification (`x/<module>/spec/`) (TODO)


app/apptesting/gamm.go
```diff
@@ -17,10 +17,20 @@ var DefaultAcctFunds sdk.Coins = sdk.NewCoins(
...
 func (s *KeeperTestHelper) PrepareBalancerPoolWithCoins(coins ...sdk.Coin) uint64 {
+	weights := make([]int64, len(coins))
+	for i := 0; i < len(coins); i++ {
+		weights[i] = 1
+	}
+	return s.PrepareBalancerPoolWithCoinsAndWeights(coins, weights)
+}
+
+// PrepareBalancerPoolWithCoins returns a balancer pool
+// consisted of given coins with equal weight.
```
nicolaslara, line 28:
```suggestion
// PrepareBalancerPoolWithCoinsAndWeights returns a balancer pool
// consisted of given coins with the specified weights.
```

x/gamm/keeper/pool_service.go
```diff
@@ -8,25 +8,52 @@ import (
...
 
 // CalculateSpotPrice returns the spot price of the quote asset in terms of the base asset,
 // using the specified pool.
-// E.g. if pool 1 traded 2 atom for 3 osmo, the quote asset was atom, and the base asset was osmo,
+// E.g. if pool 1 trades 2 atom for 3 osmo, the quote asset was atom, and the base asset was osmo,
 // this would return 1.5. (Meaning that 1 atom costs 1.5 osmo)
+//
+// This function is guaranteed to not panic, but may return an error if:
+// * An internal error within the pool occurs for calculating the spot price
+// * The returned spot price is greater
```
mattverse, line 22:
```suggestion
// * The returned spot price is greater than max spot price
```

x/gamm/pool-models/balancer/pool.go
```diff
@@ -618,25 +617,11 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (spotPric
...
-			spotPrice = sdk.Dec{}
-			err = types.ErrSpotPriceInternal
-		}
-	}()
-
 	// spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
 	// spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
 	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
 	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
-	fullRatio := supplyRatio.Mul(invWeightRatio)
```
mattverse, line 633: confirmed this logic has been correctly moved

x/twap/types/utils.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/osmosis-labs/osmosis/v11/osmoutils"
 )
 
+var MaxSpotPrice = sdk.NewDec(2).Power(128).Sub(sdk.OneDec())
```
mattverse, line 13: what's the reason we don't import it / use it directly from gamm module but declare it twice both in twap module and gamm module?

x/twap/logic.go
```diff
@@ -53,7 +53,12 @@ func getSpotPrices(ctx sdk.Context, k types.AmmInterface, poolId uint64, denom0,
 			sp1 = sdk.ZeroDec()
 		}
 	}
-	// if sp0.GT(gammtypes.MaxSpotPrice)
+	if sp0.GT(types.MaxSpotPrice) {
```
mattverse, line 56: this is just a safety check since `CalculateSpotPrice` shouldnt return values over types.MaxSpotPrice right?

mattverse (approved):
Left some questions and comments, otherwise lgtm!

x/twap/types/utils.go
ValarDragon: imports basically, introduces a dependency on its types package. (Maybe this doesn't make sense though)

I think it is ok for the two to diverge with time

x/twap/logic.go
ValarDragon: yup! In case a different spot price backend gets put in

x/twap/logic.go
```diff
@@ -53,7 +53,12 @@ func getSpotPrices(ctx sdk.Context, k types.AmmInterface, poolId uint64, denom0,
...
+		sp0 = types.MaxSpotPrice
```
p0mvn, line 57: If we go with the assumption of a possible different spot price backend, it might not return an error when reaching this case. So should we then set `latestErrTime` to the current block time here?

x/twap/logic.go
```diff
@@ -53,7 +53,12 @@ func getSpotPrices(ctx sdk.Context, k types.AmmInterface, poolId uint64, denom0,
...
+	}
+	if sp1.GT(types.MaxSpotPrice) {
+		sp1 = types.MaxSpotPrice
```
p0mvn, line 60: and here

x/gamm/keeper/pool_service.go
```diff
@@ -8,25 +8,52 @@ import (
...
 	if err != nil {
 		return sdk.Dec{}, err
 	}
 
-	return pool.SpotPrice(ctx, baseAssetDenom, quoteAssetDenom)
+	// defer to catch panics, in case something internal overflows.
+	defer func() {
+		if r := recover(); r != nil {
+			spotPrice = sdk.Dec{}
+			err = types.ErrSpotPriceInternal
```
p0mvn, line 38: Can we combine `r` with `types.ErrSpotPriceInternal` to assign to `err` so that we propagate the panic message to the caller?

x/gamm/keeper/pool_service.go
```diff
@@ -8,25 +8,52 @@ import (
...
+	spotPrice, err = pool.SpotPrice(ctx, baseAssetDenom, quoteAssetDenom)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	// if spotPrice greater than max spot price, return an error
+	if spotPrice.GT(types.MaxSpotPrice) {
+		return types.MaxSpotPrice, types.ErrSpotPriceOverflow
+	} else if !spotPrice.IsPositive() {
+		return sdk.Dec{}, types.ErrSpotPriceInternal
```
p0mvn, line 51: Can we return a more meaningful message here, indicating that the spot price was negative + its value?

x/gamm/keeper/pool_service.go
```diff
@@ -8,25 +8,52 @@ import (
...
+	}()
+
+	spotPrice, err = pool.SpotPrice(ctx, baseAssetDenom, quoteAssetDenom)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	// if spotPrice greater than max spot price, return an error
+	if spotPrice.GT(types.MaxSpotPrice) {
+		return types.MaxSpotPrice, types.ErrSpotPriceOverflow
```
p0mvn, line 49: Does it make sense to format the spot price as a string and the max spot price to have a more meaningful error for clients?

osmomath/sigfig_round.go
```diff
@@ -6,6 +6,9 @@ var pointOne = sdk.OneDec().QuoInt64(10)
 
 // SigFigRound rounds to a specified significant figure.
 func SigFigRound(d sdk.Dec, tenToSigFig sdk.Int) sdk.Dec {
+	if d.IsZero() {
+		return d
```
p0mvn, line 10: We are completely missing tests for this function. We should add them

x/gamm/keeper/pool_service.go
p0mvn, line 51: It seems that the only time we are hitting this branch during the tests is when we attempt to set up a pool. Can we have a test case directly covering this branch?

x/gamm/keeper/pool_service.go
```diff
@@ -8,25 +8,52 @@ import (
...
 	pool, err := k.GetPoolAndPoke(ctx, poolID)
 	if err != nil {
 		return sdk.Dec{}, err
 	}
 
-	return pool.SpotPrice(ctx, baseAssetDenom, quoteAssetDenom)
+	// defer to catch panics, in case something internal overflows.
+	defer func() {
+		if r := recover(); r != nil {
+			spotPrice = sdk.Dec{}
```
p0mvn, line 37: We don't have a test case covering this branch. Is it possible to add it?

x/gamm/keeper/pool_service_test.go
```diff
@@ -319,6 +319,51 @@ func (suite *KeeperTestSuite) TestPoolCreationFee() {
...
+	for name, tc := range tests {
+		suite.Run(name, func() {
+			poolId := suite.PrepareBalancerPoolWithCoinsAndWeights(tc.poolLiquidity, tc.poolWeights)
+			pool, err := suite.App.GAMMKeeper.GetPoolAndPoke(suite.Ctx, poolId)
+			suite.Require().NoError(err)
+			poolSpotPrice, poolErr := pool.SpotPrice(suite.Ctx, tc.baseAssetDenom, tc.quoteAssetDenom)
+			keeperSpotPrice, keeperErr := suite.App.GAMMKeeper.CalculateSpotPrice(suite.Ctx, poolId, tc.baseAssetDenom, tc.quoteAssetDenom)
+			if tc.overflows {
+				suite.Require().NoError(poolErr)
+				// TODO: We currently fail at internal precision calculation limits first.
```
p0mvn, line 354: I don't quite understand this TODO. Could you elaborate?

p0mvn (commented):
Great work!

Before merging, I would like to understand whether we need to update the `latestErrTime` when getting spot prices in TWAP and assuming a different backend.

Also, to discuss and plan increasing test coverage for the keeper's `CalculateSpotPrice`

x/gamm/keeper/pool_service.go
ValarDragon: I don't really get why we'd focus so much on error messages, this seems like not a big win / something that can be done later

ValarDragon: I don't really get why we'd focus so much on error messages, this seems like not a real win / something that can be done later

x/gamm/keeper/pool_service.go
ValarDragon: I don't really get why we'd focus so much on error messages, this seems like not a real win / something that can be done later

x/gamm/keeper/pool_service.go
ValarDragon: I don't really get why we'd focus so much on error messages, this seems like not a big win / something that can be done later

osmomath/sigfig_round.go
ValarDragon: Agreed, should be done in a separate PR

x/gamm/keeper/pool_service.go
ValarDragon: I don't think this matters that much to directly test, its a safety check to give a clear pre-condition for successful results.

x/gamm/keeper/pool_service.go
ValarDragon: Not easily, I tried.

x/twap/logic.go
ValarDragon: Good catch

x/gamm/keeper/pool_service_test.go
ValarDragon: ah yeah, this previously was failing at an internal error, fixing in subsequent commit

x/gamm/keeper/pool_service.go
ValarDragon: Actually I tried harder, it is!

Thanks for pointing this out :)

x/gamm/keeper/pool_service_test.go
```diff
@@ -319,6 +320,68 @@ func (suite *KeeperTestSuite) TestPoolCreationFee() {
...
+			suite.Require().NoError(err)
+			var poolSpotPrice sdk.Dec
+			var poolErr error
+			osmoassert.ConditionalPanic(suite.T(), tc.panics, func() {
+				poolSpotPrice, poolErr = pool.SpotPrice(suite.Ctx, tc.baseAssetDenom, tc.quoteAssetDenom)
+			})
+			keeperSpotPrice, keeperErr := suite.App.GAMMKeeper.CalculateSpotPrice(suite.Ctx, poolId, tc.baseAssetDenom, tc.quoteAssetDenom)
+			if tc.overflows {
+				suite.Require().NoError(poolErr)
+				// TODO: We currently fail at internal precision calculation limits first.
```
ValarDragon, line 368:
```suggestion
```

Issue #2502, opened 2022-08-24
Keeper Test Suite Init Module Account


czarcas7ic:
## Background

https://github.com/osmosis-labs/osmosis/blob/25cf6ad9a050d42a5bdcc28f46e7f8f80279c825/x/tokenfactory/keeper/genesis_test.go#L35-L37

In the keeper tests, in order to test that we are properly generating the module account, we need to first remove it.

## Suggested Design

Dev suggested the following:

Make a new suite.SetupGenesisTest() constructor that takes care of this so we don't need to worry about this with future tests.

## Acceptance Criteria

TBD

PR #2513, opened 2022-08-25
[Test] Create suite.SetupGenesisTest() that takes care of module account removal

stackman27:
Closes: #2502 

In the keeper tests, in order to test that we are properly generating the module account, we need to first remove it. However, 
Dev suggested the following: Make a new suite.SetupGenesisTest() constructor that takes care of this so we don't need to worry about this with future tests.

## Brief Changelog
n/a

## Testing and Verifying
Create new suite `TestGenesisTestSuite` for genesis_test

### DIFF
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -31,7 +31,6 @@ import (
 
 	"github.com/osmosis-labs/osmosis/v11/app"
 	"github.com/osmosis-labs/osmosis/v11/x/gamm/pool-models/balancer"
-	"github.com/osmosis-labs/osmosis/v11/x/gamm/types"
 	gammtypes "github.com/osmosis-labs/osmosis/v11/x/gamm/types"
 	lockupkeeper "github.com/osmosis-labs/osmosis/v11/x/lockup/keeper"
 	lockuptypes "github.com/osmosis-labs/osmosis/v11/x/lockup/types"
@@ -63,15 +62,17 @@ func (s *KeeperTestHelper) Setup() {
 
 	s.SetEpochStartTime()
 	s.TestAccs = CreateRandomAccounts(3)
-	s.SetupGenesisTest()
 }
 
-func (s *KeeperTestHelper) SetupGenesisTest() {
-	accountKeeper := s.App.AccountKeeper
+func (s *KeeperTestHelper) SetupTestForInitGenesis() {
+	s.App = app.Setup(true)
+	s.Ctx = s.App.BaseApp.NewContext(true, tmtypes.Header{Height: 1, ChainID: "osmosis-1", Time: time.Now().UTC()})
+	// remove module account to ensure initGenesis initializes it on its own
+	// accountKeeper := s.App.AccountKeeper
 
-	moduleAddress := accountKeeper.GetModuleAddress(types.ModuleName)
-	tokenFactoryModuleAccount := accountKeeper.GetAccount(s.Ctx, moduleAddress)
-	accountKeeper.RemoveAccount(s.Ctx, tokenFactoryModuleAccount)
+	// moduleAddress := accountKeeper.GetModuleAddress(types.ModuleName)
+	// tokenFactoryModuleAccount := accountKeeper.GetAccount(s.Ctx, moduleAddress)
+	// accountKeeper.RemoveAccount(s.Ctx, tokenFactoryModuleAccount)
 }
 
 func (s *KeeperTestHelper) SetEpochStartTime() {
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -66,13 +66,6 @@ func (s *KeeperTestHelper) Setup() {
 
 func (s *KeeperTestHelper) SetupTestForInitGenesis() {
 	s.App = app.Setup(true)
-	s.Ctx = s.App.BaseApp.NewContext(true, tmtypes.Header{Height: 1, ChainID: "osmosis-1", Time: time.Now().UTC()})
-	// remove module account to ensure initGenesis initializes it on its own
-	// accountKeeper := s.App.AccountKeeper
-
-	// moduleAddress := accountKeeper.GetModuleAddress(types.ModuleName)
-	// tokenFactoryModuleAccount := accountKeeper.GetAccount(s.Ctx, moduleAddress)
-	// accountKeeper.RemoveAccount(s.Ctx, tokenFactoryModuleAccount)
 }
 
 func (s *KeeperTestHelper) SetEpochStartTime() {
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/osmosis-labs/osmosis/v11/app"
 	"github.com/osmosis-labs/osmosis/v11/x/gamm/pool-models/balancer"
+	"github.com/osmosis-labs/osmosis/v11/x/gamm/types"
 	gammtypes "github.com/osmosis-labs/osmosis/v11/x/gamm/types"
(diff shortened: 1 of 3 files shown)


x/tokenfactory/keeper/genesis_test.go
```diff
@@ -1,14 +1,35 @@
...
+	moduleAddress := suite.App.AccountKeeper.GetModuleAddress(types.ModuleName)
+	tokenfactoryModuleAccount := suite.App.AccountKeeper.GetAccount(suite.Ctx, moduleAddress)
+	suite.App.AccountKeeper.RemoveAccount(suite.Ctx, tokenfactoryModuleAccount)
+}
+
+func TestGenesisTestSuite(t *testing.T) {
+	suite.Run(t, new(GenesisTestSuite))
+}
+
+func (suite *GenesisTestSuite) TestGenesis() {
```
czarcas7ic, line 32: I think it would be better to just define this as a part of the`KeeperTestSuite`. No need to make a whole new test suite for a single action

stackman27: Fixed!

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -25,6 +25,15 @@ func TestKeeperTestSuite(t *testing.T) {
 	suite.Run(t, new(KeeperTestSuite))
 }
 
+func (suite *KeeperTestSuite) SetupGenesisTest() {
+	suite.Setup()
+
+	// remove module account to ensure initGenesis initializes it on its own
+	moduleAddress := suite.App.AccountKeeper.GetModuleAddress(types.ModuleName)
+	tokenfactoryModuleAccount := suite.App.AccountKeeper.GetAccount(suite.Ctx, moduleAddress)
+	suite.App.AccountKeeper.RemoveAccount(suite.Ctx, tokenfactoryModuleAccount)
```
ValarDragon, line 34: Err, what I meant is that make a new thing in apptesting, that does `suite.Setup` except without calling `InitGenesis`

stackman27: wait so something like this; 

```
**apptesting.go** 

func SetupTest() {
...
s.SetupGenesisTest()
}

func (s *KeeperTestHelper) SetupGenesisTest() {
	accountKeeper := s.App.AccountKeeper

	moduleAddress := accountKeeper.GetModuleAddress(types.ModuleName)
	tokenFactoryModuleAccount := accountKeeper.GetAccount(s.Ctx, moduleAddress)
	accountKeeper.RemoveAccount(s.Ctx, tokenFactoryModuleAccount)
}
```

stackman27: Also just curious, why is `SetupGenesisTest` in Keeper_test not good?

ValarDragon: No not editing SetupTest. 

Make a new function in apptesting called `SetupTestForInitGenesis`, that does not run the call to InitChain located right here: https://github.com/osmosis-labs/osmosis/blob/5948cbc1951bb3e3c7b8e870ff75c11838b1b289/app/test_helpers.go#L26-L32

stackman27: Just pushed the update [`2bd4c23`](https://github.com/osmosis-labs/osmosis/pull/2513/commits/2bd4c2362cdd58f491ff4185172532d43d927980)

I'm pretty sure this "doesnot run the call to initChain" but the KV Store doesnot seem to be registeed 
err: `suite.go:77: test panicked: kv store with key KVStoreKey{0xc00138eb20, acc} has not been registered in stores`

stackman27: nvm i figured it out! 

should be good now [`4971d1c`](https://github.com/osmosis-labs/osmosis/pull/2513/commits/4971d1c5583b39550d7866f5b0bba7bd6288640a)

czarcas7ic (approved):
ACK

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -41,23 +40,3 @@ func (suite *KeeperTestSuite) CreateDefaultDenom() {
 	res, _ := suite.msgServer.CreateDenom(sdk.WrapSDKContext(suite.Ctx), types.NewMsgCreateDenom(suite.TestAccs[0].String(), "bitcoin"))
 	suite.defaultDenom = res.GetNewTokenDenom()
 }
-
-func (suite *KeeperTestSuite) TestCreateModuleAccount() {
```
p0mvn, line 45: Why are we deleting the test? Don't we still want to keep testing `CreateModuleAccount()`?

stackman27: oops just reverted

x/tokenfactory/keeper/genesis_test.go
```diff
@@ -31,20 +31,22 @@ func (suite *KeeperTestSuite) TestGenesis() {
 			},
 		},
 	}
+
 	app := suite.App
-	// remove module account to ensure initGenesis initializes it on its own
-	tokenfactoryModuleAccount := app.AccountKeeper.GetAccount(suite.Ctx, app.AccountKeeper.GetModuleAddress(types.ModuleName))
-	app.AccountKeeper.RemoveAccount(suite.Ctx, tokenfactoryModuleAccount)
-	suite.Ctx = app.BaseApp.NewContext(false, tmproto.Header{})
+	suite.Ctx = app.BaseApp.NewContext(true, tmproto.Header{})
```
p0mvn, line 36: Does it make sense to move this inside the helper and call the helper from here?

ValarDragon:
No, this doesn't make sense as the approach

p0mvn:
> No, this doesn't make sense as the approach

Sorry, I'm not sure I'm understanding what is the desired approach.

I thought that the goal is to make sure that the `SetupTestForInitGenesis` configures the test with empty genesis and no module account so that we can test this being done in `InitGenesis`?

Why is it not preferable to consolidate the following in the helper:
```
func (s *KeeperTestHelper) SetupTestForInitGenesis() {
	s.App = app.Setup(true)
	s.Ctx = s.App.BaseApp.NewContext(true, tmproto.Header{})
}
```

Then, in `TestGenesis`:
```
suite.SetupTestForInitGenesis()
app := suite.App
```

What am I missing?

stackman27:
> No, this doesn't make sense as the approach

Not sure what i'm missing here but as per your suggestion i believe `s.App = app.Setup(true)` doesnot run the call to initChain [here](https://github.com/osmosis-labs/osmosis/blob/5948cbc1951bb3e3c7b8e870ff75c11838b1b289/app/test_helpers.go#L26-L32), which means that `checkTx` is true

ValarDragon:
Ah apologies! I was wrong then, didn't realize setting check tx to true, also had it not run init genesis. TBH, thats an even weirder API correlation we have rn

app/apptesting/test_suite.go
```diff
@@ -64,6 +64,10 @@ func (s *KeeperTestHelper) Setup() {
 	s.TestAccs = CreateRandomAccounts(3)
 }
 
+func (s *KeeperTestHelper) SetupTestForInitGenesis() {
+	s.App = app.Setup(true)
```
ValarDragon, line 68:
```suggestion
        // Setting to True, leads to init genesis not running
	s.App = app.Setup(true)
```

ValarDragon (approved):

x/tokenfactory/keeper/genesis_test.go
stackman27: I tried that but it didn't work, i get this error `kv store with key KVStoreKey{0xc0005142c0, acc} has not been registered in stores`

stackman27: actually nvm, fixed!

p0mvn (approved):

PR #2514, opened 2022-08-26
docs: state-compatibility documentation

p0mvn:
Closes: #XXX

Add documentation for state-compatibility safety, patch and minor releases.

## Brief Changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -378,20 +378,20 @@ functioning in a deterministic way.
 
 The state-machine scope includes the following areas:
 
-- All messages supported by the chain
-
-- Transaction gas usage
-
+- All messages supported by the chain, including:
+    - Every msg's ValidateBasic method
+    - Every msg's MsgServer method
+        - Net gas usage, in all execution paths (WIP to make this not include error flows)
+        - Error result returned (TODO: Make the error code not merkelized)
+        - State changes (namely every store write)
+- AnteHandlers in "DeliverTx" mode
 - Whitelisted queries
-
 - All `BeginBlock`/`EndBlock` logic
 
 The following are **NOT** in the state-machine scope:
 
 - Events
-
 - Queries that are not whitelisted
-
 - CLI interfaces
 
 #### Validating State-Compatibility 
@@ -557,8 +557,10 @@ get successful responses while others errors, leading to state breakage.
 
 ##### Randomness
 
-Another critical property that should be avoided due to the likelihood
-of leading the nodes to result in a different state.
+Randomness cannot be used in the state machine, as the state machine definitionally must be deterministic.
+Any time you'd consider using it, instead seed a CSPRNG off of some seed.
+
+One thing to note is that in golang, iteration order over `map`s is non-deterministic, so to be deterministic you must gather the keys, and sort them all prior to iterating over all values.
 
 ##### Parallelism and Shared State
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -365,9 +365,40 @@ State-incompatibility is allowed for major upgrades because all nodes in the net
 perform it at the same time. Therefore, after the upgrade, the nodes continue
 functioning in a deterministic way.
 
-The state compatibility is ensured by taking the app hash of the state and comparing it
-to the app hash with the rest of the network. Essentially, an app hash is a hash of
-hashes of every store's Merkle root.
+#### Scope
+
+The state-machine scope includes the following areas:
+
+- All messages supported by the chain
+
+- Transaction gas usage
+
+- Whitelisted queries
+
+- All `BeginBlock`/`EndBlock` logic
+
+The following are **NOT* in the state-machine scope:
+
+- Events
+
+- Queries that are not whitelisted
+
+- CLI interfaces
+
+#### Validating State-Compatibility 
+
+Tendermint ensures state compatibility by validating a number
+of hashes that can be found here:
+https://github.com/tendermint/tendermint/blob/9f76e8da150414ce73eed2c4f248947b657c7587/proto/tendermint/types/types.proto#L70-L77
+
+Among the hashes that are commonly affected by our work and cause
+problems are the `AppHash` and `LastResultsHash`. To avoid these problems, let's now examine how these hashes work.
+
+##### App Hash
+
+Cosmos-SDK takes an app hash of the state, and propagates it to Tendermint which,
+in turn, compares it to the app hash of the rest of the network.
+An app hash is a hash of hashes of every store's Merkle root.
 
 For example, at height n, we compute:
 `app hash = hash(hash(root of x/epochs), hash(root of  x/gamm)...)`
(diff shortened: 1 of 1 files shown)


p0mvn:
Todo: mention historical queries that can be pruned

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+It is critical to avoid performing network requests since it is common
+for services to be unavailable or rate-limit. As a result, nodes
+may get diverging responses, leading to state breakage.
+
+##### Randomness
+
+Another critical property that should be avoided due to the likelihood
+of leading the nodes to result in a different state.
+
+#### Parallelism and Shared State
```
AlpinYukseloglu, line 457: Is this meant to be outside of the `Sources of State Incompatibility` section or should it have `#####`?

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+
+It is critical to avoid performing network requests since it is common
+for services to be unavailable or rate-limit. As a result, nodes
+may get diverging responses, leading to state breakage.
+
+##### Randomness
+
+Another critical property that should be avoided due to the likelihood
+of leading the nodes to result in a different state.
+
```
AlpinYukseloglu, line 456: I wonder if these two sections should be moved to a separate section on what to generally avoid during development as opposed to having them be part of a list of acceptable but state-machine breaking changes?

AlpinYukseloglu (commented):
Awesome addition, well overdue. Thanks for adding this! Left a couple comments but overall I think the main points are covered

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+Version B
+```go
+func (sk Keeper) validateAmount(ctx context.Context, amount sdk.Int) error {
+    if amount.IsNegative() || amount.IsZero() {
+        return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "amount must be positive")
+    }
+    return nil, errors.New("error")
+}
+```
```
czarcas7ic, line 406: Wouldn't this also be state incompatible for a second reason? Not only is 0 now returning a different result, but now a negative number also returns a different error response right?

czarcas7ic (approved):
LGTM! 🚀

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+  doStuff(ctx, object1, object2)
+}
+```
+- It will run out of gas with `gasUsed = 2100` where 2100 is getting merkelized
+into the tx results.
+
+Therefore, we introduced a state-incompatibility by merklezing diverging gas
+usage.
+
+##### Network Requests
```
mattverse, line 446: Can we specify what we mean by network requests here?

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+
+For example, at height n, we compute:
+`app hash = hash(hash(root of x/epochs), hash(root of  x/gamm)...)`
+
+Then, Tendermint ensures that the app hash of the local node matches the app hash
+of the network. Please note that the explanation and examples are simplified.
+
+#### Sources of State-incompatibility
+
+##### Creating Additional State
```
mattverse, line 380: We should also handle Changing Existing State(Ex changing of existing proto field)

mattverse (commented):
Looking good! This is one of the documentations that we have needed!

Would it be possible to also add docs about changes of Queries and Msgs?

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+State-incompatibility is allowed for major upgrades because all nodes in the network
+perform it at the same time. Therefore, after the upgrade, the nodes continue
+functioning in a deterministic way.
+
+The state compatibility is ensured by taking the app hash of the state and comparing it
+to the app hash with the rest of the network. Essentially, an app hash is a hash of
+hashes of every store's Merkle root.
+
+For example, at height n, we compute:
+`app hash = hash(hash(root of x/epochs), hash(root of  x/gamm)...)`
```
ValarDragon, line 373: We should also probably note the `LastResultsHash` for tx results matters as well: https://github.com/tendermint/tendermint/blob/main/proto/tendermint/types/types.proto#L70-L77

(The other hashes are unlikely to be an issue / you don't really have to separately think about)

The LastResultsHash today contains:
https://github.com/tendermint/tendermint/blob/main/types/results.go#L47-L54
* Tx GasWanted
* Tx GasUsed
* Tx response "Data"
* Tx response "Code"

The data field includes the proto message response (therefore we cannot change these in patch releases). Tbh idk what the Code field does, as far as I can tell its not even used: https://github.com/cosmos/cosmos-sdk/blob/main/baseapp/abci.go#L298-L304

The gas used field being merkelized means that:
* right now we cannot change the order of operations within a tx execution (different gas used in edge cases of out of gas'ing)
  * Therefore we must be worried about refactoring errors earlier in the tx stack
* We cannot change gas usage amounts (e.g. removing extraneous store reads)

Gas used being merkelized is slated for removal in a subsequent Tendermint release, at which point we will have more flexibility in reordering operations / erroring.

CONTRIBUTING.md
```diff
@@ -333,6 +333,137 @@ There are several steps that go into a major release
...
+into the tx results.
+
+Therefore, we introduced a state-incompatibility by merklezing diverging gas
+usage.
+
+##### Network Requests
+
+It is critical to avoid performing network requests since it is common
+for services to be unavailable or rate-limit. As a result, nodes
+may get diverging responses, leading to state breakage.
```
ValarDragon, line 450: Hrmm not sure I follow. Maybe we can clarify by stating what is the state machine scope, which is:
* All messages supported by the chain
* Queries in {whitelist file}
* All BeginBlock/Endblock logic

Importantly events are not state machine versioned

p0mvn: Will add the suggestions.

This specific part refers to querying external services, for example, some API for getting exchange rates. This is not allowed since such API might be unavailable in some regions, rate limit or just be down.

CONTRIBUTING.md
p0mvn: This specific part refers to querying external services, for example, some API for getting exchange rates. This is not allowed since such API might be unavailable in some regions, rate limit or just be down.

I haven't seen anyone doing that but I think it might make sense to mention for beginners since querying an API is a common activity in other systems

CONTRIBUTING.md
p0mvn: Agreed, I generally reshuffled things to have a "Major" and "Secondary" section. Let me know what you think

CONTRIBUTING.md
p0mvn: That's a good observation.

It seems to me that the error message does not affect things unless we store it in state for some reason which really should never happen.

However, if we were to change error code, that would cause an issue. As per Dev's comment:
https://github.com/osmosis-labs/osmosis/pull/2514#discussion_r956065300

there is a transaction response code that gets included in the `LastResultsHash`. This is what the error code here [seems to represent](https://github.com/osmosis-labs/cosmos-sdk/blob/5c9a51c277d067e0ec5cf48df30a85fae95bcd14/types/errors/abci.go#L39). So we should be really careful about error codes. I will add a section about that

CONTRIBUTING.md
p0mvn: Thanks! Great points

re: code

Apparently, it seems to be used when an error is returned. I discussed this briefly here:
https://github.com/osmosis-labs/osmosis/pull/2514#discussion_r956421045

Planning to include a few sentences on it.

p0mvn:
Pretty sure I covered all suggestions in the updates. Please let me know what you all think

CONTRIBUTING.md
AlpinYukseloglu: This version looks good!

ValarDragon (approved):
Nice work!

PR #2525, opened 2022-08-27
Minor CI tool fixes

ValarDragon:

Fixes `make test-mutation`, and ensures blocks.db from simulator is ignored regardless of directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,3 +229,4 @@ Cargo.lock
 # Ignores beaker state
 .beaker
 blocks.db
+**/blocks.db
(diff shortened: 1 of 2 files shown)


czarcas7ic:
I believe doing this leaves out some files for some reason (which is why I had to add the ! in the first place), however I am now seeing errors like you do that I wasn't seeing before. Give me a moment to look into this

czarcas7ic:
@ValarDragon  Basically the method you posted leaves out the twap folder. I have no idea why, but now am realizing that the method I committed includes things like the READMEs, so something in the substitution isnt working. I recommend that we just go back to how it was without the common variable, unless you know what the problem is off the top of your head

czarcas7ic:
Okay I think I finally figured it out, the single quotes on the file exclusions had to go -_-

Issue #2520, opened 2022-08-26
Error handling in hooks instead of panics


Anmol1696:
## Background

Epoch hooks dont need to panic since the ctx does not `write()`, can return early instead instead of panic.

## Suggested Design

Return early instead of panics
```
if err != nil {
  return
}
```

## Acceptance Criteria

- [ ] Make sure the tests pass

puneet2019:
ah, sweet #2519 already makes this very straight forward..

ValarDragon:
This is a great idea, lets do it!

puneet2019:
not really @Anmol1696 , it's do it for all, will be weird to have it included in different versions. 
I mean you can create smaller sub issues for each module hooks. but every hook should be taken care of

Anmol1696:
Yup agreed. I was planning to create smaller issues for all the other hooks as well and then tackle it. Cool let me just add the PRs to this issue itself then.

ValarDragon:
Ah my bad, I thought you were doing it for just epochs my bad. I don't think this is something we should do for all hooks, several should have the ability to halt the action. I guess we can always leave it to hook caller level to panic

puneet2019:
@ValarDragon  
 I just wanted it for epochs. But why not for all? Even returning the error halts the action. more or less same thing because we are using ApplyFuncOfNoError, just the keyword `panic` will be gone. Panic recovery is still a good to have for random store instances that might panic (this should never happen tho)
 
 ```
 func ApplyFuncIfNoError(ctx sdk.Context, f func(ctx sdk.Context) error) (err error) {
	// Add a panic safeguard
	defer func() {
		if recoveryError := recover(); recoveryError != nil {
			PrintPanicRecoveryError(ctx, recoveryError)
			err = errors.New("panic occurred during execution")
		}
	}()
	// makes a new cache context, which all state changes get wrapped inside of.
	cacheCtx, write := ctx.CacheContext()
	err = f(cacheCtx)
	if err != nil {
		ctx.Logger().Error(err.Error())
	} else {
		// no error, write the output of f
		write()
		// Temporary, should be removed once: https://github.com/cosmos/cosmos-sdk/issues/12912
		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
	}
	return err
}
 ```

Anmol1696:
[...]

PR #2519, opened 2022-08-26
Unify cache_ctx wraps

ValarDragon:
## What is the purpose of the change

In #2515 @puneet2019 fixed that epochs isn't returning events, due to https://github.com/cosmos/cosmos-sdk/issues/12912 .

This also showed that were actually duplicating panic wrapping logic. This PR adds that fix to the general osmosis helper, and makes epochs use the general osmosis helper

## Brief Changelog

- Apply events bug fix from #2515 to the general panic wrapper
- Make epochs use the general panic wrapper

## Testing and Verifying

This change is already covered by existing tests, that @puneet2019 added in #2515

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? yes
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? N/A

PR #2526, opened 2022-08-27
error returning epoch hooks

puneet2019:
Closes: #XXX
A part of #2520 only implemented for epochs module. 
If it looks good, can go ahead and implement for other modules as well. 
would wait for #2519 to be merged before. (rebase, etc. then open up)

- Hooks errors can now be handled with [`ApplyFuncIfNoError`](https://github.com/osmosis-labs/osmosis/blob/main/osmoutils/cache_ctx.go#L16) 
- This implementation allows `hooks interface` functions to return an error.
- If functions return an error any state changes will not be written to the state.
- Currently it is only done for [`EpochHooks interface`](https://github.com/osmosis-labs/osmosis/blob/main/x/epochs/types/hooks.go#L9)

## Brief Changelog

- Epoch Hooks now allow returning of error.
- All modules implementing EpochHooks also return error in place of panic.

## Testing and Verifying
- Updated and fixed test cases :::NOTE: unable to verify errorMessage.

  - Does this pull request introduce a new feature or user-facing behavior changes? - no
  - How is the feature or change documented? - not applicable

### DIFF
--- a/x/epochs/keeper/hooks.go
+++ b/x/epochs/keeper/hooks.go
@@ -5,11 +5,11 @@ import (
 )
 
 // AfterEpochEnd gets called at the end of the epoch, end of epoch is the timestamp of first block produced after epoch duration.
-func (k Keeper) AfterEpochEnd(ctx sdk.Context, identifier string, epochNumber int64) {
-	k.hooks.AfterEpochEnd(ctx, identifier, epochNumber)
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, identifier string, epochNumber int64) error {
+	return k.hooks.AfterEpochEnd(ctx, identifier, epochNumber)
 }
 
 // BeforeEpochStart new epoch is next block of epoch end block
-func (k Keeper) BeforeEpochStart(ctx sdk.Context, identifier string, epochNumber int64) {
-	k.hooks.BeforeEpochStart(ctx, identifier, epochNumber)
+func (k Keeper) BeforeEpochStart(ctx sdk.Context, identifier string, epochNumber int64) error {
+	return k.hooks.BeforeEpochStart(ctx, identifier, epochNumber)
 }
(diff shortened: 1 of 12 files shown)


x/epochs/keeper/hooks.go
```diff
@@ -5,11 +5,11 @@ import (
...
 // AfterEpochEnd gets called at the end of the epoch, end of epoch is the timestamp of first block produced after epoch duration.
-func (k Keeper) AfterEpochEnd(ctx sdk.Context, identifier string, epochNumber int64) {
-	k.hooks.AfterEpochEnd(ctx, identifier, epochNumber)
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, identifier string, epochNumber int64) error {
+	return k.hooks.AfterEpochEnd(ctx, identifier, epochNumber)
 }
 
 // BeforeEpochStart new epoch is next block of epoch end block
-func (k Keeper) BeforeEpochStart(ctx sdk.Context, identifier string, epochNumber int64) {
-	k.hooks.BeforeEpochStart(ctx, identifier, epochNumber)
```
ValarDragon, line 14: Can we ignore the error in here, instead of in abci.go?

x/mint/keeper/hooks_test.go
```diff
@@ -395,7 +395,10 @@ func (suite *KeeperTestSuite) TestAfterEpochEnd() {
 
 			osmoassert.ConditionalPanic(suite.T(), tc.expectedPanic, func() {
```
ValarDragon, line 396: We should change the variable here to `expectedError`, and then just do an if statement for checking if error vs checking no error

ValarDragon (approved):
This LGTM, thanks for doing this!

p0mvn (approved):

x/epochs/keeper/hooks.go
puneet2019: done!

x/mint/keeper/hooks_test.go
puneet2019: done!

p0mvn (approved):
All suggestions have been applied. Should be ready for merge. Thanks @puneet2019

PR #2529, opened 2022-08-27
Remove localosmosis Dockerfile

niccoloraspa:
Addresses: [#1679](https://github.com/osmosis-labs/osmosis/issues/1679) [1/2]

This PR removes the `Dockerfile` created for `localosmosis`, favoring the re-use of the main `Dockerfile`.

I have also restructured a bit the `scripts/setup.sh` script.

I will remove also the `Dockerfile` for the mainnet exported localosmosis but it will take a little bit more time as the logic is more complicated.

## Testing and Verifying

Run `localosmosis` as before:

```bash
make localnet-build
make localnet-start
# or `make localnet-startd` to run localosmosis in detached mode
make localnet-stop
```


czarcas7ic (commented):
So in my local test, there is no way to stop and start the localosmosis instance and retain the data. Is that the same for you? Imo this is a necessary capability

tests/localosmosis/docker-compose.yml
```diff
@@ -1,15 +1,25 @@
...
+    build:
+      context: ../../
+      dockerfile: Dockerfile
+      args:
+        RUNNER_IMAGE: alpine:3.16
+        GO_VERSION: 1.18
     volumes:
-      - ./.osmosisd/config:/osmosis/.osmosisd/config
-      - ./.osmosisd/data:/osmosis/.osmosisd/data
-      - ./.osmosisd/wasm:/osmosis/.osmosisd/wasm
```
ValarDragon, line 10: I think we need to bring these volumes back?

ValarDragon:
Agreed @czarcas7ic, thats def a necessary capability

niccoloraspa:
Okay! I thought It was not a requirement.
I'll revert the changes on the volumes

tests/localosmosis/docker-compose.yml
```diff
@@ -1,15 +1,25 @@
...
+      - ./scripts/setup.sh:/osmosis/setup.sh
+    entrypoint:
+      - /osmosis/setup.sh
     ports:
-      - "26657:26657"
-      - "1317:1317"
+      - 26657:26657
+      - 1317:1317
       - "9090:9090"
       - "9091:9091"
```
ValarDragon, line 21: A bit confused why lines 18 & 19 have quotes, but lines 20 & 21 don't?

niccoloraspa: I think it's due to copy and paste from another `docker-compose.yaml`.
Semantically they make no difference, but I'll use the same format for consistency.

czarcas7ic (approved):
Tested locally, good to go! 😄

Issue #2425, opened 2022-08-16
simulator: missing hash field for RequestBeginBlock


czarcas7ic:
## Background

When we generate a BeginBlock in the simulator, we leave off the Hash portion from the RequestBeginBlock type:
https://github.com/osmosis-labs/osmosis/blob/4176b287d48338870bfda3029bfa20a6e45ac126/simulation/executor/mock_tendermint.go#L140-L145

Then, even if we did provide this hash, we don't update it when we prepare the next simulator state here:
https://github.com/osmosis-labs/osmosis/blob/4176b287d48338870bfda3029bfa20a6e45ac126/simulation/executor/simulate_dev.go#L140-L160

## Suggested Design

We might have to implement tendermint type Block where we can then call the hash of the generated block (this may or may not be a really involved process). Once this is figured out, we then just have to amend the Hash to the RequestBeginBlock above and manually update the header in the next simulator state.

## Acceptance Criteria

At a minimum the app hash in retrievable for each height

PR #2530, opened 2022-08-28
simulator: app hash MVP (part 1/2)

czarcas7ic:
Closes: #2425

Within the simulator, we want to generate an app hash which we can use to compare across runs. This ends up not being a straight forward task since we implement a bare bones mock tendermint. 

This PR introduces the absolute minimum we need to produce an app hash. This app hash is not entirely useful, as it is essentially only derived from the validator hash within the block header. If the direction of this PR is correct, the next step will be to generate a datahash via all tx results, which should then produce an apphash that has meaning (this is now completed in #2531 )

As a side note, this is my first time really digging into tendermint, so my understanding could be (and is likely) incomplete. Please lmk if anything in the PR seems to not makes sense!

## Brief Changelog

- changes randomProposer to return a pubkey instead of a derived  hexbytes address
- adds apphash to the sql table
- adds version block 11 to the initial header (hashes wont get generated without this)
- derives initial apphash from the abci.ResponseInitChain as we would do on a real chain
- takes the current validator set and moves it to a validatorSet tendermint struct (this is needed to generate the validator hash, which cannot be nil otherwise you cannot generate an app hash)
- sets the validator hash to the block header
- generates an apphash from this header

## Testing and Verifying

### DIFF
--- a/simulation/executor/simulate.go
+++ b/simulation/executor/simulate.go
@@ -76,7 +76,7 @@ func SimulateFromSeed(
 		defer db.Close()
 		sts := `
 		DROP TABLE IF EXISTS blocks;
-		CREATE TABLE blocks (id INTEGER PRIMARY KEY, height INT,module TEXT, name TEXT, comment TEXT, passed BOOL, gasWanted INT, gasUsed INT, msg STRING);
+		CREATE TABLE blocks (id INTEGER PRIMARY KEY, height INT,module TEXT, name TEXT, comment TEXT, passed BOOL, gasWanted INT, gasUsed INT, msg STRING, appHash STRING);
 		`
 		_, err := db.Exec(sts)
 
@@ -148,7 +148,7 @@ func cursedInitializationLogic(
 		return nil, nil, simParams, fmt.Errorf("must have greater than zero genesis accounts")
 	}
 
-	validators, genesisTimestamp, accs := initChain(r, simParams, accs, app, initFunctions.AppInitialStateFn, config)
+	validators, genesisTimestamp, accs, res := initChain(r, simParams, accs, app, initFunctions.AppInitialStateFn, config)
 
 	fmt.Printf(
 		"Starting the simulation from time %v (unixtime %v)\n",
@@ -161,8 +161,12 @@ func cursedInitializationLogic(
 		ChainID:         config.ChainID,
 		Height:          int64(config.InitialBlockHeight),
 		Time:            genesisTimestamp,
-		ProposerAddress: validators.randomProposer(r),
+		ProposerAddress: validators.randomProposer(r).Address(),
+		AppHash:         res.AppHash,
 	}
+
+	// must set version in order to generate hashes
+	initialHeader.Version.Block = 11
 
 	simState := newSimulatorState(simParams, initialHeader, tb, w, validators).WithLogParam(config.Lean)
 
@@ -179,7 +183,7 @@ func initChain(
 	app simtypes.App,
 	appStateFn AppStateFn,
 	config *Config,
-) (mockValidators, time.Time, []simulation.Account) {
+) (mockValidators, time.Time, []simulation.Account, abci.ResponseInitChain) {
 	// TODO: Cleanup the whole config dependency with appStateFn
 	appState, accounts, chainID, genesisTimestamp := appStateFn(r, accounts, *config)
 	consensusParams := randomConsensusParams(r, appState, app.AppCodec())
@@ -200,7 +204,7 @@ func initChain(
 		config.InitialBlockHeight = 1
 	}
 
-	return validators, genesisTimestamp, accounts
+	return validators, genesisTimestamp, accounts, res
 }
 
 //nolint:deadcode,unused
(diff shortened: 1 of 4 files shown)


simulation/executor/simulate.go
```diff
@@ -161,9 +161,13 @@ func cursedInitializationLogic(
 		ChainID:         config.ChainID,
 		Height:          int64(config.InitialBlockHeight),
 		Time:            genesisTimestamp,
-		ProposerAddress: validators.randomProposer(r),
+		ProposerAddress: validators.randomProposer(r).Address(),
+		AppHash:         res.AppHash,
 	}
 
+	// must set version in order to generate hashes
+	initialHeader.Version.Block = 11
```
mattverse, line 169: Do we need to change this hard coded value for every upgrade happening? Or is this hard coded versioning independent?

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
+	// set header time
 	timeDiff := maxTimePerBlock - minTimePerBlock
 	simState.header.Time = simState.header.Time.Add(
 		time.Duration(minTimePerBlock) * time.Second)
 	simState.header.Time = simState.header.Time.Add(
 		time.Duration(int64(simCtx.GetRand().Intn(int(timeDiff)))) * time.Second)
 
 	// Draw the block proposer from proposers for n+1
-	simState.header.ProposerAddress = simState.nextValidators.randomProposer(simCtx.GetRand())
+	proposerPk := simState.nextValidators.randomProposer(simCtx.GetRand())
```
mattverse, line 158:
```suggestion
	proposerPubKey := simState.nextValidators.randomProposer(simCtx.GetRand())
```

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
+	simState.header.ProposerAddress = proposerPk.Address()
```
mattverse, line 159:
```suggestion
	simState.header.ProposerAddress = proposerPubKey.Address()
```

mattverse (approved):
utACK

simulation/executor/simulate.go
ValarDragon: No, this is the block version. For our usecase, we can literally just set it to {anything}. 

Tendermint very infrequently updates this.

It isn't really related to our Osmosis app version.

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
 
 	// now set variables in perspective of block n+1
 	simState.curValidators = simState.nextValidators
 	simState.nextValidators = nPlus2Validators
+
+	var currentValSet tmproto.ValidatorSet
+	// iterate through current validators and add them to the TM ValidatorSet struct
+	for _, key := range simState.curValidators.getKeys() {
+		var validator tmproto.Validator
+		mVal := simState.curValidators[key]
```
p0mvn, line 171: Can we give a more descriptive name to `mVal`?

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
+		validator.PubKey = mVal.val.PubKey
+		pk2, _ := cryptoenc.PubKeyFromProto(mVal.val.PubKey)
```
p0mvn, line 173:
```suggestion
		currentPubKey, _ := cryptoenc.PubKeyFromProto(mVal.val.PubKey)
```

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
+		validator.Address = pk2.Address()
+		currentValSet.Validators = append(currentValSet.Validators, &validator)
+	}
+
+	// set the proposer chosen earlier as the validator set block proposer
+	var newVal tmtypes.Validator
```
p0mvn, line 179:
```suggestion
	var proposerVal tmtypes.Validator
```

simulation/executor/simulate_dev.go
p0mvn, line 173: Not sure if the second return is an error. If yes, should we handle it?

simulation/executor/simulate.go
ValarDragon: oh wow, didn't realize the block version just happens to 11 in Tendermint, like Osmosis right now: https://github.com/tendermint/tendermint/blob/main/version/version.go#L24

simulation/executor/simulate_dev.go
```diff
@@ -142,20 +144,65 @@ func (simState *simState) prepareNextSimState(simCtx *simtypes.SimCtx, req abci.
...
+	proposerPk := simState.nextValidators.randomProposer(simCtx.GetRand())
+	simState.header.ProposerAddress = proposerPk.Address()
 	// find N + 2 valset
 	nPlus2Validators := updateValidators(simState.tb, simCtx.GetRand(), simState.simParams, simState.nextValidators, res.ValidatorUpdates, simState.eventStats.Tally)
 
 	// now set variables in perspective of block n+1
 	simState.curValidators = simState.nextValidators
 	simState.nextValidators = nPlus2Validators
+
+	var currentValSet tmproto.ValidatorSet
```
ValarDragon, line 167: Thoughts on moving this line and onward into a new function called `simstate.constructHeaderHashes(args)`?

ValarDragon (approved):
Nice work! This LGTM, beyond a desire to move the new code in `simulate_dev` into its own function

simulation/executor/simulate_dev.go
ValarDragon: Also perhaps we should link to some docs for the TM header?

simulation/executor/simulate.go
czarcas7ic: Yeah basically when I was trying to figure out the MVP, the block header version was one of the items that has to be set in order to generate a hash. The error message said it was expecting 11, so I simply set it to 11 haha. I guess this will just have to get manually changed whenever TM changes this.

simulation/executor/simulate_dev.go
czarcas7ic: Done on both! https://github.com/osmosis-labs/osmosis/pull/2530/commits/6a2cfed59221483b991093aa8586d2c0296949b4

czarcas7ic:
Thanks everyone for such quick reviews! Will likely merge soon after CI passes to work on part 2

PR #2531, opened 2022-08-28
simulator: data hash implementation (part 2/2)

czarcas7ic:

In PR #2530, we introduced a MVP which introduces a validator hash in order for us to generate an app hash. This app hash however is not very useful, because the only real information it is hashing is the current validator set.

This PR extracts the data byte results from each tx, hashes them, and adds this to the dataHash header field. Now, when we calculate the app hash, this actually provides us useful data which we can compare across runs!

NOTE: Only osmosis messages are supported. If an osmosis sim message uses the old legacy message framework (weightedMessage, see simulator and superfluid modules), they will not be supported since we decided to not make any changes to the sdk in regards to this. 

## Brief Changelog

- adds resData to the sql database (this isn't the actual data hash, its the byte data of the tx response, which is later used to calculate the block data hash)
- All resData for a block is now saved within the simState struct as a Data type. At the end of the block, this Data type is hashed, this hash is added to the block header, and that block header is used in part to calculate the appHash

## Testing and Verifying


simulation/simtypes/action.go
```diff
@@ -22,7 +22,7 @@ type Action interface {
 	// but this should not be the default.
 	Frequency() Frequency
 	Execute(*SimCtx, sdk.Context) (
-		OperationMsg simulation.OperationMsg, futureOps []simulation.FutureOperation, err error)
```
ValarDragon, line 25: Lets add a comment specifying that result data is data to be merkelized / we can compare consistency of across runs

ValarDragon (approved):
Nice job!

p0mvn (approved):

Issue #2412, opened 2022-08-15
TWAP: Overflow


czarcas7ic:
In TWAP we need to handle overflow of accumulator values, and a strategy for dealing with large spot prices. This issue details (1) how we handle large spot prices, and (2) options for handling overflow that we should decide between.

## How to handle large spot prices

Currently the `GetSpotPrice` function will panic if the spot price is too large or any intermediate term hits some upperbound. Investigation on the upperbound frontier is happening within #2207 .
The guarantees we should get from this PR are:
- GetSpotPrice should never panic (panic's caught)
- Return some "errorOverflow" type and "errorUnderflow" type, depending on internal value
- All pools always support results up to spot prices of some max bound, e.g. 2^128
  - May not be possible for weighted balancer pools right away

ValarDragon:
@nicolaslara and I went through this in some more detail last week. 

Max Accumulator value today is `2**256`

We can upperbound the value in it as `MaxPoolSpotPrice * {time interval}`

MaxPoolSpotPrice is set at `2^119` in https://github.com/osmosis-labs/osmosis/pull/2508/files#diff-d77d97fd9b8bc65f96ceb1cb6a6e8cbbe208b5838691047f8097f3a8560272abR28, but we can change this to be more convenient. For right now, lets abstract as `2^160`.

Then if time interval is measured in nanoseconds, the max time interval would be `2**256 / (2**160) = 2**96 ns = 2.5 * 10^12 years = 2.5 Gyear`

Hence we should have no concern of overflow.

czarcas7ic:
Create a test case that tests out of bounds time, then issue can be closed from there

PR #2536, opened 2022-08-29
Twap: Logic Test

mattverse:
Most of the methods in `logic_test.go` were lacking appropriate unit tests.
This PR adds 1:1 unit tests to the methods existing in `logic.go` of the twap package

## Brief Changelog
- Add unit tests for `logic.go`

## Testing and Verifying

This change added new tests.

  - Does this pull request introduce a new feature or user-facing behavior changes?  no


x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+	// prepare pool before test
+	poolId := s.PrepareBalancerPoolWithCoins(defaultTwoAssetCoins...)
+
+	tests := map[string]struct {
+		poolId        uint64
+		denom0        string
+		denom1        string
+		expectedErr   bool
+		expectedPanic bool
+	}{
```
czarcas7ic, line 56: wdyt about changing these expectedErrs from bool to error? That way we can ensure we are receiving the error we are expecting. Panic makes sense as a bool tho imo.

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			false,
+			false,
+		},
+		"denom with non-lexicographical order": {
+			poolId,
+			denom0,
+			denom1,
+			false,
+			false,
+		},
```
czarcas7ic, line 70: These are backwards. The first one is in non lexicographical order and the second one is in lexicographical order.

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			updateBlockHeight: false,
+			updateSpotPrice:   false,
+			expRecord:         newExpRecord(OneSec.MulInt64(10), OneSec.QuoInt64(10)),
+		},
+		"update block height": {
+			inputRecord:       newRecord(1, s.Ctx.BlockTime(), sdk.NewDec(10), zeroDec, zeroDec),
+			updateBlockHeight: true,
+			expRecord:         newExpRecord(OneSec.MulInt64(10), OneSec.QuoInt64(10)),
+		},
+		"update block time": {
```
czarcas7ic, line 132:
```suggestion
		"update block time and spot price": {
```

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			inputRecord:     newRecord(1, s.Ctx.BlockTime(), sdk.NewDec(10), zeroDec, zeroDec),
+			updateSpotPrice: true,
+			updateBlockTime: true,
 			expRecord:       newExpRecord(OneSec.MulInt64(10), OneSec.QuoInt64(10)),
 		},
+		"spot price changed": {
+			inputRecord:     newRecord(1, s.Ctx.BlockTime(), sdk.NewDec(10), zeroDec, zeroDec),
+			updateSpotPrice: true,
+			expRecord:       newExpRecord(OneSec.MulInt64(10), OneSec.QuoInt64(10)),
```
czarcas7ic, line 141: it seems like we aren't using the expRecord in the below tests at all

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+	recordsToPreSet := []types.TwapRecord{tPlus1Record, tMin1Record, baseRecord, tMin2Record}
+
+	expectedKeptRecords := []types.TwapRecord{baseRecord, tPlus1Record}
+	s.SetupTest()
+	s.preSetRecords(recordsToPreSet)
+
+	ctx := s.Ctx
+	twapKeeper := s.twapkeeper
+
+	ctx = ctx.WithBlockTime(baseTime)
```
czarcas7ic, line 200:
```suggestion
	ctx := s.Ctx.WithBlockTime(baseTime)
	twapKeeper := s.twapkeeper

```
nit

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
-			expRecord:       newExpRecord(oneDec.Add(OneSec.MulInt64(10)), twoDec.Add(OneSec.QuoInt64(10))),
+			record:    newRecord(poolId, time.Unix(1, 0), sdk.NewDec(10), oneDec, twoDec),
+			newTime:   time.Unix(2, 0),
+			expRecord: newExpRecord(oneDec.Add(OneSec.MulInt64(10)), twoDec.Add(OneSec.QuoInt64(10))),
 		},
 		"larger time interval": {
-			record:          newRecord(time.Unix(11, 0), sdk.NewDec(10), oneDec, twoDec),
-			interpolateTime: time.Unix(55, 0),
-			expRecord:       newExpRecord(oneDec.Add(OneSec.MulInt64(44*10)), twoDec.Add(OneSec.MulInt64(44).QuoInt64(10))),
+			record:    newRecord(poolId, time.Unix(11, 0), sdk.NewDec(10), oneDec, twoDec),
```
czarcas7ic, line 226: nit: couldn't we keep this record the same as all the others, then make a `defaultRecord` at the top just to clean things up a bit?

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
-		"same time": {
-			record:          newRecord(time.Unix(1, 0), sdk.NewDec(10), oneDec, twoDec),
-			interpolateTime: time.Unix(1, 0),
-			expRecord:       newExpRecord(oneDec, twoDec),
+		"same time, accumulator should not change": {
+			record:    newRecord(poolId, time.Unix(1, 0), sdk.NewDec(10), oneDec, twoDec),
+			newTime:   time.Unix(1, 0),
+			expRecord: newExpRecord(oneDec, twoDec),
 		},
 		// TODO: Overflow tests
```
czarcas7ic, line 235: Is there an issue for this (i think there is)? If so, can we link to it here?

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
 	}
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
 			// correct expected record based off copy/paste values
-			test.expRecord.Time = test.interpolateTime
+			test.expRecord.Time = test.newTime
+			test.expRecord.PoolId = test.record.PoolId
 			test.expRecord.P0LastSpotPrice = test.record.P0LastSpotPrice
 			test.expRecord.P1LastSpotPrice = test.record.P1LastSpotPrice
```
czarcas7ic, line 244: I'm confused, why aren't these just set in the expRecord table tests? Seems odd to manually set them here based on values that existed pre test run.

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
 		})
 	}
 }
 
-func (s *TestSuite) TestUpdateTwap() {
-	poolId := s.PrepareBalancerPoolWithCoins(defaultTwoAssetCoins...)
-	newSp := sdk.OneDec()
+func (s *TestSuite) TestGetInterpolatedRecord() {
+	baseRecord := newTwapRecordWithDefaults(baseTime, sdk.OneDec(), sdk.OneDec(), sdk.OneDec())
+	// tMin2Record, tMin1Record, baseRecord, tPlus1Record := s.createTestRecordsFromTime(baseTime)
```
czarcas7ic, line 254:
```suggestion
```
dead code?

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
-		record     types.TwapRecord
-		updateTime time.Time
-		expRecord  types.TwapRecord
+		recordsToPreSet     types.TwapRecord
+		testPoolId          uint64
+		testDenom0          string
+		testDenom1          string
+		testTime            time.Time
+		expectedAccumulator sdk.Dec
+		expectedErr         bool
```
czarcas7ic, line 263: not blocking, but same here about probably wanting specific errors rather than bool

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			testTime:        baseTime,
+		},
+		"test non lexicographical order parameter": {
+			recordsToPreSet: baseRecord,
+			testPoolId:      baseRecord.PoolId + 1,
+			testDenom0:      baseRecord.Asset1Denom,
+			testDenom1:      baseRecord.Asset0Denom,
+			testTime:        baseTime,
+			expectedErr:     true,
 		},
```
czarcas7ic, line 303: Here is an example of why expectedErrors as boolean is a bad idea. This is actually erroring because the testPoolId was increased to a non existent pool, not because the denoms were in non lexicographical order (getInterpolatedRecord calls LexicographicalOrderDenoms which orders the denoms for us)

czarcas7ic (changes requested):
Good job! Just a few things I caught, but after these are addressed should be good to go!

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			if test.expectedErr {
+				s.Require().Error(err)
+				return
+			}
+			s.Require().NoError(err)
+
+			if test.testTime.Equal(baseTime) {
+				s.Require().Equal(test.recordsToPreSet, interpolatedRecord)
+			} else {
+				s.Require().Equal(test.testTime, interpolatedRecord.Time)
```
stackman27, line 321: Just wondering why not assert the entire twapRecord instead of individual values?

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+
+	ctx = ctx.WithBlockTime(baseTime)
+
+	err := twapKeeper.PruneRecords(ctx)
+	s.Require().NoError(err)
+
+	s.validateExpectedRecords(expectedKeptRecords)
+}
+
+func TestRecordWithUpdatedAccumulators(t *testing.T) {
```
stackman27, line 208: Is there a reason this test isn't part of the testSuite?

x/twap/logic_test.go
```diff
@@ -42,76 +43,287 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			poolId + 1,
+			denom1,
+			denom0,
+			false,
+			true,
+		},
+	}
+	for name, test := range tests {
+		s.Run(name, func() {
+			if test.expectedPanic {
```
stackman27, line 88: We can replace this logic with `ConditionalPanic` https://github.com/osmosis-labs/osmosis/blob/main/app/apptesting/osmoassert/assertions.go#L14

x/twap/logic_test.go
mattverse: sgtm! Changed

x/twap/logic_test.go
mattverse: yeah the original intention was to use it at some point, but changed to testing it using other states for check! Deleted this field

x/twap/logic_test.go
mattverse: good catch! Fixed ;)

x/twap/logic_test.go
mattverse: This test case isn't using any methods from test suite, there were no reason it should be a part of test suite. Do you think it should go under test suite? If so, why do you think so?

x/twap/logic_test.go
mattverse: don't know if it's gonna have significant clean up since they all do look similar but are different records except two. 
But I still do think it makes sense, fixed it!

x/twap/logic_test.go
mattverse: There are comments all over twap module as of right now considering over flow.
https://github.com/osmosis-labs/osmosis/issues/2412

x/twap/logic_test.go
mattverse: dead code@

x/twap/logic_test.go
mattverse: I was confused after I first saw this. It becomes clearer when you see how the function is defined https://github.com/osmosis-labs/osmosis/blob/main/x/twap/logic_test.go#L35. 

We're only setting denom and accumulator in within the `newExpRecord` function, thus need to initialzie here

x/twap/logic_test.go
mattverse: Good catch! Agreeing with you, we should also walk towards changing expectedErr across our repo into specific errors.

x/twap/logic_test.go
mattverse: resolved!

x/twap/logic_test.go
mattverse: good catch on this, but I personally am not in favor of using conditional Panic for test cases simple like this, I think `conditionalPanic` is worth using when we have more complexity over testing. LMk what you think

x/twap/logic_test.go
mattverse: Good question, we don't want to test the whole field of twapRecord here(some aren't initialized in accordance to the testing env, nor do we need to initialize them to test them here). Thats why we test and check only the fields that needs testing here!

x/twap/logic_test.go
czarcas7ic: oof I didn't see the difference until a second look, you were right as it was, my bad!

x/twap/logic_test.go
```diff
@@ -43,130 +43,278 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
+			nil,
+			false,
+		},
+		"denom with lexicographical order": {
+			poolId,
+			denom1,
+			denom0,
+			nil,
+			false,
+		},
```
czarcas7ic, line 70: This is still backwards, non-lexicographical order is denom1, denom0. Lexicographical order is denom0, denom1.

x/twap/logic_test.go
```diff
@@ -43,130 +43,278 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
-			spotPriceResult0: spotPriceResOne,
-			spotPriceResult1: spotPriceResOne,
-			expRecord:        withLastErrTime(sp10OneTimeUnitAccumRecord, baseTimeMinusOne),
+		"on lexicographical order denom parameters": {
+			recordsToPreSet: baseRecord,
+			testPoolId:      baseRecord.PoolId,
+			testDenom0:      baseRecord.Asset1Denom,
+			testDenom1:      baseRecord.Asset0Denom,
+			testTime:        baseTime,
 		},
```
czarcas7ic, line 286: This one should be testDenom0 matches baseRecord.Asset0Denom, then same for testDenom1

czarcas7ic (approved):
Just a couple small issues, but overall good to go after those are addressed. Good work! 🚀

x/twap/logic_test.go
stackman27: aah gotchu

x/twap/logic_test.go
stackman27: hmm, i was thinking more from code clarity + reusability of defined methods perspective. 

Also, i'm wondering should we only use `ConditionalPanic` for just complex panic testing? It's such a simple call that performs the exact same logic. I thought it was meant for simple testing like this

x/twap/logic_test.go
stackman27: Yup, you're right, i do not have strong preference here. Donot want to force it anything, if its not being used.

x/twap/logic_test.go
mattverse: must have missed it! Thanks for double checking!

x/twap/logic_test.go
```diff
@@ -43,129 +43,254 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
 	}
 }
 
+func (s *TestSuite) TestNewTwapRecord() {
```
ValarDragon, line 46: We should add a TODO for testing this with spot prices right?

(Not v12 blocking)

x/twap/logic_test.go
ValarDragon: Things that don't use keeper methods shouldn't use the suite imo

x/twap/logic_test.go
```diff
@@ -43,129 +43,254 @@ func newExpRecord(accum0, accum1 sdk.Dec) types.TwapRecord {
...
 	}
 }
 
-func (s *TestSuite) TestUpdateTwap() {
-	poolId := s.PrepareBalancerPoolWithCoins(defaultTwoAssetCoins...)
-	programmableAmmInterface := twapmock.NewProgrammedAmmInterface(s.App.TwapKeeper.GetAmmInterface())
-	s.App.TwapKeeper.SetAmmInterface(programmableAmmInterface)
-
-	spotPriceResOne := twapmock.SpotPriceResult{Sp: sdk.OneDec(), Err: nil}
-	spotPriceResOneErr := twapmock.SpotPriceResult{Sp: sdk.OneDec(), Err: errors.New("dummy err")}
```
ValarDragon, line 93: I don't understand this test change, I think what was here with the programmable interface is better for more accurately testing what we want, than the `TestUpdateRecord` change in this PR

I feel like we should switch the TestUpdateRecord back to this approach?

ValarDragon (changes requested):

ValarDragon (approved):

ValarDragon:
awesome ty!

Issue #2518, opened 2022-08-26
TWAP: Ensure that pruning does not prune the last record


ValarDragon:
## Background

With TWAP Record pruning, we should adjust the implementation, so that it preserves one record past the pruning period.
The reason for this can be seen below.
* Suppose pruning param -48 hour
* Suppose swaps at -50 hour, -1hour
* A prune would leave us with only record at -1 hour, and we would not be able to get twaps from the [-48 hour, -1 hour] time range.

## Suggested Design / Acceptance Criteria

- [ ] TWAP pruning will preserve the newest record past -48 hour, if any records past -48 hours exist.
- [ ] Testing this

p0mvn:
Something I've been thinking about:

Let's assume that the pruning param is 48 hours like in the example.

Suppose we implemented a change like suggested. Now, we have a record at - 50 hours and -2 hours.

Technically, the API now allows getting records before parameterized -48 hours. We can also have -49 or -50 since the earliest available record is at -50.

It seems that now we cannot define the earliest available twap deterministically. The newest record before -48 hours could be at any time.

I think it might make sense to add the following check:
```
func (k Keeper) GetArithmeticTwapToNow(
	ctx sdk.Context,
	poolId uint64,
	baseAssetDenom string,
	quoteAssetDenom string,
	startTime time.Time) (sdk.Dec, error) {
	time := ctx.BlockTime()
	keepPeriod := k.GetParams(ctx).RecordHistoryKeepPeriod
	if startTime.Before(time.Add(-keepPeriod)) {
		return sdk.Dec{}, fmt.Errorf("called GetArithmeticTwapToNow with a start time (%d) that is earlier than"+
			"current block time (%d) - keep period (%d)", startTime, time, keepPeriod)
	}

```

That would limit the earliest available twap to the `block time - keep period` despite the possibility of having an earlier record.

ValarDragon:
Why is it an issue to serve an earlier record if we have it, as long as API callers know that its not guaranteed to exist?

p0mvn:
There isn't a specific issue preventing us from doing that. It just felt odd having the variability in the threshold of when twaps can be available. As a result, I was thinking of providing more specific guarantees. 

If you think this is fine, I will implement it in a way that does not guarantee that a record exists before the keep threshold. However, if it does, the twap before the keep threshold is valid.

PR #2543, opened 2022-08-29
feat(twap): pruning keep the newest record under keep threshold

p0mvn:
Closes: #2518

Modify TWAP pruning to keep the newest record younger than block time - keep period.

Please see #2518 for details.

Additionally, it switched assertions in `api_test.go` from asserting on error strings to actual errors. This required defining several custom errors.

## Brief Changelog

- Modify pruning logic
- Modify existing table-driven tests to account for the change
- Add more tests for getting TWAP from times earlier than block time - keep period
- Modify hook test

## Testing and Verifying

This change added tests mentioned in changelog

### DIFF
--- a/x/twap/api.go
+++ b/x/twap/api.go
@@ -9,22 +9,22 @@ import (
 	"github.com/osmosis-labs/osmosis/v11/x/twap/types"
 )
 
-type endTimeInFutureErr struct {
+type endTimeInFutureError struct {
 	EndTime   time.Time
 	BlockTime time.Time
 }
 
-func (e endTimeInFutureErr) Error() string {
+func (e endTimeInFutureError) Error() string {
 	return fmt.Sprintf("called GetArithmeticTwap with an end time in the future."+
 		" (end time %s, current time %s)", e.EndTime, e.BlockTime)
 }
 
-type startTimeAfterEndTimeErr struct {
+type startTimeAfterEndTimeError struct {
 	StartTime time.Time
 	EndTime   time.Time
 }
 
-func (e startTimeAfterEndTimeErr) Error() string {
+func (e startTimeAfterEndTimeError) Error() string {
 	return fmt.Sprintf("called GetArithmeticTwap with a start time that is after the end time."+
 		" (start time %s, end time %s)", e.StartTime, e.EndTime)
 }
@@ -57,12 +57,12 @@ func (k Keeper) GetArithmeticTwap(
 	startTime time.Time,
 	endTime time.Time) (sdk.Dec, error) {
 	if startTime.After(endTime) {
-		return sdk.Dec{}, startTimeAfterEndTimeErr{startTime, endTime}
+		return sdk.Dec{}, startTimeAfterEndTimeError{startTime, endTime}
 	}
 	if endTime.Equal(ctx.BlockTime()) {
 		return k.GetArithmeticTwapToNow(ctx, poolId, baseAssetDenom, quoteAssetDenom, startTime)
 	} else if endTime.After(ctx.BlockTime()) {
-		return sdk.Dec{}, endTimeInFutureErr{endTime, ctx.BlockTime()}
+		return sdk.Dec{}, endTimeInFutureError{endTime, ctx.BlockTime()}
 	}
 	startRecord, err := k.getInterpolatedRecord(ctx, poolId, startTime, baseAssetDenom, quoteAssetDenom)
 	if err != nil {
--- a/x/twap/api.go
+++ b/x/twap/api.go
@@ -1,33 +1,12 @@
 package twap
 
 import (
-	"fmt"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/osmosis-labs/osmosis/v11/x/twap/types"
 )
-
-type endTimeInFutureError struct {
-	EndTime   time.Time
-	BlockTime time.Time
-}
-
-func (e endTimeInFutureError) Error() string {
-	return fmt.Sprintf("called GetArithmeticTwap with an end time in the future."+
-		" (end time %s, current time %s)", e.EndTime, e.BlockTime)
-}
-
-type startTimeAfterEndTimeError struct {
-	StartTime time.Time
-	EndTime   time.Time
-}
-
-func (e startTimeAfterEndTimeError) Error() string {
-	return fmt.Sprintf("called GetArithmeticTwap with a start time that is after the end time."+
-		" (start time %s, end time %s)", e.StartTime, e.EndTime)
-}
 
 // GetArithmeticTwap returns an arithmetic time weighted average price.
 // The returned twap is the time weighted average price (TWAP) of:
@@ -57,12 +36,12 @@ func (k Keeper) GetArithmeticTwap(
 	startTime time.Time,
 	endTime time.Time) (sdk.Dec, error) {
 	if startTime.After(endTime) {
-		return sdk.Dec{}, startTimeAfterEndTimeError{startTime, endTime}
+		return sdk.Dec{}, types.StartTimeAfterEndTimeError{StartTime: startTime, EndTime: endTime}
 	}
 	if endTime.Equal(ctx.BlockTime()) {
 		return k.GetArithmeticTwapToNow(ctx, poolId, baseAssetDenom, quoteAssetDenom, startTime)
 	} else if endTime.After(ctx.BlockTime()) {
-		return sdk.Dec{}, endTimeInFutureError{endTime, ctx.BlockTime()}
+		return sdk.Dec{}, types.EndTimeInFutureError{EndTime: endTime, BlockTime: ctx.BlockTime()}
 	}
 	startRecord, err := k.getInterpolatedRecord(ctx, poolId, startTime, baseAssetDenom, quoteAssetDenom)
 	if err != nil {
--- a/x/twap/api.go
+++ b/x/twap/api.go
@@ -9,6 +9,26 @@ import (
 	"github.com/osmosis-labs/osmosis/v11/x/twap/types"
 )
 
+type endTimeInFutureErr struct {
+	EndTime   time.Time
+	BlockTime time.Time
+}
+
+func (e endTimeInFutureErr) Error() string {
+	return fmt.Sprintf("called GetArithmeticTwap with an end time in the future."+
+		" (end time %s, current time %s)", e.EndTime, e.BlockTime)
+}
+
+type startTimeAfterEndTimeErr struct {
+	StartTime time.Time
+	EndTime   time.Time
+}
+
+func (e startTimeAfterEndTimeErr) Error() string {
+	return fmt.Sprintf("called GetArithmeticTwap with a start time that is after the end time."+
+		" (start time %s, end time %s)", e.StartTime, e.EndTime)
+}
+
 // GetArithmeticTwap returns an arithmetic time weighted average price.
 // The returned twap is the time weighted average price (TWAP) of:
 // * the base asset, in units of the quote asset (1 unit of base = x units of quote)
@@ -37,14 +57,12 @@ func (k Keeper) GetArithmeticTwap(
 	startTime time.Time,
 	endTime time.Time) (sdk.Dec, error) {
 	if startTime.After(endTime) {
-		return sdk.Dec{}, fmt.Errorf("called GetArithmeticTwap with a start time that is after the end time."+
-			" (start time %s, end time %s)", startTime, endTime)
+		return sdk.Dec{}, startTimeAfterEndTimeErr{startTime, endTime}
 	}
 	if endTime.Equal(ctx.BlockTime()) {
 		return k.GetArithmeticTwapToNow(ctx, poolId, baseAssetDenom, quoteAssetDenom, startTime)
 	} else if endTime.After(ctx.BlockTime()) {
-		return sdk.Dec{}, fmt.Errorf("called GetArithmeticTwap with an end time in the future."+
-			" (end time %s, current time %s)", endTime, ctx.BlockTime())
+		return sdk.Dec{}, endTimeInFutureErr{endTime, ctx.BlockTime()}
 	}
 	startRecord, err := k.getInterpolatedRecord(ctx, poolId, startTime, baseAssetDenom, quoteAssetDenom)
 	if err != nil {
--- a/x/twap/api_test.go
+++ b/x/twap/api_test.go
@@ -213,25 +213,25 @@ func (s *TestSuite) TestGetArithmeticTwap() {
 			recordsToSet: []types.TwapRecord{baseRecord},
 			ctxTime:      baseTime,
 			input:        makeSimpleTwapInput(baseTime, tPlusOne, quoteAssetA),
-			expectError:  twap.EndTimeInFutureErr{BlockTime: baseTime, EndTime: tPlusOne},
+			expectError:  twap.EndTimeInFutureError{BlockTime: baseTime, EndTime: tPlusOne},
 		},
 		"start time after end time": {
 			recordsToSet: []types.TwapRecord{baseRecord},
 			ctxTime:      baseTime,
 			input:        makeSimpleTwapInput(tPlusOne, baseTime, quoteAssetA),
-			expectError:  twap.StartTimeAfterEndTimeErr{StartTime: tPlusOne, EndTime: baseTime},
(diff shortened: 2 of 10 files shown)


x/twap/api_test.go
```diff
@@ -238,9 +237,156 @@ func (s *TestSuite) TestGetArithmeticTwap() {
 	for name, test := range tests {
 		s.Run(name, func() {
 			s.SetupTest()
-			for _, record := range test.recordsToSet {
-				s.twapkeeper.StoreNewRecord(s.Ctx, record)
```
p0mvn, line 242: Note to reviewers: this is the only change in this test fixture (converting loop to using `preSetRecords` helper

x/twap/logic_test.go
```diff
@@ -210,8 +210,9 @@ func TestComputeArithmeticTwap(t *testing.T) {
 	}
 }
 
-// TestPruneRecords tests that all twap records earlier than
-// current block time - RecordHistoryKeepPeriod are pruned from the store.
+// TestPruneRecords tests that twap records earlier than
+// current block time - RecordHistoryKeepPeriod are pruned from the store
+// while keeping the newet record before the above time threshold.
```
p0mvn, line 215:
```suggestion
// while keeping the newest record before the above time threshold.
```

czarcas7ic (approved):
Well done, all the additional twap logic LGTM so I'm approving! 🚀  

Just have small nits that can be addressed at leisure

x/twap/api_test.go
```diff
@@ -238,9 +237,158 @@ func (s *TestSuite) TestGetArithmeticTwap() {
...
+			var twap sdk.Dec
+			var err error
+
+			twap, err = s.twapkeeper.GetArithmeticTwap(s.Ctx, test.input.poolId,
+				test.input.quoteAssetDenom, test.input.baseAssetDenom,
+				test.input.startTime, test.input.endTime)
+
+			if test.expErrorStr != "" {
+				s.Require().Error(err)
+				s.Require().Contains(err.Error(), test.expErrorStr)
```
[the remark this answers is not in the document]
ValarDragon: Why do we care which error was returned?

x/twap/store.go
```diff
@@ -50,10 +50,27 @@ func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
...
+// - Suppose pruning param -48 hour
+// - Suppose swaps at -50 hour, -1hour
+// - A prune would leave us with only one record at -1 hour, and we are not able to get twaps from the
+// [-48 hour, -1 hour] time range.
+func (k Keeper) pruneRecordsBeforeTimeButNewest(ctx sdk.Context, lastTime time.Time) error {
 	store := ctx.KVStore(k.storeKey)
-	iter := store.Iterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
+
+	// Reverse iterator guarantees that we get the newest records first.
+	iter := store.ReverseIterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
```
ValarDragon, line 66: I don't think this is right, to be right we'd have to have skipping logic per pool. This only maintains skipping logic for pool 0

x/twap/api_test.go
```diff
@@ -238,9 +237,158 @@ func (s *TestSuite) TestGetArithmeticTwap() {
...
+		"(1 record younger than keep threshold); to now; ctxTime = start time = end time = after keep threshold": {
+			recordsToSet: []types.TwapRecord{baseRecord},
+			ctxTime:      oneHourAfterKeepThreshold,
+			input:        makeSimpleTwapInput(oneHourAfterKeepThreshold, oneHourAfterKeepThreshold, quoteAssetA),
+			expTwap:      sdk.NewDec(10),
+		},
+		"(1 record younger than keep threshold); with end time; ctxTime = start time = end time = after keep threshold - 1ns": {
+			recordsToSet: []types.TwapRecord{baseRecord},
+			ctxTime:      oneHourAfterKeepThreshold,
+			input:        makeSimpleTwapInput(oneHourAfterKeepThreshold.Add(-time.Millisecond), oneHourAfterKeepThreshold.Add(-time.Millisecond), quoteAssetA),
```
ValarDragon, line 320: I feel like we could probably add some of these table cases more clearly, via a for loop. Don't need to change if you don't want to, but I think could make adding more cases easier

ValarDragon (changes requested):
I don't think the pruning logic works correctly, for keeping the latest record per pool

x/twap/api_test.go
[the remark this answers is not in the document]
czarcas7ic: @ValarDragon here is a recent example https://github.com/osmosis-labs/osmosis/pull/2536#discussion_r958758612

x/twap/store.go
p0mvn: Yeah that's a huge oversight on my part, thanks for catching that. Fixed now and refactored tests to be more thorough and capture coverage across pools

x/twap/store_test.go
```diff
@@ -245,11 +246,18 @@ func (s *TestSuite) TestPruneRecordsBeforeTime() {
...
 		},
-		"base time minus 1, 2 records before (deleted), 1 records at (deleted), 1 records after (not deleted)": {
+		"base time minus 1, 1 record before (not deleted, newest kept), 1 records at (not deleted), 2 records after (not deleted)": {
 			recordsToPreSet: allTestRecords,
 
 			beforeTime: tMin1Record.Time,
 
+			expectedKeptRecords: []types.TwapRecord{tMin2Record, tMin1Record, baseRecord, tPlus1Record},
+		},
+		"base time, 2 records before (one deleted, newest kept), 1 records at (not deleted), 1 records after (not deleted)": {
```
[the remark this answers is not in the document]
p0mvn: Tests were changed so this is now obsolete

x/twap/store_test.go
```diff
@@ -245,11 +246,18 @@ func (s *TestSuite) TestPruneRecordsBeforeTime() {
 
 			expectedKeptRecords: []types.TwapRecord{tPlus1Record},
 		},
-		"base time minus 1, 2 records before (deleted), 1 records at (deleted), 1 records after (not deleted)": {
+		"base time minus 1, 1 record before (not deleted, newest kept), 1 records at (not deleted), 2 records after (not deleted)": {
```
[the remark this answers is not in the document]
p0mvn: Changed tests

x/twap/api_test.go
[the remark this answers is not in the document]
p0mvn: Fixed all errors I observed in the file. Generally, let's create separate issues for such things. I think it's best to keep the scope focused on the actual change instead of fixing drive-by problems

x/twap/api_test.go
p0mvn: I explored this idea [here](https://github.com/osmosis-labs/osmosis/pull/2543/commits/e8e5084271c680ae2310f830ce4b8df1fc565968)

I don't think this is easily doable because if I generalize test cases to always subtract from end time to trigger the "with end time" path, the expected value of `expTwap` changes.

I cannot generalize by instead modifying `ctxTime` to trigger the "with end time" path because test cases' logic sometimes depends on it.

Please let me know if you see something I'm not, and I can change this

p0mvn:
All comments are addressed, please take a look

x/twap/api_test.go
```diff
@@ -213,34 +213,181 @@ func (s *TestSuite) TestGetArithmeticTwap() {
...
+		"(1 record at keep threshold); with end time; ctxTime = base keep threshold, start time = base time - 1ms (source of error); end time = base keep threshold - ms; error": {
+			recordsToSet: []types.TwapRecord{baseRecord},
+			ctxTime:      baseTimePlusKeepPeriod,
+			input:        makeSimpleTwapInput(baseTime.Add(-time.Millisecond), baseTimePlusKeepPeriod.Add(-time.Millisecond), quoteAssetA),
+			expectError:  twap.TimeTooOldError{Time: baseTime.Add(-time.Millisecond)},
+		},
+		"(2 records); to now; with one directly at threshold, interpolated": {
+			recordsToSet: []types.TwapRecord{baseRecord, recordBeforeKeepThreshold},
+			ctxTime:      baseTimePlusKeepPeriod,
+			input:        makeSimpleTwapInput(baseTime, baseTimePlusKeepPeriod, quoteAssetA),
```
mattverse, line 360: Seeing all these test cases, Im thinking if there would be a way to declare all these test cases as variables(including the expected TWAP) in the future

x/twap/listeners_test.go
```diff
@@ -201,17 +201,31 @@ func (s *TestSuite) TestEndBlock() {
...
+
+	// Assume some time has passed and new record created.
+	s.Ctx = s.Ctx.WithBlockTime(tPlus10sp5Record.Time)
+	newestRecord := tPlus10sp5Record
+
+	s.twapkeeper.StoreNewRecord(s.Ctx, newestRecord)
+
+	twapsBeforeEpoch, err := s.twapkeeper.GetAllHistoricalTimeIndexedTWAPs(s.Ctx)
 	s.Require().NoError(err)
+
```
mattverse, line 221:
```suggestion
```

x/twap/store_test.go
```diff
@@ -201,70 +201,177 @@ func (s *TestSuite) TestGetRecordAtOrBeforeTime() {
...
+func (s *TestSuite) TestPruneRecordsBeforeTimeButNewest() {
+	// These are manually created to be able to refer to them by name
+	// for convenience.
+	pool1Min2SBaseMs, pool2Min1SBaseMs, pool3BaseSecBaseMs, pool4Plus1SBaseMs := s.createTestRecordsFromTime(baseTime)
+	pool1Min2SMin1Ms, pool2Min1SMin1Ms, pool3BaseSecMin1Ms, pool4Plus1SMin1Ms := s.createTestRecordsFromTime(baseTime.Add(-time.Millisecond))
+	pool1Min2SMin2Ms, pool2Min1SMin2Ms, pool3BaseSecMin2Ms, pool4Plus1SMin2Ms := s.createTestRecordsFromTime(baseTime.Add(2 * -time.Millisecond))
+	pool1Min2SMin3Ms, pool2Min1SMin3Ms, pool3BaseSecMin3Ms, pool4Plus1SMin3Ms := s.createTestRecordsFromTime(baseTime.Add(3 * -time.Millisecond))
+
+	// Create records that match times 1:1 with other pools but have a different pool ID.
+	pool5Min2SBaseMs, pool5Min1SBaseMs, pool5BaseSecBaseMs, pool5Plus1SBaseMs := s.createTestRecordsFromTime(baseTime)
```
mattverse, line 215: im wondering if there's reason we need to test with so many pools, since I think max 3 pools would suffice testing any case possible with multiple pools

mattverse (commented):
It might just be a personal opinion, but tests in `store_test.go` seems a little bit hard to follow,(test case names are not quite intuitive on what exactly we want to test in that specific case). Other than that main logic lgtm!

x/twap/store.go
```diff
@@ -12,6 +12,25 @@ import (
...
+}
+
+type twapNotFoundError struct {
+	Asset0Denom string
+	Asset1Denom string
+}
+
+func (e twapNotFoundError) Error() string {
+	return fmt.Sprintf("TWAP not found, but there are other twaps available for this time."+
+		" Please make sure tha asset0denom and asset1denom (%s, %s) are correct, and in order (asset0 > asset1)?", e.Asset0Denom, e.Asset1Denom)
```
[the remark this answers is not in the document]
p0mvn: We can address this separately if desired. This is simply extracting errors to be reused in assertions in tests

x/twap/store.go
```diff
@@ -50,15 +69,36 @@ func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
...
+// [-48 hour, -1 hour] time range.
+func (k Keeper) pruneRecordsBeforeTimeButNewest(ctx sdk.Context, lastTime time.Time) error {
 	store := ctx.KVStore(k.storeKey)
-	iter := store.Iterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
+
+	// Reverse iterator guarantees that we get the newest records first.
+	iter := store.ReverseIterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
 	defer iter.Close()
+
+	seenPools := map[uint64]struct{}{}
```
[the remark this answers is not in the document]
p0mvn: The goal is to have a set. If a boolean was here, it would be wasting space. `struct{}{}` allocates 0 bytes for each key and `map[uint64]struct{}{}` is idiomatic with a set in Go

x/twap/api_test.go
p0mvn: These are inputs though and they are constructed based on the logic / setup of the test case. I don't think we should be extracting these since the setup logic varies form test case to test case

x/twap/store_test.go
p0mvn: They all start at the different base times:
https://github.com/osmosis-labs/osmosis/blob/2c7649f4ed193e05ae921c65cb2a037d9c3412d3/x/twap/store_test.go#L209-L212


The goal is not only to test the inter-pool correctness but also how pruning works with respect to different times

x/twap/api.go
```diff
@@ -9,6 +9,26 @@ import (
...
+
+type startTimeAfterEndTimeError struct {
+	StartTime time.Time
+	EndTime   time.Time
+}
+
+func (e startTimeAfterEndTimeError) Error() string {
+	return fmt.Sprintf("called GetArithmeticTwap with a start time that is after the end time."+
+		" (start time %s, end time %s)", e.StartTime, e.EndTime)
+}
```
ValarDragon, line 30: Lets move these to bottom of file?

x/twap/store.go
```diff
@@ -50,15 +69,36 @@ func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
...
+// For example:
+// - Suppose pruning param -48 hour
+// - Suppose swaps at -50 hour, -1hour
+// - A prune would leave us with only one record at -1 hour, and we are not able to get twaps from the
+// [-48 hour, -1 hour] time range.
+func (k Keeper) pruneRecordsBeforeTimeButNewest(ctx sdk.Context, lastKeptTime time.Time) error {
 	store := ctx.KVStore(k.storeKey)
-	iter := store.Iterator([]byte(types.HistoricalTWAPTimeIndexPrefix), types.FormatHistoricalTimeIndexTWAPKey(lastTime, 0, "", ""))
+
+	// Reverse iterator guarantees that we get the newest records first.
```
ValarDragon, line 84:
```suggestion
	// Reverse iterator guarantees that we iterate through the newest per pool first.
	// Due to how its indexed, we will only iterate times starting from lastKeptTime
```

x/twap/store.go
```diff
@@ -50,15 +69,36 @@ func (k Keeper) storeHistoricalTWAP(ctx sdk.Context, twap types.TwapRecord) {
...
+// pruneRecordsBeforeTimeButNewest prunes all records for each pool before the given time but the newest
+// record. The reason for preserving at least one record earlier than the keep period is
+// to ensure that we have a record to interpolate from in case there is only one or no records
+// within the keep period.
+// For example:
+// - Suppose pruning param -48 hour
+// - Suppose swaps at -50 hour, -1hour
+// - A prune would leave us with only one record at -1 hour, and we are not able to get twaps from the
+// [-48 hour, -1 hour] time range.
+func (k Keeper) pruneRecordsBeforeTimeButNewest(ctx sdk.Context, lastKeptTime time.Time) error {
```
ValarDragon, line 81: Should we rename lastKeptTime to firstRemoveTime due to how inclusivity works in the iterator?

ValarDragon:
Logic LGTM, have not yet reviewed tests!

x/twap/api.go
p0mvn: Moved to `types/errros,go`

x/twap/store.go
ValarDragon: ah that makes sense

ValarDragon (approved):
Nice job on the feature + thorough tests!

I somewhat feel like we went overboard with long hand written test cases here though, versus having a lot of these vectors made via loops, randomization, or more segmented.

Issue #2481, opened 2022-08-22
Move x/gov to osmosis


ValarDragon:
## Background

We have context in a few slack discussions. We should move x/gov from our fork into the osmosis repo

PR #2552, opened 2022-08-30
chore: port x/gov

alexanderbez:
Closes: #2481

## Brief Changelog

* Port x/gov from osmosis' fork of the Cosmos SDK from the `osmosis-main` branch.


alexanderbez:
This might be trickier than I originally thought, mainly because of external modules, e.g. x/params, that rely on cosmos-sdk repo x/gov.

ValarDragon:
Oh yikes, would we have to import this as an independent go.mod, and then do a replace in our main go.mod?

ValarDragon:
On skim, seems like its only these gov handlers that are the problem, which if so, we should be able to make a function that takes in an SDK type handler and returns an osmosis gov type handler?

alexanderbez:
Jumping back on this...

alexanderbez:
@ValarDragon after updating the import paths to `github.com/osmosis-labs/osmosis/v11/x/gov`, we get more build errors and the wiring just becomes to messy, mainly in tests where we use simapp for testing (which makes no sense to me)

ValarDragon:
Closing in favor of #2854 

I think we can update the tests in a second pass

PR #2556, opened 2022-08-31
fix: e2e flakiness

czarcas7ic:

It seems we have zero issue running e2e tests locally, however have intermittent issues running through CI. By looking through previous failed runs and where they fail, they seem to fail specifically when starting the new containers after upgrade. While I don't have evidence since this is hard to debug when its only happening in CI, I have a feeling this has to do with weird docker networking problems rather than the upgrade itself failing. Because of this, I wanted to see if changing from the rpcClient health call to instead a height call would make a difference. 

## Brief Changelog

- changes the run command to check the height rather than health

## Testing and Verifying

### DIFF
--- a/tests/e2e/configurer/chain/node.go
+++ b/tests/e2e/configurer/chain/node.go
@@ -58,10 +58,12 @@ func (n *NodeConfig) Run() error {
 		return err
 	}
 
+	n.rpcClient = rpcClient
+
 	require.Eventually(
 		n.t,
 		func() bool {
-			if _, err := rpcClient.Health(context.Background()); err != nil {
+			if _, err := n.QueryCurrentHeight(); err != nil {
 				return false
 			}
 
@@ -72,8 +74,6 @@ func (n *NodeConfig) Run() error {
 		time.Second,
 		"Osmosis node failed to produce blocks",
 	)
-
-	n.rpcClient = rpcClient
 
 	if err := n.extractOperatorAddressIfValidator(); err != nil {
 		return err


p0mvn:
Great point. I think this is very likely to be the cause

p0mvn (approved):

tests/e2e/configurer/chain/node.go
```diff
@@ -58,10 +58,12 @@ func (n *NodeConfig) Run() error {
 		return err
 	}
 
+	n.rpcClient = rpcClient
+
 	require.Eventually(
 		n.t,
 		func() bool {
-			if _, err := rpcClient.Health(context.Background()); err != nil {
+			if _, err := n.QueryCurrentHeight(); err != nil {
```
p0mvn, line 66: On another look, what difference does it make?

`QueryCurrentHeight` makes an RPC call under the hood

cc: @czarcas7ic

czarcas7ic: Do you know why we were defining n.rpcClient after the Eventually check?

To answer your question, I have had issues with the health endpoint in the past, whereas the current height check has never failed me. This change was less based in science and more based on gut feeling...that being said I haven't seen a false fail yes 😅 *vigorously knocks on wood*

p0mvn: > Do you know why we were defining n.rpcClient after the Eventually check?

I don't think it matters where we make an assignment in the original version. In the new one, we do have to do it before the require because that's what `n.QueryCurrentHeight` uses

Issue #2148, opened 2022-07-19
Measure and lower overhead of KeeperTestHelper.Setup()


ValarDragon:
## Background

Locally, KeeperTestHelper takes about 20ms, which adds up over the hundreds of invokations around osmosis tests, to tens of seconds of delay. More importantly, this appears to contribute a minute to CI test time. 

I added a micro-benchmark in #2143 , and it shows that 75% of this time is in heap. Its a nice / useful project for someone to investigate memoryprofiling that benchmark, and to see where we can lower heap allocation to improve our CI time.

NOTE: I'm just filing as I noticed this recently, not saying its something thats a priority

## Suggested Design

- Look into why the following functions need to allocate so much RAM:
![image](https://user-images.githubusercontent.com/6440154/179847310-2bc16c2a-5774-4577-8007-d864f13674b8.png)
- See if theres small fixes to reduce this in our CI setting
- Methods for ease of CTRL+F:
  - modulemanager.InitGenesis (on default genesis)
  - iavl.LoadVersion
  - app.MakeEncodingConfig()
    - Its all in RegisterLegacyAminoCodec
    - Presumably this can be cached / globally init once, and then copied from that?
  - app.NewDefaultGenesisState
    - This one seems basic, presumably we can just cache this and only compute it once

## Acceptance Criteria

- [ ] We understand allocation overhead root causes
- [ ] We've reduced allocation in the test setup logic

ValarDragon:
* iavl.LoadVersion likely coming from us forcing GC within iavl. We should probably remove that logic that forces garbage collection, and only make that trigger if memory usage is above some number. This is almost certainly the big root cause of slowness of our golang tests
* InitGenesis overhead seems to 75% come from JSON unmarshalling, surely theres ways we can improve these.

RusAkh:
I'd like to look into this

p0mvn:
@RusAkh Sweet, thanks!

ValarDragon:
Just fixed two of these in #7902 , gonna close for now. Lets re-open with some new benchmarks.

Issue #2554, opened 2022-08-31
Make KeeperTestHelper constructor, with a no-op suite.Sute


ValarDragon:
## Background

@iboss-ptk is working on CGO bindings to allow testing cosmwasm contracts against live state machine. One issue is that a lot of convenience methods are needed that are the same as what we do in keeper test helper. This can't currently be imported due to issues with suite.Suite depending on testing.T.

## Suggested Design

Make a constructor for KeeperTestHelper that has a no-op for testing suite

## Acceptance Criteria

- [ ] We can make a KeeperTestHelper struct and use all existing methods, without needing testing.T

hieuvubk:
Is anyone working on this issue? I think removing the suite from `KeeperTestHelper` and moving to `KeeperTestSuite` will solve this problem.
Pls lmk what u think

PR #2558, opened 2022-08-31
Cache genesis setup in App.Setup

ValarDragon:

cref #2148 , small speedup to app setup time / all tests

```
BEFORE
BenchmarkSetup-16             57          20107194 ns/op         1753442 B/op      22513 allocs/op

AFTER
BenchmarkSetup-16    	      76	  16989387 ns/op	 1087661 B/op	   13536 allocs/op
```

## Brief Changelog

deduplicate genesis creation & encoding config every time its needed

## Testing and Verifying

### DIFF
--- a/app/apptesting/test_suite.go
+++ b/app/apptesting/test_suite.go
@@ -71,7 +71,9 @@ func (s *KeeperTestHelper) SetEpochStartTime() {
 		epoch.StartTime = s.Ctx.BlockTime()
 		epochsKeeper.DeleteEpochInfo(s.Ctx, epoch.Identifier)
 		err := epochsKeeper.AddEpochInfo(s.Ctx, epoch)
-		s.Require().NoError(err)
+		if err != nil {
+			panic(err)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


app/apptesting/test_suite.go
```diff
@@ -71,7 +71,9 @@ func (s *KeeperTestHelper) SetEpochStartTime() {
 		epoch.StartTime = s.Ctx.BlockTime()
 		epochsKeeper.DeleteEpochInfo(s.Ctx, epoch.Identifier)
 		err := epochsKeeper.AddEpochInfo(s.Ctx, epoch)
-		s.Require().NoError(err)
+		if err != nil {
+			panic(err)
+		}
```
p0mvn, line 76: Why this change?

p0mvn (approved):

czarcas7ic (approved):
Simple fix, nice 😄

app/apptesting/test_suite.go
ValarDragon: We can't use any suite methods for the benchmark (until #2554 )

PR #2563, opened 2022-09-01
Add error handling for incentive hooks interface

Anmol1696:
Part of: #2520 
Not sure if this PR is needed, as there are no implementation for the hooks.

* Avoid using panics on hook functions, and instead add ability to return errors
* Hooks errors can now be handled with [ApplyFuncIfNoError](https://github.com/osmosis-labs/osmosis/blob/main/osmoutils/cache_ctx.go#L16)
* This implementation allows hooks interface functions to return an error.
* If functions return an error any state changes will not be written to the state.

## Brief Changelog
* Update the interface for IncentiveHooks hooks, to add return type error
**Note: There are no current implementations for the hook functions**

## Testing and Verifying

### DIFF
--- a/x/incentives/keeper/distribute.go
+++ b/x/incentives/keeper/distribute.go
@@ -74,6 +74,7 @@ func (k Keeper) moveActiveGaugeToFinishedGauge(ctx sdk.Context, gauge types.Gaug
 	if err := k.deleteGaugeIDForDenom(ctx, gauge.Id, gauge.DistributeTo.Denom); err != nil {
 		return err
 	}
+	
 	// Error is not handled as Hooks use osmoutils.ApplyFuncIfNoError()
 	_ = k.hooks.AfterFinishDistribution(ctx, gauge.Id)
 	return nil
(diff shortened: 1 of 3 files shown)


x/incentives/keeper/distribute.go
```diff
@@ -74,7 +74,8 @@ func (k Keeper) moveActiveGaugeToFinishedGauge(ctx sdk.Context, gauge types.Gaug
 	if err := k.deleteGaugeIDForDenom(ctx, gauge.Id, gauge.DistributeTo.Denom); err != nil {
 		return err
 	}
-	k.hooks.AfterFinishDistribution(ctx, gauge.Id)
+	// Error is not handled as Hooks use osmoutils.ApplyFuncIfNoError()
```
Anmol1696, line 77:
```suggestion
	
	// Error is not handled as Hooks use osmoutils.ApplyFuncIfNoError()
```

ValarDragon:
err not sure we want error catching / panic behavior for all hooks

I think epochs is a clear case where we do want this, since everything is in a clear scope.

Anmol1696:
> err not sure we want error catching / panic behavior for all hooks
> 
> I think epochs is a clear case where we do want this, since everything is in a clear scope. (Same comment for the other two open PRs)

Dont mind closing the PRs, but just would like to have some discussion around it in the issue.

Issue #1843, opened 2022-06-20
e2e: fix superfluid tests when upgrade is disabled


p0mvn:
## Background

Superfluid e2e tests are currently broken when run from the current branch (without upgrade logic):
```
    e2e_util_test.go:284: sending 100000000000000000000gamm/pool/1 from osmo1vr8qv9ve9dljsh3h0rpgj0e4z6f0sc783pch84 to osmo1en70tmaq3dzzy8dw372q57ltjlfdyke62sdptw on chain-id: osmo-test-a
    e2e_util_test.go:42: 
        	Error Trace:	e2e_util_test.go:42
        	            				e2e_util_test.go:286
        	            				e2e_test.go:30
        	Error:      	Condition never satisfied
        	Test:       	TestIntegrationTestSuite/TestSuperfluidVoting
        	Messages:   	tx returned a non-zero code; stdout: , stderr: 
```

The initial guess is because superfluid logic is coupled to the upgrade tests.

## Suggested Design

Superfluid tests should run independently from the upgrade tests. They should pass with and without upgrade logic

## Acceptance Criteria

- e2e `TestSuperfluidVoting` works when upgrade logic is disabled via `OSMOSIS_E2E_SKIP_UPGRADE` environment variable

Issue #2002, opened 2022-07-08
ux(e2e): Makefile steps to run e2e test suite locally


p0mvn:
## Background

Currently, there is a pre-requisites to get e2e working locally. This step includes building `osmosis:debug` image to be pre-built for testing the upgrades. This setup introduces a barrier to entry for developers new to e2e.

We should have 2 makefile steps that run e2e "out of the box" while installing everything needed in the background:
- `make test-e2e` - installs dependencies and runs a full e2e suite
- `make test-e2e-short` - installs dependencies and runs short e2e test suite with upgrade and IBC tests off
   * `OSMOSIS_E2E_SKIP_IBC` environment variable must be true
   * `OSMOSIS_E2E_SKIP_UPGRADE` environment variable must be true

## Acceptance Criteria
- 2 suggested make file steps are introduced
- Documentation added to the e2e README that Docker should still be installed manually (link added)
- each step can be run multiple times locally without issues with no extra setup

czarcas7ic:
100%, sometimes I even find myself forgetting to build the debug image locally when working on a new VM (or not rebuilding the debug image when working with a newer version)

p0mvn:
@czarcas7ic Do you think `osmosis:debug` image should be rebuilt every time `make test-e2e` is run?

I think there is a trade-off. If rebuild every time, it makes you wait for results longer. If not every time (only the image is not present), easy to forget to rebuild it and run e2e on an old Osmosis version.

I'm leaning towards only building it on the first run to avoid waiting but curious to know what anyone else prefers

czarcas7ic:
@p0mvn I for sure do not think it should be rebuild on ever run. I think we should just check for the images existence and if it doesn't exist we build it. Would also be neat if we can check for the age of the osmosis:debug image and if it is greater than x days we rebuild it. Just the first idea that came to my mind, could be much better ideas. Will think more about it

ValarDragon:
What is in the `osmosis:debug` image? Why would we not just rebuild it every time relevant source has a change?

czarcas7ic:
@ValarDragon we do rebuild this in CI every time it is run. The problem is when someone is running the e2e test suite locally, it is possible they already built the osmosis:debug image and instead of rebuilding it could just utilize this same image for multiple runs. The osmosis debug image is just osmosis build off all local changes.

[...]

PR #2568, opened 2022-09-01
fix(e2e): enable pool creation and sf tests run without upgrade

p0mvn:
Closes: #1843

The main focus of this PR is fixing pool creation and superfluid tests that were previously disabled when run without upgrade or ibc logic.

The source of the problem was that we were attempting to create a pool with IBC denoms. Since those did not exist when IBC tests were disabled, pool creation would fail. The SF tests were dependent on pool creation so they would fail as well.

Now, we still create a pool with native denoms but skip IBC denom pool creation.

## Next Steps

Currently, our e2e tests still depend on each other. For example, superfluid tests cannot run without pool creation. We should decouple them so that each test can be run in isolation. That is, the superfluid test should create a separate pool just for testing superfluid voting.

## Drive by changes

- Adds a `make test-e2e-short` command that was proposed in #2002
- Increases the difference between the expedited and regular proposal to 2 seconds because it was failing locally due to a marginal difference.

## Testing and Verifying


tests/e2e/e2e_test.go
```diff
@@ -244,10 +260,11 @@ func (s *IntegrationTestSuite) TestExpeditedProposals() {
 	}
 
 	// compare the time it took to reach pass status to expected expedited voting period
-	expeditedVotingPeriodDuration := time.Duration(chain.ExpeditedVotingPeriod * 1000000000)
+
+	expeditedVotingPeriodDuration := time.Duration(chain.ExpeditedVotingPeriod * float32(time.Second))
 	timeDelta := elapsed - expeditedVotingPeriodDuration
 	// ensure delta is within one second of expected time
-	s.Require().Less(timeDelta, time.Second)
+	s.Require().Less(timeDelta, 2*time.Second)
```
p0mvn, line 267: N.B: it was failing for me locally with a marginal difference in milliseconds. so I ended up increasing the difference to 2 seconds. Now, it passes every time

czarcas7ic (approved):
Had a similar fix locally while I was working on something e2e related, lets merge this so I can work off this new version!

Issue #2557, opened 2022-08-31
Add addToExistingLock in E2E test framework


czarcas7ic:
## Background

We recently refactored AddToExistingLock in this PR https://github.com/osmosis-labs/osmosis/pull/1979

As discussed in a call, we should have test coverage on this, and the e2e test suite seems like a good place for this

## Suggested Design

- After a lock is created in e2e, add to that existing lock pre upgrade
  - If upgrade is not skipped, after the upgrade, we should attempt to add to the lock that existed pre upgrade as well as a lock that was created post upgrade
  - If upgrade is skipped, just attempt to add to an existing lock that was created
- Ensure the above is done for both superfluid and non superfluid locks

## Acceptance Criteria

- design implemented with all tests passing

Issue #2569, opened 2022-09-01
fix(e2e): decouple tests to allow running them in isolation


p0mvn:
## Background

Currently, we have the following dependencies between tests:

- ["pool creation test"](https://github.com/osmosis-labs/osmosis/blob/ffccf80f995e70d67f24b22350214af53bc8449f/tests/e2e/e2e_test.go#L21) depends on either the [pre upgrade logic ](https://github.com/osmosis-labs/osmosis/blob/ffccf80f995e70d67f24b22350214af53bc8449f/tests/e2e/configurer/upgrade.go#L108)or [IBC token transfer](https://github.com/osmosis-labs/osmosis/blob/ffccf80f995e70d67f24b22350214af53bc8449f/tests/e2e/e2e_test.go#L33).

- ["superfluid test"](https://github.com/osmosis-labs/osmosis/blob/ffccf80f995e70d67f24b22350214af53bc8449f/tests/e2e/e2e_test.go#L47) depends on the above "pool creation test"

## Suggested Design

We should be able to run each test in isolation for faster debugging and to be able to have shorter makefile steps that can meaningfully run e2e tests without executing the long upgrade logic.

1. To decouple pool creation tests from IBC and upgrade, send IBC tokens inside `TestCreatePoolPostUpgrade` prior to calling `node.CreatePool(...)` with IBC denoms.

2. To decouple superfluid tests from pool creation, create a separate superfluid-specific pool in `TestSuperfluidVoting`

## Acceptance Criteria
- IBC test can be run in isolation
- pool creation test can be run in isolation
- superfluid voting test can be run in isolation

p0mvn:
This is relevant to all tests that we have in `e2e_test.go` today, not just the ones listed in acceptance criteria of this issue

ThanhNhann:
I think all the tests are running independently now, I have tested each of them (just one test at a time) with adjust env in `launch.json` and see that they were all passed, below are images of tests

- TestAddToExistingLock
![TestAddToExistingLock](https://user-images.githubusercontent.com/28915032/215688409-b6f26a46-1325-4f8a-ad41-82e64d731d08.png)

- TestAddToExistingLockPostUpgrade
![TestAddToExistingLockPostUpgrade](https://user-images.githubusercontent.com/28915032/215688416-c02e193a-87d2-48e8-ac43-764ca4b2cf31.png)

- TestArithemeticTWAP
[...]

PR #2571, opened 2022-09-02
feat: addToExistingLock e2e tests

czarcas7ic:
Closes: #2557 

This PR adds the necessary framework and tests to the e2e suite to properly test addToExistingLock logic works for locks and superfluid locks both pre and post upgrade.

## Brief Changelog

- If upgrade is set to True, we enable gamm/pool/1 as a superfluid asset, then lock and add to existing lock just for chainA
  - Then, after upgrade, we attempt the add to this lock that existed pre upgrade (both regular and superfluid)
  - We also create a new lock on chainB and add to this existing lock, all post upgrade (both regular and superfluid)
- If upgrade is set to False, all this logic is done at once 
- Adds `EnableSuperfluidAsset` method
- Adds `LockAndAddToExistingLock` method
  - This tests addToExistingLock for both regular and superfluid locks
- Adds `AddToExistingLock` method
  - This method automatically retrieves the desired lockID via the provided address, denom, and duration.
  - This lockID is then used to check the coin amount of the locked denom, a lock action is done, and then the method checks if that same lockID's coin amount has increased by the expected value
- Adds `GetWallet` method

## Testing and Verifying


tests/e2e/e2e_test.go
```diff
@@ -41,67 +42,98 @@ func (s *IntegrationTestSuite) Test01IBCTokenTransfer() {
 // This is the reason for prefixing the name with 02 to ensure
 // correct order.
 func (s *IntegrationTestSuite) Test02CreatePool() {
-	chain := s.configurer.GetChainConfig(0)
-	node, err := chain.GetDefaultNode()
+	chainA := s.configurer.GetChainConfig(0)
+	chainB := s.configurer.GetChainConfig(1)
```
p0mvn, line 46: Why do we need to run this on both chains? If it works on one chain, it should work on another chain as well since they are identical. I think duplication only adds to time and increases complexity

tests/e2e/e2e_test.go
```diff
@@ -41,67 +42,98 @@ func (s *IntegrationTestSuite) Test01IBCTokenTransfer() {
...
 	}
 
-	node.CreatePool("ibcDenomPool.json", initialization.ValidatorWalletName)
+	chainANode.CreatePool("ibcDenomPool.json", initialization.ValidatorWalletName)
+	chainBNode.CreatePool("ibcDenomPool.json", initialization.ValidatorWalletName)
 }
 
-// Test03SuperfluidVoting tests that superfluid voting is functioning as expected.
+// Test03AddToExistingLockPostUpgrade tests lockups to both regular and superfluid locks.
+// Specifically, we ensure addToExistingLock works for both preupgrade and postupgrade locks.
```
p0mvn, line 65: Can we decouple the logic that depends on the pre-upgrade state so that it can be disabled for running locally?

tests/e2e/e2e_test.go
```diff
@@ -41,67 +42,98 @@ func (s *IntegrationTestSuite) Test01IBCTokenTransfer() {
...
-	// superfluid delegate from non self-delegation wallet to validator 1 on chain.
-	node.SuperfluidDelegate(chain.LatestLockNumber, chain.NodeConfigs[1].OperatorAddress, walletName)
-
 	// create a text prop, deposit and vote yes
-	node.SubmitTextProposal("superfluid vote overwrite test", sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(config.InitialMinDeposit)), false)
-	chain.LatestProposalNumber += 1
-	node.DepositProposal(chain.LatestProposalNumber, false)
-	for _, node := range chain.NodeConfigs {
-		node.VoteYesProposal(initialization.ValidatorWalletName, chain.LatestProposalNumber)
+	chainANode.SubmitTextProposal("superfluid vote overwrite test", sdk.NewCoin(appparams.BaseCoinUnit, sdk.NewInt(config.InitialMinDeposit)), false)
```
p0mvn, line 124: nit: I don't really understand the value of the rename since we are only using one chain in this test

tests/e2e/configurer/upgrade.go
```diff
@@ -124,6 +127,19 @@ func (uc *UpgradeConfigurer) CreatePreUpgradeState() error {
 
 	chainANode.CreatePool("pool1A.json", initialization.ValidatorWalletName)
 	chainBNode.CreatePool("pool1B.json", initialization.ValidatorWalletName)
+
+	// enable superfluid assets on both chains
+	chainA.EnableSuperfluidAsset("gamm/pool/1")
```
p0mvn, line 132: This creates a dependency for the superfluid tests on the upgrade tests. Now, I don't think we can the superfluid test with the upgrade disabled because superfluid asset is not enabled.

Can we keep this logic in the test itself?

tests/e2e/e2e_test.go
```diff
@@ -41,67 +42,98 @@ func (s *IntegrationTestSuite) Test01IBCTokenTransfer() {
...
+	chainA := s.configurer.GetChainConfig(0)
+	chainB := s.configurer.GetChainConfig(1)
+	chainANode, err := chainA.GetDefaultNode()
+	s.NoError(err)
+	chainBNode, err := chainB.GetDefaultNode()
+	s.NoError(err)
+	if s.skipUpgrade {
+		// enable superfluid assets
+		chainA.EnableSuperfluidAsset("gamm/pool/1")
+		chainB.EnableSuperfluidAsset("gamm/pool/1")
```
p0mvn, line 80: So it seems that the superfluid tests depend on this when the upgrade is disabled. Can we make each test independent by making them operate on different assets, pools etc?

The goal is to be able to run each test in isolation

p0mvn: https://github.com/osmosis-labs/osmosis/issues/2569

tests/e2e/e2e_test.go
czarcas7ic: Discussed this offline, but for anyone else reading, I was using the ordered tests as state creation. After a discussion with Roman, we decided we will try to decouple as many of these tests as possible in order to eventually allow most of these tests to run independently.

tests/e2e/e2e_test.go
czarcas7ic: I personally like to know which chain's state it is I am modifying. I would agree that this is unnecessary if the state was reset after each run, but since it is retained I think it is helpful. 

Another reason for this is for those learning the layout of the e2e for the first time. It's not obvious that `s.configurer.GetChainConfig(0)` is actually chainA. By referring to it as chainA even in isolation, it is easier to understand as a whole imo. 

If you feel strongly against it though I can change this back!

czarcas7ic:
@p0mvn I extended the CreatePool method to return which poolID was created. This allows for this method to be called in individual tests and utilize the CreatePool method without needing to rely on other tests. Now the only two tests that rely on one another is IBC Send and Create Pool, and this is only because the Create Pool test requires an IBC denom to be present in the wallet. Lmk what you think!

tests/e2e/e2e_test.go
p0mvn: No that makes sense, ty!

tests/e2e/e2e_test.go
```diff
@@ -137,16 +126,44 @@ func (s *IntegrationTestSuite) Test03SuperfluidVoting() {
...
+	chainA := s.configurer.GetChainConfig(0)
+	chainANode, err := chainA.GetDefaultNode()
+	s.NoError(err)
+	if !s.skipUpgrade {
+		// ensure we can add to existing locks and superfluid locks that existed pre upgrade on chainA
+		// we use the hardcoded gamm/pool/1 as this was the one that was created pre upgrade
+		lockupWalletAddrA, lockupWalletSuperfluidAddrA := chainANode.GetWallet("lockup-wallet"), chainANode.GetWallet("lockup-wallet-superfluid")
+		chainANode.AddToExistingLock(sdk.NewInt(1000000000000000000), "gamm/pool/1", "240s", lockupWalletAddrA)
+		chainANode.AddToExistingLock(sdk.NewInt(1000000000000000000), "gamm/pool/1", "240s", lockupWalletSuperfluidAddrA)
+	}
```
p0mvn, line 141: Do you think we could move all this pre-upgrade dependent logic into its own test  fixture `TestAddToExistingLockPreUpgrade` while renaming this to `TestAddToExistingLock`?

The motivation is twofold:
- completely decouple upgrade dependent and independent tests
- if we were to run `make test-e2e-short` with the current design, the name of `TestAddToExistingLockPostUpgrade` would not make sense because the upgrade tests are not run at all

tests/e2e/e2e_test.go
```diff
@@ -137,16 +126,44 @@ func (s *IntegrationTestSuite) Test03SuperfluidVoting() {
...
+
+	// ensure we can add to new locks and superfluid locks
+	// create pool and enable superfluid assets
+	poolId := chainANode.CreatePool("nativeDenomPool.json", chainA.NodeConfigs[0].PublicAddress)
+	chainA.EnableSuperfluidAsset(fmt.Sprintf("gamm/pool/%d", poolId))
+
+	// setup wallets and send gamm tokens to these wallets on chainB
+	lockupWalletAddrB, lockupWalletSuperfluidAddrB := chainANode.CreateWallet("TestAddToExistingLockPostUpgrade"), chainANode.CreateWallet("TestAddToExistingLockPostUpgradeSuperfluid")
```
p0mvn, line 149: If we go with the splitting suggestion above, it might make sense to rename walltes as well

p0mvn (approved):
Excellent work on making these tests independent! I feel like the debugging experience for each test is going to be much better now

tests/e2e/e2e_test.go
czarcas7ic: Good idea! Makes sense to split these two apart even further, addressing now!

Issue #2315, opened 2022-08-05
e2e testing should clear existing containers automatically


nicolaslara:
## Background

When running e2e tests locally, if the tests have already run (and failed) the old containers aren't removed. This leads to 

```
    e2e_setup_test.go:96:
        	Error Trace:	/Users/nicolas/devel/osmosis/tests/e2e/e2e_setup_test.go:96
        	            				/Users/nicolas/devel/osmosis/tests/e2e/suite.go:132
        	            				/Users/nicolas/devel/osmosis/tests/e2e/e2e_setup_test.go:40
        	Error:      	Received unexpected error:
        	            	container already exists

        	            	github.com/ory/dockertest/v3.(*Pool).RunWithOptions
        	            		/Users/nicolas/go/pkg/mod/github.com/ory/dockertest/v3@v3.9.1/dockertest.go:453
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/containers.(*Manager).RunHermesResource
        	            		/Users/nicolas/devel/osmosis/tests/e2e/containers/containers.go:138
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/configurer.(*baseConfigurer).runIBCRelayer
        	            		/Users/nicolas/devel/osmosis/tests/e2e/configurer/base.go:110
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/configurer.(*baseConfigurer).RunIBC
        	            		/Users/nicolas/devel/osmosis/tests/e2e/configurer/base.go:77
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/configurer.withIBC.func1
        	            		/Users/nicolas/devel/osmosis/tests/e2e/configurer/setup.go:22
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/configurer.withUpgrade.func1
        	            		/Users/nicolas/devel/osmosis/tests/e2e/configurer/setup.go:32
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e/configurer.(*UpgradeConfigurer).RunSetup
        	            		/Users/nicolas/devel/osmosis/tests/e2e/configurer/upgrade.go:134
        	            	github.com/osmosis-labs/osmosis/v10/tests/e2e.(*IntegrationTestSuite).SetupSuite
        	            		/Users/nicolas/devel/osmosis/tests/e2e/e2e_setup_test.go:95
        	            	github.com/stretchr/testify/suite.Run
[...]

PR #2581, opened 2022-09-02
ux(e2e): Makefile steps to run e2e test suite locally

p0mvn:
Closes: #2002, #2315

1. Add scripts and makefile steps for rebuilding docker image and clearing resources
- `make e2e-check-image-sha`
    * compares the git sha assigned to the pre-existing tag against a local Git commit sha
    * rebuilds the image if hashes do not match
- `make remove-resources`
    * removes all e2e Docker resources, including containers and network
- `make e2e-setup`
    * combines and runs both of the above steps

Create 3 separate Makefile steps for running e2e:
- `make test-e2e` to run in CI
    * before starting runs `make e2e-setup`
- `make test-e2e-debug` to run full suite locally
    * before starting runs `make e2e-setup`
    * does not clear Docker resources at the end (for debugging)
- `make test-e2e-short` to run short test suite locally
    * before starting runs `make e2e-setup`
    * excludes state-sync, IBC and upgrade tests
    * does not clear Docker resources at the end (for debugging)

## Testing and Verifying

Tested all manually

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -258,8 +258,9 @@ Additionally, there is an ability to disable certain components of the e2e suite
 
 To get started:
 
-- Run `make test-e2e`
-- Inspect the logs of the docker containers and see if something it’s there
+- Run `make test-e2e-debug` for the full e2e test suite.
+- Run `make test-e2e-short` for the short e2e test suite that excludes upgrade, IBC and state-sync tests.
+- Inspect the logs of the docker containers and see if something is there.
 - `docker ps -a #` to list all docker containers
 - Note the container id of the one you want to see the logs
 - And then run `docker logs <CONTAINER_NAME>`  to debug via container logs
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ LEDGER_ENABLED ?= true
 SDK_PACK := $(shell go list -m github.com/cosmos/cosmos-sdk | sed  's/ /\@/g')
 DOCKER := $(shell which docker)
 BUILDDIR ?= $(CURDIR)/build
+E2E_UPGRADE_VERSION := "v12"
 
 export GO111MODULE = on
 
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -258,8 +258,9 @@ Additionally, there is an ability to disable certain components of the e2e suite
 
 To get started:
 
-- Run `make test-e2e`
-- Inspect the logs of the docker containers and see if something it’s there
+- Run `make test-e2e-debug` for the full e2e test suite.
+- Run `make test-e2e-short` for the short e2e test suite that excludes upgrade, IBC and state-sync tests.
```
ValarDragon, line 262: does `make test-e2e` run everything and clear containers?

ValarDragon (approved):

CONTRIBUTING.md
ValarDragon: answers yes! We should keep that as recommended method in the readme right?

(PR is good to merge either way imo, just wanted to bring this up in case you wanted to change it)

p0mvn: Yes, that's correct.

There are 2 differences between `make test-e2e` and `make test-e2e-debug`:
- The former attempts to clear everything after running 
- The former utilizes Go cache

p0mvn: I think recommending `make test-e2e-debug` might be preferable because it is useful to keep the containers for debugging locally. I also find that it is helpful to avoid using Go cache for local debugging.

Both of them will clear/remove the containers before running the test too

Makefile
```diff
@@ -266,11 +266,28 @@ test-sim-determinism:
...
+# Utilizes Go cache.
+test-e2e: e2e-setup
 	@VERSION=$(VERSION) OSMOSIS_E2E_UPGRADE_VERSION="v12" OSMOSIS_E2E_DEBUG_LOG=True go test -tags e2e -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
 
-test-e2e-short:
-	@VERSION=$(VERSION) OSMOSIS_E2E_SKIP_UPGRADE=True OSMOSIS_E2E_SKIP_IBC=True OSMOSIS_E2E_SKIP_STATE_SYNC=True OSMOSIS_E2E_DEBUG_LOG=True go test -tags e2e -mod=readonly -timeout=25m -v $(PACKAGES_E2E)
+# test-e2e-debug runs a full e2e test suite but does
+# not attempt to delete Docker resources at the end.
+test-e2e-debug: e2e-setup
+	@VERSION=$(VERSION) OSMOSIS_E2E_UPGRADE_VERSION="v12" OSMOSIS_E2E_DEBUG_LOG=True OSMOSIS_E2E_SKIP_CLEANUP=True go test -tags e2e -mod=readonly -timeout=25m -v $(PACKAGES_E2E) -count=1
```
czarcas7ic, line 283: Not for this PR, but now that we are defining the version in multiple places we should probably export this to a single location (and would be awesome if it could just pull the current version and add 1 to it automatically to prevent the need for manually updating this)

czarcas7ic (approved):
This is an awesome quality of life change for anyone who uses the e2e suite, thanks for doing this!

Makefile
p0mvn: Extracted a variable with the version.

I think we can't automate bumping up the version right now because it depends on the presence of the upgrade handler.

In the future, we can code gen the upgrade handler as part of the release process that would allow us to automate the version bump as well

p0mvn:
I found two problems:
- Docker capitalizes tags. As a result, hashes that start from word characters would be capitalized and fail pattern matching
- Stale e2e images would get persisted forever after a new one is built

I attempted to fix both in [b994f54](https://github.com/osmosis-labs/osmosis/pull/2581/commits/b994f5484e39adc86787ba505cf7e5e18cbd145f)

p0mvn:
Tested more on another machine. This should be good to go. If any problems. I'll fix them in the follow-ups

Issue #2509, opened 2022-08-25
Make GAMM queries follow TWAP query code format


ValarDragon:
## Background

An issue that came up in #2508 is that the keeper/grpc_query.go was calling internal pool methods, in ways that it shouldn't. All queries should have been routing through the keeper layer. 

I think we should move to the query model suggested with the querygen tool, even though its incomplete. (Moving all proto wrapping boilerplate out-of-module into the client package, ensuring all of these only run public methods)

## Suggested Design

- [ ] Make grpc_query.go into `query.go`
- [ ] Make query.yml file, and change proto target directory for gamm queries, to be like TWAP
- [ ] For each function in query.go consider:
  - [ ] Deleting it if redundant
  - [ ] Ensure it refers to keeper functions if available

## Acceptance Criteria

- [ ] GRPC overhead + protobuf overhead moved out of `x/gamm/keeper`.

stackman27:
Hi @ValarDragon i can give this a shot if you haven't started on it

ValarDragon:
Please do!

PR #2586, opened 2022-09-03
[QUERY] Make GAMM queries follow TWAP query code format

stackman27:
Closes: #2509

GAMM grpc query was calling internal pool methods, in ways that it shouldn't. All queries should have been routing through the keeper layer.

I think we should move to the query model suggested with the querygen tool, even though its incomplete. (Moving all proto wrapping boilerplate out-of-module into the client package, ensuring all of these only run public methods)

## Brief Changelog
not sure if this is a breaking change, if it is i'll write changelog

## Testing and Verifying
moved grpc_test to its own suite


stackman27:
The reason why this PR got so big was because i had to change the gamm proto definitions to make the grpc-query function consistent with what we had before 
`QuerySwapExactAmountInRequest` -> `QueryEstimateSwapExactAmountInRequest`
`QuerySwapExactAmountOutRequest` -> `QueryEstimateSwapExactAmountOutRequest`

x/gamm/keeper/keeper.go
```diff
@@ -50,7 +50,7 @@ func NewKeeper(cdc codec.BinaryCodec, storeKey sdk.StoreKey, paramSpace paramtyp
 		paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
 	}
 	return Keeper{
-		storeKey:   storeKey,
+		StoreKey:   storeKey,
```
stackman27, line 53: Note: not sure how secure it is to make `StoreKey` public as we need it outside of this package. Would really appreciate any guidance or feedback here

ValarDragon: We definitely can't do this

proto/osmosis/gamm/v1beta1/query.yml
```diff
@@ -0,0 +1,86 @@
...
+  path: "github.com/osmosis-labs/osmosis/v11/x/gamm"
+  struct: "Keeper"
+client_path: "github.com/osmosis-labs/osmosis/v11/x/gamm/client"
+queries:
+  Pool:
+    proto_wrapper:
+      default_values:
+        Req.poolId: "1"
+      query_func_name: "Pool" 
+      query_req_rep_name: "QueryPool"
```
ValarDragon, line 11: I'm a bit confused, what is this difference for?

stackman27: This is for functions like `EstimateSwapExactAmountOut` because its `function_name != funcReq/RespName` 
function name: EstimateSwapExactAmountOut
function response/request: QuerySwapExactAmountOutResponse/ QuerySwapExactAmountOutRequest

The template currently expects the function name to be the same with function response/request but there are functions like `EstimateSwapExactAmountOut` where its not the case. 

Ideally we would want to set a naming standard for that matches with what the template expects

x/gamm/keeper/keeper.go
stackman27: I see, is there an alternative to this then because i need StoreKey in `query_proto_wrap.go` ([LINK](https://github.com/osmosis-labs/osmosis/pull/2586/files#diff-94cc9713a7cdae08c57c24a5bf3ae01c62696eb0c99d4831854e973330f9eab4R65))

mattverse: Oh the pagination requires store key and the store. 
Proposed option here is to create a query method that returns the pagination as well

proto/osmosis/twap/v1beta1/query.yml
```diff
@@ -7,7 +7,7 @@ queries:
     proto_wrapper:
       default_values:
         Req.end_time: "time.Now()"
-      query_func: "k.GetArithmeticTwap"
+      query_func_name: "GetArithmeticTwap"
     cli:
       cmd: "GetArithmeticTwap"
   GetArithmeticTwapToNow:
```
mattverse, line 13: This needs to change as well if we're to change from `query_func` -> `query_func_name`

cmd/querygen/templates/grpcTemplate.go
```diff
@@ -7,14 +7,19 @@ type GrpcTemplate struct {
 }
 
 type GrpcQuery struct {
-	QueryName string
+	QueryFuncName   string
+	QueryReqRepName string
+	KeeperFuncName  string
```
mattverse, line 12: Where is this used?

mattverse (commented):
Queries Added LGTM but we need to figure out store key handling before merging this PR!

Issue #2538, opened 2022-08-29
x/twap: make tests for multi-asset pools


AlpinYukseloglu:
## Background

Currently our TWAP implementation only has tests for two-asset pools. We should expand these tests (especially for `api.go` & `logic.go`) to cover asset pools.

## Suggested Design

- Make 3 asset pool & test TWAP logic against that and/or make the records you might expect for a 3-asset pool and ensure they function as expected in relation to each other

## Acceptance Criteria

- All existing and new tests should pass

PR #2604, opened 2022-09-05
twap: three asset pool tests

czarcas7ic:
Closes: #2538 

This PR mainly adds three asset pools to both `api_test.go` and `logic_test.go`

Changing everything from using types.TwapRecord to instead using a slice of twap records felt a little messy and verbose, but as far as I can tell, it was either doing that or making a copy of every test just for multi asset pools. If there is a better way, please let me know and I can go back and address!

## Brief Changelog

- Change tests to use a slice of twap records in order to iterate over multi asset pools
- Adds three asset pools (referred to as `tap` in the tests for brevity)

## Testing and Verifying


x/twap/api_test.go
```diff
@@ -113,100 +158,201 @@ func (s *TestSuite) TestGetArithmeticTwap() {
...
 		expectError  error
 	}{
 		"(1 record) start and end point to same record": {
 			recordsToSet: []types.TwapRecord{baseRecord},
 			ctxTime:      tPlusOneMin,
 			input:        makeSimpleTwapInput(baseTime, tPlusOne, quoteAssetA),
-			expTwap:      sdk.NewDec(10),
+			expTwap:      []sdk.Dec{sdk.NewDec(10)},
+		},
+		"(1 record, three asset pool) start and end point to same record": {
```
p0mvn, line 171: Re: discussion offline. I think I see what you are saying - this is a new multi-asset test added.

I think I misinterpreted that this is a modified test case because it is still named as "1 record. I suggest renaming this to "3 records at time t".

"1 record" does not really apply because we have 3. They are all just at the same time for multiple denom pairs.

czarcas7ic: Yeah I kept debated changing it to 3 record.. since this was unclear though I will for sure change this now, thanks!

x/twap/api_test.go
```diff
@@ -95,12 +120,34 @@ type getTwapInput struct {
...
+	if isQuoteTokenA {
+		baseAssetDenom, quoteAssetDenom = quoteAssetDenom, baseAssetDenom
+	}
+	twapInput = append(twapInput, getTwapInput{2, quoteAssetDenom, baseAssetDenom, startTime, endTime})
+	quoteAssetDenom, baseAssetDenom = denom0, denom2
+	if isQuoteTokenA {
+		baseAssetDenom, quoteAssetDenom = quoteAssetDenom, baseAssetDenom
+	}
+	twapInput = append(twapInput, getTwapInput{2, quoteAssetDenom, baseAssetDenom, startTime, endTime})
+	quoteAssetDenom, baseAssetDenom = denom1, denom2
```
mattverse, line 145: Is there a reason we're switching these two although either of the denom1 and denom2 are not tokenA? Do you think we should be either creating a secondary parameter for this function isQuoteTokenC or have this as a comment?

x/twap/api_test.go
```diff
@@ -113,100 +160,201 @@ func (s *TestSuite) TestGetArithmeticTwap() {
...
+			expTwap:      []sdk.Dec{sdk.NewDec(3)}, // 5 for 5s, 2 for 10s = 45 / 15 = 3
+		},
+		"(3 pairs of 3 records, three asset pool) interpolate: get twap closer to second record": {
+			recordsToSet: []types.TwapRecord{
+				tapRecordAB, tapRecordAC, tapRecordBC,
+				tPlus10sp5TapRecordAB, tPlus10sp5TapRecordAC, tPlus10sp5TapRecordBC,
+				tPlus20sp2TapRecordAB, tPlus20sp2TapRecordAC, tPlus20sp2TapRecordBC},
+			ctxTime: tPlusOneMin,
+			input:   makeSimpleTapTwapInput(baseTime.Add(15*time.Second), baseTime.Add(30*time.Second), quoteAssetA),
+			expTwap: []sdk.Dec{sdk.NewDec(3), sdk.NewDec(3), sdk.NewDec(3)}, // 5 for 5s, 2 for 10s = 45 / 15 = 3
```
mattverse, line 356: Asking to learn: I noticed that all the test cases for three asset pool have the same values for expTwap, which I don;t understand why in the first glance. Could you explain why we have the same expTwap for the testcases here?

x/twap/api_test.go
```diff
@@ -26,63 +30,84 @@ var (
...
+		"idempotent overwrite2":                             {initStartRecord, recordWithUpdatedAccum(initStartRecord, OneSec, OneSec), tPlusOne, poolId, []string{denomA}, []string{denomB}, nil},
+		"idempotent overwrite2, three asset pool":           {tapInitStartRecord, recordWithUpdatedAccum(tapInitStartRecord, OneSec, OneSec), tPlusOne, tapPoolId, []string{tapDenomA, tapDenomA, tapDenomB}, []string{tapDenomB, tapDenomC, tapDenomC}, nil},
 		"diff spot price": {zeroAccumTenPoint1Record,
 			recordWithUpdatedAccum(zeroAccumTenPoint1Record, OneSec.MulInt64(10), OneSec.QuoInt64(10)),
-			tPlusOne, 1, denomA, denomB, nil},
+			tPlusOne, poolId, []string{denomA}, []string{denomB}, nil},
+		"diff spot price, three asset pool": {tapZeroAccumTenPoint1Record,
+			recordWithUpdatedAccum(tapZeroAccumTenPoint1Record, OneSec.MulInt64(10), OneSec.QuoInt64(10)),
+			tPlusOne, tapPoolId, []string{tapDenomA, tapDenomA, tapDenomB}, []string{tapDenomB, tapDenomC, tapDenomC}, nil},
+		// TODO: Overflow
```
mattverse, line 87:
```suggestion
```

x/twap/api_test.go
```diff
@@ -414,87 +652,133 @@ func (s *TestSuite) TestGetArithmeticTwap_PruningRecordKeepPeriod() {
...
 			input:         makeSimpleTwapToNowInput(baseTime.Add(-time.Hour), quoteAssetA),
 			expectedError: twap.TimeTooOldError{Time: baseTime.Add(-time.Hour)},
 		},
+		"three asset pool, start time too old": {
+			recordsToSet:  []types.TwapRecord{tapRecordAB, tapRecordAC, tapRecordBC},
+			ctxTime:       tPlusOne,
+			input:         makeSimpleTapTwapToNowInput(baseTime.Add(-time.Hour), quoteAssetA),
+			expectedError: twap.TimeTooOldError{Time: baseTime.Add(-time.Hour)},
+		},
+		// TODO: overflow tests
```
mattverse, line 760:
```suggestion
```

mattverse (commented):
Whoa nice job on all these massive testing additions for three asset pool!
I'd like to come back for a second round of review after I have some questions answered included in this review!

x/twap/api_test.go
```diff
@@ -26,63 +30,84 @@ var (
...
+		startRecord []types.TwapRecord
 		// We set it to have the updated time
-		expRecord  types.TwapRecord
+		expRecord  []types.TwapRecord
 		time       time.Time
 		poolId     uint64
-		quoteDenom string
-		baseDenom  string
+		quoteDenom []string
+		baseDenom  []string
```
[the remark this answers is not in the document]
czarcas7ic: Agreed, addressed this with your suggestion! 

Also your comment got me to realize that this is a misnomer; this test case does not take quote and base denoms (where order matters), rather it simply takes denom0 and denom1 (where order technically matters but in reality it doesn't because it gets auto lexicographically reordered). Thanks for drawing my attention to it!

x/twap/api_test.go
czarcas7ic: Ah I see why this is confusing, good point. In my mind I was thinking TokenA as in the first lexicographical token of the pair, but it could also be seen literally as tokenA. Fixed this here, lmk if you like the implementation (its also more customizable which is desirable)

https://github.com/osmosis-labs/osmosis/pull/2604/commits/0d9d165ecdcdc1839e895cf62f9642da957b7177

x/twap/api_test.go
czarcas7ic: So AFAIU, the reason why we have the same results across all three TwapRecords is because we are making the same modifications to pair AB as we are for AB and BC. If we modified the spot price of one of the three pairs differently than the other two, than that spot price would have a differing result. I kept all three records with the same modifications for each record in order to retain my sanity. If we want to test more unique cases (which we definitely should!), we should do so after agreeing that this is/is not the right approach we want for the base TAP test structure.

x/twap/logic_test.go
```diff
@@ -31,18 +31,55 @@ func newRecord(poolId uint64, t time.Time, sp0, accum0, accum1 sdk.Dec) types.Tw
...
 	}
+	twapAC := twapAB
+	twapAC.Asset1Denom = denom2
+	twapBC := twapAC
+	twapBC.Asset0Denom = denom1
+	return []types.TwapRecord{twapAB, twapAC, twapBC}
+}
+
+// make an expected record for math tests, we adjust other values in the test runner.
+func newTapExpRecord(poolId uint64, accum0, accum1 sdk.Dec) []types.TwapRecord {
```
mattverse, line 69: Is there a reason we cannot use newTapRecord? I feel like if we are to adjust other values in the test runner any ways, we can re use `newTapRecord`

x/twap/api_test.go
```diff
@@ -10,12 +10,27 @@ import (
 	"github.com/osmosis-labs/osmosis/v12/x/twap/types"
 )
 
+const (
```
mattverse, line 13: Can we write down comment on why these bool values are defined, and how they are defined?

x/twap/api_test.go
```diff
@@ -26,63 +41,87 @@ var (
...
 	zeroAccumTenPoint1Record := recordWithUpdatedSpotPrice(initStartRecord, sdk.NewDec(10), sdk.NewDecWithPrec(1, 1))
 
+	tapPoolId, tapDenomA, tapDenomB, tapDenomC := s.setupDefaultThreeAssetPool()
+	tapInitStartRecord := newTapRecord(tapPoolId, s.Ctx.BlockTime(), sdk.OneDec(), sdk.ZeroDec(), sdk.ZeroDec())
+	for i := range tapInitStartRecord {
+		tapInitStartRecord[i].Height = s.Ctx.BlockHeight()
+	}
+	tapZeroAccumTenPoint1Record := recordWithUpdatedSpotPrice(tapInitStartRecord, sdk.NewDec(10), sdk.NewDecWithPrec(1, 1))
+
+	denomAB := [][]string{{denomA}, {denomB}}
```
mattverse, line 70: I think two dimensional array can be replaced with a single dimension array here since all the test cases only have one element per array

mattverse (commented):
The test cases look much more cleaner! Thanks for applying the reviews! I still do think that we def need a test case that uses different spot prices amogst the three assets within the pool

Leaving second round of reviews here, I think this PR lgtm after addressing the comments in this review.

ValarDragon:
I don't understand why this is such a big edit

x/twap/api_test.go
czarcas7ic: I think that would be the case if we didn't have to call this as well 

https://github.com/osmosis-labs/osmosis/blob/2114b00d399c78176463707380103b9a66afb11b/x/twap/api_test.go#L73

alexanderbez (approved):

czarcas7ic:
I am significantly simplifying this implementation. Will close this once the new PR is up

Thanks for the reviews, that input will still matter for the new implementation!

Issue #2460, opened 2022-08-19
Add creating a swap post upgrade to e2e tests


ValarDragon:
## Background

We need to add doing a simple swap post upgrade to E2E tests. cref #2429

Issue #2602, opened 2022-09-05
chore: remove twap bindings


p0mvn:
## Background

We would like to encourage using Stargate Query for TWAP. As a result, the bindings can be removed.

## Acceptance Criteria
- TWAP Bindings are removed
- Ensure that stargate query for TWAP is created, tested and whitelisted
- modify e2e test `Test03TWAP` to get twap from the new query

PR #2606, opened 2022-09-05
test(e2e/twap): create a swap post upgrade and twap pruning

p0mvn:
Closes: #2460

Adds e2e test for twap by creating a swap and waiting for an epoch for the records to become older than the twap keep threshold.

Once we expose a query for getting TWAP, it would make sense to integrate it into the e2e test. This will be done in #2602 

## Testing and Verifying


tests/e2e/e2e_test.go
```diff
@@ -160,6 +160,67 @@ func (s *IntegrationTestSuite) TestAddToExistingLock() {
...
+	// Triggers the creation of TWAP records.
+	chainANode.SwapExactAmountIn(coinIn, minAmountOut, fmt.Sprintf("%d", poolId), denomOut, swapWalletAddr)
+
+	keepPeriodCountDown := time.NewTimer(initialization.TWAPPruningKeepPeriod)
+
+	// Make sure still producing blocks.
+	chainA.WaitUntilHeight(heightBeforeSwap + 3)
+
+	if !s.skipUpgrade {
+		// TODO: we should reduce the pruning time in the v11
```
p0mvn, line 203: TODO: make an issue for this

p0mvn: https://github.com/osmosis-labs/osmosis/issues/2607

tests/e2e/configurer/chain/queries.go
```diff
@@ -111,6 +133,36 @@ func (n *NodeConfig) QueryIntermediaryAccount(denom string, valAddr string) (int
 	return intAccountBalance, err
 }
 
+func (n *NodeConfig) QueryCurrentEpoch(identifier string) int64 {
+	// request := epochstypes.QueryCurrentEpochRequest{
+	// 	Identifier: identifier,
+	// }
```
p0mvn, line 139:
```suggestion
```

tests/e2e/configurer/chain/queries.go
```diff
@@ -111,6 +133,36 @@ func (n *NodeConfig) QueryIntermediaryAccount(denom string, valAddr string) (int
...
+
+	bz, err := n.QueryGRPCGateway(path, "identifier", identifier)
+	require.NoError(n.t, err)
+
+	var response epochstypes.QueryCurrentEpochResponse
+	err = util.Cdc.UnmarshalJSON(bz, &response)
+	require.NoError(n.t, err)
+	return response.CurrentEpoch
+}
+
```
p0mvn, line 165:
```suggestion
```

tests/e2e/configurer/chain/commands.go
```diff
@@ -36,6 +36,19 @@ func (n *NodeConfig) CreatePool(poolFile, from string) uint64 {
 	return poolID
 }
 
+// SwapExactAmountIn swaps tokenInCoint to get at least tokenOutMinAmountInt of the other token's pool out.
```
czarcas7ic, line 39:
```suggestion
// SwapExactAmountIn swaps tokenInCoin to get at least tokenOutMinAmountInt of the other token's pool out.
```

tests/e2e/e2e_test.go
```diff
@@ -60,7 +60,7 @@ func (s *IntegrationTestSuite) Test02CreatePool() {
 // - voting yes on the proposal from the validator wallet
 // - voting no on the proposal from the delegator wallet
 // - ensuring that delegator's wallet overwrites the validator's vote
-func (s *IntegrationTestSuite) TestSuperfluidVoting() {
+func (s *IntegrationTestSuite) Test03SuperfluidVoting() {
```
czarcas7ic, line 63: Superfluid Voting should no longer be required to be ran in a specific order

tests/e2e/e2e_test.go
```diff
@@ -126,8 +126,8 @@ func (s *IntegrationTestSuite) TestSuperfluidVoting() {
 	)
 }
 
-// TestAddToExistingLockPostUpgrade ensures addToExistingLock works for locks created preupgrade.
-func (s *IntegrationTestSuite) TestAddToExistingLockPostUpgrade() {
+// Test03AddToExistingLockPostUpgrade ensures addToExistingLock works for locks created preupgrade.
+func (s *IntegrationTestSuite) Test03AddToExistingLockPostUpgrade() {
```
czarcas7ic, line 130: Same with this one, it shouldn't need a number appended to it

tests/e2e/e2e_test.go
```diff
@@ -141,8 +141,8 @@ func (s *IntegrationTestSuite) TestAddToExistingLockPostUpgrade() {
 	chainANode.AddToExistingLock(sdk.NewInt(1000000000000000000), "gamm/pool/1", "240s", lockupWalletSuperfluidAddr)
 }
 
-// TestAddToExistingLock tests lockups to both regular and superfluid locks.
-func (s *IntegrationTestSuite) TestAddToExistingLock() {
+// Test03AddToExistingLock tests lockups to both regular and superfluid locks.
+func (s *IntegrationTestSuite) Test03AddToExistingLock() {
```
czarcas7ic, line 145: Same with this one

tests/e2e/e2e_test.go
```diff
@@ -160,6 +160,67 @@ func (s *IntegrationTestSuite) TestAddToExistingLock() {
...
 
+// Test03TWAP tests TWAP by creating a pool, performing a swap.
+// These two operations should create TWAP records.
+// Then, we wait until the epoch for the records to be pruned.
+// The records are guranteed to be pruned at the next epoch
+// because twap keep time = epoch time / 4 and we use a timer
+// to wait for at least the twap keep time.
+// TODO: implement querying for TWAP, once such queries are exposed:
+// https://github.com/osmosis-labs/osmosis/issues/2602
+func (s *IntegrationTestSuite) Test03TWAP() {
```
czarcas7ic, line 171: My guess is you purposefully are using 03 as the catch-all for tests that don't require a specific order. Imo its less confusing to just leave the ones that don't require an order to be without numbers, but I don't feel strongly about this.

tests/e2e/e2e_test.go
```diff
@@ -160,6 +160,67 @@ func (s *IntegrationTestSuite) TestAddToExistingLock() {
...
+
+	if !s.skipUpgrade {
+		// TODO: we should reduce the pruning time in the v11
+		// genesis to make this test run faster
+		// Currenty, we are testing the upgrade from v11 to v12,
+		// the pruning time is set to whatever is in the upgrade
+		// handler (two days). Therefore, we cannot reasonably
+		// test twap pruning post-upgrade.
+		s.T().Skip("skipping TWAP Pruning test. This can be re-enabled post v12")
+	}
```
czarcas7ic, line 210: Should we just keep this at the top of the test for now, since we have shown that this will work up to this point? Only because this will add a little bit of time to CI without actually testing twap

czarcas7ic (approved):
Only minor comments, nice work! 👍

tests/e2e/e2e_test.go
p0mvn: We should order these to be run after pool creation. Otherwise, it does not make sense having that test. If pool creation fails, we cannot rn any other tests since they all create pools

tests/e2e/e2e_test.go
p0mvn: Same reason here

tests/e2e/e2e_test.go
p0mvn: It tests twap by creating a pool and a swap. This is the core of the test - making sure that post-swap TWAP logic does not break

tests/e2e/e2e_test.go
ValarDragon: Err, you can't rely on test ordering being lexicographic ordering. But I also don't think we should be have test ordering based on naming, and should be specified elsewhere

p0mvn: We don't rely on this order. If the order was changed, the tests would still run because they are independent. 

The reason for this lexicographic naming is to be able to isolate 'TesteCreatePool' conveniently and run it first because we create pools in the majority of other tests.

If 'TestCreatePool' were faulty, we would cancel the remaining tests.

If we don't have 'TestCreatePool' run first, I think we might not need a dedicated fixture for testing pool creation because it would be covered by other flows.

Let me know what you think. If no specific opinions, I can remove the lexicographic order and consider removing 'TestCreatePool'

p0mvn: Ordering is removed

p0mvn:
Planning to merge this to unblock additional TWAP tests that are coming in the next PR in a few minutes

PR #2614, opened 2022-09-06
Revert "chore: Update Proto + ~~ORM Support~~ (#1589)"

mattverse:
This reverts commit f291e12d5f4f092ea2ccb4ef4008a2e3ba21ac47.

Some of the changes that were made in #1589 were changes that we wanted to keep(how we use buf gen to protogen, for example), making the scope of this PR not only simply revert the commit, but keep some of the updated process for proto gen that were initially included in #1589


p0mvn:
Let's wait with this PR please until the new `v12.x` is cut please

Ref: https://github.com/osmosis-labs/osmosis/pull/2612

scripts/protocgen.sh
```diff
@@ -32,4 +32,4 @@ go mod tidy -compat=1.18
 # TODO: Uncomment once ORM/Pulsar support is needed.
 #
 # Ref: https://github.com/osmosis-labs/osmosis/pull/1589
-# ./scripts/protocgen2.sh
+# ./scripts/protocgen2.sh
```
p0mvn, line 35: Should this be removed?
```suggestion
```

p0mvn (approved):

scripts/protocgen.sh
mattverse: yes! Thanks for the catch

Issue #2343, opened 2022-08-10
x/gamm: Consider unit testing balancer implementations of `JoinPool` and `JoinPoolNoSwap`


AlpinYukseloglu:
In reference to @p0mvn's comment here: https://github.com/osmosis-labs/osmosis/pull/2133#discussion_r940472270

`JoinPool` and `JoinPoolNoSwap` are currently only tested at the keeper level, but the balancer implementation is not directly unit tested. The claim is that this might lead to untested cases since it's more difficult to test many cases by testing and making assertions against a caller.

Given how `JoinPool` and `JoinPoolNoSwap` both pretty much just have two internal functions (both of which are heavily tested in balancer), unit testing them would pretty much mean copy-pasting all of those tests and checking if liquidity properly updates. I'm not confident this achieves much, but I wanted to have a place for us to discuss since we had a thread going in the PR above.

PR #2616, opened 2022-09-06
x/gamm: add direct unit tests for balancer `JoinPool` and `JoinPoolNoSwap`

AlpinYukseloglu:
Closes: #2343

While every component of our balancer `JoinPool` was thoroughly tested, we didn't have _direct_ unit tests for it. This PR adds these unit tests with minimal repeat code.

Note: unit tests for `JoinPoolNoSwap` will have to be added in a separate PR after #2133 is merged

## Brief Changelog

- Factor out test cases for multi-asset pools and edge cases from `TestCalcJoinPoolShares` into the `joinPoolTestCases` variable
- Add `JoinPool` unit tests running on the core join pool test cases (`calcSingleAssetJoinTestCases`) as well as the ones defined in the previous point

## Testing and Verifying

This change added `TestJoinPool` to `x/gamm/pool-models/balancer/pool_suite_test.go`


x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -629,169 +788,12 @@ func (suite *KeeperTestSuite) TestBalancerSpotPriceBounds() {
...
-			tokensIn: sdk.NewCoins(
-				sdk.NewInt64Coin("uosmo", 1),
-				sdk.NewInt64Coin("uatom", 1),
-			),
-			expectShares: sdk.NewInt(100_000_000),
-		},
-	}
-	testCases = append(testCases, calcSingleAssetJoinTestCases...)
+
+	testCases := append(joinPoolTestCases, calcSingleAssetJoinTestCases...)
```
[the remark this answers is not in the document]
AlpinYukseloglu: The former tests multi-assets inputs and edge cases, while the latter only tests single asset inputs. 

Splitting them up like this allows us to minimize repeated test cases as not all cases are applicable to all functions we're testing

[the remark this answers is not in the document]
mattverse: wdyt about splitting joinPoolTestCases into multi asset inputs and edge cases?
I personally think separating these out into two different testing varialbes would be a cleaner approach + less confusing when we try to use these test cases in other cases in the future, since each var would have a sole and clear purpose. lmk what you think!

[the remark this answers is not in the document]
AlpinYukseloglu: Since all the edge case tests in `joinPoolTestCases` are multi-asset I just renamed it to `multiAssetInputTestCases` – does this address your point?

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -823,25 +953,23 @@ func (suite *KeeperTestSuite) TestCalcJoinPoolShares() {
 
 			// system under test
 			sut := func() {
-				shares, liquidity, err := pool.CalcJoinPoolShares(suite.Ctx, tc.tokensIn, tc.swapFee)
+				preJoinAssets := pool.GetTotalPoolLiquidity(suite.Ctx)
+				shares, err := pool.JoinPoolNoSwap(suite.Ctx, tc.tokensIn, tc.swapFee)
```
p0mvn, line 957: q: if the caller provides uneven amounts to `JoinPoolNoSwap` such as `25000` and `50000`, how can the caller tell how many tokens were joined and how many are remaining? We might want to consider having an extra return value to represent either the "tokens joined" or "remaining tokens".

If you agree, let's make an issue for this please

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -745,75 +936,14 @@ func (suite *KeeperTestSuite) TestCalcJoinPoolShares() {
...
-					Weight: sdk.NewInt(100),
-				},
-			},
-			tokensIn: sdk.NewCoins(
-				sdk.NewInt64Coin("foo", 1),
-				sdk.NewInt64Coin("baz", 1),
+			expectShares: sdk.NewInt(1250000000000),
+			expectLiq: sdk.NewCoins(
+				sdk.NewInt64Coin("uosmo", 25_000),
+				sdk.NewInt64Coin("uatom", 12_500),
```
p0mvn, line 942: From my understanding, the ratio of tokens in relative to the pool amount is what matters for the core of the calculation in `MaximalExactRatioJoin` while the weights are irrelevant.

I know you've mentioned the "ratios" part already in the comment but can we elaborate on this please and show the calculation steps? I think it would be useful to show that the weights have no effect while the ratios do. It would help ease the onboarding on the tests.

Based on my understanding of the calculation:
```
initialShareAmount = 1 * 10**18 * 100

expectShares = 25000 / 2_000_000_000_000 * initialShareAmount = 1250000000000.0

uosmo expected liq = 25_000 because it has the minimal requested token in / current supply ration

uatom = minRatio  * current uatom supply = 25000 / 2_000_000_000_000 * 1_000_000_000_000 = 12_500
```

p0mvn (approved):
Great work on the tests!

I only have one minor request to add a comment and a suggestion to potentially have an additional return value from `JoinPoolNoSwap`.

x/gamm/pool-models/balancer/pool_suite_test.go
AlpinYukseloglu: Great point – please see my recent change. I added brief comments explaining the dynamic (ratio matters but weights don't) and how the exact results were obtained

x/gamm/pool-models/balancer/pool_suite_test.go
AlpinYukseloglu: Yeah good point – I don't think this is too critical of a change given how `JoinPoolNoSwap` _does_ return tokens joined at the keeper level, but I think it currently does so by taking a bit of a shortcut so I think this is worth adding for code hygiene

x/gamm/pool-models/balancer/pool_suite_test.go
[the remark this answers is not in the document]
mattverse: Yeah this sgtm!

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -629,109 +806,125 @@ func (suite *KeeperTestSuite) TestBalancerSpotPriceBounds() {
 
 func (suite *KeeperTestSuite) TestCalcJoinPoolShares() {
 	// We append shared calcSingleAssetJoinTestCases with multi-asset and edge
-	// test cases.
+	// test cases defined in multiAssetExactInputTestCases and multiAssetUneverInputTestCases.
+	//
+	// See calcJoinSharesTestCase struct definition for explanation why the
+	// sharing is needed.
+	testCases := append(append(multiAssetExactInputTestCases, multiAssetUnevenInputTestCases...), calcSingleAssetJoinTestCases...)
```
mattverse, line 813: Append can take multiple params! (The suggestion is from my a web editor need to double check on the code itself
```suggestion
	testCases := append(multiAssetExactInputTestCases, multiAssetUnevenInputTestCases..., calcSingleAssetJoinTestCases...)
```

mattverse (approved):
Nice work on adding `multiAssetExactInputTestCases`, let's move foward in adding these tests cases to different parts of Balancer pools uint test. 

LGTM!

x/gamm/pool-models/balancer/pool_suite_test.go
mattverse: Noticing that we can't append multiple arrays using `append`, resolving this review

Issue #2597, opened 2022-09-05
Add TWAP query for GetArithmeticTwapToNow


ValarDragon:
## Background

As title

PR #2618, opened 2022-09-06
Add Twap GetArithmeticToNow query

ValarDragon:
Closes: #2597 

* Adds GetArithmeticTwapToNow query 
* fixes proto import error in query proto file
* creates proto response struct even if error, as desired

## Testing and Verifying

### DIFF
--- a/proto/osmosis/twap/v1beta1/query.proto
+++ b/proto/osmosis/twap/v1beta1/query.proto
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package osmosis.gamm.twap.v1beta1;
+package osmosis.twap.v1beta1;
 
 import "gogoproto/gogo.proto";
 import "osmosis/twap/v1beta1/twap_record.proto";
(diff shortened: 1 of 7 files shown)


ValarDragon:
Yeah all validation should imo be handled by the go method, so all of:
* Go API
* Cli API
* grpc API

is consistent

mattverse (approved):
I think we need more basic validation catching logic for `GetArithmeticTwapToNow` in the future, although we would still error within the logic for invalid request parameters. I think we need this because we don't want to return the same error for invalid query parameters(ex. having input invalid parameter for time for `GetArithmeticTwapToNow` would now return KeySeparatorLengthError while I think we should be returning a different error for this case). lmk what you think about this

Otherwise query logic lgtm!

proto/osmosis/twap/v1beta1/query.proto
```diff
@@ -16,7 +16,12 @@ option go_package = "github.com/osmosis-labs/osmosis/v11/x/twap/client/queryprot
 service Query {
   rpc GetArithmeticTwap(GetArithmeticTwapRequest)
       returns (GetArithmeticTwapResponse) {
-    option (google.api.http).get = "/osmosis/gamm/v1beta1/pools";
+    option (google.api.http).get = "/osmosis/twap/v1beta1/GetArithmeticTwap";
+  }
+  rpc GetArithmeticTwapToNow(GetArithmeticTwapToNowRequest)
```
stackman27, line 21: Sorry late to this PR, but since we're using a template to generate grpc queries, i think it'd be better if we have a naming standard for requests and responses. 

For instance, Gamm queries start with "query" like `QueryPoolsRequest` `QueryTotalLiquidityRequest` `QuerySpotPriceResponse` 

However, Twap queries doesnot contain "query" prefix like `GetArithmeticTwapToNowRequest`

PR #2632, opened 2022-09-07
Fix: Add proto reset to Stargate Querier

mattverse:

Current Stargate querier implementation would not proto unmarshal & json marshall as expected, if the query includes a repeated field, the query simply appending on top of the existing struct. This way each time we parse the protobuf data we are simply appending to the existing array, thus having to need to zero it out every time we load. 

This PR solves the existing problem by resetting the message every time we call a query request. 

## Brief Changelog

- Reset proto message every time query is called

## Testing and Verifying

Adds a test case that includes query with iteration

Special thanks to @ethanfrey for helping with the insights here!

### DIFF
--- a/wasmbinding/query_plugin_test.go
+++ b/wasmbinding/query_plugin_test.go
@@ -226,6 +226,50 @@ func (suite *StargateTestSuite) TestDeterministicJsonMarshal() {
 		responseProtoStruct interface{}
 		expectedProto       func() proto.Message
 	}{
+
+		/**
+		   * Origin Response
+		   * balances:<denom:"bar" amount:"30" > pagination:<next_key:"foo" >
+		   * "0a090a036261721202333012050a03666f6f"
+		   *
+		   * New Version Response
+		   * The binary built from the proto response with additional field address
+		   * balances:<denom:"bar" amount:"30" > pagination:<next_key:"foo" > address:"cosmos1j6j5tsquq2jlw2af7l3xekyaq7zg4l8jsufu78"
+		   * "0a090a036261721202333012050a03666f6f1a2d636f736d6f73316a366a357473717571326a6c77326166376c3378656b796171377a67346c386a737566753738"
+		   // Origin proto
+		   message QueryAllBalancesResponse {
+		  	// balances is the balances of all the coins.
+		  	repeated cosmos.base.v1beta1.Coin balances = 1
+		  	[(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+		  	// pagination defines the pagination in the response.
+		  	cosmos.base.query.v1beta1.PageResponse pagination = 2;
+		  }
+		  // Updated proto
+		  message QueryAllBalancesResponse {
+		  	// balances is the balances of all the coins.
+		  	repeated cosmos.base.v1beta1.Coin balances = 1
+		  	[(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+		  	// pagination defines the pagination in the response.
+		  	cosmos.base.query.v1beta1.PageResponse pagination = 2;
+		  	// address is the address to query all balances for.
+		  	string address = 3;
+		  }
+		*/
+		{
+			"Query All Balances",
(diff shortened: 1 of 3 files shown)


wasmbinding/stargate_whitelist.go
```diff
@@ -36,6 +36,7 @@ func init() {
 
 	// bank
 	StargateWhitelist.Store("/cosmos.bank.v1beta1.Query/Balance", &banktypes.QueryBalanceResponse{})
+	StargateWhitelist.Store("/cosmos.bank.v1beta1.Query/AllBalances", &banktypes.QueryAllBalancesResponse{})
```
ValarDragon, line 39: I don't think we should put this in our whitelist or encourage people to use it.

mattverse: kk makes sense, moving this to test case set up

ValarDragon (approved):

wasmbinding/query_plugin_test.go
```diff
@@ -220,12 +220,60 @@ func (suite *StargateTestSuite) TestConvertProtoToJsonMarshal() {
...
+		  	string address = 3;
+		  }
+		*/
+		{
+			"Query All Balances",
+			func() {
+				wasmbinding.StargateWhitelist.Store("/cosmos.bank.v1beta1.Query/AllBalances", &banktypes.QueryAllBalancesResponse{})
+			},
+			"0a090a036261721202333012050a03666f6f",
+			"0a090a036261721202333012050a03666f6f1a2d636f736d6f73316a366a357473717571326a6c77326166376c3378656b796171377a67346c386a737566753738",
```
ValarDragon, line 265: wait just realized, why is this JSON getting hex encoded?

ValarDragon: wait no, what is this encoding. Hex decoding gives
```
\n\t\n\x03bar\x12\x0230\x12\x05\n\x03foo\x1a-cosmos1j6j5tsquq2jlw2af7l3xekyaq7zg4l8jsufu78
```
which uhh, does not make any sense to me. Why do we have so many `\x03`, and why do we start `\n\t\n` ?

iboss-ptk:
@mattverse  Since the initial problem we found is accumulated iterable (repeat in proto speak) query response, I think we should add test cases that address this problem as well.

iboss-ptk:
for:
> Adds a test case that includes query with iteration

I think we should add "running the same query that returns iterable(s) multiple times should not accumulate" into the cases since that's where the original problem occurs.

mattverse:
> @mattverse for:
> 
> > Adds a test case that includes query with iteration
> 
> I think we should add "running the same query that returns iterable(s) multiple times should not accumulate" into the cases since that's where the original problem occurs.

Good Catch! Just added some more test cases on this, re-requesting for your review on the added test cases

wasmbinding/query_plugin_test.go
mattverse: This is a hex encoded version of the response bytes that we would get from the GrpcQueryRouter

mattverse: how did you get `\n\t\n\x03bar\x12\x0230\x12\x05\n\x03foo\x1a-cosmos1j6j5tsquq2jlw2af7l3xekyaq7zg4l8jsufu78`?

Directly hex decoding the given string would give me something like [10 9 10 3 98 97 114 18 2 51 48 18 5 10 3 102 111 111], which if I attempt to convert it to string directly, would give me `bar30 foo` (which is a wrong conversion of string )

ValarDragon: The following python code:
```
>>> hex_s = "0a090a036261721202333012050a03666f6f1a2d636f736d6f73316a366a357473717571326a6c77326166376c3378656b796171377a67346c386a737566753738"
>>> bytes.fromhex(hex_s)
b'\n\t\n\x03bar\x12\x0230\x12\x05\n\x03foo\x1a-cosmos1j6j5tsquq2jlw2af7l3xekyaq7zg4l8jsufu78'
>>> chr(0x0a) # spot check that the first byte is translated correctly
'\n'
```

ValarDragon:
Whats the corresponding code to deserialize this response cosmwasm side?

ValarDragon:
Merging as this PR is clearly good, but I do still have questions pre-release

PR #2635, opened 2022-09-07
TWAP notes/fixes from code review

ValarDragon:
## What is the purpose of the change

Notes / fixes from the live TWAP code review

Two important bugs found during code review! 

* Time.Now() bug in query that was getting vendored (fixed in this PR)
  * Thankfully wouldn't have caused mainnet issue, since exec time always greater than block time, and we've well defined returning an error in this case
* Bug in pruning logic (not yet fixed)

## Testing and Verifying

Mostly comment + TODO adds

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #2638, opened 2022-09-07
test(e2e/twap): twap query before/after swap and after pruning

p0mvn:
Closes: #XXX

Basic e2e testing of TWAP query before/after swap and after pruning

## Testing and Verifying


tests/e2e/e2e_test.go
```diff
@@ -202,6 +213,11 @@ func (s *IntegrationTestSuite) TestTWAP() {
 		s.T().Logf("Current epoch number is (%d), waiting to reach (%d)", newEpochNumber, oldEpochNumber+1)
 		return newEpochNumber > oldEpochNumber
 	})
+
+	// We should still be able to get the TWAP after pruning.
+	// It should not change from the post-swap TWAP.
+	twapToNowPostPruning := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, time.Now())
+	s.Require().Equal(twapToNowPostPruning, twapToNowPostSwap)
```
p0mvn, line 220: I'm currently trying to figure out why these would not be equal.

My thinking is that they should still be since they are supposed to utilize the same start record

p0mvn: Ah right this should be addressed in https://github.com/osmosis-labs/osmosis/pull/2635

ValarDragon: Yeah, we have a bug if they're not equal. (Though this line uses time.Now() right above)

p0mvn: You are right, `time.Now()` was the problem, thanks. The new check you added exposed the invalid setup. Then, making time be at fixed intervals from queried block times helped fix this.

All tests are passing right now. I'm going to make a 3-asset pool next to try and repro the bug found in the review

tests/e2e/configurer/chain/chain.go
```diff
@@ -80,9 +80,29 @@ func (c *Config) RemoveNode(nodeName string) error {
...
+	require.NoError(c.t, err)
+	err = node.WaitUntil(func(_ coretypes.SyncInfo) bool {
+		newEpochNumber := node.QueryCurrentEpoch(epochIdentifier)
+		c.t.Logf("current epoch number is (%d), waiting to reach (%d)", newEpochNumber, epoch)
+		return newEpochNumber >= epoch
+	})
+	require.NoError(c.t, err)
+}
+
+// WaitFoeEpochs waits for the chain to to go through a given number of epochs.
```
p0mvn, line 95:
```suggestion
// WaitForEpochs waits for the chain to go through a given number of epochs.
```

tests/e2e/e2e_test.go
```diff
@@ -195,13 +234,29 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+	s.Require().Equal(twapFromAfterToAfterSwapAndBeforePruning, twapFromAfterToAfterSwapAndAfterPruning)
+
+	// TWAP "from after to after swap" should equal to "from after swap to after pruning"
+	// That is because they are suppossed to use the same start and end records.
+	// The fact that these are taken over different time periods is irrelevant.
+	timeAfterPruning := chainANode.QueryLatestBlockTime()
+	s.T().Log("querying for TWAP from after swap to after pruning")
+	twapToNowPostPruning, err := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, timeAfterSwap, timeAfterPruning)
+	s.Require().NoError(err)
+	s.Require().Equal(twapToNowPostPruning, twapFromAfterToAfterSwapAndBeforePruning)
```
p0mvn, line 261: This test sometimes fails and sometimes passes. However, I've verified through logs that it's grabbing the records from the desired heights that have correct spot prices.

The problem is likely coming from either time deltas or accumulator calculations when interpolating records.

@ValarDragon how do you think I should go about this? Is it worth investigating further or can we attribute this to setup/rounding errors? 

I might also be doing something wrong. Spent hours going in circles trying to root cause what exactly

tests/e2e/e2e_test.go
```diff
@@ -171,16 +169,59 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+	twapFromAfterToNow, err := chainANode.QueryGetArithmeticTwapToNow(poolId, denomOne, denomTwo, timeAfterSwap)
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToNow)
+
+	s.T().Log("querying for the TWAP from after swap to after swap + 10ms, must not equal to TWAPs before swap")
+	twapFromAfterToAfterSwapAndBeforePruning, err := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, timeAfterSwap, timeAfterSwap.Add(10*time.Millisecond))
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToAfterSwapAndBeforePruning)
+
+	s.Require().Equal(twapFromAfterToAfterSwapAndBeforePruning, twapFromAfterToNow)
```
p0mvn, line 224: This check is the same. Fails sometimes.

The records are as expected, only accumulators and end times are different

tests/e2e/e2e_test.go
```diff
@@ -171,16 +173,61 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToNow)
+
+	s.T().Log("querying for the TWAP from after swap to after swap + 10ms, must not equal to TWAPs before swap")
+	twapFromAfterToAfterSwapAndBeforePruning, err := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, timeAfterSwap, timeAfterSwap.Add(10*time.Millisecond))
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToAfterSwapAndBeforePruning)
+
+	// These must be equal because they are calculated over time ranges with the stable and equal spot price.
+	// There are potential rounding errors requiring us to approximate the comparison.
+	osmoassert.DecApproxEq(s.T(), twapFromAfterToAfterSwapAndBeforePruning, twapFromAfterToNow, sdk.NewDecWithPrec(1, 3))
```
p0mvn, line 230: Note to reviewers:

I've been debugging this with extra logs added.

I observed that records are picked up from the correct heights and with the correct spot prices.

I also logged time deltas and accumulator differences from:
https://github.com/osmosis-labs/osmosis/blob/7a2db79f406adfd3a5099e4acb1b1936b6e25c06/x/twap/types/utils.go#L55-L56

It seems that these are not exactly equal in cases when the time deltas are different. So I suspect that these are rounding issues caused by truncations.

Another potential source is sig fig round for spot prices. I attempted to increase precision by 4 sig figs for testing purposes. Although it reduced the delta, it did not completely eliminate it.

tests/e2e/configurer/chain/node.go
```diff
@@ -92,22 +92,20 @@ func (n *NodeConfig) Stop() error {
...
 		// let the node produce a few blocks
 		if !doneCondition(status.SyncInfo) {
 			time.Sleep(waitUntilRepeatPauseTime)
 			continue
 		}
-		return nil
+		return
 	}
-	return fmt.Errorf("node %s timed out waiting for condition, latest block height was %d", n.Name, latestBlockHeight)
+	n.t.Errorf("node %s timed out waiting for condition, latest block height was %d", n.Name, latestBlockHeight)
```
ValarDragon, line 108: Not blocking this PR, but generally confused about direction to not return an error here to flag in calling test, vs logging an error.

tests/e2e/e2e_test.go
```diff
@@ -171,16 +173,61 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+	// because the former has a higher time weight for the after swap period.
+	s.T().Log("querying for the TWAP from after swap to now, must not equal to TWAPs before swap")
+	twapFromAfterToNow, err := chainANode.QueryGetArithmeticTwapToNow(poolId, denomOne, denomTwo, timeAfterSwap)
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToNow)
+
+	s.T().Log("querying for the TWAP from after swap to after swap + 10ms, must not equal to TWAPs before swap")
+	twapFromAfterToAfterSwapAndBeforePruning, err := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, timeAfterSwap, timeAfterSwap.Add(10*time.Millisecond))
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToAfterSwap, twapFromAfterToAfterSwapAndBeforePruning)
```
czarcas7ic, line 226: I might be missing something but this check seems redundant, i.e. if the previous check passes there really should be no logical way this would fail. What do you think?

tests/e2e/e2e_test.go
```diff
@@ -171,16 +173,61 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+	// when we query for TWAP.
+	timeAfterSwap := chainANode.QueryLatestBlockTime()
+	chainA.WaitForNumHeights(2)
+
+	// TWAP "from before to after swap" should be different from "from before to before swap"
+	// because swap introduces a new record with a different spot price.
+	s.T().Log("querying for the TWAP from before swap to now after swap, must not equal to TWAPs before swap")
+	twapFromBeforeSwapToAfterSwap, err := chainANode.QueryGetArithmeticTwapToNow(poolId, denomOne, denomTwo, timeBeforeSwap)
+	s.Require().NoError(err)
+	s.Require().NotEqual(twapFromBeforeSwapToBeforeSwapOne, twapFromBeforeSwapToAfterSwap)
```
czarcas7ic, line 214: While I don't think we should be checking for exact values, I do think NotEqual might be slightly too broad here (for example, a nil record is not equal but probably not what we want to see). I think a safe middle ground would just be doing some napkin math and at least asserting one should be GT or LT the other. Thoughts?

tests/e2e/e2e_test.go
```diff
@@ -195,13 +242,31 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+
+	// Epoch end triggers the prunning of TWAP records.
+	// Records before swap should be pruned.
+	chainA.WaitForNumEpochs(1, epochIdentifier)
+
+	// We should not be able to get TWAP before swap since it should have been pruned.
+	s.T().Log("pruning is now complete, querying TWAP for period that should be pruned")
+	_, err = chainANode.QueryGetArithmeticTwapToNow(poolId, denomOne, denomTwo, timeBeforeSwap)
+	s.Require().Error(err)
+	s.Require().ErrorContains(err, "too old")
```
czarcas7ic, line 254:
```suggestion
	s.Require().ErrorContains(err, "too old")
```
nit: might be redundant

tests/e2e/e2e_test.go
```diff
@@ -195,13 +242,31 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
+
+	// TWAPs for the same time range should be the same when we query for them before and after pruning.
+	s.T().Log("querying for TWAP for period before pruning took place but should not have been pruned")
+	twapFromAfterToAfterSwapAndAfterPruning, err := chainANode.QueryGetArithmeticTwap(poolId, denomOne, denomTwo, timeAfterSwap, timeAfterSwap.Add(10*time.Millisecond))
+	s.Require().NoError(err)
+	s.Require().Equal(twapFromAfterToAfterSwapAndBeforePruning, twapFromAfterToAfterSwapAndAfterPruning)
```
czarcas7ic, line 260: I think both of these records are before pruning right? The comment and var name implies you are trying to test one before and one after pruning but I feel like they are both pre pruning.

czarcas7ic (approved):
Well done on this! I had to reread some of these quite a few times to ensure I was tracking, but I think the way you named them and the comments you added made it as clear as possible. 

After these comments are addressed I think its good to merge 👍

tests/e2e/configurer/chain/node.go
p0mvn: `Errorf` does not only log but also fails immediately.

The reason for changing this is to reduce verbosity and simplify tests.

For the majority of these methods, we fail if the error is non-nil anyways. However, constantly doing the error checks reduces the readability of the tests.

Failing immediately via this `testing` package method reduces verbosity and makes it easier to write tests

tests/e2e/e2e_test.go
p0mvn: In my opinion, the exact math should be tested at a lower level of abstraction. In e2e, we are just checking that system components work well together without diving deep into the calculation details.

Also, this is not 100% deterministic since this is twap to now where blocks and their respective time might slightly vary between runs

tests/e2e/e2e_test.go
p0mvn: This is `QueryGetArithmeticTwap` while previous one is `QueryGetArithmeticTwapToNow`. The goal here was to test that both are functioning as expected

tests/e2e/e2e_test.go
p0mvn: - `twapFromAfterToAfterSwapAndAfterPruning` is after pruning
   * renamed to `twapAfterPruning10ms`
- `twapFromAfterToAfterSwapAndBeforePruning` is before pruning
   * renamed to `twapAfterSwapBeforePruning10Ms`

renamed both for clarity. Let me know what you think

tests/e2e/e2e_test.go
czarcas7ic: Ah I didn't notice that, this makes sense then

tests/e2e/e2e_test.go
czarcas7ic: This is clearer to me, thanks!

tests/e2e/e2e_test.go
czarcas7ic: I agree that the exact math should not be tested here. I guess what I am saying is `NotEqual` will return True if the twap record we retrieve is nil. We should for sure know which twap (before or after swap) is greater than the other, so a more specific `s.Require().True(twapFromBeforeSwapToBeforeSwapOne.GT(twapFromBeforeSwapToAfterSwap))` (or whatever the properly verbiage is) might be more desirable. 

This is not PR blocking at all, just a suggestion that may save us from missing issues in the future.

tests/e2e/e2e_test.go
```diff
@@ -171,16 +173,61 @@ func (s *IntegrationTestSuite) TestTWAP() {
...
-	heightBeforeSwap := chainANode.QueryCurrentHeight()
 
-	// Triggers the creation of TWAP records.
-	chainANode.SwapExactAmountIn(coinIn, minAmountOut, fmt.Sprintf("%d", poolId), denomOut, swapWalletAddr)
+	s.T().Log("querying for the second TWAP to now before swap, must equal to first")
+	twapFromBeforeSwapToBeforeSwapTwo, err := chainANode.QueryArithmeticTwapToNow(poolId, denomOne, denomTwo, timeBeforeSwap.Add(50*time.Millisecond))
+	s.Require().NoError(err)
+
+	// Since there were no swaps between the two queries, the TWAPs should be the same.
+	s.Require().Equal(twapFromBeforeSwapToBeforeSwapOne, twapFromBeforeSwapToBeforeSwapTwo)
```
ValarDragon, line 192: This may not be reliable in edge cases due to rounding concerns

ValarDragon (approved):

tests/e2e/e2e_test.go
p0mvn: I haven't seen this one failing but I will add the approximate equality check just in case

tests/e2e/e2e_test.go
p0mvn: Will try to address this before merging. Marking this as draft for now

PR #2641, opened 2022-09-07
TWAP: Simplify historical pool index

ValarDragon:

This is a minor performance improvement, but more importantly a complexity reduction for getting entries from the pool index. Do we want to go with this approach?

It came up on the TWAP review call. I am not sure that it is worth simplifying because:
* No known bug here
* Performance improvement isn't really a big deal (reduces gas usage of TWAPs, but should not be a huge deal)
* Removes +1ns hack (though we could also remove that via other means)

Depending on folks thoughts (cc @p0mvn @czarcas7ic @AlpinYukseloglu ), we should decide whether this is a helpful simplfiication or not.

## Brief Changelog

- Simplify method of getting records in all TWAP queries
- Improve gas efficiency for multi-asset pool TWAP queries (not a usecase we care about optimizing)

## Testing and Verifying

This change is already covered by existing tests (except key suffix), including the osmoutils package change.

I can add tests for that if we agree that this is worth doing at all.

### DIFF
--- a/x/twap/api_test.go
+++ b/x/twap/api_test.go
@@ -2,7 +2,6 @@ package twap_test
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -75,9 +74,7 @@ func (s *TestSuite) TestGetBeginBlockAccumulatorRecord() {
 			actualRecord, err := s.twapkeeper.GetBeginBlockAccumulatorRecord(s.Ctx, tc.poolId, tc.baseDenom, tc.quoteDenom)
 
 			if tc.expError != nil {
-				if !strings.Contains(err.Error(), tc.expError.Error()) {
-					s.Require().Equal(tc.expError, err)
-				}
+				s.Require().ErrorContains(err, fmt.Sprintf("%s", tc.expError))
 				return
 			}
 
--- a/x/twap/api_test.go
+++ b/x/twap/api_test.go
@@ -2,6 +2,7 @@ package twap_test
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -74,7 +75,9 @@ func (s *TestSuite) TestGetBeginBlockAccumulatorRecord() {
 			actualRecord, err := s.twapkeeper.GetBeginBlockAccumulatorRecord(s.Ctx, tc.poolId, tc.baseDenom, tc.quoteDenom)
 
 			if tc.expError != nil {
-				s.Require().Equal(tc.expError, err)
+				if !strings.Contains(err.Error(), tc.expError.Error()) {
+					s.Require().Equal(tc.expError, err)
+				}
 				return
 			}
 
--- a/x/twap/store.go
+++ b/x/twap/store.go
@@ -21,16 +21,6 @@ func (e timeTooOldError) Error() string {
 		" Try storing the accumulator value. (requested time %s)", e.Time)
 }
 
-type twapNotFoundError struct {
-	Asset0Denom string
-	Asset1Denom string
-}
-
-func (e twapNotFoundError) Error() string {
-	return fmt.Sprintf("TWAP not found, but there are other twaps available for this time."+
-		" Please make sure tha asset0denom and asset1denom (%s, %s) are correct, and in order (asset0 > asset1)?", e.Asset0Denom, e.Asset1Denom)
-}
-
 // trackChangedPool places an entry into a transient store,
 // to track that this pool changed this block.
 // This tracking is for use in EndBlock, to create new TWAP records.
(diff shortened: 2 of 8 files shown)


czarcas7ic:
I find this simplification helpful and am for including this now!

x/twap/api_test.go
```diff
@@ -74,7 +75,9 @@ func (s *TestSuite) TestGetBeginBlockAccumulatorRecord() {
 			actualRecord, err := s.twapkeeper.GetBeginBlockAccumulatorRecord(s.Ctx, tc.poolId, tc.baseDenom, tc.quoteDenom)
 
 			if tc.expError != nil {
-				s.Require().Equal(tc.expError, err)
+				if !strings.Contains(err.Error(), tc.expError.Error()) {
+					s.Require().Equal(tc.expError, err)
+				}
```
ValarDragon, line 80: This basically relaxes the error checking condition to substring matching for now.

ValarDragon:
This PR is now ready for review / has intended tests added! store_test_helper is changed due to gofmt issues, only two lines changed for error message corrections (not sure why linter didn't catch the format issues before)

x/twap/api_test.go
stackman27: If the substring is not in the string, why is it necessary to check if they are equal later? Wouldn't that automatically be a false case? For instance this case: `ExpErr: F, Err: ABCD` 

Also you think  we can use `ErrorContainsf`instead? 
`s.Require().ErrorContainsf(err,  tc.expectError, err.Error(), "formatted")`

x/twap/types/keys.go
```diff
@@ -51,46 +51,19 @@ func FormatHistoricalTimeIndexTWAPKey(accumulatorWriteTime time.Time, poolId uin
...
-	}
-	t, err := osmoutils.ParseTimeString(s[2])
-	if err != nil {
-		return time.Time{}, TimeStringKeyFormatError{Key: keyS, Err: err}
-	}
-	return t, nil
+func FormatHistoricalPoolIndexTimeSuffix(poolId uint64, denom1, denom2 string, accumulatorWriteTime time.Time) []byte {
+	timeS := osmoutils.FormatTimeString(accumulatorWriteTime)
+	// . acts as a suffix for lexicographical orderings
+	return []byte(fmt.Sprintf("%s%d%s%s%s%s%s%s.", HistoricalTWAPPoolIndexPrefix, poolId, KeySeparator, denom1, KeySeparator, denom2, KeySeparator, timeS))
```
AlpinYukseloglu, line 66: Making this a suffix with `.` is really neat

AlpinYukseloglu (approved):
I actually really like these changes (including the additional cleanup you did) – would be in support of including this

x/twap/api_test.go
czarcas7ic: @stackman27 I think he just did a quick fix for CI. 

I implemented ErrorContains instead of ErrorContainsf, but had similar thinking as you, lmk if you agree

https://github.com/osmosis-labs/osmosis/pull/2641/commits/ac1796cbda18f76ddd77ef7c39db1df86ce53bd2

stackman27: looks solid!

mattverse (commented):
What is the 1ns change hack and how does this change remove the hack?

x/twap/store.go
```diff
@@ -173,40 +168,26 @@ func (k Keeper) getRecordAtOrBeforeTime(ctx sdk.Context, poolId uint64, t time.T
 		return types.TwapRecord{}, err
 	}
 	store := ctx.KVStore(k.storeKey)
-	// We make an iteration from time=t + 1ns, to time=0 for this pool.
-	// Note that we cannot get any time entries from t + 1ns, as the key would be `prefix|t+1ns`
-	// and the end for a reverse iterator is exclusive. Thus the largest key that can be returned
-	// begins with a prefix of `prefix|t`
-	startKey := types.FormatHistoricalPoolIndexTimePrefix(poolId, time.Unix(0, 0))
-	endKey := types.FormatHistoricalPoolIndexTimePrefix(poolId, t.Add(time.Nanosecond))
```
ValarDragon, line 181: @mattverse this is the +1 ns hack.

This is fixed in this PR, by managing the suffix better, in `FormatHistoricalPoolIndexTimeSuffix`

x/twap/store.go
```diff
@@ -143,7 +133,12 @@ func (k Keeper) getMostRecentRecordStoreRepresentation(ctx sdk.Context, poolId u
 	store := ctx.KVStore(k.storeKey)
 	key := types.FormatMostRecentTWAPKey(poolId, asset0Denom, asset1Denom)
 	bz := store.Get(key)
-	return types.ParseTwapFromBz(bz)
+	twap, err := types.ParseTwapFromBz(bz)
+	if err != nil {
+		err = fmt.Errorf("error in get most recent twap, likely that asset 0 or asset 1 were wrong: %s %s."+
```
p0mvn, line 138: The only possible error from the callee is proto unmarshalling. What does it mean for them to be wrong?

p0mvn (approved):

x/twap/store.go
AlpinYukseloglu: The callee also errors if there is no TWAP found, no?

https://github.com/osmosis-labs/osmosis/blob/c4766c7302de294de902d69c26e192bb4ecfa747/x/twap/types/keys.go#L109-L111

ValarDragon: yeah, thats the thing we're catching

PR #2656, opened 2022-09-08
Add type safety to query whitelist

ValarDragon:

Make type safety guaranteed for Stargate query whitelist, and make it impossible to set things on it from other packages.


alexanderbez (approved):
nit: I think the new APIs could use godocs, but non-blocking :)

ValarDragon:
Added!

mattverse (approved):
LGTM!

PR #2683, opened 2022-09-09
Remove Get from GetArithmeticTwap query & GetArithmeticTwapToNow query

ValarDragon:

Renames these queries to remove the `Get` from their name

## Testing and Verifying

### DIFF
--- a/proto/osmosis/twap/v1beta1/query.proto
+++ b/proto/osmosis/twap/v1beta1/query.proto
@@ -18,18 +18,16 @@ service Query {
   rpc Params(ParamsRequest) returns (ParamsResponse) {
     option (google.api.http).get = "/osmosis/twap/v1beta1/Params";
   }
-  rpc GetArithmeticTwap(GetArithmeticTwapRequest)
-      returns (GetArithmeticTwapResponse) {
-    option (google.api.http).get = "/osmosis/twap/v1beta1/GetArithmeticTwap";
+  rpc ArithmeticTwap(ArithmeticTwapRequest) returns (ArithmeticTwapResponse) {
+    option (google.api.http).get = "/osmosis/twap/v1beta1/ArithmeticTwap";
   }
-  rpc GetArithmeticTwapToNow(GetArithmeticTwapToNowRequest)
-      returns (GetArithmeticTwapToNowResponse) {
-    option (google.api.http).get =
-        "/osmosis/twap/v1beta1/GetArithmeticTwapToNow";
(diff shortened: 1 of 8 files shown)


proto/osmosis/twap/v1beta1/query.yml
```diff
@@ -3,18 +3,18 @@ keeper:
...
-  GetArithmeticTwap:
+  ArithmeticTwap:
     proto_wrapper:
       default_values:
         Req.end_time: "time.Now()"
       query_func: "k.GetArithmeticTwap"
     cli:
-      cmd: "GetArithmeticTwap"
-  GetArithmeticTwapToNow:
+      cmd: "ArithmeticTwap"
```
p0mvn, line 12: I know this is unrelated to the core of the work in this PR but should we also change `time.Now()` above? to `ctx.BlockTime()`?

p0mvn (approved):

proto/osmosis/twap/v1beta1/query.yml
ValarDragon: 100%

proto/osmosis/twap/v1beta1/query.proto
```diff
@@ -18,18 +18,16 @@ service Query {
   rpc Params(ParamsRequest) returns (ParamsResponse) {
     option (google.api.http).get = "/osmosis/twap/v1beta1/Params";
   }
-  rpc GetArithmeticTwap(GetArithmeticTwapRequest)
-      returns (GetArithmeticTwapResponse) {
-    option (google.api.http).get = "/osmosis/twap/v1beta1/GetArithmeticTwap";
+  rpc ArithmeticTwap(ArithmeticTwapRequest) returns (ArithmeticTwapResponse) {
```
stackman27, line 21: since we're using a template to generate grpc queries, i think it'd be better if we have a consistent naming standard 

For instance, Gamm queries start with "query" prefix like QueryPoolsRequest QueryTotalLiquidityRequest QuerySpotPriceResponse

However, Twap queries doesnot contain "query" prefix like ArithmeticTwapRequest,ArithmeticTwapToNowRequest,

ValarDragon: Good point, the Query prefix is standard within the SDK as well

ValarDragon: This decision has no state machine or client facing affects as far as I'm aware, so lets decide in an issue?

stackman27: Gotchu!

Issue #2659, opened 2022-09-08
test(twap): updateRecords function


p0mvn:
## Background

`updateRecords` is missing tests. We should add them

## Suggested Design

Utilize `AmmInterface` mock where applicable

## Acceptance Criteria

- tests added
- coverage maximized

Issue #2686, opened 2022-09-09
refactor(twap): add sanity check in `updateRecord` to compare `record` time and height against `ctx`


p0mvn:
## Background

This was observed during testing of `updateRecords`. It is possible to set up a testing environment in such a way as to be able to insert a record between the existing records.

For example, assume a two-asset pool and 2 records at the times `t` and `t + 2`. If I set up `ctx.BlockTime` to be `t + 1`, I can insert a new record in-between without error.

While this should not be possible to reproduce in real environment, having a sanity check in `updateRecord` to ensure the following invariants would be useful:
- `ctx.BlockTime` > `record.Time`
- `ctx.BlockHeight` > `record.Height`

`updateRecord` and `updateRecords` should be returning an error in these cases.

## Suggested Design

Return error if any of the above invariants are not satisfied.

## Acceptance Criteria
- invariants are added to `updateRecord`
- new code branches of `updateRecord` are covered by tests
- existing `updateRecords` tests are refactored / adjusted to cover the new behavior

hieuvubk:
https://github.com/osmosis-labs/osmosis/blob/1111d84631be02fc3c7e8ebc56a0c56a044fa476/x/twap/listeners_test.go#L26
@p0mvn We have an exception here. With sanity check added, swap on pool creation block should return err. What do u think about that?

hieuvubk:
> https://github.com/osmosis-labs/osmosis/blob/1111d84631be02fc3c7e8ebc56a0c56a044fa476/x/twap/listeners_test.go#L26
> 
> 
> @p0mvn We have an exception here. With sanity check added, swap on pool creation block should return err. What do u think about that?
I thịnk we should discuss with @mattverse owning this commit

hieuvubk:
> > https://github.com/osmosis-labs/osmosis/blob/1111d84631be02fc3c7e8ebc56a0c56a044fa476/x/twap/listeners_test.go#L26
> > 
> > @p0mvn We have an exception here. With sanity check added, swap on pool creation block should return err. What do u think about that?

I think we should discuss with @mattverse owning this commit

mattverse:
@hieuvubk  What causes the error? At a light glance, I don't see anything calling `updateRecord` with in the afterPoolCreation test

hieuvubk:
Sr to bother , I checked actually this comes from `EndBlock`( just logged the err) when updating the record created in the same block in `afterCreatePool`.
[...]

PR #2631, opened 2022-09-07
Fix markdown files using `make markdown`

AlpinYukseloglu:
## What is the purpose of the change

This PR tests the `make markdown` functionality added in #2630 to determine if there are any changes that need to be made and  to clean up our existing markdown files (so as to not clutter future outputs of `make markdown` or `make mdlint`)

## Brief Changelog

- Ran `make markdown`

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (not applicable)

PR #2687, opened 2022-09-09
test(twap): `updateRecords` function (two-asset)

p0mvn:
Closes: #2659

Covers `updateRecords` function by tests

## Brief Changelog
- added tests
- refactored amm mock
    * Reason: It was impossible to set it up in such a way as to test the case with invalid number of records.
- Moved two errors to `types` - one in `gamm`, one in `twap`. This is to be able to make sure that the desired error is returned in tests.

This change is state-breaking relative to the previous RCs.

## Follow-Ups
- https://github.com/osmosis-labs/osmosis/issues/2686
- https://github.com/osmosis-labs/osmosis/issues/2688

## Testing and Verifying

This change added tests.


x/twap/logic.go
```diff
@@ -46,12 +46,8 @@ func getSpotPrices(ctx sdk.Context, k types.AmmInterface, poolId uint64, denom0,
...
 		// so that the numbers can be still be calculated within TWAPs over error values
 		// TODO: Should we be using the last spot price?
-		if (sp0 == sdk.Dec{}) {
-			sp0 = sdk.ZeroDec()
-		}
-		if (sp1 == sdk.Dec{}) {
-			sp1 = sdk.ZeroDec()
-		}
+		sp0 = sdk.ZeroDec()
+		sp1 = sdk.ZeroDec()
```
p0mvn, line 50: Note to reviewers: please see "Changelog"  under PR description

p0mvn: Also, the above TODO makes sense IMO. Should we consider switching?

x/twap/logic.go
```diff
@@ -43,15 +43,15 @@ func getSpotPrices(ctx sdk.Context, k types.AmmInterface, poolId uint64, denom0,
...
+	}
+	// In the event of an error, we just sanity replace empty values with zero values
+	// so that the numbers can be still be calculated within TWAPs over error values
+	// TODO: Should we be using the last spot price?
+	if err0 != nil {
+		sp0 = sdk.ZeroDec()
+	}
+	if err1 != nil {
+		sp1 = sdk.ZeroDec()
 	}
```
p0mvn, line 55: Note to reviewers: please see "Changelog" under PR description

Also, the TODO makes sense. Should we consider switching?

p0mvn:
md linter seems to be broken on `main`

cc: @nicolaslara @AlpinYukseloglu might be related to #2409 or #2631

AlpinYukseloglu:
> md linter seems to be broken on `main`
> 
> 
> 
> cc: @nicolaslara @AlpinYukseloglu might be related to #2409 or #2631

I was getting issues with the general linter as well but `make markdown` should resolve any markdown-related issues (& flag any it couldn't solve)

nicolaslara:
I saw those issues on a branch after merging main. I'm guessing there was some change on what the linter checks. The rate limiter readme just needs a few blank lines around titles.

nicolaslara:
@AlpinYukseloglu Added the readme changes here: https://github.com/osmosis-labs/osmosis/pull/2691

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+
+// TestUpdateRecords tests that the records are updated correctly.
+// It tests the following:
+// - two-asset pools
+// - multi-asset pools
+// - with spot price errors
+// - without spot price errors
+// - that new records are created
+// - older historical records are not updated
+// - spot price errot times are either propagated from
```
mattverse, line 534:
```suggestion
// - spot price error times are either propagated from
```

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+// older records or set to current block time in case error occurred.
+func (s *TestSuite) TestUpdateRecords() {
+
```
mattverse, line 537:
```suggestion
```

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+	type expectedResults struct {
+		spotPriceA    sdk.Dec
+		spotPriceB    sdk.Dec
+		lastErrorTime time.Time
+	}
+
+	var spError = errors.New("spot price error")
+
+	validateRecords := func(expectedRecords []expectedResults, actualRecords []types.TwapRecord) {
```
mattverse, line 554: nit: I think this could be a suite method + can be used in different places around our test code

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+// It tests the following:
+// - two-asset pools
+// - multi-asset pools
+// - with spot price errors
+// - without spot price errors
+// - that new records are created
+// - older historical records are not updated
+// - spot price errot times are either propagated from
+// older records or set to current block time in case error occurred.
+func (s *TestSuite) TestUpdateRecords() {
```
mattverse, line 536: I think we have duplicated boundaries of testing with `TestUpdateTwap` with this test case that could be merged into one, wdyt?

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+					quoteDenom:  baseRecord.Asset0Denom,
+					overrideSp:  types.MaxSpotPrice.Add(sdk.OneDec()),
+					overrideErr: nil, // twap logic should identify the large spot price and mark it as error.
+				},
+			},
+
+			expectedMostRecentRecords: []expectedResults{
+				{
+					spotPriceA:    sdk.OneDec(),
+					spotPriceB:    types.MaxSpotPrice,               // Although the price returned from AMM was MaxSpotPrice + 1, it is reset to just MaxSpotPrice.
```
mattverse, line 712: oh interesting so we cap at max spot price regardless of having greater value than sp

x/twap/logic_test.go
```diff
@@ -521,3 +522,422 @@ func (s *TestSuite) TestPruneRecords() {
...
+					spotPriceB: sdk.OneDec().Add(sdk.OneDec()),
+				},
+				// The original record at t + 1.
+				{
+					spotPriceA: tPlus10sp5Record.P0LastSpotPrice,
+					spotPriceB: tPlus10sp5Record.P1LastSpotPrice,
+				},
+			},
+		},
+		// TODO: complete multi-asset pool tests:
```
mattverse, line 900: let's create an issue for this so that we come back to this!

mattverse (approved):
LGTM!

x/twap/logic_test.go
p0mvn: I've given our tests a quick skim and couldn't find where this function and `expectedResults` type could be reused.

The problem here is that `expectedResults` is a type that is local to this test. I think if anyone decides to make that type global / apply it in other tests, then it would make sense.

x/twap/logic_test.go
p0mvn: `TestUpdateTwap` tests `updateRecord` method while this is testing `updateRecords` (plural).

I've referenced that test while implementing it, and, unfortunately, I could not find any logic suitable for reuse or de-duplication.

x/twap/logic_test.go
p0mvn: Yes we cap it if it is greater but we set it to zero if we get an error

x/twap/logic_test.go
p0mvn: Already made: https://github.com/osmosis-labs/osmosis/issues/2688

p0mvn:
Comments addressed and waiting for one more approval

x/twap/logic_test.go
mattverse: Ah didn't notice the difference in test being plural -and singular. IMO we should change `TestUpdateTwap` -> `TestUpdateRecords` so that we can be consistent with the testing names

x/twap/logic_test.go
```diff
@@ -521,3 +522,421 @@ func (s *TestSuite) TestPruneRecords() {
...
+					overrideSp: sdk.NewDec(2),
+				},
+				{
+					baseDenom:  baseRecord.Asset1Denom,
+					quoteDenom: baseRecord.Asset0Denom,
+					overrideSp: sdk.NewDecWithPrec(2, 1),
+				},
+				{
+					baseDenom:  baseRecord.Asset1Denom,
+					quoteDenom: "extra denom B",
```
czarcas7ic, line 605: Are spaces allowed in denoms? If not, we should use a valid denom for the extra denom

x/twap/logic_test.go
```diff
@@ -521,3 +522,421 @@ func (s *TestSuite) TestPruneRecords() {
...
+					baseDenom:   baseRecord.Asset0Denom,
+					quoteDenom:  baseRecord.Asset1Denom,
+					overrideErr: spError,
+				},
+				{
+					baseDenom:   baseRecord.Asset1Denom,
+					quoteDenom:  baseRecord.Asset0Denom,
+					overrideSp:  sdk.NewDecWithPrec(2, 1),
+					overrideErr: spError,
+				},
```
czarcas7ic, line 665: Clarifying question: The description says "updated with spot price error in both denom pairs", there is a spErr in both denoms pairs, but we dont overrideSp for baseQuoteAB. Is this intentional? In my mind, if there is a spot price error in both denom pairs, then they should both return zero (as it currently stands, I am in 100 percent agreement with your recommendation that they should actually return the last known spot price)

x/twap/logic_test.go
```diff
@@ -521,3 +522,421 @@ func (s *TestSuite) TestPruneRecords() {
...
+				// The new record added.
+				{
+					spotPriceA:    sdk.ZeroDec(),
+					spotPriceB:    sdk.NewDecWithPrec(2, 1),
+					lastErrorTime: baseRecord.Time.Add(time.Second), // equals to block time
+				},
+			},
+		},
+		"two-asset; pre-set record at t; large spot price in one of the pairs": {
```
czarcas7ic, line 689: nit: Maybe instead of "large spot price" we should specify "invalidly large spot price"

x/twap/logic_test.go
```diff
@@ -521,3 +522,421 @@ func (s *TestSuite) TestPruneRecords() {
...
+				twapKeeper.SetAmmInterface(ammMock)
+			}
+
+			s.preSetRecords(tc.preSetRecords)
+
+			err := twapKeeper.UpdateRecords(ctx, tc.poolId)
+
+			if tc.expectError != nil {
+				s.Require().Error(err)
+				s.Require().ErrorIs(err, tc.expectError)
```
czarcas7ic, line 927:
```suggestion
				s.Require().ErrorIs(err, tc.expectError)
```
nit

czarcas7ic (approved):
This LGTM! 👍

ValarDragon:
Something feels like it's gone quite wrong in testing design Here. 

This function were supposed to be testing is quite small and has minimal functionality yet has 400+ lines of tests

ValarDragon (approved):

x/twap/logic_test.go
p0mvn: It is not invalid to have a large spot price. However, we do cap it at a specific maximum

x/twap/logic_test.go
p0mvn: This is intentional as it currently stands.

We only set the spot price to zero if there is a spot price error and the returned value from amm is an empty `Dec`:
https://github.com/osmosis-labs/osmosis/blob/bcda1cbe01f87111eb9320a8907b6b89d9930fb6/x/twap/logic.go#L49-L51

I'm planning to change this logic in:
https://github.com/osmosis-labs/osmosis/issues/2689

p0mvn:
> 

I noticed that it is possible to refactor and infer most recent records from the boolean flag in historical records:
[infer most recent from historical](https://github.com/osmosis-labs/osmosis/pull/2687/commits/68ede334686fcfe0aaffbe3788d2ad99587c110d)

I'm not seeing anything else that can be changed or removed. However, I'm open to suggestions.

This function does quite a bit:
- there are 2 edge cases (error branches).
- it mutates 2 database indexes. As a result, we need to test if they are updated correctly.
- there are a few extra edge cases stemming from the spot prices logic that I attempted to cover.

Also, the mock setup logic and overrides take up some space

p0mvn:
Merging this, we can discuss further improving this test in a follow-up PR

ValarDragon:
Yeah, focusing on reducing these lines of code is not at all important! But would like to figure out if we can make strategies to reduce this for similar cases in the future

Issue #2696, opened 2022-09-11
[stableswap]: Simplify two-asset CFMM solver


AlpinYukseloglu:
## Background

Our current stableswap CFMM solver is untenably complicated, to the point where it's practically impossible to visually parse or even debug. We should make an effort to simplify the solver as much as possible, prioritizing having as few terms that are approximated (sqrt etc.), taken to a large power (to avoid overflows), and negative as possible. cc: @ValarDragon since this would be replacing your original implementation 

## Suggested Design

We currently solve our CFMM as though _every single input_ is unknown, which leads to an unnecessarily complicated solver. I believe we can drastically simplify our CFMM by making the following assumptions:
- For any given swap, everything except for the amount of the output asset is known & constant
- The number of assets in the pool are always greater than zero (let's us use the original CFMM as part of our solver)
- Finding the final asset reserve amounts is as good as finding the output amount since we can subtract the original reserves at the end

Pseudocode:
```
func solveCfmm(xReserve, yReserve, yIn) {
    Calculate k using xReserve and yReserve

    yReserve = yReserve + yIn
    
    Use new yReserve to solve for new xReserve (using new solver derived from xy(x^2 + y^2) = k)

    xOut = original xReserve minus resulting output

    Return xOut
}
```

## Acceptance Criteria

- Tests comparing to other CFMM solvers pass
- Comprehensive tests are added and pass
- Desmos graph is overlayed onto `xy(x^2 + y^2) = k` and put in comments for reference

PR #2697, opened 2022-09-11
[stableswap]: simplify CFMM solver 

AlpinYukseloglu:
Closes: #2696 

This PR implements a drastically simplified solver for our two-asset stableswap CFMM.

Key differences:
- Fits on a single screen (core solver logic is ~5 lines)
- _Far_ fewer square root, product, and power terms
- Largest power is same as previous (8)
- No negative numbers in intermediate calculations
- Much easier to parse visually

## Brief Changelog

I derived this solver manually, but we can confirm on Desmos that it fits our curve exactly: https://www.desmos.com/calculator/wxl9ffxofw

As described in the issue, the key simplifcations leaned on the following assumptions:
- For any given swap, everything except for the amount of the output asset is known & constant
- The number of assets in the pool are always greater than zero (let's us use the original CFMM as part of our solver)
- Finding the final asset reserve amounts is as good as finding the output amount since we can subtract the original reserves at the end

Pseudocode for new solver:
```
func solveCfmm(xReserve, yReserve, yIn) {
    Calculate k using xReserve and yReserve

    yReserve = yReserve + yIn
    
    Use new yReserve to solve for new xReserve (using new solver derived from xy(x^2 + y^2) = k)

    xOut = original xReserve minus resulting output

    Return xOut
}
```

## Testing and Verifying

Basic tests pass, and the ones that failed on our original solver are around 3x closer to the target with this solver (likely due to fewer terms leading to fewer approximations).

Still working on more comprehensive tests, although we might first need to solve general precision issues around our CFMM.


ValarDragon:
Ah great idea, there was no need to keep `b` as a variable in the prior equations!

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -256,9 +272,59 @@ func solveCfmmMulti(xReserve, yReserve, wSumSquares, yIn sdk.Dec) sdk.Dec {
...
+	}
+
 	return a
 }
 
+// solidly CFMM is xy(x^2 + y^2) = k
+// So we want to solve for a given addition of `b` units of y into the pool,
+// how many units `a` of x do we get out.
+// So we solve the following expression for `a`
+// xy(x^2 + y^2) = (x - a)(y + b)((x - a)^2 + (y + b)^2)
```
ValarDragon, line 286:
```suggestion
// Let y' = y + b
// we solve k = (x'y')(x'^2 + y^2) for x', using the following equation {wolfram alpha link}
// Then we use that to derive the change in x as x_out = x' - x
```

AlpinYukseloglu:
Recent commit cuts the exponents of the two largest terms in half (`x^8 -> x^4` and `k^2 -> k`)

This makes meaningful progress towards #2701, but we still overflow at much smaller asset amounts than we expect to be able to accommodate (current capacity is just under $10m per asset in a 2-asset pool, and we need to be able to support $100m per asset in a 3+ asset pool)

Reference to verify new implementation: https://www.desmos.com/calculator/bx5m5wpind

The last idea/trick I have in mind is potentially dropping the +1s given the size of pools we're dealing with (maybe have two solvers and switch to the other one when x, y, or k is sufficiently large)

Ultimately we'll want to use our binary search solver, but I believe it's important to have a comprehensive and simple reference solver to benchmark against in the meantime

ValarDragon:
nice, this is great! Big win!

(The largest term is x^7 still tho =p, `k = O(x^4)`)

Heres an equation switch, basically moving `(x^2k)^{1/3}` to instead be `x (k/x)^{1/3}`, which then truly lowers the largest term to x^4 as far as I can tell!

And actually if you unwrap `k` in those equations, by making `k' = k/x = y(x^2 + y^2)`, and computing off of `k'` directly everywhere, you'd actually be at `O(x^3)` everywhere AFAICT!

AlpinYukseloglu:
This is awesome! To your second point: can't we just exponentiate wherever we see a `k / x` to take it from `O(x^4)` to `O(x^3)` without ever needing to unwrap the k?

e.g. `k / x = (sqrt(k) / sqrt(x))^2` so we reach `O(x^3)` without any intermediate terms > `O(x^2)`?

ValarDragon:
That looks great to me!

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -276,6 +281,59 @@ func solveCfmmMulti(xReserve, yReserve, wSumSquares, yIn osmomath.BigDec) osmoma
 	return a
 }
 
+// solidly CFMM is xy(x^2 + y^2) = k
+// So we want to solve for a given addition of `b` units of y into the pool,
+// how many units `a` of x do we get out.
+// Let y' = y + b
+// we solve k = (x'y')(x'^2 + y^2) for x', using the following equation {wolfram alpha link}
```
ValarDragon, line 288: Needs wolfram alpha link, or equation we test and explanation of correctness

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -276,6 +281,59 @@ func solveCfmmMulti(xReserve, yReserve, wSumSquares, yIn osmomath.BigDec) osmoma
...
+	y3 := y2.Mul(y_new)
+	y4 := y3.Mul(y_new)
+	large_term := (y4.Quo(k)).Mul(osmomath.NewBigDec(2).Quo(twentySevenRootTwo))
+	large_term2 := large_term.Mul(large_term)
+
+	// solve for new xReserve using new yReserve and old k using a solver derived from xy(x^2 + y^2) = k
+	sqrt_term, err := (osmomath.OneDec().Add(large_term2)).ApproxRoot(2)
+	if err != nil {
+		panic(err)
+	}
```
ValarDragon, line 316: Lets take the equation we want to compute, write out the variable substitutions we do, then show that each of this is very clearly computing the desired variable

AlpinYukseloglu: Took a crack at this in the recent commit!

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: I put a wolfram alpha link of the original equation, a desmos link for our simplified one, and direct equation + our key abstractions in my last commit

ValarDragon (approved):
Nice!

Issue #2566, opened 2022-09-01
V12: properly update and release fork dependencies


p0mvn:
## Background

Currently, we have our core repository depending on tags that are in development.

`osmosis-labs/osmosis` depends on `osmosis-main` of `osmosis-labs/cosmos-sdk`

Additionally, `osmosis-labs/wasmd` will temporarily depend on `osmosis-main` `osmosis-labs/cosmos-sdk` in https://github.com/osmosis-labs/osmosis/issues/2368

## Suggested Design

Before v12 release, we should:

1.Kick off a `v045.0x-osmo-v12` `osmosis-labs/cosmos-sdk` branch.
- Make a tag off of that branch.

2. Kick off a `v0.28.0x-osmo-v12` `osmosis-labs/wasmd` branch
- Update the `osmosis-labs/cosmos-sdk` dependency
- Make a tag off of that branch with SDK updated

3. Update `osmosis-labs/osmosis` to depend on the above tags

## Acceptance Criteria
- `osmosis-labs/cosmos-sdk`
    * created v12 release branch (`v045.0x-osmo-v12`)
    * release tag created from release branch
- `osmosis-labs/wasmd`
   * created v12 release branch off of upstream `v0.28.0` tag (`v0.28.0x-osmo-v12`)
   * release tag created from release branch
- `osmosis-labs/osmosis`
   * `osmosis-labs/cosmos-sdk` dependency is updated
   * `osmosis-labs/wasmd`  dependency is updated

p0mvn:
This should be done right before code freeze and is currently blocked on more v12 changes being merged across the stack

p0mvn:
@mattverse what is the decision on wasmd? are we going to use the updated upstream or our fork?

mattverse:
Apologies, I just saw this, both ways work we don't have preference over one another, I think it'd be fine we use wasmd upstream?

p0mvn:
This has been resolved

PR #2714, opened 2022-09-12
chore: upgrade wasmd to v0.28.0

p0mvn:
Part of #2566

Upgrade wasmd to v0.28.0

Branch: https://github.com/osmosis-labs/wasmd/tree/v0.28.0x-osmo-v12
Tag: https://github.com/osmosis-labs/wasmd/releases/tag/v0.28.0-osmo-v12.1

## Testing and Verifying

### DIFF
--- a/x/gamm/simulation/sim_msgs.go
+++ b/x/gamm/simulation/sim_msgs.go
@@ -355,7 +355,7 @@ func RandomExitSwapShareAmountIn(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.
 		PoolId:            pool_id,
 		TokenOutDenom:     tokenOut.Denom,
 		ShareInAmount:     gammShares.Amount.Quo(sdk.NewInt(2)),
-		TokenOutMinAmount: tokenOut.Amount,
+		TokenOutMinAmount: tokenOut.Amount.Quo(sdk.NewInt(2)),
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


p0mvn:
I've investigated the simulator failure, and it seems to be happening during fee validation in `x/txfees`:


Essentially, wasmd generates a tx with a random fee:
https://github.com/osmosis-labs/wasmd/blob/0f2794c47af3c1993f94c9f45953b3de6f4e2bf5/x/wasm/simulation/operations.go#L154

I think the solution should be to modify wasmd side and hardcode the denom to be "stake"

p0mvn:
Seems that we only allow fees in "uosmo" in our simulator too:
https://github.com/osmosis-labs/osmosis/blob/1d9c6e383a6863abac0aa90e73a817cfc82d355f/simulation/simtypes/legacy.go#L38-L39

So I'm going to go ahead and make the change wasmd side and retag

faddat:
Recommend v0.29.0 in fact.

x/gamm/simulation/sim_msgs.go
```diff
@@ -355,7 +355,7 @@ func RandomExitSwapShareAmountIn(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.
 		PoolId:            pool_id,
 		TokenOutDenom:     tokenOut.Denom,
 		ShareInAmount:     gammShares.Amount.Quo(sdk.NewInt(2)),
-		TokenOutMinAmount: tokenOut.Amount,
+		TokenOutMinAmount: tokenOut.Amount.Quo(sdk.NewInt(2)),
```
p0mvn, line 358: The simulator was failing because of this. I think the calculations above are making it flaky so this change is made to unblock v12. Once the TODO above is addressed, this should be as well

p0mvn:
> Recommend v0.29.0 in fact.
> 
> Also, I just vendored x/wasm and x/globalfees into eve, and I gotta say it feels about right, there won't be as much upstream dependency chaos that way.
> 
> I'd be happy to hack on any of that for osmo.

v0.29.0 is still an RC. Also, we really want to get v12 out, and this is one of the last missing pieces.

Since regular wasmd tag requires additional changes to be compatible with Osmosis, I think we will go with v0.28 for the sake of delivering v12. However, happy to upgrade to v0.29 as soon as stable release is out.

> Also, I just vendored x/wasm and x/globalfees into eve, and I gotta say it feels about right

Do you mean added it directly into the repository? How are you planning to integrate improvements from upstream?

czarcas7ic (approved):

PR #2721, opened 2022-09-13
Chore: Simplify test code for Stargate Queries

mattverse:
Closes: #XXX

Existing Test code for stargate query used hex as its test field, which caused unnecessary confusion within the test code. (confusion coming from the fact that we're not using hex any where within the actual implementation, but have a process of decoding hex within the test case) 

This PR changes the test code so that it uses bytes directly as a test case field/ 

## Brief Changelog
- Change hex 

## Testing and Verifying
This changes and simplifies existing test

  - Does this pull request introduce a new feature or user-facing behavior changes?no 
  - How is the feature or change documented? (not applicable


x/twap/client/grpc/grpc_query.go
```diff
@@ -47,3 +38,12 @@ func (q Querier) ArithmeticTwapToNow(grpcCtx context.Context,
 	ctx := sdk.UnwrapSDKContext(grpcCtx)
 	return q.Q.ArithmeticTwapToNow(ctx, *req)
 }
+func (q Querier) Params(grpcCtx context.Context,
```
mattverse, line 41: proto check would fail without this change

ValarDragon: huh, interesting

ValarDragon (approved):
Nice, very glad the hex is gone

PR #2725, opened 2022-09-13
Make gamm pools override addr already exists

ValarDragon:

Make gamm pools still creatable if the underlying account already exists

## Brief Changelog

- Add helper methods for creating module accounts, to override underlying ones. This is safe because module accounts as long created via the correct constructors, have domain separation and collision resistance from other addresses.
- Apply change to AMM module

## Testing and Verifying

- added unit tests that validate the new method works with prior addresses existing

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no*
  - How is the feature or change documented? code comments

### DIFF
--- a/osmoutils/module_account.go
+++ b/osmoutils/module_account.go
@@ -1,0 +1,67 @@
+package osmoutils
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
+)
+
+type AccountKeeper interface {
+	NewAccount(sdk.Context, authtypes.AccountI) authtypes.AccountI
+
+	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authtypes.AccountI
+	SetAccount(ctx sdk.Context, acc authtypes.AccountI)
+}
+
+// CanCreateModuleAccountAtAddr tells us if we can safely make a module account at
+// a given address. By collision resistance of the addr, we assume that
+// TODO: This is generally an SDK design flaw
+// code based off wasmd code: https://github.com/CosmWasm/wasmd/pull/996
+func CanCreateModuleAccountAtAddr(ctx sdk.Context, ak AccountKeeper, addr sdk.AccAddress) error {
+	existingAcct := ak.GetAccount(ctx, addr)
+	if existingAcct == nil {
+		return nil
+	}
+	if existingAcct.GetSequence() != 0 || existingAcct.GetPubKey() != nil {
+		return fmt.Errorf("cannot create module account %s, "+
+			"due to an account at that address already existing & having sent txs", addr)
+	}
+	var overrideAccountTypes = map[reflect.Type]struct{}{
+		reflect.TypeOf(&authtypes.BaseAccount{}):                 {},
+		reflect.TypeOf(&vestingtypes.DelayedVestingAccount{}):    {},
+		reflect.TypeOf(&vestingtypes.ContinuousVestingAccount{}): {},
+		reflect.TypeOf(&vestingtypes.BaseVestingAccount{}):       {},
+		reflect.TypeOf(&vestingtypes.PeriodicVestingAccount{}):   {},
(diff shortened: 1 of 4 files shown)


osmoutils/module_account_test.go
```diff
@@ -0,0 +1,56 @@
...
+			priorAccounts: []authtypes.AccountI{baseWithAddr(defaultModuleAccAddr)},
+			moduleAccAddr: defaultModuleAccAddr,
+			expErr:        false,
+		},
+		"prior user acc at addr": {
+			priorAccounts: []authtypes.AccountI{userAcc(defaultModuleAccAddr)},
+			moduleAccAddr: defaultModuleAccAddr,
+			expErr:        true,
+		},
+	}
```
p0mvn, line 44: consider a failing case with an account that has pub key existing

osmoutils/module_account_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for name, tc := range testcases {
```
p0mvn, line 45: I'm assuming that there are accounts of an overridable type since we have a black list. Can we have a test with the account that can be overridden?

osmoutils/module_account.go
```diff
@@ -0,0 +1,67 @@
...
+
+type AccountKeeper interface {
+	NewAccount(sdk.Context, authtypes.AccountI) authtypes.AccountI
+
+	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authtypes.AccountI
+	SetAccount(ctx sdk.Context, acc authtypes.AccountI)
+}
+
+// CanCreateModuleAccountAtAddr tells us if we can safely make a module account at
+// a given address. By collision resistance of the addr, we assume that
```
p0mvn, line 21: nit: seems like unfinished sentence

p0mvn (approved):
Logic LGTM. Had 2 non-blocking test suggestions to ensure full coverage

osmoutils/module_account.go
```diff
@@ -0,0 +1,67 @@
...
+	existingAcct := ak.GetAccount(ctx, addr)
+	if existingAcct == nil {
+		return nil
+	}
+	if existingAcct.GetSequence() != 0 || existingAcct.GetPubKey() != nil {
+		return fmt.Errorf("cannot create module account %s, "+
+			"due to an account at that address already existing & having sent txs", addr)
+	}
+	var overrideAccountTypes = map[reflect.Type]struct{}{
+		reflect.TypeOf(&authtypes.BaseAccount{}):                 {},
```
alexanderbez, line 34: mhh why use reflect here instead of just directly creating the module account address?

osmoutils/module_account_test.go
```diff
@@ -0,0 +1,56 @@
...
+		"no prior acc": {
+			priorAccounts: []authtypes.AccountI{},
+			moduleAccAddr: defaultModuleAccAddr,
+			expErr:        false,
+		},
+		"prior empty acc at addr": {
+			priorAccounts: []authtypes.AccountI{baseWithAddr(defaultModuleAccAddr)},
+			moduleAccAddr: defaultModuleAccAddr,
+			expErr:        false,
+		},
```
czarcas7ic, line 38: should we have a test case for "prior empty acc at addr but > 0 sequence number" which we would then expect an error for?

czarcas7ic (approved):

osmoutils/module_account.go
ValarDragon: I want to check this off of the account type, right?

How would I do that without typecasting to everything (a hardcoded series of if statements) or runtime reflection, and type checking off of that?

osmoutils/module_account_test.go
ValarDragon: This currently lists all account types to my knowledge, intended to be future proof in case other account types get added that are user-space only, and for some reason don't obey sequence number / pubkey property.

osmoutils/module_account_test.go
ValarDragon: That test already exists! The user accs are done via that

osmoutils/module_account.go
ValarDragon: ty

mattverse (commented):
What would be the user flow when we're trying to create the next pool, but the pool address we're trying to create is already an existing account?

ValarDragon:
This PR specifically is meant for fixing that case, where under normal paths to claim that address, the pool will override it.

Issue #2727, opened 2022-09-13
Force unlock message


ValarDragon:
## Background

For Mars, we need to add a "ForceUnlock" message for LP's where they can break locks they own immediately. However permission for their address to use this needs to be given by governance, based upon contract audits.

## Suggested Design

 Design of force unlock:
* There is a chain parameter for the addresses that are allowed to have force unlock for their own locks
  * (Governance can update this list through either param change proposal, or custom proposal type. I'd prefer param change)
* We add a message called "ForceUnlock" which takes in a lock ID and amount, and asserts that:
  * lock owner = msg sender
  * lock is not superfluided
  * {partial unlocks that amount immediately}

## Acceptance Criteria

- [ ] Force unlock message exists
- [ ] Above constraints are met

PR #2733, opened 2022-09-14
Create Force Unlock Messages

mattverse:
Closes: #2727 

This PR creates a msg for force unlocking for the addresses that has been passed via governance and registered as params in the lockup keeper.

This PR also includes adding additional logic to the lockup keeper since we did not support partial unlockings for locks that are already  unlocking. ForceUnlock allows force unlocking of locks that are currently being unlocked.

## Brief Changelog

- Create ForceUnlock message for lockup keeper

## Testing and Verifying

Testing WIP: Planning to do PR it separately as test cases are getting big, would prefer separate review for the test cases / edge cases dealt within the test cases

Some of the tests being worked on:
- Test that is assets being superfluid delegated, undelegated
- Test partial force unlocking


proto/osmosis/lockup/tx.proto
```diff
@@ -71,3 +72,15 @@ message MsgExtendLockup {
 }
 
 message MsgExtendLockupResponse { bool success = 1; }
+
+message MsgForceUnlock {
```
alexanderbez, line 76: nit: add docs for request and response types :)

proto/osmosis/lockup/tx.proto
```diff
@@ -71,3 +72,15 @@ message MsgExtendLockup {
...
+  string owner = 1 [ (gogoproto.moretags) = "yaml:\"owner\"" ];
```
alexanderbez, line 77: Is this an SDK address? If so, shouldn't we add `(cosmos_proto.scalar) = "cosmos.AddressString"`? Maybe our proto tooling doesn't support this yet?

alexanderbez: Also, `  option (cosmos.msg.v1.signer) = "from_address";`

x/lockup/keeper/msg_server.go
```diff
@@ -182,3 +182,43 @@ func (server msgServer) ExtendLockup(goCtx context.Context, msg *types.MsgExtend
...
+
+	// force unlock given lock
+	// This also supports the case of force unlocking lock as a whole when msg.Coins
+	// provided is empty.
+	err = server.keeper.PartialForceUnlock(ctx, *lock, msg.Coins)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+
+	return &types.MsgForceUnlockResponse{Success: true}, nil
```
alexanderbez, line 223: When would we ever return false? Is it worth having the `Success` field? Are there other fields we might want?

alexanderbez (approved):
LGTM! I can't speak much to the `splitLock` implementation, but the utilization looks sound.

app/upgrades/v12/upgrades.go
```diff
@@ -83,6 +84,7 @@ func CreateUpgradeHandler(
 
 		// Set TWAP parameters to default values.
 		keepers.TwapKeeper.SetParams(ctx, twaptypes.DefaultParams())
+		keepers.LockupKeeper.SetParams(ctx, lockuptypes.DefaultParams())
```
p0mvn, line 87: We should probably make a v13 upgrade handler and add it there

x/lockup/types/params.go
```diff
@@ -0,0 +1,63 @@
...
+	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
+}
+
+func NewParams(forceUnlockAllowedAddresses []string) Params {
+	return Params{
+		ForceUnlockAllowedAddresses: forceUnlockAllowedAddresses,
+	}
+}
+
+// default lockup module parameters.
```
p0mvn, line 28:
```suggestion
// DefaultParams returns default lockup module parameters.
```

x/lockup/types/msgs.go
```diff
@@ -165,3 +166,33 @@ func (m MsgExtendLockup) GetSigners() []sdk.AccAddress {
...
+		Owner: owner.String(),
+		ID:    id,
+		Coins: coins,
+	}
+}
+
+func (m MsgForceUnlock) Route() string { return RouterKey }
+func (m MsgForceUnlock) Type() string  { return TypeMsgBeginUnlockingAll }
+func (m MsgForceUnlock) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Owner)
```
p0mvn, line 184: Should we also validate id and coins here?

x/lockup/keeper/keeper.go
```diff
@@ -17,21 +18,44 @@ type Keeper struct {
...
+
+// SetParams sets the total set of lockup parameters.
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
+	k.paramSpace.SetParamSet(ctx, &params)
+}
+
+func (k Keeper) GetForceUnlockAllowedAddresses(ctx sdk.Context) (forceUnlockAllowedAddresses []string) {
+	return k.GetParams(ctx).ForceUnlockAllowedAddresses
+}
+
```
p0mvn, line 58: non-blocking nit: I'm generally unsure if this method is needed if we can get the addresses from params

x/lockup/keeper/msg_server.go
```diff
@@ -182,3 +182,43 @@ func (server msgServer) ExtendLockup(goCtx context.Context, msg *types.MsgExtend
 
 	return &types.MsgExtendLockupResponse{}, nil
 }
+
+// ForceUnlock ignores unlock duration and immediately unlocks the lock.
+// This message is only allowed for governance passed accounts that is kept as parameter in the lockup module.
```
p0mvn, line 187:
```suggestion
// This message is only allowed for governance-passed accounts that are kept as parameter in the lockup module.
```

x/lockup/keeper/msg_server.go
```diff
@@ -182,3 +182,43 @@ func (server msgServer) ExtendLockup(goCtx context.Context, msg *types.MsgExtend
...
+	// check for chain parameter that the address is allowed to force unlock
+	forceUnlockAllowedAddresses := server.keeper.GetForceUnlockAllowedAddresses(ctx)
+	found := false
+	for _, address := range forceUnlockAllowedAddresses {
+		if address == msg.Owner {
+			found = true
+		}
+	}
+	if !found {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "Sender not allowed to force unlock")
```
p0mvn, line 201: consider formatting the sender address for ease of debugging in the future

p0mvn (approved):

proto/osmosis/lockup/tx.proto
mattverse: what does `(cosmos_proto.scalar) = "cosmos.AddressString"` and `option (cosmos.msg.v1.signer) = "from_address";` do?

Orignial intention was to recieve it as string and then convert it to address

x/lockup/keeper/msg_server.go
mattverse: would return false on the response whenever the transaction fails due to error in msg server, I think this one field would suffice for the field of this response!

x/lockup/types/msgs.go
mattverse: good catch! added

app/upgrades/v12/upgrades.go
mattverse: good catch! just created a v13 upgrade handler

x/lockup/keeper/keeper.go
mattverse: I originally thought this looked messy(having to call params to call get the addresses), but on the second thought I tink that way makes everything more intuitive, reverting this part so that we call from params directly

x/lockup/keeper/msg_server.go
mattverse: added!

proto/osmosis/lockup/params.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package osmosis.lockup;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/osmosis-labs/osmosis/v12/x/lockup/types";
+
+// Params defines the parameters for the tokenfactory module.
```
ValarDragon, line 8:
```suggestion
```

apollo-sturdy:
Hi. Wondering if there are any updates to this issue? Mainly curious if the API is finalized so that we (Apollo DAO) can implement this in our vault contract.

Issue #2688, opened 2022-09-09
test(twap): multi-asset tests for `updateRecords`


p0mvn:
## Background

We should add multi-asset pool tests to `updateRecords`

## Suggested Design

In #2687, we added comments specifying the test cases needed:
https://github.com/osmosis-labs/osmosis/blob/6edfbefcfa5548530126c40144833373606ea473/x/twap/logic_test.go#L900-L902

#2670 adds some useful helpers for multi-asset pool twaps so it's best to wait for it to be merged

## Acceptance Criteria
- specified test added
- helpers from #2670 are reused

hieuvubk:
@p0mvn should we separate the second one into 3 individual test cases?

PR #2735, opened 2022-09-14
multi-asset tests for updateRecords

hieuvubk:
Closes: #2688

> Update more tests for twap/updateRecords( multi-asset case)

- creates new records
- with errs


x/twap/logic_test.go
```diff
@@ -1044,9 +1043,85 @@ func (s *TestSuite) TestUpdateRecords() {
 				},
 			},
 		},
-		// TODO: complete multi-asset pool tests:
-		// "multi-asset pool; pre-set at t and t + 1; creates new records": {},
-		// "multi-asset pool; pre-set at t and t + 1; pre-existing records some with error and some with too large spot price, overwrites erorr time":                        {},
+		"multi-asset pool; pre-set at t and t + 1; creates new records": {
+			preSetRecords: []types.TwapRecord{threeAssetRecordAB, tPlus10sp5ThreeAssetRecordAB},
```
p0mvn, line 1047: For multi-asset pools, there should be at least 3 records at a given time.

Assume a 3 asset pool with assets A, B and C.

Then, if swap occurs at time t, we should have AB, AC, and BC records (for each pair).

Given that this test is meant for time t and t + 1, we need 6 records.

We should also have 3 spot price overrides and 3 most recent records.

p0mvn (changes requested):

x/twap/logic_test.go
hieuvubk: Ax thank you! Already updated could u pls check

x/twap/logic_test.go
```diff
@@ -1047,9 +1047,190 @@ func (s *TestSuite) TestUpdateRecords() {
...
+				// The original record BC at t + 1.
+				{
+					spotPriceA:    tPlus10sp5ThreeAssetRecordBC.P0LastSpotPrice,
+					spotPriceB:    tPlus10sp5ThreeAssetRecordBC.P1LastSpotPrice,
+				},
+				// The new record BC added.
+				{
+					spotPriceA:    sdk.OneDec(),
+					spotPriceB:    sdk.OneDec().Add(sdk.OneDec()),                            
+					isMostRecent:  true,
```
mattverse, line 1230: should have `lastErrorTime`?

x/twap/logic_test.go
```diff
@@ -1047,9 +1047,190 @@ func (s *TestSuite) TestUpdateRecords() {
...
+				},
+				// The new record BC added.
+				{
+					spotPriceA:   sdk.OneDec().Add(sdk.OneDec()),
+					spotPriceB:   sdk.OneDec(),
+					isMostRecent: true,
+				},
+			},
+		},
+		"multi-asset pool; pre-set at t and t + 1 with err, large spot price, overwrites erorr time": {
```
mattverse, line 1140:
```suggestion
		"multi-asset pool; pre-set at t and t + 1 with err, large spot price, overwrites error time": {
```

mattverse (commented):
The test cases and the comments lgtm! Just need a question answered on adding lastErrorTime for one of the expected test cases and should be 🌮

p0mvn (approved):
LGTM. Nice work

mattverse (approved):

mattverse:
Removed unnecessary whitelines within the test cases, merging this once final CI check has finished!

PR #2746, opened 2022-09-14
fix: MsgBeginUnlocking bug

czarcas7ic:
Closes: #XXX

As discovered on the v12 integrator testnet, attempting a begin-unlock-by-id (MsgBeginUnlocking) results in an error. This is because we pass a nil Coins object to MsgBeginUnlocking when we unlock by id, and the newly added ValidateBasic requires a length of 1 (which would be used if we are unlocking a specific subset of the lock)

## Testing and Verifying

### DIFF
--- a/x/lockup/types/msgs.go
+++ b/x/lockup/types/msgs.go
@@ -111,8 +111,8 @@ func (m MsgBeginUnlocking) ValidateBasic() error {
 		return fmt.Errorf("invalid lockup ID, got %v", m.ID)
 	}
 
-	// only allow unlocks with a single denom or nil
-	if m.Coins.Len() > 1 {
+	// only allow unlocks with a single denom or empty
+	if m.Coins.Len() != 1 || !m.Coins.Empty() {
 		return fmt.Errorf("can only unlock one denom per lock ID, got %v", m.Coins)
 	}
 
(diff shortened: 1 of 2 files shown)


x/lockup/types/msgs.go
```diff
@@ -111,8 +111,8 @@ func (m MsgBeginUnlocking) ValidateBasic() error {
 		return fmt.Errorf("invalid lockup ID, got %v", m.ID)
 	}
 
-	// only allow unlocks with a single denom
-	if m.Coins.Len() != 1 {
+	// only allow unlocks with a single denom or empty
+	if m.Coins.Len() != 1 || !m.Coins.Empty() {
```
stackman27, line 115: Doesnot 'empty' just mean `coins.len() == 0`?

czarcas7ic: yep, I am still working on this because the next check is also incorrect

p0mvn (approved):

ValarDragon:
Err, I don't think we should have merged this yet.

I don't think this state break is quite right, how are we getting locks with 0 coins in them

ValarDragon:
Ah I see, good fix! Concerning we d9nt have tests at the message layer that would've caught this

PR #2749, opened 2022-09-15
chore: remove cdc from keepers; switch store get/set to osmoutils (post v12)

p0mvn:
Closes: #XXX

Removed `cdc` parameter from all Osmosis keepers but `x/gamm`. Instead, switched to using `osmoutils` helpers.

`x/gamm` is more complicated because it marshals pool interface. Planning to create a separate issue to investigate the possibility of interface marshaling in gamm.

## Testing and Verifying


czarcas7ic (approved):
Nice!

p0mvn:
I think this should be fine to backport after v12.0.0

czarcas7ic:
Agreed, should be back ported post v12.0.0 tag

mattverse (approved):
Good catch on not removing cdc from the gamm keeper! Yeah, unmarshalling and marshalling `any` types or other interfaces with proto not codec can cause problems (we should document this somewhere, can you add it to the gamm keeper caller why we are still using cdc?)

Otherwise lgtm!

ValarDragon:
Is MustGet state machine compatible for both code of the form
```
t := types.MyType{}
bz := store.Get(...)
if len(bz) == 0 {
    panic(t)
}
k.cdc.Unmarshal(bz, &t)
return t
```
and
```
t := types.MyType{}
bz := store.Get(...)
k.cdc.Unmarshal(bz, &t)
return t
```

ValarDragon:
Adding backport v11.x so we can get the state machine compatability check to quickly confirm correctness

ValarDragon:
re my compatability question, could not figure out the answer from the source code, we should write a test to check

p0mvn:
> re my compatability question, could not figure out the answer from the source code, we should write a test to check

Good catch. Seems that they aren't

`Unmarshal` - basically calls each proto's Unmarshal method
https://github.com/osmosis-labs/cosmos-sdk/blob/870132d311d162fdbe0e92fcc0a42ee3889afd0d/codec/proto_codec.go#L80-L90

I looked at this example:
https://github.com/osmosis-labs/osmosis/blob/0df6256350406b45a35dab4081f9ce89392c3b58/x/incentives/types/gauge.pb.go#L702

That does not seem to return an error if the input is empty. As a result, it does not panic.

p0mvn:
Can we merge this?

PR #2750, opened 2022-09-15
review(x/twap): minor post-review clean up (post v-12)

p0mvn:
Closes: #XXX

- Remove 2 redundant iterations from `GetAllUniqueDenomPairs`
- Obsolete comment in `updateRecords`
- Redundant casts in `MigrateExistingPools`
- Redundant allocation of `sentinelExistsValue` on every call

All changes are super minor. This is a drive-by change for best practices. None are state-breaking.

## Testing and Verifying

### DIFF
--- a/x/twap/types/utils.go
+++ b/x/twap/types/utils.go
@@ -11,8 +11,9 @@ import (
 var MaxSpotPrice = sdk.NewDec(2).Power(128).Sub(sdk.OneDec())
 
 // GetAllUniqueDenomPairs returns all unique pairs of denoms, where for every pair
-// (X, Y), X >= Y.
+// (X, Y), X > Y.
 // The pair (X,Y) should only appear once in the list
+// Panics if finds duplicate pairs.
 //
 // NOTE: Sorts the input denoms slice.
 func GetAllUniqueDenomPairs(denoms []string) ([]string, []string) {
--- a/x/twap/types/utils.go
+++ b/x/twap/types/utils.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v12/osmoutils"
 )
 
 var MaxSpotPrice = sdk.NewDec(2).Power(128).Sub(sdk.OneDec())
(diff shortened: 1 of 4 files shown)


x/twap/types/utils.go
```diff
@@ -6,36 +6,35 @@ import (
...
-			panic("input had duplicated denom")
+	for i := 0; i < len(denoms); i++ {
+		for j := i + 1; j < len(denoms); j++ {
+			pairGT = append(pairGT, denoms[i])
+			pairLT = append(pairLT, denoms[j])
+
+			// sanity check
+			if pairGT[i] == pairLT[i] {
+				panic("input had duplicated denom")
+			}
```
mattverse, line 37: This check could be placed one step below(inside the first loop, not the second loop)

mattverse (approved):
Left a minor comment on the ordering, otherwise LGTM! The comments help alot

x/twap/types/utils.go
```diff
@@ -6,36 +6,35 @@ import (
...
+// Panics if finds duplicate pairs.
 //
 // NOTE: Sorts the input denoms slice.
 func GetAllUniqueDenomPairs(denoms []string) ([]string, []string) {
 	// get denoms in descending order
-	sort.Strings(denoms)
-	reverseDenoms := osmoutils.ReverseSlice(denoms)
+	sort.Slice(denoms, func(i, j int) bool {
+		return denoms[i] > denoms[j]
+	})
```
ValarDragon, line 23: Don't really get why we changed this / don't think its worth doing. I'd take the prior easier to read situation, this is really not relevant for performance

ValarDragon (approved):

x/twap/types/utils.go
ValarDragon: https://github.com/osmosis-labs/osmosis/pull/2765/files

Seems like the preferrable approach for this fn, perhaps we merge this, and then that one?

ValarDragon: So what we should do if we care about performance is use the normal sort, then reverse indices for GT and LT

p0mvn: I think we can instead merge #2765

Then, merge `main` into this PR. That would allow us to use `sort.Strings` here.

However, I'm not feeling strongly about this change and can revert it if you don't like. The performance benefit here is negligible since there are never many denoms

Issue #2251, opened 2022-07-28
investigate(x/gamm/twap): Reconsider using fixed size unix time serialization to save space in db


p0mvn:
**Background**

Example from one of the test cases:
```
"historical_pool_index|1|2009-11-10T23:00:00.000000001|"
```

We might save db space by choosing a shorter time serialization method.

There is no suggested design yet. The person taking on this issue should analyze the alternatives and summarize them here. If a new proposed design is accepted, it can be implemented.

Relevant Discussion:
https://github.com/osmosis-labs/osmosis/pull/2199#discussion_r928045908_

**Acceptance Criteria**
- investigate alternatives related to time serialization
- summarize the findings in the comment to this issue

p0mvn:
This is a nice to have but not a blocker

RusAkh:
Hi, Jacob on Ruslan's laptop here-- this is fascinating!

I think that there are two ways that we can address this:

https://github.com/osmosis-labs/osmosis/blob/99f770f2f7969472928967c55f3e702a6fbdbd55/x/twap/types/keys.go#L49-L73

Are a number of functions returning byte slices that are just strings.  So we could go content aware, and then use "whatever" serialization method, and all of the headers should be reduced to a single byte.

There is also a lazy method that could be used here-- 


https://github.com/klauspost/compress 

has the s2 compression library, which we could use for every such case on osmosis. 

snappy and s2 are very performant (cpu time for example) and also very effective (good compression).  It might not be add good as a serialization library, but is possibly simpler and gets (nearly) the same effect.

wdyt?

ValarDragon:
unfortunately I think we do need to close unfortunately. The TWAP code is code frozen aside from bug fixes, and this would require a state migration post-inclusion that seems not worth the marginal saving

Re: compression, we would need to ensure it preserves sort and length ordering guarantees. Which seems not worth it, and the tradeoff of fixed sized byte encoding seems totally ok

PR #2753, opened 2022-09-15
Prefixes change, use of unix time to save db space

RusAkh:
Closes: #2251 

Saving database space by using unix time rather than UTC + by using shorter prefixes

## Brief Changelog

Before: "historical_pool_index|1|2009-11-10T23:00:00.000000001|"
After: "HPInd|1|UNIX_TIME_HERE|"

Prefixes change:
historical_pool_index -> HPInd
historical_time_index -> HTInd
recent_twap -> RTwap
## Testing and Verifying


mattverse (commented):
Thanks for tackling this issue! 

I generally agree on using formatInt to reduce format time, but I feel a little conflicted on changing the seperator keys into short hand(considering that it get's complicated to debug with the prefixes + I don't think much storage would be reduced by changing seprators into shorthand). Also want to hear what other reviewers think about this

p0mvn:
Thanks for this work! Although I originally created this issue, I agree with Matt now.

Now, that v12 is finalized and code freezed, changing database keys would require migration for the subsequent releases. Given that this is not a current bottleneck, I think the migration overhead is not justified.

So I think we should keep the summary of the discoveries and this PR in mind if we ever need to optimize the database layout in the future

faddat:
cool!  So, there were a number of ideas we explored behind this-- for example, making a .proto file, since I believe that protobuf would compress these items natively.

We also tried the s2 compression library, and found it actually increased the size of data stored.

ValarDragon:
Agreed with @p0mvn, unfortunately I think we do need to close unfortunately.

The TWAP code is code frozen aside from bug fixes, and this would require a state migration post-inclusion that seems not worth the marginal saving

Re: compression, that fundamentally doesn't make sense here unfortunately, since we need to preserve sort and length ordering guarantees.

Re: prefix truncation: This is definitely not worth it imo, and is an optimization that should exist at the DB level for keys (which I believe most do, via tries )

PR #2759, opened 2022-09-15
Revert "Commit timeout (#2057)"

czarcas7ic:
This reverts commit 701b391044da52e67c673f1a5f1304b98bfaa1f0.

Closes: #XXX

The PR from #2057 reduced time timeout to two seconds for all nodes, regardless of config settings. After discussion, we decided this is not desirable. I did leave the change to init.go so that any future nodes initialized will default to 2 (but can overwrite via their config file)

## Testing and Verifying

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,6 @@ which acts as a fuzz testing tool tailored for the SDK state machine.
 ### Features
 
 * [#2387](https://github.com/osmosis-labs/osmosis/pull/2387) Upgrade to IBC v3.2.0, which allows for sending/receiving IBC tokens with slashes.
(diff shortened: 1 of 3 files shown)


faddat:
Interesting!

What's the thought here?

czarcas7ic:
Hey @faddat, sorry this was a last minute change from a discussion this morning. The change in its previous form would force all validators to run with the 2 seconds timeout. We instead thought it would be best to simply set the default to 2, and allow validators to change this at their desire.

ValarDragon:
yeah this should have been default config only.

ValarDragon (approved):

p0mvn (approved):

Issue #1685, opened 2022-06-06
[Task][Dockerfile Improvements] Generalize Docker files and reuse them where applicable


p0mvn:
## Background

Most, if not all of our Dockerfiles share the same lines of code at the beginning:
https://github.com/osmosis-labs/osmosis/blob/d5cd03329b7edc028f2793f95d4c4e85e62060fb/Dockerfile#L9-L28

This can be extracted into its own base image and reused across the rest.

## Status
- June 6 - WIP: Defining Scope
- Blocked by: #1684

## Acceptance Criteria
- TBD

ValarDragon:
I believe this has been done?

PR #2761, opened 2022-09-15
build: generalize Dockerfiles and reuse common build steps

p0mvn:
Closes: #1685

Reuse common Docker build steps across all Dockerfiles

## Testing and Verifying

Tested by:
```
docker image rm -f $(docker image ls -aq)

make docker-build-setup

make docker-build-builder

make docker-build-debug

make docker-build-e2e-init-chain

make docker-build-e2e-init-node
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 ARG BUILDER_VERSION="v1"
 ARG RUNNER_IMAGE="gcr.io/distroless/static"
 
-FROM osmosis-builder:${BUILDER_VERSION} as builder
+FROM osmolabs/osmosis-builder:${BUILDER_VERSION} as builder
 
 # --------------------------------------------------------
 # Runner
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,10 @@
 # syntax=docker/dockerfile:1
 
-ARG GO_VERSION="1.18"
+# Rebuild with make docker-build-builder
+ARG BUILDER_VERSION="v1"
 ARG RUNNER_IMAGE="gcr.io/distroless/static"
 
-# --------------------------------------------------------
-# Builder
-# --------------------------------------------------------
-
-FROM golang:${GO_VERSION}-alpine as builder
-
-RUN set -eux; apk add --no-cache ca-certificates build-base; apk add git linux-headers
-
-# Download go dependencies
-WORKDIR /osmosis
-COPY go.* .
-RUN --mount=type=cache,target=/root/.cache/go-build \
-    --mount=type=cache,target=/root/go/pkg/mod \
-    go mod download
-
-# Cosmwasm - download correct libwasmvm version
-RUN WASMVM_VERSION=$(go list -m github.com/CosmWasm/wasmvm | cut -d ' ' -f 2) && \
-    wget https://github.com/CosmWasm/wasmvm/releases/download/$WASMVM_VERSION/libwasmvm_muslc.$(uname -m).a \
-      -O /lib/libwasmvm_muslc.a
-
-# Cosmwasm - verify checksum
-RUN wget https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/checksums.txt -O /tmp/checksums.txt && \
-    sha256sum /lib/libwasmvm_muslc.a | grep $(cat /tmp/checksums.txt | grep $(uname -m) | cut -d ' ' -f 1)
-
-# Copy the remaining files
-COPY . .
-
-# Build osmosisd binary
-RUN --mount=type=cache,target=/root/.cache/go-build \
-    --mount=type=cache,target=/root/go/pkg/mod \
-    VERSION=$(echo $(git describe --tags) | sed 's/^v//') && \
-    COMMIT=$(git log -1 --format='%H') && \
-    go build \
-      -mod=readonly \
-      -tags "netgo,ledger,muslc" \
-      -ldflags "-X github.com/cosmos/cosmos-sdk/version.Name="osmosis" \
-              -X github.com/cosmos/cosmos-sdk/version.AppName="osmosisd" \
-              -X github.com/cosmos/cosmos-sdk/version.Version=$VERSION \
-              -X github.com/cosmos/cosmos-sdk/version.Commit=$COMMIT \
-              -X github.com/cosmos/cosmos-sdk/version.BuildTags='netgo,ledger,muslc' \
-              -w -s -linkmode=external -extldflags '-Wl,-z,muldefs -static'" \
-      -trimpath \
-      -o /osmosis/build/ \
-      ./...
+FROM osmosis-builder:${BUILDER_VERSION} as builder
 
 # --------------------------------------------------------
 # Runner
--- a/Makefile
+++ b/Makefile
@@ -311,16 +311,6 @@ build-e2e-script:
 	mkdir -p $(BUILDDIR)
 	go build -mod=readonly $(BUILD_FLAGS) -o $(BUILDDIR)/ ./tests/e2e/initialization/$(E2E_SCRIPT_NAME)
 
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -1,52 +1,10 @@
...
-    go mod download
-
-# Cosmwasm - download correct libwasmvm version
-RUN WASMVM_VERSION=$(go list -m github.com/CosmWasm/wasmvm | cut -d ' ' -f 2) && \
-    wget https://github.com/CosmWasm/wasmvm/releases/download/$WASMVM_VERSION/libwasmvm_muslc.$(uname -m).a \
-      -O /lib/libwasmvm_muslc.a
-
-# Cosmwasm - verify checksum
-RUN wget https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0/checksums.txt -O /tmp/checksums.txt && \
-    sha256sum /lib/libwasmvm_muslc.a | grep $(cat /tmp/checksums.txt | grep $(uname -m) | cut -d ' ' -f 1)
```
p0mvn, line 28: Note to reviewers: this part is moved untouched to `setup.Dockerfile`

Dockerfile
```diff
@@ -1,52 +1,10 @@
...
-      -tags "netgo,ledger,muslc" \
-      -ldflags "-X github.com/cosmos/cosmos-sdk/version.Name="osmosis" \
-              -X github.com/cosmos/cosmos-sdk/version.AppName="osmosisd" \
-              -X github.com/cosmos/cosmos-sdk/version.Version=$VERSION \
-              -X github.com/cosmos/cosmos-sdk/version.Commit=$COMMIT \
-              -X github.com/cosmos/cosmos-sdk/version.BuildTags='netgo,ledger,muslc' \
-              -w -s -linkmode=external -extldflags '-Wl,-z,muldefs -static'" \
-      -trimpath \
-      -o /osmosis/build/ \
-      ./...
```
p0mvn, line 49: Note to reviewers: this part is moved untouched to builder.Dockerfile

Makefile
```diff
@@ -311,16 +311,6 @@ build-e2e-script:
...
-docker-build-debug:
-	@DOCKER_BUILDKIT=1 docker build -t osmosis:${COMMIT} --build-arg BASE_IMG_TAG=debug -f Dockerfile .
-	@DOCKER_BUILDKIT=1 docker tag osmosis:${COMMIT} osmosis:debug
-
-docker-build-e2e-init-chain:
-	@DOCKER_BUILDKIT=1 docker build -t osmosis-e2e-init-chain:debug --build-arg E2E_SCRIPT_NAME=chain -f tests/e2e/initialization/init.Dockerfile .
-
-docker-build-e2e-init-node:
-	@DOCKER_BUILDKIT=1 docker build -t osmosis-e2e-init-node:debug --build-arg E2E_SCRIPT_NAME=node -f tests/e2e/initialization/init.Dockerfile .
-
```
p0mvn, line 323: Note to reviewers: this is moved down to combine with other Docker related makefile steps

ValarDragon:
cc @niccoloraspa

niccoloraspa:
I think this breaks the docker CI: https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/docker.yml

niccoloraspa:
I understand the idea behind the PR, but I have mixed feelings about it.

Personally I prefer to keep `Dockerfile` flat and self-contained, and splitting a `Dockerfile`  in `3` to favour reusability I think creates confusion.

We have made a lot of progress in reducing the number of `Dockerfiles` and this creates 2 additional Dockerfiles that we have to mantain. 

I think the correct approach would be to remove the 
Also the:
- `tests/localosmosis/mainnet_state/Dockerfile-stateExport` will be removed soon 

I think the correct approach would be to remove the `e2e` Dockerfile favouring the re-use of the main Dockerfile.

niccoloraspa:
I have mixed feelings about this PR.

I don't think splitting the main `Dockerfile` into 3 to favor re-usability is the right approach. We have made a lot of progress in reducing the number of `Dockerfiles`, and this creates 2 additional `Dockerfiles` that we have to maintain.

I think the right approach would be to re-use the main `Dockerfile` in `e2e`. 
The `tests/localosmosis/mainnet_state/Dockerfile-stateExport` will be removed soon.

niccoloraspa:
I prefer `Dockerfile` to be as flat and self-contained as possible. 
A `Dockerfile` is like a contract between you and the users who run it. The simplicity favours the transparency.

p0mvn:
> I don't think splitting the main Dockerfile into 3 to favor re-usability is the right approach. We have made a lot of progress in reducing the number of Dockerfiles, and this creates 2 additional Dockerfiles that we have to maintain.

The core issue this is solving is that there are several Docker files that utilize the same logic.

As a result, when one gets updated, all other Dockerfiles aren't. They end up going out of sync. For example, both `init.Dockerfile` and `Dockerfile-stateExport` are out of sync from the main one. It also happened so that not all Dockerfiles being updated broke scripts and tests for us in the past. That's why encapsulating common Docker logic is useful.

Additionally, it is an overhead to update them in multiple places every time.

Analogous to how we encapsulate duplicated code into functions, we should do the same with Dockerfile for the same reasons of them going out of sunc and having to update them in multiple places.

> this creates 2 additional Dockerfiles that we have to maintain.

I think it is easier to maintain 2 extra Dockerfiles that are reused by 3 others rather than having 3 Dockerfiles with the same logic.
In the former case, one change would need to be made to 1 file, and in the latter case to 3 files.

> I think the right approach would be to re-use the main Dockerfile in e2e

In e2e, we are running a script, not an `osmosisd` binary. Same as in the state export Dockerfile. However, the setup logic is exactly the same.

It is not possible to utilize the main dockerfile as a builder because its last build step is a runner.

> FYI the tests/localosmosis/mainnet_state/Dockerfile-stateExport will also be removed soon.

That's great! In that case, it would be possible to consolidate `setup.Dockerfile` and `builder.Dockerfile` into one.

Currently, I still think that this change introduces an improvement rather than additional overhead. Please let me know what you think

ValarDragon:
I generally agree with the assesment that there should really just be one "build" dockerfile, and other things should just use that if they have extra steps.

niccoloraspa:
> Analogous to how we encapsulate duplicated code into functions, we should do the same with Dockerfile for the same reasons of them going out of sunc and having to update them in multiple places.

Modularity's great but for me, keeping the main Dockerfile transparent and self-contained is a priority over the rest.

Also, keep in mind that `Dockerfile-stateExport` will leave only:

- the main `Dockerfile` 
- the e2e `init.Dockerfile` 

Afaik, these two only share the "setup" part (downloading `CosmWasm` and some packages). I think the intersection would be too small to justify splitting them.

p0mvn:
> > Analogous to how we encapsulate duplicated code into functions, we should do the same with Dockerfile for the same reasons of them going out of sunc and having to update them in multiple places.
> 
> Modularity's great but, for me, keeping the main Dockerfile transparent and self-contained is a priority over the rest.
> 
> Also, keep in mind that removing the `Dockerfile-stateExport` leaves only:
> 
> * the main `Dockerfile`
> * the e2e `init.Dockerfile`
> 
> Afaiu, these two only share the "setup" part (downloading `CosmWasm` and some packages) and then they continue to build their own binary. I think the intersection would be too small to justify splitting them.
> 
> P.S. You can run scripts with the main Dockerfile if needed. This is what localOsmosis is doing: https://github.com/osmosis-labs/osmosis/blob/main/tests/localosmosis/docker-compose.yml#L17

Sounds good, thanks! Let me try running the script via a different entrypoint. If that works, we can drop this common builder.

I do think, however, that it is inevitable to end up splitting them as the system grows and new Dockerfiles continue requiring custom functionality on top of the builder

PR #2764, opened 2022-09-16
Fix: Remove balance check from EstimateSwap

mattverse:
Closes: #2337 

EstimateSwap currently checks the balance of the query sender, not because it actually moves the assets on an actual swap, but beacuse it uses the same API with the actual multi swaps that happen. 

This PR separates out the API needed for the estimation so that it does not check balance of the query sender. 

(Special thanks to @vuong177 the original author of this PR!)

## Brief Changelog

- Remove balance check from `EstimateSwap`

## Testing and Verifying
This change added tests. 

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.mxd`? yes


mattverse:
Go Test CI is failing rn because we need a new wasm reflect binary

wasmbinding/message_plugin.go
```diff
@@ -218,50 +218,22 @@ func PerformSwap(keeper *gammkeeper.Keeper, ctx sdk.Context, contractAddr sdk.Ac
...
-			return nil, wasmvmtypes.InvalidRequest{Err: "gamm perform swap negative amount in"}
-		}
-		tokenIn := sdk.Coin{
-			Denom:  swap.First.DenomIn,
-			Amount: swap.Amount.ExactIn.Input,
+		routes, tokenIn, tokenOutMinAmount, err := getSwapExactAmountInParams(swap)
+		if err != nil {
+			return nil, err
 		}
-		tokenOutMinAmount := swap.Amount.ExactIn.MinOutput
```
ValarDragon, line 238: Confirmed this is just a move

wasmbinding/message_plugin.go
```diff
@@ -218,50 +218,22 @@ func PerformSwap(keeper *gammkeeper.Keeper, ctx sdk.Context, contractAddr sdk.Ac
...
-			})
-			output = step.DenomOut
-		}
-		tokenInMaxAmount := swap.Amount.ExactOut.MaxInput
-		if swap.Amount.ExactOut.Output.IsNegative() {
-			return nil, wasmvmtypes.InvalidRequest{Err: "gamm perform swap negative amount out"}
-		}
-		tokenOut := sdk.Coin{
-			Denom:  output,
-			Amount: swap.Amount.ExactOut.Output,
```
ValarDragon, line 263: confirmed just a move

ValarDragon (commented):
Reviewed all of the wasmbinding and proto changes, looks great to me! Reviewing keeper logic now

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,256 @@
...
+		return pool, sdk.Coin{}, errors.New("cannot trade same denomination in and out")
+	}
+
+	poolOutBal := pool.GetTotalPoolLiquidity(ctx).AmountOf(tokenOut.Denom)
+	if tokenOut.Amount.GTE(poolOutBal) {
+		return pool, sdk.Coin{}, sdkerrors.Wrapf(types.ErrTooManyTokensOut,
+			"can't get more tokens out than there are tokens in the pool")
+	}
+
+	tokenIn, err = pool.SwapInAmtGivenOut(ctx, sdk.Coins{tokenOut}, tokenInDenom, swapFee)
```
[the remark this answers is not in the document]
ValarDragon: Oh hrm, this isn't so simple.

Ok I think the current method is actually better. Where the thinking is "this state may be  altered, but its up to the caller to ensure this updated state does not get saved (e.g. is in a cache ctx)"

The alternative to get code-reuse would be a deeper refactor, which does seem a bit inadvisable for this PR / release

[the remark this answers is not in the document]
mattverse: yeah exactly!

ValarDragon:
(Had a call to rename the internal `estimate` methods to better indicate that they can really change state, and move `cacheCtx` from query into the go API for public Estimate methods

x/gamm/keeper/estimate_swap.go
```diff
@@ -188,68 +154,11 @@ func (k Keeper) EstimateSwapExactAmountOut(
...
-		return pool, sdk.Coin{}, sdkerrors.Wrapf(types.ErrLimitMaxAmount, "Swap requires %s, which is greater than the amount %s", tokenIn, tokenInMaxAmount)
-	}
-
-	return pool, tokenIn, nil
-}
-
-// estimateUpdatePoolForSwap updates the pool balance.
-// This method does not include emitting hooks or sending the actual asset from
-// the sender to the pool or vice versa, as this is designed to be used by query methods.
-func (k Keeper) estimateUpdatePoolForSwap(
```
mattverse: changed to method `updatePoolForSwap` in swap.go, as this should also not be an estimating method

wasmbinding/queries.go
```diff
@@ -95,19 +95,48 @@ func (qp QueryPlugin) EstimateSwap(ctx sdk.Context, estimateSwap *bindings.Estim
...
 		return nil, wasmvmtypes.InvalidRequest{Err: "gamm estimate swap empty swap"}
 	}
 
-	estimate, err := PerformSwap(qp.gammKeeper, ctx, senderAddr, estimateSwap.ToSwapMsg())
+	subCtx, _ := ctx.CacheContext()
+	estimate, err := EstimatePerformSwap(qp.gammKeeper, subCtx, estimateSwap.ToSwapMsg())
 	return estimate, err
 }
 
+// EstimatePerformSwap can be used to query
```
p0mvn, line 108: nit: unfinished sentence

x/gamm/keeper/grpc_query_test.go
```diff
@@ -283,3 +283,124 @@ func (suite *KeeperTestSuite) TestQueryBalancerPoolSpotPrice() {
...
+	poolID := suite.PrepareBalancerPool()
+
+	testCases := []struct {
+		name      string
+		req       *types.QuerySwapExactAmountInRequest
+		expectErr bool
+		result    sdk.Int
+	}{
+		{
+			name: "multi route",
```
p0mvn, line 298: non-blocking comment / feel free to make an issue if you agree: only testing happy path here, we might want to test error cases as well

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v12/x/gamm/types"
+)
+
+// EstimateMultihopSwapExactAmountIn iterates `EstimateSwapExactAmountIn` and returns
+// the total token out given route for multihop.
+func (k Keeper) EstimateMultihopSwapExactAmountIn(
```
p0mvn, line 11: There seems to be a lot of code copied over from non-estimate versions of this method. I'm wondering if it's possible to generalize this somehow so that we are not repeating so much boilerplate.

x/gamm/keeper/swap.go
```diff
@@ -45,35 +45,59 @@ func (k Keeper) swapExactAmountIn(
...
+	ctx sdk.Context,
+	pool types.PoolI,
+	tokenIn sdk.Coin,
+	tokenOutDenom string,
+	tokenOutMinAmount sdk.Int,
+	swapFee sdk.Dec,
+) (updatedPool types.PoolI, tokenOut sdk.Coin, err error) {
 	if tokenIn.Denom == tokenOutDenom {
-		return sdk.Int{}, errors.New("cannot trade same denomination in and out")
+		return pool, sdk.Coin{}, errors.New("cannot trade same denomination in and out")
```
p0mvn, line 79:
```suggestion
		return nil, sdk.Coin{}, errors.New("cannot trade same denomination in and out")
```

x/gamm/keeper/swap.go
```diff
@@ -45,35 +45,59 @@ func (k Keeper) swapExactAmountIn(
...
 	}
 	tokensIn := sdk.Coins{tokenIn}
 
 	// Executes the swap in the pool and stores the output. Updates pool assets but
 	// does not actually transfer any tokens to or from the pool.
 	tokenOutCoin, err := pool.SwapOutAmtGivenIn(ctx, tokensIn, tokenOutDenom, swapFee)
 	if err != nil {
-		return sdk.Int{}, err
+		return pool, sdk.Coin{}, err
```
p0mvn, line 87:
```suggestion
		return nil, sdk.Coin{}, err
```

x/gamm/keeper/swap.go
```diff
@@ -45,35 +45,59 @@ func (k Keeper) swapExactAmountIn(
...
 	}
 
-	tokenOutAmount = tokenOutCoin.Amount
+	tokenOutAmount := tokenOutCoin.Amount
 
 	if !tokenOutAmount.IsPositive() {
-		return sdk.Int{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
+		return pool, sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
```
p0mvn, line 93:
```suggestion
		return nil, sdk.Coin{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "token amount must be positive")
```

x/gamm/keeper/swap.go
```diff
@@ -45,35 +45,59 @@ func (k Keeper) swapExactAmountIn(
...
+// and potentially mutates state. It does not save the new pool struct, or do token transfers.
+// Returns the altered pool after swap, and the token out when swapped.
+func (k Keeper) swapExactAmountInNoTokenSend(
+	ctx sdk.Context,
+	pool types.PoolI,
+	tokenIn sdk.Coin,
+	tokenOutDenom string,
+	tokenOutMinAmount sdk.Int,
+	swapFee sdk.Dec,
+) (updatedPool types.PoolI, tokenOut sdk.Coin, err error) {
```
p0mvn, line 77: Do we even need to return `updatedPool` here? `pool` is a pointer so it should be mutated if that's desired

x/gamm/keeper/swap.go
```diff
@@ -105,44 +129,66 @@ func (k Keeper) swapExactAmountOut(
...
+// and potentially mutates state. It does not save the new pool struct, or do token transfers.
+// Returns the altered pool after swap, and the token in when swapped.
+func (k Keeper) swapExactAmountOutNoTokenSend(
+	ctx sdk.Context,
+	pool types.PoolI,
+	tokenInDenom string,
+	tokenInMaxAmount sdk.Int,
+	tokenOut sdk.Coin,
+	swapFee sdk.Dec,
+) (updatedPool types.PoolI, tokenIn sdk.Coin, err error) {
```
p0mvn, line 160: Same question about `updatedPool`

x/gamm/keeper/swap.go
```diff
@@ -105,44 +129,66 @@ func (k Keeper) swapExactAmountOut(
...
 	if tokenInDenom == tokenOut.Denom {
-		return sdk.Int{}, errors.New("cannot trade same denomination in and out")
+		return pool, sdk.Coin{}, errors.New("cannot trade same denomination in and out")
```
p0mvn, line 162: We should also consider returning `nil` in all error cases

x/gamm/keeper/grpc_query_test.go
mattverse: added test cases for empty requests, I don't think error catching paths for estimate swap itself have to go here

x/gamm/keeper/estimate_swap.go
mattverse: It ranges over the same for loop for the multihops, the only difference is that it would call different internal methods, `swapExactAmountOut` for the actual swap and`estimateSwapExactAmountOut` for the estimated multihops.

This could be solved by passing in the internal func we want to use as a function parameter, but I think such refactor could be done in a separate PR

x/gamm/keeper/swap.go
mattverse: ur right, they do get mutated if desired, removing returning the updatedPool

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	tokenOut, err := k.swapExactAmountInNoTokenSend(ctx, pool, tokenIn, tokenOutDenom, tokenOutMinAmount, swapFee)
+	if err != nil {
+		return tokenOut.Amount, err
+	}
+
+	err = k.updatePoolForSwap(ctx, pool, tokenIn, tokenOut)
+	return tokenOut.Amount, err
+}
```
czarcas7ic, line 94: Can we make it clearer what the diff is between these two? Both say "this method does not include full steps of an actual swap", but it seems like if cache context is false then we actually mutate state

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,200 @@
...
+	insExpected, err := k.createMultihopExpectedSwapOuts(cacheCtx, routes, tokenOut, swapFeeMultiplier)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	if len(insExpected) == 0 {
+		return sdk.Int{}, nil
+	}
+
+	insExpected[0] = tokenInMaxAmount
+
```
czarcas7ic, line 122: Same about these needing to change with this PR https://github.com/osmosis-labs/osmosis/pull/3352

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,200 @@
...
+	useCacheCtx bool,
+) (tokenInAmount sdk.Int, err error) {
+	if useCacheCtx {
+		ctx, _ = ctx.CacheContext()
+	}
+	pool, err := k.getPoolForSwap(ctx, poolId)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+
```
czarcas7ic, line 192: Same point about comments for cacheCtx

x/gamm/keeper/estimate_swap_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+			// calculate token out amount using `EstimateMultihopSwapExactAmountIn`
+			estimateMultihopTokenOutAmount, err := keeper.EstimateMultihopSwapExactAmountIn(
+				suite.Ctx,
+				test.param.estimateRoutes,
+				test.param.tokenIn,
+				test.param.tokenOutMinAmount)
+			suite.Require().NoError(err)
+
+			// assure that the token out amount is same
```
czarcas7ic, line 136:
```suggestion
			// ensure that the token out amount is same
```

czarcas7ic (commented):
Left some comments, main thing is changing the multi-hop logic to match the new design here https://github.com/osmosis-labs/osmosis/pull/3352

Also, unclear what is going on with the proto gen CI check

wasmbinding/message_plugin.go
```diff
@@ -218,50 +218,22 @@ func PerformSwap(keeper *gammkeeper.Keeper, ctx sdk.Context, contractAddr sdk.Ac
...
-			Amount: swap.Amount.ExactIn.Input,
+		routes, tokenIn, tokenOutMinAmount, err := getSwapExactAmountInParams(swap)
+		if err != nil {
+			return nil, err
 		}
-		tokenOutMinAmount := swap.Amount.ExactIn.MinOutput
+
 		tokenOutAmount, err := keeper.MultihopSwapExactAmountIn(ctx, contractAddr, routes, tokenIn, tokenOutMinAmount)
 		if err != nil {
 			return nil, sdkerrors.Wrap(err, "gamm perform swap exact amount in")
```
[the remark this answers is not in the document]
mattverse: lol I actually do agree with you that its pretty unclear, but also didnt want to touch anything outside the scope of this PR. But sinced its addressed, changing this to something more intuitive

wasmbinding/queries.go
```diff
@@ -95,19 +95,49 @@ func (qp QueryPlugin) EstimateSwap(ctx sdk.Context, estimateSwap *bindings.Estim
...
 	return estimate, err
 }
 
+// EstimatePerformSwap can be used to either estimate a exact amount in swap, or an exact amount out swap.
+// This function requires requires either swap.Amount.ExactIn or swap.Amount.ExactOut to be non nil, or else it will error.
+func EstimatePerformSwap(keeper *gammkeeper.Keeper, ctx sdk.Context, swap *bindings.SwapMsg) (*bindings.SwapAmount, error) {
+	if swap == nil {
+		return nil, wasmvmtypes.InvalidRequest{Err: "gamm perform swap null swap"}
+	}
+	if swap.Amount.ExactIn != nil {
```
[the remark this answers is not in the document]
mattverse: Why do you think that's the case? AFAIU, i think `swap.Amount.ExactIn != nil && swap.Amount.ExactOut == nil` can be a possible scenario, for example, when a pool is 1:1 ratio(10 uosmo, 10 uatom), you could be expecting 1 in 1out or similar

x/gamm/keeper/estimate_swap.go
mattverse: Ahh, yeah this comment does need some better explanation. Added some more in line comments on this, tell me if it makes more sense now!

x/gamm/keeper/estimate_swap.go
mattverse: ditto for the comment above

x/gamm/keeper/estimate_swap.go
mattverse: Merged this with the most recent main and then made the according changes, can you review it and confirm?

x/gamm/keeper/estimate_swap.go
```diff
@@ -0,0 +1,200 @@
...
+	tokenOutMinAmount sdk.Int,
+) (tokenOutAmount sdk.Int, err error) {
+	// use cache context so that pool state is not mutated after estimation
+	cacheCtx, _ := ctx.CacheContext()
+	for i, route := range routes {
+		swapFeeMultiplier := sdk.OneDec()
+		if types.SwapAmountInRoutes(routes).IsOsmoRoutedMultihop() {
+			swapFeeMultiplier = types.MultihopSwapFeeMultiplierForOsmoPools.Clone()
+		}
+
```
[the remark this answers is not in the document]
mattverse: Merged this with the most recent main and then made the according changes, can you review it and confirm?

x/gamm/keeper/estimate_swap.go
mattverse: https://github.com/osmosis-labs/osmosis/issues/3386

wasmbinding/queries.go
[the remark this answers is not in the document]
czarcas7ic: Right, not only CAN it be a possible scenario, it should be the only scenario for the one sided argument. What I am trying to prevent here is if someone provides a SwapMsg with both a swap.Amount.ExactIn and swap.Amount.ExactOut, instead of erroring they would go through the first for loop. EstimatePerformSwap should only return a valid response if one side is given, not both sides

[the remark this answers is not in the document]
czarcas7ic: AFAIU, we should never call an estimate method with both the in and the out. We either give it the in to calculate the out or we give it the out to calculate the in

czarcas7ic:
Wait, but wouldn't it be as simple as adding a "useCacheCtx" boolean here https://github.com/osmosis-labs/osmosis/blob/bef8a4b3b107b383676d953a9e89e409b2e4f828/x/gamm/keeper/multihop.go#L13-L19 and then we could get rid of the entire estimate_swap.go file?

mattverse:
@czarcas7ic I actaully think that's a great idea. Removed the whole file of estimate swap in my last commit and changed the existing swap methods to have a boolean filed for the estimate swaps.

I think this requires detailed, in-depth review as it touches the core swap logic

czarcas7ic:
@mattverse will review in depth tomorrow!

p0mvn:
Is this v13 blocking?

p0mvn:
NVM, I see that we went with the code duplication approach for safety here: https://github.com/osmosis-labs/osmosis/pull/3424

wasmbinding/queries.go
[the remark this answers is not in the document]
czarcas7ic: I still feel this is the case

ValarDragon:
@mattverse should we close this PR, and re-PR with elements you want added in still?

ValarDragon:
Closing per last comment

Issue #2497, opened 2022-08-23
Domain separate event types by module name Overlaps


czarcas7ic:
## Background

While doing mutation tests on tokenfactory, I realized BurnDenom was emitting two events. After looking into this with Roman, we realized that this was due to overlap between the sdk's bank "burn" message and the tokenfactory "burn" message. To fix this, I changed the tokenfactory message name from "burn" to "tf_burn".

## Suggested Design

As Roman suggested, it would probably be best for all module messages to have a module name prefix to prevent overlaps like this in the future. This is especially true for messages that might exist between multiple modules (such as burning and minting)

## Acceptance Criteria

TBD

hieuvubk:
Is anyone working on this issue?? I could close it

czarcas7ic:
@hieuvubk no one is working on this! I will assign you now, thanks!

hieuvubk:
@czarcas7ic  With the `tokenfactory` module, we use TypeMsg to emit events and msg names both, but unlike other modules which have their own msg name, their own event name. With such modules, should we just change the events name

czarcas7ic:
@hieuvubk sorry I am not following, can you explain further?

hieuvubk:
> @hieuvubk sorry I am not following, can you explain further?

https://github.com/osmosis-labs/osmosis/blob/main/x/tokenfactory/keeper/msg_server.go#L67-L72
Here use msg type to emit events.

And the difference:
https://github.com/osmosis-labs/osmosis/blob/main/x/superfluid/keeper/internal/events/emit.go#L23-L27

https://github.com/osmosis-labs/osmosis/blob/main/x/superfluid/types/events.go#L7
https://github.com/osmosis-labs/osmosis/blob/main/x/superfluid/types/msgs.go#L12

So I mean we just need to change the TypeEvt right?

PR #2767, opened 2022-09-16
Separate events by module

hieuvubk:
Closes: #2497

## Brief Changelog
 
  - *Add module name prefix before event & msg name*


ValarDragon:
This an integrator breaking change that I do not think is worth it, if hand done.

This should come from the SDK framework, and is blocked on the SDK adding a simple run-time for modules

PR #2771, opened 2022-09-17
lint: tighten linting

faddat:
Closes: #1939 

* This PR will tighten linting across osmosis v12. 
* style
  * I changed instances of Asset and Basset to assetA and assetB since they didn't need to be exported, and the latter reads more easily. 

## Brief Changelog

* enable revive linter, with below exclusions:

```yaml
  exclude:
    - "var-naming"
```

## Testing and Verifying

This PR should not change any APIs or the function of any code.  If it does, it should be rejected. 

  - How is the feature or change documented? in changelog

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,81 +10,46 @@ linters:
   enable:
     - asciicheck
     - bidichk
-    #    - bodyclose <- needs go 1.18 support
-    # - contextcheck <- needs go 1.18 support
-    #    - cyclop
     - depguard
-    #    - dogsled <- disabled because we would like to participate in the iditerod
-    # - dupl <- really just catches cli stub text
     - durationcheck
     - errcheck
     - errname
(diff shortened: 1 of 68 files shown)


faddat:
@ValarDragon @p0mvn -- hey :D.

So, I removed the exclusion of stableswap from the unused linter, but there and still many things unused.  

I'm thinking that I should tag them like:

```go
//nolint:unused
```

ValarDragon:
I found the golangci lint comments quite helpful!

I also don't agree with most of the changes as net beneficial. (Though I also don't think deliberation around them is particularly helpful either)

faddat:
I agree about not deliberating, so if this doesn't seem real useful, I can just close this, OR I can revert some of the changes and comment them.

ValarDragon:
Want to PR just revive + your exclusions?

Issue #2776, opened 2022-09-19
[stableswap]: Convert all core stableswap arithmetic to use bigdec


AlpinYukseloglu:
## Background

The SDK's decimal bitlength and precision are currently insufficient to meet our needs for our stableswap CFMM and the SDK currently does not support conversion to arbitrary precision decimals, so we need to move our core stableswap arithmetic over to our `BigDec` type where we can safely change bitlength and decimal precision without risking breaking a bunch of things.

## Suggested Design

- Keep the pool interface the same, and simply convert incoming decimals to `BigDec` and truncate additional precision values from outgoing decimals (e.g. if `BigDec` has 19 decimal places and `sdk.Dec` has 18, truncate the last decimal value when converting from `BigDec` to `sdk.Dec`)

## Acceptance Criteria

- All core stableswap arithmetic is switched to using `BigDec`
- All current stableswap tests pass and nothing gamm-level breaks

PR #2777, opened 2022-09-19
[stableswap]: Convert all core stableswap arithmetic to use `BigDec`

AlpinYukseloglu:
Closes: #2776 

This PR updates all core arithmetic operations in stableswap to use our `BigDec` type instead of the regular `sdk.Dec` type.

## Brief Changelog

- Update all uses of `sdk.Dec` for internal calculations to use `BigDec`
- Add methods for converting from `sdk.Dec` to `BigDec` and vice versa to our `BigDec` implementation

## Testing and Verifying

- The core change is currently a trivial refactor as `sdk.Dec` and `BigDec` have the same precision, but this change will require more thorough testing once we increase precision bounds in `BigDec`
- The conversion methods for `sdk.Dec` <-> `BigDec` only have basic sanity tests, as testing them more thoroughly requires finalizing the specific decimal precision we want for `BigDec` (left for a future PR to not block this one)


ValarDragon (approved):

osmomath/decimal_test.go
```diff
@@ -148,6 +149,56 @@ func (s *decimalTestSuite) TestDecFloat64() {
...
+		expPanic bool
+	}{
+		{NewBigDec(0), sdk.MustNewDecFromStr("0.000000000000000000"), false},
+		{NewBigDec(1), sdk.MustNewDecFromStr("1.000000000000000000"), false},
+		{NewBigDec(10), sdk.MustNewDecFromStr("10.000000000000000000"), false},
+		{NewBigDec(12340), sdk.MustNewDecFromStr("12340.000000000000000000"), false},
+		{NewDecWithPrec(12340, 4), sdk.MustNewDecFromStr("1.234000000000000000"), false},
+		{NewDecWithPrec(12340, 5), sdk.MustNewDecFromStr("0.123400000000000000"), false},
+		{NewDecWithPrec(12340, 8), sdk.MustNewDecFromStr("0.000123400000000000"), false},
+		{NewDecWithPrec(1009009009009009009, 17), sdk.MustNewDecFromStr("10.090090090090090090"), false},
```
ValarDragon, line 165: Doesn't this test case seem wrong?

osmomath/decimal.go
```diff
@@ -494,6 +496,27 @@ func (d BigDec) MustFloat64() float64 {
...
+// Values in any additional decimal places are truncated.
+func (d BigDec) SdkDec() sdk.Dec {
+	precisionDiff := Precision - sdk.Precision
+	precisionFactor := new(big.Int).Exp(big.NewInt(10), big.NewInt(int64(precisionDiff)), nil)
+
+	if precisionDiff < 0 {
+		panic("invalid decimal precision")
+	}
+
+	truncatedDec := sdk.NewDecFromBigIntWithPrec(new(big.Int).Quo(d.BigInt(), precisionFactor), sdk.Precision)
```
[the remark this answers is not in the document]
AlpinYukseloglu: Yes, this is something we need to discuss, in addition to the potential issues related to the ultimate conversion back to SDK dec at the keeper level (since even though stableswap calcs use `BigDec` the pool interface remains unchanged with sdk.Dec)

The approach I've implemented here basically truncates any decimal places that go beyond sdk.Dec's precision (e.g. if `BigDec` has 19 decimal precision and `sdk.Dec` has 18, the 19th decimal is truncated). I defaulted to this to avoid having the pool send more tokens out than it should, but I'm not convinced this is a robust solution

osmomath/decimal_test.go
AlpinYukseloglu: Definitely an odd case yes but it does seem correct to me (note that `NewDecWithPrec` shifts by 17 places, not 18 or 36). Since both the expected and actual values in each case pad zeroes for unused precision at the end these cases seem to be working as intended

osmomath/decimal.go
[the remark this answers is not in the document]
ValarDragon: I think truncation is the right thing to do as well! Its just not obvious to me from the code + test (perhaps I'm missing something though)

Are we relying on big.Int Quo to be floor division? (If so we should add a comment for this. And move it into a couple lines)

osmomath/decimal_test.go
ValarDragon: Err, isn't LHS: `1.009009009009009009`, and RHS `10.090090090090090090`, which is different values?

p0mvn (approved):

osmomath/decimal_test.go
p0mvn: Looks good to me. I think LHS is `10.09009009009009009` because of the precision of 17. RHS seems to only add an extra zero on the right since precision is 18

alexanderbez (approved):

osmomath/decimal.go
```diff
@@ -494,6 +496,27 @@ func (d BigDec) MustFloat64() float64 {
 	}
 }
 
+// SdkDec returns the Sdk.Dec representation of a BigDec.
+// Values in any additional decimal places are truncated.
+func (d BigDec) SdkDec() sdk.Dec {
```
alexanderbez, line 501: nit: Would ensure all `sdk` usage in methods and functions is named `SDK`

osmomath/decimal_test.go
AlpinYukseloglu: @ValarDragon https://www.wolframalpha.com/input?i2d=true&i=Divide%5B1009009009009009009%2CPower%5B10%2C17%5D%5D
Screen Shot 2022-09-19 at 8 00 03 AM

osmomath/decimal.go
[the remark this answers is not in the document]
AlpinYukseloglu: Yes, we do rely on `big.Int` Quo being floor div (which it is) – I'll add a comment and spread the calculation out to be more readable!

osmomath/decimal.go
AlpinYukseloglu: Done!

Issue #2757, opened 2022-09-15
Properly export incentives


czarcas7ic:
## Background

When creating a state exported testnet, AFAIK incentives are not properly exported. This leaves us with null upcoming gauges and we cannot properly QA the frontend.

## Suggested Design

Investigate export for incentives module.

## Acceptance Criteria

Incentives are properly exported for state exported testnets.

czarcas7ic:
Correction: pool-incentives

But I did notice that we don't export last gauge ID in incentives which could also be problematic

czarcas7ic:
I looked into this a bit further and it appears the pool incentives are being properly exported. so the issue lies in the import

czarcas7ic:
![Screen Shot 2022-09-15 at 3 29 38 PM](https://user-images.githubusercontent.com/40078083/190502669-b12164fd-54cc-4a6c-98a6-b959c0d9caf5.png)

Further narrowing down the problem, the gauge ID exists in incentives, but is not properly linked to its respective pool in pool-incentives

PR #2787, opened 2022-09-20
x/incentives: gauge method patch

czarcas7ic:
Closes: #2757

When running state exported testnets, we ran into an issue where upcoming gauges would begin distributing after the first epoch post state export. It turns out the method we were using to determine Active or Upcoming was incorrect.

## Brief Changelog

- Fixes IsActive and IsUpcoming

## Testing and Verifying

### DIFF
--- a/x/incentives/types/gauge.go
+++ b/x/incentives/types/gauge.go
@@ -31,12 +31,12 @@ func NewGauge(id uint64, isPerpetual bool, distrTo lockuptypes.QueryCondition, c
 
 // IsUpcomingGauge returns true if the gauge's distribution start time is after the provided time.
 func (gauge Gauge) IsUpcomingGauge(curTime time.Time) bool {
-	return curTime.After(gauge.StartTime)
+	return curTime.Before(gauge.StartTime) || curTime.Equal(gauge.StartTime)
 }
 
 // IsActiveGauge returns true if the gauge is in an active state during the provided time.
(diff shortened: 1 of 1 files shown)


x/incentives/types/gauge.go
```diff
@@ -31,12 +31,12 @@ func NewGauge(id uint64, isPerpetual bool, distrTo lockuptypes.QueryCondition, c
 
 // IsUpcomingGauge returns true if the gauge's distribution start time is after the provided time.
 func (gauge Gauge) IsUpcomingGauge(curTime time.Time) bool {
-	return curTime.After(gauge.StartTime)
+	return curTime.Before(gauge.StartTime) || curTime.Equal(gauge.StartTime)
```
p0mvn, line 34: I believe active gauge instead of upcoming should be inclusive of curTime, based on:

https://github.com/osmosis-labs/osmosis/blob/dd0c5a7c2ca253cac493991dac438806ae2527b2/x/incentives/keeper/distribute.go#L51-L53

ValarDragon:
Great catch, I've also confirmed that this only applies to genesis logic

ValarDragon (approved):

x/incentives/types/gauge.go
ValarDragon: Agreed

mattverse (approved):
Confirmed as well this was only called in genesis.go

PR #2788, opened 2022-09-20
feat(osmomath): log2 approximation

p0mvn:
Closes: #3086

This is an implementation of a base 2 logarithm.

It is based on: https://stackoverflow.com/questions/4657468/fast-fixed-point-pow-log-exp-and-sqrt
Currently, there are several optimizations and improvements possible to improve performance and accuracy.

The core idea is based on:
$$log_{2}{x} = log_{2}{(2*x/2)} = log_{2}{2} + log_{2}{(x / 2)} = 1 + log_{2}{(x / 2)}$$

We iteratively repeat this expansion, calculating the number of iterations (`y`), until $$1 <= x < 2$$

The number of iterations after normalization (`y`) equals to the whole part of the exponent.

Additionally, we calculate the logarithm of the remaining mantissa by following the approach described here:
http://www.claysturner.com/dsp/BinaryLogarithm.pdf

## Testing and Verifying

This change added tests and benchmarks


osmomath/decimal.go
```diff
@@ -54,6 +56,27 @@ func init() {
...
 		precisionMultipliers[i] = calcPrecisionMultiplier(int64(i))
 	}
+
+	log2LookupTable = buildLog2LookupTable()
+}
+
+// buildLog2LookupTable returns a lookup table for log values
+// ranging from [1, 2)
+// the keys are multiplied by 10 to simplify the rounding logic
+// in the log function.s
```
p0mvn, line 66:
```suggestion
// in the log function
```

p0mvn:
Benchmark at commit: [d6324f4](https://github.com/osmosis-labs/osmosis/pull/2788/commits/d6324f4867570af2c42910c64ec1f41bfcaf82bd)

```
goos: linux
goarch: amd64
pkg: github.com/osmosis-labs/osmosis/v12/osmomath
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
BenchmarkLog2-16    	   10000	    101134 ns/op	   67700 B/op	    1231 allocs/op
BenchmarkLog2-16    	   10000	    145882 ns/op	   67965 B/op	    1237 allocs/op
BenchmarkLog2-16    	    5715	    177965 ns/op	   68150 B/op	    1241 allocs/op
BenchmarkLog2-16    	    6957	    417276 ns/op	   67861 B/op	    1235 allocs/op
BenchmarkLog2-16    	    5710	    188281 ns/op	   67831 B/op	    1234 allocs/op
BenchmarkLog2-16    	    6952	    181742 ns/op	   68041 B/op	    1239 allocs/op
BenchmarkLog2-16    	    7924	    142498 ns/op	   68228 B/op	    1242 allocs/op
BenchmarkLog2-16    	    8320	    145017 ns/op	   67667 B/op	    1231 allocs/op
BenchmarkLog2-16    	    7587	    152439 ns/op	   67778 B/op	    1233 allocs/op
BenchmarkLog2-16    	    7418	    139766 ns/op	   67921 B/op	    1236 allocs/op
PASS
ok  	github.com/osmosis-labs/osmosis/v12/osmomath	27.615s

```

p0mvn:
According to benchmarks, reducing the log to be between `[1, 1.0001)` ends up being much less performant. More than 5500 iterations for log base 2 of 3. With the current implementations, there are approximately 300 iterations.

The reason is that we have to continuously be looping and dividing by the upper bound to normalize the number to be below `1.0001`. Since the denominator is small, it ends up taking many iterations, reducing the performance. 

I don't have the exact benchmark in nanoseconds for the `[1, 1.0001)` approach but it is higher than the current implementation even before reaching Taylor series step.

Upon exploring the properties of log base 2, I discovered the current method that required fewer iterations:
http://www.claysturner.com/dsp/BinaryLogarithm.pdf

It is implemented here and provides strong accuracy guarantees relative to the Wolfram Alpha results.

The benchmark for the latest commit is added up top

czarcas7ic (approved):
This LGTM!

osmomath/decimal.go
```diff
@@ -856,3 +866,56 @@ func DecApproxEq(t *testing.T, d1 BigDec, d2 BigDec, tol BigDec) (*testing.T, bo
...
+	if xCopy.LTE(ZeroDec()) {
+		panic(fmt.Sprintf("log is not defined at <= 0, given (%s)", xCopy))
+	}
+
+	// Normalize x to be 1 <= x < 2
+
+	// y is the exponent that results in a whole multiple of 2.
+	y := ZeroDec()
+
+	// invariant: x >= 1
```
nicolaslara, line 889: this is just a wording comment, but doesn't invariant imply that x>=1 is true throughout the iteration? I think you mean that after the iteration x>=1.  Are we using these invariants for formal verification somewhere?

osmomath/decimal.go
```diff
@@ -856,3 +866,56 @@ func DecApproxEq(t *testing.T, d1 BigDec, d2 BigDec, tol BigDec) (*testing.T, bo
...
+
+// LogBase2 returns log_2 {x}.
+// Rounds down by truncations during division and right shifting.
+// Accurate up to 32 precision digits.
+// Implementation is based on:
+// https://stm32duinoforum.com/forum/dsp/BinaryLogarithm.pdf
+func (x BigDec) LogBase2() BigDec {
+	// create a new decimal to avoid mutating
+	// the receiver's int buffer.
+	xCopy := ZeroDec()
```
nicolaslara, line 878: renaming xCopy to x, and x to xInput (or something similar) might make this easier to read. Not a blocker, just my stylistic preference.

p0mvn: Yeah I've thought about that but we have linter enabled to keep receiver names the same. So another stylistic thing preventing this change

osmomath/decimal.go
p0mvn: I think I was misusing the term to imply that we are aiming to normalize the value to be >= 1 after the iteration complete. Going to change this wording

osmomath/decimal.go
```diff
@@ -856,3 +866,56 @@ func DecApproxEq(t *testing.T, d1 BigDec, d2 BigDec, tol BigDec) (*testing.T, bo
...
+	// https://stm32duinoforum.com/forum/dsp/BinaryLogarithm.pdf
+	// This has shown precision of 32 digits relative
+	// to Wolfram Alpha in tests
+	for i := 0; i < maxLog2Iterations; i++ {
+		xCopy = xCopy.Mul(xCopy)
+		if xCopy.GTE(twoBigDec) {
+			xCopy.i = xCopy.i.Rsh(xCopy.i, 1)
+			y = y.Add(b)
+		}
+		b.i = b.i.Rsh(b.i, 1)
```
nicolaslara, line 917: If I understand the source correctly, `b` is used as a breaking condition (terminate once enough mantisa bits are found). Here we're just running a fixed number of iterations. Should we short circuit once a certain number of these bits are found? I'm assuming we know in advance how many bits ensures the aprox converges to the log

nicolaslara (commented):
Added some comments on the diff

osmomath/decimal.go
```diff
@@ -856,3 +866,56 @@ func DecApproxEq(t *testing.T, d1 BigDec, d2 BigDec, tol BigDec) (*testing.T, bo
...
+
+	// Normalize x to be 1 <= x < 2
+
+	// y is the exponent that results in a whole multiple of 2.
+	y := ZeroDec()
+
+	// invariant: x >= 1
+	// while x < 1
+	for xCopy.LT(OneDec()) {
+		xCopy.i = xCopy.i.Lsh(xCopy.i, 1)
```
nicolaslara, line 892: I think these assignments may be redundant

nicolaslara (approved):
Discussed it with @p0mvn. LGTM

osmomath/decimal.go
p0mvn: addressed

PR #2795, opened 2022-09-20
Refactors testnetify and remove last localOsmosis Dockerfile

niccoloraspa:

As per the title, this PR:
1. refactors the `testnetify` scripts
2. removes the localOsmosis `Dockerfile`

Main changes:

- Give code a better structure
- Create a CLI tool with multiple arguments
- Decouple the `testnetify` process from the `osmosisd` binary. All the needed addresses are now expected as input.
- Reuse the main `Dockerfile` to avoid creating a dedicated one
- Significant speedup was obtained by:
    - reading/writing  from disk only at the beginning/end 
    - Replace the `sed_inplace` with the `replace` function that works in-memory
    - `IBC` module pruning
- Run the `testnetify.py` at `docker run` time and not at `docker build` time

## Brief Changelog

1.  Refactor `testnetify` scripts
2. Remove the localOsmosis `Dockerfile`

## Testing and Verifying

I have updated the documentation that can be followed to test the process.

Main parts:

1. Copy `state_export.json` in `tests/localosmosis/state_export/state_export.json` 
2. `localnet-build-state-export`
3. `localnet-start-state-export`


niccoloraspa:
IMHO, a good way to review changes and understand the new flow is to go to top-down:

1. Inspect `docker-compose.yaml`
2. Inspect `start.sh`
3. Inspect `testnetify.py`

tests/localosmosis/state_export/scripts/testnetify.py
```diff
@@ -0,0 +1,361 @@
...
+        print("\tSet governance_voting_period to", config["governance_voting_period"])
+    genesis['app_state']['gov']['voting_params']['voting_period'] = config["governance_voting_period"]
+
+    # Update epochs module
+    if args.verbose:
+        print("⌛ Update epochs module")
+        print("\tSet epoch_duration to", config["epoch_duration"])
+        print("\tReset current_epoch_start_time")
+    genesis['app_state']['epochs']['epochs'][0]['duration'] = config["epoch_duration"]
+    genesis['app_state']['epochs']['epochs'][0]['current_epoch_start_time'] = datetime.now().isoformat() + 'Z'
```
czarcas7ic, line 218: Can we modify these to show what we are switching from? It just makes things easier when debugging / when dealing with multiple testnets

tests/localosmosis/state_export/scripts/testnetify.py
```diff
@@ -0,0 +1,361 @@
...
+    genesis['app_state']["ibc"]["channel_genesis"]["acknowledgements"] = []
+    genesis['app_state']["ibc"]["channel_genesis"]["channels"] = []
+    genesis['app_state']["ibc"]["channel_genesis"]["commitments"] = []
+    genesis['app_state']["ibc"]["channel_genesis"]["receipts"] = []
+    genesis['app_state']["ibc"]["channel_genesis"]["recv_sequences"] = []
+    genesis['app_state']["ibc"]["channel_genesis"]["send_sequences"] = []
+
+    genesis['app_state']["ibc"]["client_genesis"]["clients"] = []
+    genesis['app_state']["ibc"]["client_genesis"]["clients_consensus"] = []
+    genesis['app_state']["ibc"]["client_genesis"]["clients_metadata"] = []
```
czarcas7ic, line 234: This should not happen by default, it should only happen if explicitly requested by caller

tests/localosmosis/state_export/scripts/start.sh
```diff
@@ -0,0 +1,73 @@
...
+}
+
+if [[ ! -d $CONFIG_FOLDER ]]
+then
+
+    install_prerequisites
+
+    echo "Chain ID: $CHAIN_ID"
+    echo "Moniker:  $MONIKER"
+
```
czarcas7ic, line 43: should we start with an unsafe-reset-all? my main minor concern is if they have a filled addrbook, this will carry over and they will have a ton of p2p logs trying to call (probably mainnet) IPs

czarcas7ic (changes requested):
Awesome job with this!! Its really awesome to see something that is actually understandable compared to the previous testnetify script. Would like to discuss some of these points and then do another review before merging if possible!

tests/localosmosis/state_export/scripts/testnetify.py
```diff
@@ -0,0 +1,361 @@
...
+    )
+
+    parser.add_argument(
+        '-o',
+        '--output',
+        type = str,
+        default="testnet_genesis.json",
+        dest='output_genesis',
+        help='Path to output genesis'
+    )
```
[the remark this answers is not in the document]
niccoloraspa: yes, definitely makes sense

tests/localosmosis/state_export/scripts/testnetify.py
niccoloraspa: Agree. I had `--prune-ibc` argument before, not sure why I removed it

Makefile
```diff
@@ -363,26 +363,32 @@ localnet-keys:
...
+
+localnet-startd-state-export:
+	@docker-compose -f tests/localosmosis/state_export/docker-compose.yml up -d
+
+localnet-stop-state-export:
+	@docker-compose -f tests/localosmosis/docker-compose.yml down
+
 localnet-remove-state-export:
-	@docker-compose -f tests/localosmosis/mainnet_state/docker-compose-state-export.yml down
+	rm -rf $(PWD)/tests/localosmosis/state_export/.osmosisd
```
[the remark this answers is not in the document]
niccoloraspa: Yes, we're using a `.osmosisd` folder within the repo.
Same behavior of regular localOsmosis.

Should we change both?
TBH I prefer keeping it inside the repo but it's not a problem changing.

tests/localosmosis/state_export/scripts/testnetify.py
niccoloraspa: Sure!

tests/localosmosis/state_export/scripts/start.sh
niccoloraspa: Yeah, it has some verbose p2p logs at the beginning.
Let's test it and see if it's better

Makefile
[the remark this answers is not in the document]
czarcas7ic: Can we change this on both? From personal experience (today actually), I was confused where it was pulling the data from and had to remake a testnet instead of upgrading it. With the knowledge I have now I realize I could have done the following: 

1. Move the .osmosisd repo to the home folder
2. Checkout v12
3. Move the .osmosisd repo back to the in repo location

This seems like a hassle when we could just keep it using .osmosisd. What are your thoughts though? Totally down to keep it this way if there is a good reason!

tests/localosmosis/state_export/scripts/start.sh
czarcas7ic: For sure is not a big deal though and we can skip this issue for later! I'm going to mark resolved unless you feel it needs further exploration

czarcas7ic:
Side note, does args.verbose default to True? If not I think it should imo. 

Only question remaining is discussion on .osmosisd folder, other than that everything else looks good to go!

niccoloraspa:
> Side note, does args.verbose default to True? If not I think it should imo.
> 
> Only question remaining is discussion on .osmosisd folder, other than that everything else looks good to go!

verbose defaults to `false`

Makefile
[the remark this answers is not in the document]
niccoloraspa: The osmosis home makes more sense, to be honest.

I like the `.osmosisd` folder within the repo because the initialization step (running `testnetify` and modifying the various tomls) runs only if no  `.osmosisd/config` folder it's found: https://github.com/osmosis-labs/osmosis/blob/main/tests/localosmosis/scripts/setup.sh#L87

It's less likely to have a `.osmosisd` folder already in the GitHub repository, while it's very possible to have already a `.osmosisd` folder.

czarcas7ic:
@niccoloraspa do you agree that verbose should default to true? Also, would you like to just merge this as is and address the .osmosisd issue in a subsequent PR or do you want to do this in this PR? I'll approve now so you can decide whatever works best for you!

czarcas7ic (approved):

niccoloraspa:
> @niccoloraspa do you agree that verbose should default to true? Also, would you like to just merge this as is and address the .osmosisd issue in a subsequent PR or do you want to do this in this PR? I'll approve now so you can decide whatever works best for you!

No strong preference, I'm happy with verbose output as a default.
I will invert the logic and rename the `-v, --verbose` flag to `-q, --quiet` to make the output less verbose if needed.

czarcas7ic:
As discussed, merging this, then fixing/merging https://github.com/osmosis-labs/osmosis/pull/2813. From there we can make a fix for utilizing the home folder

PR #2801, opened 2022-09-20
bump IBC version

ValarDragon:
Bump to IBC v3.3.0, which has a minor security fix.


p0mvn (approved):
Pre-ACK for when the tests are passing

mattverse (commented):
Also agreeing bumping ibc-go version to v3.3.0, having confirmed the patch

ValarDragon:
Have to correct the SDK version in the backport

Issue #2794, opened 2022-09-20
[stableswap]: Generalize stableswap binary search solver to osmoutils binary search


AlpinYukseloglu:
## Background

Factoring out this binary search logic would simplify stableswap solver code drastically and would give us access to a `BigDec` binary search solver to use elsewhere later if needed

## Suggested Design

- Create a `BigDec` version of `BinarySearch` here:
https://github.com/osmosis-labs/osmosis/blob/0378e54207698098555b241306ebea9a0e23d1df/osmoutils/binary_search.go#L62-L68

## Acceptance Criteria

- Binary search unit tests and stableswap tests pass

Issue #2702, opened 2022-09-12
[stableswap]: Resolve large inaccuracies that arise from solver approximations


AlpinYukseloglu:
## Background

Our CFMM solver has many terms that are heavily approximated, mainly due square root approximations. This leads to an ever-increasing inaccuracy in the output of the CFMM that scales with the number of tokens in pool reserves. We need to find a way to minimize or bound these inaccuracies if we want to be able to accommodate pools with large amounts of assets.

The simplified solver implemented in #2697 reduces these inaccuracies by around 3x for our current tests, but they still fall outside of our intended thresholds and, importantly, still scale with the number of assets in the pool. We can continue trying to simplify the solver, but I believe this path will have significantly diminishing returns.

## Suggested Design

- Interpolate the output value given the difference in k values between the beginning and end of the solver
- If there is a clear overestimate, truncate the output to minimize security risks in the short term

## Acceptance Criteria

- Trades of all sizes against pools with $100M in each asset fall within our expected thresholds

ValarDragon:
I feel like the binary search approach is probably the move.

Also we really need to raise the fractional exponentiation precision bound, for the cubic root.

AlpinYukseloglu:
Closed by #2845

PR #2802, opened 2022-09-20
Implement `BigDec` binary search in osmoutils

AlpinYukseloglu:
Progress towards: #2794 and #2702

This PR implements a `BigDec` binary search function for our general osmoutils package.

## Brief Changelog

- Add `BinarySearchBigDec` to osmoutils
- Add `CompareBigDec` to osmoutils
- Refactor existing and new tests to be more readable (string map names, `t.Run` etc.)

## Testing and Verifying

This change added tests for both new functions in `osmoutils/binary_search_test.go`

### DIFF
--- a/osmoutils/binary_search.go
+++ b/osmoutils/binary_search.go
@@ -132,8 +132,13 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
 	return sdk.Int{}, errors.New("hit maximum iterations, did not converge fast enough")
 }
 
-// Binary search BigDec inputs between [lowerbound, upperbound] to a monotonic increasing function f
-// We stop once f(found_input) meets the ErrTolerance constraints.
+// BinarySearchBigDec takes as input:
+// * an input range [lowerbound, upperbound] 
+// * an increasing function f
+// * a target output x
+// * max number of iterations (for gas control / handling does-not-converge cases)
+// 
+// It binary searches on the input range, until it finds an input y s.t. f(y) meets the err tolerance constraints for how close it is to x.
 // If we perform more than maxIterations (or equivalently lowerbound = upperbound), we return an error.
 func BinarySearchBigDec(f func(input osmomath.BigDec) (osmomath.BigDec, error),
 	lowerbound osmomath.BigDec,
--- a/osmoutils/binary_search.go
+++ b/osmoutils/binary_search.go
@@ -58,7 +58,7 @@ func (e ErrTolerance) Compare(expected sdk.Int, actual sdk.Int) int {
 	return 0
 }
 
-// CompareBigDec returns if actual is within errTolerance of expected.
+// CompareBigDec validates if actual is within errTolerance of expected.
 // returns 0 if it is
 // returns 1 if not, and expected > actual.
 // returns -1 if not, and expected < actual
--- a/osmoutils/binary_search.go
+++ b/osmoutils/binary_search.go
@@ -120,7 +120,7 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
 		} else if compRes < 0 {
 			lowerbound = curEstimate
 		} else {
-			break
+			return curEstimate, nil
 		}
 		curEstimate = lowerbound.Add(upperbound).QuoRaw(2)
 		curOutput, err = f(curEstimate)
@@ -128,10 +128,8 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
 			return sdk.Int{}, err
 		}
 	}
-	if curIteration == maxIterations {
-		return sdk.Int{}, errors.New("hit maximum iterations, did not converge fast enough")
-	}
-	return curEstimate, nil
+
+	return sdk.Int{}, errors.New("hit maximum iterations, did not converge fast enough")
 }
 
 // Binary search BigDec inputs between [lowerbound, upperbound] to a monotonic increasing function f
@@ -158,7 +156,7 @@ func BinarySearchBigDec(f func(input osmomath.BigDec) (osmomath.BigDec, error),
 		} else if compRes < 0 {
 			lowerbound = curEstimate
 		} else {
-			break
+			return curEstimate, nil
 		}
 		curEstimate = lowerbound.Add(upperbound).Quo(osmomath.NewBigDec(2))
 		curOutput, err = f(curEstimate)
@@ -166,8 +164,6 @@ func BinarySearchBigDec(f func(input osmomath.BigDec) (osmomath.BigDec, error),
 			return osmomath.BigDec{}, err
 		}
 	}
-	if curIteration == maxIterations {
-		return osmomath.BigDec{}, errors.New("hit maximum iterations, did not converge fast enough")
-	}
-	return curEstimate, nil
+
+	return osmomath.BigDec{}, errors.New("hit maximum iterations, did not converge fast enough")
 }
--- a/osmoutils/binary_search.go
+++ b/osmoutils/binary_search.go
@@ -2,7 +2,6 @@ package osmoutils
 
 import (
 	"errors"
-	"fmt"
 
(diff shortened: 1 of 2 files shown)


osmoutils/binary_search.go
```diff
@@ -93,3 +133,41 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
...
+	}
+	curIteration := 0
+	for ; curIteration < maxIterations; curIteration += 1 {
+		compRes := errTolerance.CompareBigDec(curOutput, targetOutput)
+		if compRes > 0 {
+			upperbound = curEstimate
+		} else if compRes < 0 {
+			lowerbound = curEstimate
+		} else {
+			break
```
p0mvn, line 161: Consider simply returning here

osmoutils/binary_search.go
```diff
@@ -93,3 +133,41 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
...
+		curEstimate = lowerbound.Add(upperbound).Quo(osmomath.NewBigDec(2))
+		curOutput, err = f(curEstimate)
+		if err != nil {
+			return osmomath.BigDec{}, err
+		}
+	}
+	if curIteration == maxIterations {
+		return osmomath.BigDec{}, errors.New("hit maximum iterations, did not converge fast enough")
+	}
+	return curEstimate, nil
```
p0mvn, line 172: If you go with the previous suggestion, you should be able to simply return error here, without doing the iteration check

osmoutils/binary_search.go
AlpinYukseloglu: This is a good point – after a bit of thought this does seem to be the simplest approach. Updated!

osmoutils/binary_search.go
AlpinYukseloglu: Updated, please see my comment above! 

https://github.com/osmosis-labs/osmosis/pull/2802/files#r975799847

osmoutils/binary_search.go
```diff
@@ -56,6 +58,44 @@ func (e ErrTolerance) Compare(expected sdk.Int, actual sdk.Int) int {
 	return 0
 }
 
+// CompareBigDec returns if actual is within errTolerance of expected.
+// returns 0 if it is
+// returns 1 if not, and expected > actual.
+// returns -1 if not, and expected < actual
```
p0mvn, line 64:
```suggestion
// CompareBigDec validates if actual is within errTolerance of expected.
// returns 0 if it is
// returns 1 if not, and expected > actual.
// returns -1 if not, and expected < actual
```

osmoutils/binary_search_test.go
```diff
@@ -41,26 +43,89 @@ func TestBinarySearch(t *testing.T) {
...
+		"exponential f, large additive err tolerance, does not converge": {expF, osmomath.ZeroDec(), osmomath.NewBigDec(1 << 50), osmomath.NewBigDec((1 << 30)), testErrToleranceAdditive, 10, osmomath.BigDec{}, true},
+		"exponential f, large multiplicative err tolerance, converges": {expF, osmomath.ZeroDec(), osmomath.NewBigDec(1 << 50), osmomath.NewBigDec(1 + (1 << 25)), testErrToleranceMultiplicative, 51, osmomath.NewBigDec(322539792367616), false},
+		"exponential f, large multiplicative err tolerance, does not converge": {expF, osmomath.ZeroDec(), osmomath.NewBigDec(1 << 50), osmomath.NewBigDec(1 + (1 << 25)), testErrToleranceMultiplicative, 10, osmomath.BigDec{}, true},
+		"exponential f, both err tolerances, converges": {expF, osmomath.ZeroDec(), osmomath.NewBigDec(1 << 50), osmomath.NewBigDec((1 << 15)), testErrToleranceBoth, 51, osmomath.NewBigDec(1 << 45), false},
+		"exponential f, both err tolerances, does not converge": {expF, osmomath.ZeroDec(), osmomath.NewBigDec(1 << 50), osmomath.NewBigDec((1 << 30)), testErrToleranceBoth, 10, osmomath.BigDec{}, true},
```
p0mvn, line 116: Consider test cases with different lower and upper bounds. Happy for this to be done in another issue / PR.

p0mvn (approved):
LGTM. What are our plans for the binary search solver in `stableswap`? Is it going to be used anywhere?

AlpinYukseloglu:
Yes! After we merge this and #2697 (which has most of our stableswap CFMM tests in it) I will migrate our core solvers to use this implementation @p0mvn

osmoutils/binary_search_test.go
AlpinYukseloglu: Created issue here! #2807

osmoutils/binary_search.go
```diff
@@ -56,6 +58,44 @@ func (e ErrTolerance) Compare(expected sdk.Int, actual sdk.Int) int {
...
+	// Check additive tolerance equations
+	if !e.AdditiveTolerance.IsNil() {
+		// if no error accepted, do a direct compare.
+		if e.AdditiveTolerance.IsZero() {
+			if expected.Equal(actual) {
+				return 0
+			}
+		}
+
+		if diff.GT(osmomath.NewBigDec(e.AdditiveTolerance.Int64())) {
```
ValarDragon, line 84: I don't think this quite makes sense, to increase the precision, we first truncate to int64 and then increase the size.

Should probably e.AdditiveTolerance.ToDec() and then make a big dec from that.

osmoutils/binary_search.go
```diff
@@ -80,16 +120,50 @@ func BinarySearch(f func(input sdk.Int) (sdk.Int, error),
...
 	}
-	if curIteration == maxIterations {
-		return sdk.Int{}, errors.New("hit maximum iterations, did not converge fast enough")
+
+	return sdk.Int{}, errors.New("hit maximum iterations, did not converge fast enough")
+}
+
+// Binary search BigDec inputs between [lowerbound, upperbound] to a monotonic increasing function f
+// We stop once f(found_input) meets the ErrTolerance constraints.
+// If we perform more than maxIterations (or equivalently lowerbound = upperbound), we return an error.
```
ValarDragon, line 137:
```suggestion
// BinarySearchBigDec takes as input:
// * an input range [lowerbound, upperbound] 
// * an increasing function f
// * a target output x
// * max number of iterations (for gas control / handling does-not-converge cases)
// 
// It binary searches on the input range, until it finds an input y s.t. f(y) meets the err tolerance constraints for how close it is to x.
// If we perform more than maxIterations (or equivalently lowerbound = upperbound), we return an error.
```

osmoutils/binary_search_test.go
```diff
@@ -41,26 +43,89 @@ func TestBinarySearch(t *testing.T) {
...
+	}
+}
+
+func TestBinarySearchBigDec(t *testing.T) {
+	// straight line function that returns input. Simplest to binary search on,
+	// binary search directly reveals one bit of the answer in each iteration with this function.
+	lineF := func(a osmomath.BigDec) (osmomath.BigDec, error) {
+		return a, nil
+	}
+	expF := func(a osmomath.BigDec) (osmomath.BigDec, error) {
```
ValarDragon, line 77: Shouldn't be called `expF`, this is a cubic function

ValarDragon (approved):

osmoutils/binary_search.go
AlpinYukseloglu: Ah so we didn't have the `BigDecFromSDKDec` implemented in `BigDec` when I made this PR so what you suggested wasn't possible then, but I actually agree with this change! Updating now

osmoutils/binary_search_test.go
AlpinYukseloglu: Great point, I was extending the variable names used in the integer binary search but you're completely right – updating!

Issue #2689, opened 2022-09-09
proposal(twap): consider using last spot price instead of zero on error in `getSpotPrices`


p0mvn:
## Background

Context:
https://github.com/osmosis-labs/osmosis/blob/4bbd3fb45cef4d9c88207822858f6eeaf9fe95cf/x/twap/logic.go#L46-L54

I think this is a worthwhile improvement to make before v12.

## Acceptance Criteria
- decide if including in v12

mattverse:
Let me think about the possible edge case here, is it going to error when the pool does not have the last spot price? By last spot price, are we referring to the spot price recorded in the previous epoch?

p0mvn:
> Let me think about the possible edge case here, is it going to error when the pool does not have the last spot price? By last spot price, are we referring to the spot price recorded in the previous epoch?

Assume that we have a twap record at time `t`. Then, at `t + 1` we have a swap so a new record needs to be created with the updated spot prices. Let us get a spot price error.

Currently, we would set the spot price to 0 at `t + 1` since it errored. The problem with that is that it negatively impacts the calculations by diverging them significantly from the true value.

Instead of setting it to zero, we might want to reuse the spot price from the previous record (at time ``t`)

One edge case I can think of is when we get an error but there is no older record. 

I don't think it should be hard to implement the change. I will try getting it out before Monday to further discuss

mattverse:
> Currently, we would set the spot price to 0 at t + 1 since it errored. The problem with that is that it negatively impacts the calculations by diverging them significantly from the true value

If this is the case, I'm definitely for what this issue is proposing! (I wonder if we have a test case showing that rn)

p0mvn:
Decided on the call that this is not v12 blocking but good to have in the future releases

PR #2809, opened 2022-09-21
Using last spot price instead of zero on error in `getSpotPrices`

hieuvubk:
Closes: #2689 

> Refer from #2689 . Instead of updating price=0 if `getSpotPrices` got error as it is. We will update the price with the previous record.

## Brief Changelog
 
  - *Update `getSpotPrices` logic*
  - *Update related function & test*
  - *Got a special case that needs to be discussed. What if the pair have no previous record? Should spot price = 0( Following this way in my code)*

### DIFF
--- a/x/twap/logic.go
+++ b/x/twap/logic.go
@@ -10,13 +10,13 @@ import (
 	"github.com/osmosis-labs/osmosis/v12/x/twap/types"
 )
 
-func newTwapRecord(k types.AmmInterface, ctx sdk.Context, poolId uint64, denom0, denom1 string) (types.TwapRecord, error) {
+func newTwapRecord(k types.AmmInterface, keeper Keeper, ctx sdk.Context, poolId uint64, denom0, denom1 string) (types.TwapRecord, error) {
 	denom0, denom1, err := types.LexicographicalOrderDenoms(denom0, denom1)
 	if err != nil {
 		return types.TwapRecord{}, err
 	}
 	previousErrorTime := time.Time{} // no previous error
-	sp0, sp1, lastErrorTime := getSpotPrices(ctx, k, poolId, denom0, denom1, previousErrorTime)
+	sp0, sp1, lastErrorTime := getSpotPrices(ctx, k, keeper, poolId, denom0, denom1, previousErrorTime)
 	return types.TwapRecord{
 		PoolId:                      poolId,
 		Asset0Denom:                 denom0,
@@ -39,6 +39,7 @@ func newTwapRecord(k types.AmmInterface, ctx sdk.Context, poolId uint64, denom0,
 func getSpotPrices(
 	ctx sdk.Context,
 	k types.AmmInterface,
+	keeper Keeper,
 	poolId uint64,
 	denom0, denom1 string,
 	previousErrorTime time.Time,
@@ -47,15 +48,21 @@ func getSpotPrices(
 	sp0, err0 := k.CalculateSpotPrice(ctx, poolId, denom0, denom1)
 	sp1, err1 := k.CalculateSpotPrice(ctx, poolId, denom1, denom0)
 	if err0 != nil || err1 != nil {
+		previousRecord, err := keeper.getRecordAtOrBeforeTime(ctx, poolId, previousErrorTime, denom0, denom1)
 		latestErrTime = ctx.BlockTime()
-		// In the event of an error, we just sanity replace empty values with zero values
-		// so that the numbers can be still be calculated within TWAPs over error values
-		// TODO: Should we be using the last spot price?
 		if (sp0 == sdk.Dec{}) {
-			sp0 = sdk.ZeroDec()
+			if err != nil {
+				sp0 = sdk.ZeroDec()
+			} else {
+				sp0 = previousRecord.P0LastSpotPrice
+			}
 		}
 		if (sp1 == sdk.Dec{}) {
-			sp1 = sdk.ZeroDec()
+			if err != nil {
+				sp1 = sdk.ZeroDec()
+			} else {
+				sp1 = previousRecord.P1LastSpotPrice
+			}
 		}
 	}
 	if sp0.GT(types.MaxSpotPrice) {
@@ -72,7 +79,7 @@ func (k Keeper) afterCreatePool(ctx sdk.Context, poolId uint64) error {
 	denoms, err := k.ammkeeper.GetPoolDenoms(ctx, poolId)
 	denomPairs0, denomPairs1 := types.GetAllUniqueDenomPairs(denoms)
 	for i := 0; i < len(denomPairs0); i++ {
-		record, err := newTwapRecord(k.ammkeeper, ctx, poolId, denomPairs0[i], denomPairs1[i])
+		record, err := newTwapRecord(k.ammkeeper, k, ctx, poolId, denomPairs0[i], denomPairs1[i])
 		// err should be impossible given GetAllUniqueDenomPairs guarantees
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


ValarDragon:
This is not eligible for v12 inclusion

mattverse:
assuming that we're putting this in in the upcoming future releases, I'm wondering if there needs to be migration work needed to be done for the existing twap records?

ValarDragon:
Nah this wouldn't need a state migration, it'll only affect new records

x/twap/logic.go
```diff
@@ -10,13 +10,13 @@ import (
 	"github.com/osmosis-labs/osmosis/v12/x/twap/types"
 )
 
-func newTwapRecord(k types.AmmInterface, ctx sdk.Context, poolId uint64, denom0, denom1 string) (types.TwapRecord, error) {
+func newTwapRecord(k types.AmmInterface, keeper Keeper, ctx sdk.Context, poolId uint64, denom0, denom1 string) (types.TwapRecord, error) {
```
mattverse, line 13: In this case I suggest we just make this a keeper method instead of passing keeper as a parameter

x/twap/logic.go
```diff
@@ -47,15 +48,21 @@ func getSpotPrices(
...
 	sp1, err1 := k.CalculateSpotPrice(ctx, poolId, denom1, denom0)
 	if err0 != nil || err1 != nil {
+		previousRecord, err := keeper.getRecordAtOrBeforeTime(ctx, poolId, previousErrorTime, denom0, denom1)
 		latestErrTime = ctx.BlockTime()
-		// In the event of an error, we just sanity replace empty values with zero values
-		// so that the numbers can be still be calculated within TWAPs over error values
-		// TODO: Should we be using the last spot price?
 		if (sp0 == sdk.Dec{}) {
-			sp0 = sdk.ZeroDec()
+			if err != nil {
```
mattverse, line 54:
```suggestion
                        // use zero if we fail to get previous record
			if err != nil {
```

x/twap/logic.go
```diff
@@ -47,15 +48,21 @@ func getSpotPrices(
 	sp0, err0 := k.CalculateSpotPrice(ctx, poolId, denom0, denom1)
 	sp1, err1 := k.CalculateSpotPrice(ctx, poolId, denom1, denom0)
 	if err0 != nil || err1 != nil {
+		previousRecord, err := keeper.getRecordAtOrBeforeTime(ctx, poolId, previousErrorTime, denom0, denom1)
 		latestErrTime = ctx.BlockTime()
-		// In the event of an error, we just sanity replace empty values with zero values
-		// so that the numbers can be still be calculated within TWAPs over error values
-		// TODO: Should we be using the last spot price?
 		if (sp0 == sdk.Dec{}) {
```
mattverse, line 53: IMO, with so much if and else cases here, I think this reduces code readability. What do you think of separating out the cases for err0 != nil and err1!= nil?

x/twap/logic_test.go
```diff
@@ -961,7 +961,7 @@ func (s *TestSuite) TestUpdateRecords() {
 				// The new record added.
 				{
 					spotPriceA:    sdk.OneDec(),
-					spotPriceB:    sdk.ZeroDec(),
+					spotPriceB:    baseRecord.P1LastSpotPrice,       // getSpotPrice failed, should equal to previous record LastSpotPrice
 					lastErrorTime: baseRecord.Time.Add(time.Second), // equals to block time
```
mattverse, line 965: I think we need further test cases, with having more variant in scenarios. 

Can we add test cases where
1. getting last spot prices errored, so we use last spot price from the previous record
2. getting last spot prices errored,and getting previous record also errored, resulting in exp spot price to be zero value?

mattverse (commented):
Nice job on changing these! Thanks for the contribution! 
Left some minor comments I think would be great to be addressed before merging this, otherwise LGTM!

hieuvubk:
Thank you for leaving comments. Updated almost them. Also add new test cases and update related tests at `TestUpdateRecords` & `TestUpdateRecord`

x/twap/logic_test.go
```diff
@@ -24,16 +24,17 @@ var (
 )
 
 func (s *TestSuite) TestGetSpotPrices() {
-	currTime := time.Now()
+	currTime := time.Now().UTC()
+	// set up pool at block t + 2, test have previous record
```
mattverse, line 28:
```suggestion
	// set up pool at curr block + 2 to test previous records
```

x/twap/logic_test.go
```diff
@@ -42,28 +43,43 @@ func (s *TestSuite) TestGetSpotPrices() {
...
-		"zero sp": {
+		"use last spot price on err": {
 			poolID:                poolID,
 			prevErrTime:           currTime,
+			blockTime:             s.Ctx.BlockTime(),
 			mockSp0:               sdk.ZeroDec(),
 			mockSp1:               sdk.ZeroDec(),
 			mockSp0Err:            fmt.Errorf("foo"),
-			expectedSp0:           sdk.ZeroDec(),
+			expectedSp0:           sdk.OneDec(),
```
mattverse, line 53: why does expectedSp0 become oneDec? (both mocksp0 and mocksp1) being zero dec, I was expecting expectedSp1 to be zero dec as well?

x/twap/logic_test.go
```diff
@@ -42,28 +43,43 @@ func (s *TestSuite) TestGetSpotPrices() {
...
+		"sp zero, when err & no previous record": {
+			poolID:                poolID,
+			prevErrTime:           currTime,
+			blockTime:             currTime.Add(5 * time.Second),
+			mockSp0:               sdk.OneDec(),
+			mockSp1:               sdk.OneDec(),
+			mockSp0Err:            fmt.Errorf("foo"),
+			expectedSp0:           sdk.ZeroDec(),
+			expectedSp1:           sdk.OneDec(),
+			expectedLatestErrTime: currTime.Add(5 * time.Second),
```
mattverse, line 67: same question here, why does expected SP0 result in zerodec and expected sp1 result in one dec?

x/twap/logic_test.go
hieuvubk: In this case, we got err when getting sp0. `blockTime` coincides with pool creation time so we have a record with sp0 = sp1 = 1. 
expectedSp0 take the old record sp0.

x/twap/logic_test.go
hieuvubk: `expectedSp0` = 0 cause got err when getting sp0 & no previous record (`blockTime` = t + 1 while pool create at block t + 2)

hieuvubk:
I leave some comments to discuss more

x/twap/logic_test.go
mattverse: hrrm but the last sp1 before the update was zero as well right? Shouldnt expectedSp1 stay zero as well?

x/twap/logic_test.go
```diff
@@ -74,10 +90,10 @@ func (s *TestSuite) TestGetSpotPrices() {
 
 	for name, tc := range testCases {
 		s.Run(name, func() {
+			ctx := s.Ctx.WithBlockTime(tc.blockTime)
```
mattverse, line 93: what's the significance of setting ctx with blocktime when there are no arguments to compare it with? (what / why are we setting blocktime for?)

x/twap/logic_test.go
mattverse: Sorry if im missing anything, we still have the previous record for sp0 which is onedec right?

x/twap/logic_test.go
hieuvubk: For case getting sp err & no previous record. Actually, this case never happen cause pool already have a record when created. So I use blockTime before pool creation's time to make sure no record.

x/twap/logic_test.go
hieuvubk: At block `t + 1` we dont have any record

x/twap/logic_test.go
hieuvubk: With sp1 we don't get any error so I think it's updated normally (is zero as mock)

PR #2813, opened 2022-09-21
feat: add gamm state to localosmosis

czarcas7ic:
Closes: #XXX

After discussion with @p0mvn, we agreed it would be very useful to provide an option (non-default) for LocalOsmosis to start with some pre-loaded state. This MVP allows a user to start their LocalOsmosis instance with `make localnet-start-with-state` and have two pools automatically created (one two asset pool and one three asset pool).

## Brief Changelog

- Adds state flag to setup.sh, when set:
  - Creates a two asset pool
  - Creates a three asset pool 

## Testing and Verifying

### DIFF
--- a/tests/localosmosis/docker-compose.yml
+++ b/tests/localosmosis/docker-compose.yml
@@ -11,10 +11,14 @@ services:
         RUNNER_IMAGE: alpine:3.16
         GO_VERSION: 1.18
     volumes:
+      - ../e2e/scripts/nativeDenomPool.json:/osmosis/nativeDenomPool.json
+      - ../e2e/scripts/nativeDenomThreeAssetPool.json:/osmosis/nativeDenomThreeAssetPool.json
       - ./scripts/setup.sh:/osmosis/setup.sh
       - ./.osmosisd/:/osmosis/.osmosisd/
     entrypoint:
       - /osmosis/setup.sh
+    command:
+      - $STATE
     ports:
       - 26657:26657
       - 1317:1317
--- a/tests/localosmosis/scripts/setup.sh
+++ b/tests/localosmosis/scripts/setup.sh
@@ -95,23 +95,33 @@ edit_config () {
 create_two_asset_pool () {
     # Create default pool
     substring='code: 0'
-    string=$(osmosisd tx gamm create-pool --pool-file=nativeDenomPool.json --from $MONIKER --chain-id=$CHAIN_ID --home $OSMOSIS_HOME --keyring-backend=test -b block --yes  2>&1)
-    if [ "$string" != "${string%"$substring"*}" ]; then
-        echo "create two asset pool: successful"
-    else
-        create_two_asset_pool
-    fi
+    COUNTER=0
+    while [ $COUNTER -lt 15 ]; do
+        string=$(osmosisd tx gamm create-pool --pool-file=nativeDenomPool.json --from $MONIKER --chain-id=$CHAIN_ID --home $OSMOSIS_HOME --keyring-backend=test -b block --yes  2>&1)
+        if [ "$string" != "${string%"$substring"*}" ]; then
+            echo "create two asset pool: successful"
+            break
+        else
+            let COUNTER=COUNTER+1
+            sleep 0.5
+        fi
+    done
 }
 
 create_three_asset_pool () {
     # Create three asset pool
     substring='code: 0'
-    string=$(osmosisd tx gamm create-pool --pool-file=nativeDenomThreeAssetPool.json --from $MONIKER --chain-id=$CHAIN_ID --home $OSMOSIS_HOME --keyring-backend=test -b block --yes 2>&1)
-    if [ "$string" != "${string%"$substring"*}" ]; then
-        echo "create three asset pool: successful"
-    else
-        create_three_asset_pool
-    fi
+    COUNTER=0
+    while [ $COUNTER -lt 15 ]; do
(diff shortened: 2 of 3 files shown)


czarcas7ic:
Note to reviewers, the only "issue" I have had with this implementation is when closing out the instance with Ctrl+C, it does not gracefully shut down and instead you either wait 5 seconds and it shuts down with a message or you Ctrl+C again to force it. I think this is because of the "wait" at the end, but AFAIU that is required to keep osmosisd running after running the make pool commands.

@niccoloraspa also tagging you for awareness because I know you made changes to this file as well. Would like to conjoin the two in some way!

czarcas7ic:
One more thing to note: Right now this is utilizing the same json files we use in e2e to create pools. This was to prevent code duplication. I can move these files into this folder as well, but I personally don't see these files within e2e going anywhere anytime soon, so I feel its fine for reuse. Would like to know other's opinions though!

tests/localosmosis/scripts/setup.sh
```diff
@@ -84,6 +92,28 @@ edit_config () {
...
 }
 
+create_two_asset_pool () {
+    # Create default pool
+    substring='code: 0'
+    string=$(osmosisd tx gamm create-pool --pool-file=nativeDenomPool.json --from $MONIKER --chain-id=$CHAIN_ID --home $OSMOSIS_HOME --keyring-backend=test -b block --yes  2>&1)
+    if [ "$string" != "${string%"$substring"*}" ]; then
+        echo "create two asset pool: successful"
+    else
+        create_two_asset_pool
```
p0mvn, line 102: Does it make sense to have a retry limit of some kind to avoid it looping infinitely?

tests/localosmosis/scripts/setup.sh
```diff
@@ -84,6 +92,28 @@ edit_config () {
...
+}
+
+create_three_asset_pool () {
+    # Create three asset pool
+    substring='code: 0'
+    string=$(osmosisd tx gamm create-pool --pool-file=nativeDenomThreeAssetPool.json --from $MONIKER --chain-id=$CHAIN_ID --home $OSMOSIS_HOME --keyring-backend=test -b block --yes 2>&1)
+    if [ "$string" != "${string%"$substring"*}" ]; then
+        echo "create three asset pool: successful"
+    else
+        create_three_asset_pool
```
p0mvn, line 113: ditto

p0mvn (approved):

tests/localosmosis/scripts/setup.sh
czarcas7ic: I forgot to do this! Thanks for catching, just implemented

tests/localosmosis/docker-compose.yml
```diff
@@ -11,10 +11,14 @@ services:
...
         GO_VERSION: 1.18
     volumes:
+      - ../e2e/scripts/nativeDenomPool.json:/osmosis/nativeDenomPool.json
+      - ../e2e/scripts/nativeDenomThreeAssetPool.json:/osmosis/nativeDenomThreeAssetPool.json
       - ./scripts/setup.sh:/osmosis/setup.sh
       - ./.osmosisd/:/osmosis/.osmosisd/
     entrypoint:
       - /osmosis/setup.sh
+    command:
+      - $STATE
```
niccoloraspa, line 21: Could we use environment variables instead of this?

In the other script, you can modify the `MONIKER` or `CHAIN_ID` via environment variables:

https://github.com/osmosis-labs/osmosis/pull/2795/files#diff-90a6bf7fea85b94101791de0f55d6ae2f059f900df53f24687e87ebeff4b5c7fR13-R15

and in the `docker-compose.yaml`:

https://github.com/osmosis-labs/osmosis/pull/2795/files#diff-19e0664081fa6ecc70118cd2c3b4204fcf23a2d63e2aba5a4d6fc294dfa342b0R20-R22

I think it's a little bit cleaner and consistent

niccoloraspa:
> One more thing to note: Right now this is utilizing the same json files we use in e2e to create pools. This was to prevent code duplication. I can move these files into this folder as well, but I personally don't see these files within e2e going anywhere anytime soon, so I feel its fine for reuse. Would like to know other's opinions though!


I suggest moving he `json` files in the `localOsmosis` and keep the environments isolated.
I'm sure that In two months we will forget about this and break localOsmosis :D

niccoloraspa:
The idea of creating an initial state it's great and the changes made here are perfectly fine.

The problem with this approach is that it doesn't scale well. 
In the future, we might want to add a more complex state,s and we would have to keep increasing the complexity of the setup scripts and the number of input variables to control which changes we want to add.

We're also facing this problem of  _initial state manipulation_ in different places, and in each one, we're trying to solve it differently:

- `e2e/init` uses Go
- `localOsmosis` uses a bash script
- `localOsmosis-state-export` uses bash + python script
- `testnets` bash scripts

I think it's worth spending some development effort to create something to address this problem because it's a requirement that we have.

What I have in mind is to build a `config.yaml`  (DevOps loves yamls sorry) where you would describe the state you want to have (accounts, pools, validators, genesis) and have a tool that creates/modify the state to match the configuration.

Ignite CLI is doing something similar for scaffolding chains:

```yaml
# config.yaml
# reference: https://docs.ignite.com/kb/genesis

accounts:
  - name: alice
    coins: ["1000token", "100000000stake"]
  - name: bob
    coins: ["500token"]
    address: cosmos1adn9gxjmrc3hrsdx5zpc9sj2ra7kgqkmphf8yw

genesis:
  chain_id: "foobar"

  # Change the denom in staking module
  app_state:
    staking:
      params:
        bond_denom: "denom"
```

I like this declarative approach, and it's very powerful.
We would only have one yaml to describe the environment and we could re-use this tool in all the scenarios mentioned above.

What are your thoughts on this?

ValarDragon (approved):
Yml configuration for stages is a cool idea, but a separate issue/feature

This LGTM

Issue #2615, opened 2022-09-06
x/gamm: make balancer `JoinPool` empty return value on error consistent with that of `CalcJoinPoolShares`


AlpinYukseloglu:
## Background

Note: this is a minor code hygiene change

Our current balancer implementation for `JoinPool` returns `sdk.Int{}` as `numShares` upon hitting an error, while all of its subsidiary functions return `sdk.ZeroInt()`. `sdk.Int{}` looks really messy in error messages and in general we want consistent error returns along a chain of function calls, so I think this would be a worthwhile change to make.

cc: @ValarDragon would this break anything?

## Suggested Design

- Replace balancer `JoinPool`'s empty return value on error with `sdk.ZeroInt()`
https://github.com/osmosis-labs/osmosis/blob/0b260495245c77d1967d5ef1379387357350d226/x/gamm/pool-models/balancer/pool.go#L656

## Acceptance Criteria

- All existing tests pass

hieuvubk:
@AlpinYukseloglu @ValarDragon should we close this issue, I can work on it

AlpinYukseloglu:
Feel free to pick this up!

AlpinYukseloglu:
Looks like we aren't going to be implementing this – thanks for taking a crack at it anyway @RusAkh!

https://github.com/osmosis-labs/osmosis/pull/2820#issuecomment-1256620893

PR #2820, opened 2022-09-22
from Int{} to ZeroInt() balancer pool

RusAkh:
Closes: #2615 

For the sake of cleanness returns in balancer/pool.go now return sdk.ZeroInt()

## Brief Changelog
Same as purpose

## Testing and Verifying

Changed value in TestJoinPool -> after that all current tests passed

  - Does this pull request introduce a new feature or user-facing behavior changes? No
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? No

### DIFF
--- a/x/gamm/pool-models/balancer/pool.go
+++ b/x/gamm/pool-models/balancer/pool.go
@@ -449,7 +449,7 @@ func (p *Pool) PokePool(blockTime time.Time) {
 func (p Pool) GetTokenWeight(denom string) (sdk.Int, error) {
 	PoolAsset, err := p.GetPoolAsset(denom)
 	if err != nil {
-		return sdk.Int{}, err
+		return sdk.ZeroInt(), err
 	}
 
 	return PoolAsset.Weight, nil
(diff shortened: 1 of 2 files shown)


AlpinYukseloglu (approved):
LGTM!

p0mvn (approved):

mattverse (commented):
Thanks for this PR! Glad that we brought this topic up!

I personally would prefer the other way around, where we would change existing zeroints to Int{}. The reason for this is because although we would have error catchers on the callers, I think we double ensure saftey for this within the method by returning a empty struct. IMO using zero int values can lead to using the zero value as the actual value in some cases.

Thoughts? cc. @AlpinYukseloglu @p0mvn

ValarDragon:
I'm not a fan of this, and don't understand why we would change this behavior.

p0mvn:
Thanks for the PR @RusAkh . Based on the discussions, we would like to instead be returning `sdk.Int{}` in similar cases so I'm going to close this

Issue #2791, opened 2022-09-20
Cleanup Balancer Pools


mattverse:
## Background

We currently have many unimportant pools, that were created before pool creation fees were created.
This burdens front end developers and node operators, causing high network fees in the pools query.

## Suggested Design

Proposed way of doing this is referencing https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/pool.go#L115, and including it in V13. 

The code snippet we've created back in the days. (We need a sincere review on this (maybe rewrite the entire function since most of the pool APIs are now outdated) as this was created long time ago). 

## Acceptance Criteria

- Rewrite or have deep code review on `CleanupBalancerPool`
- Add test cases for `CleanupBalancerPool`

JohnnyWyles:
Bumping this as it has been raised again.
We have governance approval for this from the early days of the chain.
https://www.mintscan.io/osmosis/proposals/85

ValarDragon:
We have decided that this does not make sense to do on-chain given the current state of Osmosis. The goal can be achieved by making a new query that skips these pools.

PR #2821, opened 2022-09-22
Cleanup Balancer Pools

mattverse:
Closes: #2791 

The purpose of this change is to add functionality to delete unnecessary pools that were created before pool creation fees were created. Unnecessary pools should be removed, as it burdens front end developers and node operators, causing high network fees in the pools query.

The actual implementation, which includes testing and implementations on the upgrade handler is TBD once this PR has passed. 

## Brief Changelog
- Add method to cleanup existing pools

## Testing and Verifying
Added new tests


x/gamm/keeper/pool_test.go
```diff
@@ -1,222 +1,65 @@
...
+		{
+			name:           "create pool with default coins, join pool with default coins",
+			createPoolFund: defaultCoins,
+			joinPoolFund:   defaultCoins,
+		},
+		{
+			name:           "create pool with default coins, join pool larger coins",
+			createPoolFund: defaultCoins,
+			joinPoolFund:   defaultCoins.Add(defaultCoins...),
+		},
```
mattverse, line 32: TODO: think about cases where we can have rounding errors

x/gamm/keeper/pool.go
```diff
@@ -105,97 +105,110 @@ func (k Keeper) DeletePool(ctx sdk.Context, poolId uint64) error {
...
+		}
+
+		// Refund assets
+		// lpShareEquivalentTokens = (amount in pool) * (your shares) / (total shares)
+		for _, asset := range poolLiquidity[shareDenom] {
+			lpShareEquivalentTokens := asset.Amount.Mul(coin.Amount).Quo(totalShares[shareDenom])
+			if lpShareEquivalentTokens.IsZero() {
+				continue
+			}
+
```
p0mvn, line 166:
```suggestion
```

x/gamm/keeper/pool.go
```diff
@@ -105,97 +105,110 @@ func (k Keeper) DeletePool(ctx sdk.Context, poolId uint64) error {
...
+	for _, poolId := range poolIds {
+		pool, err := k.GetPoolAndPoke(ctx, poolId)
+		if err != nil {
+			return err
+		}
+		shareDenom := types.GetPoolShareDenom(poolId)
+
+		// sanity check that we have no remaining shares
+		if !poolLiquidity[shareDenom].IsZero() {
+			panic("pool total share should be zero after cleanup")
```
p0mvn, line 194: consider returning an error here fore easier testing. I'm assuming the caller (upgrade handler) would panic on any error anyways

x/gamm/keeper/pool.go
```diff
@@ -105,97 +105,110 @@ func (k Keeper) DeletePool(ctx sdk.Context, poolId uint64) error {
...
-
-// 		err = k.DeletePool(ctx, pool.GetId())
-// 		if err != nil {
-// 			return err
-// 		}
-// 	}
-
-// 	return nil
-// }
+func (k Keeper) CleanupBalancerPool(ctx sdk.Context, poolIds []uint64) (err error) {
```
p0mvn, line 114: I think we should consider renaming this to `CleanupPools` because this should work for stableswap pools as well. There are currently no type checks that this is balancer-specific

x/gamm/keeper/pool.go
```diff
@@ -105,97 +105,110 @@ func (k Keeper) DeletePool(ctx sdk.Context, poolId uint64) error {
...
+	// we use maps here because we can't alter the state of pool directly
+	poolAddresses := make(map[string]sdk.AccAddress)
+	totalShares := make(map[string]sdk.Int)
+	poolLiquidity := make(map[string]sdk.Coins)
```
p0mvn, line 118: Wdyt about having:
```
type poolInfo struct {
    address sdk.AccAddress\
    totalShares sdk.Int
    liquidity sdk.Coins
}

poolInfo := map[string]poolInfo{}
```

It seems to be a common pattern to need more than one of this field on each iteration so by consolidating this in a struct, we would do one map lookup instead of 3

p0mvn: We could get all 3 at the beginning of iteration and update the map with the new amounts at the end to minimize the number of map accesses.

x/gamm/keeper/pool.go
mattverse: makes sense1

x/gamm/keeper/pool.go
mattverse: makes sense to me! fixed!

mattverse (commented):
Suppose we have 13uosmo, 13uatom in the pool. 
Also suppose
- `accountA` has 3 shares, `accountB` has 3 shares, `accountC` has 4shares, with the total pool shares being 11 shares. 

Iterating over three users in the order of account C, account B, account A, each account would end up having (4uatom, 4uosmo), (5uatom, 5uosmo), (4uatom, 4uosmo) respectively. 

Now suppose we iterate in the order of account A, account B, account C. Each account would end up having (3uatom, 3uosmo), (5uatom, 5uosmo), (5uatom, 5uosmo),  respectively. 

This means that the order of exiting pool makes difference in how much each account would get refunded, and the current way of calculating pool lqiduidity to refund would not work. 

Marking this PR as a draft, changing this to simply calculating how much pool liquidity each share accounts for and them refunding (num_shares) * (pool_liquidity_per_share)

mattverse:
Updated how we calculate pool share per liquidity for liquidity refund. R4R!

PR #2825, opened 2022-09-22
[stableswap]: Switch multi-asset CFMM solver to use BigDec binary search

AlpinYukseloglu:
Progress towards: #2794 

This PR replaces our stableswap multi-asset CFMM solver to use the `BigDec` binary search implementation in #2802 

Note: most of the diff in this PR is new and refactored test cases

## Brief Changelog

- Replace multi asset binary search solver logic with our `BigDec` binary search implementation

## Testing and Verifying

- All two-asset CFMM test cases were set up to test this implementation in `amm_test.go`
- There is one important set of tests that currently overflow (4-asset pools with $100M in each), so I left those tests commented pending on increasing `BigDec` precision even further


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -321,22 +322,26 @@ func solveCFMMBinarySearchMulti(constantFunction func(osmomath.BigDec, osmomath.
...
-			x_est = (x_high_est.Add(x_low_est)).Quo(twodec)
-			cur_k = constantFunction(x_est, yf, uReserve, wSumSquares)
+		yFinal := yReserve.Add(yIn)
+		xLowEst := osmomath.ZeroDec()
+		xHighEst := xReserve
+		maxIterations := 256
+		errTolerance := osmoutils.ErrTolerance{AdditiveTolerance: sdk.OneInt(), MultiplicativeTolerance: sdk.Dec{}}
+
+		// create single-input CFMM to pass into binary search
+		calc_x_est := func(xEst osmomath.BigDec) (osmomath.BigDec, error) {
```
p0mvn, line 336: nit: casing

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -321,22 +322,26 @@ func solveCFMMBinarySearchMulti(constantFunction func(osmomath.BigDec, osmomath.
...
+			return constantFunction(xEst, yFinal, uReserve, wSumSquares), nil
 		}
+
+		x_est, err := osmoutils.BinarySearchBigDec(calc_x_est, xLowEst, xHighEst, k, errTolerance, maxIterations)
```
p0mvn, line 340: ditto

x/gamm/pool-models/stableswap/amm_test.go
```diff
@@ -12,6 +13,253 @@ import (
...
+	"even 4-asset medium pool, small input": {
+		xReserve: osmomath.NewBigDec(100000),
+		yReserve: osmomath.NewBigDec(100000),
+		// represents a 4-asset pool with 100,000 in each reserve
+		uReserve:    osmomath.NewBigDec(10000000000),
+		wSumSquares: osmomath.NewBigDec(20000000000),
+		yIn:         osmomath.NewBigDec(1),
+		expectPanic: false,
+	},
+	/* TODO: increase BigDec precision (36 -> 72) to be able to accommodate this
```
p0mvn, line 65: random question: how much is this going to affect performance? I'm assuming this will make some math operations to converge slower. Do we have benchmarks / should we be comparing trade-offs?

x/gamm/pool-models/stableswap/amm_test.go
```diff
@@ -12,6 +13,253 @@ import (
...
+		wSumSquares: osmomath.NewBigDec(20000),
+		yIn:         osmomath.NewBigDec(1),
+		expectPanic: true,
+	},
+	"wSumSquares overflows in 4-asset pool": {
+		xReserve: osmomath.NewBigDec(100),
+		yReserve: osmomath.NewBigDec(100),
+		// represents a 4-asset pool with 100 in each reserve
+		uReserve:    osmomath.NewBigDec(200),
+		wSumSquares: osmomath.NewDecFromBigInt(new(big.Int).Sub(new(big.Int).Exp(big.NewInt(2), big.NewInt(1024), nil), big.NewInt(1))),
```
p0mvn, line 248: nit: consider extracting this common "overflow" variable

x/gamm/pool-models/stableswap/amm_test.go
```diff
@@ -92,93 +340,37 @@ func TestCFMMInvariantTwoAssets(t *testing.T) {
...
-			osmomath.NewBigDec(1000),
-			false,
-		},
-		// This test fails due to a bug in our original solver
-		// "large pool, large input": {
-		// 	sdk.NewDec(100000),
-		// 	sdk.NewDec(100000),
-		// 	sdk.NewDec(10000),
-		// },
+	// TODO: switch solveCfmmMulti to binary search and replace this with test case suite
```
p0mvn, line 343: Sorry not sure I'm following the plan here. 

If we are going to replace it with the binary search why would we have two sets of tests?

x/gamm/pool-models/stableswap/amm_test.go
AlpinYukseloglu: Great point, I also grouped the two asset and multi asset variable declarations together into `var(...)` – looks much cleaner now!

x/gamm/pool-models/stableswap/amm_test.go
AlpinYukseloglu: Originally we had 3 distinct solvers for our CFMM (general direct, simplified direct, and binary search), so we kept `solveCfmm` and `solveCfmmMulti` as a layer of abstraction so we wouldn't have to manually edit all pool logic as we worked on the different solver types

If/when we decide to enshrine one solver as our main one (likely binary search as indicated in here), then it might make sense to consolidate these into one, but I think we should do that after all of the loose ends are tied with our solvers

x/gamm/pool-models/stableswap/amm_test.go
AlpinYukseloglu: I'm not sure about performance impact, but the max number of assets per pool we want to support is still an open question that will determine the final precision (72 isn't enough for large 5+ asset pools)

Once we finalize precision, we can expand `amm_bench_test.go` tests to see what the performance effect is, but it is upper bounded at `maxIterations` so it likely won't be too bad

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: fixed

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: fixed

x/gamm/pool-models/stableswap/amm_test.go
AlpinYukseloglu: Please see this issue: https://github.com/osmosis-labs/osmosis/issues/2845

ValarDragon (approved):

ValarDragon:
I believe these changes were a subset of #2847 please re-open if this isn't right

Issue #2739, opened 2022-09-14
x/gamm: Make a `PoolType()` query, that returns a string type name


ValarDragon:
## Background & suggested design

We should make a `PoolType()` query, so users & contracts can easily know the type of a pool, between Balancer and Stableswap pools. This can be done via the `pool` query, but thats intended for clients as a fallback if they need the raw representation. We don't want that to be the stable API people are depending upon, and hence don't want that to be contract-accessible.

RusAkh:
I think it would be good to add a method to PoolI interface for example .Type() so we won't need to use typecasting to access type
```
_, ok := pool.(*balancer.Pool)
if ok {
      return &types.QueryPoolTypeResponse{Type: "Balancer"}, nil
} 
return &types.QueryPoolTypeResponse{Type: "Stableswap"}, nil
```

PR #2832, opened 2022-09-23
Add Pool Type Query

mattverse:
Closes: #2739 

Adds pool type query to query the type of pool. The query would return a string of the type of pool in a string, depending on the type of pool, would return an error otherwise.

## Brief Changelog
- Add pool type query 

## Testing and Verifying
Added new test cases 

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?yes


x/gamm/keeper/grpc_query.go
```diff
@@ -127,6 +127,27 @@ func (q Querier) NumPools(ctx context.Context, _ *types.QueryNumPoolsRequest) (*
...
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	pool, err := q.Keeper.GetPoolAndPoke(sdkCtx, req.PoolId)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	switch pool := pool.(type) {
+	case *balancer.Pool:
+		return &types.QueryPoolTypeResponse{PoolType: "Balancer"}, nil
```
ValarDragon, line 144: We should have this be exposed as a method on the pool interface, right?

Or at least, lets make a go API for `keeper.PoolType(ctx, poolId) (string, error)`, to have it in a more discoverable point

mattverse: Changed to having go API, I would prefer having further discussion in adding pool type to `PoolI`, I'd like to go ahead and open an issue for this for further discussion for this if you agree!

ValarDragon: SGTM

ValarDragon (approved):
Perfect!

Can you make a followup issue for discussing adding it to the poolI api?

mattverse:
https://github.com/osmosis-labs/osmosis/issues/2868

Issue #2845, opened 2022-09-24
[stableswap]: Combine `solveCfmm` and `solveCfmmMulti`


AlpinYukseloglu:
## Background

All pool-level CFMM calls currently only use `solveCfmm`, which is only connected to our two-asset CFMM solver. Many of these functions also hard code the assumption that there are only two assets (e.g. by always specifically taking the [first two](https://github.com/osmosis-labs/osmosis/blob/39b32018d736f93fbb4ecdc85d53dae457eec5a1/x/gamm/pool-models/stableswap/amm.go#L379) assets in pool reserves).

We should combine these two solver functions and properly direct inputs towards the correct solver depending on the number of assets they have.

We should also change any hard-coded assumptions of two-asset pools so pool functions all accommodate multi-asset pools.

Note: any PR for this issue should either be made after #2843 is merged or should be branched from it

## Suggested Design

- Merge `solveCfmmMulti` into `solveCfmm`
- Change `solveCfmm` to take a single `sdk.Coins` or `PoolAssets` input (latter might be safer if possible)
- Add logic in `solveCfmm` that routes to two-asset solver if only two assets and to multi-asset solver otherwise. This logic should calculate `uReserve` and `wSumSquares` internally from the input tokens to prevent unintuitive calculations elsewhere in the codebase and to prevent user error around these calcs in the future
- Update pool function logic to accommodate multi-asset pools e.g. here (there are others):
https://github.com/osmosis-labs/osmosis/blob/39b32018d736f93fbb4ecdc85d53dae457eec5a1/x/gamm/pool-models/stableswap/amm.go#L379

## Acceptance Criteria

- All existing and new tests pass

PR #2843, opened 2022-09-24
[stableswap]: Wire up binary search solvers to core stableswap logic

AlpinYukseloglu:
Closes: #1341, #2702, #2826

## What is the purpose of the change

This PR wires up our new binary search solvers to core stableswap logic through `solveCfmm` and `solveCfmmMulti`.

Note: majority of the diff is branched from #2825 – this PR makes minimal core logic changes and is meant mainly to tie up loose ends at the CFMM level so we can start thoroughly cleaning up and testing pool-level logic.

## Brief Changelog

- Wires up `solveCfmm` and `solveCfmmMulti` to use our new `BigDec` binary search solvers

## Testing and Verifying

- The existing suite of test cases for two-asset and multi-asset CFMMs is applied to both solvers in `amm_test.go`

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (not documented)

PR #2847, opened 2022-09-24
[stableswap]: Combine `solveCfmm` and `solveCfmmMulti`

AlpinYukseloglu:
Closes: #2845

This PR merges `solveCfmmMulti` into `solveCfmm` and updates tests and pool logic accordingly.

Note: majority of diff is from #2843 and test refactors (one side benefit of this change is that it allows our tests to be _drastically_ cleaner and more readable (no more manual `uReserve` and `wSumSquares` calculations!))

Context from issue:

> All pool-level CFMM calls currently only use solveCfmm, which is only connected to our two-asset CFMM solver. Many of these functions also hard code the assumption that there are only two assets (e.g. by always specifically taking the [first two](https://github.com/osmosis-labs/osmosis/blob/39b32018d736f93fbb4ecdc85d53dae457eec5a1/x/gamm/pool-models/stableswap/amm.go#L379) assets in pool reserves).
> 
> We should combine these two solver functions and properly direct inputs towards the correct solver depending on the number of assets they have.

## Brief Changelog

- Merge `solveCfmmMulti` into `solveCfmm`
- Change `solveCfmm` to take a single `sdk.Coins` or `PoolAssets` input (latter might be safer if possible)
- Add logic in `solveCfmm` that routes to two-asset solver if only two assets and to multi-asset solver otherwise. This logic should calculate `uReserve` and `wSumSquares` internally from the input tokens to prevent unintuitive calculations elsewhere in the codebase and to prevent user error around these calcs in the future
- Update pool function logic to accommodate multi-asset pools (previously would just take first two tokens in token reserves)

## Testing and Verifying

All tests previously relating to `solveCfmm` and `solveCfmmMulti` should now be properly refactored to use the new `solveCfmm`


ValarDragon:
Do we not want to keep the direct solvers at all?

ValarDragon (approved):
LGTM, aside from question of direct solver being kept around

AlpinYukseloglu:
> Do we not want to keep the direct solvers at all?
> 
> 
> 
> If nothing else, just for reference testing our binary search code?

We do! The direct solver in #2697 is called `solveCfmmDirect`, so I went ahead and deleted the old version in this PR expecting that one to get merged and take its place

PR #2857, opened 2022-09-25
Add a simulator executor types package

ValarDragon:

Adds a new package for executor types, and moves private simulator code out of simtypes into this new package. This further clarifies package responsibilities.

This also starts deleting legacy code for genesis instantiation, so we can simplify the genesis instantiation, and start passing things we are concerned about. Furthermore, we switch the genesis initialization to dealing with abci.RequestInitGenesis more directly

## Brief Changelog

  - Delete some unused codepaths in AppStateFn
  - Start moving some simtypes logic to an executor types package
  - Delete SimulationManager from being defined on app.go, instead we just expose the module manager.

## Testing and Verifying

make test-sim still passes

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no
  - How is the feature or change documented? docs beginning in README's


czarcas7ic (approved):
This all LGTM!

Issue #2287, opened 2022-08-03
Turn off all markdown lints by default, make explicit whitelist


ValarDragon:
## Background

Lets turn off all markdown lints, and then one by one enable markdown lints. Imo the markdown linter has generally been a huge waste of time, providing very little quality guarantees, and tons of quality flow time burned.

From multiple conversations, theres _very few_ markdown lints we care about. Lets specifically only enable those, and ensure that every single file in this repo passes those lints, 

## Suggested Design

Explicitly disable every markdown lint, make markdown lint apply to every single file, and then one by one re-enable things.

## Acceptance Criteria

- [ ] all but whitelist of Markdown lints disabled in all prior branches
- [ ] Markdown lint is set in GH to apply to every single file
- [ ] If anyone likes particular lints, must one by one PR re-adding it

alexanderbez:
Agree with the proposal!

>  Imo the markdown linter has generally been a huge waste of time, providing very little quality guarantees, and tons of quality flow time burned.

Don't necessarily agree with this though. The linter should guarantee consistency in style across markdown files :)

p0mvn:
What are you proposing to whitelist?

ValarDragon:
I personally only find URL lints, and codeblocks should have language specified useful.

p0mvn:
I haven't had any issues with the markdown lints besides some already disabled so I would say that all are fine with me.

I find heading levels, indentations, and line limits for non-code blocks, particularly useful.

I also would like to second Bez here that the main benefit I see is consistency and. I'm afraid that with the majority of the linters disabled, we would have a variety of styles in our markdowns, making them more challenging to follow. 

To me, the major cause of frustration with markdownlints occurs when someone merges a broken markdown CI lint PR. It doesn't break the CI for the next immediate pull request. However, It then breaks linter CI for someone else working on the same file in the future. 

We could mitigate that by turning [this](https://github.com/osmosis-labs/osmosis/blob/e33e776f4f45d63d4d9c8fa50e09992d80b36d4c/.github/workflows/lint.yml#L58) to true and fixing all existing issues. Then, if someone breaks the markdown linter - it is trivial to `git blame` and resolve the problem. That would help to prevent another person from dealing with issues they didn't cause in the future and remove a lot of frustration about the markdown lints.

What do you think?

PR #2863, opened 2022-09-26
Change markdownlinter to default rules to false, and we have to add more in

ValarDragon:

Closes: #2287 

Stops the time burn on things that don't matter, wrt most markdown lints. Most lints are things that all renderers handle, and aren't particularly obtrusive in any IDE.


.markdownlint.yml
```diff
@@ -1,16 +1,22 @@
...
-MD024: false
-MD029: false
-MD033: false
-MD037: false # breaks on latex
+MD018: true
+MD026: true
+MD034: true
+MD040: true
+MD042: true
+MD048: true
```
p0mvn, line 11: Link fragments should be valid

```suggestion
MD048: true
MD051: true
```

.markdownlint.yml
```diff
@@ -1,16 +1,22 @@
...
-MD007: false
-MD009: false
-MD010:
-  code_blocks: false
-MD013: false
-MD024: false
-MD029: false
-MD033: false
-MD037: false # breaks on latex
+MD018: true
```
p0mvn, line 6: - Heading levels should only increment by one level at a time

I think this helps to ensure that headings make sense / are consistent

```suggestion
MD001: true
MD018: true
```

p0mvn (approved):

.markdownlint.yml
```diff
@@ -1,16 +1,22 @@
...
+MD026: true
+MD034: true
+MD040: true
+MD042: true
+MD048: true
+# MD004: false
+# # Can't disable MD007 :/
```
p0mvn, line 13: Are we keeping these just in case we want to switch back?

alexanderbez (approved):

.markdownlint.yml
ValarDragon: I think so, we can def delete them later tho

PR #2867, opened 2022-09-26
Stop double running gamm tests

ValarDragon:

Caught this in re-reviewing all gamm changes for v12

Gamm keeper tests were getting double ran, due to:
```
keeper/keeper_test.go
23:	suite.Run(t, new(KeeperTestSuite))
```
and 
```
keeper/msg_server_test.go
20:	suite.Run(t, new(KeeperTestSuite))
```

This PR fixes this. 

## Testing and Verifying

Manually verified by doing `go test -v ./x/gamm/keeper/.` before and after this change.

### DIFF
--- a/x/gamm/keeper/msg_server_test.go
+++ b/x/gamm/keeper/msg_server_test.go
@@ -1,10 +1,7 @@
 package keeper_test
 
 import (
-	"testing"
-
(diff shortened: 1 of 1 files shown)


mattverse (approved):
Ah surprised that we still have some of the tests running twice. 

LGTM

Issue #2858, opened 2022-09-25
[stableswap][bug]: Fix int conversion bug that allows pools with negative scaled assets


AlpinYukseloglu:
## Background

Scaling factors for pool creation are of type `uint64`, but during the actual scaling process they are [converted to](https://github.com/osmosis-labs/osmosis/blob/bf93ac70226212c9f8ce1f9796688e9d6c118353/x/gamm/pool-models/stableswap/pool.go#L126) `int64`, making it possible for sufficiently large `uint64` scaling factors to flip the signed bit and become negative.

## Suggested Design

- Add a `maxScalingFactor` of `1 << 62` to pool creation logic (`1 << 63` flips the signed bit and makes the number negative)
- Add checks before the `uint64` -> `int64` conversion in scaling and descaling logic that panics if a scaling factor is greater than `1 << 62`
https://github.com/osmosis-labs/osmosis/blob/bf93ac70226212c9f8ce1f9796688e9d6c118353/x/gamm/pool-models/stableswap/pool.go#L126 https://github.com/osmosis-labs/osmosis/blob/bf93ac70226212c9f8ce1f9796688e9d6c118353/x/gamm/pool-models/stableswap/pool.go#L137
- Add relevant check to `ValidateBasic()` in `msgs.go`

## Acceptance Criteria

- Panic catching tests for `1 << 62` scaling factors are created and pass
- Error catching tests for scaling factors greater than `maxScalingFactor` are created and pass

Issue #2861, opened 2022-09-26
[stableswap]: Add scaling factor tests


AlpinYukseloglu:
## Background

Scaling factors currently have a basic test skeleton commented out in the keeper tests that needs to be moved to our `pool_test.go` and filled with test cases for scaling/descaling pool assets

## Suggested Design

- Move commented tests from keeper tests to `pool_test.go`
- Add thorough test cases (even/uneven pools, even/uneven scaling factors, min and max scaling factors, overflow/underflow, panic catching etc.)

## Acceptance Criteria

- All new and existing tests pass

AlpinYukseloglu:
@hieuvubk I saw that you self-assigned - I was working on this alongside a few other scaling factor changes so will unassign you for now (I hope you weren't too far into it!).

In general I will probably be picking up (or already have significant progress for) the stableswap-related issues I create unless I tag them as 'Good First Issue' :)

hieuvubk:
Okay got it, thank you!

PR #2874, opened 2022-09-26
[stableswap]: Add checks and tests for scaling factors and extend for multi-asset pools

AlpinYukseloglu:
Closes: #2858 and #2861

This PR adds checks and tests for core scaling factor logic. It also extends core pool logic to support scaling factors for multi-asset pools.

Note: 95% of the commit history is past PRs that are already merged (was working off of them in parallel)

## Brief Changelog

- Ensure that it is not possible to reach negative scaling factors by setting scaling factors large enough to flip the sign bit during `uint64` -> `int64` conversions
- Add thorough tests for scaling and descaling logic
- Expand core pool logic to allow for multi-asset scaling factors (instead of just taking the first two in the array)
- Move testing logic from commented out scaling factor tests in `pool_service_test.go` at the keeper layer
- Minor refactor for `solveCfmm` (unrelated but simple enough to tie in with other checks)

## Testing and Verifying

- All tests in this PR are in `pool_test.go`


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -58,15 +58,17 @@ func cfmmConstantMulti(xReserve, yReserve, uReserve, vSumSquares osmomath.BigDec
...
+	} else if len(remReserves) > 0 {
+		uReserve := osmomath.OneDec()
+		wSumSquares := osmomath.ZeroDec()
+		for _, assetReserve := range remReserves {
+			uReserve = uReserve.Mul(assetReserve)
+			wSumSquares = wSumSquares.Add(assetReserve.Mul(assetReserve))
+		}
+		return solveCFMMBinarySearchMulti(cfmmConstantMulti)(xReserve, yReserve, uReserve, wSumSquares, yIn)
+	} else {
+		panic("invalid input reserves for CFMM solver")
```
p0mvn, line 70: This case should not be impossible since the previous two check cover all possible range of values of `len(remReserves)`

x/gamm/pool-models/stableswap/msgs.go
```diff
@@ -58,6 +58,14 @@ func (msg MsgCreateStableswapPool) ValidateBasic() error {
 		return types.ErrInvalidScalingFactors
 	}
 
+	// iterates len(msg.InitialPoolLiquidity) < 10 times
```
p0mvn, line 61: Can we cover this with test please?

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -123,8 +127,14 @@ func (p Pool) getScaledPoolAmts(denoms ...string) ([]sdk.Dec, error) {
 			return []sdk.Dec{}, fmt.Errorf("denom %s does not exist in pool", denom)
 		}
 		scalingFactor := p.GetScalingFactorByLiquidityIndex(liquidityIndex)
+
+		// should not be possible after pool creation checks but we keep this here as a guardrail
+		if int64(scalingFactor) <= 0 {
```
p0mvn, line 132: nit / non-blocking: at first sight, feels that this might be an overkill / additional complexity given that we already test for this in the constructor, `ValidateBasic`  and `SetStableSwapScalingFactors`

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -58,15 +58,17 @@ func cfmmConstantMulti(xReserve, yReserve, uReserve, vSumSquares osmomath.BigDec
...
 	}
```
ValarDragon, line 71: We should undo this change back to the old code?

This is an unnecessary else if

x/gamm/pool-models/stableswap/msgs.go
ValarDragon, line 61:
```suggestion
```

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -24,8 +24,12 @@ var _ types.PoolI = &Pool{}
...
 		scalingFactors = []uint64{1, 1}
-	} else if scalingFactors[0] == 0 || scalingFactors[1] == 0 {
-		return Pool{}, types.ErrInvalidStableswapScalingFactors
+	}
+
+	for _, scalingFactor := range scalingFactors {
+		if scalingFactor == 0 || int64(scalingFactor) <= 0 {
+			return Pool{}, types.ErrInvalidStableswapScalingFactors
+		}
 	}
```
ValarDragon, line 33: This is reused, lets extract this to a function in stableswap for validate scaling factors

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -299,15 +315,16 @@ func (p *Pool) PokePool(blockTime time.Time) {}
 // It should only be able to be successfully called by the pool's ScalingFactorGovernor
 // TODO: move commented test for this function from x/gamm/keeper/pool_service_test.go once a pool_test.go file has been created for stableswap
 func (p *Pool) SetStableSwapScalingFactors(ctx sdk.Context, scalingFactors []uint64, scalingFactorGovernor string) error {
-	if scalingFactorGovernor != p.ScalingFactorGovernor {
+	if scalingFactorGovernor != p.ScalingFactorGovernor || len(scalingFactors) != p.PoolLiquidity.Len() {
```
p0mvn, line 318: Non-blocking for this PR:

I think checking for governor here is not compatible with the abstraction.

I would like to suggest moving this governor check up to the caller. Then, we can also reuse `SetStableSwapScalingFactors` method in the constructor to reduce code duplication for checking scaling factor overflow

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -123,8 +127,14 @@ func (p Pool) getScaledPoolAmts(denoms ...string) ([]sdk.Dec, error) {
 			return []sdk.Dec{}, fmt.Errorf("denom %s does not exist in pool", denom)
 		}
 		scalingFactor := p.GetScalingFactorByLiquidityIndex(liquidityIndex)
+
+		// should not be possible after pool creation checks but we keep this here as a guardrail
+		if int64(scalingFactor) <= 0 {
+			panic("invalid scaling factor encountered when scaling pool assets")
+		}
```
ValarDragon, line 134: Not sure how I feel about these being everywhere, but if you want this lets add it into the get method, since the property is that this should never get into state.

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -299,15 +315,16 @@ func (p *Pool) PokePool(blockTime time.Time) {}
 // It should only be able to be successfully called by the pool's ScalingFactorGovernor
 // TODO: move commented test for this function from x/gamm/keeper/pool_service_test.go once a pool_test.go file has been created for stableswap
 func (p *Pool) SetStableSwapScalingFactors(ctx sdk.Context, scalingFactors []uint64, scalingFactorGovernor string) error {
-	if scalingFactorGovernor != p.ScalingFactorGovernor {
+	if scalingFactorGovernor != p.ScalingFactorGovernor || len(scalingFactors) != p.PoolLiquidity.Len() {
 		return types.ErrNotScalingFactorGovernor
 	}
```
ValarDragon, line 320: Err should be a different error message and therefore a second check

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -299,15 +315,16 @@ func (p *Pool) PokePool(blockTime time.Time) {}
...
 
-	if len(scalingFactors) != p.PoolLiquidity.Len() {
-		return types.ErrInvalidStableswapScalingFactors
+	for _, scalingFactor := range scalingFactors {
+		if int64(scalingFactor) <= 0 {
+			return types.ErrInvalidStableswapScalingFactors
+		}
 	}
```
ValarDragon, line 326: Another point to reuse the scaling factor validation method

ValarDragon (approved):
LGTM once my comments are addressed

x/gamm/pool-models/stableswap/pool.go
ValarDragon: The argument should be `sender` here, and this is the entry point of the `SetStableSwapScalingFactors` function right?

x/gamm/pool-models/stableswap/msgs.go
AlpinYukseloglu: This was a note that I intended to remove, not sure why it's still there (the number of iterations is capped at the max number of pool assets by definition)

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: Reverted

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: Reverted

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: Great point – extracted into `validateScalingFactors` function

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: Check moved into `validateScalingFactors` so this should be addressed

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: I ended up removing this check and the one in `getDescaledPoolAmt` to reduce complexity as all cases should be covered by higher level checks

AlpinYukseloglu:
Thanks for the thorough review! All comments should be addressed

p0mvn (approved):

alexanderbez (approved):

Issue #2282, opened 2022-08-02
Create SuperfluidDelegationsByValidator and SuperfluidDelegationsByAsset GRPC Queries


Pharaon1993:
## Background

ATM, the SFS module has [ABCI queries](https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/superfluid/query.proto) for the total amount of Osmos SFS staked & some more granular queries. Specifically, there are queries for SFSDelegationsByDelegator and SuperFluidDelegationsByValidatorDenom.

I run an [analytics dashboard](https://hathornodes.com/super_fluid_staking/osmosis) for tracking SFS usage. To get the total number of Osmos SFS delegated by a pool or validator, I have to make `V*A` ABCI Queries where V is the number of Validators and A is the number of SFS Assets. Even if I limit my searches to the active set, that is `135 * 15 = 2025` ABCI Queries ATM. 

If either number grows, which is a reasonable expectation, then this number increases dramatically. I can't use SFSDelegationsByDelegators as there's no ABCI Query for the entire list of delegators using SFS. I'd have to get a state export and then make an ABCI query for every user w/ a 14 day lock to see if they have any SFS Delegations. The number of users w/ 14 day locks far exceeds `V*A`, ~156k 14 Day Locks when I last checked.

## Suggested Design

Implement SFSDelegationsByValidator and SFSDelegationsByDenom ABCI Queries. They don't have to return the estimation of Osmos delegated. Simply returning the delegations for a Validator/Asset. This would reduce the number of queries to V + A and remove any scalability concerns.

An alternative would an ABCI Query that lists all SFSDelegations. I can't think of a similar query in the CosmosSDK so that may break from the usual convention.  There also may be bandwidth/spam concerns (i.e. someone spam requesting a large ABCI Query from a node to increase its bandwidth out significantly) with this solution.

ValarDragon:
Listing all SFS delegations is going to be a bit hard, because then you have to paginate it. Right now pagination in the SDK is too naive, and means you have to literally read pages 0 through N, in order to return page N+1...

Sounds like we just need one of the two queries supported right?

If I understand correctly, we could make the query `TotalDelegationByValidatorForAsset( asset: string ) [](validator addr, amount of asset SFS'd, equivalent osmo amount)` right?

Pharaon1993:
That works! Users can aggregate from there if needed and it reduces the number of queries needed too.

PR #2881, opened 2022-09-28
[superfluid] TotalDelegationByValidatorForDenom query

RusAkh:
Closes: #2282

Query for getting all sfs'd assets for each validator

## Brief Changelog

added query and test

## Testing and Verifying

unit test


x/superfluid/keeper/grpc_query.go
```diff
@@ -392,6 +392,41 @@ func (q Querier) EstimateSuperfluidDelegatedAmountByValidatorDenom(goCtx context
...
+		val, found := q.Keeper.sk.GetValidator(ctx, valAddr)
+		if !found {
+			return nil, stakingtypes.ErrNoValidatorFound
+		}
+		equivalentAmountOSMO := q.Keeper.GetSuperfluidOSMOTokens(ctx, req.Denom, val.Tokens)
+		result := &types.Delegations{
+			ValAddr:        valAddr.String(),
+			AmountSfsd:     &val.Tokens,
+			OsmoEquivalent: &equivalentAmountOSMO,
+		}
```
mattverse, line 420: hrmm this seems wrong to me. Returning val.Tokens would return all the tokens on the validator, not the amount on the intermediary account. Instead I think we should be doing
```
delegation, found := suite.App.StakingKeeper.GetDelegation(suite.Ctx, expAcc.GetAccAddress(), valAddr)
```

and then be using delegation.shares to track how much is sfsd

x/superfluid/keeper/grpc_query_test.go
```diff
@@ -16,6 +16,58 @@ func (suite *KeeperTestSuite) TestGRPCParams() {
 	suite.Require().True(res.Params.MinimumRiskFactor.Equal(types.DefaultParams().MinimumRiskFactor))
 }
 
+func (suite *KeeperTestSuite) TestTotalDelegationByValidatorForAsset() {
+	suite.SetupTest()
+	ctx := suite.Ctx
+	querier := suite.querier
+	delegation_amount := int64(1000000)
+
+	var denom string
```
mattverse, line 25:
```suggestion

```

mattverse (commented):
Hey @RusAkh! Thanks for this PR! I do agree with you that this query is something that we def need!

I think ur on the right track on this PR, just that how sfsd asset amount is calculated seems that it needs further changes to me. Left some comment, please take a look!

RusAkh:
Thanks for the review! I will work on it!

x/superfluid/keeper/grpc_query.go
RusAkh: @mattverse I do not really understand why do we need to get amount of an intermediary account if we want to get assets of each validator for a specific denomination? I would be very thankful for further clarification, please

RusAkh: other comments were addressed

mattverse: Awesome! Good question, here we want to return the amount superfluid staked, not the amount total staked on the validator right? Say for example valA has total of 100uosmo staked, 50uosmo staked natively, 50uosmo staked via superfluid staking . `val.Tokens` would be returning 100uosmo, whilst using getDelegation would be returning 50uosmo, only the amount that has been staked

RusAkh: oh, that was an awesome example! Thank you! I will try to figure out how to fix that!

RusAkh: @mattverse this comment was also addressed

x/superfluid/keeper/grpc_query_test.go
```diff
@@ -16,6 +16,55 @@ func (suite *KeeperTestSuite) TestGRPCParams() {
...
+		{0, 0, 0, delegation_amount},
+		{0, 1, 1, delegation_amount},
+		{1, 0, 1, delegation_amount},
+		{1, 1, 0, delegation_amount},
+	}
+
+	suite.setupSuperfluidDelegations(valAddrs, superfluidDelegations, denoms)
+
+	for _, denom := range denoms {
+		req, err := querier.TotalDelegationByValidatorForDenom(sdk.WrapSDKContext(ctx), &types.QueryTotalDelegationByValidatorForDenomRequest{Denom: denom})
```
mattverse, line 38:
```suggestion
		res, err := querier.TotalDelegationByValidatorForDenom(sdk.WrapSDKContext(ctx), &types.QueryTotalDelegationByValidatorForDenomRequest{Denom: denom})
```

x/superfluid/keeper/grpc_query_test.go
```diff
@@ -16,6 +16,55 @@ func (suite *KeeperTestSuite) TestGRPCParams() {
...
+
+		suite.Require().NoError(err)
+		suite.Require().Equal(len(valAddrs), len(req.AssetResponse))
```
mattverse, line 41:
```suggestion
		suite.Require().Equal(len(valAddrs), len(res.AssetResponse))
```

x/superfluid/keeper/grpc_query_test.go
```diff
@@ -16,6 +16,55 @@ func (suite *KeeperTestSuite) TestGRPCParams() {
...
+
+		for res_ind, res := range req.AssetResponse {
+			// check osmo equivalent is correct
+			actual_response_osmo := *req.AssetResponse[res_ind].OsmoEquivalent
+			needed_response_osmo := suite.App.SuperfluidKeeper.GetSuperfluidOSMOTokens(ctx, denom, sdk.NewInt(delegation_amount))
+
+			suite.Require().Equal(actual_response_osmo, needed_response_osmo)
+
```
mattverse, line 49: should we also add check for correct osmo equivilent amount in the res?

x/superfluid/keeper/grpc_query.go
```diff
@@ -392,6 +392,46 @@ func (q Querier) EstimateSuperfluidDelegatedAmountByValidatorDenom(goCtx context
 	}, nil
 }
 
+func (q Querier) TotalDelegationByValidatorForDenom(goCtx context.Context, req *types.QueryTotalDelegationByValidatorForDenomRequest) (*types.QueryTotalDelegationByValidatorForDenomResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	var delegationsByValidator = []*types.Delegations{}
```
mattverse, line 398: So I figured using pointers here is awkward, which made me follow into the rabbit hole until I noticed that the response struct itself is taking pointer referenced array. 

```
type QueryTotalDelegationByValidatorForDenomResponse struct {
	AssetResponse []*Delegations `protobuf:"bytes,1,rep,name=asset_response,json=assetResponse,proto3" json:"asset_response,omitempty"`
}
```

whilst I think it should instead be 

```

type QueryTotalDelegationByValidatorForDenomResponse struct {
	AssetResponse []Delegations `protobuf:"bytes,1,rep,name=asset_response,json=assetResponse,proto3" json:"asset_response,omitempty"`
}
```
just like other proto response structs do. I do think that the possible solution for this is adding `[ (gogoproto.nullable) = false ];` to the proto field. Please lmk if it works out

mattverse (changes requested):
Going through a second round of review, I think we need change in how the proto struct response returns a pointer referenced slice.

proto/osmosis/superfluid/query.proto
```diff
@@ -53,6 +53,10 @@ service Query {
         "/osmosis/superfluid/v1beta1/connected_intermediary_account/{lock_id}";
   }
 
+  rpc TotalDelegationByValidatorForDenom(
```
alexanderbez, line 56: nit: can we add docs for new proto types please 🙏

x/superfluid/keeper/grpc_query.go
alexanderbez: We should definitely do this ^

alexanderbez (changes requested):

RusAkh:
@alexanderbez @mattverse changes addressed

proto/osmosis/superfluid/query.proto
```diff
@@ -153,6 +158,18 @@ message ConnectedIntermediaryAccountResponse {
...
+  repeated Delegations asset_response = 1;
+}
+
+message Delegations {
+  string val_addr = 1;
+  string amount_sfsd = 2
+      [ (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int" ];
+  string osmo_equivalent = 3
+      [ (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int" ];
+}
```
mattverse, line 172:
```suggestion
  string amount_sfsd = 2 [
      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
      (gogoproto.moretags) = "yaml:\"amount_sfsd\""
    ];
  string osmo_equivalent = 3 [
      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int" ],
      (gogoproto.moretags) = "yaml:\"osmo_equivalent\""
 ];  
}
```

x/superfluid/keeper/grpc_query.go
mattverse: Could we address this before merging? Thanks!

RusAkh:
@mattverse done

x/superfluid/keeper/grpc_query.go
```diff
@@ -392,6 +392,46 @@ func (q Querier) EstimateSuperfluidDelegatedAmountByValidatorDenom(goCtx context
 	}, nil
 }
 
+func (q Querier) TotalDelegationByValidatorForDenom(goCtx context.Context, req *types.QueryTotalDelegationByValidatorForDenomRequest) (*types.QueryTotalDelegationByValidatorForDenomResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	var delegationsByValidator = []types.Delegations{}
+	var intermediaryAccount types.SuperfluidIntermediaryAccount
```
alexanderbez, line 399: nit: group vars :p

alexanderbez (approved):

proto/osmosis/superfluid/query.proto
```diff
@@ -153,6 +158,24 @@ message ConnectedIntermediaryAccountResponse {
   SuperfluidIntermediaryAccountInfo account = 1;
 }
 
+message QueryTotalDelegationByValidatorForDenomRequest { string denom = 1; }
+message QueryTotalDelegationByValidatorForDenomResponse {
+  repeated Delegations asset_response = 1 [ (gogoproto.nullable) = false ];
```
mattverse, line 163:
```suggestion
  repeated Delegations assets = 1 [ (gogoproto.nullable) = false ];
```

mattverse (approved):
LGTM! Nice work!

Issue #2807, opened 2022-09-21
[osmoutils]: Add tests with different upper/lower bounds to osmoutil binary search


AlpinYukseloglu:
## Background

Ref: https://github.com/osmosis-labs/osmosis/pull/2802#discussion_r975861292

We should test our binary search implementation(s) on different bounds (currently all of them use the same min and max value).

## Suggested Design

Cases to add:
- Nonzero lower bounds
- Equivalent upper and lower bounds (should panic)
- Lower bound greater than upper bound (should panic)
- Negative bounds (should panic)
- Overflowing bounds (should panic)
- Different upper bounds (just so all the tests aren't identical)

## Acceptance Criteria

- Most/all of the cases above are added and pass

RusAkh:
I was working on this issue: every test case seemed fine and passed fine except for this one: 
```
"different upper bounds":     {cubicF, osmomath.OneDec(), osmomath.NewBigDec(1 << 10), osmomath.NewBigDec(1 + (1 << 25)), lowErrTolerance, 20000, false},
```
So basically the problem is when you specify upperbounds  (3rd struct field) to be less than ```1 << 49``` it throws this error:
```
 hit maximum iterations, did not converge fast enough
```
As seen from this test case, I tried increasing the maxIterations field (20000), however, it did not work. 



PS: I removed ```actualSolvedInput``` field from ```tests``` map, so now test routine looks like this: 
```
actualSolvedInput, err := BinarySearchBigDec(tc.f, tc.lowerbound, tc.upperbound, tc.targetOutput, tc.errTolerance, tc.maxIterations)

			if tc.expectErr {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
				output, err := tc.f(actualSolvedInput)
				require.NoError(t, err)

				require.Equal(t, tc.errTolerance.CompareBigDec(output, tc.targetOutput), 0) // 0 - within errTolerance
			}
```

AlpinYukseloglu:
Do you have your work on a fork/branch by any chance? The likeliest scenario is that the values being considered were too large for even our `BigDec` precision, preventing the search from ever converging (if this is the case then it would have triggered an infinite loop so increasing `maxIterations` wouldn't do anything)

If this is the case then it is intended behavior as precision will always be a bottleneck for sufficiently large binary search inputs

RusAkh:
[...]

PR #2884, opened 2022-09-28
[osmoutils] Added test cases to binary search

RusAkh:
Closes: #2807 

Additional test cases for binary search tests

## Testing and Verifying


osmoutils/binary_search_test.go
```diff
@@ -77,8 +86,8 @@ func TestBinarySearchBigDec(t *testing.T) {
 	cubicF := func(a osmomath.BigDec) (osmomath.BigDec, error) {
 		// these precision shifts are done implicitly in the int binary search tests
 		// we keep them here to maintain parity between test cases across implementations
-		calculation := a.Quo(osmomath.NewBigDec(10).Power(18))
-		result := calculation.Power(3)
+		calculation := a.Power(3)
+		result := calculation.Quo(osmomath.NewBigDec(10).Power(18))
 		output := result.Mul(osmomath.NewBigDec(10).Power(18))
```
AlpinYukseloglu, line 91: Why do we flip these two here? 1. The original ordering was intentional to have the tests aligned with our regular binary search tests and 2. the final version of this divides by 10^18 then multiplies by it right after

I suggest we revert this to maintain parity across binary search tests

osmoutils/binary_search_test.go
```diff
@@ -33,8 +33,7 @@ func TestBinarySearch(t *testing.T) {
 		errTolerance  ErrTolerance
 		maxIterations int
 
-		expectedSolvedInput sdk.Int
```
AlpinYukseloglu, line 36: I like the idea of directly checking `targetOutput`, but I don't see why we need to remove `expectedSolvedInput` to ensure intermediate values are correct as well. I would personally prefer to keep both checks

AlpinYukseloglu (changes requested):
Good work! LGTM once my comments are addressed

osmoutils/binary_search_test.go
RusAkh: oh, I left a comment under an initial issue https://github.com/osmosis-labs/osmosis/issues/2807#issuecomment-1260660342. Long story short this function seemed not to work correctly, so I decided to do some swaps and it started working fine afterwards. More details about it are in that comment. I really hope it will clarify my decision!

PS: my bad I forgot to mention that cubicF also seemed to work incorrectly in TestBinarySearch as well, for example:

```
DEBUG:  1024195081273344 (function input) -> 0.001024195081273344 (dec version) -> 1074355614 (output)
```

osmoutils/binary_search_test.go
RusAkh: I thought there is no actual need for this intermediary check, because by checking that 
```
require.Equal(t, tc.errTolerance.CompareBigDec(output, tc.targetOutput), 0) // 0 - within errTolerance
```
outputs match, we can imply that an actualSolvedInput is correct indeed, because an output's value depends on it
```
output, err := tc.f(actualSolvedInput)
```
I might be wrong about this and if that is the case I will return an expectedSolvedInput back

Issue #1850, opened 2022-06-22
Make a daily workflow to check that every link isn't broken


ValarDragon:
## Background

A common problem is that links to other sites in the code / documentation break. We should have a daily CI action that can scan every markdown file, and report on breaking links in a github issue / slack bot

## Acceptance Criteria

- Automated reporting of link breakage {exists}

MasterPi-2124:
that's a coincidence that I was also working on this a few minutes ago. I created a workflow in notional fork and here is the run: https://github.com/notional-labs/osmosis/actions/runs/4498700168/jobs/7915641640

PR #2893, opened 2022-09-29
fix joining mainnet link in readme

Nostradamus411:
The link for join mainnet was directing to a no longer found page, updated with current docs location.


mattverse (approved):

ValarDragon:
Thank you!

cref #1850 for fixing this general problem

PR #2896, opened 2022-09-29
stableswap: Add multi-asset support threading, via scaled reserve API change

ValarDragon:

All of the stableswap functions assumed 2 asset pools still for getting the scaled reserves. This PR fixes this, by making methods to get all scaled reserves at once, but reordering the first and second field as needed.

We may want to change the result type from DecCoin to BigDec in a second pass though

## Testing and Verifying

This change is already covered by existing tests, `TestScaledSortedPoolReserves`.

Added tests for multi-asset pools


AlpinYukseloglu (approved):
LGTM!

Issue #2902, opened 2022-09-29
[x/gamm][stableswap]: Create scaling/descaling functions for token inputs/outputs


AlpinYukseloglu:
## Background

We currently have functions to scale/descale pool reserves, but not ones to scale/descale inputs. This leads to odd behavior where unscaled inputs are passed into scaled pool reserves. 

## Suggested Design

- Apply same logic used in pool reserve scaling but modified to accommodate specific token inputs and outputs (notes/ideas: return slice with length of input []DecCoin, loop through pool liquidity and add scaled output to return slice if denom exists in input array)

## Acceptance Criteria

- All existing and new tests pass

PR #2904, opened 2022-09-29
[x/gamm][stableswap]: Create scaling/descaling functions for token inputs/outputs

AlpinYukseloglu:
Closes: #2902

This PR adds a scaling function for token inputs. Depending on what level we want to handle rounding at (at the scaling function level vs. at the `SwapExactAmountIn` level), we can either use this function for outputs as well or create a second one that handles outputs with different rounding.

I personally think having one function with rounding handled at a higher level of abstraction is cleaner because it has less repeat code, but I can see the security argument for having multiple scaling functions that round differently for each specific use (cc: @ValarDragon)

## Brief Changelog

- Add `scaledInput` function and tests

## Testing and Verifying

- The function added in this PR is tested in `pool_test.go` under `TestScaledInput`


AlpinYukseloglu:
Switched to draft to add descaling as well

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -132,6 +133,23 @@ func (p Pool) scaledPoolReserves() ([]sdk.DecCoin, error) {
 	return scaledReserves, nil
 }
 
+// scaledInput returns scaled input tokens for usage in AMM equations
+func (p Pool) scaledInput(input sdk.Coins) ([]sdk.DecCoin, error) {
```
ValarDragon, line 137:
```suggestion
func (p Pool) scaledInput(input sdk.Coin) (sdk.DecCoin, error) {
```

ValarDragon (approved):

ValarDragon:
Should change division to method in #2908 once thats merged

x/gamm/pool-models/stableswap/pool.go
AlpinYukseloglu: I originally had this as `sdk.Coins` to be able to accommodate inputs for all-asset joins as well – will leave it as `sdk.Coin` for now as it's simpler and aligned with the descaling function (which deals with 1 asset) and handle all asset joins at a higher level of abstraction

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -104,6 +105,22 @@ func (p Pool) NumAssets() int {
 	return len(p.PoolLiquidity)
 }
 
+// scaledInput returns scaled input tokens for usage in AMM equations
+func (p Pool) scaledInput(input sdk.Coin, roundingDirection osmomath.RoundingDirection) (sdk.DecCoin, error) {
```
ValarDragon, line 109:
```suggestion
func (p Pool) scaleInputAmount(input sdk.Coin, roundingDirection osmomath.RoundingDirection) (osmomath.BigDec, error) {
```

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -104,6 +105,22 @@ func (p Pool) NumAssets() int {
...
+	scaledAmount, err := osmomath.DivIntByU64ToBigDec(input.Amount, scalingFactor, roundingDirection)
+	if err != nil {
+		return sdk.DecCoin{}, err
+	}
+	scaledInput := sdk.NewDecCoinFromDec(
+		input.Denom,
+		scaledAmount.SDKDec())
+
+	// TODO: correctly handle rounding
+	return scaledInput, nil
```
ValarDragon, line 121:
```suggestion
	if err != nil {
		return osmomath.BigDec{}, err
	}
	return scaledAmount, nil
```

PR #2912, opened 2022-09-30
Stableswap: swap, scale rounding & swapfee alignment w/ spec

ValarDragon:

- Adds spec for swap methods
- Aligns swap code's scale rounding calls, and swap fee to be in line with spec

## Testing and Verifying

This code remains untested


AlpinYukseloglu (approved):
LGTM! I added tests + a few other major changes as a PR to this branch (to not block this one and because it ended up leading to some other broader changes)

Issue #2487, opened 2022-08-22
ApplyFuncIfNoErr: catch out of gas panics, write gas amount anyway


ValarDragon:
## Background

Right now if a tx out of gas panics during an ApplyFuncIfNoErr block, we log a spooky looking message, and don't correctly write the out of gas up, to ensure the tx gets fully reverted. In concrete instantiations today, due to superfluid, we are safe. But this should be addressed to make the API more generally safe, and to remove the ominous logs.

## Suggested Design

https://github.com/osmosis-labs/osmosis/blob/main/osmoutils/cache_ctx.go#L16-L34

- [ ] Find a reliable way to detect if a panic is an out of gas (we may need to edit SDK fork ante handler, or fork our own gas handler -- tbh SDK out of gas handling is pretty insane)
- [ ] If out of gas, panic catch, then out of gas panic again. Write a test that this gets caught as a normal out of gas panic would get caught in a tx

08d2:
It's insane that running out of gas triggers a panic instead of returning an error. Does that happen in the SDK? Is it infeasible to fix it there?

ValarDragon:
Yeah it happens in the SDK. In general the SDK's guarantees around gas are pretty scant, so its a huge pain to fix things like this. The decision is also fairly complex, because the SDK framework only has the opportunity at the end of a message exec to cleanly alter execution flow, without a panic to abort the flow. (This is unlike with a VM interpreter, where you can just halt at the next instruction. The base 'instruction' unit of the SDK is poorly documented, but is the sdk.Msg)

I don't want to derail this issue too much with the SDK's design to panic for gas control flow. I am also in favor of fixing it, its just not the easiest.

PR #2915, opened 2022-09-30
Remove logging in out of gas panic recovery behavior

ValarDragon:
Closes: #2487 

In an apply func if no err block, if theres an out of gas panic, we currently abort the concerning code block and the gas change. 

This PR changes it to be such that if were in such a situation, we still commit the gas amount. (And give better log messages)

So during tx flow, we should re-panic and get old behavior.
In Begin/EndBlock, nothing should really happen, since were just adding the diff in consumption which should not overflow.

## Brief Changelog

- Fix ApplyFuncIfNoErr edge cases with gas limit (no current safety issue due to atomic reversion just cleans up edge case)

## Testing and Verifying

This requires tests to be written

  - How is the feature or change documented? Code comments


ValarDragon:
Wait what, why did this get closed, this was a great PR :(

ValarDragon:
On further investigation, the cache ctx and base ctx (very confusingly) share the same gas meter. I feel like thats an SDK bug (or at least undocumented behavior), but given that reverting the actual consumption logic, and its purely a matter of cosmetically improving logs then.

ValarDragon:
Closing this, since last two commits change implementation / invalidate original PR goal. making as a new PR

Issue #2909, opened 2022-09-30
Possible unreported error from GetArithmeticTwap


migueldingli1997:
#### System information

Osmosis version: `v12.0.0`
OS & Version: OSX
Commit hash: `aa727b5bf25019d561da5d6f82dd9a3892fda2be`

#### Expected behaviour

Assume `t0 < t1 < t2`, where these are time values

Say the store has the following two TWAP records:
- `{time: t0, lastErrorTime: t0}` (erroneous)
- `{time: t2, lastErrorTime: t0}`

And say we want to [GetArithmeticTwap](https://github.com/osmosis-labs/osmosis/blob/v12.0.0/x/twap/api.go#L31) with `startTime=t1` and `endTime=t2`. When GetArithmeticTwap is called, the [getInterpolatedRecord](https://github.com/osmosis-labs/osmosis/blob/aa727b5bf25019d561da5d6f82dd9a3892fda2be/x/twap/logic.go#L195) function will:
- Obtain the record at or before t1, which is `{time: t0, lastErrorTime: t0}`
- Interpolate this record to `{time: t1, lastErrorTime: t0}` (time updated)
- Obtain the record at or before t2, which is `{time: t2, lastErrorTime: t0}`
- Interpolate this record to `{time: t2, lastErrorTime: t0}`

My understanding here is that the first record should still be considered erroneous even after it was interpolated, since the interpolation is based on erroneous values. The interpolated records are fed into [computeArithmeticTwap](https://github.com/osmosis-labs/osmosis/blob/aa727b5bf25019d561da5d6f82dd9a3892fda2be/x/twap/logic.go#L228), which I expect to return an error due to the use of interpolated records based on erroneous values.

#### Actual behaviour

The computeArithmeticTwap function errors only if `endRecord.LastErrorTime` (i.e. `t0`) is after or equal to `startRecord.Time` (i.e. `t1`). In our case, `t0 < t1` and so the function does **not** error.

#### Steps to reproduce the behaviour

Add this failing test to [x/twap/api_test.go](https://github.com/osmosis-labs/osmosis/blob/aa727b5bf25019d561da5d6f82dd9a3892fda2be/x/twap/api_test.go):

```go
func (s *TestSuite) TestGetArithmeticTwap_ExpectError() {
	t0 := baseTime
	t1 := tPlusOne
	t2 := tPlusOne.Add(time.Second)
	
[...]

PR #2923, opened 2022-10-01
Fix: Last error time now updates during record interpolation

migueldingli1997:
Closes: #2909

This pull request resolves the issue that interpolated records that have an error were no longer considered erroneous after interpolation. In other words, if the last error time matches the record's time, these will also match in the interpolated record.

## Brief Changelog

  - Last error time is updated during record interpolation if record has an error

## Testing and Verifying

  - Added unit test that ensures the last error time is updated if the record's time matches the last error time

  - Does this pull request introduce a new feature or user-facing behavior changes? **no**
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? **yes**
  - How is the feature or change documented? **not documented**


ValarDragon:
This looks good to me!

Can you add the test case you had in the github issue, which ensures the needed property at the api level?

migueldingli1997:
> This looks good to me!
> 
> Can you add the test case you had in the github issue, which ensures the needed property at the api level?

Sure thing

migueldingli1997:
I have not had the time to continue working on this, however please note that that after I added some additional rough test cases, there seems to be an issue at the API level as a result of my changes.

x/twap/api_test.go
```diff
@@ -719,7 +742,7 @@ func (s *TestSuite) TestGetArithmeticTwapToNow() {
 
 			if test.expectedError != nil {
 				s.Require().Error(err)
-				s.Require().ErrorIs(err, test.expectedError)
+				s.Require().Equal(test.expectedError, err)
```
migueldingli1997, line 745: The `ErrorIs` does not work for `spotPriceError` so I changed it to an `Equals`

migueldingli1997:
Updated @ValarDragon :)

x/twap/logic.go
```diff
@@ -225,14 +229,17 @@ func (k Keeper) getMostRecentRecord(ctx sdk.Context, poolId uint64, assetA, asse
...
 // if (endRecord.Time == startRecord.Time) returns endRecord.LastSpotPrice
 // else returns
 // (endRecord.Accumulator - startRecord.Accumulator) / (endRecord.Time - startRecord.Time)
 func computeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quoteAsset string) (sdk.Dec, error) {
 	// see if we need to return an error, due to spot price issues
 	var err error = nil
-	if endRecord.LastErrorTime.After(startRecord.Time) || endRecord.LastErrorTime.Equal(startRecord.Time) {
+	if endRecord.LastErrorTime.After(startRecord.Time) ||
+		endRecord.LastErrorTime.Equal(startRecord.Time) ||
+		startRecord.LastErrorTime.Equal(startRecord.Time) {
```
mattverse, line 242: curious, do we not need to add `startRecord.LastErrorTime.Equal(startRecord.Time)`?

mattverse (commented):
All the core logic for this LGTM, nice catch btw!

Left some minor comments for this PR, please take a look

ValarDragon (approved):
This LGTM!

Apologies for delay in review, the dragonberry security bug got in the way :sweat_smile:

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#2803](https://github.com/osmosis-labs/osmosis/pull/2803) Fix total pool liquidity CLI query.
 * [#2914](https://github.com/osmosis-labs/osmosis/pull/2914) Remove out of gas panics from node logs
 * [#2937](https://github.com/osmosis-labs/osmosis/pull/2937) End block ordering - staking after gov and module sorting.
+* [#2923](https://github.com/osmosis-labs/osmosis/pull/2923) Last error time is updated during record interpolation if record has an error
```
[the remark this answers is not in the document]
migueldingli1997:
```suggestion
* [#2923](https://github.com/osmosis-labs/osmosis/pull/2923) TWAP calculation now errors if it uses records that have errored previously.
```

x/twap/logic.go
migueldingli1997: Consider the case where:
- `endRecord = {Time: 2, LastErrTime: 0}`
- `startRecord = {Time: 1, LastErrTime: 1}`

The conditions will evaluate to the following:
- `endRecord.LastErrorTime.After(startRecord.Time)` - False
- `endRecord.LastErrorTime.Equal(startRecord.Time)` - False
- `startRecord.LastErrorTime.Equal(startRecord.Time)` - True

Without the third condition, the calculation proceeds. And we do not want it to proceed because the startRecord has an error at its time.

The case is definitely possible because the startRecord might have been interpolated before being passed to this function (e.g. from `{Time: 0, LastErrTime: 0}` to `{Time: 1, LastErrTime: 1}`) while the endRecord is untouched.

x/twap/api_test.go
```diff
@@ -702,6 +718,13 @@ func (s *TestSuite) TestGetArithmeticTwapToNow() {
 			input:         makeSimpleTwapToNowInput(baseTime.Add(time.Hour), baseQuoteAB),
 			expectedError: types.StartTimeAfterEndTimeError{StartTime: baseTime.Add(time.Hour), EndTime: tPlusOne},
 		},
+		"spot price error in record at record time (start time > record time)": {
```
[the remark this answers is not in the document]
migueldingli1997: They are in different test functions though, one is `TestGetArithmeticTwap` and the other is `TestGetArithmeticTwapToNow`. Should they still be unique?

x/twap/logic.go
```diff
@@ -205,6 +206,10 @@ func (k Keeper) getInterpolatedRecord(ctx sdk.Context, poolId uint64, t time.Tim
 	if err != nil {
 		return types.TwapRecord{}, err
 	}
+        // if it had errored on the last record, make this record inherit the error
```
migueldingli1997, line 209: Fix lint issue
```suggestion
	// if it had errored on the last record, make this record inherit the error
```

alexanderbez (approved):

x/twap/api_test.go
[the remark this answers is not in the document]
mattverse: no, I think you have a point, thanks for the explanation

PR #2931, opened 2022-10-02
fix: skip unmarshaling of blank modules

czarcas7ic:

Now that we have introduced partial state exports, we have instances where we want to derive account balances for specific modules. For instance, if we just export the staking module, we dont want to error at deriving account bank balances because we expect this to be blank. This simply adds a check to see if the modules is blank, and if it is we do not require unmarshaling.

## Brief Changelog

- Checks length of module when exporting balances. If zero, don't force unmarshaling and go to next module.

## Testing and Verifying


mattverse (approved):
LGTM

ValarDragon (approved):

Issue #2940, opened 2022-10-04
State Export Logic Audit


czarcas7ic:
## Background

If it ever came down to needing a new genesis file for mainnet, we need to be certain that the current state export logic created a genesis file that replicated mainnet 1:1. As it stands, we have had to make many changes over the months after realizing things were left out or were incorrect in osmo-test-4 due to improper state export logic.

## Suggested Design

Go through every module and ensure state export logic is sound.

We can even make a new testnet (osmo-test-5) to be the new main testnet, and we can query various things on this testnet and see if there is anything off in terms of what should have been exported and what really was exported.

## Acceptance Criteria

Every module has been checked against a running testnet to verify all state export logic is sound.

ValarDragon:
Unneeded due to in place testnet

PR #2944, opened 2022-10-04
fix: remove interchain accounts from state export in testnetify.py

niccoloraspa:
Related to: https://github.com/osmosis-labs/osmosis/issues/2940

This PR implements the quick fix suggested by @ValarDragon to circumvent the issue related to re-starting a chain from a mainnet state export:

```bash
10:49PM INF ABCI Replay Blocks appHeight=0 module=consensus stateHeight=0 storeHeight=0
10:50PM INF initializing blockchain state from genesis.json
10:50PM INF running initialization for module module=capability
10:50PM INF running initialization for module module=auth
panic: unable to resolve type URL /ibc.applications.interchain_accounts.v1.InterchainAccount
```

The fix removes from the `state_export.json` all accounts of type `/ibc.applications.interchain_accounts.v1.InterchainAccount` present in the `auth` module.

## Brief Changelog

- Fix`testnetify.py`

## Testing and Verifying

Run state-exported localosmosis as usual/

### DIFF
--- a/tests/localosmosis/state_export/scripts/testnetify.py
+++ b/tests/localosmosis/state_export/scripts/testnetify.py
@@ -369,6 +369,11 @@ def main():
                 print("\tUpdate total uosmo supply from {} to {}".format(supply["amount"], str(int(supply["amount"]) + 2000000000000000 - DISTRIBUTION_MODULE_OFFSET)))
             supply["amount"] = str(int(supply["amount"]) + 2000000000000000 - DISTRIBUTION_MODULE_OFFSET)
             break
+
+    # State export bug
+    INTERCHAIN_ACCOUNT_TYPE = "/ibc.applications.interchain_accounts.v1.InterchainAccount"
+    genesis['app_state']['auth']["accounts"] = [account for account in genesis['app_state']['auth']["accounts"] if account['@type'] != INTERCHAIN_ACCOUNT_TYPE]
+
     
     print("📝 Writing {}... (it may take a while)".format(args.output_genesis))
     with open(args.output_genesis, 'w') as f:


czarcas7ic:
Can someone explain to me why this is something we cannot support in a state exported network? In the event of a chain reboot, does this imply all ICA would be removed?

ValarDragon (approved):

ValarDragon:
This is a bug that I dont understand, that we need to sort out, this is just a temporary workaround

(Would like us to be extra sure that were on v12 binaries for import though)

czarcas7ic:
Lets hold on merging this until I double confirm that this is the case :)

czarcas7ic:
This is a common mistake that comes with using cosmovisor that is my only gripe. When cosmovisor upgrades itself, it doesn't also upgrade osmosisd. So if you do something like osmosisd export after an upgrade it still uses the previous binary

czarcas7ic:
@niccoloraspa if you try this again with the newest state export I posted on our DO space then it should work without error! Please lmk and then we can close this PR!

ValarDragon:
Ah amazing, this problem did sound sus.

Thank you for investigating :)

Maybe we can have cosmovisor emit a big warning if you have a non-same version binary at `osmosisd` ?

czarcas7ic:
Even better would be for cosmovisor to simply update osmosisd as well at time of upgrade block (unsure how feasible this would be). The warning would be nice but I can still see people app hashing themselves with the following:

1. Cosmovisor upgrades, node runs for a month.
2. Some routine maintenance causes them to shut their daemon down
3. Post maintenance, muscle memory has them type osmosisd start instead of cosmovisor. Gets app hashed. Is sad while they download snapshot

niccoloraspa:
Can confirm that the fix worked. Closing this issue.
Thanks @czarcas7ic

Issue #2485, opened 2022-08-22
Small changes/chores for IBC rate limiting contract


nicolaslara:
This issue tracks the changes/chores left for https://github.com/osmosis-labs/osmosis/pull/2408.

 - [x] Add a readme
 - [ ] Calculate how the rate limit would behave in the two different scenarios from [these comments](https://github.com/osmosis-labs/osmosis/pull/2408#discussion_r949369768) (base limit on current value vs value at the beginning of the period) and adapt the code for the best option 
 - [x] Use Uint256 internally for amounts ([comment](https://github.com/osmosis-labs/osmosis/pull/2408#discussion_r947722594))
 - [x] Upgrade to cosmwasm 1.1
 - [x] Add a schema

nicolaslara:
Should probably also upgrade to cosmwasm 1.1 so that we get nice Uint256 math

ValarDragon:
Closing this tracking issue!

PR #2950, opened 2022-10-05
IBC rate limiting: Updated the contract to cosmwasm 1.1 and Uint256 for amounts

nicolaslara:
Small contract upgrades. Ref https://github.com/osmosis-labs/osmosis/issues/2485


nicolaslara:
I'm merging this into https://github.com/osmosis-labs/osmosis/pull/2410 since it's needed for the rate limiting changes around channel value calculations

ValarDragon (approved):

Issue #2060, opened 2022-07-14
Helper functions for handling sdk.Coins in osmosutil


catShaark:
## Background

We need functions to do division and multiplication of sdk.Coins and a number. For example 
DivCoin( 88stake, 2 ) -> 44stake, this will be a useful helper function for our tests

## Suggested Design

- Create function for division and multiplication of sdk.Coin in osmosutil.

ThanhNhann:
Hey @ValarDragon, I saw you coded the division of sdk.Coins, what about multiplication? if not, I will handle it

PR #2961, opened 2022-10-07
#2060 update multiple coin with uint64

ThanhNhann:
Closes: #2060 

Add multiplication of sdk.Coins with a number


osmomath/mul_coins_with_number_test.go
```diff
@@ -0,0 +1,36 @@
...
+		"5 * 1": {
+			sdk.NewInt(5), 1, sdk.NewInt(5),
+		},
+		"5 * 10": {
+			sdk.NewInt(5), 10, sdk.NewInt(50),
+		},
+	}
+
+	for name, tt := range tests {
+		fmt.Println("Start")
```
mattverse, line 30: TODO : erase this

mattverse (commented):
Thanks for this PR! 

Would this PR be able to address some of the places where we can replace existing code with this method? I'm concerned about these methods not getting used and ending up as extra boiler plates

osmomath/mul_coins_with_number_test.go
ThanhNhann: thanks for your reply. This code can be used for testing the cases relevant to the coin first sir

mattverse (commented):
oh, apologies if I was a little unclear. 

I think this implementation itself lgtm, but I was wondering if we could actually replace some existing parts or test cases in the codebase with this new method.( don't want this method to be extra overhead that remains unused)

ThanhNhann:
@mattverse maybe not now, we can close it now and reopen it when need

Issue #2622, opened 2022-09-06
Update TWAP responses, to make errors for incorrect error in time range, exposed in response


ValarDragon:
## Background

We want the marker for 'potentially unsafe result'  to be exposed in the query response. A go API error won't get threaded through to cosmwasm, so this data has to be in the query response field.

## Suggested Design

Add a bool field for "is response unstable" to the twap queries, at the GRPC proto layer, and a test for this threading through correctly

hieuvubk:
@ValarDragon could pls add more details? Like the return condition of variable

PR #2963, opened 2022-10-07
Update twap response

hieuvubk:
Closes: #2622

> Add a bool field for "is response unstable" to the twap queries

## Brief Changelog

*(for example:)*
 
  - *Update proto*
  - *Check the err in the queries to return the appropriate response*
  - *Add tests*


mattverse (commented):
Oh i see the `is_response_unstable` fields is for when the twap query fails.

Don't know how others feel about this, but my suggestion would be changing the field name from `is_response_unstable` to `success`

hieuvubk:
> Oh i see the `is_response_unstable` fields is for when the twap query fails.
> 
> Don't know how others feel about this, but my suggestion would be changing the field name from `is_response_unstable` to `success`

need to change some logic but yess

mattverse:
@hieuvubk  Can you look into proto gen ( if it's correctly updated) before we merge this? (Upon seeing check proto CI failing)

hieuvubk:
> @hieuvubk Can you look into proto gen ( if it's correctly updated) before we merge this? (Upon seeing check proto CI failing)

I already ran both `make proto-all` & `make proto-gen` but it still happen. Seem we got this CI err before.

mattverse:
@hieuvubk interesting, can you try merging with current main on ur branch and then running it?

hieuvubk:
> @hieuvubk interesting, can you try merging with current main on ur branch and then running it?

Still got the `check-proto` err

faddat:
trying a quick `make proto-all` here

faddat:
hey @hieuvubk -- I suspect that you may have found a bug with this one. 

Like you, when I try it, I end up in the same spot that you did.

Issue #2959, opened 2022-10-06
Change expectPass with bool fields in test cases to `expectedError` fields with error types


mattverse:
## Background

Most of the current test cases have an `expectPass` or an `expectFail`, `expectError` with a boolean value set as its field. (ex. https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/msg_server_test.go). 

Although these test cases are still valid, these test cases sometimes can turn out to be inaccurate, as the test case itself can omit different errors from different parts of the code from different reasons, not exactly from the logic we're trying to test, and not exactly from the reason we're expecting it to error. 

## Suggested Design

Although we do have **alot** of these test cases with the test cases of having these expectPass boolean fields as test cases, we should move towards chagning these to `expectedError` with the type of `Error` and precisely check the Error that's happening to ensure that we're erroring for reasons that we're expecting. 

A good example to reference is https://github.com/osmosis-labs/osmosis/blob/main/x/mint/keeper/keeper_test.go#L136

For tracking:
- [x] [gamm](https://github.com/osmosis-labs/osmosis/pull/2968)
- [ ] [epoch]()
- [x] [incentives](https://github.com/osmosis-labs/osmosis/pull/2986)
- [x] [lockup](https://github.com/osmosis-labs/osmosis/pull/2981)
- [ ] [mint]()
- [x] [pool-incentives](https://github.com/osmosis-labs/osmosis/pull/2985)
- [x] [superfluid](https://github.com/osmosis-labs/osmosis/pull/2990)
- [ ] [tokenfactory]()
- [x] [twap](https://github.com/osmosis-labs/osmosis/pull/2987)
- [x] [txfees](https://github.com/osmosis-labs/osmosis/pull/2989)

hieuvubk:
I think we should break this down by each module to make it easier to follow

hieuvubk:
And is anyone working on? I can start with it

mattverse:
@hieuvubk please go ahead! That'd be great! Might be almost impossible to do this in a single PR since the scope is too big, might want to do it by module to module scope.

ValarDragon:
Wait I thought we decided not to do this

czarcas7ic:
Correct, I am going to close these respective PRs and we can re address at another time.

czarcas7ic:
Respective PRs closed, closing this issue. Still think we should discuss internally to get a shared understanding of what is the best endstate for tests is

PR #2968, opened 2022-10-08
gamm: Use expectedErr fields with error types

hieuvubk:
Partially closes: #2959

> Update tests that are using expectPass or expectErr as the bool variable to the error type in the gamm module

## Brief Changelog

*(for example:)*
 
  - *Coverage gamm module*
  - *All tests pass*
  - *Commented at test cases where the error returned is not as expected*

### DIFF
--- a/x/gamm/keeper/msg_server_test.go
+++ b/x/gamm/keeper/msg_server_test.go
@@ -252,9 +252,9 @@ func (suite *KeeperTestSuite) TestJoinPool_Events() {
 			poolId:         1,
 			shareOutAmount: sdk.NewInt(shareOut),
 			tokenInMaxs:    sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(tokenInMaxAmount))),
-			expectError:    sdkerrors.Wrapf(types.ErrLimitMaxAmount, "TokenInMaxs is less than the needed LP liquidity to this JoinPoolNoSwap,"+
-			" upperbound: %v, needed %v", sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(tokenInMaxAmount))), 
-			sdk.NewCoins(sdk.NewCoin("bar", sdk.OneInt()), sdk.NewCoin("baz", sdk.OneInt()), sdk.NewCoin("foo", sdk.OneInt()), sdk.NewCoin("uosmo", sdk.OneInt()))),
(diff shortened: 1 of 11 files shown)


x/gamm/keeper/msg_server_test.go
```diff
@@ -305,7 +316,7 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
 			poolId:        1,
 			shareInAmount: sdk.NewInt(shareIn),
 			tokenOutMins:  sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(tokenOutMinAmount))),
-			expectError:   true,
+			expectError:   types.ErrLimitMinAmount,
```
mattverse, line 319: Isn't this error returned wrapped with the sdk error? e.g `sdkerrors.Wrapf(types.ErrLimitMaxAmount, sharesLargerThanMaxErrFormat, sharesIn.Int64(), shareInMaxAmount.Uint64())` ?

x/gamm/keeper/pool_service_test.go
```diff
@@ -419,7 +418,7 @@ func (suite *KeeperTestSuite) TestJoinPoolNoSwap() {
 			tokenInMaxs: sdk.Coins{
```
mattverse, line 418: Ah this test case was wrong, glad that we have this PR + change. This shouldn't be testing with negative shares, it should have positivie shares and we're expecting error related to insufficient funds, not negative share ratio. Would you be able to change this test case within this PR as well? We can also do it in a separate PR if you would prefer that way!

mattverse (commented):
Looks nice! Left some few comments that I think should be addressed before merging!

x/gamm/keeper/pool_service_test.go
hieuvubk: I updated in this pr

x/gamm/keeper/msg_server_test.go
hieuvubk: ah missed, updated

czarcas7ic:
Copying my previous message here:

There seems to be some miscommunication internally on desired endstate for tests. While you completed what was requested from the github issue, we had internally decided to stick with bool error checking (for now) and possibly discuss moving to this at a later time. cc @mattverse

czarcas7ic:

Issue #2956, opened 2022-10-06
[x/gamm]: Add RPC query endpoints for joining and exiting pools


georgemc98:
## Background

It would be helpful to have a RPC query that returns the lp shares when providing liquidity to a pool.  It would also be helpful to have a query that returns the assets returned withdrawing liquidity from a liquidity pool.  These queries will be useful when using  a `StargateQuery` in `cosmwasm` smart contracts.

## Suggested Design

Here is the commit for the PoolType query that was added recently: https://github.com/osmosis-labs/osmosis/commit/e55e13d709628cff2075ae1576f993ed2a8c310e.  The design should follow the steps taken in this commit to add RPC end points for the four following queries:

- `QueryJoinPool`
     - Parameters: QueryJoinPoolSharesRequest{ PoolId, TokenInMaxs}
     - Returns:  QueryJoinPoolSharesResponse { ShareOutAmount, TokenIn }
- `QueryJoinSwapExactAmountIn`
     - Parameters: QueryJoinSwapExactAmountInRequest{ PoolId, TokenIn}
     - Returns:  QueryJoinSwapExactAmountInResponse { ShareOutAmount }
- `QueryExitPool`
     - Parameters: QueryExitPoolSharesRequest{ PoolId, ShareInAmount }
     - Returns:  QueryExitPoolSharesResponse { TokenOut }
- `QueryExitSwapShareAmountIn`
     - Parameters: QueryExitSwapShareAmountInRequest{ PoolId, TokenOutDenom, ShareInAmount }
     - Returns:  QueryExitSwapShareAmountInResponse { TokenOutAmount }

RusAkh:
I would like to work on this issue!

georgemc98:
> I would like to work on this issue!


Excellent.  I am going to start on it today.

PR #2972, opened 2022-10-10
feat(gamm): CalcJoinPoolShares and CalcExitPoolCoinsFromShares queries

RusAkh:
Closes: #2956

Adds two queries: 

1 JoinSwapExactAmountIn allows you to query the amount of shares you get by providing X tokens
2 ExitSwapShareAmountIn allows you to query the amount of tokens you get when exiting pool with X shares

## Testing and Verifying

added two tests

  - How is the feature or change documented? changelog


RusAkh:
@georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?

georgemc98:
> @georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?

This is more difficult than the other queries.  The query estimation will be inaccurate since the liquidity is never taken out before the other token in the pool is swapped for the request token denom.  This is what we have so far: https://github.com/apollodao/osmosis/tree/dev/pool-shares-rpc-queries

georgemc98:
> @georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?

Could I push my `QueryExitPool` message into your branch?

RusAkh:
> > @georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?
> 
> Could I push my `QueryExitPool` message into your branch?

Sure

RusAkh:
> > @georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?
> 
> This is more difficult than the other queries. The query estimation will be inaccurate since the liquidity is never taken out before the other token in the pool is swapped for the request token denom. This is what we have so far: https://github.com/apollodao/osmosis/tree/dev/pool-shares-rpc-queries
> 
> I have a query for `QueryExitPool` and I'm debating if we even need the `QueryExitSwapShareAmountIn` now.

So what do you think, we leave it or remove it for good?

georgemc98:
> > > @georgemc98 I want to clarify one thing about the second query. Currently it works by equally splitting shares among all tokens in pool and then returns respectful amount of tokens of specific denom. My question is: should it work like this or should it transform all shares just into one specific denom?
> > 
> > 
> > This is more difficult than the other queries. The query estimation will be inaccurate since the liquidity is never taken out before the other token in the pool is swapped for the request token denom. This is what we have so far: https://github.com/apollodao/osmosis/tree/dev/pool-shares-rpc-queries
> > I have a query for `QueryExitPool` and I'm debating if we even need the `QueryExitSwapShareAmountIn` now.
> 
> So what do you think, we leave it or remove it for good?

I recommend that we add two endpoints, JoinPool and ExitPool.  JoinPool can take a single asset, or both assets.  Check out my branch where I implement `ExitPool` and `JoinPool`.  Your `JoinSwapExactAmountIn` query is identical to my `JoinPool` query.  I think it would be better to name it `JoinPool` for simplicity's sake.    I cannot push to your branch, so could you allow me access please?

RusAkh:
@georgemc98 invited you as a collaborator, please accept

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
 	}, err
 }
 
+// JoinSwapExactAmountIn queries the amount of shares you get by providing specific amount of tokens
+func (q Querier) JoinSwapExactAmountIn(ctx context.Context, req *types.QueryJoinSwapExactAmountInRequest) (*types.QueryJoinSwapExactAmountInResponse, error) {
```
p0mvn, line 144: Can we call this `CalcJoinPoolShares`?

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	numShares, _, err := pool.CalcJoinPoolShares(sdkCtx, req.TokensIn, pool.GetSwapFee(sdkCtx))
+	if err != nil {
+		return nil, err
+	}
+	return &types.QueryJoinSwapExactAmountInResponse{
+		ShareOutAmount: numShares,
+	}, nil
+}
+
+func (q Querier) ExitSwapShareAmountIn(ctx context.Context, req *types.QueryExitSwapShareAmountInRequest) (*types.QueryExitSwapShareAmountInResponse, error) {
```
p0mvn, line 167: godoc please

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	}
+
+	denom_liquidity := q.Keeper.GetDenomLiquidity(sdkCtx, req.TokenOutDenom)
+	if denom_liquidity == sdk.NewInt(0) {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("no denom (%s) in pool", req.TokenOutDenom))
+	}
+
+	totalSharesAmount := pool.GetTotalShares()
+	if req.ShareInAmount.GTE(totalSharesAmount) || req.ShareInAmount.LTE(sdk.ZeroInt()) {
+		return &types.QueryExitSwapShareAmountInResponse{TokenOutAmount: 0}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "share ratio is zero or negative")
```
p0mvn, line 185: please return nil

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	}
+	exitFee := pool.GetExitFee(sdkCtx)
+
+	exitCoins, err := pool.CalcExitPoolCoinsFromShares(sdkCtx, *req.ShareInAmount, exitFee)
+	if err != nil {
+		return &types.QueryExitSwapShareAmountInResponse{TokenOutAmount: 0}, err
```
p0mvn, line 191: ditto

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+
+func (q Querier) ExitSwapShareAmountIn(ctx context.Context, req *types.QueryExitSwapShareAmountInRequest) (*types.QueryExitSwapShareAmountInResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	pool, err := q.Keeper.GetPoolAndPoke(sdkCtx, req.PoolId)
+	if err != nil {
+		return &types.QueryExitSwapShareAmountInResponse{TokenOutAmount: 0}, err
```
p0mvn, line 175: nil please

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,65 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	}
+
+	for _, coin := range exitCoins {
+		if coin.Denom == req.TokenOutDenom {
+			return &types.QueryExitSwapShareAmountInResponse{TokenOutAmount: coin.Amount.Uint64()}, nil
+		}
+	}
+	return &types.QueryExitSwapShareAmountInResponse{}, nil
```
p0mvn, line 199: Should we return error if denom is not found?

x/gamm/keeper/grpc_query_test.go
```diff
@@ -9,6 +9,32 @@ import (
...
+
+	poolId := suite.PrepareBalancerPool()
+	defaultShares := sdk.NewInt(50000000000000)
+
+	out, err := queryClient.ExitSwapShareAmountIn(gocontext.Background(), &types.QueryExitSwapShareAmountInRequest{
+		PoolId:        poolId,
+		TokenOutDenom: "foo",
+		ShareInAmount: &defaultShares,
+	})
+	suite.Require().NotNil(out)
```
p0mvn, line 23: please add table driven tests to cover all edge cases

x/gamm/keeper/grpc_query_test.go
```diff
@@ -9,6 +9,32 @@ import (
...
+	suite.Require().NoError(err)
+}
+func (suite *KeeperTestSuite) TestJoinSwapExactAmountIn() {
```
p0mvn, line 26:
```suggestion

func (suite *KeeperTestSuite) TestJoinSwapExactAmountIn() {
```

x/gamm/keeper/grpc_query_test.go
```diff
@@ -9,6 +9,32 @@ import (
...
+	queryClient := suite.queryClient
+
+	poolId := suite.PrepareBalancerPool()
+
+	out, err := queryClient.JoinSwapExactAmountIn(gocontext.Background(), &types.QueryJoinSwapExactAmountInRequest{
+		PoolId:   poolId,
+		TokensIn: sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(500))),
+	})
+	suite.Require().NotNil(out)
+	suite.Require().NoError(err)
```
p0mvn, line 36: Table driven tests with edge cases please

p0mvn:
The discussed approach of having fewer queries that take more inputs makes sense. I would like to suggest considering naming queries w/o "Join/Exit" but with "Calc". We've been trying to use the former abstraction to imply mutations. Since queries are non-mutative, "Calc" abstraction is preferable

RusAkh:
@georgemc98 I fixed some things right now but not finished. basically what is left is to finish with tests. If you decide to work on this issue please feel free to push to my branch (recall that I added you to collaborators) thank you!

georgemc98:
> 

I will add table driven tests for `TestCalcJoinPoolShares` today.

georgemc98:
> @georgemc98 I fixed some things right now but not finished. basically what is left is to finish with tests. If you decide to work on this issue please feel free to push to my branch (recall that I added you to collaborators) thank you!

I will add table driven tests for `TestCalcJoinPoolShares` today.

RusAkh:
> > @georgemc98 I fixed some things right now but not finished. basically what is left is to finish with tests. If you decide to work on this issue please feel free to push to my branch (recall that I added you to collaborators) thank you!
> 
> I will add table driven tests for `TestCalcJoinPoolShares` today.

Thank you very much! I will finish TestCalcExitPoolCoinsFromShares as soon as I can

georgemc98:
> > > @georgemc98 I fixed some things right now but not finished. basically what is left is to finish with tests. If you decide to work on this issue please feel free to push to my branch (recall that I added you to collaborators) thank you!
> > 
> > 
> > I will add table driven tests for `TestCalcJoinPoolShares` today.
> 
> Thank you very much! I will finish TestCalcExitPoolCoinsFromShares as soon as I can

Sounds good, I pushed my `CalcExitPoolCoinsFromShares` test earlier

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,67 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	pool, err := q.Keeper.GetPoolAndPoke(sdkCtx, req.PoolId)
+	if err != nil {
+		return nil, types.ErrPoolNotFound
+	}
+	denom_liquidity := q.Keeper.GetDenomLiquidity(sdkCtx, req.TokenOutDenom)
```
p0mvn, line 180: let's use camelCase please

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,67 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	if denom_liquidity == sdk.NewInt(0) {
```
p0mvn, line 181:
```suggestion
	if denom_liquidity == sdk.ZeroInt() {
```

x/gamm/keeper/grpc_query_test.go
```diff
@@ -4,11 +4,151 @@ import (
...
+		},
+		{
+			"pool id does not exist",
+			poolId + 1,
+			"foo",
+			sdk.NewInt(1000000000000000000),
+			types.ErrPoolNotFound,
+		},
+		{
+			"token in denom does not exist",
```
p0mvn, line 41: consider adding a test case with an invalid `ShareInAmount`

x/gamm/keeper/grpc_query_test.go
```diff
@@ -4,11 +4,151 @@ import (
...
+		name          string
+		poolId        uint64
+		tokenOut      string
+		shareInAmount sdk.Int
+		expectedErr   error
+	}{
+		{
+			"valid test case",
+			poolId,
+			"foo",
```
p0mvn, line 29: please aim to use `sdk.DefaultBondDenom` or a reusable consant instead of hardcoding the denom name. The reason is it is more difficult to maintain if we keep hard coding more things

x/gamm/types/errors.go
```diff
@@ -16,18 +16,19 @@ func (e PoolDoesNotExistError) Error() string {
...
+	ErrTooFewPoolAssets     = sdkerrors.Register(ModuleName, 4, "pool should have at least 2 assets, as they must be swapping between at least two assets")
+	ErrTooManyPoolAssets    = sdkerrors.Register(ModuleName, 5, "pool has too many assets (currently capped at 8 assets per balancer pool and 2 per stableswap)")
+	ErrLimitMaxAmount       = sdkerrors.Register(ModuleName, 6, "calculated amount is larger than max amount")
+	ErrLimitMinAmount       = sdkerrors.Register(ModuleName, 7, "calculated amount is lesser than min amount")
+	ErrInvalidMathApprox    = sdkerrors.Register(ModuleName, 8, "invalid calculated result")
+	ErrAlreadyInvalidPool   = sdkerrors.Register(ModuleName, 9, "destruction on already invalid pool")
+	ErrInvalidPool          = sdkerrors.Register(ModuleName, 10, "attempting to create an invalid pool")
+	ErrDenomNotFoundInPool  = sdkerrors.Register(ModuleName, 11, "denom does not exist in pool")
+	ErrDenomAlreadyInPool   = sdkerrors.Register(ModuleName, 12, "denom already exists in the pool")
+	ErrInvalidNumberOfCoins = sdkerrors.Register(ModuleName, 13, "balancer pool only supports LP'ing with one asset or all assets in pool")
```
p0mvn, line 31: I suggest to consider defining a Go native error instead of SDK error. We are trying to move away from using SDK errors.

p0mvn (commented):
Great work. Can we please also rename the PR to account for the API change?

x/gamm/keeper/grpc_query_test.go
RusAkh: unfortunately suite.prepareBalancerPool does not include stake denom, this is the reason I did not use it. If you agree, I can change this method to creatte a pool with stake denom

x/gamm/types/errors.go
georgemc98: The error is originally a Go native error, but I changed it to match the other tests in the table.  The other tests match SDK errors, and line 161, https://github.com/RusAkh/osmosis-fork/blob/issue2956/x/gamm/keeper/grpc_query_test.go#L160, requires the error to be registered.  The last test in the join pool test table would return the error: `Target error should be in err chain` when returning a Go error: https://github.com/RusAkh/osmosis-fork/blob/issue2956/x/gamm/keeper/grpc_query_test.go#L135.

x/gamm/keeper/grpc_query_test.go
p0mvn: Makes sense. Let's not block this then

x/gamm/types/errors.go
p0mvn: Sounds good, thanks for clarifying. We will refactor that separately when deprecating skderrors

p0mvn (approved):
LGTM. I would like one more reviewer to approve this before merging please

x/gamm/types/errors.go
georgemc98: > LGTM. I would like one more reviewer to approve this before merging please

Awesome, thank you.

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -110,6 +121,39 @@ message QueryPoolTypeResponse {
...
+    (gogoproto.moretags) = "yaml:\"share_out_amount\"",
+    (gogoproto.nullable) = false
+  ];
+  repeated cosmos.base.v1beta1.Coin tokens_out = 2 [
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (gogoproto.nullable) = false
+  ];
+}
+
+//=============================== QueryCalcExitPoolCoinsFromSharesRequest
```
alexanderbez, line 144: nit: these are not necessary as they're superfluous IMO

alexanderbez (approved):

apollo-pacman:
@p0mvn @alexanderbez what is the process to determine whether these queries can be added to the stargate whitelist or not?

alexanderbez:
> @p0mvn @alexanderbez what is the process to determine whether these queries can be added to the stargate whitelist or not?

AFAIK, they have to be deterministic and constant in gas usage.

georgemc98:
> > @p0mvn @alexanderbez what is the process to determine whether these queries can be added to the stargate whitelist or not?
> 
> 
> 
> AFAIK, they have to be deterministic and constant in gas usage.

Is there anything I should do on my end to initiate these queries  being added to the StargateQuery whitelist?

apollo-pacman:
> > > @p0mvn @alexanderbez what is the process to determine whether these queries can be added to the stargate whitelist or not?
> > 
> > 
> > AFAIK, they have to be deterministic and constant in gas usage.
> 
> @alexanderbez Is there anything I should do on my end to initiate these queries being added to the StargateQuery whitelist?

@p0mvn @alexanderbez any update on this?

georgemc98:
@alexanderbez @RusAkh I'm seeing there is a merge conflict now in the `x/gamm/types/errors.go` file.  Would you like me to resolve it, or will the Osmosis team fix it?

alexanderbez:
Can you resolve the conflicts?

georgemc98:
> Can you resolve the conflicts?

Yes, happy to.

apollo-sturdy:
@alexanderbez @p0mvn Should whitelisting of these queries be done in this PR or do we need to create a separate PR for that? Have you guys reviewed the determinism and whatever else is needed to whitelist it?

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,69 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+	exitCoins, err := pool.CalcExitPoolCoinsFromShares(sdkCtx, req.ShareInAmount, exitFee)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, coin := range exitCoins {
+		if coin.Denom == req.TokenOutDenom {
+			return &types.QueryCalcExitPoolCoinsFromSharesResponse{TokenOutAmount: coin.Amount.Uint64()}, nil
+		}
+	}
```
p0mvn, line 201: What is the behavior of multi-asset pools?

Also, what if we want to query for all tokens out and not just one?

Can we cover the above with tests?

x/gamm/keeper/grpc_query_test.go
```diff
@@ -4,11 +4,165 @@ import (
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/types/query"
 
 	"github.com/osmosis-labs/osmosis/v12/x/gamm/types"
 )
 
+func (suite *KeeperTestSuite) TestCalcExitPoolCoinsFromShares() {
+	queryClient := suite.queryClient
+	ctx := suite.Ctx
+	poolId := suite.PrepareBalancerPool()
```
p0mvn, line 16: Can we add a test with a multi-asset pool?

x/gamm/pool-models/balancer/pool.go
```diff
@@ -725,7 +725,7 @@ func (p *Pool) CalcJoinPoolShares(ctx sdk.Context, tokensIn sdk.Coins, swapFee s
 		tokensJoined = tokensIn
 		return numShares, tokensJoined, nil
 	} else if tokensIn.Len() != p.NumAssets() {
-		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("balancer pool only supports LP'ing with one asset or all assets in pool")
+		return sdk.ZeroInt(), sdk.NewCoins(), types.ErrInvalidNumberOfCoins
```
p0mvn, line 728: This actually causes a state-break. I would prefer that we keep this as a Go error please

p0mvn:
> @alexanderbez @p0mvn Should whitelisting of these queries be done in this PR or do we need to create a separate PR for that? Have you guys reviewed the determinism and whatever else is needed to whitelist it?

This looks good to me to whitelist. However, let's do a separate PR to get more eyes from the rest of the team.

x/gamm/keeper/grpc_query.go
RusAkh: working on these issues

x/gamm/keeper/grpc_query_test.go
RusAkh: PrepareBalancerPool creates a pool with 4 assets

x/gamm/keeper/grpc_query.go
RusAkh: covered @p0mvn

RusAkh:
@p0mvn changes addressed:
- now takes a []string with denoms and returns sdk.Coins

x/gamm/keeper/grpc_query_test.go
georgemc98: > > @alexanderbez @p0mvn Should whitelisting of these queries be done in this PR or do we need to create a separate PR for that? Have you guys reviewed the determinism and whatever else is needed to whitelist it?
> 
> This looks good to me to whitelist. However, let's do a separate PR to get more eyes from the rest of the team.

Excellent.  I made a new branch based off this branch and created the PR here: https://github.com/osmosis-labs/osmosis/pull/3217.  Please let me know if there is anything else i need to do to get them whitelisted.

georgemc98:
> > @alexanderbez @p0mvn Should whitelisting of these queries be done in this PR or do we need to create a separate PR for that? Have you guys reviewed the determinism and whatever else is needed to whitelist it?
> 
> This looks good to me to whitelist. However, let's do a separate PR to get more eyes from the rest of the team.

Excellent. I made a new branch based off this branch and created the PR here: https://github.com/osmosis-labs/osmosis/pull/3217. Please let me know if there is anything else i need to do to get them whitelisted.

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,72 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+		return nil, err
+	}
+
+	return &types.QueryCalcJoinPoolSharesResponse{
+		ShareOutAmount: numShares,
+		TokensOut:      newLiquidity,
+	}, nil
+}
+
+// CalcExitPoolCoinsFromShares queries the amount of tokens you get by exiting specific amouint of shares
```
p0mvn, line 169:
```suggestion
// CalcExitPoolCoinsFromShares queries the number of tokens you get by exiting a specific amount of shares.
// Only returns the coins requested by `TokensOutDenoms`. If `TokensOutDenoms` is empty, nothing is returned
// despite potentially having exiting coins.
```

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,72 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+		return nil, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "share ratio is zero or negative")
+	}
+
+	exitCoins, err := pool.CalcExitPoolCoinsFromShares(sdkCtx, req.ShareInAmount, exitFee)
+	if err != nil {
+		return nil, err
+	}
+
+	coinsOut := sdk.NewCoins()
+	for _, req_denom := range req.TokensOutDenoms {
```
p0mvn, line 194: I'm slightly confused about the purpose of `TokensOutDenoms` for filtering. Is this a requirement for this query?

Why can't we let clients filter out the denoms that they want? I think that filtering out by this input here is likely to lead to erroneous use where clients provide only one `TokensOutDenoms` while there are more existing

x/gamm/keeper/grpc_query_test.go
```diff
@@ -2,13 +2,201 @@ package keeper_test
...
+				suite.Require().NoError(err)
+
+				var pool types.PoolI
+				err = suite.App.InterfaceRegistry().UnpackAny(poolRes.Pool, &pool)
+				suite.Require().NoError(err)
+
+				exitCoins, err := pool.CalcExitPoolCoinsFromShares(ctx, tc.shareInAmount, exitFee)
+				suite.Require().NoError(err)
+
+				// For each coin in exitCoins we are looking for a match in out response
```
p0mvn, line 105:
```suggestion
				// For each coin in exitCoins we are looking for a match in our response
```

p0mvn (commented):
Looks good to me after discussing the filtering

apollo-sturdy:
@georgemc98 @p0mvn @RusAkh
I noticed that in `JoinPoolNoSwap` it first does:
```go
neededLpLiquidity, err := getMaximalNoSwapLPAmount(ctx, pool, shareOutAmount)
```
before passing this result to:
```go
sharesOut, err = pool.JoinPoolNoSwap(ctx, neededLpLiquidity, pool.GetSwapFee(ctx))
```
which internally calls `CalcJoinPoolNoSwapShares` which this new query also calls.

`getMaximalNoSwapLPAmount` seems to calculate the reverse of `CalcJoinPoolNoSwapShares`, so in theory the use case of being able to use this query to get `sharesOut` and pass this value to `JoinPoolNoSwap` should work, but have we verified that there are no rounding issues or similar that would lead to issues?

For example, let's say I have some amounts of the assets in a pool, and use this new query `QueryCalcJoinPoolSharesRequest` to receive a `share_out_amount` and a `tokens_out` (should maybe be called `tokens_used`...?).
Now I call `MsgJoinPool` with arguments `ShareOutAmount` = `share_out_amount` returned from the query and `TokenInMaxs` = `tokens_out` returned from the query, will this work without erroring due to rounding differences between `CalcJoinPoolNoSwapShares` and `getMaximalNoSwapLPAmount`?

apollo-sturdy:
> For example, let's say I have some amounts of the assets in a pool, and use this new query `QueryCalcJoinPoolSharesRequest` to receive a `share_out_amount` and a `tokens_out` (should maybe be called `tokens_used`...?). Now I call `MsgJoinPool` with arguments `ShareOutAmount` = `share_out_amount` returned from the query and `TokenInMaxs` = `tokens_out` returned from the query, will this work without erroring due to rounding differences between `CalcJoinPoolNoSwapShares` and `getMaximalNoSwapLPAmount`?

To expand on this, the reason we would want to pass the exact amounts returned from the query to `MsgJoinPool` is so that we can be sure that exactly this amount of tokens is drawn from our account, so that we can then do a single sided join with the remaining amount. Of course it would be great if this functionality was supported out of the box, but @ValarDragon recommended that we implement it at the smart contract level (https://github.com/osmosis-labs/osmosis/issues/2994#issuecomment-1289240340), so for that to work this query needs to be an exact simulation.

x/gamm/keeper/grpc_query.go
georgemc98: Agreed, let me remove that parameter and remove this for loop.

georgemc98: > I'm slightly confused about the purpose of `TokensOutDenoms` for filtering. Is this a requirement for this query?
> 
> Why can't we let clients filter out the denoms that they want? I think that filtering out by this input here is likely to lead to erroneous use where clients provide only one `TokensOutDenoms` while there are more existing

I've removed this parameter and fixed the typos that you also suggested.

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,72 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+		return nil, status.Error(codes.InvalidArgument, "no tokens in")
+	}
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	pool, err := q.Keeper.getPoolForSwap(sdkCtx, req.PoolId)
+	if err != nil {
+		return nil, err
+	}
+
+	numShares, newLiquidity, err := pool.CalcJoinPoolShares(sdkCtx, req.TokensIn, pool.GetSwapFee(sdkCtx))
```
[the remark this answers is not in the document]
apollo-sturdy: Ah I see now looking closer at `CalcJoinPoolShares`. Well in that case we would need two queries. As per the original issue https://github.com/osmosis-labs/osmosis/issues/2956 what we actually want is the ability to simulate each type of liquidity provision (`MsgJoinPool` and `MsgJoinSwapExternAmountIn`). These are the only use cases I'm aware of or can think of anyone wanting.

It looks like for `MsgJoinSwapExternAmountIn` it only really does `CalcJoinPoolShares`, so the current query that this PR introduces should work as an exact simulation of `MsgJoinSwapExternAmountIn`?

For `MsgJoinPool` as mentioned it seems to do more than just `CalcJoinPoolNoSwapShares`. Would you be able to help with introducing an exact simulation query for this function?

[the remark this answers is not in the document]
p0mvn: > For MsgJoinPool as mentioned it seems to do more than just CalcJoinPoolNoSwapShares. Would you be able to help with introducing an exact simulation query for this function?

I think we should be able to do the following:
- create a new query `CalcJoinPoolNoSwapShares` similar to the one introduced in this PR
- add [this logic](https://github.com/osmosis-labs/osmosis/blob/f5c7125f8cc7e58e0ed7912e62333758a9194870/x/gamm/keeper/pool_service.go#L79-L82) to get maximal no swap lp amounts
- call pool's [`CalcJoinPoolNoSwapShares`](https://github.com/osmosis-labs/osmosis/blob/f5c7125f8cc7e58e0ed7912e62333758a9194870/x/gamm/types/pool.go#L79)
- add table-driven unit tests
- ideally, add a functional test that makes sure that this new query behaves exactly the same as [`JoinPoolNoSwap`](https://github.com/osmosis-labs/osmosis/blob/f5c7125f8cc7e58e0ed7912e62333758a9194870/x/gamm/keeper/pool_service.go#L64)
   * I think this can be also tested contract side so is optional

@RusAkh would you be interested in making a follow-up PR with this new query?

p0mvn (approved):

x/gamm/keeper/grpc_query.go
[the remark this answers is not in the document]
RusAkh: yes, I would like to work on it

RusAkh:
hello @p0mvn. Are you sure we need to update sdk version in go.mod in this PR?

p0mvn:
> hello @p0mvn. Are you sure we need to update sdk version in go.mod in this PR?

Yes, that should be fine. We use a replace directive for cosmos-sdk so the true version is determined by that. Thanks for checking though

x/gamm/keeper/grpc_query.go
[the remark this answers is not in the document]
p0mvn: Awesome. Please feel free to make a follow-up issue and start working on it

p0mvn:
I'm going to merge this. `CalcJoinPoolNoSwapShares` can be addressed in a follow-up PR

x/gamm/keeper/grpc_query.go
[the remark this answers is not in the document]
RusAkh: @p0mvn I want to clarify something, why would we include [this logic](https://github.com/osmosis-labs/osmosis/blob/f5c7125f8cc7e58e0ed7912e62333758a9194870/x/gamm/keeper/pool_service.go#L79-L82)? I see we are simulating pool's CalcJoinPoolNoSwapShares method, right? I want to make sure I understand the task correctly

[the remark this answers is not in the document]
apollo-sturdy: Sounds great. Thank you @p0mvn and @RusAkh!

@RusAkh the reason to include that logic is that we want an exact simulation of `MsgJoinPool`. @p0mvn perhaps it would be clearer if this new query was called something else? E.g. `QueryJoinPool`, `SimulateJoinPool` or something.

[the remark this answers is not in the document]
apollo-sturdy: @RusAkh Can you tag me in the PR once you open it? Thanks!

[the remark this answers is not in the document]
p0mvn: > I want to clarify something, why would we include [this logic](https://github.com/osmosis-labs/osmosis/blob/f5c7125f8cc7e58e0ed7912e62333758a9194870/x/gamm/keeper/pool_service.go#L79-L82)? I see we are simulating pool's CalcJoinPoolNoSwapShares method, right?

> the reason to include that logic is that we want an exact simulation of MsgJoinPool

Yes, that's correct. Based on the requirements from @apollo-sturdy , the new query's API should be:
```
shareOutAmount -> tokensIn, sharesOut
```  

In other words, given  the desired shareOutAmount we want to answer:
- how many tokens do we actually need to put in without performing a swap?
- how many shares would that actually return? (there is small rounding difference from shareOutAmount)

By using `getMaximalNoSwapLPAmount` that you highlighted @RusAkh , we can answer the first question.

By calling pool's `CalcJoinPoolNoSwapShares` and giving it the result of  `getMaximalNoSwapLPAmount`, we can answer second question.

Please let me know if that makes sense.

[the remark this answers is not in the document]
RusAkh: hello @p0mvn, thanks you, it absolutely makes sense, I thought the point of this query was actually the opposite of what you just said, thanks again!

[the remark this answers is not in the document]
RusAkh: I think there is a problem in getMaximalNoSwapLPAmount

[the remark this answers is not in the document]
RusAkh:
```
totalSharesAmount := pool.GetTotalShares()
	// shareRatio is the desired number of shares, divided by the total number of
	// shares currently in the pool. It is intended to be used in scenarios where you want
	shareRatio := shareOutAmount.ToDec().QuoInt(totalSharesAmount)
	if shareRatio.LTE(sdk.ZeroDec()) || shareRatio.GT(sdk.OneDec()) {
		return sdk.Coins{}, sdkerrors.Wrapf(types.ErrInvalidMathApprox, "share ratio is zero or negative")
	}
```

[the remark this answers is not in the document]
RusAkh: It only checks if share ratio is LTE 0, I think it should also check if it is bigger than one, because in case it is >= 1, you are trying to get more shares than there is available. Wdyt @p0mvn @apollo-sturdy ?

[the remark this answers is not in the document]
apollo-sturdy: > 

Why should you not be able to provide more liquidity than is already in the pool? The amount of "shares available" is infinite, since they are minted upon liquidity provision?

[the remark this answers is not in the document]
RusAkh: > > It only checks if share ratio is LTE 0, I think it should also check if it is bigger than one, because in case it is >= 1, you are trying to get more shares than there is available. Wdyt @p0mvn @apollo-sturdy ?
> 
> Why should you not be able to provide more liquidity than is already in the pool? The amount of "shares available" is infinite, since they are minted upon liquidity provision?

indeed, thank you for clarification

x/gamm/keeper/grpc_query.go
```diff
@@ -140,6 +140,59 @@ func (q Querier) PoolType(ctx context.Context, req *types.QueryPoolTypeRequest)
...
+func (q Querier) CalcJoinPoolShares(ctx context.Context, req *types.QueryCalcJoinPoolSharesRequest) (*types.QueryCalcJoinPoolSharesResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+	if req.TokensIn == nil {
+		return nil, status.Error(codes.InvalidArgument, "no tokens in")
+	}
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	pool, err := q.Keeper.getPoolForSwap(sdkCtx, req.PoolId)
```
p0mvn, line 153: For balancer pools, this method may call `PokePool()` which updates pool weights. Is this acceptable?

CC: @ValarDragon

ValarDragon: Yeah this is fine, because:
* this won't get saved
* Even if it did, cosmwasm would revert the result

PR #2980, opened 2022-10-12
Add servers to openapi.yaml swagger file

### LINKED ISSUE #2979: Add mainnet & testnet server to openapi swagger docs
Openapi spec files supports specifying servers, it would be useful to add the mainnet/testnet servers to to our swagger docs by default in order to make a working swagger ui in our docs automatically.

daniel-farina:
This change on the codegen scripts allows adding default servers on this openapi swagger file which will be used on the official docs. 

closes #2979

### DIFF
--- a/scripts/generate-docs.sh
+++ b/scripts/generate-docs.sh
@@ -40,6 +40,10 @@ yarn install
 yarn combine
 yarn convert
 yarn build
+
+#Add public servers to spec file for Osmosis testnet and mainnet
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):

ValarDragon:
LGTM, though do we have any way to track if our endpoints are getting overloaded due to this?

mattverse (approved):
I'm not worried about the spec file being overloaded due to this, but I would love to see this in a testnet node

PR #2981, opened 2022-10-12
lockups: use expectedErr with error types

RusAkh:
Closes partially: #2959

What is the purpose of the change:
Update tests that are using expectPass or expectErr as the bool variable to the error type in the lockup module

all tests pass


x/lockup/types/errors.go
```diff
@@ -12,4 +12,9 @@ var (
 	ErrSyntheticLockupAlreadyExists      = sdkerrors.Register(ModuleName, 2, "synthetic lockup already exists for same lock and suffix")
 	ErrSyntheticDurationLongerThanNative = sdkerrors.Register(ModuleName, 3, "synthetic lockup duration should be shorter than native lockup duration")
 	ErrLockupNotFound                    = sdkerrors.Register(ModuleName, 4, "lockup not found")
+	ErrNewDurationNotGreater             = sdkerrors.Register(ModuleName, 5, "new duration should be greater than the original")
+	ErrRequestedExceedsTokens            = sdkerrors.Register(ModuleName, 6, "requested amount to unlock exceeds locked tokens")
+	ErrNotStartedUnlocking               = sdkerrors.Register(ModuleName, 7, "lock hasn't started unlocking yet")
+	ErrNotUnlockableYet                  = sdkerrors.Register(ModuleName, 8, "lock is not unlockable yet")
+	ErrSyntheticLockupNotExists          = sdkerrors.Register(ModuleName, 9, "synthetic lockup does not exist")
```
mattverse, line 19: I think these should be handled in a separate PR + needs further discussion on adding these, as adding new errors would make this PR state breaking

RusAkh: I can remove them for now but the tests would be pretty dirty without those, what do you think? @mattverse

czarcas7ic: Since this PR is not urgent, we can probably just leave these here and not backport it right?

czarcas7ic:

PR #2985, opened 2022-10-13
convert from shoudlErr to specific error pool-incentives/keeper unit tests

tantv4668:
Closes: #XXX

- Convert from shouldErr (boolean) to specific error in x/pool-incentives/keeper unit tests

### DIFF
--- a/x/pool-incentives/keeper/distr_test.go
+++ b/x/pool-incentives/keeper/distr_test.go
@@ -1,6 +1,8 @@
 package keeper_test
 
 import (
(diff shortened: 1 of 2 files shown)


czarcas7ic:
Thanks so much for making this PR! There seems to be some miscommunication internally on desired endstate for tests. While you completed exactly what was requested from the github issue, we had internally decided to stick with bool error checking (for now) and possibly discuss moving to this at a later time. cc @mattverse

PR #2986, opened 2022-10-13
incentives: expectErr with error fields not bool

RusAkh:
Closes partially: #2959 

incentives tests now have error fields in expectingErr and not bool


czarcas7ic:

Issue #2457, opened 2022-08-19
Make testing suite to ensure queries never alter state


ValarDragon:
## Background

Similar to how we made a test battery for `messages` to check things like encoding, we should make a battery of tests for queries.

One thing in the battery, should be ensuring that all queries do not alter state. (perhaps attempting to commit the block and check app hash, or iterating over every state entry)

## Acceptance Criteria

- [ ] We have a test in AppTesting to ensure that queries aren't state altering
- [ ] We make tests in client/* that are table driven where we make a proto input, run the query, and check that no state changed

vuong177:
Hi @ValarDragon could I do this work? It is related to my PR about estimate swap.

ValarDragon:
Sure!

PR #3001, opened 2022-10-14
Make testing suite to ensure queries never alter state

hieuvubk:
Closes: #2457 

## Brief Changelog

Create a func in apptesting to make sure state not change after query.
Create test in modules/client

## Testing and Verifying

Added new test
All test passed

### DIFF
--- a/x/epochs/client/cli/query_test.go
+++ b/x/epochs/client/cli/query_test.go
@@ -1,0 +1,75 @@
+package cli_test
+
+import (
+	gocontext "context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/osmosis-labs/osmosis/v12/app/apptesting"
+	"github.com/osmosis-labs/osmosis/v12/x/epochs/types"
+)
+
+type QueryTestSuite struct {
+	apptesting.KeeperTestHelper
+	queryClient types.QueryClient
+}
+
+func (s *QueryTestSuite) SetupSuite() {
+	s.Setup()
+	s.queryClient = types.NewQueryClient(s.QueryHelper)
+
+	// add new epoch
+	epoch := types.EpochInfo{
+		Identifier:              "weekly",
+		StartTime:               time.Time{},
+		Duration:                time.Hour,
+		CurrentEpoch:            0,
+		CurrentEpochStartHeight: 0,
+		CurrentEpochStartTime:   time.Time{},
+		EpochCountingStarted:    false,
+	}
+	s.App.EpochsKeeper.AddEpochInfo(s.Ctx, epoch)
+
+	s.Commit()
+}
+
+func (s *QueryTestSuite) TestQueriesNeverAlterState() {
+	s.SetupSuite()
+	testCases := []struct {
+		name  string
+		query func()
+	}{
+		{
+			"Query current epoch",
+			func() {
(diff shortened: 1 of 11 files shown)


x/epochs/client/cli/query_test.go
```diff
@@ -0,0 +1,74 @@
...
+func (s *QueryTestSuite) TestQueriesNeverAlterState() {
+	s.SetupSuite()
+	testCases := []struct {
+		name  string
+		query func()
+	}{
+		{
+			"Query current epoch",
+			func() {
+				res, err := s.queryClient.CurrentEpoch(gocontext.Background(), &types.QueryCurrentEpochRequest{Identifier: "weekly"})
```
p0mvn, line 48: nit: Let's try to avoid these callbacks in table-driven test cases if possible please. While this test case is not complex, for bigger test cases this pattern becomes confusing. I think having a completely separate test fixture is preferable to avoid encouraging this style.

p0mvn: I know it's a lot to refactor so I discussed this with Matt, and he agrees with the suggestion to split it up if possible, please

p0mvn:
@hieuvubk thanks for the change. Please apply a state-compatibility label. These start with "V:"

x/epochs/client/cli/query_test.go
hieuvubk: So should split each query into a separate test function?

hieuvubk: To avoid table-driven callbacks, I think there is another way to use queryClient.cc.Invoke and pass query path, request,...

p0mvn: I think that sounds reasonable. Do you mind changing one test to see how it compares?

p0mvn: Thanks, IMO the epoch change looks good.

hieuvubk: Thank you, I'll apply it to other modules

p0mvn (approved):
LGTM.

I have some points of concern but IMO we can address them separately:
- Difficult to maintain hard-coded proto paths (I've thought about this after we discussed the approach)
- We rely upon that export genesis being correct and exporting all states. I think that the app hash verification approach might be more accurate

hieuvubk:
> Agreed with @p0mvn assessment in terms of reliance on ExportGenesis. For now I think using exportGenesis is "good enough", but has been known to contain issues in terms of not exporting the right thing. We should both:
> 
> 1. Audit exportGenesis code for all modules and
> 2. Look into utilizing appHash instead of exportGenesis comparison
> 
> Nice work on this though! 😄

Thank you, will create an issue about it

p0mvn:
Hi @hieuvubk . Thanks for the great work here. Can you please link the newly created issue once it is ready?

I would like to merge this only after we track future work to avoid getting anything missed

hieuvubk:
> Hi @hieuvubk . Thanks for the great work here. Can you please link the newly created issue once it is ready?
> 
> I would like to merge this only after we track future work to avoid getting anything missed

Thank you, I have created 2 separate issues so that we can easily track and discuss them in the future.

p0mvn:
Awesome, thanks!

p0mvn:
discussed with @czarcas7ic and this looks good, going to continue improving on the approach in the created issues

Issue #3004, opened 2022-10-16
[x/gamm][stableswap]: Expand inverse relation tests to accommodate multi-asset operations


AlpinYukseloglu:
## Background

Follow-up issue from #2921 to expand our inverse relation tests to accommodate multi-asset pool functions.

## Suggested Design

- Same pool sizing for test cases as two-asset, except for swaps in 3-5 asset pools

## Acceptance Criteria

- All new and existing tests should pass

Issue #3005, opened 2022-10-16
[x/gamm][stableswap]: Add scaling factors to inverse relationship tests


AlpinYukseloglu:
## Background

Our inverse relationship tests for stableswap currently do not incorporate scaling factors (hard-coded to factors of 1). We should expand these tests to ensure that uneven scaling factors do not break our stableswap inverse relationship.

## Suggested Design

- Abstract scaling factors into test field and add cases

## Acceptance Criteria

- All new and existing tests pass

PR #2921, opened 2022-10-01
[x/gamm][stableswap]: Fix and expand gamm inverse relationship tests & fix bug in `CalcInAmtGivenOut`

AlpinYukseloglu:
## What is the purpose of the change

This PR fixes our inverse relationship tests which were previously not being run on both Balancer and Stableswap.

It also fixes a bug in our stableswap `CalcInAmtGivenOut`, and a separate bug in how the inverse relationship tests handles our stableswap CFMM for imbalanced pools.

As a byproduct of these changes, this PR also does basic level testing for the rounding-related changes added in #2912

TODO: Expand this functionality for the multi asset solver/calcs

## Brief Changelog

- Fix bug in inverse relationship testing suite that caused none of the tests to run
- Update the tests and testing methodology to properly accommodate base and edge cases for both stableswap and balancer
- Fix bug in `CalcInAmtGivenOut` that caused specific small inputs to be negated due to a failure to acknowledge error bounds
- As a simpler/more elegant solution to prevent the previous point from accidentally happening again, make solver support negative inputs

## Testing and Verifying

- All changes in this PR are tested in `amm_test.go` as part of the inverse relationship test suite

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (not documented)

PR #3006, opened 2022-10-16
[x/gamm][stableswap]: Expand inverse relation tests to multi assets and add scaling factors

AlpinYukseloglu:
Closes: #3004 and #3005

This PR expands our inverse relationship tests to accommodate multi-asset pools and scaling factors for stableswap.

## Brief Changelog

- Clean up and update inverse relation tests to accommodate multi pools and scaling factors
- Add test cases for balanced and imbalanced multi pools and scaling factors
- Update multi-asset solver to accommodate negative inputs (as done with our two-asset solver in #2921

## Testing and Verifying

- All new and existing functionality is tested in `stableswap/amm_test.go`

### DIFF
--- a/x/gamm/pool-models/stableswap/amm.go
+++ b/x/gamm/pool-models/stableswap/amm.go
@@ -174,7 +174,6 @@ func solveCFMMBinarySearch(constantFunction func(osmomath.BigDec, osmomath.BigDe
 		yFinal := yReserve.Add(yIn)
 		xLowEst := osmomath.ZeroDec()
 		// we set upper bound at 2 * xReserve to accommodate negative yIns
-		// TODO: update this to be in accordance with spec (linear estimation)
 		xHighEst := xReserve.Mul(osmomath.NewBigDec(2))
 		maxIterations := 256
 		errTolerance := osmoutils.ErrTolerance{AdditiveTolerance: sdk.OneInt(), MultiplicativeTolerance: sdk.Dec{}}
--- a/x/gamm/pool-models/stableswap/amm.go
+++ b/x/gamm/pool-models/stableswap/amm.go
@@ -174,6 +174,7 @@ func solveCFMMBinarySearch(constantFunction func(osmomath.BigDec, osmomath.BigDe
 		yFinal := yReserve.Add(yIn)
 		xLowEst := osmomath.ZeroDec()
 		// we set upper bound at 2 * xReserve to accommodate negative yIns
+		// TODO: update this to be in accordance with spec (linear estimation)
 		xHighEst := xReserve.Mul(osmomath.NewBigDec(2))
 		maxIterations := 256
 		errTolerance := osmoutils.ErrTolerance{AdditiveTolerance: sdk.OneInt(), MultiplicativeTolerance: sdk.Dec{}}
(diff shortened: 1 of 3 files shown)


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -174,6 +174,7 @@ func solveCFMMBinarySearch(constantFunction func(osmomath.BigDec, osmomath.BigDe
 		yFinal := yReserve.Add(yIn)
 		xLowEst := osmomath.ZeroDec()
 		// we set upper bound at 2 * xReserve to accommodate negative yIns
+		// TODO: update this to be in accordance with spec (linear estimation)
```
AlpinYukseloglu, line 177: Flagged in #3007

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -199,15 +200,30 @@ func solveCFMMBinarySearch(constantFunction func(osmomath.BigDec, osmomath.BigDe
 // solveCFMMBinarySearch searches the correct dx using binary search over constant K.
 // added for future extension
 func solveCFMMBinarySearchMulti(xReserve, yReserve, wSumSquares, yIn osmomath.BigDec) osmomath.BigDec {
-	if !xReserve.IsPositive() || !yReserve.IsPositive() || wSumSquares.IsNegative() || !yIn.IsPositive() {
+	if !xReserve.IsPositive() || !yReserve.IsPositive() || wSumSquares.IsNegative() {
 		panic("invalid input: reserves and input must be positive")
-	} else if yIn.GTE(yReserve) {
+	} else if yIn.Abs().GTE(yReserve) {
 		panic("cannot input more than pool reserves")
 	}
```
ValarDragon, line 207: We should revisit this in the future, since I don't think we need it for protection on the over-join side. (But also maybe doesn't matter)

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -221,7 +237,7 @@ func solveCFMMBinarySearchMulti(xReserve, yReserve, wSumSquares, yIn osmomath.Bi
 		panic(err)
 	}
 
-	xOut := xReserve.Sub(xEst)
+	xOut := xReserve.Sub(xEst).Abs()
```
ValarDragon, line 240: Err, are we sure this abs is correct

x/gamm/pool-models/stableswap/amm.go
ValarDragon, line 177: I believe this is now done!

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: Yes – it's there to cover the case where `yIn` is negative and `xEst > xReserve`

Alternatively, we can have an if/else that does `xOut = xReserve - xEst` or `xOut = xEst -  xReserve` but this seemed cleaner to me

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu: Good point! Changes are in #3067 now – removing this comment

x/gamm/pool-models/stableswap/amm.go
ValarDragon: I definitely think this abs should be done by the caller.

We are equivocating two different things in the function output with the abs (caller needs to give tokens into the pool, vs caller gets tokens out of the pool)

ValarDragon (approved):
LGTM once we remove the abs

ValarDragon:
Removed the approve, I tried doing this locally, and removing the abs is actually giving weird errors. It appears the output of `solveCfmm` in the call stack of `calcInAmtGivenOut` is sometims positive, and sometimes negative, which is concerning / shouldn't be correct.

ValarDragon (approved):
Re-approving, I messed up doing it from convo w/ alpin!

Issue #2906, opened 2022-09-30
Proposal: Support better query for pools


mattverse:
## Background
We currently only have a single query for querying all Pools, namely `osmosis/gamm/v1beta1/pools`. 
But with different pools existing under different circumstances, and with over 500+ pools existing in state as of right now, there are situations where a person would like to query pools with a certain filter, or a whitelist. 

I would like to propose two different ways of supporting better queries for pools. 

## Suggested Design
1. The first way of doing this can be done via creating a pool query that accepts different filter as its request, something like this 
```
message QueryPoolWithFilterRequest {
       int64 minLiquidity = 1;
       string poolType = 2;
      int64 minPoolAssetNum = 3;
      ...
}
```
and then whoever's requesting can use the field that they want to use as a filter, leaving other fields with null value.  The response would be returned via iterating over pool state, only returning the pools that goes under the filtered request.

2. The second way of doing this is a quite untraditional way of supporting queries, which is done via using `app.toml`. We create a field in `app.toml` where the node operator can have their own list of whitelist pool ids. Then we have a query, say `QueryPoolsUnsafe`, which would non-deterministically only return the state of the pools that are whitelisted in app.toml. Using the `QueryPoolsUnsafe` query, who ever's querying does not have to go through the process of having to query all states of pools first, and the filtering them within the client side. 

Open for discussions on the pros and cons of these ways, and also if we even do need these queries!

RusAkh:
Hello @mattverse! This issue seems interesting, I would like to work on it. May I? (I am more of a fan of the first suggested design)

mattverse:
Yes please!

PR #3009, opened 2022-10-16
gamm: query to get pools with specified parameters 

RusAkh:
Closes: #2906 

adds query to get pools with specified parameters

## Testing and Verifying

added unit test


RusAkh:
hello @mattverse. Currently this query supports only 4 "filters": min_liquidity, includes_denom, num_tokens, pool_type. If you think some other things might be added here I would love to hear about them. Or if there is anything in my code worth additional work

mattverse:
Hey @RusAkh Sounds awesome, I don't think it's worth supporting num_tokens for now. 
I'm also curious how the implementation of min_liquidity would work, since the chain itself does not know the actual prices, just the relatvie spot price. 

Also lmk or feel free to go ahead more options for the filters that you would like to add!

RusAkh:
@mattverse currently min_liquidity is used in a format of sdk coins such that an amount of every token in pool should be greater or equal than each sdk coin in min_liquidity request. I thought about using an actual price for liquidity in this query (for example 1 mil $) but do not really know how to implement this. What do you think should be done here?

mattverse:
> @mattverse currently min_liquidity is used in a format of sdk coins such that an amount of every token in pool should be greater or equal than each sdk coin in min_liquidity request. I thought about using an actual price for liquidity in this query (for example 1 mil $) but do not really know how to implement this. What do you think should be done here?

So I don't thimk the sdk coin amt represents liquidity. I propose we drop the min_liquidity feature here

RusAkh:
> > @mattverse currently min_liquidity is used in a format of sdk coins such that an amount of every token in pool should be greater or equal than each sdk coin in min_liquidity request. I thought about using an actual price for liquidity in this query (for example 1 mil $) but do not really know how to implement this. What do you think should be done here?
> 
> So I don't thimk the sdk coin amt represents liquidity. I propose we drop the min_liquidity feature here

Hello @mattverse. I currently spoke with other developers and we currently think only two "filters" namely includes_denom and pool_type is enough for now. Just want to make sure it is ok with you

RusAkh:
actually I did not really understand why do not we use a filter for a minimum number of tokens. I think this could actually replace includes_denom filter with something like minimum_amount_of_asset, which will basically be an sdk.coin, which will allow us to query pools with specific asset (includes_denom is basically already doing this), but also with a minimum number of this asset in a pool. If I am wrong, please, I am open to discussion :)

mattverse:
> actually I did not really understand why do not we use a filter for a minimum number of tokens. I think this could actually replace includes_denom filter with something like minimum_amount_of_asset, which will basically be an sdk.coin, which will allow us to query pools with specific asset (includes_denom is basically already doing this), but also with a minimum number of this asset in a pool. If I am wrong, please, I am open to discussion :)

Oh I see, the feature ur suggesting does not support a single filter, but multiple filters at once, 

good point, I am very supportive of this

RusAkh:
hello @mattverse done

x/gamm/keeper/grpc_query.go
```diff
@@ -170,6 +170,96 @@ func (q Querier) PoolParams(ctx context.Context, req *types.QueryPoolParamsReque
...
 
+// PoolsWithFilter query allows to query pools with specific parameters
+func (q Querier) PoolsWithFilter(ctx context.Context, req *types.QueryPoolsWithFilterRequest) (*types.QueryPoolsWithFilterResponse, error) {
+	res, err := q.Pools(ctx, &types.QueryPoolsRequest{
+		Pagination: &query.PageRequest{},
+	})
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	if err != nil {
+		panic(err)
```
mattverse, line 181: this should return error, not panic!

x/gamm/keeper/grpc_query_test.go
```diff
@@ -87,6 +89,143 @@ func (suite *KeeperTestSuite) TestQueryPools() {
...
+		}
+	)
+
+	testCases := []struct {
+		name               string
+		num_pools          int
+		expected_num_pools int
+		min_liquidity      sdk.Coins
+		pool_type          string
+		include_denom      string
```
mattverse, line 112:
```suggestion
```

unused field

x/gamm/keeper/grpc_query_test.go
```diff
@@ -87,6 +89,143 @@ func (suite *KeeperTestSuite) TestQueryPools() {
...
+		poolAssets         []balancertypes.PoolAsset
+		expectedErr        error
+	}{
+		{
+			name:               "valid tc",
```
mattverse, line 117:
```suggestion
			name:               "valid tc with both filters for min liquidity and pool type",
```

x/gamm/keeper/grpc_query_test.go
```diff
@@ -87,6 +89,143 @@ func (suite *KeeperTestSuite) TestQueryPools() {
...
+			num_pools:          2,
```
mattverse, line 118: can we add two test cases for:
- only querying using filter for min liquidity
- only querying using fitler for pool type

mattverse (commented):
Left some initial comments for this PR! Looking Good! Please take a look!

x/gamm/keeper/grpc_query.go
```diff
@@ -170,6 +170,96 @@ func (q Querier) PoolParams(ctx context.Context, req *types.QueryPoolParamsReque
...
+	}
+	pools := res.Pools
+
+	var response = []*codectypes.Any{}
+
+	// set filters
+	min_liquidity := req.MinLiquidity
+	pool_type := req.PoolType
+	checks_needed := 0
```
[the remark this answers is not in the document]
RusAkh: I do not think we can do that:
```
if min_liquidity_flag || pool_type_flag {
			response = append(response, p)
}
```
I just tried to do something like that and as you can see this logic will not work, we still need to somehow know how many checks we need.

RusAkh:
hello @mattverse done

mattverse (approved):
LGTM, @RusAkh  can you -help with resolving the final merge conflict and then we can merge

RusAkh:
hello @mattverse I think it is done

ValarDragon:
I haven't reviewed this, but I'm happy to merge as it only adds a new queries

jonator:
Can we eventually create a query that will calculate spot prices and compare them amongst pools? That would make things much simpler on frontend. Then, we'd be able to just supply one amount, say OSMO or USDC, and the query would calculate and only return pools with a minimum equivalent amount of the given asset. Then, we could just supply 100,000 USDC and the query would return pools with a TVL of at least $100,000

RusAkh:
@jonator If I understand you correctly, initially I had this idea of `minimum liquidity` filter, however it was rejected. Is it what you mean, right?

If so, I would like to work on it and bring back the filter if @mattverse approves

RusAkh:
Messages from this PR:
```
currently min_liquidity is used in a format of sdk coins such that an amount of 
every token in pool should be greater or equal than each sdk coin in min_liquidity request.
I thought about using an actual price for liquidity in this query (for example 1 mil $) 
but do not really know how to implement this. 
What do you think should be done here?
```
```
So I don't think the sdk coin amt represents liquidity. I propose we drop the min_liquidity feature here
```

I think the actual liquidity feature could be implemented on an application level using oracles and in this query we could leave `min_liquidity` as `sdk.Coins`. Even so, I guess an actual price$ won't be always necessary for a user. I think, it is also good to be able to specify the number of tokens the pool in response would need to contain in spite of their actual price

PR #3010, opened 2022-10-16
fix issue_2959 for superfluid module

tantv4668:
- Convert expectErr (bool) to specific error in superfluid module (issue 2959)


tantv4668:
duplicate task

Issue #3031, opened 2022-10-17
[CL]: Refactor and generalize pool interface


p0mvn:
## Background

There are several methods currently defined on the pool interface that are applicable to balancer and stableswap pools but not the concentrated liquidity pool.

Therefore, we should identify what methods should be moved out of the interface into a pool interface extension similar to:
https://github.com/osmosis-labs/osmosis/blob/babeebe957986ed13b6a3e7b2dc01f04da876f10/x/gamm/types/pool.go#L94

## Suggested Design

Upon initial inspection of the current pool interface methods, the following should be moved out to a separate extension:
- `GetTotalPoolLiquidity`
   * According to the concentrated liquidity pool design, we aren't going to keep track of each token individually. Instead, we will keep track of $$\sqrt P$$ and $$L$$
- `PokePool`
   * This method is concerned with weights that only seem to be applicable to the balancer pool

The remaining methods can be split into 3 categories:
- needed for LP
- needed for swapping
- generalized getters

## Acceptance Criteria
- Pool interface is refactored and generalized
- The listed above methods are moved to a separate interface extension
- Balancer pool continues to function as expected
- Stableswap pool continues to function as expected

ValarDragon:
Concentrated liquidity at most should only share a swap interface with normal AMM's. They should not share any other methods / should otherwise have distinct interfaces.

p0mvn:
> Concentrated liquidity at most should only share a swap interface with normal AMM's. They should not share any other methods / should otherwise have distinct interfaces.

Agreed, the goal here is to move out any balancer and stable swap-specific methods into an interface extension and only keep swapping/joining methods.

Do you also agree about join pool no swap and exit pool methods being applicable? These also seem to be valid for adding/removing liquidity

p0mvn:
@mattverse  has determined that the API change for joining and exiting pool would have to be significantly different due to ticks. As a result, we are exploring the idea of implementing the concentrated liquidity pool with its own API. Likely in another module (not gamm).

That being said, I removed the `PokePool` method from the pool interface because it only seems to make sense for balancer pools: https://github.com/osmosis-labs/osmosis/pull/3035

p0mvn:
After discussing and thinking about the architecture more, we should be able to extract an `AmmInterface` that is implemented by `balancer` and `stableswap` pools. It should contain the methods only compatible with the amm pools.

[...]

PR #3035, opened 2022-10-18
refactor: remove PokePool from the PoolI interface, define on extension instead

p0mvn:
Closes: #3031

This PR removes `PokePool` method from the pool interface and defines it on the `PoolI` instead.

The reason is that only balancer pool has the concept of weights. Therefore, it does not make sense to have this method on stableswap pool as well.

Instead, we can define this via extension and verify via type checking.

Additionally, I regenerated the mock pool upon changing the interface.

## Brief Changelog
- removed `PokePool` from `PoolI`
- defined on extension
- balancer pool implements extension
- do type check in keeper
- tests

## Testing and Verifying

This change added tests and can be verified as follows


x/gamm/module.go
```diff
@@ -88,7 +89,7 @@ func (b AppModuleBasic) GetQueryCmd() *cobra.Command {
 func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 	types.RegisterInterfaces(registry)
 	balancer.RegisterInterfaces(registry)
-	// stableswap.RegisterInterfaces(registry)
+	stableswap.RegisterInterfaces(registry)
```
p0mvn, line 92: Note to reviewers: registered stable swap interfaces that are needed for tests.

p0mvn: Applied state breaking label due to this

x/gamm/keeper/pool_test.go
```diff
@@ -220,3 +231,93 @@ package keeper_test
...
+
+			pool, err := k.GetPoolAndPoke(ctx, tc.poolId)
+
+			if tc.expectError {
+				suite.Require().Error(err)
+				return
+			}
+
+			suite.Require().NoError(err)
+
```
mattverse, line 299:
```suggestion
```

x/gamm/keeper/pool_test.go
```diff
@@ -220,3 +231,93 @@ package keeper_test
...
+// TestGetPoolAndPoke tests that the right pools is returned from GetPoolAndPoke.
+// For the pools implementing the weighted extension, asserts that PokePool is called.
+func (suite *KeeperTestSuite) TestGetPoolAndPoke() {
+	const (
+		startTime = 1000
+		blockTime = startTime + 100
+	)
+
+	// N.B.: We make a copy because SmoothWeightChangeParams get mutated.
+	// We would like to avoid mutating global params
```
p0mvn, line 244:
```suggestion
	// We would like to avoid mutating global pool assets that are used in other tests.
```

x/gamm/types/pool.go
```diff
@@ -126,6 +122,19 @@ type PoolAmountOutExtension interface {
...
 }
 
+// WeightedPoolExtension is an extension of the PoolI interface
+// That defines an additional API for handling the pool's weights.
+type WeightedPoolExtension interface {
+	PoolI
+
+	// PokePool determines if a pool's weights need to be updated and updates
+	// them if so.
+	PokePool(blockTime time.Time)
```
ValarDragon, line 132: Why is poke pool on this? Its more general than for a weighted pool?

p0mvn: According to the [spec](https://github.com/osmosis-labs/osmosis/blob/babeebe957986ed13b6a3e7b2dc01f04da876f10/x/gamm/pool-models/balancer/pool.go#L395-L396) and balancer implementation, `PokePool` is only concerned with weights.

Stableswap did not have this method implemented, and I don't see why we would need it since there are no weights. So it seems to be weight-specific

x/gamm/keeper/pool.go
```diff
@@ -22,8 +22,8 @@ func (k Keeper) UnmarshalPool(bz []byte) (types.PoolI, error) {
 	return acc, k.cdc.UnmarshalInterface(bz, &acc)
 }
 
-// GetPoolAndPoke returns a PoolI based on it's identifier if one exists. Prior
-// to returning the pool, the weights of the pool are updated via PokePool.
+// GetPool returns a PoolI based on it's identifier if one exists. If poolId corresponds
+// to a pool with weights (e.g. balancer), the weights of the pool are updated via PokePool prior to returning.
```
czarcas7ic, line 26: What do you think about just keeping the name? It seems like its easier in terms of API to understand that the poke is a no-op for stableswap rather than leaving poke out of the name and having to look into the function to see it also pokes. No strong opinion either way so close as desired.

x/gamm/keeper/pool_test.go
```diff
@@ -220,3 +231,93 @@ package keeper_test
...
+		{
+			Weight: defaultPoolAssets[1].Weight.Mul(sdk.NewInt(3)),
+			Token:  defaultPoolAssets[1].Token,
+		},
+	}
+
+	tests := map[string]struct {
+		isPokePool  bool
+		poolId      uint64
+		expectError bool
```
czarcas7ic, line 262: seems we are not testing an expectError case?

x/gamm/types/pool.go
czarcas7ic: It seems like our current implementation of PokePool only concerns itself with updating weights. Is there some extension of PokePool that is planned that would make it useful for stableswap/concentrated-liquidity pools?

czarcas7ic (approved):
Approving this, but would say merging is contingent on consensus with entire chain dev team for way forward with concentrated liquidity

x/gamm/keeper/pool.go
p0mvn: whoops forgot to change this back after trying things but yeah should be `GetPoolAndPoke`

p0mvn:
This should be independent of the concentrated liquidity direction because, according to the spec and the balancer implementation of `PokePool`, it only seems to be concerned about pools with weights. So stableswap doesn't seem to need it.

However, I would like to get an ack from @ValarDragon to make sure I'm not missing anything

p0mvn:
I'm going to merge this since this is unrelated to concentrated liquidity + 2 approvals.

If issues arise, will address them separately.

Issue #3041, opened 2022-10-18
[Story][CL]: Concentrated Liquidity Module


p0mvn:
## Background

> As an engineer, I would like the concentrated liquidity logic to exist in its own module so that I can easily reason about the concentrated liquidity abstraction that is different from the existing pools.

Since concentrated liquidity architecture diverges from the existing amm logic, we would like to encapsulate the new implementation in its own module.

## Suggested Design

Specifically, we would like to implement the following keeper methods:
- `CreatePool`
- `JoinPool`
- `ExitPool`
- `SwapExactAmountIn`
- `SwapExactAmountOut`

This new module will **not** have any messages or the message server. Instead, we will implement separate modules that will call into the keeper methods of the `concentrated-liquidity` module.

## Tasks

- https://github.com/osmosis-labs/osmosis/issues/3044
- https://github.com/osmosis-labs/osmosis/issues/3045
- https://github.com/osmosis-labs/osmosis/issues/3059
- https://github.com/osmosis-labs/osmosis/issues/3063
- https://github.com/osmosis-labs/osmosis/issues/3069
- https://github.com/osmosis-labs/osmosis/issues/3108
- https://github.com/osmosis-labs/osmosis/issues/3134
- https://github.com/osmosis-labs/osmosis/issues/3140
- https://github.com/osmosis-labs/osmosis/issues/3243
- https://github.com/osmosis-labs/osmosis/issues/3245

## Acceptance Criteria
- new module is implemented
- follows `twap` folder structure
- no message server
- all keeper methods implemented
- utilizes the `concentrated` pool model
- test coverage is maximized

p0mvn:
TODO:
- create tasks for milestone 2:
   * barebones module structure
   * documentation of the store layout

p0mvn:
The core of the new concentrated liquidity module is now implemented so we consider this story as complete

PR #3050, opened 2022-10-19
Separate out concentrated liquidity as separate module

mattverse:
Closes: #3041 

Thing to note regarding this PR is that 
- Swapping related logics are implemented as Pool methods so that they are compatible with the `PoolI`, 
- Lping related logics are implemented as keeper methods so that we can use storage from keeper for them in the future


czarcas7ic (approved):
LGTM

PR #3056, opened 2022-10-19
CL: remove exponentiation from milestone 1 swaps

p0mvn:
Title


x/concentrated-liquidity/pool_test.go
```diff
@@ -19,23 +19,23 @@ func TestCalcOutAmtGivenIn(t *testing.T) {
 	swapFee := sdk.NewDec(0)
 
 	amountOut, _ := pool.CalcOutAmtGivenIn(ctx, tokensIn, tokenOutDenom, swapFee)
-	require.Equal(t, sdk.NewDec(664366278), amountOut.Amount.ToDec())
+	require.Equal(t, sdk.NewDec(663944647).String(), amountOut.Amount.ToDec().String())
```
czarcas7ic, line 22: Why was this made a string but the others were kept the same?

czarcas7ic (approved):

x/concentrated-liquidity/pool_test.go
p0mvn: I was just debugging and string is more readable but it doesn't really make much difference

Issue #3030, opened 2022-10-17
[Tracking Issue][CL]: Misc Tasks


p0mvn:
## Background

This issue keeps track of any concentrated liquidity task that does not belong to any of the user stories. These might be some configurations, pool wiring, dev ops, or integration tasks.

- https://github.com/osmosis-labs/osmosis/issues/3031
- https://github.com/osmosis-labs/osmosis/issues/3032 (closed, did not pursue)
- https://github.com/osmosis-labs/osmosis/issues/3039 (in progress: on refactored interface)
- https://github.com/osmosis-labs/osmosis/issues/3046
- https://github.com/osmosis-labs/osmosis/issues/3086
- https://github.com/osmosis-labs/osmosis/issues/3088
- https://github.com/osmosis-labs/osmosis/issues/3089
- https://github.com/osmosis-labs/osmosis/issues/3090
- https://github.com/osmosis-labs/osmosis/issues/3132
- https://github.com/osmosis-labs/osmosis/issues/3178
- https://github.com/osmosis-labs/osmosis/issues/3240
- https://github.com/osmosis-labs/osmosis/issues/3559
- https://github.com/osmosis-labs/osmosis/issues/3556
- https://github.com/osmosis-labs/osmosis/issues/3154
- https://github.com/osmosis-labs/osmosis/issues/3987
- https://github.com/osmosis-labs/osmosis/issues/3218
- https://github.com/osmosis-labs/osmosis/issues/3291
- https://github.com/osmosis-labs/osmosis/issues/3989
- https://github.com/osmosis-labs/osmosis/issues/3988
- https://github.com/osmosis-labs/osmosis/issues/3986
- https://github.com/osmosis-labs/osmosis/issues/3985
- https://github.com/osmosis-labs/osmosis/issues/3984
- https://github.com/osmosis-labs/osmosis/issues/3983
- https://github.com/osmosis-labs/osmosis/issues/3891
- https://github.com/osmosis-labs/osmosis/issues/3890
- https://github.com/osmosis-labs/osmosis/issues/3889
- https://github.com/osmosis-labs/osmosis/issues/3791
- https://github.com/osmosis-labs/osmosis/issues/3742
- https://github.com/osmosis-labs/osmosis/issues/3684
- https://github.com/osmosis-labs/osmosis/issues/3675
- https://github.com/osmosis-labs/osmosis/issues/3644
- https://github.com/osmosis-labs/osmosis/issues/3623
[...]

PR #3057, opened 2022-10-19
Add basic pool state, Make outlines for JoinPool

mattverse:
This PR includes
- adding basic pool state( this can change in the future as we're going through development)
- Adding store keys and getters, setters for the store
- Utilizing them for structuring join pool


proto/osmosis/concentrated-liquidity/concentratedPool.proto
```diff
@@ -17,4 +17,56 @@ message Pool {
...
+
+message Position {
+  string liquidity = 1 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"liquidity\"",
+    (gogoproto.nullable) = false
+  ];
 }
+
+message CurrentTick {
```
p0mvn, line 61: Do we need this? Seems that we define `current_sqrt_price` and `current_tick` directly in the pool

x/concentrated-liquidity/lp.go
```diff
@@ -1,6 +1,32 @@
...
-import sdk "github.com/cosmos/cosmos-sdk/types"
+import (
+	fmt "fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	types "github.com/osmosis-labs/osmosis/v12/x/concentrated-liquidity/types"
+)
+
+func (k Keeper) JoinPool(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, tokenIn sdk.Int, lowerTick sdk.Int, upperTick sdk.Int) (numShares sdk.Int, err error) {
```
p0mvn, line 11: Should we add some basic tests right away? Just making sure that tick / liquidity are updated etc

p0mvn, line 11: Wdyt about changing `JoinPool` to `Mint` or `ProvideLiquidity` since we've discussed wanting to incrementally move away from the pool abstraction in concentrated-liquidity

x/concentrated-liquidity/position.go
```diff
@@ -0,0 +1,41 @@
+package concentrated_liquidity
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v12/osmoutils"
+	types "github.com/osmosis-labs/osmosis/v12/x/concentrated-liquidity/types"
+)
+
+func (k Keeper) UpdatePositionWithLiquidity(ctx sdk.Context,
```
p0mvn, line 10: does it make sense to unexport this?

x/concentrated-liquidity/types/constants.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var (
+	MaxTick = sdk.NewIntFromUint64(887272)
```
p0mvn, line 8: Can we add TODO / issue to decide on the value here

x/concentrated-liquidity/types/keys.go
```diff
@@ -0,0 +1,34 @@
...
+	// Contract: Coin denoms cannot contain this character
+	KeySeparator = "|"
+)
+
+var (
+	TickPrefix     = "tick_prefix" + KeySeparator
+	PositionPrefix = "position_prefix" + KeySeparator
+)
+
+// uses pool Id and tick index
```
p0mvn, line 26:
```suggestion
// KeyTick uses pool Id and tick index
```

x/concentrated-liquidity/types/keys.go
```diff
@@ -0,0 +1,34 @@
...
+func KeyTick(poolId uint64, tickIndex sdk.Int) []byte {
+	return []byte(fmt.Sprintf("%s%d%s", TickPrefix, poolId, tickIndex.String()))
+}
+
+// uses pool Id, owner, lower tick and upper tick for keys
```
p0mvn, line 31:
```suggestion
// KeyPosition uses pool Id, owner, lower tick and upper tick for keys
```

p0mvn (approved):
LGTM.

I think would be great to start outlining some basic test structure asap so that we can build up on it in the subsequent PRs and milestones

czarcas7ic (approved):

mattverse:
@p0mvn  wdym by basic test structures?

x/concentrated-liquidity/tick.go
```diff
@@ -0,0 +1,37 @@
...
+	tickInfo.Liquidity = liquidityAfter
+
+	if liquidityBefore == sdk.ZeroInt() {
+		tickInfo.Initialized = true
+	}
+
+	k.setTickInfo(ctx, poolId, tickIndex, tickInfo)
+}
+
+func (k Keeper) getTickInfo(ctx sdk.Context, poolId uint64, tickIndex sdk.Int) TickInfo {
```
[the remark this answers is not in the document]
mattverse: Nah this is actually referring to the tick itself

x/concentrated-liquidity/lp.go
mattverse: I actually disagree with adding test at this point, considering that it doesn't have any logic or calculation besides updating state, which is not worth it considering time needed to do this. I would propose adding test on the next milestone (the second milestone)

proto/osmosis/concentrated-liquidity/concentratedPool.proto
mattverse: good catch, we actually dont

x/concentrated-liquidity/lp.go
mattverse: will go with Mint for now, we can decide on which method name would suit the best later on

x/concentrated-liquidity/types/constants.go
mattverse: Added in https://github.com/osmosis-labs/osmosis/issues/3030 as well as comment in code

x/concentrated-liquidity/lp.go
```diff
@@ -1,6 +1,32 @@
...
+
+	if tokenIn.IsZero() {
+		return sdk.Int{}, fmt.Errorf("token in amount is zero")
+	}
+
+	k.UpdateTickWithNewLiquidity(ctx, poolId, lowerTick, tokenIn)
+	k.UpdateTickWithNewLiquidity(ctx, poolId, upperTick, tokenIn)
+
+	// update tick with new liquidity
+	k.UpdatePositionWithLiquidity(ctx, poolId, owner.String(), lowerTick, upperTick, tokenIn)
```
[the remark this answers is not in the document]
mattverse: I suggest we postphone this convo as the methods can change in the process of development

p0mvn:
> 

@mattverse  Like stub test fixture, maybe some basic assertions that the state is updated but if you think it's better to do in the next PR, that's cool with me

Issue #3059, opened 2022-10-19
[CL][Concentrated Liquidity Module]: Prototype Tick Bitmap


p0mvn:
## Background

Univ3 paper talks about using tick bitmap for tracking the presence / absence of initialized ticks. We need to investigate and prototype this bitmap in our concentrated-liquidity module

## Acceptance Criteria
- bitmap prototype is made

p0mvn:
@alexanderbez FYI made this issue to track bitmap investigations

mattverse:
Closing Per the original PR being closed

PR #3065, opened 2022-10-19
feat(CL): Tick Bitmap

alexanderbez:
Closes: #3059

### DIFF
--- a/x/concentrated-liquidity/tick_bitmap.go
+++ b/x/concentrated-liquidity/tick_bitmap.go
@@ -1,0 +1,22 @@
+package concentrated_liquidity
+
+// TickBitmap defines a bitmap used to represent price ticks. It contains a
+// mapping of 64-bit words, where each bit in a word corresponds to a unique
+// tick. A set bit, i.e. a bit set to 1, indicates liquidity for that tick.
+// Conversely, an unset bit means there is no liquidity for that tick.
+type TickBitmap struct {
+	bitmap map[int16]uint64
+}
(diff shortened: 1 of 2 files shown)


alexanderbez:
Marking this R4R so we can start testing finding ticks for dynamic swaps...still need to implement tests. Could be that my approach is off here using 64-bit words or maybe even the translation of `NextInitializedTickWithinOneWord` from Solidity to Go.

x/concentrated-liquidity/tick_bitmap.go
```diff
@@ -0,0 +1,117 @@
...
+// TickBitmap defines a bitmap used to represent price ticks. It contains a
+// mapping of 64-bit words, where each bit in a word corresponds to a unique
+// tick. A set bit, i.e. a bit set to 1, indicates liquidity for that tick.
+// Conversely, an unset bit means there is no liquidity for that tick. Note,
+// ticks are in the range [−887272, 887272].
+//
+// Ref: https://uniswapv3book.com/docs/introduction/uniswap-v3/#ticks
+// Ref: https://uniswapv3book.com/docs/milestone_2/tick-bitmap-index/#bitmap
+type TickBitmap struct {
+	bitmap map[int16]uint64
```
p0mvn, line 18: How can we serialize and store this structure in the state?

Does it make sense to use a "prefix" map that we can store on disk?
e.g. we make a prefix of the format `bm< tick index >< key separator>< data >`

PR #3072, opened 2022-10-20
feat: adds tick price functions

czarcas7ic:

Adds tickToPrice and priceToTick functions


czarcas7ic:
Only note here, since our log function is still not optimal, the retrieved price and tick is off. Once log function is fixed, this function will work as intended

p0mvn (approved):

alexanderbez (approved):

Issue #3092, opened 2022-10-21
Add query for lockup params


mattverse:
## Background

We need to add query for the lockup params that was added in https://github.com/osmosis-labs/osmosis/pull/2733

## Acceptance Criteria

- Follow the format of param query supported in other modules and create a query for it.

hieuvubk:
Is anyone working on this issue? I can start with it

mattverse:
@hieuvubk That'd be awesome! Thanks!

hieuvubk:
> @hieuvubk That'd be awesome! Thanks!

Created a pr above could u pls check it out!

PR #3098, opened 2022-10-22
Query lockup params

hieuvubk:
Closes: #3092 

Add query for lockup module params

## Brief Changelog

Add grpc & cli cmd for new query
Add tests

## Testing and Verifying

All exist test & new tests passed


alexanderbez (approved):

x/lockup/keeper/grpc_query_test.go
```diff
@@ -513,3 +513,10 @@ func (suite *KeeperTestSuite) TestLockedDenom() {
 	testTotalLockedDuration("2h", 0)
 	testTotalLockedDuration("1h", 10)
 }
+
+func (suite *KeeperTestSuite) TestParams() {
```
mattverse, line 517: Can we add test cases for setting params and then querying it?

x/lockup/keeper/genesis.go
```diff
@@ -9,6 +9,7 @@ import (
 // InitGenesis initializes the capability module's state from a provided genesis
 // state.
 func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
+	k.SetParams(ctx, types.DefaultParams())
```
mattverse, line 12: good catch!

mattverse (commented):
Nice job! Thanks for this PR! 
Left some comments I would like to see before getting merged.

x/lockup/keeper/grpc_query_test.go
hieuvubk: Yeah sure

x/lockup/keeper/genesis.go
ValarDragon, line 12: oh wow, thanks!

ValarDragon (approved):

ValarDragon:
Nice work, ty!

Issue #2609, opened 2022-09-05
feat(release): automate post-upgrade tasks by code generating upgrade handler and auto-increasing e2e version


p0mvn:
## Background

The post-release process in https://github.com/osmosis-labs/osmosis/issues/2608 can be fully automated.

## Suggested Design

- Once a release a created, we can trigger the post-release GitHub Action
- The GitHub Action runs a script to generate an empty "next version" upgrade handler in `app/upgrades`
   - Some earlier code gen examples can be found here: https://github.com/osmosis-labs/osmosis/blob/roman/twap-e2e/cmd/querygen/README.md#L0-L1
- The e2e upgrade version is increased by the script
- Pull request with the changes is automatically created
   - Our import paths action can be a helpful reference: https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/import_paths.yml

## Acceptance Criteria
- the script is tested on a personal fork
- Proof PR from personal fork is attached
- E2E test is passing
- E2E test is testing the post-release upgrade handler

RusAkh:
hello @p0mvn is this available to work on?

p0mvn:
Yes, please go ahead. Thank you

PR #3054, opened 2022-10-19
feat(release): Automated post-upgrade tasks by code generating upgrade handler and auto-increasing e2e version

RusAkh:
Closes: #2609 

## Brief Changelog

On release creation will be triggered GitHub action which will make a pull request with added new file for new version in app/upgrades and increase E2E_UPGRADE_VERSION in Makefile

https://github.com/RusAkh/osmosis-fork/pull/8 Here is a pull request from a fork which shows what exactly this action does

I am not sure if this should be documented or not, so making it draft for now. What do you think @p0mvn?

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? no

PR #3100, opened 2022-10-22
feat(release): Automated post-upgrade tasks by code generating upgrade handler and auto-increasing e2e version

RusAkh:
Closes: https://github.com/osmosis-labs/osmosis/issues/2609

Now after a release is created a GitHub action will be triggered which will make a pull request with changes:

1) create a new folder for upgrade handler in app/upgrades
2) increase E2E_UPGRADE_VERSION in Makefile by one
3) change app/app.go file

an example of such pull request can be seen [here](https://github.com/RusAkh/osmosis-fork/pull/15)

the same, but draft pull request as this one can be found [here](https://github.com/osmosis-labs/osmosis/pull/3054) with changes from GitHub action included, so you can view in CI that all e2e tests pass

##Documentation and Release Note

Does this pull request introduce a new feature or user-facing behavior changes? no
Is a relevant changelog entry added to the Unreleased section in CHANGELOG.md? no
How is the feature or change documented? no


scripts/empty_upgrade_handler_gen.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+# this script creates an empty directory in app/upgrades called "vX" where X is a previous version + 1 and
+# copies upgrade handler from the previous release with increased version where needed. Also insures that all the imports
```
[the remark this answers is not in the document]
RusAkh: hello @p0mvn. I addressed this. However, the first approach would be hard to implement, so I did not decide to use it, and for the second one, I personally do not think that using snippets here would be appropriate. Correct me if I am wrong, please. 
I rewrote a script file such that now basically it constructs new files line by line using echo. That is a pretty dirty approach, but it works

.github/workflows/auto-update-upgrade.yml
```diff
@@ -0,0 +1,32 @@
+# When new release is created this workflow will be triggered and will do 3 things:
+# 1) it will create a directory with an empty upgrade handler in app/upgrades folder
+# 2) will increase an E2E_UPGRADE_VERSION variable in Makefile
+# 3) create a pull request with these changes to main 
+
+name: On Release Auto Upgrade
+
+on:
+  release: 
+    types: [created]
```
p0mvn, line 10: Does this only run on a major upgrade only?

We would like to avoid this running for minor/patch upgrades

scripts/empty_upgrade_handler_gen.sh
```diff
@@ -0,0 +1,93 @@
...
+ done
+ version_create=$((latest_version+1))
+ new_file=./app/upgrades/v${version_create}
+
+ mkdir $new_file
+ CONSTANTS_FILE=$new_file/constants.go
+ UPGRADES_FILE=$new_file/upgrades.go
+ touch $CONSTANTS_FILE
+ touch $UPGRADES_FILE
+
```
p0mvn, line 29:
```suggestion
```

p0mvn (commented):
This LGTM. Great work!

Would like to make sure two things before merge:
- only runs on major version releases
- the latest changes have been tested

Please confirm if those are true

.github/workflows/auto-update-upgrade.yml
RusAkh: hello @p0mvn. So basically what you mean is that it should run when in version vX.Y.Z of osmosis only X changes, for example from 12 to 13, right?

RusAkh: if so, then unfortunately this script will run when any of X Y or Z changes

but this can be addressed

RusAkh: also, are you sure you want this action to be run when a release is created and not published?

p0mvn: I think when the release is published is preferable

RusAkh: got it!

RusAkh:
@p0mvn changes addressed. Now PR will be created only if the first number in a tag is higher than existing upgrade handler's version

RusAkh:
so, replying to your previous comments:
1) now as I mentioned PR is only created at major updates
2) nothing changed except when is the script being ran, so there is nothing new to test basically (well, only that it indeed runs when intended and that is a yes)

RusAkh:
hello @p0mvn. I was wondering, do you think it is worth automatically assign someone for the pr created?

czarcas7ic (approved):
This looks great to me! Would love to see a secondary option in a later PR for forks as well, as forks usually mean bad times which usually means the more automation we can get the better 😄

p0mvn (approved):
Awesome work

p0mvn:
> hello @p0mvn. I was wondering, do you think it is worth automatically assign someone for the pr created?

I think we shouldn't assign anyone, someone can just pick it up

Issue #3062, opened 2022-10-19
[CL][Swapping]: Generalize Swaps


p0mvn:
## Background

In milestone 3, we are going to generalize swapping to work with any values (not hardcoded)

## Acceptance Criteria
- minting works for any value
- tests added

mattverse:
Closed by https://github.com/osmosis-labs/osmosis/pull/3137

PR #3101, opened 2022-10-22
concentrated liquidity: further swap/lp work

czarcas7ic:
Closes: #3062 

Adds stepState and swapState

I have generalized everything for swap except for liquidity for asset a and asset b which is hardcoded to 1:5000. I am stumped on how this is supposed to be retrieved from the pool itself.


x/concentrated-liquidity/pool.go
```diff
@@ -143,25 +165,34 @@ func (p Pool) CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coin, tokenInDenom
...
+		return sdk.Coin{}, fmt.Errorf("issue calculating square root of minPrice")
+	}
+	sqrtPCurTick := curSqrtPrice
+	sqrtPUpperTick, err := maxPrice.ApproxSqrt()
+	if err != nil {
+		return sdk.Coin{}, fmt.Errorf("issue calculating square root of maxPrice")
+	}
 
-	// TODO: Roman change out with query to pool to get this info
+	// TODO: How do we remove/generalize this? I am stumped.
```
alexanderbez, line 186: Do these values reflect the current spot prices or liquidity?

czarcas7ic: Current liquidity which should be the spot price

x/concentrated-liquidity/lp.go
```diff
@@ -24,26 +24,22 @@ func (k Keeper) Mint(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, liqui
...
-	currentSqrtPrice := pool.CurrentSqrtPrice
-	sqrtRatioUpperTick, err := k.getSqrtRatioAtTick(upperTick)
-	if err != nil {
-		return sdk.Int{}, sdk.Int{}, err
-	}
-	sqrtRatioLowerTick, err := k.getSqrtRatioAtTick(lowerTick)
-	if err != nil {
-		return sdk.Int{}, sdk.Int{}, err
-	}
+	currentSqrtPrice := sdk.NewDecWithPrec(int64(pool.CurrentSqrtPrice.Uint64()), 6)
```
p0mvn, line 37: Please avoid using  `uint64` to avoid the risk of overflows

x/concentrated-liquidity/lp.go
```diff
@@ -24,26 +24,22 @@ func (k Keeper) Mint(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, liqui
...
+	sqrtRatioUpperTick, _ := k.tickToPrice(sdk.NewInt(upperTick)).ApproxSqrt()
+	sqrtRatioLowerTick, _ := k.tickToPrice(sdk.NewInt(lowerTick)).ApproxSqrt()
 
-	amtDenom0 = calcAmount0Delta(currentSqrtPrice.ToDec(), sqrtRatioUpperTick, liquidityIn.ToDec()).RoundInt()
-	amtDenom1 = calcAmount1Delta(currentSqrtPrice.ToDec(), sqrtRatioLowerTick, liquidityIn.ToDec()).RoundInt()
+	amtDenom0 = calcAmount0Delta(liquidityIn.ToDec(), currentSqrtPrice, sqrtRatioUpperTick).RoundInt()
```
p0mvn, line 41: Why do we round and not truncate?

x/concentrated-liquidity/pool.go
```diff
@@ -9,20 +9,39 @@ import (
 	"github.com/osmosis-labs/osmosis/v12/x/gamm/types"
 )
 
+// maintains the current swap state
```
p0mvn, line 12:
```suggestion
// SwapState maintains the current swap state
```

x/concentrated-liquidity/pool.go
```diff
@@ -9,20 +9,39 @@ import (
...
+type SwapState struct {
+	amountSpecifiedRemaining sdk.Dec // remaining amount of tokens that need to be bought by the pool
+	amountCalculated         sdk.Dec // amount out
+	sqrtPrice                sdk.Dec // new current price when swap is done
+	tick                     sdk.Int // new tick when swap is done
+}
+
+// tracks state of one iteration of an order filling
```
p0mvn, line 20:
```suggestion
// StepState tracks state of one iteration of an order filling
```

x/concentrated-liquidity/pool.go
```diff
@@ -76,127 +95,177 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
+	swapState := SwapState{
+		amountSpecifiedRemaining: tokenAmountInAfterFee,
+		amountCalculated:         sdk.ZeroDec(),
+		sqrtPrice:                sqrtPCurTick,
+		tick:                     priceToTick(sqrtPCurTick.Power(2)),
+	}
+	// TODO: This should be GT 0 but some instances have very small remainder
+	// need to look into fixing this
+	for swapState.amountSpecifiedRemaining.GT(sdk.NewDecWithPrec(1, 6)) {
+		stepState := StepState{}
```
p0mvn, line 155: IMO these `StepState` and `SwapState` abstractions add more complexity. I think that declaring each variable directly would be more readable

However, let's not block on this comment

x/concentrated-liquidity/lp.go
mattverse, line 37: what's the reason we are using 6 prec points? I thought the goal was to remove them frmo code base

x/concentrated-liquidity/tick.go
```diff
@@ -12,13 +12,9 @@ import (
...
 
-func (k Keeper) getSqrtRatioAtTick(tickIndex int64) (sdk.Dec, error) {
-	sqrtRatio, err := sdk.NewDecWithPrec(10001, 4).Power(uint64(tickIndex)).ApproxSqrt()
-	if err != nil {
-		return sdk.Dec{}, nil
-	}
-
-	return sqrtRatio, nil
+func (k Keeper) tickToPrice(tickIndex sdk.Int) sdk.Dec {
+	price := sdk.NewDecWithPrec(10001, 4).Power(tickIndex.Uint64())
```
mattverse, line 16: what's the reason we don't do sqrt any more?

x/concentrated-liquidity/pool.go
```diff
@@ -76,127 +95,177 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
 // this only works on a single directional trade, will implement bi directional trade in next milestone
-func (p Pool) CalcOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coin, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.Coin, err error) {
+func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coin, tokenOutDenom string, swapFee sdk.Dec, minPrice, maxPrice sdk.Dec, poolId uint64) (newTokenIn, tokenOut sdk.Coin, err error) {
+	p := k.getPoolbyId(ctx, poolId)
 	asset0 := p.Token0
 	asset1 := p.Token1
+	tokenAmountInAfterFee := tokenIn.Amount.ToDec().Mul(sdk.OneDec().Sub(swapFee))
+
+	// get current sqrt price from pool
+	curSqrtPrice := sdk.NewDecWithPrec(int64(p.CurrentSqrtPrice.Uint64()), 6)
```
mattverse, line 105: ditto here for the precision points

x/concentrated-liquidity/pool.go
```diff
@@ -76,127 +95,177 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
+			liq,
+			swapState.amountSpecifiedRemaining,
+			lte,
+		)
+		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(stepState.amountIn)
+		swapState.amountCalculated = swapState.amountCalculated.Add(stepState.amountOut)
+		swapState.tick = priceToTick(swapState.sqrtPrice.Power(2))
+	}
+	newTokenIn.Amount = tokenIn.Amount.Sub(swapState.amountSpecifiedRemaining.RoundInt())
+	return sdk.NewCoin(tokenIn.Denom, newTokenIn.Amount), sdk.NewCoin(tokenOutDenom, swapState.amountCalculated.RoundInt()), nil
```
p0mvn, line 175: I don't think we should be r using bankers rounding anywhere here. It should either be rounding up or down.

For example, for tokenIn, I would assume we should round up so that we don't undercharge user.

x/concentrated-liquidity/math.go
```diff
@@ -52,3 +54,20 @@ func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 	diff := sqrtPriceB.Sub(sqrtPriceA)
 	return liq.Mul(diff)
 }
+
+func computeSwapStep(sqrtPriceCurrent, sqrtPriceTarget, liquidity, amountRemaining sdk.Dec, lte bool) (sqrtPriceNext sdk.Dec, amountIn sdk.Dec, amountOut sdk.Dec) {
```
nicolaslara, line 58: does `lte` represent if we're moving up or down in the x axis? Can we document this to make it clearer?

x/concentrated-liquidity/pool.go
```diff
@@ -76,127 +95,177 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
+	sqrtPCurTick := curSqrtPrice
+	sqrtPUpperTick, err := maxPrice.ApproxSqrt()
+	if err != nil {
+		return sdk.Coin{}, sdk.Coin{}, fmt.Errorf("issue calculating square root of maxPrice")
+	}
 
-	// TODO: Roman change out with query to pool to get this info
+	// TODO: How do we remove/generalize this? I am stumped.
 	amountETH := int64(1000000)
 	amountUSDC := int64(5000000000)
```
p0mvn, line 137: Can we add a comment that these are virtual amounts?

x/concentrated-liquidity/lp.go
mattverse: Why not round? IMO rounding is more accurate no?

nicolaslara (approved):
LGTM

x/concentrated-liquidity/pool.go
```diff
@@ -75,128 +77,185 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
+		amountCalculated:         sdk.ZeroDec(),
+		sqrtPrice:                curSqrtPrice,
+		tick:                     priceToTick(curSqrtPrice.Power(2)),
+	}
+
+	// TODO: This should be GT 0 but some instances have very small remainder
+	// need to look into fixing this
+	for swapState.amountSpecifiedRemaining.GT(sdk.NewDecWithPrec(1, 6)) {
+		lte := tokenIn.Denom == asset1
+		nextTick, _ := k.NextInitializedTick(ctx, poolId, swapState.tick.Int64(), lte)
```
alexanderbez, line 147: This could return `false` if there is no tick found. maybe return an error in that case?

x/concentrated-liquidity/pool.go
czarcas7ic: changed the logic to round up and down as requested

x/concentrated-liquidity/pool_test.go
```diff
@@ -1,94 +1,100 @@
...
+	_, amountOut, err = s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, minPrice, maxPrice, pool.Id)
 	s.Require().NoError(err)
 	s.Require().Equal(sdk.NewDec(805287), amountOut.Amount.ToDec())
 
+	// test asset b to a logic
+	tokenIn = sdk.NewCoin("usdc", sdk.NewInt(42000000))
+	tokenOutDenom = "eth"
+	swapFee = sdk.NewDec(0)
+
+	_, amountOut, err = s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, minPrice, maxPrice, pool.Id)
```
mattverse, line 37: utack on these test cases, I would love to see detailed comments on where these numbers are from or how the math is done to get these numbers, don't think it has to be in the scope of this PR

x/concentrated-liquidity/pool.go
```diff
@@ -75,128 +77,197 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
...
+		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(amountIn)
+		swapState.amountCalculated = swapState.amountCalculated.Add(amountOut.Quo(sdk.OneDec().Sub(swapFee)))
+		swapState.tick = priceToTick(sqrtPrice.Power(2))
 	}
-	return sdk.Coin{}, fmt.Errorf("tokenOut (%s) does not match any asset in pool", tokenOut.Denom)
+	return sdk.NewCoin(tokenInDenom, swapState.amountCalculated.RoundInt()), sdk.NewCoin(tokenOut.Denom, tokenOut.Amount), nil
 }
 
-func (p *Pool) orderInitialPoolDenoms(denom0, denom1 string) error {
+func (k Keeper) orderInitialPoolDenoms(denom0, denom1 string) (string, string, error) {
```
mattverse, line 264: maybe we should just move this method to type package?

Issue #3155, opened 2022-10-25
[CL][Liquidity Provision] Cross tick minting


mattverse:
## Background

Previously the minting method, or the JoinPool method supported only single tick minting. 

Now we want to support cross-tick minting when joining the pool

PR #3158, opened 2022-10-26
Support cross tick create position

mattverse:
Closes: #3155

Supports cross tick for creating a position in a concentrated liquidity pool.


czarcas7ic (approved):
2ez

PR #3163, opened 2022-10-26
Change `lte` to `zeroToOne`

mattverse:
Existing code base introduces a `lte` parameter, which was suppose to indicate if we're going down the tick or up the tick, depending on if we're setting the base asset as asset 0 or asset 1. 

The proposal is to change `lte` parameter to a varaiable `zeroToOne` to make the variable more clear and intuitive. 
ZeroToOne indicates that we're changing from asset zero to one if true. If false, changing asset one to zero

### DIFF
--- a/x/concentrated-liquidity/math.go
+++ b/x/concentrated-liquidity/math.go
@@ -79,8 +79,8 @@ func computeSwapStep(sqrtPriceCurrent, sqrtPriceTarget, liquidity, amountRemaini
 	return sqrtPriceNext, amountIn, amountOut
 }
 
-func getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec, lte bool) (sqrtPriceNext sdk.Dec) {
-	if lte {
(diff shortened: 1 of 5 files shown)


x/concentrated-liquidity/math.go
```diff
@@ -55,35 +55,35 @@ func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+	} else {
+		amountIn = calcAmount1Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent)
+		amountOut = calcAmount0Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent)
 	}
 
 	return sqrtPriceNext, amountIn, amountOut
 }
 
 func getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec, lte bool) (sqrtPriceNext sdk.Dec) {
 	if lte {
```
czarcas7ic, line 83: remove

czarcas7ic (approved):

Issue #3132, opened 2022-10-24
osmomath(log/CL): implement and test natural logarithm, custom base logarithm and `1.0001` logarithm 


p0mvn:
## Background

With log base 2 implemented in https://github.com/osmosis-labs/osmosis/pull/2788, we should add more bases for logarithms.

### Natural logarithm

`ln x = \frac{log_{2}{x}}{log_{2}{1.0001}}`

- estimate `{log_{2}{e}` in WolframAlpha (add link to the estimate)
- hard code the estimate
- create a method `Ln` that utilizes `LogBase2` and hardcoded `e` value to solve the above

## Base 1.0001
This is what we need for concentrated liquidity
$$ln x = \frac{log_{2}{x}}{log_{2}{1.0001}}$$ 

- estimate `{log_{1.0001}{x}` in WolframAlpha (add link to the estimate)
- hard code the estimate
- create a method `LogBaseHundredthBasisPoint` that utilizes `LogBase2` and hardcoded `log_{1.0001}{x}` value to solve the above

## Custom Base

Implement `LogCustomBase` that utilizes similar idea to above

## Acceptance Criteria
- tests added for each method
- wolfram links with calculations
- benchmarks
- similar to https://github.com/osmosis-labs/osmosis/pull/2788 
- make PR against main with backport to `concentrated-liquidity-main`

RusAkh:
I would like to work on it

p0mvn:
Thank you. Please go ahead

PR #3169, opened 2022-10-26
osmomath(log/CL): ln(x), log_1.0001(x), log_custom(x)

RusAkh:
Closes: #3132 

`log` functions

## Testing and Verifying

tests added (2 test cases end with too big absolute error, I commented them in code)

### DIFF
--- a/osmomath/decimal_test.go
+++ b/osmomath/decimal_test.go
@@ -741,7 +741,7 @@ func (s *decimalTestSuite) TestLog2() {
 }
 
 func (s *decimalTestSuite) TestLn() {
-	var expectedErrTolerance = MustNewDecFromStr("0.000000000000000000000000000000000140")
+	var expectedErrTolerance = MustNewDecFromStr("0.000000000000000000000000000000000100")
 
 	tests := map[string]struct {
 		initialValue BigDec
@@ -803,7 +803,7 @@ func (s *decimalTestSuite) TestLn() {
 		},
 		"log_e{912648174127941279170121098210.92821920190204131121} = 68.986147965719214790400745338243805015": {
 			initialValue: MustNewDecFromStr("912648174127941279170121098210.92821920190204131121"),
-			// From: https://www.wolframalpha.com/input?i=68.986147965719214790400745338243805015&assumption=%22ClashPrefs%22+-%3E+%7B%22Math%22%7D
+			// From: https://www.wolframalpha.com/input?i=log912648174127941279170121098210.92821920190204131121+to+38+digits
 			expected: MustNewDecFromStr("68.986147965719214790400745338243805015"),
 		},
 	}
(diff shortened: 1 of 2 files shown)


osmomath/decimal_test.go
```diff
@@ -739,3 +739,263 @@ func (s *decimalTestSuite) TestLog2() {
...
+		"log_0.2(0.99) = 0.006244624769837438271878639001855450": {
+			initialValue: MustNewDecFromStr("0.99"),
+			base:         MustNewDecFromStr("0.2"),
+			// From: https://www.wolframalpha.com/input?i=log_0.2%280.99%29+to+34+digits
+			expected: MustNewDecFromStr("0.006244624769837438271878639001855450"),
+		},
+		// THIS TEST CASE FAILS WITH 0.722861284485807159166346369373275313 ERROR
+		// "log_0.0001(500000) = -1.424742501084004701196565276318876743": {
+		// 	initialValue: NewDecWithPrec(1, 4),
+		// 	base:         NewBigDec(500000),
```
p0mvn, line 942: Shouldn't `base` and `initialValue` be switched?

osmomath/decimal_test.go
```diff
@@ -739,3 +739,263 @@ func (s *decimalTestSuite) TestLog2() {
...
+		// 	// From: https://www.wolframalpha.com/input?i=log_0.0001%28500000%29+to+37+digits
+		// 	expected: MustNewDecFromStr("-1.424742501084004701196565276318876743"),
+		// },
+
+		// THIS TEST CASE FAILS WITH 0.722861284485807159166346369373275311 ERROR
+		// "log_500000(0.0001) = -0.701881216598197542030218906945601429": {
+		// 	initialValue: NewBigDec(500000),
+		// 	base:         NewDecWithPrec(1, 4),
```
p0mvn, line 950: I suspect the same might be happening here

osmomath/decimal_test.go
```diff
@@ -739,3 +739,263 @@ func (s *decimalTestSuite) TestLog2() {
...
+		},
+		// Fails with 0.000000000000000000000000000000007686!!!!!
+		"log_1.123{1024.987654321} = 59.760484327223888489694630378785099461": {
+			initialValue: NewDecWithPrec(1024987654321, 9),
+			base:         NewDecWithPrec(1123, 3),
+			// From: https://www.wolframalpha.com/input?i=log_1.123%281024.987654321%29+to+38+digits
+			expected: MustNewDecFromStr("59.760484327223888489694630378785099461"),
+		},
+		// Fails with 0.000000000000000000000000000000077705!!!!!
+		"log_1.123{912648174127941279170121098210.92821920190204131121} = 594.689327867863079177915648832621538986": {
```
p0mvn, line 979: I think this is fine. Let's do the following:
- add a test case field `expectedErrorTolerance`
- select the minimal `expectedErrorTolerance` value for each test case

osmomath/decimal_test.go
```diff
@@ -739,3 +739,282 @@ func (s *decimalTestSuite) TestLog2() {
...
+			expectedPanic: true,
+		},
+		"log_1.0001{0}; invalid; panic": {
+			initialValue:  ZeroDec(),
+			expectedPanic: true,
+		},
+		"log_1.0001{0.001} = -69081.006609899112313305835611219486392199": {
+			initialValue: MustNewDecFromStr("0.001"),
+			// From: https://www.wolframalpha.com/input?i=log_1.0001%280.001%29+to+41+digits
+			expectedErrTolerance: MustNewDecFromStr("0.000000000000000000000000000143031879"),
```
czarcas7ic, line 847: Do we think it makes more sense to have varying error tolerance levels for each test case or just a single overarching one for all the tests? The later seems more readable but I could be missing some important reason why this is not being done.

czarcas7ic (approved):
Just have a comment on the error tolerance, all else LGTM!

osmomath/decimal_test.go
p0mvn: While I agree that the latter is more readable, I think it's helpful to know the error tolerance for some subsets of inputs that we can infer from test cases.

I would prefer this knowledge over readability here

p0mvn (approved):
Great work

PR #3175, opened 2022-10-28
ibc-go v3.3.1

faddat:
Upgrades osmosis to ibc-go v3.3.1

## Brief Changelog

Changes go.mod and go.sum to use a later version of ibc.


ValarDragon:
Doesn't matter for us, as its just an SDK version bump, that we already replace.

Issue #3196, opened 2022-11-01
[CL][Swapping] Add Test cases for `CalcOutAmtGivenIn`


mattverse:
## Background

Current `CalcOutAmtGivenIn` only has test cases for doing a swap within one price range. 

We want to add more test cases so that we have the following test cases for the swappings

## Suggested Design
The following are the test cases that are planned to be added:
- Swapping on two equal price ranges
- Swapping on consecutive price ranges
- Swapping on partially overlapping price ranges

## Acceptance Criteria

- [ ] Test cases listed above are added 
- [ ] In depth comment on the test cases are added(how we get the calculation and maths)

czarcas7ic:
Closing as addressed in previous PRs

PR #3202, opened 2022-11-01
Concentrated Liquidity: fix `computeSwapStep` and add test case with two positions

mattverse:
Closes: #3196 

This PR is the attempt to fix `computeSwapStep` to use correct method of calculation, and add a test for a case with two positions. 

Currently the tests are failing (inclduing the `calcOutAmtGivenIn` for a single position that was existing before) investigateing into this


mattverse:
Her, I think the rebase of the base branch messed up the diffs between concentrated-liquidity-main branch and this branch. Would create a new PR instead tmrw

x/concentrated-liquidity/math.go
```diff
@@ -56,7 +56,9 @@ func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 
 // computeSwapStep calculates the amountIn, amountOut, and the next sqrtPrice given current price, price target, tick liquidity, and amount available to swap
 // lte is reference to "less than or equal", which determines if we are moving left or right of the current price to find the next initialized tick with liquidity
-func computeSwapStep(sqrtPriceCurrent, sqrtPriceTarget, liquidity, amountRemaining sdk.Dec, zeroForOne bool) (sqrtPriceNext, amountIn, amountOut sdk.Dec) {
+func computeSwapStep(sqrtPriceCurrent, sqrtPriceTarget, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext, amountIn, amountOut sdk.Dec) {
```
mattverse, line 59: Main logic change for this PR

Issue #3178, opened 2022-10-29
Concentrated Liquidity Chore Tracking issue


mattverse:
This Issue serves the purpose of tracking the chores needed to be done for concentrated liquidity

- [ ] Remove `currTick` from argument of `CreateNewConcentratedLiquidityPool`: we should be calculating currTick inside the method instead of having it passed on as an argument
- [ ] Make `CreateNewConcentratedLiquidityPool` return pool id instead of pool object
- [ ] Add function to mint lp shares and send coins from account to pool
- [ ] Do calculation for `MinSqrtRatio`
- [ ] #3203 
- [ ] Add logic to handle if provided sqrt price limit is 0, use min sqrt ratio and max sqrt ratio for amountSpecified(tokenIn)

hieuvubk:
Hi @mattverse is `CreateNewConcentratedLiquidityPool` issue still needed? Id like to work with it

p0mvn:
I'm going to close this due to being out-of-date. Most of the tasks here seem to be addressed

PR #3203, opened 2022-11-01
Concentrated Liquidity chore: Fix `CalcOutAmtGivenIn` to return correct amount of tokenIn

mattverse:
Sub component of : #3178

Previously we were returning the delta amount of token in from `CalcOutAmtGivenIn` , not the actual amount of token in. Instead, the case should have been that we return the token amount in, by doing the calculation of tokenInMin - delta amount. 

## Testing and Verifying
Changed existing tests to confirm correct calculation

### DIFF
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -133,7 +133,7 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 	// at first, we use the pool liquidity
 	swapState := SwapState{
 		amountSpecifiedRemaining: tokenAmountInAfterFee,
-		amountCalculated:         sdk.ZeroDec(),
+		amountCalculated:         tokenAmountInAfterFee,
 		sqrtPrice:                curSqrtPrice,
 		tick:                     priceToTick(curSqrtPrice.Power(2)),
 		liquidity:                p.Liquidity,
--- a/x/concentrated-liquidity/pool_test.go
+++ b/x/concentrated-liquidity/pool_test.go
@@ -40,7 +40,7 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 	swapFee := sdk.ZeroDec()
 	priceLimit := sdk.NewDec(5004)
 	tokenIn, tokenOut, updatedTick, updatedLiquidity, err := s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenIn(s.Ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, pool.Id)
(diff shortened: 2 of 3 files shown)


x/concentrated-liquidity/pool_test.go
```diff
@@ -40,7 +40,7 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 	swapFee := sdk.ZeroDec()
 	priceLimit := sdk.NewDec(5004)
 	tokenIn, tokenOut, updatedTick, updatedLiquidity, err := s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenIn(s.Ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, pool.Id)
-	expectedTokenIn := sdk.NewCoin("usdc", sdk.NewInt(0))
+	expectedTokenIn := sdk.NewCoin("usdc", sdk.NewInt(42))
 	expectedTokenOut := sdk.NewCoin("eth", sdk.NewInt(42))
```
czarcas7ic, line 44: I am a little confused on why the expected in and out would be the same here

mattverse: The swap here is in 42 usdc, out 42 eth

czarcas7ic: ser where can I find this deal lol

mattverse: ser no changes with the expected numbers of test cases

mattverse: The current tests also puts in 42 usdc, gets 42 eth out right?

czarcas7ic: No, the current test puts in 42000000usdc (42usdc) and gets out 008396 eth (price of 5000usd per eth)

x/concentrated-liquidity/pool.go
```diff
@@ -133,7 +133,7 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 	// at first, we use the pool liquidity
 	swapState := SwapState{
 		amountSpecifiedRemaining: tokenAmountInAfterFee,
-		amountCalculated:         sdk.ZeroDec(),
+		amountCalculated:         tokenAmountInAfterFee,
```
czarcas7ic, line 136: This should start at zero, right?

czarcas7ic: This should get updated each loop on how much has been calculated thus far

mattverse:
Having significent difference in #3205 , planning to close this PR and re open one for this

PR #3205, opened 2022-11-01
CL: fix CalcOutAmtGivenIn and other logic

czarcas7ic:

This corrects faulty logic throughout the CL module, mostly within CalcOut and ComputeSwapStep

I think for now we should leave these tests as is, and have a subsequent issue convert it to a clean table driven test

This confirms OneToZero logic for CalcOut is good, I still need to check if ZeroToOne logic is sound (there is no reason why it shouldn't, but I didn't have time to check and wanted to get this merged to main ASAP)

### DIFF
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -67,7 +67,6 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		// this means position is solely made up of asset0
 		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = sdk.ZeroInt()
-
 	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
 		// outcome two: the current price falls within the position
 		// if this is the case, we attempt to provide liquidity evenly between asset0 and asset1
@@ -75,7 +74,6 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick, false).RoundInt()
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
-
 	} else {
 		// outcome three: position is above current price
 		// this means position is solely made up of asset1
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -62,23 +62,23 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
 	}
 
-	// outcome one: position is below current price
-	// this means position is solely made up of asset0
 	if pool.CurrentTick.LT(sdk.NewInt(lowerTick)) {
+		// outcome one: position is below current price
+		// this means position is solely made up of asset0
 		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = sdk.ZeroInt()
 
+	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
 		// outcome two: the current price falls within the position
 		// if this is the case, we attempt to provide liquidity evenly between asset0 and asset1
 		// we also update the pool liquidity since the virtual liquidity is modified by this position's creation
-	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
 		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick, false).RoundInt()
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
 
+	} else {
 		// outcome three: position is above current price
 		// this means position is solely made up of asset1
-	} else {
 		amtDenom0 = sdk.ZeroInt()
 		amtDenom1 = calcAmount1Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 	}
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -77,7 +77,7 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
 
 		// outcome three: position is above current price
-		// tis means position is solely made up of asset1
+		// this means position is solely made up of asset1
 	} else {
 		amtDenom0 = sdk.ZeroInt()
 		amtDenom1 = calcAmount1Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -62,14 +62,22 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
 	}
 
-	// update pool state as well only if we're injecting liquidity at the current tick of the pool
+	// outcome one: position is below current price
+	// this means position is solely made up of asset0
 	if pool.CurrentTick.LT(sdk.NewInt(lowerTick)) {
 		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = sdk.ZeroInt()
+
+		// outcome two: the current price falls within the position
+		// if this is the case, we attempt to provide liquidity evenly between asset0 and asset1
+		// we also update the pool liquidity since the virtual liquidity is modified by this position's creation
 	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
 		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick, false).RoundInt()
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
+
+		// outcome three: position is above current price
+		// tis means position is solely made up of asset1
 	} else {
 		amtDenom0 = sdk.ZeroInt()
 		amtDenom1 = calcAmount1Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -64,15 +64,15 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 
 	// update pool state as well only if we're injecting liquidity at the current tick of the pool
 	if pool.CurrentTick.LT(sdk.NewInt(lowerTick)) {
-		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick).RoundInt()
+		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 		amtDenom1 = sdk.ZeroInt()
 	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
-		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick).RoundInt()
-		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick).RoundInt()
+		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick, false).RoundInt()
+		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick, false).RoundInt()
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
 	} else {
 		amtDenom0 = sdk.ZeroInt()
-		amtDenom1 = calcAmount1Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick).RoundInt()
+		amtDenom1 = calcAmount1Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
 	}
 
 	k.setPoolById(ctx, pool.Id, pool)
--- a/x/concentrated-liquidity/math.go
+++ b/x/concentrated-liquidity/math.go
@@ -62,8 +62,12 @@ func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
 	// if calculating for amountIn, we round up
-	// if calculating for amountOut, we don't round
+	// if calculating for amountOut, we don't round at all
 	// this is to prevent removing more from the pool than expected due to rounding
+	// example: we calculate 1000000.9999999 uusdc (~$1) amountIn and 2000000.999999 uosmo amountOut
+	// we would want the used to put in 1000001 uusdc rather than 1000000 uusdc to ensure we are charging enough for the amount they are removing
+	// additionally, without rounding, there exists cases where the swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) for loop within
+	// the CalcOut/In functions never actually reach zero due to dust that would have never gotten counted towards the amount (numbers after the 10^6 place)
 	if roundUp {
 		return liq.Mul(diff).Ceil()
 	}
--- a/x/concentrated-liquidity/math.go
+++ b/x/concentrated-liquidity/math.go
@@ -39,6 +39,9 @@ func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
 	mult := liq
+	// if calculating for amountIn, we round up
+	// if calculating for amountOut, we don't round
+	// this is to prevent removing more from the pool than expected due to rounding
 	if roundUp {
 		return ((mult.Mul(diff)).QuoRoundUp(sqrtPriceB).QuoRoundUp(sqrtPriceA)).Ceil()
 	}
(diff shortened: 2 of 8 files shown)


x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 // sqrtPriceA is the smaller of sqrtpCur and the nextPrice
 // sqrtPriceB is the larger of sqrtpCur and the nextPrice
 // calcAmount0Delta = (liquidity * (sqrtPriceB - sqrtPriceA)) / (sqrtPriceB * sqrtPriceA)
-func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
+func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec {
```
mattverse, line 36: What's the reason behind in somecases we round up, some cases we just round down? (Which cases do we round up which cases do we round down and why?)

x/concentrated-liquidity/pool.go
```diff
@@ -100,7 +100,7 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 	tokenOutDenom string,
 	swapFee sdk.Dec,
 	priceLimit sdk.Dec,
-	poolId uint64) (tokenIn, tokenOut sdk.Coin, updatedTick sdk.Int, updatedLiquidity sdk.Dec, err error) {
+	poolId uint64) (tokenInDelta, tokenOutDelta sdk.Int, updatedTick sdk.Int, updatedLiquidity sdk.Dec, err error) {
```
mattverse, line 103: Im wondering if we should do tokenIn and tokenOut instead? ( I personally think this way would be more useful as API) since the caller would have to recalculate on the actual token In and token Out. What do you think?

x/concentrated-liquidity/pool.go
```diff
@@ -139,15 +139,17 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 		liquidity:                p.Liquidity,
 	}
 
-	for swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) && !swapState.sqrtPrice.Equal(sqrtPriceLimit) {
+	// had to reintroduce the dust check, we need to figure out a cleaner way to do this
+	for swapState.amountSpecifiedRemaining.GT(sdk.NewDecWithPrec(1, 8)) && !swapState.sqrtPrice.Equal(sqrtPriceLimit) {
```
mattverse, line 143: So I personally think we should not reintroduce dust check.
I was digging around with this for loop yesterday, and the closest answer I got to was that it only works with the dust check not because of some sort of minor rounding issues, its because that the calculation being done is off(incorrect) and we're just force exiting from the loop. 

I, myself have to do more investigation on this, and I'm also fine with leaving the dust check for now, but just wanted to let you know that having the dust check means that we still have a point that we missed somewhere!

x/concentrated-liquidity/pool_test.go
```diff
@@ -12,50 +12,182 @@ import (
...
+	swapFee = sdk.ZeroDec()
+	// limit max price impact to 5004 usdc per eth
+	priceLimit = sdk.NewDec(5004)
+
+	// run calculation
+	tokenInDelta, tokenOutDelta, updatedTick, updatedLiquidity, err = s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, pool.Id)
+	s.Require().NoError(err)
+
+	// we expect to put 41999999 usdc in and in return get .008396 eth back
+	expectedTokenInDelta = sdk.NewInt(41999999)
```
mattverse, line 111: This doesnt seem like the "delta" but the actual amount of token in?

x/concentrated-liquidity/pool_test.go
```diff
@@ -12,50 +12,182 @@ import (
...
+	// check pool liquidity
+	lowerSqrtPrice, err = s.App.ConcentratedLiquidityKeeper.TickToSqrtPrice(lowerTick)
+	s.Require().NoError(err)
+	upperSqrtPrice, err = s.App.ConcentratedLiquidityKeeper.TickToSqrtPrice(upperTick)
+	s.Require().NoError(err)
+	expectedLiquidity = cl.GetLiquidityFromAmounts(currSqrtPrice, lowerSqrtPrice, upperSqrtPrice, amount0, amount1)
+	s.Require().Equal(expectedLiquidity.String(), updatedLiquidity.String())
+
+	//
+	// TEST 3: two consecutive price ranges
```
mattverse, line 130: Consider separating these tests out

x/concentrated-liquidity/pool_test.go
czarcas7ic: It is the delta. Token in delta is what the pool supply changes by, and the token out is also what the pool supply changes by (what you get out), but we take the abs value of that which is why they are both positive

x/concentrated-liquidity/pool.go
czarcas7ic: Agreed this is not something we want long term, but my vs code was having a seizure every time I accidentally reintroduced an inf loop so wanted something to prevent this. Maybe we should introduce a max 100 loop counter for now, wdyt?

x/concentrated-liquidity/pool.go
czarcas7ic: These terms are actually sysnonymous, since the delta is what we are putting into and taking out of the pool. But agreed that tokenIn and tokenOut are just simpler to understand / better API

x/concentrated-liquidity/pool_test.go
czarcas7ic: can you explain what you mean by separating the tests out?

x/concentrated-liquidity/pool.go
czarcas7ic: Changed!

x/concentrated-liquidity/math.go
czarcas7ic: Discussed offline

x/concentrated-liquidity/pool_test.go
czarcas7ic: Changed the name

x/concentrated-liquidity/pool_test.go
mattverse: Nm, I will create a PR for making this TDD

x/concentrated-liquidity/pool.go
```diff
@@ -100,7 +100,7 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 	tokenOutDenom string,
 	swapFee sdk.Dec,
 	priceLimit sdk.Dec,
-	poolId uint64) (tokenIn, tokenOut sdk.Coin, updatedTick sdk.Int, updatedLiquidity sdk.Dec, err error) {
+	poolId uint64) (tokenIn, tokenOut sdk.Int, updatedTick sdk.Int, updatedLiquidity sdk.Dec, err error) {
```
mattverse, line 103: Extension from my previous Q: What's the reason for the change to sdk.Int?

IMO it should be sdk.Coin considering that existing CalcOutGiven in returns a coin and the point that the we would be expecting or constructing a coin any way in the caller of this method

x/concentrated-liquidity/pool.go
```diff
@@ -139,15 +139,17 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 		liquidity:                p.Liquidity,
 	}
 
+	// had to reintroduce the dust check, we need to figure out a cleaner way to do this
```
mattverse, line 142:
```suggestion
```

x/concentrated-liquidity/pool.go
```diff
@@ -164,42 +166,43 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
...
 			zeroForOne,
 		)
+		swapState.sqrtPrice = sqrtPrice
 
 		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(amountIn)
 		swapState.amountCalculated = swapState.amountCalculated.Add(amountOut)
 
 		// if we have moved to the next tick,
-		if nextSqrtPrice.Equal(sqrtPrice) {
+		if nextSqrtPrice.Equal(swapState.sqrtPrice) {
```
mattverse, line 175: I think this can stay the same? (use sqrtPrice)

x/concentrated-liquidity/pool.go
```diff
@@ -164,42 +166,43 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
...
 	if zeroForOne {
-		tokenIn = sdk.NewCoin(tokenInMin.Denom, amt0)
-		tokenOut = sdk.NewCoin(tokenOutDenom, amt1)
+		tokenIn = (swapState.amountCalculated).TruncateInt()
+		tokenOut = tokenAmountInAfterFee.Add(swapState.amountSpecifiedRemaining).TruncateInt()
 	} else {
-		tokenIn = sdk.NewCoin(tokenInMin.Denom, amt1)
-		tokenOut = sdk.NewCoin(tokenOutDenom, amt0)
+		tokenIn = tokenAmountInAfterFee.Add(swapState.amountSpecifiedRemaining).TruncateInt()
+		tokenOut = swapState.amountCalculated.TruncateInt()
```
mattverse, line 205: consider de-duplicating these logic by using a variable above

mattverse (commented):
Overall Main logic LGTM, I just have final round of reivews I wanted to adress / ask before merge

x/concentrated-liquidity/pool.go
czarcas7ic: they are synonymous now since we we now set swapState.sqrtPrice = sqrtPrice but I will change it back!

x/concentrated-liquidity/pool.go
czarcas7ic: Yeah I think I agree with you, thinking back to why I changed this in the first place I think it was for troubleshooting purposes. Will change back

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+		amountIn = calcAmount1Delta(liquidity, sqrtPriceTarget, sqrtPriceCurrent, true)
 	}
-
 	if amountRemaining.GTE(amountIn) {
 		sqrtPriceNext = sqrtPriceTarget
 	} else {
 		sqrtPriceNext = getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining, zeroForOne)
 	}
 
+	max := sqrtPriceNext.Equal(sqrtPriceTarget)
```
p0mvn, line 77: Please add comments about this

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+
 	if zeroForOne {
```
p0mvn, line 79: I suggest exploring the option of using separate methods for `zero` and `one`.

Also, exploring whether using a design pattern like strategy makes sense here

x/concentrated-liquidity/pool.go
```diff
@@ -187,19 +191,19 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
...
-		} else {
+		} else if !sqrtPriceStart.Equal(sqrtPrice) {
 			swapState.tick = priceToTick(sqrtPrice.Power(2))
 		}
 	}
 
-	amt0 := tokenAmountInAfterFee.Add(swapState.amountSpecifiedRemaining.Abs()).TruncateInt()
-	amt1 := swapState.amountCalculated.TruncateInt()
+	amt0 := tokenAmountInAfterFee.Add(swapState.amountSpecifiedRemaining).TruncateInt()
+	amt1 := (swapState.amountCalculated).TruncateInt()
```
p0mvn, line 200:
```suggestion
	amt1 := swapState.amountCalculated.TruncateInt()
```

p0mvn, line 200: Can we add comments to explain truncations here?

x/concentrated-liquidity/pool_test.go
```diff
@@ -12,50 +12,182 @@ import (
 func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 	ctx := s.Ctx
 
+	//
```
p0mvn, line 15: Let's convert to table-driven please

x/concentrated-liquidity/tick.go
```diff
@@ -87,7 +87,7 @@ func (k Keeper) NextInitializedTick(ctx sdk.Context, poolId uint64, tickIndex in
 	}
 
 	var iter db.Iterator
-	if zeroForOne {
+	if !zeroForOne {
```
p0mvn, line 90: I think we should also have separate methods here or a strategy design pattern instead of checking `zeroForOne` everywhere

p0mvn (changes requested):

x/concentrated-liquidity/tick.go
p0mvn: Reason: https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior

x/concentrated-liquidity/lp.go
```diff
@@ -64,15 +64,15 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 
 	// update pool state as well only if we're injecting liquidity at the current tick of the pool
 	if pool.CurrentTick.LT(sdk.NewInt(lowerTick)) {
-		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick).RoundInt()
+		amtDenom0 = calcAmount0Delta(liquidity, sqrtRatioLowerTick, sqrtRatioUpperTick, false).RoundInt()
```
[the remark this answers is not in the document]
czarcas7ic: As discussed offline, am going to revert the rounding until we can create a test case that actually causes an overflow

x/concentrated-liquidity/math.go
czarcas7ic: was just gas efficiency, removing now

x/concentrated-liquidity/lp.go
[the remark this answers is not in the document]
czarcas7ic: Further discussion offline, rounding both protects from inf loop as well as rounding up on amountIn to prevent removing more from the pool than expected.

x/concentrated-liquidity/lp.go
```diff
@@ -62,17 +62,25 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+		// we also update the pool liquidity since the virtual liquidity is modified by this position's creation
 	} else if pool.CurrentTick.LT(sdk.NewInt(upperTick)) {
-		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick).RoundInt()
-		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick).RoundInt()
+		amtDenom0 = calcAmount0Delta(liquidity, currentSqrtPrice, sqrtRatioUpperTick, false).RoundInt()
+		amtDenom1 = calcAmount1Delta(liquidity, currentSqrtPrice, sqrtRatioLowerTick, false).RoundInt()
 		pool.Liquidity = pool.Liquidity.Add(liquidity)
+
+		// outcome three: position is above current price
+		// tis means position is solely made up of asset1
```
p0mvn, line 80:
```suggestion
		// this means position is solely made up of asset1
```

p0mvn, line 80: nit: please consider moving comments inside their branches. Now it might look like this is referring to outcome two branch

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
-func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
+func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec {
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
 	mult := liq
+	// if calculating for amountIn, we round up
+	// if calculating for amountOut, we don't round
+	// this is to prevent removing more from the pool than expected due to rounding
```
p0mvn, line 44: Could you elaborate a bit on this please? Why would we remove more than expected from pool? Where exactly the difference leads? For example, offline we discussed an infinite loop. Where is that loop?

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
 // calcAmount1Delta = liq * (sqrtPriceB - sqrtPriceA)
-func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
+func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec {
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
+	// if calculating for amountIn, we round up
+	// if calculating for amountOut, we don't round
+	// this is to prevent removing more from the pool than expected due to rounding
```
p0mvn, line 62: ditto to expand on rounding

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,59 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
 		sqrtPriceNext = sqrtPriceTarget
 	} else {
 		sqrtPriceNext = getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining, zeroForOne)
 	}
 
 	if zeroForOne {
-		amountIn = calcAmount0Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent)
-		amountOut = calcAmount1Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent)
+		amountIn = calcAmount0Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent, true)
+		amountOut = calcAmount1Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent, false)
```
p0mvn, line 85: Why do we round up `amountIn` calculation here but round down `amountOut`? I think this deserves a comment

x/concentrated-liquidity/pool_test.go
czarcas7ic: Just merged Matts branch that did this!

x/concentrated-liquidity/math.go
czarcas7ic: elaborated on this further. this should also address your concern in math.go. If it doesn't, I don't have any other way I can further explain this so will have to get on a call to further hash out what we can say to make it clearer.

x/concentrated-liquidity/math.go
```diff
@@ -33,48 +33,67 @@ func liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
 	mult := liq
+	// if calculating for amountIn, we round up
+	// if calculating for amountOut, we don't round at all
+	// this is to prevent removing more from the pool than expected due to rounding
+	// example: we calculate 1000000.9999999 uusdc (~$1) amountIn and 2000000.999999 uosmo amountOut
+	// we would want the used to put in 1000001 uusdc rather than 1000000 uusdc to ensure we are charging enough for the amount they are removing
```
p0mvn, line 46:
```suggestion
	// we would want the user to put in 1000001 uusdc rather than 1000000 uusdc to ensure we are charging enough for the amount they are removing
```

x/concentrated-liquidity/math.go
p0mvn: This is perfect, thank you

p0mvn (approved):
Great work! Comments are super helpful.

I focused only on core logic and only skimmed the tests. Planning to review the tests later outside of the PR.

Also, as discussed offline, it would be great to investigate separately whether this approach would reduce the complexity from passing in `zeroForOne` booleans 
https://github.com/osmosis-labs/osmosis/issues/3218

PR #3207, opened 2022-11-02
fix(e2e): various e2e build issues breaking debugging on linux amd64

p0mvn:
Closes: #XXX

Fixed the following problems:
- tests folder excluded by dockerignore, disallowing Docker init image build
- init image being built for aarch64 when built from macs
   * mitigated by specifying platform
- print debug logs for `test-e2e-debug`
- use `Require()`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@ docs/
 networks/
 proto/
 scripts/
-tests/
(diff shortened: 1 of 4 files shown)


czarcas7ic (approved):
utACK

PR #3214, opened 2022-11-02
Add TWAP CLI

ValarDragon:

Add CLI query support for TWAP. I've eye ball-tested the values for correctness against imperator API. Ideally we add this logic to our querygen, or use proto-reflect from SDK to automate CLI boilerplate going for.

## Testing and Verifying

Manually tested

  - How is the feature or change documented? CLI help


x/twap/client/cli/query.go
```diff
@@ -0,0 +1,138 @@
...
+			baseDenom := strings.TrimSpace(args[1])
+
+			// <UNIX TIME PARSE>
+			startTime, err := parseUnixTime(args[2], "start time")
+			if err != nil {
+				return err
+			}
+
+			// END TIME PARSE: ONEOF {<UNIX TIME PARSE>, <DURATION>}
+			endTime, err := parseUnixTime(args[3], "end time")
```
mattverse, line 69:
```suggestion
			// try parsing in unix time, if failed try parsing in duration  
			endTime, err := parseUnixTime(args[3], "end time")
```

x/twap/client/cli/query.go
```diff
@@ -0,0 +1,138 @@
...
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand(GetQueryTwapCommand())
+
+	return cmd
+}
+
+// GetCmdAssetMultiplier returns multiplier of an asset by denom.
+func GetQueryTwapCommand() *cobra.Command {
```
mattverse, line 36: nit: don't know if its worth overhead, but would love to have some test cases on this since this includes lot of internal parsing. We could also make it a separate issue for this if u agree with the needs of testing for this

x/twap/client/cli/query.go
```diff
@@ -0,0 +1,138 @@
...
+	v, err := strconv.ParseUint(arg, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("could not parse %s as uint for field %s: %w", arg, fieldName, err)
+	}
+	return v, nil
+}
+
+func parseUnixTime(arg string, fieldName string) (time.Time, error) {
+	timeUnix, err := strconv.ParseInt(arg, 10, 64)
+	if err != nil {
```
mattverse, line 133: Maybe we should bring these into osmoutils at some point since we're doing the same thing in all cli files 🤔

mattverse (approved):
Overall LGTM, left some nit comments!

x/twap/client/cli/query.go
ValarDragon: yeah, I think we either make querygen create this, or we use the protoreflect automation tooling from SDK to remove all this code in the first place.

x/twap/client/cli/query.go
ValarDragon: Happy if someone else wants to write tests for this, I think of it as pretty low priority.

I'd much rather allocate those mental cycles to getting most CLI code killed / better testing structures. (I believe SDK has progress on better CLI testing structure)

PR #3219, opened 2022-11-02
Calculate channel value across all channels for native tokens

nicolaslara:
The channel value for native tokens shouldn't be only the value in escrow for that channel, but the value for *all* channels. This is similar to what we were doing before: getting the total supply, but focuses on the IBC supply of native tokens.


x/ibc-rate-limit/rate_limit.go
```diff
@@ -151,17 +152,19 @@ func GetLocalDenom(denom string) string {
...
-		escrowAddress := transfertypes.GetEscrowAddress(port, channel)
-		return bankKeeper.GetBalance(ctx, escrowAddress, denom).Amount
+	// For native tokens, obtain the balance held in escrow for all potential channels
+	channels := channelKeeper.GetAllChannels(ctx)
+	balance := sdk.NewInt(0)
+	for _, channel := range channels {
+		escrowAddress := transfertypes.GetEscrowAddress("transfer", channel.ChannelId)
+		balance = balance.Add(bankKeeper.GetBalance(ctx, escrowAddress, denom).Amount)
+
 	}
```
ValarDragon, line 168: I'm worried about this being too slow for native assets right now.

Do we want to comment out for native assets until we can get this into a pre-computed state entry?

nicolaslara: yeah, I was wondering about that too. 

We need the value to be calculated across all channels if we want to be able to filter the tokens for any channel (like we discussed). I can add the pre-computed state and a query for it, but it's a detour. 

We could also do the queries concurrently, but we still need to do one query per channel.

ValarDragon: Maybe we do native assets = bank supply, non-native = escrow account for now?

And then once we get an efficient state entry for the sum of IBC'd out native assets across all channels, we switch the go side?

ValarDragon: Made this issue in IBC-go to track this: https://github.com/cosmos/ibc-go/issues/2664

nicolaslara: That makes sense to me. I'll make the changes.

It's pretty much what we had before 😅, but should be easier to modify now since the tests are more flexible.

ValarDragon: sorry :(

nicolaslara: haha, no worries. Now we know what to do :)

x/ibc-rate-limit/contracts/rate-limiter/src/state.rs
```diff
@@ -243,7 +243,9 @@ fn calculate_channel_value(
             if denom.contains("ibc") {
                 channel_value + funds // Non-Native tokens get removed from the supply on send. Add that amount back
             } else {
-                channel_value - funds // Native tokens increase escrow amount on send. Remove that amount here
+                // The commented-out code bellow is what we want, but we're currently using the whole supply temporarily for efficiency. see rate_limit.go/CalculateChannelValue(..)
```
ValarDragon, line 246:
```suggestion
                // The commented-out code in the golang calculate channel value is what we want, but we're currently using the whole supply temporarily for efficiency. see rate_limit.go/CalculateChannelValue(..)
```

ValarDragon (approved):

Issue #3222, opened 2022-11-02
[x/gamm][stableswap]: Add default scaling factor multiplier


AlpinYukseloglu:
## Background

Currently default pools set scaling factors to 1, which risks having the scenario where a pool is accidentally created with low maximum capacity (this is hard to reverse if it goes unnoticed until this cap is hit). We should add a default multiplier to all scaling factors and apply this to default scaling factors as well to avoid this scenario.

## Suggested Design

- Apply sufficiently large multiplier to scaling factors to accommodate a minimum of 100B per asset assuming 6 units of token decimal precision

## Acceptance Criteria

All existing scaling factor tests pass

PR #3223, opened 2022-11-02
[x/gamm][stableswap]: Add default scaling factor multiplier, geometric error tolerance, and minimum post-scaled asset amount

AlpinYukseloglu:
Closes: #3222, #3254, #3255

This PR adds a multiplier to scaling factors that scales all scaling factors for pools, changes the error tolerance for our solver to be geometric with `10^-12` multiplicative tolerance, and adds a check that ensures post-scaled asset amounts cannot be less than 1 (to fix a bug where k would go to zero during calculations and the solver would do nothing)

**Note to reviewer:** most of the diff is updating pool tests to accommodate variable scaling factor multipliers

## Brief Changelog

- Add `ScalingFactorMultiplier` constant to `types.go`
- Apply multiplier at pool creation and in `SetScalingFactors` function
- Add basic guardrail to `validateScalingFactors` to ensure this multiplier is correctly applied everywhere
- Remove additive error tolerance and add multiplicative tolerance of `10^-12`
- Add check to `validatePoolLiquidity` that ensure the minimum post-scaled amount for any asset is at least 1

## Testing and Verifying

- Existing tests already thoroughly test scaling factor functionality and pass with these changes


x/gamm/types/constants.go
```diff
@@ -15,6 +15,9 @@ const (
 	BalancerGasFeeForSwap = 10_000
 
 	StableswapMaxScaledAmtPerAsset = 10_000_000_000
+	StableswapMinScaledAmtPerAsset = 1
+	// TODO: increase this to 1_000_000 (aligned with 6 decimal place tokens) and update tests accordingly
```
AlpinYukseloglu, line 19: Note: since most of our tests currently use < 1_000_000 tokens per asset, most of them would need to be refactored after this change

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -263,7 +266,9 @@ func solveCFMMBinarySearchMulti(xReserve, yReserve, wSumSquares, yIn osmomath.Bi
 	}
 
 	maxIterations := 256
-	errTolerance := osmoutils.ErrTolerance{AdditiveTolerance: sdk.OneInt(), MultiplicativeTolerance: sdk.Dec{}}
+
+	// we use a geometric error tolerance that guarantees approximately 10^-8 precision on outputs
```
ValarDragon, line 270:
```suggestion
	// we use a geometric error tolerance that guarantees approximately 10^-12 precision on outputs
```

Maybe move this to a constant in types?

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -287,7 +292,7 @@ func solveCFMMBinarySearchMulti(xReserve, yReserve, wSumSquares, yIn osmomath.Bi
 	}
 
 	xOut := xReserve.Sub(xEst)
-	if xOut.GTE(xReserve) {
+	if xOut.Abs().GTE(xReserve) {
```
ValarDragon, line 295: Lets add a comment as to why its abs?

ValarDragon: Lets add a comment that this ensures that:
* We can't swap more than the pool supply out
* A swap won't double the input asset supply

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -414,8 +428,19 @@ func validatePoolLiquidity(liquidity sdk.Coins, scalingFactors []uint64) error {
 		scaledAmount := asset.Amount.Quo(sdk.NewInt(int64(scalingFactors[i])))
 		if scaledAmount.GT(sdk.NewInt(types.StableswapMaxScaledAmtPerAsset)) {
 			return types.ErrHitMaxScaledAssets
+		} else if asset.Amount.Quo(sdk.NewInt(int64(scalingFactors[i]))).LT(sdk.NewInt(types.StableswapMinScaledAmtPerAsset)) {
```
ValarDragon, line 431:
```suggestion
		} else if scaledAmount.LT(sdk.NewInt(types.StableswapMinScaledAmtPerAsset)) {
```

ValarDragon (approved):

x/gamm/pool-models/stableswap/amm.go
AlpinYukseloglu, line 295:
```suggestion
        // we check the absolute value of the output against the xReserve amount to ensure that a swap cannot output more than the output token's pool supply and cannot more than double the input token's pool supply
	if xOut.Abs().GTE(xReserve) {
```

PR #3229, opened 2022-11-03
IBC: rate limit native tokens on all channels

nicolaslara:
This PR extends https://github.com/osmosis-labs/osmosis/pull/3219 to unify all quotas for native denoms into one: the "any" channel. 

When sending or receiving a native denom, the contract is not queried for the specific ibc channel used, but instead the catch-all "any", which means transfers through any channel will increase or reduce the quota. 

As an example, if someone sends 1 osmo to juno and 1 osmo to the akash, then the used quota for osmo will be 2. If the max quota setup for the channel "any" was 2, then no other osmo sends will be allowed. 

Right now only the any channel is checked. This could be extended to check both "any" and the channel id. That way we could add different limits to each channel beyond the global limit. Since we're planning on moving some of this logic to the contract I don't think it makes sense to do it on this PR, but we can discuss it in case we want it later.


x/ibc-rate-limit/ibc_module.go
```diff
@@ -134,15 +135,22 @@ func (im *IBCModule) OnRecvPacket(
...
 
-	channelValue := im.ics4Middleware.CalculateChannelValue(ctx, denom, packet)
+	channelValue := im.ics4Middleware.CalculateChannelValue(ctx, denom)
+
+	channel := packet.GetSourceChannel()
+	if !strings.HasPrefix(denom, "ibc/") {
+		// For native channels, we want to have only one limit across all channels. We can use a quota for
+		//channel "any" for this
+		channel = "any"
+	}
```
ValarDragon, line 145: Shouldn't this be for all denoms now?

nicolaslara: I thought we wanted to do this only for native tokens. IBC tokens are mostly limited to a channel anyway, but we could end up in the case where someone "parks" the tokens on another network through a different channel. 

If we want the all tokens to have an "any" channel, then we can probably remove the channel abstraction altogether (though it might make sense to keep it in case we want to add it back in the future)

nicolaslara: I've moved this logic to the contract. Now the actual channel and "any" are checked independently. This way we can get one behaviour or the other depending on how we configure the quotas. Marking as resolved.

nicolaslara:
This PR can probably be closed in favor of https://github.com/osmosis-labs/osmosis/pull/3234. The only point of having them separately was to be able to develop and review it in steps

Issue #3218, opened 2022-11-02
[CL][Swapping]: refactor to use Strategy design pattern instead of `zeroForOne` boolean


p0mvn:
## Background

Currently, we use `zeroForOne` boolean within swapping logic that leads to code like this:

```go
func getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec, zeroForOne bool) (sqrtPriceNext sdk.Dec) {
	if zeroForOne {
		sqrtPriceNext = getNextSqrtPriceFromAmount0RoundingUp(sqrtPriceCurrent, liquidity, amountRemaining)
	} else {
		sqrtPriceNext = getNextSqrtPriceFromAmount1RoundingDown(sqrtPriceCurrent, liquidity, amountRemaining)
	}
	return sqrtPriceNext
}
```

This reduces readability and increases the likelihood of making a mistake by providing a wrong flag. Our functions and methods should be cohesive.

## Suggested Design

Instead, we should use a strategy design pattern. By doing so, we can initialize a specific "strategy" either `zeroForOne` or `oneForZero` at the beginning of the execution flow of the swap.

The concrete strategies have the same interfaces but different implementations. As a result, we should be able to call each strategy's implementation and avoid passing in the `zeroForOne` boolean throughout the codebase.

The previous example would turn into:

```go
type swapStrategy interface {
    getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext sdk.Dec)
}
...

newSwapStrategy(zeroForOne bool) swapStrategy {
    if zeroForOne {
        return &zeroForOneStrategy{}
    }
    return &oneForZeroStrategy{}
}

func (s *zeroForOneStrategy) getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext sdk.Dec) {
	return getNextSqrtPriceFromAmount0RoundingUp(sqrtPriceCurrent, liquidity, amountRemaining)
}

func (s *oneForZeroStrategy) getNextSqrtPriceFromInput(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext sdk.Dec) {
	return getNextSqrtPriceFromAmount1RoundingDown(sqrtPriceCurrent, liquidity, amountRemaining)
}
```
- note that the interface and each strategy will have more methods defined

[...]

PR #3232, opened 2022-11-03
Use strategy pattern in CL swaps

hieuvubk:
Closes: #3218 

Referring from suggesting design in the original issue, strategy pattern is integrated. 

## Testing and Verifying

All test passed


p0mvn:
Nice @hieuvubk ! Can you follow this approach to remove any remaining `zeroForOne` parameter propagation?

That is, refactor any remaining methods and functions that have `zeroForOne` as a parameter

hieuvubk:
> Nice @hieuvubk ! Can you please follow this approach to remove any remaining `zeroForOne` parameter propagation?
> 
> That is, refactor any remaining methods and functions that have `zeroForOne` as a parameter



> Nice @hieuvubk ! Can you please follow this approach to remove any remaining `zeroForOne` parameter propagation?
> 
> That is, refactor any remaining methods and functions that have `zeroForOne` as a parameter

Yes, I will

hieuvubk:
@p0mvn Hi I refactor `computeSwapStep` to the strategy pattern. There is only func `NextInitializedTick` takes `zeroForOne` as param. But I don't think we should refactor this func, what do u think pls lmk.

p0mvn:
hey @hieuvubk .

What do you think about refactoring logic here as well?

https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L173-L191

We can encapsulate this logic into function `setLiquidityDeltaSign`:
https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L178-L180

and this logic into `setNextTick`:
https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L187-L191

Let me know what you think

hieuvubk:
> hey @hieuvubk .
> 
> What do you think about refactoring logic here as well?
> 
> https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L173-L191
> 
> We can encapsulate this logic into function `setLiquidityDeltaSign`:
> 
> https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L178-L180
> 
> and this logic into `setNextTick`:
> 
> https://github.com/osmosis-labs/osmosis/blob/a77ddc12c05f836bd1bd970beefb0133ecadd4fc/x/concentrated-liquidity/pool.go#L187-L191
> 
> Let me know what you think

Thanks for the workaround, I will follow it. 
The logic of the two functions `CalcInAmtGivenIn` and `CalcInAmtGivenOut` is flipped. So I add a bool variable to determine the `In` or `Out` case.
I'm not sure if removing a bool variable and adding another violates what we want. So if u have any ideas pls let me know. Thankssss!!

x/concentrated-liquidity/math.go
```diff
@@ -76,36 +98,66 @@ func calcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
+		return deltaLiquidity.Neg()
+	}
+	return deltaLiquidity
+}
+
+func (s *zeroForOneStrategy) setNextTick(nextTick int64, swapIn bool) sdk.Int {
+	if swapIn {
+		return sdk.NewInt(nextTick - 1)
+	}
+	return sdk.NewInt(nextTick)
```
p0mvn, line 153: Why do we need to pass in the `swapIn` flag? Can we execute one or the other depending on the strategy?

The point of using the strategy pattern is to be able to avoid passing in these flags

hieuvubk: Yes I see, thanks you

faddat:
I think there's some upstream changes here that have made this one kinda-not-fresh. 

I did a merge, but I'm rather sure that it is not straight yet, so please feel free to revert.

p0mvn (approved):
Thank you @hieuvubk 

This does show me that the strategy pattern is an improvement. I think we can continue with integrating it into tick logic in follow-up PRs.

I would also like to move each strategy into its own file. That is not blocking the work here though

hieuvubk:
> Thank you @hieuvubk
> 
> This does show me that the strategy pattern is an improvement. I think we can continue with integrating it into tick logic in follow-up PRs.
> 
> I would also like to move each strategy into its own file. That is not blocking the work here though

Yes  I'll create an issue about it, should move strategy to `concentrated-liquidity/internal/strategy` & refactor `NextInitializedTick`

PR #3235, opened 2022-11-03
CL: test expansion

czarcas7ic:
Closes: #XXX

Through this PR's test expansion, we realized that we are not reaching appropriate precision. This PR introduces wolframalpha links to the math behind each test. If a test is off, the correct result is commented beside the result that currently is returning

## Brief Changelog

- Changes TickToSqrtPrice from a keeper method to a regular func
- Changes TickToSqrtPrice to no longer use approxSqrt and instead just use (1.0001)^(n/2) formula
- Adds math tests
- Adds tick tests


x/concentrated-liquidity/tick.go
```diff
@@ -13,13 +13,10 @@ import (
 )
 
 // tickToSqrtPrice takes the tick index and returns the corresponding sqrt of the price
-func (k Keeper) tickToSqrtPrice(tickIndex sdk.Int) (sdk.Dec, error) {
-	price, err := sdk.NewDecWithPrec(10001, 4).Power(tickIndex.Uint64()).ApproxSqrt()
-	if err != nil {
-		return sdk.Dec{}, err
-	}
+func tickToSqrtPrice(tickIndex sdk.Int) sdk.Dec {
+	sqrtPrice := sdk.NewDecWithPrec(10001, 4).Power(tickIndex.Uint64() / 2)
```
p0mvn, line 17: Q:

$$2^{3 /2} = 2^1.5 = 1.6817928305074292 $$

but with our `Power()` function it would be:

$$2^{3/2} = 2^1 = 2$$

Would it not? Am I missing anything?

p0mvn, line 17: Q:

`2^{3 /2} = 2^1.5 = 1.6817928305074292`

but with our `Power(...)` function it would be:

`2^{3/2} = 2^1 = 2`

Would it not? Am I missing anything?

x/concentrated-liquidity/tick_test.go
```diff
@@ -132,26 +132,28 @@ func (suite *KeeperTestSuite) TestTickToSqrtPrice() {
...
 			sdk.NewInt(85176),
-			"70.710004849206351867",
-			false,
+			"70.710004849206351861", // 70.710004849206120647
+			// https://www.wolframalpha.com/input?i2d=true&i=Power%5B1.0001%2CDivide%5B85176%2C2%5D%5D
+		},
+		{
+			"happy path 2",
+			sdk.NewInt(86129),
+			"74.157016832801088770", // 74.160724590950847046
```
p0mvn, line 145: I think the problem unfortunately exists for odd powers.

From python:

```
>>> 1.0001**(43064.5)
74.16072459091568
>>> 1.0001**(43064)
74.15701683276568
```

It seems that here we round down when dividing. That's where we lose a lot of precision

czarcas7ic: Good catch, we for sure have to do something about the odd ones at the very least, would like to improve precision all around if at all possible

p0mvn: Maybe for odd we use `ApproxRoot` for now?

I can look into implementing a more suitable root or power functions later

x/concentrated-liquidity/tick.go
czarcas7ic: Yep, you are right, I think I am going to just change it back for now instead of even using this and odd using ApproxSqrt. Good catch

x/concentrated-liquidity/tick_test.go
czarcas7ic: Ended up going back to using ApproxRoot for both for now

x/concentrated-liquidity/pool_test.go
```diff
@@ -55,10 +55,9 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 			tokenOutDenom: "eth",
 			priceLimit:    sdk.NewDec(5004),
 			// we expect to put 42 usdc in and in return get .008398 eth back
-			// due to limited liquidity, we actually put in 41.99 usdc and in return get .008396 eth back
-			expectedTokenIn:  sdk.NewCoin("usdc", sdk.NewInt(41999999)),
-			expectedTokenOut: sdk.NewCoin("eth", sdk.NewInt(8396)),
-			expectedTick:     sdk.NewInt(85184),
+			expectedTokenIn:  sdk.NewCoin("usdc", sdk.NewInt(41999999)), // TODO: should be 42
+			expectedTokenOut: sdk.NewCoin("eth", sdk.NewInt(8396)),      // TODO: should be 0.008371593947078467
+			expectedTick:     sdk.NewInt(85184),                         // TODO: should be 85183
```
mattverse, line 60: Did we ever get to a wolframalpha for this or is this something that we should do in the future as well?

czarcas7ic: The pool tests might be a little harder to replicate in wolfram because of the iterations we go through via the currently existing positions. There may be a simple way I am not seeing though

x/concentrated-liquidity/math.go
```diff
@@ -47,9 +47,9 @@ func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	// additionally, without rounding, there exists cases where the swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) for loop within
 	// the CalcOut/In functions never actually reach zero due to dust that would have never gotten counted towards the amount (numbers after the 10^6 place)
 	if roundUp {
-		return ((mult.Mul(diff)).QuoRoundUp(sqrtPriceB).QuoRoundUp(sqrtPriceA)).Ceil()
+		return liq.Mul((diff.QuoRoundUp(denom))).Ceil()
```
p0mvn, line 50:
```suggestion
		return liq.Mul(diff.QuoRoundUp(denom)).Ceil()
```

x/concentrated-liquidity/math.go
```diff
@@ -47,9 +47,9 @@ func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
 	}
-	return (mult.Mul(diff)).Quo(sqrtPriceB).Quo(sqrtPriceA)
+	return liq.Mul((diff.Quo(denom)))
```
p0mvn, line 52:
```suggestion
	return liq.Mul(diff.Quo(denom))
```

x/concentrated-liquidity/pool_test.go
```diff
@@ -203,96 +202,6 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 	}
 }
 
-// func (s *KeeperTestSuite) TestCalcInAmtGivenOut() {
```
p0mvn, line 206: Is the plan to reintroduce these later when implemented accurately?

p0mvn (commented):
LGTM but would like to spend some time looking at the tests in more detail

x/concentrated-liquidity/pool_test.go
czarcas7ic: Yeah, just found it to be very distracting / cumbersome when we don't even have Calc In implemented yet

mattverse: What do you think of leaving the test cases at the bottom? I think its still good to leave it as we're going to use them any way, and to remove any duplicated / unnecessary workload in the future

mattverse (approved):
LGTM, nice work on all the calculations! 

Noticed that alot of the small calculation diffs comes from the 10^6 when we're providing some numbers as inputs lol

x/concentrated-liquidity/pool_test.go
```diff
@@ -52,9 +52,14 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 			tokenIn:       sdk.NewCoin("usdc", sdk.NewInt(42000000)),
 			tokenOutDenom: "eth",
 			priceLimit:    sdk.NewDec(5004),
-			// we expect to put 42 usdc in and in return get .008398 eth back
-			// due to truncations and precision loss, we actually put in 41.99 usdc and in return get .008396 eth back
-			expectedTokenIn:  sdk.NewCoin("usdc", sdk.NewInt(41999999)),
+			// params
+			// liquidity: 		 1517818840.967515822610790519
+			// sqrtPriceNext:    70.738349405152439867 which is 5003.914076565430543175
```
mattverse, line 57: Can we add a wolfram alpha link to how we got the sqrt price next?

x/concentrated-liquidity/math.go
```diff
@@ -47,9 +47,9 @@ func calcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	// additionally, without rounding, there exists cases where the swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) for loop within
 	// the CalcOut/In functions never actually reach zero due to dust that would have never gotten counted towards the amount (numbers after the 10^6 place)
 	if roundUp {
-		return ((mult.Mul(diff)).QuoRoundUp(sqrtPriceB).QuoRoundUp(sqrtPriceA)).Ceil()
+		return liq.Mul(diff.Quo(denom)).Ceil()
```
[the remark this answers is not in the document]
czarcas7ic: The change here was we were essentially rounding up twice when we should only do this with Ceil

x/concentrated-liquidity/math.go
```diff
@@ -89,10 +90,10 @@ func computeSwapStep(sqrtPriceCurrent, sqrtPriceTarget, liquidity, amountRemaini
 	}
 
 	if zeroForOne {
-		amountIn = calcAmount0Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent, true)
+		amountIn = calcAmount0Delta(liquidity, sqrtPriceNext, sqrtPriceCurrent, false)
```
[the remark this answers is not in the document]
czarcas7ic: So basically what I am trying to do is phase out the need for rounding up by only rounding at the very end when we return amt0 and amt1 at the end of the CalcFunc. For now, I say we leave these here though, because its possible we will end up needing them and I would rather not unimplement just to reimplement when we can just set to false for now. But agreed, lets aim to deprecate this

x/concentrated-liquidity/lp_test.go
```diff
@@ -5,52 +5,63 @@ import (
...
+			s.SetupTest()
 
-	s.App.ConcentratedLiquidityKeeper.CreateNewConcentratedLiquidityPool(s.Ctx, poolId, denom0, denom1, currentSqrtP, currentTick)
+			s.App.ConcentratedLiquidityKeeper.CreateNewConcentratedLiquidityPool(s.Ctx, tc.poolId, denom0, denom1, tc.currentSqrtP, tc.currentTick)
 
-	asset0, asset1, _, err := s.App.ConcentratedLiquidityKeeper.CreatePosition(s.Ctx, poolId, s.TestAccs[0], amount0Desired, amount1Desired, sdk.ZeroInt(), sdk.ZeroInt(), lowerTick, upperTick)
-	s.Require().NoError(err)
-	s.Require().Equal(amount0Desired.String(), asset0.String())
-	s.Require().Equal(amount1Desired.String(), asset1.String())
+			asset0, asset1, liquidityCreated, err := s.App.ConcentratedLiquidityKeeper.CreatePosition(s.Ctx, tc.poolId, s.TestAccs[0], tc.amount0Desired, tc.amount1Desired, sdk.ZeroInt(), sdk.ZeroInt(), tc.lowerTick, tc.upperTick)
```
p0mvn, line 43: Why is the returned liquidity (`liquidityCreated`) a Dec and not an int?

If we really want Dec, should we assert the exact decimal amount?

p0mvn: Looking at the way it is stored in state, it is an `Int`. So maybe we convert the return here to an `Int` as well?

x/concentrated-liquidity/math_test.go
```diff
@@ -7,30 +7,28 @@ import (
...
+			expectedLiquidity: "1517882343.751510418088349649",
 		},
 	}
 
-	for _, tc := range testCases {
+	for name, tc := range testCases {
 		tc := tc
 
-		suite.Run(tc.name, func() {
+		suite.Run(name, func() {
```
p0mvn, line 31: Consider adding comments about what is happening inside `getLiquidityFromAmounts`.

x/concentrated-liquidity/math_test.go
```diff
@@ -79,27 +75,25 @@ func (suite *KeeperTestSuite) TestLiquidity1() {
...
-			sdk.MustNewDecFromStr("70.710678"),
-			sdk.MustNewDecFromStr("74.161984"),
-			sdk.NewInt(1),
-			"1519.437618821730672389",
-			// https://www.wolframalpha.com/input?i=1+*+%2870.710678+*+74.161984%29+%2F+%2874.161984+-+70.710678%29
+		"happy path": {
+			currentSqrtP:      sdk.MustNewDecFromStr("70.710678118654752440"), // 5000
+			sqrtPHigh:         sdk.MustNewDecFromStr("74.161984870956629487"), // 5500
+			amount0Desired:    sdk.NewInt(1000000),
+			expectedLiquidity: "1519437308.014768571721000000", // TODO: should be 1519437308.014768571720923239
```
p0mvn, line 88: Convert `liquidity0` to:

```
func liquidity0(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
	}
	product := sqrtPriceA.Mul(sqrtPriceB)
	return amount.ToDec().Mul(product).Quo(diff)
}
```

That should fix it. `Quo` loses precision. As a result, if we apply it as the last operation, we aren't going to lose as much

x/concentrated-liquidity/math_test.go
```diff
@@ -180,29 +170,27 @@ func (suite *KeeperTestSuite) TestGetNextSqrtPriceFromAmount1RoundingDown() {
...
-			sdk.NewDec(1377927219),
-			sdk.MustNewDecFromStr("70.710678"),
-			sdk.MustNewDecFromStr("74.161984"),
-			"906866",
-			// https://www.wolframalpha.com/input?i=%281377927219+*+%2874.161984+-+70.710678+%29%29+%2F+%2870.710678+*+74.161984%29
+		"happy path": {
+			liquidity:       sdk.MustNewDecFromStr("1517882343.751510418088349649"), // we use the smaller liquidity between liq0 and liq1
+			sqrtPCurrent:    sdk.MustNewDecFromStr("70.710678118654752440"),         // 5000
+			sqrtPUpper:      sdk.MustNewDecFromStr("74.161984870956629487"),         // 5500
+			amount0Expected: "998976.618347426747968399",                            // TODO: should be 998976.618347426388356630
```
p0mvn, line 183: Please check my previous comment to see if that fixes it here as well

x/concentrated-liquidity/pool.go
```diff
@@ -108,99 +111,125 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
...
+		amountSpecifiedRemaining: tokenAmountInAfterFee, // tokenIn
+		amountCalculated:         sdk.ZeroDec(),         // tokenOut
 		sqrtPrice:                curSqrtPrice,
 		tick:                     priceToTick(curSqrtPrice.Power(2)),
 		liquidity:                p.Liquidity,
 	}
 
-	for swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) && !swapState.sqrtPrice.Equal(sqrtPriceLimit) {
+	// iterate and update swapState until we swap all tokenIn or we reach the specific sqrtPriceLimit
+	for swapState.amountSpecifiedRemaining.GT(sdk.MustNewDecFromStr("0.0000001")) && !swapState.sqrtPrice.Equal(sqrtPriceLimit) {
```
p0mvn, line 156: let's store `sdk.MustNewDecFromStr("0.0000001")` as a global variable to avoid initialization every iteration or every time the swap is called

p0mvn: why do we even have this value and not zero?

czarcas7ic: Imagine we are running through the for loop and we end up with an amount remaining of 0.000000000000456459. This is completely insignificant because we end up rounding at 10^6, and when we try to calculate an amountIn/amountOut for this value, its too small to actually generate anything, so we infinite loop on this

This may not be the most elegant way to prevent this, but it works for now

p0mvn: I see, thanks. Let's add a comment explaining this please with a TODO to come back to it

x/concentrated-liquidity/lp_test.go
czarcas7ic: I ended up actually changing the protos to Decs in place of Ints

x/concentrated-liquidity/math_test.go
czarcas7ic: As discussed offline, your suggestion fixed 1/2 TODOs, this one was not fixed

PR #3253, opened 2022-11-05
Fix Spot Price

mattverse:

This PR fixes the current spot price API by switching the orders of the parameters for base asset and quote asset. This is done in both `PoolI` level and the pool model API level.

This PR also addresses the issue for the reversed accumulator stores and spot price store entry in the twap module, and this is done by simply switching the names of each state entry. This allows us to successfully have the correct states with the correct names without having the need to do separate migration for the existing states.

## Brief Changelog
- Fix Spot price to return with correct value given base asset and quote asset

## Testing and Verifying
Existing tests has been fixed if necessary, also added a new integration test case in the twap module.

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -48,6 +48,7 @@ func (s *KeeperTestHelper) PrepareBalancerPool() uint64 {
 		ExitFee: sdk.NewDec(0),
 	})
 
+	// note to reviewers: please double check if these are correct
 	spotPrice, err := s.App.GAMMKeeper.CalculateSpotPrice(s.Ctx, poolId, "foo", "bar")
 	s.NoError(err)
 	s.Equal(sdk.NewDec(2).String(), spotPrice.String())
--- a/proto/osmosis/twap/v1beta1/twap_record.proto
+++ b/proto/osmosis/twap/v1beta1/twap_record.proto
@@ -37,20 +37,20 @@ message TwapRecord {
 
   // We store the last spot prices in the struct, so that we can interpolate
   // accumulator values for times between when accumulator records are stored.
-  string p0_last_spot_price = 6 [
+  string p1_last_spot_price = 6 [
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
-  string p1_last_spot_price = 7 [
+  string p0_last_spot_price = 7 [
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
 
-  string p0_arithmetic_twap_accumulator = 8 [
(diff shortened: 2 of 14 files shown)


proto/osmosis/twap/v1beta1/twap_record.proto
```diff
@@ -37,20 +37,20 @@ message TwapRecord {
 
   // We store the last spot prices in the struct, so that we can interpolate
   // accumulator values for times between when accumulator records are stored.
-  string p0_last_spot_price = 6 [
```
mattverse, line 40: Change of state entry names to avoid secondary migrations in upgrade

x/gamm/pool-models/balancer/pool.go
```diff
@@ -625,10 +625,10 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAsset, quoteAsset string) (spotPric
 		return sdk.Dec{}, errors.New("pool is misconfigured, got 0 weight")
 	}
 
-	// spot_price = (Base_supply / Weight_base) / (Quote_supply / Weight_quote)
-	// spot_price = (weight_quote / weight_base) * (base_supply / quote_supply)
-	invWeightRatio := quote.Weight.ToDec().Quo(base.Weight.ToDec())
-	supplyRatio := base.Token.Amount.ToDec().Quo(quote.Token.Amount.ToDec())
+	// spot_price = (Quote_supply / Weight_quote) / (base_supply / Weight_base)
```
mattverse, line 628: Note to reviewers: Please double check if this equation is correct

app/apptesting/gamm.go
```diff
@@ -48,6 +48,7 @@ func (s *KeeperTestHelper) PrepareBalancerPool() uint64 {
 		ExitFee: sdk.NewDec(0),
 	})
 
+	// note to reviewers: please double check if these are correct
```
mattverse, line 51: Note to reviewers: Please confirm that these values are correct

proto/osmosis/twap/v1beta1/twap_record.proto
ValarDragon: Niec, this LGTM

x/gamm/pool-models/balancer/pool.go
ValarDragon: This LGTM, thoughts on adding a simple example in the code comment?

ValarDragon:
We need to break queries to get the expected, in a second PR right?

This one should be backportable?

mattverse:
> We need to break queries to get the expected, in a second PR right?
>

Yes, that's the current plan


> This one should be backportable?

This should be query breaking, but not state breaking need double confirmation on this

ValarDragon:
Hrmm, my suspition is that this relates to twap variable name change

czarcas7ic:
This should fix e2e, the logic was flipped previously to account for wrong var names, this flips it back

Issue #2579, opened 2022-09-02
Validator set preferences module


ValarDragon:
## Background

We should make a new module, for validator set preferences. 

The goal is to give users & contracts a better UX for staking to a set of validators. E.g. a one click button, that stakes to multiple people. Then a user can set (or realistically a front-end provides) a list of recommended defaults. (E.g. governooors, relayers, chain/stack contributors, etc) This can technically be done off-chain today with front-ends, having preference lists stored locally. Having this on chain:
- eases front-end code burden that is already very high in cosmos
- reduces front-end lock-in, giving you the same UX across multiple front-ends

What this module should do is:
- Allow an address to set a list of {val addr, weight} in state, called their validator set preference
- Give users a single message to stake X osmo, according to their validator set preference distribution
- Give users a single message to claim rewards from everyone on their preference list (or should it be all their staking positions)
- If a user does not have a validator set preference list set, and has staked, make their preference list default to their current staking distribution
  - Expose a query for this
- If a user has no preference list and no staking, then return error for messages

## Suggested design

couple minute initial sketch of a proto for this
```proto
message ValidatorPreference {
  string val_oper_address = 1 [ (gogoproto.moretags) = "yaml:\"address\"" ];
  // two options for how we build weight.
  // Either as a decimal between 0 and 1, and they all sum to 1
  // OR normal int, and percentage to each validator is int/total weight
  string weight = 2 [
    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
    (gogoproto.moretags) = "yaml:\"autostake_rate\"",
    (gogoproto.nullable) = false
  ];
}

message ValidatorSetPreference {
  repeated ValidatorPreference preferences = 1 [ (gogoproto.nullable) = false ];
}
```

## Acceptance Criteria

- [ ] Above listed features built out
- [ ] We follow a module standard similar to x/TWAP (we should do so for new modules going forward)
- [ ] We document what rounding behavior for splitting stake amounts across validators were taking. Any particular decision is fine, just need to document / be clear about what we did

PR #3260, opened 2022-11-07
[ValSet-Pref] Add Msgs for Delegate and Undelegate Tokens

stackman27:
Part of : https://github.com/osmosis-labs/osmosis/issues/2579

This message `MsgDelegateToValidatorSet` allows delegators to delegate to a set of defined validator set.  For ex: delegate 10osmo with validator-set {ValA -> 0.5, ValB -> 0.3, ValC -> 0.2} our delegate logic would attempt to delegate 5osmo to A , 2osmo to B, 3osmo to C 

## Brief Changelog
n/a

## Testing and Verifying

Added tests for TestDelegateToValidatorSet


x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,45 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
...
+	for _, val := range existingSet.Preferences {
+		_, validator, err := k.getValAddrAndVal(ctx, val.ValOperAddress)
+		if err != nil {
+			return err
+		}
+
+		// tokenAmt takes the amount to delegate, calculated by {val_distribution_weight * tokenAmt}
+		// amountToDelegate tokenAmt takes the smallest denom so, 1 usmo = 10^6 osmo
+		tokenAmt := val.Weight.Mul(coin.Amount.ToDec())
+		amountToDelegate := tokenAmt.Mul(sdk.NewDec(1000000)).RoundInt()
```
stackman27, line 58: Any suggestion on how i can retrieve this multiplier in a better way? I know in superfluid we store the existing multiplier (https://github.com/osmosis-labs/osmosis/blob/main/x/superfluid/keeper/twap_price.go#L36)

stackman27: Resolved! No need to multiply manually, expect everything to be in smallest denom

faddat:
why?

not that I have any issue with it, it seems cool,  but curious on thinking here

stackman27:
> why?
> 
> not that I have any issue with it, it seems cool, but curious on thinking here

Hi @faddat! This feature can technically be done off-chain today with front-ends, having preference lists stored locally, but Having this on chain:
1. eases front-end code burden that is already very high in cosmos
2. reduces front-end lock-in, giving you the same UX across multiple front-ends
3. Opens door for features such as auto-staking

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+	tests := []struct {
+		name           string
+		delegator      sdk.AccAddress
+		coin           sdk.Coin
+		expectedShares []sdk.Dec
+		expectPass     bool
+		valSetExists   bool
+	}{
+		{
+			name:           "Delegate to valid validators!",
```
czarcas7ic, line 143:
```suggestion
			name:           "Delegate to valid validators",
```

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+		},
+		{
+			name:           "Delegate Decimal Amounts",
+			delegator:      sdk.AccAddress([]byte("addr2---------------")),
+			coin:           sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(25_000_000)),                       // amount to delegate
+			expectedShares: []sdk.Dec{sdk.NewDec(12_500_000), sdk.NewDec(7_500_000), sdk.NewDec(5_000_000)}, // expected shares after delegation
+			expectPass:     true,
+		},
+		{
+			name:       "User doesnot have enough tokens to stake",
```
czarcas7ic, line 165:
```suggestion
			name:       "User does not have enough tokens to stake",
```

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+		coin           sdk.Coin
+		expectedShares []sdk.Dec
+		expectPass     bool
+		valSetExists   bool
+	}{
+		{
+			name:           "Delegate to valid validators!",
+			delegator:      sdk.AccAddress([]byte("addr1---------------")),
+			coin:           sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(10_000_000)),                      // amount to delegate
+			expectedShares: []sdk.Dec{sdk.NewDec(5_000_000), sdk.NewDec(3_000_000), sdk.NewDec(2_000_000)}, // expected shares after delegation
```
czarcas7ic, line 146: Just fyi this abstraction made me read it twice before I got what was happening. Maybe in the comments just explicitly state in the comments that test's val set is defined as 50%,30%,and 20% instead of inferring it :)

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+			_, err := msgServer.DelegateToValidatorSet(c, types.NewMsgDelegateToValidatorSet(test.delegator, test.coin))
+			if test.expectPass {
+				suite.Require().NoError(err)
+
+				// check if the expectedShares matches after delegation
+				for i, val := range preferences {
+					valAddr, err := sdk.ValAddressFromBech32(val.ValOperAddress)
+					suite.Require().NoError(err)
+
+					//gurantees that the delegator exist because we check it in DelegateToValidatorSet
```
czarcas7ic, line 195:
```suggestion
					// guarantees that the delegator exists because we check it in DelegateToValidatorSet
```

x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,44 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
 	return nil
 }
 
+// TODO MAYBE: Check if there are any banned assets and ways to handle them
```
czarcas7ic, line 32: Mmm, idk if this TODO is worth leaving in the code base. Discussion should happen first with community before considering any implementation

x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,44 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
...
+		// expect tokenAmt to be in the smallest denom for ex, 1 usmo = 10^6 osmo
+		tokenAmt := val.Weight.Mul(coin.Amount.ToDec()).RoundInt()
+
+		// TODO: What happens here if validator is jailed, tombstoned, or unbonding
+		// Delegate the unbonded tokens
+		_, err = k.stakingKeeper.Delegate(ctx, delegator, tokenAmt, stakingtypes.Unbonded, validator, true)
+		if err != nil {
+			return err
+		}
+	}
```
czarcas7ic, line 65: The only issue I see with this is if delegating to a tombstoned validator produces an error. If this is the case, If we are halfway through the loop and we hit a tombstoned validator, it seems like we would error and the remainder of the delegation doesn't get delegated. There should be some kind of contract here that we either delegate all or none. Should check if delegating to tombstoned validators produces error. If it does, we need to check if its tombstoned prior to this. If not, can probably leave as is.

stackman27: hmm one option would be to loop through the val-set before delegation and check if there are any jailed / invalid validators or not using `[IsValidatorJailed](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/validator.go#L480)`  or [GetValidator](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/validator.go#L15) 

And if there is any jailed validator we DONOT delegate anything. lmk what you think

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+					del, _ := suite.App.StakingKeeper.GetDelegation(suite.Ctx, test.delegator, valAddr)
+					suite.Require().Equal(del.Shares, test.expectedShares[i])
```
mattverse, line 197: Can we add a check on user balance and confirmed that it has been reduced as well?

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+	preferences := suite.PrepareDelegateToValidatorSet()
+
+	tests := []struct {
+		name           string
+		delegator      sdk.AccAddress
+		coin           sdk.Coin
+		expectedShares []sdk.Dec
+		expectPass     bool
+		valSetExists   bool
+	}{
```
mattverse, line 141: would it be possible to add a test case that deals delegating coins that wouldn't be calculated perfectly under the ratio and test the rounding conditions? Ex) 739 or similar

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,82 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+		{
```
mattverse, line 142: Ditto with the comment above regarding jailed validator. Consider adding a test case on it as well

x/valset-pref/validator_set.go
stackman27: @mattverse we talked about this in the chain dev call. If we loop through the validators and check their jail, tombstoned &  unbonding status it might be very gas intensive, therefore i believe we can just have simple validator check like `GetValidator` and leave this as is. 

Another reason why we are being lenient with this check is that if in any iteration of the loop the validator is faulty, then the function returns an error and the entire message is reverted. **We are doing many iterations of `Delegate` function using only 1 message so we might have a leeway here**

mattverse: Just to confirm, that means that we're still adding a check to check if a validator is jailed, and if so, reverts the entire message?

x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,44 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
...
+	// loop through the validatorSetPreference and delegate the proportion of the tokens based on weights
+	for _, val := range existingSet.Preferences {
+		_, validator, err := k.getValAddrAndVal(ctx, val.ValOperAddress)
+		if err != nil {
+			return err
+		}
+
+		// tokenAmt takes the amount to delegate, calculated by {val_distribution_weight * tokenAmt}
+		// expect tokenAmt to be in the smallest denom for ex, 1 usmo = 10^6 osmo
+		tokenAmt := val.Weight.Mul(coin.Amount.ToDec()).RoundInt()
```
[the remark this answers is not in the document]
stackman27: aah gotchu will do

x/valset-pref/validator_set.go
stackman27: we have `GetValidator` that runs in ever iteration of `Val-Set` which checks that

x/valset-pref/msg_server_test.go
stackman27: hmm not sure how i can create a test scenario where a validator is jailed and user tries to delegate to that validator  🤔

x/valset-pref/msg_server_test.go
stackman27: added it here [`45964da`](https://github.com/osmosis-labs/osmosis/pull/3260/commits/45964da70bb84d553d3cc72ce3a796411b58217f)

x/valset-pref/msg_server_test.go
stackman27: added [here](https://github.com/osmosis-labs/osmosis/pull/3260/commits/09ee056004a254b10660d332b42cdbce1d2e047a#)

stackman27:
Thank you @mattverse and @czarcas7ic for the feedback. Everything should be resolved now :))

x/valset-pref/msg_server_test.go
stackman27: So i did some digging around and found out that "It is possible to delegate to a jailed validator, the only difference being it will not be added to the power index until it is unjailed" therefore, i dont think we can add a very indepth test

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,142 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+	amountToFund := sdk.Coins{sdk.NewInt64Coin(sdk.DefaultBondDenom, 100_000_000)} // 100 osmo
+
+	// prepare validators to delegate to
+	preferences := suite.PrepareDelegateToValidatorSet()
+
+	tests := []struct {
+		name           string
+		delegator      sdk.AccAddress
+		coin           sdk.Coin  // amount to delegate
+		expectedShares []sdk.Dec //expected shares after delegation
```
czarcas7ic, line 140:
```suggestion
		expectedShares []sdk.Dec // expected shares after delegation
```

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,142 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+			}
+
+			// call the create validator set preference
+			_, err := msgServer.DelegateToValidatorSet(c, types.NewMsgDelegateToValidatorSet(test.delegator, test.coin))
+			if test.expectPass {
+				suite.Require().NoError(err)
+
+				// check if the user balance decreased
+				balance := suite.App.BankKeeper.GetBalance(suite.Ctx, test.delegator, sdk.DefaultBondDenom)
+				suite.Require().True(balance.IsLT(amountToFund[0]))
```
czarcas7ic, line 187: why not check the exact amount rather than inequalities?

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,142 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+
+				// check if the expectedShares matches after delegation
+				for i, val := range preferences {
+					valAddr, err := sdk.ValAddressFromBech32(val.ValOperAddress)
+					suite.Require().NoError(err)
+
+					//guarantees that the delegator exists because we check it in DelegateToValidatorSet
```
czarcas7ic, line 194:
```suggestion
					// guarantees that the delegator exists because we check it in DelegateToValidatorSet
```

x/valset-pref/msg_server_test.go
```diff
@@ -124,3 +124,142 @@ func (suite *KeeperTestSuite) TestSetValidatorSetPreference() {
...
+func (suite *KeeperTestSuite) TestUnDelegateFromValidatorSet() {
+	tests := []struct {
+		name          string
+		delegator     sdk.AccAddress
+		coinToStake   sdk.Coin
+		coinToUnStake sdk.Coin
+		expectPass    bool
+	}{
+		{
+			name:          "Unstake x amount from a validator Set",
```
czarcas7ic, line 215: since its only one test case here you can just say "unstake half from a"

x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,92 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
...
+	for _, val := range existingSet.Preferences {
+		_, validator, err := k.getValAddrAndVal(ctx, val.ValOperAddress)
+		if err != nil {
+			return err
+		}
+
+		// tokenAmt takes the amount to delegate, calculated by {val_distribution_weight * tokenAmt}
+		tokenAmt := val.Weight.Mul(coin.Amount.ToDec()).TruncateInt()
+
+		// TODO: What happens here if validator is jailed, tombstoned, or unbonding
```
czarcas7ic, line 57: I think the only one we still are curious about is tombstoned

czarcas7ic (approved):
This LGTM, well done! Once we get another review here and this gets merged, please ensure to test this on a localosmosis instance with all the applicable CLI commands you used. This way we can easily test come next upgrade!

x/valset-pref/msg_server_test.go
stackman27: Yup will do

stackman27:
Thank you @czarcas7ic addressed all you comments and added more test coverage [`06f64d`](https://github.com/osmosis-labs/osmosis/commit/06f64dc0f7e410f1d7a9fdd6d037defdd787de79)

still looking on tombstoned validators and ways to approach them

czarcas7ic:
@stackman27 that comment was more to add it to the TODO, and not a blocker for this PR IMO

x/twap/api.go
```diff
@@ -58,17 +93,16 @@ func (k Keeper) GetArithmeticTwap(
 	if err != nil {
 		return sdk.Dec{}, err
 	}
-	return computeArithmeticTwap(startRecord, endRecord, quoteAssetDenom)
+	return strategy.computeTwap(startRecord, endRecord, quoteAssetDenom)
```
mattverse, line 96: doesnt this change to `getTwapToNow` anyways if `endTime.Equal(ctx.BlockTime())`?

x/valset-pref/msg_server_test.go
mattverse: how about unbonded or unbonding validators?

x/valset-pref/validator_set.go
mattverse: TODO: change comments

x/valset-pref/validator_set.go
```diff
@@ -29,6 +29,92 @@ func (k Keeper) SetupValidatorSetPreference(ctx sdk.Context, delegator string, p
...
+	// the total amount the user wants to undelegate
+	tokenAmt := sdk.NewDec(coin.Amount.Int64())
+
+	totalAmountFromWeights := sdk.NewDec(0)
+	for _, val := range existingSet.Preferences {
+		totalAmountFromWeights = totalAmountFromWeights.Add(val.Weight.Mul(tokenAmt))
+	}
+
+	if !totalAmountFromWeights.Equal(tokenAmt) {
+		return fmt.Errorf("The undelegate total donot add up with the amount calculated from weights expected %s got %s", tokenAmt, totalAmountFromWeights)
```
mattverse, line 93:
```suggestion
		return fmt.Errorf("The undelegate total do not add up with the amount calculated from weights expected %s got %s", tokenAmt, totalAmountFromWeights)
```

x/valset-pref/msg_server_test.go
mattverse: cannot see it due to force push :( , can you point me to the specific test case?

mattverse (approved):
LGTM! 

Left some minor comments I would love to see before merging!

x/twap/api.go
stackman27: oops this change wasn't supposed to be in this PR, but endTime = BlockTime logic is covered in line 84. We can follow up this conversation in this PR https://github.com/osmosis-labs/osmosis/pull/3529

x/valset-pref/msg_server_test.go
stackman27: Yup, i changed the set-validator weights to have upto 3 decimal places https://github.com/osmosis-labs/osmosis/pull/3260/files#diff-102acff6e2fd0c8c3b2b95afe0c166f91100a1dcbc1204c40a0ff981823504a3R31

x/valset-pref/msg_server_test.go
stackman27: Here's what i found; 
1. unbonded: Validators will be removed from active set but **still can receive delegations**
2. unbonding: Validators leaves active set either through slashing, jailing etc, **all delegations can be unbonded but will have to wait for unbonding time to exhaust** 

I am hoping to test all these in either e2e or testnet

ValarDragon:
Merging due to approvals, and this PR not being merged blocking things

PR #3275, opened 2022-11-07
Delete unused binary search

ValarDragon:
Deletes deadcode for a binary search over sdk.Int, if we want this in the future we should do it via a generic.


ValarDragon:
Ah I had an incorrect understanding, this is actually used

PR #3311, opened 2022-11-09
CL: Partially overlapping / gap price range test/logic

czarcas7ic:
Closes: #XXX

Slight adjust to logic to allow for gap positions

Adds gap position pool test and concurrent position pool test

### DIFF
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -215,9 +215,12 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 			// update the swapState's liquidity with the new tick's liquidity
 			newLiquidity := addLiquidity(swapState.liquidity, liquidityDelta)
 			swapState.liquidity = newLiquidity
-			if swapState.liquidity.LTE(sdk.ZeroDec()) || swapState.liquidity.IsNil() {
-				return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
-			}
+
+			// TODO: Investigate proper break of loop given the following conditions
+			// if swapState.liquidity.LTE(sdk.ZeroDec()) || swapState.liquidity.IsNil() {
+			// 	return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
+			// }
+
 			// update the swapState's tick with the tick we retrieved liquidity from
 			if zeroForOne {
 				swapState.tick = sdk.NewInt(nextTick - 1)
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -205,22 +205,16 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 		// tick has been consumed and we must move on to the next tick to complete the swap
 		if nextSqrtPrice.Equal(sqrtPrice) {
 			// retrieve the liquidity held in the next closest initialized tick
-			liquidityDelta, err := k.crossTick(ctx, p.Id, nextTick)
+			liquidityNet, err := k.crossTick(ctx, p.Id, nextTick)
 			if err != nil {
 				return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
 			}
 			if zeroForOne {
-				liquidityDelta = liquidityDelta.Neg()
+				liquidityNet = liquidityNet.Neg()
 			}
 			// update the swapState's liquidity with the new tick's liquidity
-			newLiquidity := addLiquidity(swapState.liquidity, liquidityDelta)
+			newLiquidity := addLiquidity(swapState.liquidity, liquidityNet)
 			swapState.liquidity = newLiquidity
-
-			// TODO: Investigate proper break of loop given the following conditions
-			// SEE https://github.com/osmosis-labs/osmosis/pull/3311/files/93c4bd0590278ef6d901f39bab73b123ddfc0da2#r1017566233
-			// if swapState.liquidity.LTE(sdk.ZeroDec()) || swapState.liquidity.IsNil() {
-			// 	return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
-			// }
 
 			// update the swapState's tick with the tick we retrieved liquidity from
 			if zeroForOne {
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -217,6 +217,7 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 			swapState.liquidity = newLiquidity
 
 			// TODO: Investigate proper break of loop given the following conditions
+			// SEE https://github.com/osmosis-labs/osmosis/pull/3311/files/93c4bd0590278ef6d901f39bab73b123ddfc0da2#r1017566233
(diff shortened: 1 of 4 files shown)


x/concentrated-liquidity/pool.go
```diff
@@ -215,9 +215,12 @@ func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
 			// update the swapState's liquidity with the new tick's liquidity
 			newLiquidity := addLiquidity(swapState.liquidity, liquidityDelta)
 			swapState.liquidity = newLiquidity
-			if swapState.liquidity.LTE(sdk.ZeroDec()) || swapState.liquidity.IsNil() {
-				return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
-			}
+
+			// TODO: Investigate proper break of loop given the following conditions
```
mattverse, line 219: Do the new test cases fail when we enable these checks?

x/concentrated-liquidity/pool_test.go
```diff
@@ -163,7 +163,7 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 				// create second position parameters
 				newLowerPrice := sdk.NewDec(5500)
 				s.Require().NoError(err)
-				newLowerTick := cl.PriceToTick(newLowerPrice) // 84222
+				newLowerTick := cl.PriceToTick(newLowerPrice) // 86129
```
mattverse, line 166: What's the reasoning behind this change?

x/concentrated-liquidity/pool_test.go
```diff
@@ -371,6 +469,70 @@ func (s *KeeperTestSuite) TestSwapOutAmtGivenIn() {
...
+			expectedTokenOut: sdk.NewCoin("eth", sdk.NewInt(1864112)),
+			expectedTick:     sdk.NewInt(87092),
+			newLowerPrice:    sdk.NewDec(5001),
+			newUpperPrice:    sdk.NewDec(6250),
+		},
+		//  Sequential price ranges with a gap
+		//
+		//          5000
+		//  4545 -----|----- 5500
+		//              5501 ----------- 6250
```
mattverse, line 508:
```suggestion
		//                           5501 ----------- 6250
```

x/concentrated-liquidity/pool_test.go
czarcas7ic: Was previously just a copy paste error, 5500 should correspond to tick 86129 (this is only a change in comment, so no logic here was touched)

x/concentrated-liquidity/pool.go
czarcas7ic: Specifically the test case where there is a break in position. So if we have a 4545-5500 and a 5502-6750 position, if we have this check then we we stop at tick 5500 instead of looking for another tick (in this case 5502), which imo is a bug

x/concentrated-liquidity/pool_test.go
czarcas7ic: Oh the way I am visualizing it, the start of the "-" is where liquidity starts/stops

x/concentrated-liquidity/pool.go
czarcas7ic: Agreed on context loss concern

This error path was never hit in all other test cases because all other test cases have overlapping ticks. When ticks overlap, liquidity can bleed into the next. The problem occurred with non overlapping fields where we use the entire liquidity of the upper tick, then we need to skip past empty ticks to the next initialized tick. I will need to think about this more to figure out what we should do

czarcas7ic: As discussed offline, included link to this discussion in code comment

x/concentrated-liquidity/pool_test.go
```diff
@@ -190,6 +191,245 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 			newLowerPrice:    sdk.NewDec(5500),
 			newUpperPrice:    sdk.NewDec(6250),
 		},
+		//  Consecutive price ranges
+		//
+		//                     5000
+		//             4545 -----|----- 5500
+		//  4000 ----------- 4545
```
p0mvn, line 198: Should the top one be a bit more to the right? Feels a bit confusing

x/concentrated-liquidity/pool_test.go
```diff
@@ -190,6 +191,245 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
...
+		//
+		"two positions with consecutive price ranges: eth -> usdc": {
+			addPositions: func(ctx sdk.Context, poolId uint64) {
+				// add first position
+				_, _, _, err = s.App.ConcentratedLiquidityKeeper.CreatePosition(ctx, poolId, s.TestAccs[0], defaultAmt0, defaultAmt1, sdk.ZeroInt(), sdk.ZeroInt(), lowerTick.Int64(), upperTick.Int64())
+				s.Require().NoError(err)
+				// params
+				// liquidity (1st):  1517818840.967515822610790519
```
p0mvn, line 206: Are these the expected values after creating position?

x/concentrated-liquidity/pool_test.go
```diff
@@ -190,6 +191,245 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
...
+			expectedTokenIn:  sdk.NewCoin("usdc", sdk.NewInt(10000000000)),
+			expectedTokenOut: sdk.NewCoin("eth", sdk.NewInt(1820536)),
+			expectedTick:     sdk.NewInt(87173),
+			newLowerPrice:    sdk.NewDec(5501),
+			newUpperPrice:    sdk.NewDec(6250),
+		},
+		// Slippage protection doesn't cause a failure but interrupts early.
+		"single position within one tick, trade completes but slippage protection interrupts trade early: eth -> usdc": {
+			addPositions: func(ctx sdk.Context, poolId uint64) {
+				// add first position
```
p0mvn, line 393: there's only one
```suggestion
				// add position
```

x/concentrated-liquidity/pool_test.go
```diff
@@ -190,6 +191,245 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenIn() {
...
+				_, _, _, err = s.App.ConcentratedLiquidityKeeper.CreatePosition(ctx, poolId, s.TestAccs[0], defaultAmt0, defaultAmt1, sdk.ZeroInt(), sdk.ZeroInt(), lowerTick.Int64(), upperTick.Int64())
+				s.Require().NoError(err)
+			},
+			tokenIn:       sdk.NewCoin("usdc", sdk.NewInt(5300000000)),
+			tokenOutDenom: "eth",
+			priceLimit:    sdk.NewDec(6000),
+			expectErr:     true,
+		},
+		"single position within one tick, trade does not complete due to lack of liquidity: eth -> usdc": {
```
p0mvn, line 422: Do we have a test over two positions where it does not consume all of the liquidity of the second position?

Also, might be good to have a test over 3 or more positions in the future

p0mvn (approved):
Great work. Comments are super helpful

x/concentrated-liquidity/pool_test.go
czarcas7ic: I took the value that CreatePosition returns, so this is assuming CreatePosition returns the correct number for liquidity

x/concentrated-liquidity/pool_test.go
czarcas7ic: Added two cases for two positions where we don't not consume all of the liquidity of the second position. Will add the later test in some eventual PR

czarcas7ic:
Going to merge this now, can address any further comments in future PRs (wanting to work on swaps now without large branch divergence)

Issue #3291, opened 2022-11-07
[CL][Swapping]: avoid redundant conversions between sqrt price and tick


p0mvn:
## Background

We perform redundant conversions:
https://github.com/osmosis-labs/osmosis/blob/b582a52bcdb7c390c520b99e3b757f2a68ea4429/x/concentrated-liquidity/pool.go#L138

These conversions utilize logarithm that is expensive to compute.

## Suggested Design

We store current tick in state and can provide it directly to `swapState` instead of recomputing

## Acceptance Criteria
- redundant conversions between sqrt price and tick are removed

p0mvn:
This is now addressed

PR #3319, opened 2022-11-10
Use current tick instead of recomputing

hieuvubk:
Closes: #3291 

Use `pool.currentTick` has been saved in the last `applySwap` instead of recalculating the new one


mattverse (approved):
LGTM!

PR #3321, opened 2022-11-10
IBC wasm hooks

nicolaslara:

This pull request adds a hooks middleware that allows us to easily add custom hooks on for IBC transfer.

It also adds `WasmHooks` that will parse an ICS20 packet and execute a contract if the memo contains instructions to do so in json (see README). That hook is configured in this PR.

This should have no effect on IBC transfers without the memo.

Potential future improvements:

 * Add a MultipleHooks struct that would allow us to configure multiple hooks to be pipelined
 * Move rate limits to an IBC hook. This would unify the way we deal with middlewares into the hooks

## Brief Changelog

 * Added an IBCHooks middleware that allows us to easily customize how the chain reacts to an IBC transfer. 
 * Added the first hook (WasmHooks) which execute a contract call based on the data sent in the ICS20 packet's memo

## Testing and Verifying

It would be great if we could replay the existing IBC traffic during a period of time to be sure that this change doesn't affect any IBC packets that we normally receive. 

README.md contains the spec


x/ibc-hooks/wasm_hook.go
```diff
@@ -0,0 +1,173 @@
...
+	fullAck := ContractAck{ContractResult: response.Data, IbcAck: ack.Acknowledgement()}
+	bz, err = json.Marshal(fullAck)
+	if err != nil {
+		return channeltypes.NewErrorAcknowledgement(fmt.Sprintf(types.ErrBadResponse, err.Error()))
+	}
+
+	return channeltypes.NewResultAcknowledgement(bz)
+}
+
+func ValidateMemo(memo string, receiver string) (shouldBreak bool, errStr string, contractAddr sdk.AccAddress, msgBytes []byte) {
```
ValarDragon, line 110:
```suggestion
func ValidateMemo(memo string, receiver string) (isWasmRouted bool, errStr string, contractAddr sdk.AccAddress, msgBytes []byte) {
```

ValarDragon (approved):
Awesome job, excited for this getting in!

PR #3330, opened 2022-11-10
fix: fix multihop gamm cli

czarcas7ic:
Closes: #XXX

When testing osmosis multihop discount, I found that utilizing multiple pool ids and denoms in the cli was not getting parsed correctly. This fixes that issue

### DIFF
--- a/x/gamm/client/cli/flags.go
+++ b/x/gamm/client/cli/flags.go
@@ -67,16 +67,16 @@ type smoothWeightChangeParamsInputs struct {
 func FlagSetQuerySwapRoutes() *flag.FlagSet {
 	fs := flag.NewFlagSet("", flag.ContinueOnError)
 
-	fs.StringArray(FlagSwapRoutePoolIds, []string{""}, "swap route pool id")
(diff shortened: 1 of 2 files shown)


p0mvn (approved):

ValarDragon:
Hrmm, lets not merge this in v13 just yet.

I'm pretty sure that `id1,id2` works, Viper does support reading arrays and a change here would break folks

czarcas7ic:
I have tried it a myriad of ways and was unable to get multihop to work via CLI prior to this PR

czarcas7ic:
Can anyone determine if this is in fact a required change? I would be shocked if the current CLI command works in its current form without this change

ValarDragon:
huh yeah, I am super confused. The Viper tests have string array working, but it doesn't work for me either via testing: https://github.com/spf13/viper/blob/202060b3a253bbac94795be51ca0b8e7432bf0fb/viper_test.go#L1008-L1011

czarcas7ic:
So, can we merge? 😅

PR #3337, opened 2022-11-11
fix: limit force unlock perms and add cli

czarcas7ic:
Closes: #XXX

Prevents addresses with force unlock perms from unlocking any address. Limits their perms to only force unlock locks they create.

Also adds CLI for force unlock to lockup module.


ValarDragon (approved):

ValarDragon:
yikes that we didn't have this before

PR #3347, opened 2022-11-11
fix(stableswap): require sorted initial liquidity

p0mvn:
Closes: #XXX

Currently, the CLI auto-sorts the initial liquidity. As a result, denoms end up mismatching the desired scaling factor.

This PR  removes the sorting in CLI and adds the requirement in `ValidateBasic()` for initial liquidity coins to be sorted. Fails if not.

## Testing and Verifying

Tested on localnet. Summary: https://hackmd.io/40cR4cL5SBOuX7APZKTWIw?edit

### DIFF
--- a/x/gamm/client/cli/tx.go
+++ b/x/gamm/client/cli/tx.go
@@ -856,6 +856,8 @@ func NewStableSwapAdjustScalingFactorsMsg(clientCtx client.Context, txf tx.Facto
 	return txf, msg, nil
 }
 
+// ParseCoinsNoSort parses coins from coinsStr but does not sort them.
+// Returns error if parsing fails.
 func ParseCoinsNoSort(coinsStr string) (sdk.Coins, error) {
 	coinStrs := strings.Split(coinsStr, ",")
 	decCoins := make(sdk.DecCoins, len(coinStrs))
--- a/x/gamm/client/cli/tx.go
+++ b/x/gamm/client/cli/tx.go
@@ -479,22 +479,22 @@ func NewBuildCreateBalancerPoolMsg(clientCtx client.Context, txf tx.Factory, fs 
 
 // Apologies to whoever has to touch this next, this code is horrendous
 func NewBuildCreateStableswapPoolMsg(clientCtx client.Context, txf tx.Factory, fs *flag.FlagSet) (tx.Factory, sdk.Msg, error) {
-	pool, err := parseCreateStableswapPoolFlags(fs)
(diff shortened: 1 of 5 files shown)


x/gamm/client/cli/tx.go
```diff
@@ -855,3 +855,17 @@ func NewStableSwapAdjustScalingFactorsMsg(clientCtx client.Context, txf tx.Facto
 
 	return txf, msg, nil
 }
+
+func ParseCoinsNoSort(coinsStr string) (sdk.Coins, error) {
```
p0mvn, line 859:
```suggestion
// ParseCoinsNoSort parses coins from coinsStr but does not sort them.
// Returns error if parsing fails.
func ParseCoinsNoSort(coinsStr string) (sdk.Coins, error) {
```

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -418,13 +418,22 @@ func validateScalingFactors(scalingFactors []uint64, numAssets int) error {
 
 // assumes liquidity is all pool liquidity, in correct sorted order
 func validatePoolLiquidity(liquidity sdk.Coins, scalingFactors []uint64) error {
+	liquidityCopy := make(sdk.Coins, len(liquidity))
+	copy(liquidityCopy, liquidity)
+	liquidityCopy.Sort()
+
```
ValarDragon, line 424: Can we add a test for the go logic?

ValarDragon (approved):
LGTM, I think we should add a test for the go logic though

x/gamm/pool-models/stableswap/pool.go
```diff
@@ -418,13 +418,22 @@ func validateScalingFactors(scalingFactors []uint64, numAssets int) error {
 
 // assumes liquidity is all pool liquidity, in correct sorted order
 func validatePoolLiquidity(liquidity sdk.Coins, scalingFactors []uint64) error {
+	liquidityCopy := make(sdk.Coins, len(liquidity))
```
alexanderbez, line 421: Any reason `sdk.NewCoins(liquidity...)` would work?

alexanderbez (approved):

x/gamm/pool-models/stableswap/pool.go
ValarDragon: agreed, that seems better to me as well

p0mvn: It also sorts the input coins. We don't want that

As a result, we would have to make a new copy anyways

PR #3354, opened 2022-11-12
Break up osmo router code more

ValarDragon:
Breaks up the osmo router code more

### DIFF
--- a/x/gamm/types/route.go
+++ b/x/gamm/types/route.go
@@ -27,6 +27,30 @@ func (routes SwapAmountInRoutes) Validate() error {
 	return nil
 }
 
+func (routes SwapAmountInRoutes) IntermediateDenoms() []string {
+	if len(routes) < 2 {
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


x/gamm/types/route.go
```diff
@@ -27,6 +27,30 @@ func (routes SwapAmountInRoutes) Validate() error {
 	return nil
 }
 
+func (routes SwapAmountInRoutes) IntermediateDenoms() []string {
+	if len(routes) < 2 {
+		return nil
+	}
+	intermediateDenoms := make([]string, 0, len(routes)-1)
+	for _, route := range routes[:len(routes)-1] {
```
czarcas7ic, line 35: for my learning, why specify routes[:len(routes)-1] when we determined it has to be < 2 already?

czarcas7ic (approved):
👍 thank you!

x/gamm/types/route.go
ValarDragon: ah this is because we don't want the last token out, only wanted the intermediate one

PR #3356, opened 2022-11-12
Add stableswap to pooltype, and pooltype query to stargate whitelist

ValarDragon:

Resolves an issue pointed out by @jonator 

## Brief Changelog

- Add stableswap support for pooltype query
- Make pooltype query callable from cosmwasm
- Fix bug in stableswap msg validate basic, when no scaling factors provided

## Testing and Verifying

- Test has been added

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -5,6 +5,7 @@ import (
 
 	gammkeeper "github.com/osmosis-labs/osmosis/v12/x/gamm/keeper"
 	"github.com/osmosis-labs/osmosis/v12/x/gamm/pool-models/balancer"
+	"github.com/osmosis-labs/osmosis/v12/x/gamm/pool-models/stableswap"
 	gammtypes "github.com/osmosis-labs/osmosis/v12/x/gamm/types"
 )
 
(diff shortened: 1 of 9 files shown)


jonator:
@ValarDragon I don't think an individual pool query for type is a good performance move here (prob good to have in general, though). IMO we should include the type, and include nullable scaling factors in the existing pools query and pool model

ValarDragon:
That definitely doesn't make sense for the general query. Its bad design to have fields that have no sensical interpretation for a given type configuration.

 What is instead available right now is interpreting the result of the pool query as a protobuf `oneof` between `Balancer pool` and `Stableswap pool` that can then be type casted. (This is a poor man's version of a sum type)
 
 I'd prefer for the single query, we handle the pool enum complexity for it.

jonator:
OK. I didn't know protobuf could handle a union/sum type like that. Yeah then that would be ideal to have a union of each of the types of pools. Then the pools (plural) query could contain a list of that union type.

app/apptesting/gamm.go
```diff
@@ -81,6 +87,21 @@ func (s *KeeperTestHelper) PrepareBalancerPool() uint64 {
 	return poolId
 }
 
+func (s *KeeperTestHelper) PrepareBasicStableswapPool() uint64 {
```
mattverse, line 90: Nice!

mattverse (approved):
LGTM!

PR #3358, opened 2022-11-12
Move around some wasmhook wiring code

ValarDragon:

Some minor moving around of wasmhook wiring code & file names

## Testing and Verifying


app/keepers/keepers.go
```diff
@@ -123,11 +123,11 @@ type AppKeepers struct {
...
 	// IBC modules
 	// transfer module
-	TransferModule          transfer.AppModule
-	RateLimitingICS4Wrapper *ibcratelimit.ICS4Wrapper
-	HooksMiddleware         *ibchooks.IBCMiddleware
-	WasmHooks               *ibchooks.WasmHooks
-	HooksICS4Wrapper        ibchooks.ICS4Middleware
+	RawIcs20TransferAppModule transfer.AppModule
+	RateLimitingICS4Wrapper   *ibcratelimit.ICS4Wrapper
+	CombinedTransferIBCModule *ibchooks.IBCMiddleware
```
nicolaslara, line 128: IBC-go uses the term "transfer stack' maybe we should use that? This is a pointer to the top of the transfer stack, which right now just happens to be the ibc_hooks middleware. Or maybe `TransferModuleWithHooks`? 

Though I don't want to bikeshed this too much, so whatever works.

ValarDragon: TransferStack SGTM!

nicolaslara (approved):

PR #3374, opened 2022-11-14
replace bad bytecode

nicolaslara:
Main doesn't have the correct bytecode for rate limiting. This is likely due to a bad merge. This PR replaces it with the proper bytecode


mattverse (approved):

mattverse (commented):
utACK

p0mvn (approved):

PR #3217, opened 2022-11-02
Add CalcJoinPoolShares and CalcExitPoolCoinsFromShares pool queries to Stargate whitelist

georgemc98:
Related to #2956 and dependent on PR #2972  

## What is the purpose of the change

This PR adds the `CalcJoinPoolShares` and `CalcExitPoolCoinsFromShares` queries added in PR #2972 to the whitelisted Stargate queries in the wasmbindings.

## Brief Changelog

*(for example:)*
 
  - Add CalcJoinPoolShares query to whitelisted Stargate queries
  - Add CalcExitPoolCoinsFromShares query to whitelisted Stargate queries

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? yes
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? yes
  - How is the feature or change documented? not documented

PR #3383, opened 2022-11-15
test(gamm): add stableswap tests for new stargate queries

p0mvn:
Closes: #XXX

Most of our query and other gamm tests only test balancer pool. It is safer to test any supported pool and make sure that the same outcome is returned given the same inputs. Asserting this might be particularly useful for stargate queries that smart contracts rely upon.

This PR extends existing tests with the stableswap pool model for new queries whitelisted here: https://github.com/osmosis-labs/osmosis/pull/3217

Some error checks are shifted around to make sure that both pool models return the same error.

The way we test both pool modes is by creating shared test cases and a list of pool model ids. For each pool id, we create a new shared test case. This allows to reduce code duplication and to reuse the same test cases.
See this function for details:
https://github.com/osmosis-labs/osmosis/blob/e16c8ee62f76e6f1613df6450c8dd2d461ebfad2/x/gamm/keeper/keeper_test.go#L138-L153

and this is how we merge tests:
https://github.com/osmosis-labs/osmosis/blob/e16c8ee62f76e6f1613df6450c8dd2d461ebfad2/x/gamm/keeper/grpc_query_test.go#L38-L59

In the future, we can refactor to extract common validation into the `types` package.

## State Breakage

This is state-breaking due to removing `sdkerrors` and shifting around error checks.

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -37,6 +37,7 @@ var DefaultStableswapLiquidity = sdk.NewCoins(
 	sdk.NewCoin("foo", sdk.NewInt(10000000)),
 	sdk.NewCoin("bar", sdk.NewInt(10000000)),
 	sdk.NewCoin("baz", sdk.NewInt(10000000)),
+	sdk.NewCoin("uosmo", sdk.NewInt(10000000)),
 )
 
 // PrepareBalancerPoolWithCoins returns a balancer pool
--- a/x/gamm/pool-models/balancer/amm.go
+++ b/x/gamm/pool-models/balancer/amm.go
@@ -234,7 +234,7 @@ func ensureDenomInPool(poolAssetsByDenom map[string]PoolAsset, tokensIn sdk.Coin
 	for _, coin := range tokensIn {
 		_, ok := poolAssetsByDenom[coin.Denom]
 		if !ok {
-			return sdkerrors.Wrapf(types.ErrDenomNotFoundInPool, invalidInputDenomsErrFormat, coin.Denom)
+			return types.ErrInputDenomsNotInPool
 		}
 	}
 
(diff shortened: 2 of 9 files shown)


x/gamm/pool-models/balancer/amm.go
```diff
@@ -234,7 +234,7 @@ func ensureDenomInPool(poolAssetsByDenom map[string]PoolAsset, tokensIn sdk.Coin
 	for _, coin := range tokensIn {
 		_, ok := poolAssetsByDenom[coin.Denom]
 		if !ok {
-			return sdkerrors.Wrapf(types.ErrDenomNotFoundInPool, invalidInputDenomsErrFormat, coin.Denom)
+			return types.ErrInputDenomsNotInPool
```
p0mvn, line 237: Note to reviewer: reusing this error with stableswap doesn't allow for formatting a denom. Planning to create an issue to refactor stableswap side to improve this.

x/gamm/pool-models/balancer/pool.go
```diff
@@ -792,11 +796,6 @@ func (p *Pool) CalcJoinPoolNoSwapShares(ctx sdk.Context, tokensIn sdk.Coins, swa
 		return sdk.ZeroInt(), sdk.NewCoins(), err
 	}
 
-	// ensure that there aren't too many or too few assets in `tokensIn`
-	if tokensIn.Len() != p.NumAssets() {
-		return sdk.ZeroInt(), sdk.NewCoins(), errors.New("no-swap joins require LP'ing with all assets in pool")
-	}
-
```
p0mvn, line 799: Note to reviewer: this was moved up so that both pool models return the same error given the same inputs

app/apptesting/gamm.go
```diff
@@ -37,6 +37,7 @@ var DefaultStableswapLiquidity = sdk.NewCoins(
 	sdk.NewCoin("foo", sdk.NewInt(10000000)),
 	sdk.NewCoin("bar", sdk.NewInt(10000000)),
 	sdk.NewCoin("baz", sdk.NewInt(10000000)),
+	sdk.NewCoin("uosmo", sdk.NewInt(10000000)),
```
ValarDragon, line 40: I think the main concern with this PR for me, is adding this right here

PR #3393, opened 2022-11-15
[CL]: Calculation for `MinSqrtRatio`

hieuvubk:
Sub component of : #3178

Add a function to calculate `MinSqrtRatio` instead of using zero value.
I'm using `BigDec` to save value but limit 36 ​​numbers prec (a little confused at this point)

### DIFF
--- a/x/concentrated-liquidity/types/constants.go
+++ b/x/concentrated-liquidity/types/constants.go
@@ -23,7 +23,6 @@ var (
 
 // Calculates MinSqrtPrice = sqrt(1.0001^MinTick)
 func GetMinSqrtRatio() sdk.Dec {
-	minSqrtRatio := osmomath.MustNewDecFromStr(strconv.FormatFloat(math.Pow(1.0001, -887272 / 2), 'f', 36, 64))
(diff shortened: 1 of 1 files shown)


x/concentrated-liquidity/types/constants.go
```diff
@@ -12,5 +17,11 @@ var (
 	MaxSqrtRatio = sdk.MustNewDecFromStr("18446050711097703529.7763428")
 	// TODO: this is a temp value, figure out math for this.
 	// we basically want getSqrtRatioAtTick(MIN_TICK)
-	MinSqrtRatio = sdk.MustNewDecFromStr("0")
+	MinSqrtRatio = GetMinSqrtRatio()
```
p0mvn, line 20: We should hardcode this value instead

czarcas7ic: Yeah, we can just have GetMinSqrtRatio take an input, in this case of -887272

p0mvn:
Thanks for the PR @hieuvubk . We are not going to go in this direction so I'm going to close this to keep the PR list short. Please let me know if any concerns

PR #3394, opened 2022-11-15
Added rate limiter setup to upgrade handler

nicolaslara:

Sets up the rate limits on the upgrade handler.

Open Question: Are the quotas setup here enough? do we want to add others? 

To deal with the "attach around rest time" issue we can add a quota with a 1.5w timeout too. 

## Brief Changelog

Added rate limit quotas

## Testing and Verifying

Added a test that validates the contract has been properly instantiated and has quotas.


app/upgrades/v13/upgrades.go
```diff
@@ -1,15 +1,84 @@
...
+	}
+
+	transferModule := keepers.AccountKeeper.GetModuleAddress(transfertypes.ModuleName)
+
+	denoms := []string{
+		"uosmo",
+		"ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2", // atom
+		"ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858", // usdc
+		"ibc/EA1D43981D5C9A1C4AAEA9C23BB1D4FA126BA9BC7020A25E0AE4AA841EA25DC5", // weth
+		"ibc/EA1D43981D5C9A1C4AAEA9C23BB1D4FA126BA9BC7020A25E0AE4AA841EA25DC5", // wbtc
```
p0mvn, line 42: I think this is the right one but please double-check

```suggestion
		"ibc/D1542AA8762DB13087D8364F3EA6509FD6F009A34F00426AF9E4F9FA85CBBF1F", // wbtc
```

app/upgrades/v13/upgrades.go
```diff
@@ -1,15 +1,84 @@
...
+		// Remove the trailing comma because JSON can't handle it
+		quotas = strings.TrimRight(quotas, ",")
+	}
+
+	initMsgBz := []byte(fmt.Sprintf(`{
+           "gov_module":  "%s",
+           "ibc_module":"%s",
+           "paths": [%s]
+        }`,
+		govModule, transferModule, quotas))
```
czarcas7ic, line 60: What is the transferModule allowing here? Did we decide to remove this from the smart contract or are we leaving it in for a particular reason?

app/upgrades/v13/upgrades.go
czarcas7ic, line 42: Side note: When we test this on testnet, we should code in pre existing ibc denoms from the testnet to ensure this works as expected

app/upgrades/v13/upgrades.go
```diff
@@ -1,15 +1,84 @@
...
+
+	addr, _, err := contractKeeper.Instantiate(ctx, codeID, govModule, govModule, initMsgBz, "rate limiting contract", nil)
```
czarcas7ic, line 62: I'm confused on this line. From my memory it seems like you are passing in the gov module twice, the second time in place of the ibc_module address, but above you are using the transferModule address in the initMsgBz. Can you expand further here on what is actually happening?

app/upgrades/v13/upgrades.go
```diff
@@ -1,15 +1,84 @@
...
+	}
+	params, err := ibcratelimittypes.NewParams(addrStr)
+	if err != nil {
+		return err
+	}
+	paramSpace, ok := keepers.ParamsKeeper.GetSubspace(ibcratelimittypes.ModuleName)
+	if !ok {
+		return sdkerrors.New("rate-limiting-upgrades", 2, "can't create paramspace")
+	}
+	paramSpace.SetParamSet(ctx, &params)
```
czarcas7ic, line 78: For my learning, how does GetSubspace pass if we SetParamSet below this? Is SetParamSet even required if NewParams is setting this for us?

czarcas7ic (commented):
Should be good to go after addressing Roman's comment on WBTC ibc denom

app/upgrades/v13/upgrades.go
mattverse: @czarcas7ic  wdym by coding it in the testnet?

czarcas7ic: What I mean is we should change these denoms to five testnet ibc denoms when we upgrade testnet and check rate limiting works on them

app/upgrades/v13/upgrade_test.go
```diff
@@ -53,25 +71,32 @@ func (suite *UpgradeTestSuite) TestUpgrade() {
...
+				paramSpace, ok := suite.App.ParamsKeeper.GetSubspace(ibcratelimittypes.ModuleName)
+				suite.Require().True(ok)
+				var contract string
+				paramSpace.GetIfExists(suite.Ctx, ibcratelimittypes.KeyContractAddress, &contract)
+				suite.Require().NotEmpty(contract)
+
+				// The quotas are configured?
+				contractAddr, err := sdk.AccAddressFromBech32(contract)
+				suite.Require().NoError(err)
+				bytes, err := suite.App.WasmKeeper.QuerySmart(suite.Ctx, contractAddr, []byte(`{"get_quotas": {"channel_id": "any", "denom": "uosmo"}}`))
```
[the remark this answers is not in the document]
nicolaslara: I figured if one was there then the transaction succeeded, but I'll add the other checks

app/upgrades/v13/upgrades.go
nicolaslara: I was gonna remove it and then found why I had added that in the first place. The transfer module is allowed to add or remove quotas. The idea is that in the future we may want to automatically add or remove channels as they get created/deleted

app/upgrades/v13/upgrades.go
nicolaslara: Those fields on `Instantiate` refer to `creator` and `admin`. The creator is who created the contract and will be the sender of the instantiate message. The admin is who is allowed to upgrade the contract. These should both be set to the gov module.

The transfer module passed above is just part of the message and it's the contract (not wasmd) who knows how to deal with it.

app/upgrades/v13/upgrades.go
nicolaslara: As I understand it, NewParams gives is a `Params{}` struct (as we define it on `types/params.go` for the rate limit middleware), which is what `SetParamSet` expects. We could add a `SetParam()` function but it would basically do this internally. 

`GetSubspace()` just gives us a Subspace struct, which knows how to get the right keys from a kvstore. 

TBH, I find the whole params implementation super confusing, so if someone knows a better way to deal with this I'm open to suggestions

app/upgrades/v13/upgrade_test.go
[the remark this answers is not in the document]
nicolaslara: Added the tests for all denoms

p0mvn (approved):

czarcas7ic (approved):

PR #3396, opened 2022-11-15
chore: modify multi-hop table driven tests

czarcas7ic:
Closes: #XXX

With modifications to the osmo multi-hop fee reduction logic, we needed to also modify the table driven tests. This PR simplifies the current table driven test logic as well as implements the new fee logic.

### DIFF
--- a/x/gamm/keeper/multihop_test.go
+++ b/x/gamm/keeper/multihop_test.go
@@ -11,6 +11,13 @@ import (
 	poolincentivestypes "github.com/osmosis-labs/osmosis/v12/x/pool-incentives/types"
 )
 
+var (
+	fooCoin   = sdk.NewCoin("foo", sdk.NewInt(1000000000))
+	barCoin   = sdk.NewCoin("bar", sdk.NewInt(1000000000))
+	bazCoin   = sdk.NewCoin("baz", sdk.NewInt(1000000000))
+	uosmoCoin = sdk.NewCoin("uosmo", sdk.NewInt(1000000000))
+)
+
 func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn() {
 	type param struct {
 		routes            []types.SwapAmountInRoute
@@ -47,9 +54,9 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn() 
 				tokenIn:           sdk.NewCoin("foo", sdk.NewInt(100000)),
 				tokenOutMinAmount: sdk.NewInt(1),
 			},
-			coinA:      sdk.NewCoin("foo", sdk.NewInt(1000000000)),
-			coinB:      sdk.NewCoin("bar", sdk.NewInt(1000000000)),
-			coinC:      sdk.NewCoin("baz", sdk.NewInt(1000000000)),
+			coinA:      fooCoin,
+			coinB:      barCoin,
+			coinC:      bazCoin,
 			expectPass: true,
 		},
 		{
@@ -69,9 +76,9 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn() 
 				tokenIn:           sdk.NewCoin("foo", sdk.NewInt(100000)),
 				tokenOutMinAmount: sdk.NewInt(1),
 			},
-			coinA:             sdk.NewCoin("foo", sdk.NewInt(1000000000)),
-			coinB:             sdk.NewCoin("uosmo", sdk.NewInt(1000000000)),
-			coinC:             sdk.NewCoin("baz", sdk.NewInt(1000000000)),
+			coinA:             fooCoin,
+			coinB:             uosmoCoin,
+			coinC:             bazCoin,
 			reducedFeeApplied: true,
 			expectPass:        true,
 		},
@@ -92,9 +99,9 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn() 
 				tokenIn:           sdk.NewCoin("foo", sdk.NewInt(100000)),
 				tokenOutMinAmount: sdk.NewInt(1),
 			},
-			coinA:             sdk.NewCoin("foo", sdk.NewInt(1000000000)),
-			coinB:             sdk.NewCoin("uosmo", sdk.NewInt(1000000000)),
-			coinC:             sdk.NewCoin("baz", sdk.NewInt(1000000000)),
+			coinA:             fooCoin,
+			coinB:             uosmoCoin,
+			coinC:             bazCoin,
 			reducedFeeApplied: true,
 			expectPass:        true,
 		},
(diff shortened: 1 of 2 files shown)


x/gamm/keeper/multihop_test.go
```diff
@@ -55,32 +65,55 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
 
-			// Prepare pools
-			suite.PrepareBalancerPoolWithPoolParams(balancer.PoolParams{
-				SwapFee: poolDefaultSwapFee,
-				ExitFee: sdk.NewDec(0),
-			})
-			suite.PrepareBalancerPoolWithPoolParams(balancer.PoolParams{
-				SwapFee: poolDefaultSwapFee,
-				ExitFee: sdk.NewDec(0),
-			})
```
czarcas7ic, line 83: Note to reviewer, these methods created four asset pools, making the test unnecessarily difficult to test osmo swap fee reduction. Switched to two asset pools for simplicity

x/gamm/keeper/multihop_test.go
```diff
@@ -94,35 +127,28 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
-					pool, err := keeper.GetPoolAndPoke(suite.Ctx, route.PoolId)
-					suite.NoError(err, "test: %v", test.name)
-
-					sp, err := pool.SpotPrice(suite.Ctx, tokenInDenom, route.TokenOutDenom)
-					suite.NoError(err, "test: %v", test.name)
-					dec = dec.Mul(sp)
-				}
-				return dec
-			}
-
```
czarcas7ic, line 114: Note to reviewer: I got rid of these spot price calculations, it was unclear to me why they were needed for multi-hop tests

x/gamm/keeper/multihop_test.go
```diff
@@ -216,9 +232,36 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
+						TokenInDenom: "uosmo",
+					},
+				},
+				poolFee:          []sdk.Dec{poolDefaultSwapFee, sdk.NewDecWithPrec(1, 1)},
+				tokenInMaxAmount: sdk.NewInt(90000000),
+				tokenOut:         sdk.NewCoin("baz", sdk.NewInt(100000)),
+			},
+			coinA:             sdk.NewCoin("foo", sdk.NewInt(1000000000)),
+			coinB:             sdk.NewCoin("uosmo", sdk.NewInt(1000000000)),
+			coinC:             sdk.NewCoin("baz", sdk.NewInt(1000000000)),
```
p0mvn, line 264: nit: consider defining these up top and reusing

x/gamm/keeper/multihop_test.go
```diff
@@ -295,40 +327,21 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountOut()
...
-					diffA := tokenInCalculatedAsSeparateSwaps.Amount.Sub(multihopTokenInAmount)
-					diffB := multihopTokenInAmount.Sub(tokenInCalculatedAsSeparateSwapsWithoutFee.Amount)
-					diffDistinctionPercent := diffA.Sub(diffB).Abs().ToDec().Quo(diffA.Add(diffB).ToDec())
-					suite.Require().True(diffDistinctionPercent.LT(sdk.MustNewDecFromStr("0.01")))
+					// calculate the swap as separate swaps with the new reduced swap fee
+					expectedMultihopTokenOutAmount = calcInAmountAsSeparateSwaps(true)
 				} else {
-					suite.Require().True(multihopTokenInAmount.Equal(tokenInCalculatedAsSeparateSwaps.Amount))
+					// calculate the swap as separate swaps with the default swap fee
+					expectedMultihopTokenOutAmount = calcInAmountAsSeparateSwaps(false)
```
p0mvn, line 336: you can also do the following instead of if / else:
```
expectedMultihopTokenOutAmount = calcInAmountAsSeparateSwaps(test.reducedFeeApplied)
```

p0mvn (approved):

x/gamm/keeper/multihop_test.go
```diff
@@ -55,32 +65,55 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
 				tokenOutMinAmount: sdk.NewInt(1),
 			},
+			coinA:             sdk.NewCoin("foo", sdk.NewInt(1000000000)),
+			coinB:             sdk.NewCoin("uosmo", sdk.NewInt(1000000000)),
+			coinC:             sdk.NewCoin("baz", sdk.NewInt(1000000000)),
+			reducedFeeApplied: true,
+			expectPass:        true,
+		},
+		{
+			name: "Swap - [foo -> uosmo](pool 1) - [uosmo -> baz](pool 2) with a half fee applied, (pool 1) 1 percent fee, (pool 2) 10 percent fee",
```
p0mvn, line 79: q: do we have a test where one route has a reduced fee but the next one does not or vice versa?

czarcas7ic: So I might be misunderstanding your question, but the reduced fee is an all or nothing deal. If the route is greater than two hops, the reduced fee does not get applied. Please lmk if this addresses your concern

p0mvn: Makes sense but do we have a test to cover this case?

That is, more than 2 hoops AND the first hop is osmo routed BUT the half fee is not applied

czarcas7ic: Agreed, will add this!

czarcas7ic:
@p0mvn I made some relatively significant changes to accommodate for that fourth test case. If you could give it another quick glance it would be much appreciated!

x/gamm/keeper/multihop_test.go
```diff
@@ -11,21 +11,33 @@ import (
...
+	uosmoCoin = sdk.NewCoin("uosmo", sdk.NewInt(1000000000))
+)
+
 func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn() {
 	type param struct {
-		routes            []types.SwapAmountInRoute
-		tokenIn           sdk.Coin
-		tokenOutMinAmount sdk.Int
+		routes             []types.SwapAmountInRoute
+		incentivizedGauges []uint64
```
mattverse, line 24: just curious, whats the significance of having incentivized Gauges for testing multihop?

x/gamm/keeper/multihop_test.go
```diff
@@ -55,74 +92,92 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
+			calcOutAmountAsSeparateSwaps := func(osmoFeeReduced bool) sdk.Coin {
 				cacheCtx, _ := suite.Ctx.CacheContext()
 
-				if adjustedPoolSwapFee != poolDefaultSwapFee {
+				if osmoFeeReduced {
 					for _, hop := range test.param.routes {
-						err := suite.updatePoolSwapFee(cacheCtx, hop.PoolId, adjustedPoolSwapFee)
-						suite.NoError(err, "test: %v", test.name)
+						// determine the expected reduced swap fee
+						hopPool, _ := keeper.GetPoolAndPoke(cacheCtx, hop.PoolId)
```
mattverse, line 167: nit: consider doing the err check here just to be safe

x/gamm/keeper/multihop_test.go
```diff
@@ -55,74 +92,92 @@ func (suite *KeeperTestSuite) TestBalancerPoolSimpleMultihopSwapExactAmountIn()
...
+						currentPoolSwapFee := hopPool.GetSwapFee(cacheCtx)
```
mattverse, line 168: would it be possible to add inline comments on what is happening here? It was kind of hard for me to follow why we need to update the pool swap fees directly in the beginning

mattverse: From my understanding, I think it's getting / calculating the swap fees manually and then updating the pool swap fee manually in accordance to the calculated swap fee, and then compare it with the calculation that has been done via multihop swap method.

mattverse (approved):
LGTM, left some minor comments that I would like to see, but shouldn't be a big blocker for this PR

czarcas7ic:
@ValarDragon I can get this added tomorrow!

ValarDragon:
awesome!

x/gamm/keeper/multihop_test.go
czarcas7ic: Part of the new spec is reduced fees don't happen to ALL osmo-x pool pairs, but instead only incentivized osmo-x pool pairs! (this prevents potential misuse)

czarcas7ic:
@ValarDragon was able to combine four asset pool tests and two asset pool tests, and was also able to test the osmo fee reduction with four asset pools now after the test refactor. Will merge today if no further comments!

ValarDragon (approved):

Issue #3251, opened 2022-11-04
Move binary search to its own package


ValarDragon:
## Background

We should move binary search out of osmoutils into its own package. Its a nice self-contained, reusable unit. I suggest we make this a sub-package within `osmomath`.

## Acceptance Criteria

It is contained in its own subpackage

PR #3405, opened 2022-11-16
Move binary search to osmomath

ValarDragon:
Closes: #3251 

Simple file move, to move binary search to the math package which makes more sense imo. It could eventually have a sub-component within math if we feel like math is getting too big.


p0mvn (approved):

ValarDragon:
Easier to just remake this PR

Issue #3399, opened 2022-11-15
fix: import paths CI job


p0mvn:
## Background

Import paths ci job is failing:
https://github.com/osmosis-labs/osmosis/actions/runs/3473319104/jobs/5805205259

We should update the relevant scripts to unblock the CI.

## Acceptance Criteria
- import paths CI job is working
- updates all relevant import paths correctly
- test on a personal fork added

RusAkh:
I'll fix it

PR #3408, opened 2022-11-16
CI: Fix import path

RusAkh:
Closes: #3399

For some reason this [line](https://github.com/osmosis-labs/osmosis/blob/main/scripts/replace_import_paths.sh#L22) with normal paths returned `./x/twap/TWAP` file which does not exist. 

`Changes`: just added if statement that checks if a file exists when iterating them.

`Checks`: [successful action run](https://github.com/RusAkh/osmosis-fork/actions/runs/3476923028/jobs/5812582037)
Note that in this run I removed pull request job


p0mvn (approved):
Thank you

Issue #3348, opened 2022-11-11
Github CI: Update deprecated go test cache work


ValarDragon:
## Background

We see this warning `GitHub Actions: Deprecating save-state and set-output commands` on every PR now, in the load from cache step, linking to: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Haven't read through yet, but seems like something we need to fix

RusAkh:
id like to fix it

RusAkh:
https://github.com/mheap/github-action-required-labels/issues/42#issuecomment-1315784425

PR #3409, opened 2022-11-16
CI: undeprecate set-output

RusAkh:
Closes: #3348

[Changes according to this](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

However, there are still many actions which depend on other actions, which were not updated. It is fine for now, but will break `1st of June 2023` Example: https://github.com/osmosis-labs/osmosis/issues/3348#issuecomment-1316364643


ValarDragon (approved):
Oh cool, thanks!

PR #3412, opened 2022-11-16
Remove old SpotPrice query from stargate query

mattverse:
The old spot price stargate query should have been removed at one point, but seems that it has been added along in a different PR. 

This PR re-removes the old spot price query


ValarDragon:
oh hrm, given that it caused in E2E break, I probably just never removed it

mattverse:
@ValarDragon should we go ahead and fix the e2e test as well for this?

ValarDragon (approved):

PR #3413, opened 2022-11-16
recompiling wasm

nicolaslara:
For some reason the `.wasm` file that ended up in main is not the one that compiles from the latest rate limiting code.

This PR updates the bytecode.


czarcas7ic:
I might be wrong but this PR appears to merge no diff

nicolaslara:
huh! weird. That's probably related to why the `.wasm` files aren't getting properly merged.

nicolaslara:
ok, so the checksum of this and main are the same. That leads me to think that the issue with the checks is on the checks themselves and not the bytecode. I'll keep debugging and can probably close this for now

Issue #3013, opened 2022-10-17
[EPIC][x/twap] Geometric TWAP


p0mvn:
## Background

> While the arithmetic mean TWAPs are much more widely used, they should theoretically be less accurate in measuring a geometric Brownian motion process (which is how price movements are usually modeled)

Arithmetic TWAP tends to overweight higher prices relative to lower ones.

Therefore, we should implement Geometric TWAP.

## Suggested Design

The geometric mean can be expressed as the exponential of the arithmetic mean of logarithms.

Ultimately, we would like to calculate:

![image](https://user-images.githubusercontent.com/34196718/196110459-fea9cba4-5f22-4afd-95a3-8b7ed7ff3e32.png)

Therefore, we should be able to utilize our existing arithmetic accumulator system in the log scale.

As the first step, I propose finalizing the logarithm implementation: https://github.com/osmosis-labs/osmosis/pull/2788

Next, integrate the geometric TWAP into the `twap` module by utilizing the log-average relationship described above.

## Core Progress

- https://github.com/osmosis-labs/osmosis/pull/3420
- https://github.com/osmosis-labs/osmosis/pull/3620
- https://github.com/osmosis-labs/osmosis/pull/3676
- https://github.com/osmosis-labs/osmosis/issues/3443
- https://github.com/osmosis-labs/osmosis/issues/3514
- https://github.com/osmosis-labs/osmosis/issues/3540
- https://github.com/osmosis-labs/osmosis/issues/3541
- https://github.com/osmosis-labs/osmosis/issues/3542
- https://github.com/osmosis-labs/osmosis/issues/3543
- https://github.com/osmosis-labs/osmosis/issues/3728
- https://github.com/osmosis-labs/osmosis/issues/3770
- https://github.com/osmosis-labs/osmosis/issues/3941

## Acceptance Criteria
- Geometric TWAP is implemented using log-average
- Logarithm implementation is efficient and accurate
- The log-average rounding is insignificant
- Integration into the existing TWAP module is sound

## Sources
- https://en.wikipedia.org/wiki/Geometric_mean#Relationship_with_logarithms
- https://en.wikipedia.org/wiki/Geometric_Brownian_motion#Use_in_finance
- https://uniswap.org/whitepaper-v3.pdf

p0mvn:
Some notes on the investigations so far:

[...]

PR #3420, opened 2022-11-17
feat(x/twap): Geometric Mean TWAP POC

p0mvn:
Part of: #3013

This is a proof-of-concept Geometric Mean TWAP implementation.

While the arithmetic mean TWAPs are much more widely used, they should theoretically be less accurate in measuring a geometric Brownian motion process (which is how price movements are usually modeled)

Arithmetic TWAP tends to overweight higher prices relative to lower ones.

Therefore, we also support a geometric mean TWAP.

The core functionality stays similar to the arithmetic mean TWAP. However, instead of computing the geometric mean TWAP naively, we use the following property:

`GeometricTwap(P) = 1.0001^{ArithmeticTwap(log_{1.0001}{P})}`

Naive computation is expensive and easily overflows. As a result, we track logarithms of prices instead of prices themselves in the accumulators.
When geometric twap is requested, we first compute the arithmetic mean of the logarithms, and then exponentiate it with the same base as the logarithm

## Brief Changelog

- added geometric twap accumulator to the twap record
- implemented geometric accumulator update
- refactored `computeArithmeticTwap` method to be more general and named it `computeTwap`
- implemented `computedGeometricTwap` method that is called by `computeTwap`
- added tests

## Testing and Verifying

This change added tests.

### Blocking TODOs

- overflow tests, potentially choose different log base and improve `Pow()` function
- https://github.com/osmosis-labs/osmosis/issues/3443
- hand-calculated tests for `computeGeometricTwap`
- twap record migrations in the upgrade handler

### Clean up TODOs

- address all TODOs for prices and add pruning tests
- spec and tests for osmomath Dec's `TwapLog`


x/twap/README.md
```diff
@@ -12,6 +12,22 @@ Notice that the latest price `p_n` isn't used, as it has lasted for a time inter
...
+
+Therefore, we also support a geometric mean TWAP.
+
+The core functionality stays similar to the arithmetic mean TWAP. However, instead of computing the geometric mean TWAP naively, we use the following property:
+
+$$GeometricMean(P) = 1.0001^{ArithmeticMean(log_{1.0001}{P})}$$
+
+Naive computation is expensive and easily overflows. As a result, we track logarithms of prices instead of prices themselves in the accumulators.
+When geometric twap is requested, we first compute the arithmetic mean of the logarithms, and then exponentiate it with the same base as the logarithm
+to get the final result.
```
nicolaslara, line 29: Can we add a line or two about why this holds? 

IIUC, it's basically:

$$  {(\prod_{i=a}^{b} P_i)}^{\frac{1}\{b-a}}  =  exp(\sum_{i=a}^{b}{\frac{1}{b-a} log_{1.0001}{(P_i)}}) = exp(log_{1.0001}{(\frac{1}{b-a}\sum{P_i})})$$

nicolaslara: Just realized this is the case by definition. So maybe we don't need it

p0mvn: The only difference is that, currently, we take 1.0001 to that power, not Euler's number. It might change to a larger value down the road due to the risk of overflows with such a small base for the logarithm. A small base leads us to have a large exponent in `Power()` function that might overflow

Yeah while I tried to explain in words, I agree that it's nice to have the formula.

I didn't add it in the first place because having so many issues with rendering math on GitHub 90% of the time they get rendered on Notion but not here 😢

p0mvn: The base of `1.0001` wouldn't format so I added this:

https://github.com/osmosis-labs/osmosis/pull/3420/commits/6fd661d33bd9ce8ae44a711fcb95410acef4c821

nicolaslara: yeah, I took `exp()` to represent $$exp_{1.0001}()$$ here just to avoid writing it. That makes sense (and the comment on the new commit looks great!)

x/twap/logic.go
```diff
@@ -190,6 +192,12 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 	p1NewAccum := types.SpotPriceMulDuration(record.P1LastSpotPrice, timeDelta)
 	newRecord.P1ArithmeticTwapAccumulator = newRecord.P1ArithmeticTwapAccumulator.Add(p1NewAccum)
 
+	// logP0SpotPrice = log_{1.0001}{P_0}
+	logP0SpotPrice := osmomath.TwapLog(record.P0LastSpotPrice)
+	// p0NewGeomAccum = log_{1.0001}{P_0} * timeDelta
+	p0NewGeomAccum := types.SpotPriceMulDuration(logP0SpotPrice, timeDelta)
+	newRecord.GeometricTwapAccumulator = newRecord.GeometricTwapAccumulator.Add(p0NewGeomAccum)
```
nicolaslara, line 199: Why is it enough to store this for P0? Is it cause we can later obtain p1 via `1/p0`?

x/twap/logic.go
```diff
@@ -219,15 +227,16 @@ func (k Keeper) getMostRecentRecord(ctx sdk.Context, poolId uint64, assetA, asse
...
+// type - arithmetic or geometric.
+// Between two records given the quote asset.
 // precondition: endRecord.Time >= startRecord.Time
 // if (endRecord.LastErrorTime >= startRecord.Time) returns an error at end + result
 // if (startRecord.LastErrorTime == startRecord.Time) returns an error at end + result
 // if (endRecord.Time == startRecord.Time) returns endRecord.LastSpotPrice
 // else returns
 // (endRecord.Accumulator - startRecord.Accumulator) / (endRecord.Time - startRecord.Time)
-func computeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quoteAsset string) (sdk.Dec, error) {
+func computeTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quoteAsset string, twapType twapType) (sdk.Dec, error) {
```
nicolaslara, line 239: In this function we short circuit if timeDelta == 0 and return `endRecord.PxLastSpotPrice`. Shouldn't that be different for Geometric twap? i.e.:  `osmomath.Pow(osmomath.Tick, endRecord.PxLastSpotPrice)`

x/twap/logic.go
p0mvn: Exactly. We use this property: Geometric mean of the reciprocals is the reciprocal of the geometric mean.

I attached the proof in the computeGeometricTwap function.

osmomath/math.go
```diff
@@ -171,3 +173,8 @@ func PowApprox(base sdk.Dec, exp sdk.Dec, precision sdk.Dec) sdk.Dec {
 	}
 	return sum
 }
+
+// TODO: spec and tests
+func TwapLog(x sdk.Dec) sdk.Dec {
+	return BigDecFromSDKDec(x).TickLog().SDKDec()
```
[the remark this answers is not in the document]
p0mvn: TickLog refers to 1.0001 base log which we call as tick.

I'm open to suggestions. This is the best semantics I could think of to refer to this specific choice of base.

x/twap/logic.go
p0mvn: The idea is that if time delta from the last sample is 0, we return the sample itself.

TWAP is an estimate of the true price. If the request lands at the time of sample, the most accurate estimate is the price itself.

So I think this is what we want in both cases

p0mvn: The last price is not stored in log scale. Only the accumulator is. So returning the actual price should be what we want

nicolaslara: ahh, of course... that makes sense.

x/twap/logic.go
```diff
@@ -243,11 +252,43 @@ func computeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRec
 		}
 		return endRecord.P1LastSpotPrice, err
 	}
+
+	if twapType == arithmeticTwapType {
+		return computeArithmeticTwap(startRecord, endRecord, quoteAsset), err
+	}
+	return computeGeometricTwap(startRecord, endRecord, quoteAsset), err
+}
+
```
czarcas7ic, line 261: we should explicitly take twapType == geometricTwapType as and else if here. Otherwise we might accidentally compute geomtric twap for an invalid type input

x/twap/store_test.go
```diff
@@ -162,11 +162,11 @@ func (s *TestSuite) TestGetRecordAtOrBeforeTime() {
...
 	wrongPoolIdInputAt := func(t time.Time) getRecordInput { return getRecordInput{2, t, denom0, denom1} }
 	defaultRevInputAt := func(t time.Time) getRecordInput { return getRecordInput{1, t, denom1, denom0} }
-	baseRecord := newEmptyPriceRecord(1, baseTime, denom0, denom1)
+	baseRecord := withPrice0Set(newEmptyPriceRecord(1, baseTime, denom0, denom1), sdk.OneDec())
 	tMin1 := baseTime.Add(-time.Second)
-	tMin1Record := newEmptyPriceRecord(1, tMin1, denom0, denom1)
+	tMin1Record := withPrice0Set(newEmptyPriceRecord(1, tMin1, denom0, denom1), sdk.OneDec())
 	tPlus1 := baseTime.Add(time.Second)
-	tPlus1Record := newEmptyPriceRecord(1, tPlus1, denom0, denom1)
+	tPlus1Record := withPrice0Set(newEmptyPriceRecord(1, tPlus1, denom0, denom1), sdk.OneDec())
```
czarcas7ic, line 169: why are we now setting price0 instead of initializing with empty price record?

czarcas7ic (commented):
Still grasping the concepts, but overall logic appears sound to me

osmomath/math.go
```diff
@@ -15,6 +15,8 @@ var powPrecision, _ = sdk.NewDecFromStr("0.00000001")
 var zero sdk.Dec = sdk.ZeroDec()
 
 var (
+	Tick = sdk.MustNewDecFromStr("1.0001")
```
ValarDragon, line 18: Why are we doing logs respective to this base, rather than just 2

osmomath/math.go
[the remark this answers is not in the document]
ValarDragon: Agreed, I also recommend we just do log2

x/twap/README.md
```diff
@@ -12,6 +12,28 @@ Notice that the latest price `p_n` isn't used, as it has lasted for a time inter
...
 
+## Geometric mean TWAP
+
+While the arithmetic mean TWAPs are much more widely used, they should theoretically be less accurate in measuring a geometric Brownian motion process (which is how price movements are usually modeled)
+
+Arithmetic TWAP tends to overweight higher prices relative to lower ones.
+
+Therefore, we also support a geometric mean TWAP.
+
+The core functionality stays similar to the arithmetic mean TWAP. However, instead of computing the geometric mean TWAP naively, we use the following property:
```
ValarDragon, line 23: First the naive should be defined

Then how we compute it =p

x/twap/README.md
```diff
@@ -12,6 +12,28 @@ Notice that the latest price `p_n` isn't used, as it has lasted for a time inter
...
+
+
+$$GeometricMean(P) = 1.0001^{ArithmeticMean(log_{1.0001}{P})}$$
```
ValarDragon, line 26: Very confused as to why the choice of this base

osmomath/math.go
p0mvn: I'm tracking this as one of the "Blocking TODOs" in the PR description to revisit next.

I've been following uniswap whitepaper. However, I've also come to the conclusion that we don't need this choice of base and plan to choose the new one while working on overflow tests.

With the base of 2, I've also run into overflow problems so I would like to tackle the correct choice of base next in a separate PR.

x/twap/api_test.go
```diff
@@ -52,21 +62,27 @@ var (
...
 		baseTime.Add(20*time.Second),
 		sdk.NewDec(2),                // spot price 0
 		OneSec.MulInt64(10*10+5*10),  // accum A
 		OneSec.MulInt64(3),           // accum B
-		OneSec.MulInt64(20*10+10*10)) // accum C
+		OneSec.MulInt64(20*10+10*10), // accum C
+		sdk.ZeroDec(),                // TODO: choose correct
+		sdk.ZeroDec(),                // TODO: choose correct
+		sdk.ZeroDec(),                // TODO: choose correct
+	)
```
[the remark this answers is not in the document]
p0mvn: I need to add more tests where these values would be used.

Currently, this is set to zero to avoid existing arithmetic twap tests from failing due to geometric accumulators not being initialized.

x/twap/store_test.go
p0mvn: price0 of 0 would panic because the logarithm of 0 is undefined.

This is safe because, logically, spot price of 0 should be undefined as well

x/twap/logic.go
p0mvn: Good point. I changed `twapType` to be a boolean so this is not possible anymore

osmomath/math.go
p0mvn: Changed base to 2 but still might need to revisit once overflow tests are added

x/twap/api_test.go
```diff
@@ -22,28 +22,38 @@ var (
...
 
 	// accumA = 10 seconds * (spot price = 10) = OneSec * 10 * 10
 	// accumB = 10 seconds * (spot price = 0.1) = OneSec
 	// accumC = 10 seconds * (spot price = 20) = OneSec * 10 * 20
 	accumA, accumB, accumC sdk.Dec = OneSec.MulInt64(10 * 10), OneSec, OneSec.MulInt64(10 * 20)
 
+	// geomAccumAB = 10 seconds * (log_{1.0001}{spot price = 10})
+	geomAccumAB = geometricTenSecAccum.MulInt64(10)
+	geomAccumAC = geomAccumAB
```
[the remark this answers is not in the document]
p0mvn: It's due to how our tests are set up. Specifically, `newThreeAssetPoolTwapRecordWithDefaults`

x/twap/README.md
p0mvn: Changed

x/twap/README.md
p0mvn: [36841f5](https://github.com/osmosis-labs/osmosis/pull/3420/commits/36841f5ee92e1370aeb98ced873b9faa7b3d8ceb)

p0mvn:
@czarcas7ic @ValarDragon @nicolaslara all requested changes are added

osmomath/math.go
ValarDragon: SG! (FWIW im not at all married to the base of 2! Just if we don't have a reason for any particular base, or some number being superior to 2, would prefer to default to 2)

czarcas7ic (approved):
LGTM as first step, acknowledging the remaining TODOs. Nice work 👍

p0mvn:
While this is not urgent, this PR is still waiting for one more approval

x/twap/logic.go
```diff
@@ -190,6 +201,12 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 	p1NewAccum := types.SpotPriceMulDuration(record.P1LastSpotPrice, timeDelta)
 	newRecord.P1ArithmeticTwapAccumulator = newRecord.P1ArithmeticTwapAccumulator.Add(p1NewAccum)
 
+	// logP0SpotPrice = log_{1.0001}{P_0}
```
stackman27, line 204: nit: shouldn't the comment be log_2{P_0}

x/twap/logic.go
```diff
@@ -243,11 +261,55 @@ func computeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRec
...
+
+// twapLog returns the logarithm of the given spot price, base 2.
+// TODO: basic test
+func twapLog(price sdk.Dec) sdk.Dec {
+	return osmomath.BigDecFromSDKDec(price).LogBase2().SDKDec()
+}
+
+// twapPow exponentiates the geometricTwapMathBase to the given exponent.
+// TODO: basic test and benchmark.
+func twapPow(exponent sdk.Dec) sdk.Dec {
```
stackman27, line 313: just want to confirm we are exponentiating by e right? so that our result is something like `e^(exponent)`

osmomath/math.go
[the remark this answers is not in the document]
stackman27: I'm trying to understand this why was 1.0001 base log used instead of log_2 in first place?

x/twap/logic.go
p0mvn: The base for both power and log is `2`. So it is $$2^{ArithmeticMean(log_2{Prices})}$$

x/twap/logic.go
```diff
@@ -243,11 +261,55 @@ func computeArithmeticTwap(startRecord types.TwapRecord, endRecord types.TwapRec
...
+// - compute arithmetic mean of logarithms of spot prices.
+// - exponentiate the result to get the geometric mean.
+// - if quoted asset is asset 1, take reciprocal of the exponentiated result.
+func computeGeometricTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quoteAsset string) sdk.Dec {
+	accumDiff := endRecord.GeometricTwapAccumulator.Sub(startRecord.GeometricTwapAccumulator)
+
+	timeDelta := endRecord.Time.Sub(startRecord.Time)
+	arithmeticMeanOfLogPrices := types.AccumDiffDivDuration(accumDiff, timeDelta)
+
+	geometrciMeanDenom0 := twapPow(arithmeticMeanOfLogPrices)
```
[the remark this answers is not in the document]
p0mvn: addressed

x/twap/logic.go
p0mvn: addressed

stackman27 (approved):
LGTM!

p0mvn:
Merging this to continue making progress on geometrict TWAP. The next items are tracked in: https://github.com/osmosis-labs/osmosis/issues/3013

PR #3440, opened 2022-11-17
chore: add param query to ibc rate limit

czarcas7ic:
Closes: #XXX

ibc-rate-limit had a v10 import path. This fixes that as well as adds the param query to CLI


p0mvn (approved):

ValarDragon (approved):
LGTM!

PR #3442, opened 2022-11-17
clean up stableswap comments

p0mvn:
Closes: #XXX

- Remove dead code
- Elaborate on godoc
- Add TODO to clarify solver spec


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -76,6 +76,8 @@ func cfmmConstantMulti(xReserve, yReserve, u, v osmomath.BigDec) osmomath.BigDec
 // xy(x^2 + y^2) = (x - a)(y + b)((x - a)^2 + (y + b)^2)
 // and the following expression for `a` in multi-asset pools:
 // xyz(x^2 + y^2 + w) = (x - a)(y + b)z((x - a)^2 + (y + b)^2 + w)
+// TODO: explain where z comes from, it seems that we only use sum squares
+// in solveCFMMBinarySearchMulti (w).
```
p0mvn, line 80: I thought that we are solving `h(x,y,w)` w/o `z`:

Ref:
https://github.com/osmosis-labs/osmosis/blob/593f0d2c101f2dc970255366333291a6c6a66203/x/gamm/pool-models/stableswap/README.md?plain=1#L129-L130

ValarDragon: Ah yeah that z is the doc is an accident, we need to delete it. Its not used within the function

ValarDragon, line 80:
```suggestion
// xy(x^2 + y^2) = (x - a)(y + b)((x - a)^2 + (y + b)^2 + w)
// with w set to 0 for 2 asset pools
```

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -317,7 +315,7 @@ func solveCFMMBinarySearchMulti(xReserve, yReserve, wSumSquares, yIn osmomath.Bi
 	} else if yIn.Abs().GTE(yReserve) {
 		panic("cannot input more than pool reserves")
 	}
-	// fmt.Printf("solve cfmm xreserve %v, yreserve %v, w %v, yin %v\n", xReserve, yReserve, wSumSquares, yIn)
```
ValarDragon, line 320: do you think its bad form to leave this println's? I find them helpful when going through and debugging, to already have something w/ the info I need available

p0mvn: I think that any form of dead code adds some level of complexity. Also, I usually inspect variables with a debugger and add them to watch so the value of prints wasn't obvious to me.

However, not a strong opinion. I can revert if you find it useful.

ValarDragon: Can we revert?

p0mvn: okay done

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -73,9 +73,8 @@ func cfmmConstantMulti(xReserve, yReserve, u, v osmomath.BigDec) osmomath.BigDec
 // So we want to solve for a given addition of `b` units of y into the pool,
 // how many units `a` of x do we get out.
 // So we solve the following expression for `a` in two-asset pools:
```
ValarDragon, line 75:
```suggestion
// So we solve the following expression for `a`:
```

ValarDragon (approved):

PR #3458, opened 2022-11-18
Update our SDK fork

ValarDragon:

Update our SDK fork to revert the changes to cache Ctx for writing events there. While the changes are right, it changes events for IBC (adding duplicates), as IBC did a workaround similar to us. 

External things depend on IBC events, so the duplicate events causes problems for them. This is a blocker for updating to cosmwasm v0.29

After this PR, I'll PR an update to cosmwasm v0.29.x


p0mvn (approved):
confirmed hash

PR #3464, opened 2022-11-19
[x/gamm][stableswap]: Fix swap fee charge on single asset joins

AlpinYukseloglu:

This PR fixes a bug where swap fees for single asset joins are charged on the full amount instead of just the portion that is "swapped."

## Brief Changelog

- Update single asset join share calculation to only charge swap fee on the portion of the input tokens made of non-input assets
- Update scaling factor validation logic to prevent edge case where individual scaling factors pass but the sum of scaling factors overflows or turns negative when converted to int64

## Testing and Verifying

This PR passes our single asset join test suite and inverse relation tests, which ensure that any errors favor the pool. The exact swap fee charged is not directly tested.

### DIFF
--- a/x/gamm/pool-models/stableswap/amm.go
+++ b/x/gamm/pool-models/stableswap/amm.go
@@ -441,9 +441,24 @@ func (p *Pool) calcSingleAssetJoinShares(tokenIn sdk.Coin, swapFee sdk.Dec) (sdk
 		return &paCopy
 	}
 
-	// We apply the swap fee by multiplying by (1 - swapFee) and then truncating to int
-	oneMinusSwapFee := sdk.OneDec().Sub(swapFee)
-	tokenInAmtAfterFee := tokenIn.Amount.ToDec().Mul(oneMinusSwapFee).TruncateInt()
+	// Find the portion of the pool made up by non-input assets
+	sumScalingFactors := int64(0)
+	for _, scalingFactor := range p.ScalingFactors {
+		sumScalingFactors += int64(scalingFactor)
+	}
+
+	// We ensure that uint64 -> int64 conversions are safe in scaling factor validation logic
+	liquidityIndexes := p.getLiquidityIndexMap()
+	inputScalingFactor := int64(p.GetScalingFactorByLiquidityIndex(liquidityIndexes[tokenIn.Denom]))
+
+	// Round the ratio up to overestimate the charged fee
+	nonInputRatio := (sdk.NewDec(sumScalingFactors).Sub(sdk.NewDec(inputScalingFactor))).Quo(sdk.NewDec(sumScalingFactors)).Ceil()
+
+	// Find amount charged in swap fee on non-input assets and subtract from total
+	chargedSwapFee := (nonInputRatio.Mul(tokenIn.Amount.ToDec())).Mul(swapFee)
+
+	// We truncate to ensure we always underestimate the number of calculated shares (by passing in a lower input)
+	tokenInAmtAfterFee := tokenIn.Amount.ToDec().Sub(chargedSwapFee).TruncateInt()
 
 	return cfmm_common.BinarySearchSingleAssetJoin(p, sdk.NewCoin(tokenIn.Denom, tokenInAmtAfterFee), poolWithAddedLiquidityAndShares)
 }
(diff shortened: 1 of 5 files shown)


x/gamm/pool-models/stableswap/pool.go
```diff
@@ -442,7 +442,7 @@ func validateScalingFactors(scalingFactors []uint64, numAssets int) error {
 	}
 
 	for _, scalingFactor := range scalingFactors {
-		if scalingFactor == 0 || int64(scalingFactor) <= 0 {
+		if scalingFactor == 0 || scalingFactor >= (1<<(62-types.MaxPoolAssets)) {
```
AlpinYukseloglu, line 445: Note to reviewer: this update overshoots by a little bit (adds a 1 bit buffer per asset when we actually only need log(maxNumAssets)), but I think it's more important to eliminate the class of scaling factor-related bugs than to get an exact bound since we don't expect people to use up that many bits for scaling factors anyway

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -441,9 +441,24 @@ func (p *Pool) calcSingleAssetJoinShares(tokenIn sdk.Coin, swapFee sdk.Dec) (sdk
 		return &paCopy
 	}
 
-	// We apply the swap fee by multiplying by (1 - swapFee) and then truncating to int
-	oneMinusSwapFee := sdk.OneDec().Sub(swapFee)
-	tokenInAmtAfterFee := tokenIn.Amount.ToDec().Mul(oneMinusSwapFee).TruncateInt()
+	// Find the portion of the pool made up by non-input assets
+	sumScalingFactors := int64(0)
```
p0mvn, line 445: Consider encapsulating the new logic into a function with a godoc and adding a couple of basic tests

p0mvn, line 445: Not sure if there are more reasons for performing these conversions but if there aren't any, we could just use `uint64` throughout function

```suggestion
	sumScalingFactors := uint64(0)
```

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -441,9 +441,24 @@ func (p *Pool) calcSingleAssetJoinShares(tokenIn sdk.Coin, swapFee sdk.Dec) (sdk
...
+	for _, scalingFactor := range p.ScalingFactors {
+		sumScalingFactors += int64(scalingFactor)
+	}
+
+	// We ensure that uint64 -> int64 conversions are safe in scaling factor validation logic
+	liquidityIndexes := p.getLiquidityIndexMap()
+	inputScalingFactor := int64(p.GetScalingFactorByLiquidityIndex(liquidityIndexes[tokenIn.Denom]))
```
p0mvn, line 452:
```suggestion
	inputScalingFactor := p.GetScalingFactorByLiquidityIndex(liquidityIndexes[tokenIn.Denom])
```

x/gamm/pool-models/stableswap/amm.go
```diff
@@ -441,9 +441,24 @@ func (p *Pool) calcSingleAssetJoinShares(tokenIn sdk.Coin, swapFee sdk.Dec) (sdk
...
+
+	// Round the ratio up to overestimate the charged fee
+	nonInputRatio := (sdk.NewDec(sumScalingFactors).Sub(sdk.NewDec(inputScalingFactor))).Quo(sdk.NewDec(sumScalingFactors)).Ceil()
```
p0mvn, line 455:
```suggestion
	nonInputRatio := (sdk.NewDecFromInt(sdk.NewIntFromUint64(sumScalingFactors)).Sub(sdk.NewDec(inputScalingFactor))).Quo(sdk.NewDecFromInt(sdk.NewIntFromUint64(sumScalingFactors))).Ceil()
```

x/gamm/pool-models/stableswap/amm.go
p0mvn: Suggest elaborating in the spec on this point: `only charge swap fee on the portion of the input tokens made of non-input assets`. Also, answering why we need to charge the fee only on the proportion of non-input assets would be helpful.

ValarDragon:
I think this needs to be based on liquidity ratio, post-scaling. Rather than the scaling factors, which would assume a 1:1 price peg for swap ratios.

https://github.com/osmosis-labs/osmosis/pull/3478

ValarDragon:
Closing in favor of 3478

PR #3468, opened 2022-11-20
refactor(txfees): remove redundant keeper fields and constructor params

p0mvn:
Closes: #XXX

Removed obsolete:
- `feeCollectorName
- `nonNativeFeeCollectorName`

These are defined globals in `types`. The keeper fields were left unused


ValarDragon:
Hrmm, actually I think its a bug that were using the constant instead of the keeper value in the live code! At least for the fee collector name

ValarDragon (approved):

PR #3475, opened 2022-11-21
fix: initialize last gauge ID in init gen

czarcas7ic:
Closes: #XXX

We forget to initialize LastGaugeID despite properly exporting it. This properly initializes it with the exported param.


ValarDragon (approved):
Pretty embarassing that we missed this lol

PR #3477, opened 2022-11-21
chore: fix import paths CI vendoring failure

p0mvn:
Closes: #XXX

In some cases `make proto-gen` modifies `go,mod`. As a result, that causes subsequent query-gen step to fail:
https://github.com/osmosis-labs/osmosis/actions/runs/3518190249/jobs/5896827173

Therefore, we should revendor after running proto-gen

## Testing and Verifying

Tested the script locally


ValarDragon (approved):
Oh wow, interesting (and annoying) interaction

PR #3478, opened 2022-11-21
Fix swapfee for single asset joins

ValarDragon:

We were accidentally over-charging swap fee for single asset joins. This PR causes us to no longer significantly over-charge. We now charge swapfee at `swapfee * fraction of pool ratio that is not the token your joining`. This is a slight over-charge, as your swap does change this ratio itself, but this is quite low. (Namely bips on the % change in liquidity ratio itself)

## Brief Changelog

Reduce the amount of stableswap singleassetjoin swap fee

## Testing and Verifying

* Updates the test that ensures that we don't over-return funds on exit.
* Added tests that we return what is expected from the new internal function.

We should add more tests that:
* we don't return an insufficient amount, but I have spot-checked manually that this is tightly-close to the intended bound.
* Increase coverage of the join use case
* Split out single asset join test setup from multi-asset joins

But I leave those to future PR's.

  - How is the feature or change documented? Spec

### DIFF
--- a/x/gamm/pool-models/internal/cfmm_common/lp.go
+++ b/x/gamm/pool-models/internal/cfmm_common/lp.go
@@ -151,7 +151,7 @@ func BinarySearchSingleAssetJoin(
 			return sdk.Int{}, err
 		}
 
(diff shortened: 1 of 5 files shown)


AlpinYukseloglu (approved):
Nice, LGTM!

PR #3490, opened 2022-11-22
CL: add swaprouter tests

czarcas7ic:
Closes: #XXX

Adds swaprouter tests ported from main

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -59,6 +59,16 @@ func (s *KeeperTestHelper) PrepareBalancerPoolWithCoins(coins ...sdk.Coin) uint6
 	return s.PrepareBalancerPoolWithCoinsAndWeights(coins, weights)
 }
 
+// PrepareBalancerPoolWithCoinsAndSwapFee returns a balancer pool
+// consisted of given coins with equal weight set with provided swap fee.
+func (s *KeeperTestHelper) PrepareBalancerPoolWithCoinsAndSwapFee(swapFee sdk.Dec, coins ...sdk.Coin) uint64 {
+	weights := make([]int64, len(coins))
+	for i := 0; i < len(coins); i++ {
+		weights[i] = 1
+	}
+	return s.PrepareBalancerPoolWithCoinsWeightsAndSwapFee(coins, weights, swapFee)
+}
+
 // PrepareBalancerPoolWithCoins returns a balancer pool
 // PrepareBalancerPoolWithCoinsAndWeights returns a balancer pool
 // consisted of given coins with the specified weights.
@@ -73,6 +83,22 @@ func (s *KeeperTestHelper) PrepareBalancerPoolWithCoinsAndWeights(coins sdk.Coin
 	}
 
 	return s.PrepareBalancerPoolWithPoolAsset(poolAssets)
+}
+
+// PrepareBalancerPoolWithCoins returns a balancer pool
+// PrepareBalancerPoolWithCoinsWeightsAndSwapFee returns a balancer pool
+// consisted of given coins with the specified weights and swap fee.
+func (s *KeeperTestHelper) PrepareBalancerPoolWithCoinsWeightsAndSwapFee(coins sdk.Coins, weights []int64, swapFee sdk.Dec) uint64 {
+	var poolAssets []balancer.PoolAsset
+	for i, coin := range coins {
+		poolAsset := balancer.PoolAsset{
+			Weight: sdk.NewInt(weights[i]),
+			Token:  coin,
+		}
+		poolAssets = append(poolAssets, poolAsset)
+	}
+
+	return s.PrepareBalancerPoolWithPoolAssetAndSwapFee(poolAssets, swapFee)
 }
 
 // PrepareBalancerPool returns a Balancer pool's pool-ID with pool params set in PrepareBalancerPoolWithPoolParams.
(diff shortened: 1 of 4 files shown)


czarcas7ic:
@p0mvn the only thing I am not satisfied with in this PR was the inability to use SetPool from gamm, so I had to make a bunch of different functions to set the swap fee. I had originally exported setPool to SetPool but wasnt sure if that was desired. If that is desired, I will happily undo this specific change and directly SetPool rather than this current work around

x/swaprouter/router_test.go
```diff
@@ -88,30 +245,77 @@ func (suite *KeeperTestSuite) TestMultihopSwapExactAmountIn() {
...
+						tokenOut, err := gammKeeper.SwapExactAmountIn(cacheCtx, suite.TestAccs[0], hopPool, nextTokenIn, hop.TokenOutDenom, sdk.OneInt(), updatedPoolSwapFee)
+						suite.Require().NoError(err)
+						nextTokenIn = sdk.NewCoin(hop.TokenOutDenom, tokenOut)
+					}
+					return nextTokenIn
+				}
+			}
 
 			if tc.expectError {
+				_, err := swaprouterKeeper.RouteExactAmountIn(suite.Ctx, suite.TestAccs[0], tc.routes, tc.tokenIn, tc.tokenOutMinAmount)
```
p0mvn, line 309: We can move this up and share between if / else branches

x/swaprouter/router_test.go
```diff
@@ -33,45 +36,199 @@ func (suite *KeeperTestSuite) TestMultihopSwapExactAmountIn() {
...
+				},
+				{
+					PoolId:        3,
+					TokenOutDenom: "bar",
+				},
+			},
+			incentivizedGauges:      []uint64{1, 2, 3, 4, 5, 6, 7, 8, 9},
+			tokenIn:                 sdk.NewCoin("foo", sdk.NewInt(100000)),
+			tokenOutMinAmount:       sdk.NewInt(1),
+			expectReducedFeeApplied: false,
```
p0mvn, line 231: Please remove TODOs below that are addressed in this PR

p0mvn, line 231: Did we have any tests for stableswap pools?

If not, could you add a TODO to add stableswap test cases please

x/swaprouter/router_test.go
```diff
@@ -161,30 +539,79 @@ func (suite *KeeperTestSuite) TestMultihopSwapExactAmountOut() {
...
-			// if we expect a reduced fee to apply, we set both pools in DistrInfo to replicate it being an incentivized pool
-			// each pool has three gauges, hence 6 gauges for 2 pools
-			if tc.reducedFeeApplied {
-				test := poolincentivestypes.DistrInfo{
-					TotalWeight: sdk.NewInt(2),
-					Records: []poolincentivestypes.DistrRecord{
-						{GaugeId: 1, Weight: sdk.OneInt()}, {GaugeId: 2, Weight: sdk.OneInt()}, {GaugeId: 3, Weight: sdk.OneInt()},
-						{GaugeId: 4, Weight: sdk.OneInt()}, {GaugeId: 5, Weight: sdk.OneInt()}, {GaugeId: 6, Weight: sdk.OneInt()}},
+			// if test specifies incentivized gauges, set them here
+			if len(tc.incentivizedGauges) > 0 {
```
p0mvn, line 548: Can this be moved into a helper method defined on the suite and shared between tests?

x/swaprouter/router_test.go
```diff
@@ -161,30 +539,79 @@ func (suite *KeeperTestSuite) TestMultihopSwapExactAmountOut() {
...
+					TotalWeight: totalWeight,
+					Records:     records,
+				}
+				suite.App.PoolIncentivesKeeper.SetDistrInfo(suite.Ctx, distInfo)
 			}
 
-			tokenIn, err := swaprouterKeeper.RouteExactAmountOut(suite.Ctx, suite.TestAccs[0], tc.routes, tc.tokenInMaxAmount, tc.tokenOut)
+			// calcOutAmountAsSeparateSwaps calculates the multi-hop swap as separate swaps, while also
+			// utilizing a cacheContext so the state does not change
+			calcOutAmountAsSeparateSwaps := func(osmoFeeReduced bool) sdk.Coin {
```
p0mvn, line 563: Consider moving this function definition either to be defined at the top of the test fixture or as a separate helper function in the test-global scope.

Similar suggestion for the other test.

The reason is to keep the actual test simple and easy to follow

x/swaprouter/router_test.go
```diff
@@ -125,27 +329,201 @@ func (suite *KeeperTestSuite) TestMultihopSwapExactAmountIn() {
...
+				{
+					PoolId:       3,
+					TokenInDenom: "baz",
+				},
+			},
+			incentivizedGauges:      []uint64{1, 2, 3, 4, 5, 6, 7, 8, 9},
+			tokenOut:                sdk.NewCoin("bar", sdk.NewInt(100000)),
+			tokenInMaxAmount:        sdk.NewInt(90000000),
+			expectReducedFeeApplied: false,
+		},
```
p0mvn, line 526: ditto for TODOs:
- remove addressed
- add TODO for stableswap

x/swaprouter/router_test.go
```diff
@@ -6,19 +6,22 @@ import (
...
 )
 
 const (
-	denomA = "denomA"
-	denomB = "denomB"
-	denomC = "denomC"
+	foo   = "foo"
+	bar   = "bar"
+	baz   = "baz"
+	uosmo = "uosmo"
```
p0mvn, line 19: nit: noticed that there are places where hardcoded values are still used. Please make sure to use these variables everywhere where applicable

app/apptesting/gamm.go
```diff
@@ -156,6 +182,24 @@ func (s *KeeperTestHelper) PrepareBalancerPoolWithPoolAsset(assets []balancer.Po
 	return poolId
 }
 
+// PrepareBalancerPoolWithPoolAssetAndSwapFee sets up a Balancer pool with an array of assets and a set swap fee.
+func (s *KeeperTestHelper) PrepareBalancerPoolWithPoolAssetAndSwapFee(assets []balancer.PoolAsset, swapFee sdk.Dec) uint64 {
```
p0mvn, line 186: Just commenting to think about this in the future.

There are so many variations of these helpers now. We should think about reducing their number while keeping the same functionality.

p0mvn (approved):
Great work. Thank you

p0mvn:
> 

I think it's better to make more test-specific helpers than expose API just for the purpose of testing. So I agree with the approach here

x/swaprouter/router_test.go
czarcas7ic: We can only do this if we use a cache context for routeExactAmount in as well, otherwise the second swap that happens happens with an unbalanced pool. I personally would rather use cache ctx for the expected and then run the actual swap after and compare the results. lmk what you think

x/swaprouter/router_test.go
czarcas7ic: Good call 👍

app/apptesting/gamm.go
czarcas7ic: Agreed, should I make a "good first issue" to combine them into a single helper that allows modification of all pool traits?

p0mvn: Yes! That would be great, thank you!

x/swaprouter/router_test.go
p0mvn: Ahh I see. Wasn't aware of this constraint. I'm fine with whatever you think is best

p0mvn (approved):

Issue #3240, opened 2022-11-04
[CL][Misc]: consider removing "pool_count" index and `NumPools` query from `gamm` 


p0mvn:
## Background

It is questionable whether `NumPools` query in `x/gamm` has any use to clients.
Currently, we store an additional index to keep track of this count.

## Suggested Design

We should consider removing this query from `x/gamm` and the `pool_count` store index.

## Acceptance Criteria
- Decision is made
- If decided to remove, all of the store index, relevant methods, queries and CLI are deleted

p0mvn:
If no argument for keeping the "NumPools" query in `x/gamm` is made in the next few days, I propose we remove it.

hieuvubk:
Hi if we got final decision about this issue, I'm willing to work with it

PR #3492, opened 2022-11-22
refactor(CL): remove NumPools query from gamm

p0mvn:
Closes: #3240

Since we are planning to merge gamm refactor to `main` soon, it might be best to apply this change prior to that so that the extra complexity stemming from `NumPools` query is not present in `main`.

Context: https://github.com/osmosis-labs/osmosis/issues/3240


czarcas7ic:
It appears the 'osmosisd tx swaprouter create-pool' command is failing in e2e

x/gamm/keeper/genesis_test.go
```diff
@@ -111,7 +109,6 @@ func TestGammExportGenesis(t *testing.T) {
 
 	genesis := app.GAMMKeeper.ExportGenesis(ctx)
 	require.Len(t, genesis.Pools, 2)
-	require.Equal(t, genesis.PoolCount, uint64(2))
```
czarcas7ic, line 114: shouldn't we be checking NextPoolID here?

x/gamm/simulation/sim_msgs.go
```diff
@@ -297,6 +300,10 @@ func getRandPool(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (uint64
 	poolCoins = simtypes.RemoveIndex(poolCoins, index)
 	coinOut := poolCoins[0]
 	poolDenoms := osmoutils.CoinsDenoms(pool.GetTotalPoolLiquidity(ctx))
-	gammDenom := types.GetPoolShareDenom(pool_id)
-	return pool_id, pool, coinIn, coinOut, poolDenoms, gammDenom, err
+
+	// pool ids start from 1.
+	poolId := poolIndex + 1
```
czarcas7ic, line 305: if we are max bound by the upcoming pool ID, its possible the poolIndex is the current largest poolID, so if we add one to here wouldn't it be a non existent pool?

app/upgrades/v13/upgrades.go
```diff
@@ -98,7 +98,6 @@ func migrateNextPoolId(ctx sdk.Context, gammKeeper *gammkeeper.Keeper, swaproute
 	// N.B: pool id in gamm is to be deprecated in the future
 	// Instead,it is moved to swaprouter.
 	nextPoolId := gammKeeper.GetNextPoolId(ctx)
-	gammKeeper.SetPoolCount(ctx, nextPoolId-1)
 	swaprouterKeeper.SetNextPoolId(ctx, nextPoolId)
```
czarcas7ic, line 101: Without going through the very fun e2e debugging process yet, my first reaction to the e2e error is the nextPoolID is not getting properly set

czarcas7ic (commented):
Leaving comments for now while we look into e2e failure

app/upgrades/v13/upgrades.go
p0mvn: @czarcas7ic hoping this: [0cf44a3](https://github.com/osmosis-labs/osmosis/pull/3492/commits/0cf44a3cef30d26bdc72da8bc1e3d5467c29ce20) fixed e2e. It was calling the removed query

x/gamm/keeper/genesis_test.go
p0mvn: Next pool id is tracked in the swap router now. All increment logic was removed from gamm. The reason we still keep the index is for migrations

x/gamm/simulation/sim_msgs.go
p0mvn: I think this is right

Assume we have 3 pools with ids: 1, 2, 3.

So, the max `poolIndex` we can generate is 2.

Then, max `poolId` is 3 which is the correct one

p0mvn:
If e2e passes, this is g2g since all other comments are addressed

p0mvn:
@czarcas7ic this is ready

x/gamm/simulation/sim_msgs.go
czarcas7ic: yeah I think you are correct, the verbiage in the comment "max bound by next poolID" was confusing which is why I had the initial impression

czarcas7ic (approved):

PR #3497, opened 2022-11-23
CL: add swap logic

czarcas7ic:
Closes: #XXX

Adds SwapExactAmountIn/Out are respective logic for bank sends to the concentrated liquidity module

This likely will need one or two more iterations, but wanted to get some reviews on this so we can merge to main and get rid of the concentrated-liquidity-main branch.


x/gamm/types/pool.go
```diff
@@ -141,6 +137,16 @@ type WeightedPoolExtension interface {
...
+type BalancerPool interface {
+	PoolI
+	TraditionalAmmInterface
+}
+
+type StableSwapExtension interface {
+	PoolI
+	TraditionalAmmInterface
+}
+
```
czarcas7ic, line 149: Note to reviewer, need to decide what we are adding to these extensions. Was not clear to me what is not shared since I did not work on stable swap

x/concentrated-liquidity/types/pool.go
```diff
@@ -3,10 +3,12 @@ package types
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	proto "github.com/gogo/protobuf/proto"
+
+	gammtypes "github.com/osmosis-labs/osmosis/v13/x/gamm/types"
 )
 
 type ConcentratedPoolExtension interface {
-	// gammtypes.PoolI
+	gammtypes.PoolI
```
p0mvn, line 11: We should remove the `PoolI` methods redefined below

x/swaprouter/router.go
```diff
@@ -130,7 +132,12 @@ func (k Keeper) MultihopEstimateOutGivenExactAmountIn(
 			swapFee = routeSwapFee.Mul((swapFee.Quo(sumOfSwapFees)))
 		}
 
-		tokenOut, err := pool.CalcOutAmtGivenIn(ctx, sdk.Coins{tokenIn}, route.TokenOutDenom, swapFee)
+		poolTrad, ok := pool.(gammtypes.TraditionalAmmInterface)
```
p0mvn, line 135: Is it possible to make this more general to work with CL pool as well?

We should aim to have everything in `swaprouter` to be generalized for any kind of pool

x/gamm/types/pool.go
p0mvn: Most logic is shared and we shouldn't need these.

Balancer pool supports more methods but those are defined on other / additional interface extensions such as `PoolAmountOutExtension`

x/gamm/keeper/swap.go
```diff
@@ -31,6 +31,11 @@ func (k Keeper) SwapExactAmountIn(
 	}
 	tokensIn := sdk.Coins{tokenIn}
 
+	pool, ok := poolI.(types.TraditionalAmmInterface)
+	if !ok {
+		return sdk.Int{}, errors.New("type cast failed")
```
p0mvn, line 36: consider formatting actual and expected types and also covering this with a test

x/gamm/keeper/pool.go
```diff
@@ -120,14 +125,20 @@ func (k Keeper) GetPoolsAndPoke(ctx sdk.Context) (res []types.TraditionalAmmInte
...
 		}
 
-		if pokePool, ok := pool.(types.WeightedPoolExtension); ok {
+		if pokePool, ok := poolI.(types.WeightedPoolExtension); ok {
 			pokePool.PokePool(ctx.BlockTime())
 		}
+
+		pool, ok := poolI.(types.TraditionalAmmInterface)
+		if !ok {
+			return nil, fmt.Errorf("type cast failed")
```
p0mvn, line 139: ditto for error. Maybe we could define this error in x/gamm/types/errors.go and reuse where applicable?

x/concentrated-liquidity/types/expected_keepers.go
```diff
@@ -0,0 +1,23 @@
...
+// creating a x/concentrated-liquidity keeper.
+type BankKeeper interface {
+	SendCoinsFromModuleToAccount(ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
+
+	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
+
+	SendCoins(ctx sdk.Context, fromAddr sdk.AccAddress, toAddr sdk.AccAddress, amt sdk.Coins) error
+
+	MintCoins(ctx sdk.Context, moduleName string, amt sdk.Coins) error
+	BurnCoins(ctx sdk.Context, name string, amt sdk.Coins) error
```
p0mvn, line 18: Do we need these?

x/gamm/keeper/swap.go
czarcas7ic: Im not quite sure how to go about formatting the error to show an "expected" vs "actual" type without just hard coding it.

x/gamm/keeper/pool.go
czarcas7ic: Same question as above

x/concentrated-liquidity/types/expected_keepers.go
czarcas7ic: Removed

x/swaprouter/router.go
czarcas7ic: Can we talk about this further? Do you for see just checking if the msg is CL and use the CL keeper method, otherwise use the TraditionalAmmInterface CalcOut?

x/gamm/keeper/swap.go
p0mvn: I'd do something like

```
 fmt.Errorf("failed cast to TraditionalAmmInterface, actual type: %T, poolI)
```

p0mvn: but not a blocker, feel free to skip if you want

x/swaprouter/router.go
p0mvn: I would instead propagate this logic to the keeper (`swapModule`). Similar to how we route swaps. The keeper methods can deal with that instead

x/gamm/keeper/swap.go
czarcas7ic: ah, duh, thanks 😅

x/concentrated-liquidity/lp.go
```diff
@@ -91,6 +106,26 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAd
...
+	// send withdraw amount from pool to position owner
+	if actualAmount0.IsPositive() && actualAmount1.IsPositive() {
+		err = k.bankKeeper.SendCoins(ctx, pool.GetAddress(), owner, sdk.NewCoins(sdk.NewCoin(denom0, actualAmount0), sdk.NewCoin(denom1, actualAmount1)))
+	} else if actualAmount0.IsPositive() && actualAmount1.IsZero() {
+		err = k.bankKeeper.SendCoins(ctx, pool.GetAddress(), owner, sdk.NewCoins(sdk.NewCoin(denom0, actualAmount0)))
+	} else if actualAmount0.IsZero() && actualAmount1.IsPositive() {
+		err = k.bankKeeper.SendCoins(ctx, pool.GetAddress(), owner, sdk.NewCoins(sdk.NewCoin(denom1, actualAmount0)))
+	}
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, err
```
czarcas7ic, line 126: Note to reviewer: does a normal send coins work here or do we need to eventually change this to a module account send

x/concentrated-liquidity/lp.go
```diff
@@ -56,6 +56,21 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, types.InsufficientLiquidityCreatedError{Actual: actualAmount1, Minimum: amount1Min}
 	}
 
+	denom0 := pool.GetToken0()
+	denom1 := pool.GetToken1()
+
+	// send deposit amount from position owner to pool
+	if actualAmount0.IsPositive() && actualAmount1.IsPositive() {
```
p0mvn, line 63: Is this a requirement to be implemented for testing swaps? If not, can we submit this in another PR to keep this one focused on swaps only and to avoid blocking it on this?

Some notes:
- I'm not sure why we have so many if /else. Should be possible to:

```
if amount 0 is not positive => error
if amount 1 is not positive => error

bank send denom0 with amount 0
bank send denom 1 with amount 1
```
- this logic should be tested, might also make sense to encapsulate it in a helper

x/concentrated-liquidity/lp.go
p0mvn: `SendCoinsFromModuleToAccount`

x/concentrated-liquidity/internal/emit.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+)
+
+// TODO: change as needed for CL rather than gamm
+func EmitSwapEvent(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, input sdk.Coin, output sdk.Coin) {
```
p0mvn, line 12: Can we instead move the swap event logic to swap router, make it public and import it to gamm and here?

x/concentrated-liquidity/internal/emit.go
```diff
@@ -0,0 +1,71 @@
...
+
+func newRemoveLiquidityEvent(sender sdk.AccAddress, poolId uint64, liquidity sdk.Coins) sdk.Event {
+	return sdk.NewEvent(
+		types.TypeEvtPoolExited,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		sdk.NewAttribute(sdk.AttributeKeySender, sender.String()),
+		sdk.NewAttribute(types.AttributeKeyPoolId, strconv.FormatUint(poolId, 10)),
+		sdk.NewAttribute(types.AttributeKeyTokensOut, liquidity.String()),
+	)
+}
```
p0mvn, line 71: I don't think we need these yet. Only swaps should be needed for now to make the tests pass. We should define concentrated-liqudity specific events separately

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -857,7 +857,10 @@ func (suite *KeeperTestSuite) TestJoinPool() {
 		tc := tc
 
 		suite.T().Run(tc.name, func(t *testing.T) {
-			pool := createTestPool(t, tc.swapFee, sdk.ZeroDec(), tc.poolAssets...)
+			poolI := createTestPool(t, tc.swapFee, sdk.ZeroDec(), tc.poolAssets...)
```
p0mvn, line 860: let's instead make the `createTestPool` return balancer right away instead of type casting everywhere

x/concentrated-liquidity/types/expected_keepers.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// BankKeeper defines the banking contract that must be fulfilled when
+// creating a x/concentrated-liquidity keeper.
+type BankKeeper interface {
+	SendCoinsFromModuleToAccount(ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
+
+	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
+
+	SendCoins(ctx sdk.Context, fromAddr sdk.AccAddress, toAddr sdk.AccAddress, amt sdk.Coins) error
```
p0mvn, line 15: Do we need all 3?

x/concentrated-liquidity/lp.go
czarcas7ic: I actually dont think we use FromModuleToAccount here. I believe in gamm we use SendCoins. Plus SendCoinsFromModuleToAccount only takes a module name, which we wouldn't be able to do for each pool

x/concentrated-liquidity/internal/model/pool.go
```diff
@@ -182,3 +182,25 @@ func (p *Pool) ApplySwap(newLiquidity sdk.Dec, newCurrentTick sdk.Int, newCurren
...
+
+// TODO: Figure out what we are going to do with these stubs since we call these in the keeper but are needed to use PoolI
+func (p Pool) CalcInAmtGivenOut(
+	ctx sdk.Context, tokensOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.Coin, err error) {
+	return sdk.Coin{}, nil
+}
+
+func (p Pool) CalcOutAmtGivenIn(ctx sdk.Context, tokensIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (sdk.Coin, error) {
+	return sdk.Coin{}, nil
+}
```
p0mvn, line 194: Why? We shouldn't need these as they should be defined on the `TraditionalAmmInterface`

x/concentrated-liquidity/internal/model/pool.go
```diff
@@ -182,3 +182,25 @@ func (p *Pool) ApplySwap(newLiquidity sdk.Dec, newCurrentTick sdk.Int, newCurren
...
+	return sdk.Coins{}
+}
+
+func (p *Pool) SwapInAmtGivenOut(ctx sdk.Context, tokensOut sdk.Coins, tokenInDenom string, swapFee sdk.Dec) (tokenIn sdk.Coin, err error) {
+	return sdk.Coin{}, nil
+}
+
+func (p *Pool) SwapOutAmtGivenIn(ctx sdk.Context, tokensIn sdk.Coins, tokenOutDenom string, swapFee sdk.Dec) (tokenOut sdk.Coin, err error) {
+	return sdk.Coin{}, nil
+}
```
p0mvn, line 206: These too

x/concentrated-liquidity/swaps.go
```diff
@@ -47,13 +52,53 @@ func (k Keeper) CreateNewConcentratedLiquidityPool(ctx sdk.Context, poolId uint6
...
 	tokenOutMinAmount sdk.Int,
 	swapFee sdk.Dec,
-) (sdk.Int, error) {
-	return sdk.Int{}, nil
+) (tokenOutAmount sdk.Int, err error) {
+	if tokenIn.Denom == tokenOutDenom {
+		return sdk.Int{}, errors.New("cannot trade same denomination in and out")
+	}
+
+	// get concentratedPool extension from poolI
```
p0mvn, line 65: Let's do a type cast instead of accessing the store

x/concentrated-liquidity/swaps.go
```diff
@@ -47,13 +52,53 @@ func (k Keeper) CreateNewConcentratedLiquidityPool(ctx sdk.Context, poolId uint6
...
+	asset0 := pool.GetToken0()
+	zeroForOne := tokenIn.Denom == asset0
+	var tokenOutCoin sdk.Coin
+
+	// change priceLimit based on which direction we are swapping
+	if zeroForOne {
+		tokenOutCoin, err = k.SwapOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, sdk.NewDec(1), pool.GetId())
+	} else {
+		tokenOutCoin, err = k.SwapOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, sdk.NewDec(999999999999), pool.GetId())
+	}
```
p0mvn, line 81: Let's define `sdk.NewDec(999999999999)` as `upperPriceLimit` and `sdk.OneDec()` as `lowerPriceLimit` 

```suggestion
        var priceLimit sdk.Dec
	// change priceLimit based on which direction we are swapping
	if zeroForOne {
		priceLimit = lowerPriceLimit
	} else {
		priceLimit = upperPriceLimit
	}
	tokenOutCoin, err = k.SwapOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, pool.GetId())
```

x/concentrated-liquidity/swaps.go
```diff
@@ -47,13 +52,53 @@ func (k Keeper) CreateNewConcentratedLiquidityPool(ctx sdk.Context, poolId uint6
...
+	if err != nil {
+		return sdk.Int{}, err
+	}
+
+	// check that the tokenOut calculated is both valid and less than specified limit
+	tokenOutAmount = tokenOutCoin.Amount
+	if !tokenOutAmount.IsPositive() {
+		return sdk.Int{}, sdkerrors.Wrapf(gammtypes.ErrInvalidMathApprox, "token amount must be positive")
```
p0mvn, line 89: Let's avoid using sdkerrors please

x/concentrated-liquidity/swaps.go
```diff
@@ -84,10 +129,31 @@ func (k Keeper) SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coin, tokenOutDen
...
+	swapFee sdk.Dec,
+) (tokenIn sdk.Coin, err error) {
+	return sdk.Coin{}, nil
+}
+
+// CalcOutAmtGivenInOld calculates tokens to be swapped out given the provided amount and fee deducted. It also returns
 // what the updated tick, liquidity, and currentSqrtPrice for the pool would be after this swap.
 // Note this method is non-mutative, so the values returned by CalcOutAmtGivenIn do not get stored
-func (k Keeper) CalcOutAmtGivenIn(ctx sdk.Context,
+func (k Keeper) CalcOutAmtGivenInOld(ctx sdk.Context,
```
p0mvn, line 156: Let's call this `calcOutAmtGivenIn`.

x/concentrated-liquidity/swaps.go
```diff
@@ -242,7 +308,7 @@ func (k *Keeper) SwapInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coin, tokenInDe
 	return tokenInCoin, nil
 }
 
-func (k Keeper) CalcInAmtGivenOut(ctx sdk.Context, tokenOut sdk.Coin, tokenInDenom string, swapFee sdk.Dec, minPrice, maxPrice sdk.Dec, poolId uint64) (sdk.Coin, sdk.Dec, sdk.Int, sdk.Dec, error) {
+func (k Keeper) CalcInAmtGivenOutOld(ctx sdk.Context, tokenOut sdk.Coin, tokenInDenom string, swapFee sdk.Dec, minPrice, maxPrice sdk.Dec, poolId uint64) (sdk.Coin, sdk.Dec, sdk.Int, sdk.Dec, error) {
```
p0mvn, line 311: ditto for lowercase

x/gamm/keeper/swap.go
```diff
@@ -123,6 +123,36 @@ func (k Keeper) SwapExactAmountOut(
 	return tokenInAmount, nil
 }
 
+func (k Keeper) CalcOutAmtGivenIn(
```
p0mvn, line 126: Let's add basic success and failure test cases. To cover given pool not implementing `TraditionalAmmInterface`

x/swaprouter/router.go
```diff
@@ -377,13 +379,18 @@ func (k Keeper) createOsmoMultihopExpectedSwapOuts(
...
 		if err != nil {
 			return nil, err
 		}
 
-		swapFee := pool.GetSwapFee(ctx)
-		tokenIn, err := pool.CalcInAmtGivenOut(ctx, sdk.NewCoins(tokenOut), route.TokenInDenom, cumulativeRouteSwapFee.Mul((swapFee.Quo(sumOfSwapFees))))
+		poolTrad, ok := poolI.(gammtypes.TraditionalAmmInterface)
+		if !ok {
+			return nil, fmt.Errorf("failed cast to TraditionalAmmInterface, actual type: %T", poolI)
+		}
```
p0mvn, line 390: Should this cast be removed?

x/concentrated-liquidity/internal/model/pool.go
czarcas7ic: They are defined in PoolI within swaprouter here, not TraditionalAmmInterface

https://github.com/osmosis-labs/osmosis/blob/dba9df0db5a8330de4204e52ec0369cea31c570c/x/swaprouter/types/pool.go#L30-L42

p0mvn: Ahh I see, these should be defined on the `TraditionalAmmInterface`. I'll try to see if moving these is a quick change. If so, maybe we can merge that change first to avoid creating the stubs here unnecessarily

czarcas7ic: Didn't we have these defined on TAMM previously and then decided to move them up to PoolI?

p0mvn: No, based on the component diagram, the goal has been to have it on `TraditionalAmmInterface`. The only reason why we would keep these methods on `PoolI` until now it is to be able to port `EstimateSwap` to swaprouter. However, since we now call keeper's "Calc" method instead of the pool's, we can move these to the right place.

I've looked into this problem, and moving the methods to `TraditionalAmmInterface` depends on this PR. Specificcally, on the change I mentioned above (where swaprouter uses keeper's calc as opposed to pool's).

So I think we can merge this first and refactor separately.

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	tests := []struct {
+		name      string
+		param     param
+		expectErr bool
+	}{
+		{
+			name: "Proper swap foo > bar",
+			param: param{
+				tokenIn:           sdk.NewCoin("foo", sdk.NewInt(42000000)),
```
p0mvn, line 44: Is there a Wolfram link you are using to QA this? How can I be sure that the values are correct here?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+				tokenIn:           sdk.NewCoin("bar", sdk.NewInt(13370)),
+				tokenOutDenom:     "foo",
+				tokenOutMinAmount: sdk.NewInt(1),
+				expectedTokenOut:  sdk.NewInt(66808387),
+			},
+		},
+		{
+			name: "out is lesser than min amount",
+			param: param{
+				tokenIn:           sdk.NewCoin("foo", sdk.NewInt(42000000)),
```
p0mvn, line 62: nit: seems that a  lot of the same values are reused throughout the test - `42000000`, `8397` etc. Consider initializing constants with these values and reusing them where applicable

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+			s.Require().NoError(err)
+
+			// Check the test case to see if we are swapping asset0 for asset1 or vice versa
+			asset0 := pool.GetToken0()
+			zeroForOne := test.param.tokenIn.Denom == asset0
+
+			// Fund the test account with foo and bar, then create a default position to the pool created earlier
+			s.FundAcc(s.TestAccs[0], sdk.NewCoins(sdk.NewCoin("bar", sdk.NewInt(10000000000000)), sdk.NewCoin("foo", sdk.NewInt(1000000000000))))
+			_, _, _, err = s.App.ConcentratedLiquidityKeeper.CreatePosition(s.Ctx, 1, s.TestAccs[0], defaultAmt0, defaultAmt1, sdk.ZeroInt(), sdk.ZeroInt(), lowerTick.Int64(), upperTick.Int64())
+			s.Require().NoError(err)
```
p0mvn, line 113: If this is a default positon and assuming that this is going to be reused frequently, should we make a test helper for this?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+			if test.expectErr {
+				pool, err = s.App.ConcentratedLiquidityKeeper.GetPoolById(s.Ctx, pool.GetId())
```
p0mvn, line 116: Why do we get pool here again? Seems that we have one initialized above and the "else" case does not re-get it

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+				prevGasConsumed := s.Ctx.GasMeter().GasConsumed()
+
+				// Execute the swap directed in the test case
+				tokenOutAmount, err := s.App.ConcentratedLiquidityKeeper.SwapExactAmountIn(s.Ctx, s.TestAccs[0], pool.(swaproutertypes.PoolI), test.param.tokenIn, test.param.tokenOutDenom, test.param.tokenOutMinAmount, swapFee)
+				s.Require().NoError(err)
+				s.Require().Equal(test.param.expectedTokenOut.String(), tokenOutAmount.String())
+
+				gasConsumedForSwap := s.Ctx.GasMeter().GasConsumed() - prevGasConsumed
+				// TODO: make a CLGasFeeForSwap
+				// Check that we consume enough gas that a CL pool swap warrants
```
p0mvn, line 130: nit: seems like a quick change. Do you mind adding it right away instead of having to track this?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+				spotPriceAfter := pool.GetCurrentSqrtPrice().Power(2)
+
+				// Ratio of the token out should be between the before spot price and after spot price.
+				var tradeAvgPrice sdk.Dec
+				if zeroForOne {
+					tradeAvgPrice = tokenOutAmount.ToDec().Quo(test.param.tokenIn.Amount.ToDec())
+				} else {
+					tradeAvgPrice = test.param.tokenIn.Amount.ToDec().Quo(tokenOutAmount.ToDec())
+				}
```
p0mvn, line 149:
```suggestion
				tradeAvgPrice := tradeAvgPrice = tokenOutAmount.ToDec().Quo(test.param.tokenIn.Amount.ToDec())
				if !zeroForOne {
					tradeAvgPrice = sdk.OneDec().Quo(tradeAvgPrice)
				}
```

p0mvn (approved):
LGTM once remaining comments are addressed.

x/concentrated-liquidity/swaps_test.go
czarcas7ic: "trust me bro"

jk, I am using the same base test cases from pool_test, I can copy over the wolfram links from there though

x/concentrated-liquidity/swaps_test.go
czarcas7ic: I kind of feel displaying the number itself, despite being repeated, is the best way. If it was instead something like "baseCaseEthIn" and "baseCaseUSDCOut", we would just need to look at what that is anyway to confirm the math is right, right? Please lmk if I am misunderstanding, I very well could be

p0mvn: Fair, that makes sense! Let's keep it

x/concentrated-liquidity/swaps_test.go
p0mvn: Yeah I think duplicating a link would be helpful. While I assumed that there is probably another test case using the same values, it is just much more easier having it available in this test directly

x/gamm/keeper/swap.go
czarcas7ic: This is the last point I have not addressed, are you saying instead of returning a blank coin and error, you want me to check if we can type cast the provided poolI to TAMM, and then make a test in swaps_test to test for this?

p0mvn: I just want to make sure that we have test coverage for the following:
- balancer pool given -> swap is done (minimal assertions to ensure correctness)
- stableswap pool given -> swap is done (minimal assertions to ensure correctness)
- cl pool given -> swap not done and error returned

czarcas7ic: But this comment is point to Calc, not swap. Was this meant to be for Swap?

p0mvn: On a related note, in another PR, I would like to suggest making a helper:

```go
func convertToCFMMPooll() (types.TraditionalAmmInterface, error) {
    	cfmmPool, ok := poolI.(types.TraditionalAmmInterface)
	if !ok {
		return nil, fmt.Errorf("given pool does not implement TraditionalAmmInterface, implements %T", poolI)
	}
       return cfmmPool, nil
}
```

and a similar helper for the concentrated pool conversion

p0mvn: @czarcas7ic sorry just saw your reply. I meant calc, not swap. Edited my previous comment

p0mvn (approved):

PR #3505, opened 2022-11-24
feat(CL): x/concentrated-liquidity protos (merge to main 5)

p0mvn:
Closes: #XXX

We would like to drop `concentrated-liquidity-main` branch to avoid branch synching overhead. As part of that, we are going to be incrementally merging the current concentrated liquidity feature state to `main`.

All this logic has already been given a round of reviews since we treated `concentrated-liqudiity-main` branch as `main` with appropriate review processes.

This particular PR does the following:
- merges `concentrated-liquidity` module protos

For the latest state of the feature that we are merging, see: https://github.com/osmosis-labs/osmosis/tree/concentrated-liquidity-main

See spec: https://github.com/osmosis-labs/osmosis/pull/3052/files


p0mvn:
Not sure why linter is failing. Maybe it makes more sense to merge the entire `concentrated-liquidity` module all at once

czarcas7ic:
https://github.com/osmosis-labs/osmosis/pull/3506

This is the entire cl module merged instead of just the protos (and by definition the last part of concentrated-liquidity-main that needs to get merged)

If you agree with this approach we can close this PR and merge 3506. Normally I would be against a PR this large, but since we reviewed every PR on their own, I don't think this is problematic.

PR #3145, opened 2022-10-25
add tests for GetIfFound

hieuvubk:
Closes: #3138 

## What is the purpose of the change

Add testcases for `GetIfFound` functions

## Testing and Verifying

All test added & passed

PR #3136, opened 2022-10-25
Add state for Mint: tick and Position

mattverse:
This PR mainly serves two purposes:
- Adding State for tick 
- Adding State for position 
in the process of Minting. 

1. Adding State for Tick
We have two state entries for TickInfo now, `liquidityGross` and `liquidityNet`. Gross checks the inclined amount regardless of liquidityDelta being positive or negative, while liquidityNet tracks liquidity with regarding the sign of liquidityDelta. 

2. Adding state for Position 
We add state for position given address, upper tick, lower tick and liquidity In whenever a mint is happening. 

Added tests to verify that state entry is correctly being stored.

PR #3512, opened 2022-11-24
feat(superfluid/gov): configure unpool whitelist via gov prop

p0mvn:
Closes: #XXX

Add the ability to configure the unpool whitelist via gov prop.

The new gov prop takes the list of pool ids and `isOverwriteFlag`. If the flag is true, the old pool ids are discarded. If false, the old pool ids are merged with the new.

## Brief Changelog
- wired up gov prop boilerplate
- implemented validation in the handler
- unit tested
- Took store helper and tests from CL branch:
   * https://github.com/osmosis-labs/osmosis/pull/3136
   * https://github.com/osmosis-labs/osmosis/pull/3145

## Testing and Verifying

- [x] uni tests added
- [x] localnet test

```
# Create 3 pools by running the below 3 times
osmosisd tx gamm create-pool --pool-file=tests/e2e/scripts/nativeDenomPool.json  --from lo-test1 --keyring-backend test -b=block

# Fails because pool id 4 does not exist
osmosisd tx gov submit-proposal update-unpool-whitelist --pool-ids "4" --title "test" --description "test"  --from lo-test1 --keyring-backend test -b=block

# Succesfully creates proposal
osmosisd tx gov submit-proposal update-unpool-whitelist --pool-ids "1, 2, 3" --title "test" --description "test"  --from lo-test1 --keyring-backend test -b=block
```

Upon enabling whitelist, it is still impossible to run:
```
# Fails with message not activated.
 osmosisd tx superfluid unpool-whitelisted-pool 1 --from lo-test1 --keyring-backend test -b=bloc
```
We should enable that message in a separate PR.

  - How is the feature or change documented? godoc


x/superfluid/keeper/gov/suite_test.go
```diff
@@ -2,30 +2,26 @@ package gov_test
...
+	"github.com/osmosis-labs/osmosis/v13/x/superfluid/types"
 )
 
 type KeeperTestSuite struct {
 	apptesting.KeeperTestHelper
 
-	ctx     sdk.Context
-	querier keeper.Querier
-	app     *app.OsmosisApp
+	// ctx     sdk.Context
```
p0mvn, line 16:
```suggestion
```

x/superfluid/keeper/gov/suite_test.go
```diff
@@ -2,30 +2,26 @@ package gov_test
...
+	querier types.QueryServer
+	// queryClient types.QueryClient
+	// app     *app.OsmosisApp
```
p0mvn, line 19:
```suggestion
```

osmoutils/store_helper.go
```diff
@@ -137,3 +137,16 @@ func MustGetDec(store store.KVStore, key []byte) sdk.Dec {
 	MustGet(store, key, result)
 	return result.Dec
 }
+
+// GetIfFound gets key from store
+// returns a boolean indicating whether value exists for the given key and error
+func GetIfFound(store store.KVStore, key []byte, result proto.Message) (found bool, err error) {
```
ValarDragon, line 143: Can be changed later / can discuss in an issue, should not block this PR. Perhaps this should be called `HasWithType` ?

e.g. if we were masters of the store interface
```
store.HasProtoMsg("myKey", types.Pool)
```

Has to indicate its a bool (rather than assuming you'd get the object back)

OR is proto.Message here actually mutated? If so lets update the comment, but this doesn't actually make sense to me as an API.

Why don't just make a method called `Get` thats the same as MustGet, but returns an error instead of panics?

x/superfluid/client/rest/tx.go
```diff
@@ -10,23 +10,25 @@ import (
 func ProposalSetSuperfluidAssetsRESTHandler(clientCtx client.Context) govrest.ProposalRESTHandler {
 	return govrest.ProposalRESTHandler{
 		SubRoute: "set-superfluid-assets",
-		Handler:  newSetSuperfluidAssetsHandler(clientCtx),
```
ValarDragon, line 13: idk whats going on here, but IIRC this rest folder doesn't matter for anything so not blocking

p0mvn: For each of these, the handler is the same - no-op. I just made a shared no-op function for all of these but yes it's unimportant

ValarDragon: ah that makes sense!

x/superfluid/types/gov.go
```diff
@@ -108,3 +112,46 @@ func (p RemoveSuperfluidAssetsProposal) String() string {
...
+		if id == 0 {
+			return fmt.Errorf("pool id cannot be 0")
+		}
+	}
+
+	return nil
+}
+
+func (p UpdateUnpoolWhiteListProposal) String() string {
+	return fmt.Sprintf(`Set Superfluid Assets Proposal:
```
ValarDragon, line 151:
```suggestion
	return fmt.Sprintf(`Update Unpool Whitelist Assets Proposal:
```

ValarDragon (approved):
Looks great to me! Can we just sanity check that we have an existing test for ensuring that unpooling doesn't work for pools not in whitelist?

x/superfluid/keeper/gov/gov_test.go
```diff
@@ -135,22 +134,200 @@ func (suite *KeeperTestSuite) TestHandleSetSuperfluidAssetsProposal() {
...
 			}
 		})
 	}
 }
+
+func (suite *KeeperTestSuite) TestHandleUnpoolWhiteListChange() {
+	const (
+		testTitle       = "test title"
+		testDescription = "test description"
+	)
```
ValarDragon, line 161: Great job on the test!

x/superfluid/client/cli/tx.go
```diff
@@ -402,3 +405,92 @@ func NewCmdUnPoolWhitelistedPool() *cobra.Command {
...
 }
+
+// NewCmdUpdateUnpoolWhitelistProposal defines the command to create a new update unpool whitelist proposal command.
+func NewCmdUpdateUnpoolWhitelistProposal() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-unpool-whitelist [flags]",
+		Args:  cobra.ExactArgs(0),
+		Short: "Update unpool whitelist proposal",
+		Long: "This proposal will update the unpool whitelist if passed. " +
+			"Every pool id must be valid. If proposal passes, and the pool id is invalid, the proposal will not take any effect. " +
```
czarcas7ic, line 416: Why don't we run GetPoolAndPoke in ValidateBasic to prevent this from happening?

czarcas7ic (approved):
Well done 👍

x/superfluid/client/cli/tx.go
p0mvn: Unfortunately, we don't have access to the gamm keeper in `ValidateBasic()`.

However, I found out that gov module attempts to test-run the proposal with cache context before allowing it to be submitted:
https://github.com/osmosis-labs/cosmos-sdk/blob/870132d311d162fdbe0e92fcc0a42ee3889afd0d/x/gov/keeper/proposal.go#L22-L29

So if pool doesn't exist, it will not be submitted. I will change the wording

p0mvn:
> Looks great to me! Can we just sanity check that we have an existing test for ensuring that unpooling doesn't work for pools not in whitelist?

Test added: https://github.com/osmosis-labs/osmosis/pull/3532

Issue #3517, opened 2022-11-25
State sync is broken in v12.2.1+


faddat:
## Context

I've never been able to state sync osmosis.  I've contributed state sync scripts, some with pebbledb, some without, to > a dozen other cosmos chains.  

There's been progress lately, namely that errors are reduced to those described here:

* https://github.com/osmosis-labs/osmosis/pull/3507
* https://github.com/osmosis-labs/osmosis/issues/1090

My initial guess was that it was related to the use of upstream iavl, but that's proven incorrect.  I've purchased a server, that will allow me to investigate this more thoroughly.

faddat:
So I just spent ... pretty much way too long making state sync work.

this was going to be a writeup of problems, but I **just** got it working. 

I don't know why it frequently doesn't work,  **but** my node just reached the tip.  

In order to make it work, I set a server to make a snapshot every 100 blocks.  This actually takes longer than every 100 blocks takes to pass by, but basically I:

used the Notional snapshot, which is pruned, but not heavily, and is pebble flavored:

https://snapshot.notional.ventures/

I applied #3522 to Osmosis v12.3.0, and compiled on that server with the pebble code (mainly for speed reasons -- this took over 8 hours to get rolling) and then from my desktop, I began a state sync where that server was my only peer. 

Previously, when I tried to state sync, I got block.Header.AppHash errors.  

Now, it works.  I'm pushing my working state sync branch here, and it is possible that we can finally make it work in CI.

PR #3522, opened 2022-11-25
remove settings that configure state sync

faddat:
closes: #3517 

## Use default state sync configuration

This PR removes settings that should no longer need to be set, and maybe never should have been set.  Specifically, this PR relies more heavily on upstream defaults:

* using the default state sync trust period, instead of 112 hours
* taking a snapshot every 1000 blocks, instead of 1500 (iirc)
* disabling serving state sync snapshots by default

## Testing and Verifying

* Manually verified the change by running a node with these settings, then state syncing to my desktop against it.

Script to reproduce the findings is here:

Working:
https://github.com/osmosis-labs/osmosis/blob/faddat/working-state-sync-example/scripts/statesync-resolution.bash

Not Working:
https://github.com/osmosis-labs/osmosis/blob/faddat/working-state-sync-example/scripts/statesync.bash

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no) yes: 


faddat:
closing this since it didn't target the correct branch.

PR #3525, opened 2022-11-25
refactor(CL): deprecate gamm messages and queries instead of removing

p0mvn:
Closes: #XXX

We should avoid removing old messages and queries from x/gamm to avoid breaking smart contracts, frontends, and other clients.

A decision was made to deprecate the old messages and queries instead of removing them. That would imply keeping the client-facing APIs but routing the implementation into the new `x/swaprouter` module.

This PR brings back the client-facing APIs. The routing itself will be handled in the PR after.

## Important Changes

- removed `SwapAmountInRoute` and `SwapAmountOutRoute` proto definitions in gamm. Instead, imported duplicates from swaprouter


proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -231,6 +259,50 @@ message QuerySpotPriceResponse {
...
+// Deprecated: use QuerySwapExactAmountInRequest in x/swaprouter.
+message QuerySwapExactAmountInRequest {
+  // TODO: CHANGE THIS TO RESERVED IN A PATCH RELEASE
+  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
+  uint64 pool_id = 2 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string token_in = 3 [ (gogoproto.moretags) = "yaml:\"token_in\"" ];
+  repeated osmosis.swaprouter.v1beta1.SwapAmountInRoute routes = 4 [
+    (gogoproto.moretags) = "yaml:\"routes\"",
+    (gogoproto.nullable) = false
+  ];
```
czarcas7ic, line 272: Should we set all these fields to [deprecated = true]?

x/gamm/keeper/grpc_query.go
```diff
@@ -415,3 +420,13 @@ func (q Querier) TotalLiquidity(ctx context.Context, _ *types.QueryTotalLiquidit
...
+
+// Deprecated: use EstimateSwapExactAmountIn in x/swaprouter.
+func (q Querier) EstimateSwapExactAmountIn(ctx context.Context, req *types.QuerySwapExactAmountInRequest) (*types.QuerySwapExactAmountInResponse, error) {
+	panic("not implemented")
+}
+
+// Deprecated: use EstimateSwapExactAmountOut in x/swaprouter.
+func (q Querier) EstimateSwapExactAmountOut(ctx context.Context, req *types.QuerySwapExactAmountOutRequest) (*types.QuerySwapExactAmountOutResponse, error) {
+	panic("not implemented")
+}
```
czarcas7ic, line 432: Can you explain this a little further in regards to future steps? Do these panic because they are not client facing?

x/gamm/keeper/msg_server.go
```diff
@@ -25,6 +27,22 @@ func NewStableswapMsgScalingFactorModifierServerImpl(keeper *Keeper) stableswap.
...
+// Deprecated: use CreateBalancerPool in x/swaprouter.
+func (server msgServer) CreateBalancerPool(goCtx context.Context, msg *balancer.MsgCreateBalancerPool) (*balancer.MsgCreateBalancerPoolResponse, error) {
+	panic("not implemented")
+}
+
+// Deprecated: use CreateStableswapPool in x/swaprouter.
+func (server msgServer) CreateStableswapPool(goCtx context.Context, msg *stableswap.MsgCreateStableswapPool) (*stableswap.MsgCreateStableswapPoolResponse, error) {
+	panic("not implemented")
+}
+
```
czarcas7ic, line 45: Same question on these

x/gamm/types/codec.go
```diff
@@ -14,6 +14,8 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 	cdc.RegisterInterface((*TraditionalAmmInterface)(nil), nil)
 	cdc.RegisterConcrete(&MsgJoinPool{}, "osmosis/gamm/join-pool", nil)
 	cdc.RegisterConcrete(&MsgExitPool{}, "osmosis/gamm/exit-pool", nil)
+	cdc.RegisterConcrete(&MsgSwapExactAmountIn{}, "osmosis/gamm/swap-exact-amount-in", nil)
+	cdc.RegisterConcrete(&MsgSwapExactAmountOut{}, "osmosis/gamm/swap-exact-amount-out", nil)
```
czarcas7ic, line 18: Why are only these getting registered and not the others? (num pools, create pool, etc)

x/gamm/types/msgs.go
```diff
@@ -60,6 +64,82 @@ func ValidateFutureGovernor(governor string) error {
...
+	}
+
+	err = swaproutertypes.SwapAmountInRoutes(msg.Routes).Validate()
+	if err != nil {
+		return err
+	}
+
+	if !msg.TokenIn.IsValid() || !msg.TokenIn.IsPositive() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, msg.TokenIn.String())
+	}
```
czarcas7ic, line 84: Do we continue to use sdkerrors here? Just remember you saying something about avoiding their use.

x/gamm/types/msgs_test.go
```diff
@@ -7,13 +7,273 @@ import (
...
+			Routes: []swaproutertypes.SwapAmountInRoute{{
+				PoolId:        0,
+				TokenOutDenom: "test",
+			}, {
+				PoolId:        1,
+				TokenOutDenom: "test2",
+			}},
+			TokenIn:           sdk.NewCoin("test", sdk.NewInt(100)),
+			TokenOutMinAmount: sdk.NewInt(200),
+		}
```
czarcas7ic, line 35: I'm a little confused on this as well, I thought we returned a panic when calling a swapExactAmountIn msg from gamm side?

proto/osmosis/swaprouter/v1beta1/query.proto
```diff
@@ -4,6 +4,7 @@ package osmosis.swaprouter.v1beta1;
 import "gogoproto/gogo.proto";
 import "osmosis/swaprouter/v1beta1/genesis.proto";
 import "osmosis/swaprouter/v1beta1/tx.proto";
+import "osmosis/swaprouter/v1beta1/swap_route.proto";
```
czarcas7ic, line 7: Why does swap route proto get added here to query?

czarcas7ic (commented):
Waiting to approve until I can further grasp the next steps. I don't doubt this is following a logical path, but its just not entirely clear to me yet. Once some of these questions are addressed I am sure it will be more clear to me

proto/osmosis/gamm/v1beta1/query.proto
p0mvn: Yeah, I will be trying that in the future PR. Here, I would like to bring back the previously removed messages and queries as is.

The plan is to implement the entrypoints to route to swaprouter and try marking them as proto-deprecated in the next PR.

x/gamm/keeper/grpc_query.go
p0mvn: Here I'm only bringing back the client-facing APIs. The routing to swap router will be handled in the next PR

x/gamm/types/codec.go
p0mvn: create pool is handled per-model:
https://github.com/osmosis-labs/osmosis/blob/f0bb20932d8c3468926f4b3a4935d6ad245d3f37/x/gamm/pool-models/balancer/codec.go#L17

That was unchanged. Queries do not need this registration, only messages

x/gamm/types/msgs.go
p0mvn: I'm not adding new code, only bringing back what was removed. For ease of merging to main, we should keep things as they used to be

x/gamm/types/msgs_test.go
p0mvn: These tests are being re-added. See main: https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/types/msgs_test.go

I'm not adding anything new, only bringing back previously removed messages and tests

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -231,6 +259,50 @@ message QuerySpotPriceResponse {
...
+}
+
+//=============================== EstimateSwapExactAmountOut
+// Deprecated: use QuerySwapExactAmountOutRequest in x/swaprouter.
+message QuerySwapExactAmountOutRequest {
+  // TODO: CHANGE THIS TO RESERVED IN A PATCH RELEASE
+  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
+  uint64 pool_id = 2 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  repeated osmosis.swaprouter.v1beta1.SwapAmountOutRoute routes = 3 [
```
p0mvn, line 290: Here we use route

proto/osmosis/swaprouter/v1beta1/query.proto
p0mvn: https://github.com/osmosis-labs/osmosis/pull/3525/files#r1033004282

p0mvn:
@czarcas7ic all comments are addressed. Please take a look

czarcas7ic (approved):
Thanks for the explanations. Approving with the understanding that the panics will get replaced with wiring to the swap router in a subsequent PR

PR #3526, opened 2022-11-25
feat(CL): add bank sends to positions

czarcas7ic:
Closes: #XXX

As part of an offline conversation, we agreed to remove the BankSend logic from [this PR](https://github.com/osmosis-labs/osmosis/pull/3497) and create a separate PR for it. This PR introduces the BankSend logic for creating and leaving a position.

### DIFF
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -56,6 +56,12 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, types.InsufficientLiquidityCreatedError{Actual: actualAmount1, Minimum: amount1Min}
 	}
 
+	// send deposit amount from position owner to pool
+	err = k.sendCoinsBetweenPoolAndUser(cacheCtx, pool.GetToken0(), pool.GetToken1(), actualAmount0, actualAmount1, owner, pool.GetAddress())
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+	}
+
 	// only persist updates if amount validation passed.
 	writeCacheCtx()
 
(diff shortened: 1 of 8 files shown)


x/concentrated-liquidity/lp.go
```diff
@@ -147,3 +164,19 @@ func (k Keeper) updatePosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 	// The returned amounts are rounded down to avoid returning more to clients than they actually deposited.
 	return actualAmount0.TruncateInt(), actualAmount1.TruncateInt(), nil
 }
+
+func (k Keeper) sendCoinsBetweenPoolAndUser(ctx sdk.Context, denom0, denom1 string, amount0, amount1 sdk.Int, sender, receiver sdk.AccAddress) error {
```
p0mvn, line 168: godoc and tests please

p0mvn (commented):
Nice work! Looks great to me once godoc and tests added!

x/concentrated-liquidity/lp_test.go
```diff
@@ -276,3 +278,108 @@ func mergeConfigs(dst *lpTest, overwrite *lpTest) {
...
+		"only asset0 is greater than sender has, position creation (user to pool)": {
+			coin0:       sdk.NewCoin("eth", sdk.NewInt(100000000000000)),
+			coin1:       sdk.NewCoin("usdc", sdk.NewInt(1000000)),
+			expectError: true,
+		},
+		// "only asset1 is greater than sender has, position creation (user to pool)": {
+		// 	coin0:       sdk.NewCoin("eth", sdk.NewInt(1000000)),
+		// 	coin1:       sdk.NewCoin("usdc", sdk.NewInt(100000000000000)),
+		// 	expectError: true,
+		// },
```
czarcas7ic, line 311: for some reason, everything works as expected but this...

if the second asset is too large, we end up BankSending the coin0 but not coin1...

czarcas7ic: Actually, I suppose this makes sense. While this test case will be odd and show a partial send, it wouldn't actually get committed and would get reverted due to the error right? Yeah I guess for error cases I should just not check the resulting amounts

p0mvn (approved):
Great work! I'm good with merging this now (before merge to `main` is complete)

PR #3527, opened 2022-11-25
refactor(CL): move PoolI to swaprouter; swaprouter only depends on PoolI

p0mvn:
Closes: #XXX

`PoolI` is the most general pool interface. Every pool implements it.

`x/swaprouter` is the module that deals with the most general pool logic. As a result, `PoolI` belongs there.

Additionally, there are currently several import cycles that are blocking #3525 . This PR needs to be merged to unblock mergin to `main`.

Lastly, `swaprouter` previously depended on the `gammtypes.TraditionalAmmInterface` when it shouldn't have. Now, all dependencies have been removed so that `x/swaprouter` is only concerned with the generalized `PoolI`.

### DIFF
--- a/x/gamm/types/msgs_test.go
+++ b/x/gamm/types/msgs_test.go
@@ -675,14 +675,6 @@ func TestAuthzMsg(t *testing.T) {
 				InitialPoolLiquidity: sdk.NewCoins(coin),
 			},
 		},
-		{
-			name: "MsgCreateStableswapPool",
-			gammMsg: &stableswap.MsgCreateStableswapPool{
-				Sender:               addr1,
-				PoolParams:           &stableswap.PoolParams{},
-				InitialPoolLiquidity: sdk.NewCoins(coin),
-			},
-		},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/x/gamm/types/pool.go
+++ b/x/gamm/types/pool.go
@@ -3,60 +3,16 @@ package types
 import (
 	"time"
 
-	proto "github.com/gogo/protobuf/proto"
-
 	"github.com/cosmos/cosmos-sdk/types/address"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	swaproutertypes "github.com/osmosis-labs/osmosis/v13/x/swaprouter/types"
 )
-
-// PoolI defines an interface for pools that hold tokens.
-type PoolI interface {
-	proto.Message
-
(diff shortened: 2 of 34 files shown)


x/gamm/types/msgs_test.go
```diff
@@ -675,14 +675,6 @@ func TestAuthzMsg(t *testing.T) {
 				InitialPoolLiquidity: sdk.NewCoins(coin),
 			},
 		},
-		{
-			name: "MsgCreateStableswapPool",
```
p0mvn, line 679: Note: there is a duplicate right above

x/gamm/types/pool.go
```diff
@@ -3,60 +3,16 @@ package types
...
-}
+	swaproutertypes "github.com/osmosis-labs/osmosis/v13/x/swaprouter/types"
+)
 
 // TraditionalAmmInterface defines an interface for pools representing traditional AMM.
 type TraditionalAmmInterface interface {
-	PoolI
-
-	// GetTotalPoolLiquidity returns the coins in the pool owned by all LPs
-	GetTotalPoolLiquidity(ctx sdk.Context) sdk.Coins
```
p0mvn, line 59: Note: this was moved to `PoolI`. Although this getter might not be very useful for CL pools, it is still possible to have it. In the context of this PR, moving it simplifies the refactor here if the method is defined on `PoolI`. As a result, I would like to revisit this later

x/swaprouter/types/routes.go
```diff
@@ -31,14 +29,14 @@ type CommunityPoolI interface {
...
+	InitializePool(ctx sdk.Context, pool PoolI, creatorAddress sdk.AccAddress) error
 
-	GetPool(ctx sdk.Context, poolId uint64) (gammtypes.PoolI, error)
+	GetPool(ctx sdk.Context, poolId uint64) (PoolI, error)
 
 	SwapExactAmountIn(
 		ctx sdk.Context,
 		sender sdk.AccAddress,
-		poolId gammtypes.PoolI,
+		poolId PoolI,
```
czarcas7ic, line 39: Same request for poolId change

x/swaprouter/types/routes.go
```diff
@@ -48,7 +46,7 @@ type SwapI interface {
 	SwapExactAmountOut(
 		ctx sdk.Context,
 		sender sdk.AccAddress,
-		poolId gammtypes.PoolI,
+		poolId PoolI,
```
czarcas7ic, line 49: same

czarcas7ic (approved):
LGTM!

PR #3528, opened 2022-11-26
remove settings that configure state sync

faddat:
closes: #3517 

## Use default state sync configuration

This PR removes settings that should no longer need to be set, and maybe never should have been set.  Specifically, this PR relies more heavily on upstream defaults:

* using the default state sync trust period, instead of 112 hours
* taking a snapshot every 1000 blocks, instead of 1500 (iirc)
* disabling serving state sync snapshots by default

## Testing and Verifying

* Manually verified the change by running a node with these settings, then state syncing to my desktop against it.

Script to reproduce the findings is here:

Working:
https://github.com/osmosis-labs/osmosis/blob/faddat/working-state-sync-example/scripts/statesync-resolution.bash

Not Working:
https://github.com/osmosis-labs/osmosis/blob/faddat/working-state-sync-example/scripts/statesync.bash

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no) yes: 

### DIFF
--- a/cmd/osmosisd/cmd/init.go
+++ b/cmd/osmosisd/cmd/init.go
@@ -102,8 +102,6 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 			config.P2P.MaxNumOutboundPeers = 40
 			config.Consensus.TimeoutCommit = 2 * time.Second
 			config.Mempool.Size = 10000
-			config.StateSync.TrustPeriod = 112 * time.Hour
-			config.FastSync.Version = "v0"
 
 			config.SetRoot(clientCtx.HomeDir)
 
--- a/cmd/osmosisd/cmd/root.go
+++ b/cmd/osmosisd/cmd/root.go
@@ -102,8 +102,6 @@ func initAppConfig() (string, interface{}) {
 	// server config.
 	srvCfg := serverconfig.DefaultConfig()
 	srvCfg.API.Enable = true
-	srvCfg.StateSync.SnapshotInterval = 1500
-	srvCfg.StateSync.SnapshotKeepRecent = 2
 	srvCfg.MinGasPrices = "0uosmo"
 
 	// 128MB IAVL cache


cmd/osmosisd/cmd/root.go
```diff
@@ -102,8 +102,6 @@ func initAppConfig() (string, interface{}) {
 	// server config.
 	srvCfg := serverconfig.DefaultConfig()
 	srvCfg.API.Enable = true
-	srvCfg.StateSync.SnapshotInterval = 1500
-	srvCfg.StateSync.SnapshotKeepRecent = 2
```
ValarDragon, line 106: I think this is good for snapshot interval, tbh I'd even consider making it longer

cmd/osmosisd/cmd/init.go
```diff
@@ -102,8 +102,6 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 			config.P2P.MaxNumOutboundPeers = 40
 			config.Consensus.TimeoutCommit = 2 * time.Second
 			config.Mempool.Size = 10000
-			config.StateSync.TrustPeriod = 112 * time.Hour
```
ValarDragon, line 105: Hrmm, this should be parameterized to be upperbounded by your unbonding period. Perhaps a bit lower, what is the default parameterizing it to?

cmd/osmosisd/cmd/init.go
```diff
@@ -102,8 +102,6 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
-			config.FastSync.Version = "v0"
```
ValarDragon, line 106: I have no idea what this parameter is

ValarDragon:
Of the 4 changes, the only one I think we should be really considering is the fast sync one, if we have some understanding of whats getting changed. (Though it confuses me how that relates to state sync)

```
	srvCfg.StateSync.SnapshotInterval = 1500
	srvCfg.StateSync.SnapshotKeepRecent = 2
```

Seems better than upstream setting? If anything we'd want these to be less often ideally imo

PR #3532, opened 2022-11-26
test(superfluid): unpool when not whitelisted

p0mvn:
Closes: #XXX

Add a test for unpooling when not whitelisted

### DIFF
--- a/x/superfluid/keeper/msg_server.go
+++ b/x/superfluid/keeper/msg_server.go
@@ -126,6 +126,11 @@ func (server msgServer) UnPoolWhitelistedPool(goCtx context.Context, msg *types.
 		return nil, err
 	}
 
+	err = server.keeper.checkUnpoolWhitelisted(ctx, msg.PoolId)
+	if err != nil {
+		return nil, err
+	}
+
 	// We get all the lockIDs to unpool
 	lpShareDenom := gammtypes.GetPoolShareDenom(msg.PoolId)
 	minimalDuration := time.Millisecond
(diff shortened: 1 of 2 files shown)


x/superfluid/keeper/msg_server.go
```diff
@@ -126,6 +126,11 @@ func (server msgServer) UnPoolWhitelistedPool(goCtx context.Context, msg *types.
 		return nil, err
 	}
 
+	err = server.keeper.checkUnpoolWhitelisted(ctx, msg.PoolId)
```
p0mvn, line 129: Note: this is also checked down below in `UnpoolAllowedPools` but wouldn't hurt having it here too

x/superfluid/keeper/unpool_test.go
```diff
@@ -252,3 +251,37 @@ func (suite *KeeperTestSuite) TestUnpool() {
 		})
 	}
 }
+
+// TestUnpoolAllowedPools_WhiteList tests that unpooling does not work for pools not in a whitelist.
+// Should fail immediately if pool is not in whitelist.
+func (suite *KeeperTestSuite) TestUnpoolAllowedPools_WhiteList() {
+	// lock id does not matter in the context of this test.
+	const (
+		testLockId        = 1
```
czarcas7ic, line 260: whats the point of this const? I don't see it used elsewhere

czarcas7ic (approved):

x/superfluid/keeper/unpool_test.go
p0mvn: fixed

PR #3533, opened 2022-11-26
feat(superfluid): cli query for unpool whitelist

p0mvn:
Closes: #XXX

CLI query for unpool whitelist


czarcas7ic (approved):
utACK, were you able to test this on LocalOsmosis?

p0mvn:
> utACK, were you able to test this on LocalOsmosis?

I only tested the client (that query exists and can be run)

p0mvn:
Going to merge this since it's just CLI. I will be localnet testing unpool logic in the next 2 days

Issue #3159, opened 2022-10-26
[CL][Swapping]: implement `SwapExactAmountIn` in `concentrated-liquidity`


p0mvn:
## Background

We should implement the method with validations and test cases

## Acceptance Criteria
- implemented
- table-driven tests added

p0mvn:
Blocked by https://github.com/osmosis-labs/osmosis/issues/3156

PR #3535, opened 2022-11-27
feat[CL]: Implement `SwapExactAmountIn`

czarcas7ic:
Closes: #3159

Adds SwapExactAmtIn as well as all logic/tests that depends on this

Implements CalcIn


czarcas7ic:
Just going to refactor this to be all of SwapIn

x/concentrated-liquidity/pool_test.go
```diff
@@ -7,23 +7,7 @@ import (
...
-	upperPrice := sdk.NewDec(5500)
-	s.Require().NoError(err)
-	upperTick := math.PriceToTick(upperPrice) // 86129
-
-	defaultAmt0 := sdk.NewInt(1000000)
-	defaultAmt1 := sdk.NewInt(5000000000)
-
-	swapFee := sdk.ZeroDec()
-
+func (s *KeeperTestSuite) TestCalcAndSwapOutAmtGivenIn() {
```
czarcas7ic, line 10: Note to reviewer: Significantly cleaned up pool tests by merging Calc and Swap tests into one. I know we generally don't want to do this, but this makes complete sense for two reasons:

1. These functions work hand in hand. Calc is run first and should be non mutative. If it mutates, we will see in the following Swap test.
2. Prior to this merge, these two tests used the exact same test cases and was a lot of unnecessary boilerplate

x/concentrated-liquidity/swaps.go
```diff
@@ -134,6 +182,30 @@ func (k Keeper) SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coin, tokenOutDen
...
+	swapFee sdk.Dec,
+	priceLimit sdk.Dec,
+	poolId uint64,
+) (tokenIn sdk.Coin, err error) {
+	tokenIn, actualTokenOut, updatedTick, updatedLiquidity, updatedSqrtPrice, err := k.calcInAmtGivenOut(ctx, desiredTokenOut, tokenInDenom, swapFee, priceLimit, poolId)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	if err := k.applySwap(ctx, tokenIn, actualTokenOut, poolId, updatedLiquidity, updatedTick, updatedSqrtPrice); err != nil {
```
p0mvn, line 198: We should not have this here.

The caller of this function - `SwapExactAmountOut` checks against `tokenInMaxAmount` and fails if the max is exceeded.

However, here we might be mutating the state before checking against max.

In general, there is no reason for having 2 swap methods for each kind of swap. We should do the following - remove `SwapInAmtGivenOut` and `SwapOutAmtGivenIn`. Move all logic to `SwapExactAmountOut` and `SwapExactAmountIn` respectively.

After moving, call `applySwap` after we check against the `tokenInMaxAmount` and `tokenOutMinAmount` respectively.

p0mvn: In this PR, let's move the call to `k.applySwap()` to `k.updatePoolForSwap()`.

Let's create an issue for deleting swap methods in a separate PR please

x/concentrated-liquidity/swaps.go
```diff
@@ -134,6 +182,30 @@ func (k Keeper) SwapOutAmtGivenIn(ctx sdk.Context, tokenIn sdk.Coin, tokenOutDen
...
+		return sdk.Coin{}, err
+	}
+
+	if tokenIn.Amount.GT(tokenIn.Amount) {
```
p0mvn, line 202: This is checking against the same value. Do we have tests covering this logic?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -158,3 +152,145 @@ func (s *KeeperTestSuite) TestCLPoolSimpleSwapExactAmountIn() {
...
+		tokenOut         sdk.Coin
+		tokenInDenom     string
+		tokenInMaxAmount sdk.Int
+		expectedTokenIn  sdk.Int
+	}
+
+	tests := []struct {
+		name      string
+		param     param
+		expectErr bool
```
p0mvn, line 167: Can we start checking for specific errors?

x/concentrated-liquidity/export_test.go
```diff
@@ -58,3 +58,13 @@ func (k Keeper) GetTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64) (ti
 func (k Keeper) SendCoinsBetweenPoolAndUser(ctx sdk.Context, denom0, denom1 string, amount0, amount1 sdk.Int, sender, receiver sdk.AccAddress) error {
 	return k.sendCoinsBetweenPoolAndUser(ctx, denom0, denom1, amount0, amount1, sender, receiver)
 }
+
+// temp test exports until we move calcInAmtGivenOut to CalcInAmtGivenOut
```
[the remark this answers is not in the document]
czarcas7ic: You are correct, we shouldn't, I am still in the mindset of calc and swap being used in PoolI which it will not be anymore, thanks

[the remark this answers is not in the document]
czarcas7ic: I will still have to use the word Temp on the export until we deprecate calc and swap from PoolI though

x/concentrated-liquidity/swaps.go
czarcas7ic: This is an issue from a rebase that I missed, removed

x/concentrated-liquidity/swaps.go
czarcas7ic: This is following the design of what we do for Balancer. Additionally, we updatePoolForSwap in the SwapExact methods. Shouldn't we keep the design as similar as possible between the pool types to prevent confusion?

x/concentrated-liquidity/swaps.go
```diff
@@ -109,18 +109,66 @@ func (k Keeper) SwapExactAmountIn(
...
 	tokenOut sdk.Coin,
 	swapFee sdk.Dec,
 ) (tokenInAmount sdk.Int, err error) {
-	return sdk.Int{}, nil
+	if tokenOut.Denom == tokenInDenom {
+		return sdk.Int{}, errors.New("cannot trade same denomination in and out")
+	}
+
+	// type cast PoolI to ConcentratedPoolExtension
+	pool, ok := poolI.(types.ConcentratedPoolExtension)
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/3559

Done!

x/concentrated-liquidity/swaps.go
p0mvn: It's incorrect to update and persist the pool state when we end up not doing the swap due to price impact protection. That should be changed.

We should aim to follow balancer implementation only when it makes sense. There is no reason for having 2 swap methods. The balancer/gamm has 2 because there is one at the keeper layer and one at the pool layer. Here, we have 2 on the keeper layer when one would suffice.

x/concentrated-liquidity/swaps.go
```diff
@@ -109,18 +109,68 @@ func (k Keeper) SwapExactAmountIn(
...
+		return sdk.Int{}, err
+	}
+
+	// check that the tokenOut calculated is both valid and less than specified limit
+	tokenInAmount = tokenIn.Amount
+	if !tokenInAmount.IsPositive() {
+		return sdk.Int{}, fmt.Errorf("token amount must be positive: got %v", tokenInAmount)
+	}
+	fmt.Printf("tokenInAmount %v \n", tokenInAmount)
+	fmt.Printf("tokenInMaxAmount %v \n", tokenInMaxAmount)
```
p0mvn, line 150:
```suggestion
```

x/concentrated-liquidity/export_test.go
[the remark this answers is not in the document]
p0mvn: Should be removed from `PoolI` now

czarcas7ic:
@p0mvn decided to make it so that if priceLimit is not set (aka it is set to ZeroDec) then we use min or max value depending on swap direction. I felt like this makes things much simpler to deal with. If you are good with that change then this should be good to merge.

x/concentrated-liquidity/swaps.go
czarcas7ic: I didnt see this comment, let me address this before last review

czarcas7ic:
@p0mvn all comments addressed. While applySwap still is in SwapInAmtGivenIn/Out, I also moved it to updatePoolForSwap. We can remove it from SwapInAmtGivenIn/Out once we fully deprecate it's use

x/concentrated-liquidity/swaps_test.go
```diff
@@ -28,7 +22,7 @@ func (s *KeeperTestSuite) TestCLPoolSimpleSwapExactAmountIn() {
 	tests := []struct {
 		name      string
 		param     param
-		expectErr bool
+		expectErr string
```
p0mvn, line 25: Can we make these error types, please?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -158,3 +153,145 @@ func (s *KeeperTestSuite) TestCLPoolSimpleSwapExactAmountIn() {
...
+		tokenOut         sdk.Coin
+		tokenInDenom     string
+		tokenInMaxAmount sdk.Int
+		expectedTokenIn  sdk.Int
+	}
+
+	tests := []struct {
+		name      string
+		param     param
+		expectErr string
```
p0mvn, line 168: ditto for error type

x/concentrated-liquidity/swaps_test.go
```diff
@@ -158,3 +153,145 @@ func (s *KeeperTestSuite) TestCLPoolSimpleSwapExactAmountIn() {
...
+				s.Require().NoError(err)
+				s.Require().Equal(test.param.expectedTokenIn.String(), tokenIn.String())
+
+				gasConsumedForSwap := s.Ctx.GasMeter().GasConsumed() - prevGasConsumed
+				// Check that we consume enough gas that a CL pool swap warrants
+				// We consume `types.GasFeeForSwap` directly, so the extra I/O operation mean we end up consuming more.
+				s.Require().Greater(gasConsumedForSwap, uint64(cltypes.ConcentratedGasFeeForSwap))
+
+				// Assert events
+				s.AssertEventEmitted(s.Ctx, cltypes.TypeEvtTokenSwapped, 1)
```
p0mvn, line 272: nit: I'm wondering if we should remove the swap events from `cltypes` and instead import them from `swaproutertypes`. The reason being - this event should apply for all pools so we should have a single definition for it.

x/concentrated-liquidity/swaps_test.go
```diff
@@ -158,3 +153,145 @@ func (s *KeeperTestSuite) TestCLPoolSimpleSwapExactAmountIn() {
...
+				pool, err = s.App.ConcentratedLiquidityKeeper.GetPoolById(s.Ctx, pool.GetId())
+				s.Require().NoError(err)
+
+				spotPriceAfter := pool.GetCurrentSqrtPrice().Power(2)
+
+				// Ratio of the token out should be between the before spot price and after spot price.
+				tradeAvgPrice := tokenIn.ToDec().Quo(test.param.tokenOut.Amount.ToDec())
+				if !zeroForOne {
+					tradeAvgPrice = sdk.OneDec().Quo(tradeAvgPrice)
+				}
```
p0mvn, line 284: nit: there is another check for `zeroForOne` right below, we can move this logic there

p0mvn (approved):
Some nits. LGTM when addressed!

Issue #3530, opened 2022-11-26
osmoutils.GetIfFound semantics discussion


p0mvn:
Can be changed later / can discuss in an issue, should not block this PR. Perhaps this should be called `HasWithType` ?

e.g. if we were masters of the store interface
```
store.HasProtoMsg("myKey", types.Pool)
```

Has to indicate its a bool (rather than assuming you'd get the object back)

OR is proto.Message here actually mutated? If so lets update the comment, but this doesn't actually make sense to me as an API.

Why don't just make a method called `Get` thats the same as MustGet, but returns an error instead of panics?

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/3512#discussion_r1032655165_

p0mvn:
I think that the method does what we expect it to do - that is return a value if it is found by mutating a `proto.Message` parameter.

This has the same signature as `MustGet` with an extra return boolean indicating whether a value was found at that key.

I think that we should not be returning an error when a value at a given key is not found because it would be difficuly to distinguish a database error from "not found" error for clients.

In my opinion, the problem here is the confusing spec that should be changed to:

```
// GetIfFound returns a value at key by mutating the result parameter. Returns true if the value was found and the
// result mutated correctly. False otherwise. Returns error only when database or serialization errors occur
// returns a boolean indicating whether value exists for the given key and error
func GetIfFound(store store.KVStore, key []byte, result proto.Message) (found bool, err error) {
    // stub
}
```

@ValarDragon let me know what you think

ValarDragon:
Not sure why we don't just make the API

```
func Get(store, key, result) (found bool, err error)
```

p0mvn:
> Not sure why we don't just make the API
> 
> ```
> func Get(store, key, result) (found bool, err error)
> ```
> 
> AFAIK, this is not uncommon

I'm supportive of renaming it to just `Get` and improving the godoc. Will PR a change for this

PR #3537, opened 2022-11-27
refactor(osmoutils): rename GetIfFound to Get, simplify godoc

p0mvn:
Closes: #3530

We would like to rename the `osmoutils` "get" helper and simplify its spec according to the conclusion in #3530

### DIFF
--- a/osmoutils/store_helper.go
+++ b/osmoutils/store_helper.go
@@ -139,7 +139,7 @@ func MustGetDec(store store.KVStore, key []byte) sdk.Dec {
 }
 
 // Get returns a value at key by mutating the result parameter. Returns true if the value was found and the
-// result mutated correctly. False otherwise. Returns false and error only when database or serialization errors occur
+// result mutated correctly. If the value is not in the store, returns false. Returns error only when database or serialization errors occur. (And when an error occurs, returns false)
 func Get(store store.KVStore, key []byte, result proto.Message) (found bool, err error) {
 	b := store.Get(key)
 	if b == nil {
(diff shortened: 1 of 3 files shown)


osmoutils/store_helper.go
```diff
@@ -138,10 +138,9 @@ func MustGetDec(store store.KVStore, key []byte) sdk.Dec {
 	return result.Dec
 }
 
-// GetIfFound returns a value at key by mutating the result parameter. Returns true if the value was found and the
-// result mutated correctly. False otherwise. Returns error only when database or serialization errors occur
-// returns a boolean indicating whether value exists for the given key and error
-func GetIfFound(store store.KVStore, key []byte, result proto.Message) (found bool, err error) {
+// Get returns a value at key by mutating the result parameter. Returns true if the value was found and the
+// result mutated correctly. False otherwise. Returns false and error only when database or serialization errors occur
```
stackman27, line 142: small nit: it was a bit difficult to read
```suggestion
// Get returns a value at key by mutating the result parameter. Returns true if the value was found and the
// result mutated correctly. Returns False otherwise and errors only when database or serialization errors occur
```

stackman27 (approved):
LGTM

ValarDragon (approved):

osmoutils/store_helper.go
ValarDragon: Err, this actually changes the semantic meaning!

ValarDragon, line 142:
```suggestion
// Get returns a value at key by mutating the result parameter. Returns true if the value was found and the
// result mutated correctly. If the value is not in the store, returns false. Returns error only when database or serialization errors occur. (And when an error occurs, returns false)
```

ValarDragon: Made another suggestion for what I think you intended

stackman27: Right on point!

PR #3539, opened 2022-11-27
refactor(CL): gamm msg and query deprecation (pt2 - route to swaprouter)

p0mvn:
Closes: #XXX

This is the next step towards deprecating some gamm messages that have been moved to swaprouter instead of completely eliminating them from the codebase.

This PR implements the deprecated messages and queries by restoring their logic to being as close to main as possible with one main difference: they call into the swaprouter implementation.

It also adds deprecation proto tags to all applicable queries and messages.


x/swaprouter/client/query_proto_wrap.go
```diff
@@ -41,29 +44,16 @@ func (q Querier) Params(ctx sdk.Context,
...
-		return nil, status.Errorf(codes.InvalidArgument, "invalid address: %s", err.Error())
-	}
-
 	tokenIn, err := sdk.ParseCoinNormalized(req.TokenIn)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "invalid token: %s", err.Error())
 	}
 
-	tokenOutAmount, err := q.K.RouteExactAmountIn(ctx, sender, req.Routes, tokenIn, sdk.NewInt(1))
+	tokenOutAmount, err := q.K.MultihopEstimateOutGivenExactAmountIn(ctx, req.Routes, tokenIn)
```
p0mvn, line 56: Note: this calling "route" was a bug

x/swaprouter/client/query_proto_wrap.go
```diff
@@ -87,17 +77,12 @@ func (q Querier) EstimateSwapExactAmountOut(ctx sdk.Context, req queryproto.Esti
...
-		return nil, status.Errorf(codes.InvalidArgument, "invalid address: %s", err.Error())
-	}
-
 	tokenOut, err := sdk.ParseCoinNormalized(req.TokenOut)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "invalid token: %s", err.Error())
 	}
 
-	tokenInAmount, err := q.K.RouteExactAmountOut(ctx, sender, req.Routes, sdkIntMaxValue, tokenOut)
+	tokenInAmount, err := q.K.MultihopEstimateInGivenExactAmountOut(ctx, req.Routes, tokenOut)
```
p0mvn, line 85: Note: this calling "route" was a bug

x/swaprouter/client/query_proto_wrap.go
```diff
@@ -14,6 +14,9 @@ import (
 
 // This file should evolve to being code gen'd, off of `proto/swaprouter/v1beta/query.yml`
 
+type swaprouterQuerierI interface {
+}
+
```
p0mvn, line 19:
```suggestion
```

x/gamm/keeper/keeper.go
```diff
@@ -30,6 +30,7 @@ type Keeper struct {
 	accountKeeper       types.AccountKeeper
 	bankKeeper          types.BankKeeper
 	communityPoolKeeper types.CommunityPoolKeeper
+	swaprouterKeeper    types.SwaprouterKeeper
 }
 
 func NewKeeper(cdc codec.BinaryCodec, storeKey sdk.StoreKey, accountKeeper types.AccountKeeper, bankKeeper types.BankKeeper, communityPoolKeeper types.CommunityPoolKeeper) Keeper {
```
czarcas7ic, line 36: don't we need to add to the return statement:

`swaprouterKeeper: swaprouterKeeper,`

x/gamm/keeper/msg_server.go
```diff
@@ -129,12 +157,54 @@ func (server msgServer) ExitPool(goCtx context.Context, msg *types.MsgExitPool)
...
+	if err != nil {
+		return nil, err
+	}
+
+	tokenInAmount, err := server.keeper.swaprouterKeeper.RouteExactAmountOut(ctx, sender, msg.Routes, msg.TokenInMaxAmount, msg.TokenOut)
+	if err != nil {
+		return nil, err
+	}
+
+	// Swap event is handled elsewhere
```
czarcas7ic, line 198: nit: instead of "elsewhere" we can just say "in swap router" right?

czarcas7ic (approved):
LGTM!

x/gamm/keeper/msg_server.go
p0mvn: This is a direct copy of the old logic. I would like to avoid making changes to avoid more merge conflicts when merging cl-main to main.

Also, the event is emitted somewhere in gamm keeper after persisting pool mutations to state

x/gamm/keeper/keeper.go
p0mvn: No, because swaprouter keeper is being set by the `SetSwapRouterKeeper` function due to the ordering of keeper initialization in `app/keepers`

Issue #3538, opened 2022-11-27
feat(CL): convertToCFMMPool helper


p0mvn:
**Background**

We tend to cast from `PoolI` to `TraditionalAmmInterface` in many places. We should add a helper instead:

```go
func convertToCFMMPooll() (types.TraditionalAmmInterface, error) {
    	cfmmPool, ok := poolI.(types.TraditionalAmmInterface)
	if !ok {
		return nil, fmt.Errorf("given pool does not implement TraditionalAmmInterface, implements %T", poolI)
	}
       return cfmmPool, nil
}
```

and a similar helper for the concentrated pool conversion

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/3497#discussion_r1033010263_

**A/C**
- helper introduced
- tests added
- replace all casts with the use of this helper

ThanhNhann:
Can I handle this ?

p0mvn:
@ThanhNhann thanks for checking in. This is already being addressed by: https://github.com/osmosis-labs/osmosis/pull/3546

PR #3546, opened 2022-11-28
refactor(CL): pool interfaces

p0mvn:
Closes: #3538

Please note that the diff comes from regenerating pool mocks from new interfaces. Actual changes are quite small.

Refactoring pool interfaces:
- Renamed `TraditionalAmmInterface` to `CFMMI`
- Moved "Calc" and "Swap" methods from `PoolI` to `CFMMI`
- Created `convertToCFMMPool` to cast from `PoolI` to `CFMMI`
    * added tests
- Regenerated mocks from the new interfaces


x/gamm/keeper/pool_test.go
```diff
@@ -316,3 +320,38 @@ func (suite *KeeperTestSuite) TestGetPoolAndPoke() {
...
+
+func (suite *KeeperTestSuite) TestConvertToCFMMPool() {
+	ctrl := gomock.NewController(suite.T())
+
+	tests := map[string]struct {
+		pool        swaproutertypes.PoolI
+		expectError bool
+	}{
+		"cfmm pool": {
+			pool: mocks.NewMockCFMMPoolI(ctrl),
```
czarcas7ic, line 332: For my learning, whats the benefit of doing this over running a KeeperTestSuite helper func to make a pool?

x/swaprouter/simulation/sim_msgs.go
```diff
@@ -152,10 +152,15 @@ func createPoolRestriction(k swaprouter.Keeper, sim *simtypes.SimCtx, ctx sdk.Co
 	}
 }
 
-func getRandPool(k simulationKeeper, sim *simtypes.SimCtx, ctx sdk.Context) (uint64, types.PoolI, sdk.Coin, sdk.Coin, []string, string, error) {
+// TODO: refactor this to work with simulateion pools
```
czarcas7ic, line 155:
```suggestion
// TODO: refactor this to work with simulation pools
```

czarcas7ic (approved):
LGTM 👍

x/gamm/keeper/pool_test.go
p0mvn: Our helpers create a pool but don't return it. Instead, only return their id.

So we have to also call `GetPoolById()` to get the pool itself. It's more verbose.

Here, the goal is to simply test that given one pool implementation, the function passes. Given the other implementation, the function fails. So using mocks seemed to be the least verbose and simplest way to achieve that.

Issue #3157, opened 2022-10-25
[CL][Liquidity Provision]: add table-driven tests for all core methods


p0mvn:
## Background

I propose using this as a tracking issue. For each core method, create a separate PR with tests.

**Please aim to avoid testing more than one method / function per PR to make sure that we can do a detailed review.**

The list of things to test:
- [ ] `initOrUpdateTick`
- [ ] `GetTickInfo`
- [ ] `getPoolById`
- [ ] `initOrUpdatePosition`
- [ ] `GetPosition`
- [ ] `createPosition`

Add godoc for each method you are testing

## Acceptance Criteria
- tests added
- coverage maximized
- godocs added
- aim for 1 method per PR

RusAkh:
Hello. Is anyone working on it?

p0mvn:
@RusAkh I think @mattverse and @czarcas7ic have the most context on the test vectors here.

@mattverse have you made any progress on this?

ThanhNhann:
Hey @mattverse, can I handle one of the tests above?

PR #3558, opened 2022-11-28
tests[CL]: initOrUpdateTick table driven tests (1/x)

czarcas7ic:
Partially Closes: #3157 

Adds table driven test for initOrUpdateTick

- [x] tests added
- [x] coverage maximized
- [x] godocs added
- [x] aim for 1 method per PR

### DIFF
--- a/x/concentrated-liquidity/keeper_test.go
+++ b/x/concentrated-liquidity/keeper_test.go
@@ -9,6 +9,22 @@ import (
 	"github.com/osmosis-labs/osmosis/v13/app/apptesting"
 )
 
+var (
+	DefaultLowerPrice       = sdk.NewDec(4545)
+	DefaultLowerTick        = int64(84222)
+	DefaultUpperPrice       = sdk.NewDec(5500)
+	DefaultUpperTick        = int64(86129)
+	DefaultCurrPrice        = sdk.NewDec(5000)
+	DefaultCurrTick         = int64(85176)
+	DefaultCurrSqrtPrice, _ = DefaultCurrPrice.ApproxSqrt() // 70.710678118654752440
+	DefaultZeroSwapFee      = sdk.ZeroDec()
+	ETH                     = "eth"
+	DefaultAmt0             = sdk.NewInt(1000000)
+	USDC                    = "usdc"
+	DefaultAmt1             = sdk.NewInt(5000000000)
+	DefaultLiquidityAmt     = sdk.NewDec(50000000000)
+)
+
 type KeeperTestSuite struct {
 	apptesting.KeeperTestHelper
 }
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -31,7 +31,7 @@ func (k Keeper) getPoolById(ctx sdk.Context, poolId uint64) (types.ConcentratedP
 	return &pool, nil
 }
 
-// TODO: spec and test
+// poolExists returns true if a pool with the given id exists. False otherwise.
 func (k Keeper) poolExists(ctx sdk.Context, poolId uint64) bool {
 	store := ctx.KVStore(k.storeKey)
 	pool := model.Pool{}
--- a/x/concentrated-liquidity/pool.go
+++ b/x/concentrated-liquidity/pool.go
@@ -32,6 +32,18 @@ func (k Keeper) getPoolById(ctx sdk.Context, poolId uint64) (types.ConcentratedP
 }
 
 // TODO: spec and test
+func (k Keeper) poolExists(ctx sdk.Context, poolId uint64) bool {
+	store := ctx.KVStore(k.storeKey)
+	pool := model.Pool{}
+	key := types.KeyPool(poolId)
+	found, err := osmoutils.GetIfFound(store, key, &pool)
+	if err != nil {
+		panic(err)
+	}
+	return found
+}
+
+// TODO: spec and test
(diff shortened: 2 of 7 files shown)


x/concentrated-liquidity/tick.go
```diff
@@ -14,7 +14,15 @@ import (
 	types "github.com/osmosis-labs/osmosis/v13/x/concentrated-liquidity/types"
 )
 
+// initOrUpdateTick retrieves the tickInfo from the specified tickIndex and updates both the liquidityNet and LiquidityGross.
+// if we are initializing or updating an upper tick, we subtract the liquidityIn from the LiquidityNet
+// if we are initializing or updating an lower tick, we add the liquidityIn from the LiquidityNet
 func (k Keeper) initOrUpdateTick(ctx sdk.Context, poolId uint64, tickIndex int64, liquidityIn sdk.Dec, upper bool) (err error) {
+	_, err = k.getPoolById(ctx, poolId)
```
p0mvn, line 21: Should we implement `hasPool` method?

x/concentrated-liquidity/keeper_test.go
```diff
@@ -9,6 +9,22 @@ import (
...
+	DefaultUpperTick        = int64(86129)
+	DefaultCurrPrice        = sdk.NewDec(5000)
+	DefaultCurrTick         = int64(85176)
+	DefaultCurrSqrtPrice, _ = DefaultCurrPrice.ApproxSqrt() // 70.710678118654752440
+	DefaultZeroSwapFee      = sdk.ZeroDec()
+	ETH                     = "eth"
+	DefaultAmt0             = sdk.NewInt(1000000)
+	USDC                    = "usdc"
+	DefaultAmt1             = sdk.NewInt(5000000000)
+	DefaultLiquidityAmt     = sdk.NewDec(50000000000)
```
p0mvn, line 25: Love this. Have we switched all hard-coded occurrences to utilize these?

x/concentrated-liquidity/tick_test.go
```diff
@@ -121,3 +121,206 @@ func (s *KeeperTestSuite) TestNextInitializedTick() {
...
+		upper       bool
+	}
+
+	tests := []struct {
+		name                   string
+		param                  param
+		tickExists             bool
+		expectedLiquidityNet   sdk.Dec
+		expectedLiquidityGross sdk.Dec
+		expectErr              bool
```
p0mvn, line 139: Do you think we can start testing specific errors?

x/concentrated-liquidity/tick_test.go
```diff
@@ -121,3 +121,206 @@ func (s *KeeperTestSuite) TestNextInitializedTick() {
...
+	}{
+		{
+			name: "Init tick 50 with 50000000000 liquidity, upper",
+			param: param{
+				poolId:      1,
```
p0mvn, line 144: nit: we can make it as `const validPoolId = 1` and reuse

p0mvn (approved):
Awesome work

x/concentrated-liquidity/keeper_test.go
czarcas7ic: Its from [this PR](https://github.com/osmosis-labs/osmosis/pull/3535) which switches to the use of these!

x/concentrated-liquidity/tick_test.go
czarcas7ic: Sorry I may be missing something, but we cant just remove this as a param and set it as a constant since we have a test case that tests an invalid poolId of 2 right? How would I overwrite the const in this case?

czarcas7ic: Ohh I see what you are saying now, yeah I can change this

x/concentrated-liquidity/tick.go
czarcas7ic: Do you mean just create a func in pool.go such as `func (k Keeper) hasPool(ctx sdk.Context, poolId uint64)` that just returns a bool if it exists or not?

p0mvn: yeah

czarcas7ic: okay, added a `poolExists` method!

x/concentrated-liquidity/pool.go
```diff
@@ -31,6 +31,18 @@ func (k Keeper) getPoolById(ctx sdk.Context, poolId uint64) (types.ConcentratedP
 	return &pool, nil
 }
 
+// TODO: spec and test
+func (k Keeper) poolExists(ctx sdk.Context, poolId uint64) bool {
```
p0mvn, line 35: Let's test it right away

x/concentrated-liquidity/tick_test.go
```diff
@@ -121,3 +121,207 @@ func (s *KeeperTestSuite) TestNextInitializedTick() {
...
+		upper       bool
+	}
+
+	tests := []struct {
+		name                   string
+		param                  param
+		tickExists             bool
+		expectedLiquidityNet   sdk.Dec
+		expectedLiquidityGross sdk.Dec
+		expectedErr            string
```
p0mvn, line 140: Can we have error here instead of string?

If someone changes an error message, all tests have to be updated. On the other hand, if we expect an error here, the additional changes wouldn't be necessary

czarcas7ic:
@p0mvn comments addressed!

p0mvn (approved):
🥪

Issue #3556, opened 2022-11-28
tracking(CL): merge to main


p0mvn:
## Background

This issue is for tracking the progress of merging `concentrated-liqudiity-main` to main.

As we are merging in chunks with the goal of splitting state-break logic from state-compatible, we end up commenting out some functionality and tests.

This issue tracks these areas to make sure nothing is left uncommented

## TODOs

- [x] `x/swaprouter/types`
   * [x] merge in `codec.go`
   * [x] uncomment authz test: https://github.com/osmosis-labs/osmosis/blob/bc4ed47025b5a363d5e3194f981eb15d1b6ecbff/x/swaprouter/types/msgs_test.go#L279-L280
   * [x] uncomment `GetSizeBytes()` for `MsgSwapExactAmountIn`: https://github.com/osmosis-labs/osmosis/blob/bc4ed47025b5a363d5e3194f981eb15d1b6ecbff/x/swaprouter/types/msgs.go#L43
   * [x] uncomment `GetSizeBytes()` for `MsgSwapExactAmountOut`: https://github.com/osmosis-labs/osmosis/blob/bc4ed47025b5a363d5e3194f981eb15d1b6ecbff/x/swaprouter/types/msgs.go#L83
   * godoc: https://github.com/osmosis-labs/osmosis/pull/3764/files#r1052738361

- `x/swaprouter/cli`
   * re-enable cli tests: https://github.com/osmosis-labs/osmosis/pull/3557/files#r1034133974
   * re-enable query tests: https://github.com/osmosis-labs/osmosis/pull/3557/files#r1034134838

- [x] x/swaprouter/module`
   * merge module.go to enable type registrations and simulation
          * [x] https://github.com/osmosis-labs/osmosis/pull/3557/files#r1034136480
          * [x] https://github.com/osmosis-labs/osmosis/pull/3557/files#r1034137022

- [x] `x/swaprouter`
   * re-enable: https://github.com/osmosis-labs/osmosis/pull/3557/files#r1034135904

- [x] `app/keepers/keepers.go`
   * remove and initialize keeper: https://github.com/osmosis-labs/osmosis/blob/72925e82fae4106d36d67828b32ab752fe4a123a/app/keepers/keepers.go#L124-L125

- [x] `x/gamm/simulation`:
   * move pool creation fee to swaprouter simulation: https://github.com/osmosis-labs/osmosis/pull/3738#discussion_r1050093787

- `x/gamm`
[...]

PR #3560, opened 2022-11-29
feat(CL): create CFMMPoolI interface, refactor PoolI

p0mvn:
Closes: #XXX

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


x/gamm/keeper/pool.go
```diff
@@ -284,3 +285,15 @@ func (k Keeper) setStableSwapScalingFactors(ctx sdk.Context, poolId uint64, scal
 
 	return k.setPool(ctx, stableswapPool)
 }
+
+// convertToCFMMPool converts PoolI to CFMMPoolI by casting the input.
+// Returns the pool of the CFMMPoolI or error if the given pool does not implement
+// CFMMPoolI.
+// nolint: unused
```
p0mvn, line 292: Note: removing this is tracked in https://github.com/osmosis-labs/osmosis/issues/3556

x/gamm/keeper/pool.go
```diff
@@ -12,21 +12,22 @@ import (
...
 
-func (k Keeper) MarshalPool(pool types.PoolI) ([]byte, error) {
+func (k Keeper) MarshalPool(pool swaproutertypes.PoolI) ([]byte, error) {
 	return k.cdc.MarshalInterface(pool)
 }
 
-func (k Keeper) UnmarshalPool(bz []byte) (types.PoolI, error) {
-	var acc types.PoolI
+func (k Keeper) UnmarshalPool(bz []byte) (types.CFMMPoolI, error) {
+	var acc types.CFMMPoolI
```
czarcas7ic, line 23: Just to clarify, is it intended to Marshal from PoolI but Unmarshal into CFMMPoolI?

x/gamm/keeper/share.go
```diff
@@ -3,11 +3,12 @@ package keeper
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
-	"github.com/osmosis-labs/osmosis/v13/x/gamm/keeper/internal/events"
 	"github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+	"github.com/osmosis-labs/osmosis/v13/x/swaprouter/events"
+	swaproutertypes "github.com/osmosis-labs/osmosis/v13/x/swaprouter/types"
 )
 
-func (k Keeper) applyJoinPoolStateChange(ctx sdk.Context, pool types.PoolI, joiner sdk.AccAddress, numShares sdk.Int, joinCoins sdk.Coins) error {
+func (k Keeper) applyJoinPoolStateChange(ctx sdk.Context, pool types.CFMMPoolI, joiner sdk.AccAddress, numShares sdk.Int, joinCoins sdk.Coins) error {
```
czarcas7ic, line 11: why does this specifically call for CFMMPoolI but the others use PoolI?

czarcas7ic (approved):
LGTM, approving and then can merge on @ValarDragon approval due to state break as discussed previously 🏄🏻‍♂️

x/gamm/keeper/share.go
p0mvn: There was no reason for this being `CFMMPoolI`. So I updated it to `PoolI` which has fewer methods.

x/gamm/keeper/pool.go
ValarDragon: Its unclear to me that its safe to Marshal to CFMMPoolI, I think we may have to always marshal and unmarshal from PoolI.

(Or at least need to double check the serialized bytecode. IIRC, the proto import path + struct name does get serialized into @type)

x/gamm/pool-models/balancer/codec.go
```diff
@@ -20,8 +21,13 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
...
 	registry.RegisterInterface(
-		"osmosis.gamm.v1beta1.PoolI",
-		(*types.PoolI)(nil),
+		"osmosis.swaprouter.v1beta1.PoolI",
+		(*swaproutertypes.PoolI)(nil),
+		&Pool{},
+	)
+	registry.RegisterInterface(
+		"osmosis.gamm.v1beta1.CFMMPoolI",
+		(*types.CFMMPoolI)(nil),
```
ValarDragon, line 30: imo we need to keep the old names here, otherwise we would need a relatively complex state migration.

ValarDragon: Alternative is we make state migration code that takes the old type and casts it to a new type. This seems like a complex integrator break, that doesn't seem worth it imo

x/gamm/keeper/pool.go
p0mvn: I'm going to look into this.

x/gamm/pool-models/balancer/codec.go
p0mvn: Something to keep in mind on this matter - we have an e2e test that pre-creates pools before the upgrade, runs some swaps against them, performs the upgrade, and runs some swaps again.

If the e2e test is passing, do you think still that the migrations would be necessary?

p0mvn: Either way, I will try to understand whether this is unsafe by looking into `RegisterInterface(...)` logic

x/gamm/keeper/pool.go
p0mvn: Please see test: [d7d388d](https://github.com/osmosis-labs/osmosis/pull/3560/commits/d7d388d30cf8d773f327d455d6e7c672b0cdad4d)

p0mvn: Just noticed that you specified serializing specifically to JSON. Why would JSON matter? Would the test above cover your concern?

x/gamm/pool-models/balancer/codec.go
p0mvn: Reverted to old name to avoid breaking clients:

[9f01762](https://github.com/osmosis-labs/osmosis/pull/3560/commits/9f01762419a4c8941bd637579e59e41ab0892d1f)

x/gamm/types/codec.go
```diff
@@ -24,8 +24,8 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
 
 func RegisterInterfaces(registry types.InterfaceRegistry) {
 	registry.RegisterInterface(
-		"osmosis.gamm.v1beta1.PoolI",
-		(*PoolI)(nil),
+		"osmosis.gamm.v1beta1.CFMMPoolI",
```
p0mvn, line 27:
```suggestion
		"osmosis.gamm.v1beta1.PoolI",
```

x/gamm/types/codec.go
```diff
@@ -24,8 +24,8 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
...
+		(*CFMMPoolI)(nil),
```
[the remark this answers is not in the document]
p0mvn: fixed

x/gamm/keeper/pool.go
ValarDragon: JSON was suggested just for debugging

x/gamm/keeper/pool.go
```diff
@@ -12,21 +12,22 @@ import (
 	"github.com/osmosis-labs/osmosis/v13/x/gamm/pool-models/balancer"
 	"github.com/osmosis-labs/osmosis/v13/x/gamm/pool-models/stableswap"
 	"github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+	swaproutertypes "github.com/osmosis-labs/osmosis/v13/x/swaprouter/types"
 )
 
-func (k Keeper) MarshalPool(pool types.PoolI) ([]byte, error) {
+func (k Keeper) MarshalPool(pool swaproutertypes.PoolI) ([]byte, error) {
 	return k.cdc.MarshalInterface(pool)
```
ValarDragon, line 19: err does this work without casting to CFMMPoolI ?

x/gamm/types/codec.go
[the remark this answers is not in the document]
ValarDragon: Now im even more confused, we had one entry at time of this comment, now were back at two entries.

I thought this PR was jsut renaming PoolI -> CFMMPoolI. If so, then there should just be one interface registered, with the interface route being the old one.

ValarDragon (approved):
Reviewed synchronously! LGTM, nice job

Theres a test confirming these are serialization equivalent (so the interface serialization didn't include this), we just can't have another interface, that uses this as a constituent interface.

I think for backwards consistency, we should still make the pairs in all 3 spots:
"osmosis.gamm.v1beta1.PoolI" -> CFMMPoolI 

"osmosis.swaprouter.v1beta1.PoolI" -> swaprouter.PoolI

p0mvn:
Paths updated as discussed so merging this to continue progressing with the swaprouter merge

PR #3563, opened 2022-11-29
Add Pagination to Pool With Filter Query

mattverse:
Closes: #XXX

This PR adds pagination to pools with filter query.
As of right now, pagination is not included in the query for pools with filter, whist pagination is necessary as front end query heavily depends on pagination.

## Brief Changelog

- Add pagination to `PoolsWithFilter` query in gamm


x/gamm/keeper/grpc_query.go
```diff
@@ -184,14 +175,14 @@ func (q Querier) PoolsWithFilter(ctx context.Context, req *types.QueryPoolsWithF
 		checks_needed++
 	}
 
-	for _, p := range pools {
+	var response = []*codectypes.Any{}
+	pageRes, err := query.FilteredPaginate(poolStore, req.Pagination, func(_, value []byte, accumulate bool) (bool, error) {
```
mattverse, line 179: Using `FilteredPaginate` instead of `Paginate` method allows us to iterate until pagination count has been reached

x/gamm/keeper/grpc_query.go
```diff
@@ -184,14 +175,14 @@ func (q Querier) PoolsWithFilter(ctx context.Context, req *types.QueryPoolsWithF
...
 		var checks = 0
-		var pool types.PoolI
-
-		err := q.cdc.UnpackAny(p, &pool)
+		pool, err := q.Keeper.UnmarshalPool(value)
```
czarcas7ic, line 181: where does this `value` as the input to UnmarshalPool come from?

mattverse: It comes from the FilteredPagination method, where it iterates over the given kv store!

x/gamm/keeper/grpc_query.go
```diff
@@ -226,23 +198,37 @@ func (q Querier) PoolsWithFilter(ctx context.Context, req *types.QueryPoolsWithF
...
 		}
 
 		if checks == checks_needed {
-			response = append(response, p)
+			any, err := codectypes.NewAnyWithValue(pool)
+			if err != nil {
+				return false, err
+			}
+
+			if accumulate {
```
mattverse, line 215: required for `offset` in pagination

x/gamm/keeper/grpc_query.go
ValarDragon: https://github.com/cosmos/cosmos-sdk/issues/12242

We can backport the fix for that to clean this up a little if we want.
(Or just copy/paste https://github.com/cosmos/cosmos-sdk/pull/12253 into osmoutils)

mattverse: I was also checking generic pagination as well! Sad that we don't have it yet. I say we refactor this to use generic pagination later on when it gets backported and for now leave as is

ValarDragon (approved):

Issue #2962, opened 2022-10-07
Call to GetArithmeticTwap with wrong order of arguments


migueldingli1997:
#### System information

Osmosis version: main (at time of writing)
OS & Version: n/a
Commit hash: `8a19d8b3edb53c18c2607750ac8e9b5b3b37e71b`

#### Expected behaviour

All callers of this function provide the base asset followed by the quote asset

https://github.com/osmosis-labs/osmosis/blob/8a19d8b3edb53c18c2607750ac8e9b5b3b37e71b/x/twap/api.go#L31-L38

#### Actual behaviour

Some callers under `wasmbinding/queries.go` call the function with quote asset followed by base asset

https://github.com/osmosis-labs/osmosis/blob/8a19d8b3edb53c18c2607750ac8e9b5b3b37e71b/wasmbinding/queries.go#L122

and

https://github.com/osmosis-labs/osmosis/blob/8a19d8b3edb53c18c2607750ac8e9b5b3b37e71b/wasmbinding/queries.go#L140

#### Steps to reproduce the behaviour

n/a

#### Backtrace

n/a

mattverse:
Oh good catch! Thanks for this issue, we should be solving this by the next upgrade, meanwhile, querying via correct arguments should be able to be done via stargate queries instead of bindings

migueldingli1997:
I don't know how problematic this is, but I noticed that the order is incorrect in the unit tests as well. And if the order is switched to `base, quote` some unit tests start failing.

https://github.com/osmosis-labs/osmosis/blob/8a19d8b3edb53c18c2607750ac8e9b5b3b37e71b/x/twap/api_test.go#L329-L331

p0mvn:
I believe this is now solved. Closing for now. Thank you @migueldingli1997

p0mvn:
Actually, this is not solved. I mistook it for another refactor we did. Keeping open until further investigations

p0mvn:
@mattverse seeing that you are assigned here. Any updates on this? I'm still seeing some twap tests having assets misplaced

PR #3564, opened 2022-11-29
fix/test(twap): remove v2 query proto wrap, undeprecate v1, test v1 twap API

p0mvn:
Closes: #2962

We added v2 twap query proto wrap that misplaced parameters. This PR removes the faulty v2.

Additionally, it covers the original queries with tests to ensure that parameters are correct now.

Lastly, although `wasmbindings` are deprecated, they were also broken so this PR fixes them.


x/twap/client/query_proto_wrap_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(QueryTestSuite))
+}
+
+func (suite *QueryTestSuite) SetupTest() {
+	suite.Setup()
+}
+
+func (suite *QueryTestSuite) TestV2QueryTwap() {
```
p0mvn, line 31:
```suggestion
func (suite *QueryTestSuite) TestQueryTwap() {
```

x/twap/client/query_proto_wrap.go
```diff
@@ -18,7 +18,10 @@ type Querier struct {
 func (q Querier) ArithmeticTwap(ctx sdk.Context,
 	req queryproto.ArithmeticTwapRequest, // nolint: staticcheck
 ) (*queryproto.ArithmeticTwapResponse, error) {
-	if (req.EndTime == nil || *req.EndTime == time.Time{}) {
+	if req.EndTime == nil {
+		req.EndTime = &time.Time{}
+	}
+	if (*req.EndTime == time.Time{}) {
```
ValarDragon, line 24: Why did this get changed? Old behavior looks correct to me

ValarDragon (approved):
This LGTM, as long as were very sure Twap v1 got it right

x/twap/client/query_proto_wrap.go
p0mvn: if `req.EndTime` is nil, the old behavior would try to dereference a nil pointer.

PR #3565, opened 2022-11-29
test[CL]: getPoolById table driven tests

czarcas7ic:
Closes: #XXX

Partially Closes: #3157 

Adds table driven test for initOrUpdateTick

- [x] tests added
- [x] coverage maximized
- [x] godocs added
- [x] aim for 1 method per PR

### DIFF
--- a/x/concentrated-liquidity/pool_test.go
+++ b/x/concentrated-liquidity/pool_test.go
@@ -1021,112 +1021,53 @@ func (s *KeeperTestSuite) TestOrderInitialPoolDenoms() {
 }
 
 func (s *KeeperTestSuite) TestGetPoolById() {
-	s.SetupTest()
+	const validPoolId = 1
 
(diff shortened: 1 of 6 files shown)


x/concentrated-liquidity/pool_test.go
```diff
@@ -1058,93 +1089,3 @@ func (s *KeeperTestSuite) TestPoolExists() {
 	// ensure that this returns false
 	s.Require().False(poolExists)
 }
-
-// func (s *KeeperTestSuite) TestCalcInAmtGivenOut() {
```
p0mvn, line 1062: I'm guessing that these are re-introduced in: https://github.com/osmosis-labs/osmosis/pull/3535?

p0mvn (approved):
Amazing! Love how well-tested and documented CL is becoming!

x/concentrated-liquidity/pool_test.go
czarcas7ic: Yep!

PR #3566, opened 2022-11-29
fix/test: misplaced assets in twap tests

p0mvn:
Closes: #2962

This is a continuation of #3564 that fixes twap test asset misplace. There are no state-breaking changes here

### DIFF
--- a/x/twap/api_test.go
+++ b/x/twap/api_test.go
@@ -154,18 +154,19 @@ func makeSimpleTwapInput(startTime time.Time, endTime time.Time, isQuoteTokenA b
 
 func makeSimpleThreeAssetTwapInput(startTime time.Time, endTime time.Time, baseQuoteAB, baseQuoteCA, baseQuoteBC bool) []getTwapInput {
 	var twapInput []getTwapInput
-	formatSimpleTwapInput(twapInput, startTime, endTime, baseQuoteAB, denom0, denom1, 2)
-	formatSimpleTwapInput(twapInput, startTime, endTime, baseQuoteAB, denom2, denom0, 2)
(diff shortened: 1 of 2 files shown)


x/twap/api_test.go
```diff
@@ -154,18 +154,19 @@ func makeSimpleTwapInput(startTime time.Time, endTime time.Time, isQuoteTokenA b
...
 }
 
-func formatSimpleTwapInput(twapInput []getTwapInput, startTime time.Time, endTime time.Time, baseQuoteAB bool, denomA, denomB string, poolID uint64) {
+func formatSimpleTwapInput(twapInput []getTwapInput, startTime time.Time, endTime time.Time, baseQuoteAB bool, denomA, denomB string, poolID uint64) []getTwapInput {
 	quoteAssetDenom, baseAssetDenom := denomA, denomB
 	if baseQuoteAB {
 		baseAssetDenom, quoteAssetDenom = quoteAssetDenom, baseAssetDenom
 	}
 	twapInput = append(twapInput, getTwapInput{poolID, quoteAssetDenom, baseAssetDenom, startTime, endTime})
+	return twapInput
```
czarcas7ic, line 169: I think this was fine as it was previously, unless I am missing something further in the PR

czarcas7ic (approved):
Change makes sense to me 👍

x/twap/api_test.go
p0mvn: The slice only gets mutated if we are modifying pre-allocated elements in it.

The reason is that slice is just a pointer to buffer with length and capacity under the hood.

So when we pass in a slice we are making a copy of these 3 values.

Appends end up allocating a new buffer that the caller's slice pointer isn't referencing.

These values never end up being updated without changes here. And some tests weren't even running previously

PR #3568, opened 2022-11-29
app/apptesting: Reduce memory allocations in `KeeperTestHelper.Setup()`

RusAkh:
Component of: #2148 

Memprofile `KeeperTestHelper.Setup()` and reduce memory allocs 

Using Benchmark added in #2143

## Steps Taken 
- [x] Cache encoding config in `app/encoding.go`
 ```
Initial benchmark:
	BenchmarkSetup-8             514           2258866 ns/op         1364230 B/op      17732 allocs/op
Benchmark after caching encoding config: 
	BenchmarkSetup-8             534           1902947 ns/op         1077572 B/op      14312 allocs/op
```
- [ ] modulemanager.InitGenesis
     `encoding/json.Unmarshal()` takes too much resources and time to do its job. There are alternatives to json's `Unmarshal()`, one of which is [jsoniter](https://github.com/json-iterator/go), which claims to be faster than `encoding/json` package
- [ ] iavl.LoadVersion()
     Starting up of archive nodes (improving performance of `iavl.Load.Version()`) is tackled in this PR: https://github.com/cosmos/iavl/pull/638

### CODE REVIEW

**Comment by ValarDragon** [comment]
Thanks for picking this up! Apologies though, having a hard time understanding the benchmark pasted.

Is that the output for a particular function?

We can see overall impact by doing 
`go test -benchmem -run=^$ -bench ^BenchmarkSetup$ github.com/osmosis-labs/osmosis/v13/app/apptesting` before and after. 

You'd get something like:
```
BenchmarkSetup-16    	     398	   2977221 ns/op	 1356578 B/op	   17590 allocs/op
```
And we should see both time & allocs/op go down

RusAkh:
oh, it is from `go tool pprof`. Yes, you are right, it shows memory allocations for a particular function. I will rerun it the way you specified and add it here :)

ValarDragon:
This LGTM, nice! Other followups involve work in other repos, so I recommend marking this ready for review, and we get this merged?

ValarDragon (approved):

PR #3583, opened 2022-11-30
test[CL]: getTickInfo table driven tests (5/6)

czarcas7ic:
Partially Closes: #3157 

Adds table driven test for getTickInfo

- [x] tests added
- [x] coverage maximized
- [x] godocs added
- [x] aim for 1 method per PR

### DIFF
--- a/x/concentrated-liquidity/tick.go
+++ b/x/concentrated-liquidity/tick.go
@@ -57,6 +57,9 @@ func (k Keeper) getTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64) (ti
 	store := ctx.KVStore(k.storeKey)
 	tickStruct := model.TickInfo{}
 	key := types.KeyTick(poolId, tickIndex)
+	if !k.poolExists(ctx, poolId) {
+		return tickInfo, types.PoolNotFoundError{PoolId: poolId}
+	}
 
 	found, err := osmoutils.GetIfFound(store, key, &tickStruct)
 	// return 0 values if key has not been initialized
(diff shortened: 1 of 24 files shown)


x/concentrated-liquidity/tick_test.go
```diff
@@ -263,3 +264,51 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
...
+			expectedTickInfo: model.TickInfo{LiquidityGross: DefaultLiquidityAmt, LiquidityNet: DefaultLiquidityAmt.Neg()},
+		},
+		{
+			name:             "Get tick info on existing pool with no existing tick",
+			poolToGet:        validPoolId,
+			tickToGet:        DefaultCurrTick.Int64() + 1,
+			expectedTickInfo: model.TickInfo{LiquidityGross: sdk.ZeroDec(), LiquidityNet: sdk.ZeroDec()},
+		},
+		{
+			name:             "Get tick info on a non-existing pool with no existing tick",
```
p0mvn, line 288: Why do we not return an error in such a case instead? Doesn't make sense to return a valid tick info with zero values in such a case

czarcas7ic: Okay, I added a poolExists check. Basically if the pool does not exist, it returns an error with nil values for tickInfo. If the pool exists but the ticks just arent initialized, it returns a valid tickInfo object but with zero values for liquidity

x/concentrated-liquidity/tick.go
```diff
@@ -57,6 +57,9 @@ func (k Keeper) getTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64) (ti
 	store := ctx.KVStore(k.storeKey)
 	tickStruct := model.TickInfo{}
 	key := types.KeyTick(poolId, tickIndex)
+	if !k.poolExists(ctx, poolId) {
+		return tickInfo, types.PoolNotFoundError{PoolId: poolId}
```
p0mvn, line 61: nit: I know that these are equivalent. However, the change guarantees that we return an empty struct even if someone accidentally modifies the `tickInfo` prior to returning

```suggestion
		return model.TickInfo{}, types.PoolNotFoundError{PoolId: poolId}
```

p0mvn (approved):
Great work!

PR #3587, opened 2022-11-30
ProtoRev: Proto code and Init scaffolding code

davidterpay:

This PR is the first step in building and deploying the `x/protorev` module. Currently, we have implemented an end to end working V1 of the module, but are breaking down the implementation into modular PRs. The current V1 can be found [here](https://github.com/skip-mev/osmosis-protorev/tree/v1BackUp).

This PR includes all of the proto and basic app configuration code. Much of the boilerplate code for the module is also in this PR. Although this includes all of the proto code that will be used end to end, some of it will be implemented in subsequent PRs.

## Brief Changelog
 
  - *Init proto code for the module*
  - *Hook up `x/protorev` to the app*
  - *Genesis code + stateless validation*
  - *Keeper interfaces (dependencies with gamm, epoch, etc.)*
  - *Implementing the Msg types + stateless validation*
  - *Governance proposals handlers + stateless validation*
  - *`TokenPairArbRoutes` stateless validation* (read more about this in the spec linked below)

## Testing and Verifying

  - *Added unit tests for stateless validation for the genesis state*
  - *Added unit tests for stateless validation for Msgs*
  - *Added unit tests for stateless validation of governance proposals*

The current WIP spec for `x/protorev` can be found [here](https://skip-protocol.notion.site/x-protorev-Spec-3ba5cb368f314120bc514eb2d0a6a0dd). Most answers surrounding design decisions can be found in the spec (feel free to ask if something is missing). 

#### Genesis State
Genesis state can be configured with a preset number of hot routes. Additionally, it can be configured to be on or off by default (through params).

#### TokenPairArbRoutes
There are two methods of which the module will create cyclic arbitrage routes (i.e. trades). One will use the highest liquidity pools and the other will use a manually entered set of searcher routes (hot routes) that are typically used on chain but may not be included in the highest liquidity pool method. TokenPairArbRoutes is how all of the hot routes are stored in the module.

#### Governance Proposals
There are two proposals that will be implemented:
1. Enabling/Disabling `x/protorev`
2. Setting the Admin account

The first proposal will allow the chain to turn the module on/off if desired. The second proposal allows the chain to select an admin account that will be setting hot routes.

#### Queries
There are several queries that will be implemented in a later PR. Most of them allow users to request information about how profitable `x/protorev` has been.

#### Msgs
There are two valid messages that will be implemented:
1. SetHotRoutes
2. SetDeveloperAccount

Both of these Msgs can only be called by the admin account which must be initialized through governance.

#### Expected Keepers
The module primarily interacts with the `x/gamm` module. However, it also needs access to the `bankkeeper` in order to mint and burn coins for trades and send profits to the developer address (which will be Skip's address). Additionally, there are stateful changes that need to occur every so often so the epoch keeper is also expected.


proto/osmosis/protorev/v1beta1/protorev.proto
```diff
@@ -0,0 +1,51 @@
...
+  // The denom of token A that is traded
+  string denomA = 2;
+  // The denom of token B that is traded
+  string denomB = 3;
+}
+
+// PoolStatistics tracks the arbitrarge statistics for a given pool
+message PoolStatistics {
+  // Mapping between denom and the profits earned for this pool
+  map<string, cosmos.base.v1beta1.Coin> profits = 1;
```
ValarDragon, line 48: Need to be careful that this is deterministic under current SDK serialization :(

https://github.com/cosmos/cosmos-sdk/issues/12816

x/protorev/types/errors.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	ErrInvalidRoute       = sdkerrors.Register(ModuleName, 1, "This is not a valid cyclic route")
+	ErrDuplicateTokenPair = sdkerrors.Register(ModuleName, 2, "This token pair has already been added")
+	ErrInvalidTokenName   = sdkerrors.Register(ModuleName, 3, "This is not a valid token name")
+	ErrInvalidParams      = sdkerrors.Register(ModuleName, 4, "Invalid params")
+	ErrInvalidArbDenom    = sdkerrors.Register(ModuleName, 5, "Invalid denom for arbitrage")
```
ValarDragon, line 12: I recommend not doing registered errors, since the codespaces make error type not safely changeable in patch releases. Perhaps just an errors.New or fmt.ErrorF() ?

ValarDragon (approved):
LGTM, aside from my 2 comments (in particular the one about maps)

proto/osmosis/protorev/v1beta1/protorev.proto
bpiv400: Thanks for highlighting this. 

We'll follow the same standards as you use in the rest of the code base. Can you point us to a canonical example where you're taking data stored in a map and serializing it safely? 

Is the best way to just order by keys and store as a pair of arrays?

ValarDragon: We don't actually use proto maps anywhere in the codebase at the moment. We annoyingly always do some manual "unrolling" of this, of making a data structure with both items as fields, and make a repeated list of them where we want to include them. We then make a linear time `Get` function. (Asymptotically were still at linear time, because deserializing the map requires reading all of its fields as well)

Super silly that this is the status quo right now :/

bpiv400: That makes sense. 

I mean could you point us to a place where you can do this unrolling, so we can adopt this approach. 

Sounds like you don't unpack the whole list of key, value pairs into an in-memory map if I'm understanding correctly? You just iterate over the list each time you need a key. This seems fine (same asymptotic time) if you only access 1 key in expectation each time you unserialize. 

But if we access potentially many in expectation, it feels like it would make more sense to: 
- store using the key value approach you all use 
- iterate over the entire list to store in an in-memory map o(n) 
- do all our logic with o(1) accesses in memory instead of o(n) accesses 
- if there are changes, recreate the list of key, value pairs with a deterministic ordering when we need to store again 

(Disclaimer: I haven't done SDK development, so maybe there's not a way to have this separation of stored data and in memory usage of it as I'm imagining it)

x/protorev/types/errors.go
davidterpay: Got it. We will refactor and replace the registered errors with fmt.ErrorF().

proto/osmosis/protorev/v1beta1/protorev.proto
davidterpay: Thats good to know! We can do the manual unrolling (as mentioned) and have everything handled in the KV stores by adding a new (key, value) as oppose to using the map. Given that, I'll deprecate PoolStatistics.

davidterpay:
Pushed changes to remove map usage from the PoolStatistics proto and refactored error handling. @ValarDragon would you happen you know why the `e2e` test suite is failing? Is it safe to merge without it passing?


Screen Shot 2022-12-04 at 5 15 54 PM

czarcas7ic:
This is not safe to merge yet, I believe its a bug. Here are the logs from the container running the test that it halted on

```
1:34AM INF indexed block exents height=125 module=txindex
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd519d4]

goroutine 3991 [running]:
github.com/cosmos/iavl.(*nodeDB).incrVersionReaders(0x2a8edc8?, 0x20?)
        github.com/cosmos/iavl@v0.19.4/nodedb.go:943 +0x24
github.com/cosmos/iavl.newExporter(0x400a52ce60)
        github.com/cosmos/iavl@v0.19.4/export.go:46 +0xd0
github.com/cosmos/iavl.(*ImmutableTree).Export(...)
        github.com/cosmos/iavl@v0.19.4/immutable_tree.go:159
github.com/cosmos/cosmos-sdk/store/iavl.(*Store).Export(0x400a527758?, 0x2a7a3e0?)
        github.com/cosmos/cosmos-sdk@v0.46.6/store/iavl/store.go:284 +0x54
github.com/cosmos/cosmos-sdk/store/rootmulti.(*Store).Snapshot(0x4001059080, 0x7d, {0x2a6f340, 0x400a25c930})
        github.com/cosmos/cosmos-sdk@v0.46.6/store/rootmulti/store.go:733 +0x538
github.com/cosmos/cosmos-sdk/snapshots.(*Manager).createSnapshot(0x40001d0a20, 0x48a3f4?, 0x40018a4740?)
        github.com/cosmos/cosmos-sdk@v0.46.6/snapshots/manager.go:216 +0x88
created by github.com/cosmos/cosmos-sdk/snapshots.(*Manager).Create
        github.com/cosmos/cosmos-sdk@v0.46.6/snapshots/manager.go:203 +0x1f8
```

It is failing where we lock and add to existing lock

app/keepers/keepers.go
```diff
@@ -263,6 +266,12 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 
 	appKeepers.EpochsKeeper = epochskeeper.NewKeeper(appKeepers.keys[epochstypes.StoreKey])
 
+	protorevKeeper := protorevkeeper.NewKeeper(
+		appCodec, appKeepers.keys[protorevtypes.StoreKey],
+		appKeepers.GetSubspace(protorevtypes.ModuleName),
+		appKeepers.AccountKeeper, appKeepers.BankKeeper, appKeepers.GAMMKeeper, appKeepers.EpochsKeeper)
+	appKeepers.ProtoRevKeeper = &protorevKeeper
```
alexanderbez, line 273: Why do we need a reference to this keeper?

proto/osmosis/protorev/v1beta1/protorev.proto
```diff
@@ -0,0 +1,57 @@
...
+// TokenPairArbRoutes tracks all of the hot routes for a given pair of tokens
+message TokenPairArbRoutes {
+  option (gogoproto.equal) = true;
+
+  // Stores all of the possible hot paths for a given pair of tokens
+  repeated Route arb_routes = 1;
+  // Token denomination of the first asset
+  string token_a = 2;
+  // Token denomination of the second asset
+  string token_b = 3;
```
alexanderbez, line 20: Could we consider better nomenclature for these? Perhaps in/out?

alexanderbez: also nit: consistent Proto fields using `_` (snake-case).

x/protorev/types/expected_keepers.go
```diff
@@ -0,0 +1,39 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	gammtypes "github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+
```
alexanderbez, line 7:
```suggestion
```

x/protorev/types/genesis.go
```diff
@@ -0,0 +1,74 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// InputAmountList contains a list of input amounts to test when creating the optimal amount to swap in the
+// binary search method
+var InputAmountList []sdk.Int
+
+// AtomDenomination stores the native denom name for Atom on chain used for route building
+var AtomDenomination string = "ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2"
```
alexanderbez, line 14: Should this be hard-coded as such? What about an on-chain parameter?

x/protorev/types/keys.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	// ModuleName defines the module name
+	ModuleName = "protorev"
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// RouterKey defines the module's message routing key
+	RouterKey = ModuleName
```
alexanderbez, line 11: I don't think we need this.

x/protorev/types/utils.go
```diff
@@ -0,0 +1,46 @@
...
+func UInt64ToBytes(number uint64) []byte {
+	bz := make([]byte, 8)
+	binary.LittleEndian.PutUint64(bz, number)
+	return bz
+}
+
+// Converts a []byte into a uint64
+func BytesToUInt64(bz []byte) uint64 {
+	return binary.LittleEndian.Uint64(bz)
+}
```
alexanderbez, line 15: There should already exist functions like this in `types`.  If note, we should define them there.

alexanderbez (approved):

x/protorev/types/genesis.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+
+	return gs.Params.Validate()
+}
+
+func init() {
+	// Init all of the input amounts
+	InputAmountList = make([]sdk.Int, 0)
+	for i := 0; i < 1000000000; i += 100000 {
```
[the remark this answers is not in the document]
NotJeremyLiu: @alexanderbez V1 of protorev uses a binary search algorithm to determine the optimal input amount for max profits given a cylic arb route. Our range of input amounts is from 0-1000 osmo, with .1 osmo-sized steps.

x/protorev/types/genesis.go
davidterpay: These two denoms must be known by module at run-time since the trading executed by protorev will always start and end with either atom or osmo. We can configure this so that we initialize this in the genesis state instead of having it be a hardcoded value.

x/protorev/types/utils.go
davidterpay: correct, sdk implements `Uint64ToBigEndian` which will do the job. deprecating this now

x/protorev/types/keys.go
davidterpay: Is this not needed for query/message services (which are implemented in later PRs)?

x/protorev/types/genesis.go
bpiv400: IMO, probably better in this case for it to be hardcoded, as we don't want it to use anything other than osmo or atom, for the foreseeable future

proto/osmosis/protorev/v1beta1/protorev.proto
bpiv400: @davidterpay @NotJeremyLiu my understanding was there wasn't a consistent "in/out" associated with these?

app/keepers/keepers.go
```diff
@@ -507,6 +516,7 @@ func (appKeepers *AppKeepers) initParamsKeeper(appCodec codec.BinaryCodec, legac
 	paramsKeeper.Subspace(incentivestypes.ModuleName)
 	paramsKeeper.Subspace(lockuptypes.ModuleName)
 	paramsKeeper.Subspace(poolincentivestypes.ModuleName)
+	paramsKeeper.Subspace(protorevtypes.ModuleName).WithKeyTable(protorevtypes.ParamKeyTable())
```
bpiv400, line 519: QQ: What's the reason protorev is the only module using ParamKeyTable?

bpiv400 (approved):

app/keepers/keepers.go
davidterpay: Its essentially just initializing the subspace and parameters here instead of initializing on genesis. I'll change it so that it conforms to the standard.

proto/osmosis/protorev/v1beta1/protorev.proto
davidterpay: We made a design decision to make the route building logic cleaner and constraining the type of data that is inputted to protorev instead of having reversal logic built in as well. With this design, anytime we see a swap with a -> b we can instantly find all hot routes that are valid without having to do any additional logic (or compute). If we also want hot routes for b -> a, we have to pass in a additional hot routes (i.e. the hot routes are always ordered and directed in the correct direction of the trade).

Issue #3540, opened 2022-11-27
test(twap): add overflow tests for geometric TWAP


p0mvn:
## Background

Geometric TWAP needs overflow tests [here](https://github.com/osmosis-labs/osmosis/blob/7ad6f2ed7bae04227e29c3d196b20747a4431962/x/twap/logic_test.go#L728-L731)

## Suggested Design

Add potentially overflowing tests with:
- max spot price
- large time delta
   * 2 days
   * 100 days
- both

If math functions panic, identify and document the maximum values at which the tests do not panic.

## Acceptance Criteria
- requested tests added
- if panics, occur find the maximum values that do not and create a follow-up issue to investigate

PR #3588, opened 2022-12-01
test(osmomath): log tests around max spot price

p0mvn:
Closes: #XXX

This is related to: https://github.com/osmosis-labs/osmosis/issues/3540

Upon adding tests around max spot price for geometric twap, overflows started to occur.

To eliminate the possibility of log function being at fault, the tests around max spot price values are added directly to the log. Benchmark is also added to ensure that large values do not inflate the time it takes to run the log. The difference was indeed negligible.


p0mvn:
Hi @czarcas7ic . Could you please review this change when you get the chance?

stackman27 (approved):
ACK

Issue #3543, opened 2022-11-27
test(twap): ensuere test coverage for `twapLog` and `twapPow`


p0mvn:
## Background

We should test the helper math functions introduced in #3420 

## Suggested Design

The test should be as simple as validating that the base for pow and log a chosen from `geometricTwapMathBase` defined in `logic.go`

## Acceptance Criteria
- test for `twapPow` added
- test for `twapLog` added
- validating that the base for math function is as defined by `geometricTwapMathBase`

ThanhNhann:
Can I handle this?

p0mvn:
@ThanhNhann please do. thanks!

PR #3591, opened 2022-12-01
Nhan/test twappow twaplog

ThanhNhann:
Closes: #3543 

Write table driven test for `twapLog` and `twapPow` function

## Testing and Verifying
All tests are pass

  - How is the feature or change documented? ( not documented)


x/twap/logic_test.go
```diff
@@ -1596,3 +1596,129 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+			priceValue: osmomath.MustNewDecFromStr("912648174127941279170121098210.928219201902041311"),
+			// From: https://www.wolframalpha.com/input?i2d=true&i=log+base+2+of+912648174127941279170121098210.928219201902041311+with+20+digits
+			expected: osmomath.MustNewDecFromStr("99.525973560175362367"),
+		},
+	}
+	for name, test := range tests {
+		s.Run(name, func() {
+			osmoassert.ConditionalPanic(s.T(), test.expectedPanic, func() {
+				result := twap.TwapLog(test.priceValue.SDKDec())
+				require.True(osmomath.DecApproxEq(s.T(), test.expected, osmomath.BigDecFromSDKDec(result), expectedErrTolerance))
```
p0mvn, line 1663: We need to make sure that we are testing that the following value is used for the base: `geometricTwapMathBase`. This would also require two simple tests case as opposed to many table driven.

That would imply exporting it for tests via `export_test.go`.

Test case 1:
- Performing `twap.Log` and `osmomath.CustomLog` with `geometricTwapMathBase` and comparing the values.

Test case 2: using one of the hardcoded values above. For example, `Twaplog{912648174127941279170121098210.928219201902041311} = 99.525973560175362367`

No need to keep other test cases to avoid increasing LOC. The underlying log implementation should be testing this thoroughly

x/twap/logic_test.go
```diff
@@ -1596,3 +1596,129 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+			expected: osmomath.MustNewDecFromStr("1.99999999"),
+		},
+		"TwapPow(1); invalid; panic": {
+			exponentValue: osmomath.OneDec(),
+			expectedPanic: true,
+		},
+	}
+	for name, test := range tests {
+		s.Run(name, func() {
+			osmoassert.ConditionalPanic(s.T(), test.expectedPanic, func() {
```
p0mvn, line 1718: Similar idea here for pow

x/twap/logic_test.go
```diff
@@ -1596,3 +1596,129 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+				result := twap.TwapPow(test.exponentValue.SDKDec())
+				require.True(osmomath.DecApproxEq(s.T(), test.expected, osmomath.BigDecFromSDKDec(result), expectedErrTolerance))
```
p0mvn, line 1720: Similar suggestion for the log test above
```suggestion
				suite.Require().True(test.expected, osmomath.BigDecFromSDKDec(result), expectedErrTolerance))
```

p0mvn:
Nice work @ThanhNhann . Please see my earlier comments

x/twap/logic_test.go
ThanhNhann: I have updated the code with your suggestions about comparing values, can you plz check again?

x/twap/logic_test.go
```diff
@@ -1596,3 +1597,135 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+		"TwapLog(1024) = -4.3219280948873623479": {
+			priceValue: osmomath.MustNewDecFromStr("0.05"),
+			// From:  https://www.wolframalpha.com/input?i2d=true&i=log+base+2+of+0.05+with+20+digits
+			expected: osmomath.MustNewDecFromStr("-4.3219280948873623479"),
+		},
+		"Twaplog{1024.987654321} = 10.00139081765414132": {
+			priceValue: osmomath.NewDecWithPrec(1024987654321, 9),
+			// From: https://www.wolframalpha.com/input?i2d=true&i=log+base+2+of+1024.987654321+with+19+digits
+			expected: osmomath.MustNewDecFromStr("10.00139081765414132"),
+		},
```
p0mvn, line 1653: Can we keep just one test case, please? I even think we can convert it from table-driven to a regular Go test case.

The reason for not needing detailed tests here is that low-level math is already being tested in `osmomath`.

The goal here is to verify that the desired base is used AND one sanity hand-calculated check.

I appreciate the hard work of adding all these test cases! However, asking to remove it with the goal of aiming to avoid complexity unnecessarily.

```suggestion
```

x/twap/logic_test.go
p0mvn: Can you please apply these changes everywhere where applicable?

x/twap/logic_test.go
p0mvn: Nice work!

x/twap/logic_test.go
ThanhNhann: Yeah, I'm understanding the code inside to be able to change according to your suggestion is like this, right?
`s.Require().True(test.expected.Sub(osmomath.BigDecFromSDKDec(result)).Abs().LTE(expectedErrTolerance))`
or I can write another func for this test's type because `osmomath.DecApproxEq` is not being used for `suite`, tell me if I am wrong something, thanks!

x/twap/logic_test.go
```diff
@@ -1596,3 +1597,57 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+		"Twaplog{912648174127941279170121098210.928219201902041311} = 99.525973560175362367": {
+			priceValue: osmomath.MustNewDecFromStr("912648174127941279170121098210.928219201902041311"),
+			// From: https://www.wolframalpha.com/input?i2d=true&i=log+base+2+of+912648174127941279170121098210.928219201902041311+with+20+digits
+			expected: osmomath.MustNewDecFromStr("99.525973560175362367"),
+		},
+	}
+	for name, test := range tests {
+		s.Run(name, func() {
+			osmoassert.ConditionalPanic(s.T(), test.expectedPanic, func() {
+				result := twap.TwapLog(test.priceValue.SDKDec())
```
p0mvn, line 1618: Since we've only kept one test case, can we convert this from table-driven into a regular Go test case now?

x/twap/logic_test.go
```diff
@@ -1596,3 +1597,57 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+		},
+	}
+	for name, test := range tests {
+		s.Run(name, func() {
+			osmoassert.ConditionalPanic(s.T(), test.expectedPanic, func() {
+				result := twap.TwapPow(test.exponentValue.SDKDec())
+				result_by_mathPow := math.Pow(twap.GeometricTwapMathBase.MustFloat64(), test.exponentValue.SDKDec().MustFloat64())
+
+				s.Require().True(test.expected.Sub(osmomath.BigDecFromSDKDec(result)).Abs().LTE(expectedErrTolerance))
+				s.Require().True(osmomath.MustNewDecFromStr(fmt.Sprint(result_by_mathPow)).Sub(osmomath.BigDecFromSDKDec(result)).Abs().LTE(expectedErrTolerance))
```
p0mvn, line 1649: I think this might work. Please double-check

```suggestion
				s.Require().True(osmomath.BigDecFromSDKDec(result_by_mathPow).Sub(osmomath.BigDecFromSDKDec(result)).Abs().LTE(expectedErrTolerance))
```

x/twap/logic_test.go
```diff
@@ -1596,3 +1597,57 @@ func geometricTestCaseFromDeltas0(startAccum, accumDiff sdk.Dec, timeDelta time.
...
+	tests := map[string]struct {
+		exponentValue osmomath.BigDec
+		expected      osmomath.BigDec
+		expectedPanic bool
+	}{
+		"TwapPow(0.5) = 1.41421356": {
+			exponentValue: osmomath.MustNewDecFromStr("0.5"),
+			//From: https://www.wolframalpha.com/input?i2d=true&i=power+base+2+exponent+0.5+with+9+digits
+			expected: osmomath.MustNewDecFromStr("1.41421356"),
+		},
```
p0mvn, line 1640: Same comment for removing table-driven boilerplate. The reason is that we are only keeping one test case

x/twap/logic_test.go
p0mvn: The current change LGTM!

x/twap/logic_test.go
ThanhNhann: Updated

x/twap/logic_test.go
p0mvn: What's the reason for making it float64 in the first place? Why not keep it as `sdk.Dec`?

p0mvn: Ah it's because we are using `math.Pow`. OK that sounds good!

p0mvn (approved):

PR #3600, opened 2022-12-02
Reduce memory consumption for state exports

ValarDragon:
Closes: #XXX

We've recently been hitting memory consumption issues with state exports, as we maintain several caches during state export. (IAVL internal cache, a second IAVL cache within the SDK, CacheKV cache, exported values, JSON marshal overhead)

This eliminates the `CacheKV cache` which is itself some degree of memory amplification over the size of all state, thus hopefully meaningfully lowering our overall memory consumption. We need some more profiling to know for sure where the highest time/RAM overhead is. 

This still needs to be sanity tested for correctness, and that it actually speeds things up.

There are other actions we can do to reduce the overall RAM overhead. Will try to make an issue outlining this.

## Brief Changelog

Make state export not use CacheKV

## Testing and Verifying

I plan to manually verify by getting a state export from a node with and without this, and check sha-sum being the same (post sorting)

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? TODO

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -29,7 +29,7 @@ func (app *OsmosisApp) ExportAppStateAndValidators(
 	}
 
 	genState := app.mm.ExportGenesisForModules(ctx, app.appCodec, modulesToExport)
-	appState, err := json.MarshalIndent(genState, "", "  ")
+	appState, err := json.Marshal(genState)
 	if err != nil {
 		return servertypes.ExportedApp{}, err
 	}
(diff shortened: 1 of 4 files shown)


niccoloraspa:
great! is it ready to be tested?

ValarDragon:
Yeah it is ready for testing! Just has to be cherry-picked onto a node with the respective version

Issue #3245, opened 2022-11-04
[CL][Concentrated Liquidity Module]: implement `MsgCreateConcentratedPool`


p0mvn:
## Background

This issue is blocked on: https://github.com/osmosis-labs/osmosis/issues/3243

Upon successfully implementing the above refactor, and implement the new `MsgCreateConcentratedPool` in the `concentratedpool` package.

Follow the structure and design similar to `MsgCreateBalancerPool`, `MsgCreateStableswapPool`

Implement the message server method:

```go
func (server msgServer) CreateConcentratedPool(goCtx context.Context, msg *balancer.MsgCreateConcentratedPool) (*balancer.MsgCreateConcentratedPoolResponse, error) {
}
```

Add tests

## Acceptance Criteria
- The message is implemented
- Follows the similar folder structure and design to `MsgCreateBalancerPool`, `MsgCreateStableswapPool`
- tests added

PR #3607, opened 2022-12-03
feat[CL]: add MsgCreateConcentratedPool

czarcas7ic:
Closes: #3245 

Implementation of MsgCreateConcentratedPool and required wiring

## Brief Changelog

*(for example:)*
 

x/concentrated-liquidity/keeper_test.go
```diff
@@ -39,6 +40,13 @@ func (suite *KeeperTestSuite) SetupTest() {
 	suite.Setup()
 }
 
+// SetupDefaultPool sets up a eth usdc concentrated liquid pool with pool ID 1 and no liquidity
+func (s *KeeperTestSuite) SetupDefaultPool(ctx sdk.Context) types.ConcentratedPoolExtension {
```
p0mvn, line 44: Because most of the existing pool creation helpers are using this naming convention.

```suggestion
func (s *KeeperTestSuite) PrepareDefaultPool(ctx sdk.Context) types.ConcentratedPoolExtension {
```

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
 	}
 
-	// calculate liquidity created from this position
+	// N.B. we only write cache context if actual amounts
+	// returned are greater than the given minimums.
+	cacheCtx, writeCacheCtx := ctx.CacheContext()
```
p0mvn, line 42: Let's move this to be right before the call to `updatePosition`. I don't see a reason to be initializing this before validation checks that can fail

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+
+	currSqrtPrice := pool.GetCurrentSqrtPrice()
+	currentTick := pool.GetCurrentTick()
+
+	// If the currentSqrtPrice and currentTick are zero, then this is the first position to be created for this pool.
+	// We therefore calculate the sqrtPrice and currentTick based on the inputs of this position
+	// TODO: We need to review this logic very carefully, as I am not 100 percent convinced it is safe / correct
+	if currSqrtPrice.Equal(sdk.ZeroDec()) && currentTick.Equal(sdk.ZeroInt()) {
+		if amount0Desired.Equal(sdk.ZeroInt()) || amount1Desired.Equal(sdk.ZeroInt()) {
```
p0mvn, line 51: Is it not possible for me to create a single-asset position below or above the current tick?

p0mvn: Ah I see that's because we want to establish the current price

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+	// returned are greater than the given minimums.
+	cacheCtx, writeCacheCtx := ctx.CacheContext()
+
+	currSqrtPrice := pool.GetCurrentSqrtPrice()
+	currentTick := pool.GetCurrentTick()
+
+	// If the currentSqrtPrice and currentTick are zero, then this is the first position to be created for this pool.
+	// We therefore calculate the sqrtPrice and currentTick based on the inputs of this position
+	// TODO: We need to review this logic very carefully, as I am not 100 percent convinced it is safe / correct
+	if currSqrtPrice.Equal(sdk.ZeroDec()) && currentTick.Equal(sdk.ZeroInt()) {
```
p0mvn, line 50: Can we encapsulate all this into a method `isInitialPosition(...)` and cover by tests?

x/concentrated-liquidity/lp.go
p0mvn, line 51: And all the logic below encapsulate into `createInitialPosition` and cover by tests

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+		if amount0Desired.Equal(sdk.ZeroInt()) || amount1Desired.Equal(sdk.ZeroInt()) {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, types.InitialLiquidityZeroError{Amount0: amount0Desired, Amount1: amount1Desired}
+		}
+		currentSpotPrice := amount1Desired.Quo(amount0Desired).ToDec()
+		currSqrtPrice, err = currentSpotPrice.ApproxSqrt()
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+		}
+		newTick := math.PriceToTick(currentSpotPrice)
```
p0mvn, line 59:
```suggestion
		initialTick := math.PriceToTick(initialSpotPrice)
```

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+	currentTick := pool.GetCurrentTick()
+
+	// If the currentSqrtPrice and currentTick are zero, then this is the first position to be created for this pool.
+	// We therefore calculate the sqrtPrice and currentTick based on the inputs of this position
+	// TODO: We need to review this logic very carefully, as I am not 100 percent convinced it is safe / correct
+	if currSqrtPrice.Equal(sdk.ZeroDec()) && currentTick.Equal(sdk.ZeroInt()) {
+		if amount0Desired.Equal(sdk.ZeroInt()) || amount1Desired.Equal(sdk.ZeroInt()) {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, types.InitialLiquidityZeroError{Amount0: amount0Desired, Amount1: amount1Desired}
+		}
+		currentSpotPrice := amount1Desired.Quo(amount0Desired).ToDec()
```
p0mvn, line 54:
```suggestion
		initialSpotPrice := amount1Desired.Quo(amount0Desired).ToDec()
```

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+		currSqrtPrice, err = currentSpotPrice.ApproxSqrt()
```
p0mvn, line 55:
```suggestion
		initialSqrtPrice, err = initialSpotPrice.ApproxSqrt()
```

x/concentrated-liquidity/lp.go
```diff
@@ -37,16 +37,39 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+		}
+		newTick := math.PriceToTick(currentSpotPrice)
+		pool.SetCurrentSqrtPrice(currSqrtPrice)
+		pool.SetCurrentTick(newTick)
```
p0mvn, line 61:
```suggestion
		pool.SetCurrentSqrtPrice(initialSqrtPrice)
		pool.SetCurrentTick(initialTick)
```

x/concentrated-liquidity/lp.go
p0mvn: I see not that this is being used by `k.setPool`.

Maybe in the future PR, we can make the `updatePosition` method take `pool` instead of `poolId`. That way we wouldn't have to call `k.setPool` prior to `updatePosition`.

I think it is best to gather all mutative function in one place - `updatePosition`.

x/concentrated-liquidity/msg_server.go
```diff
@@ -6,21 +6,22 @@ import (
...
-func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
+func NewMsgServerImpl(keeper *Keeper) clmodel.MsgServer {
 	return &msgServer{
 		keeper: keeper,
 	}
 }
 
 // TODO: tests, including events
-func (server msgServer) CreatePosition(goCtx context.Context, msg *types.MsgCreatePosition) (*types.MsgCreatePositionResponse, error) {
+func (server msgServer) CreatePosition(goCtx context.Context, msg *clmodel.MsgCreatePosition) (*clmodel.MsgCreatePositionResponse, error) {
```
p0mvn, line 24: Can we keep this message in `types`? Only `MsgCreatePool` should be in the `model` package since it's the only one related to the pool model

x/concentrated-liquidity/msg_server.go
```diff
@@ -52,11 +53,11 @@ func (server msgServer) CreatePosition(goCtx context.Context, msg *types.MsgCrea
 		),
 	})
 
-	return &types.MsgCreatePositionResponse{Amount0: actualAmount0, Amount1: actualAmount1}, nil
+	return &clmodel.MsgCreatePositionResponse{Amount0: actualAmount0, Amount1: actualAmount1}, nil
 }
 
 // TODO: tests, including events
-func (server msgServer) WithdrawPosition(goCtx context.Context, msg *types.MsgWithdrawPosition) (*types.MsgWithdrawPositionResponse, error) {
+func (server msgServer) WithdrawPosition(goCtx context.Context, msg *clmodel.MsgWithdrawPosition) (*clmodel.MsgWithdrawPositionResponse, error) {
```
p0mvn, line 60: Same for this

x/concentrated-liquidity/msg_server.go
p0mvn: This is alligned with gamm module where we only keep `MsgCreatePool` definitions in the model packages

x/swaprouter/creator.go
```diff
@@ -64,7 +67,10 @@ func (k Keeper) CreatePool(ctx sdk.Context, msg types.CreatePoolMsg) (uint64, er
 		return 0, err
 	}
 
-	k.poolCreationListeners.AfterPoolCreated(ctx, sender, pool.GetId())
+	// TODO: Ensure this is correct decision. Without this hack, we try to make gauges for the CL pool which isn't needed
+	if poolType != types.Concentrated {
+		k.poolCreationListeners.AfterPoolCreated(ctx, sender, pool.GetId())
+	}
```
p0mvn, line 73: Let's create an issue for this and link it here directly. We should create a separate hook called `AfterCFMMPoolCreated`, call it in `gamm`'s `InitializePool`. The gauges should be created upon that new hook being called, not the `AfterPoolCreated`.

x/swaprouter/creator.go
```diff
@@ -84,14 +90,17 @@ func validateCreatePoolMsg(ctx sdk.Context, msg types.CreatePoolMsg) error {
...
-	if numAssets < types.MinPoolAssets {
-		return types.ErrTooFewPoolAssets
-	}
-	if numAssets > types.MaxPoolAssets {
-		return errors.Wrapf(
-			types.ErrTooManyPoolAssets,
-			"pool has too many PoolAssets (%d)", numAssets,
-		)
+
+	if msg.GetPoolType() != types.Concentrated {
```
p0mvn, line 94: Why would we not run this for the concentrated pool? I suggest keeping this as it was. If we want the CL pool have stricted bounds, we can have an  extra check in `InitializePool`.

The swaproter logic should not be concerned with what kind of pool is being created.

That is, we should avoid `msg.GetPoolType() != types.Concentrated` in swaprouter as much as possible

x/swaprouter/creator.go
```diff
@@ -110,17 +120,21 @@ func (k Keeper) validateCreatedPool(
...
+		if !pool.GetTotalPoolLiquidity(ctx).IsEqual(initialPoolLiquidity) {
+			return errors.Wrap(types.ErrInvalidPool,
+				"Pool was attempted to be created, with initial liquidity not equal to what was specified.")
+		}
+		// TODO: this check should be moved
+		// This check can be removed later, and replaced with a minimum.
+		if !pool.GetTotalShares().Equal(gammtypes.InitPoolSharesSupply) {
+			return errors.Wrap(types.ErrInvalidPool,
+				"Pool was attempted to be created with incorrect number of initial shares.")
+		}
```
p0mvn, line 137: Can we move this into `gamm` under `validateCFMMPool` method and call it right at the beginning of `InitializePool`?

p0mvn (commented):
Nice work!

x/concentrated-liquidity/lp.go
czarcas7ic: Yeah, right now it needs to be initialized there for setting the pool for the first position. Agreed on taking pool for a future PR, I will make an issue, add it here, then resolve conversation.

x/swaprouter/creator.go
czarcas7ic: Yea, I suppose as long as we check num assets in the Validate of CL itself then this is fine, will add it back

czarcas7ic: Actually, after I removed it I remembered why it was needed. The length of pool is checked via pool liquidity, and since we make CL pools with zero liquidity, this check wont pass. Lmk if you want me to approach this in another way, but I personally don't see a problem with adding this if statement here

x/swaprouter/creator.go
czarcas7ic: I dont think so, because we would need to pass initialPoolLiquidity into it somehow

czarcas7ic:
@p0mvn addressed most of your comments with a few followups!

x/swaprouter/creator.go
p0mvn: I think we might want to move this to each of `MsgCreateBalancerPool` and `MsgCreateStableswapPool`'s `ValidateBasic`.

We can still create a reusable helper that does these checks and simply call that helper from every applicable `ValidateBasic`.

Let me know what you think

p0mvn:
@czarcas7ic thanks, replied to the remaining unresolved discussions

x/swaprouter/creator.go
czarcas7ic: But then we would be checking if pool.GetTotalLiquidity() isEqual to pool.GetTotalLiquidity() 😅

p0mvn: Ah I see. In that case, this check seems unnecessary. Validation such as this (checking that pool was initialized with correct liquidity and shares) should be done by unit tests, not at runtime.

We might want to make sure that the unit tests exist and remove these

x/swaprouter/creator.go
czarcas7ic: It isn't reusable since StableSwap calls Coins direct and Balancer calls a PoolAssets array. I do agree with defining directly in the validate basic and will do so now

czarcas7ic: Oh actually they are already defined!

https://github.com/osmosis-labs/osmosis/blob/244b28a450d3e64c24d2c8a71c4fe56cf7865348/x/gamm/pool-models/stableswap/msgs.go#L63-L67

and 

https://github.com/osmosis-labs/osmosis/blob/244b28a450d3e64c24d2c8a71c4fe56cf7865348/x/gamm/pool-models/balancer/msgs.go#L42-L45

czarcas7ic: I could actually generalize this if we move it up to the swaprouter level, since we then have commonality at the msg.InitialLiquidity level. Otherwise I will just leave the respective logic to be done at the ValidateBasic level (i think this is the best approach imo)

p0mvn: I'm supportive of leaving it as is in gamm for this PR to limit the scope. However, is my understanding correct that we would still remove this from the swap router?

czarcas7ic: Yes, otherwise we are just validating twice

czarcas7ic:
@p0mvn all concerns addressed!

czarcas7ic:
@p0mvn I got rid of the internal model package but for some reason proto check is failing on this as well as the go test. Is this normal?

p0mvn:
@czarcas7ic I think there might be an import that is still referencing the old package

czarcas7ic:
@p0mvn Just double checked and there isnt :/

czarcas7ic:
I think that its a false error. It was passing before, and the changes I made dont effect this..

czarcas7ic:
Yeah, I reverted back to a point where it was passing and it still fails, so I think its something to do with CI

czarcas7ic:
Closing in favor of https://github.com/osmosis-labs/osmosis/pull/3687

PR #3620, opened 2022-12-05
feat(scripts): polynomial and rational approximations

p0mvn:
Progress towards: https://github.com/osmosis-labs/osmosis/issues/3013

# Math Operations Approximations

## Context

This is a set of scripts to approximate a mathematical operation using polynomial
and rational approximations.

The `main` script is in its respective function of `main.py`. It does the following:

1. Computes polynomial and rational approximations of a given function (e^x by default), 
returning the coefficients.

1. Computes (x,y) coordinates for every kind of approximation given the same x coordinates.
Plots the results for rough comparison.

1. Plots the results for rough comparison.

2. Computes the max error for every approximation given the same x coordinates.

3. Computes and plots max errors for every approximation with a varying number of parameters.

In other words, this script runs various approximation methods, plots their results and deltas
from actual function values. It can be configured to print the maximum error.
The exact behavior is controlled by the global variables at the top of `main.py`.

The following are the resources used to create this:
- <https://xn--2-umb.com/22/approximation>
- <https://sites.tufts.edu/atasissa/files/2019/09/remez.pdf>

In `main.py`, there is also an `exponent_approximation_choice` script.

This is a shorter and simpler version of `main` that isolates the 13-parameter
Chebyshev Rational approximation of e^x. We are planning to use it in production.
Therefore, we need to perform coefficient truncations to 36 decimal points
(the max `osmomath` supported precision). This truncation is applied
to `exponent_approximation_choice` but not `main`.

## Configuration

Several parameters can be changed on the needs basis at the
top of `main.py`.

Some of the parameters include the function we are approximating, the [x_start, x_end] range of
the approximation, and the number of terms to be used. For the full parameter list, see `main.py`.

## Usage

Assuming that you are in the root of the repository and have Sympy installed:

```bash
# Create a virtual environment.
python3 -m venv ~/approx-venv

# Start the environment
source ~/approx-venv/bin/activate

# Install dependencies in the virtual environment.
pip install -r scripts/approximations/requirements.txt

# Run the script.
python3 scripts/approximations/main.py

# Run tests
python3 scripts/approximations/rational_test.py
```

# Example Output

## Plotting Functions 

This script plots the results to sanity check that they look right.

On the screenshot below, "50000 terms Equispaced poly" aims to be the most accurate representation of the function $$e^x$$

Visually, the results are as expected where "Chebyshev Rational" is plotted as the most accurate, followed by "Chebyshev Poly" and, lastly, "Equispaced Poly".

image
 
## Plotting Approximation Errors For Each Approximation

image

## Plotting Chosen Approximation Errors

13 parameter Chebyshev Rational errors with coefficient truncations done to match `osmomath.BigDec` precision of 36:

image


scripts/approximations/main.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # start of the interval to calculate the approximation on
+    x_start = 0
+    # end of the interval to calculate the approximation on
+    x_end = 3
+    # number of terms in the polynomial approximation / numerator of the rational approximation.
+    num_terms_approximation = 3
+
+    # number of (x,y) coordinates used to plot the resulting approximation.
+    num_points_plot = 10
```
ValarDragon, line 29: Hrmm, this should be way higher / so dense we don't notice, no?

scripts/approximations/main.py
```diff
@@ -0,0 +1,90 @@
...
+    # number of (x,y) coordinates used to plot the highly accurate
+    # equispaced approximation that we treat as "the closest to truth".
+    num_points_plot_accurate = 50000
```
ValarDragon, line 32: Why wouldn't we just plot the actual function were approximating directly?

scripts/approximations/polynomial.py
```diff
@@ -0,0 +1,41 @@
...
+
+def solve(x: np.ndarray, coeffs: np.ndarray) -> np.ndarray:
+    """ Solves the polynomial. Given a list of x coordinates and a list of coefficients, returns a list of
+    y coordinates, one for each x coordinate. The coefficients must be in ascending order.
+    """
+    o = len(coeffs)
+    y = 0
+    for i in range(o):
+        y += coeffs[i]*x**i
+    return y
```
ValarDragon, line 41: If this is too slow, lmk, we can easily speed this up

scripts/approximations/rational.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+
+def construct_vandermonde_matrix(x_list: list, y_list: list) -> list[list]:
+    """ Constructs a Vandermonde matrix for a rational approximation.
```
ValarDragon, line 4: Pretty sure this is not a vandermonde matrix.

Perhaps just call it rational_poly_eval_matrix

scripts/approximations/rational.py
```diff
@@ -0,0 +1,47 @@
...
+        for j in range(num_terms):
+            # denominator terms
+            if j > 0:
+                row.append(-1 * x_list[i]**j * y_list[i])
+
+        matrix.append(row)
+
+    return matrix
+
+def solve(x: np.ndarray, coefficients_numerator: list, coefficients_denominator: list) -> np.ndarray:
```
ValarDragon, line 31: This should probably be renamed `evaluate`, were not solving the system of equations in this fn

scripts/approximations/polynomial.py
```diff
@@ -0,0 +1,41 @@
...
+    for i in range(num_terms):
+        row = []
+        for j in range(num_terms):
+            row.append(x_list[i]**j)
+        matrix.append(row)
+
+    return matrix
+
+def solve(x: np.ndarray, coeffs: np.ndarray) -> np.ndarray:
+    """ Solves the polynomial. Given a list of x coordinates and a list of coefficients, returns a list of
```
ValarDragon, line 34: This should probably be renamed `evaluate_polynomial`, were not solving the system of equations in this fn

ValarDragon: I'm generally confused. Why don't we just do this as:

image

ValarDragon: (you can change np.dot to just be matrix_inv @ y)

scripts/approximations/rational.py
ValarDragon: EDIT: Same as comment as what I made for polynomial, I don't think this works.

scripts/approximations/polynomial.py
p0mvn: We are solving for the coefficients that are unknown. x and y are both known in this system of equations and we treat them as constants.

scripts/approximations/main.py
p0mvn: I chose a lower number to see the difference in accuracy between each of the approximations. But it could be anything, depending on what is your goal for visualizing it

scripts/approximations/main.py
p0mvn: I think the comment is wrong and requires changes.

This is indeed the actual function.

What we do is spread the X-axis into equal intervals. For each, X we compute the true value of Y.

I incorrectly called this "equispaced" because of the similarity to splitting X-axis into intervals with the equispaced approximation.

scripts/approximations/rational.py
p0mvn: This should be a Vandermonde matrix where the terms follow the geometric progression based on the Remco article. I also compared to the Wikipedia definition, and it seemed to be right

p0mvn:
The graph for errors is still TBD

scripts/approximations/rational.py
p0mvn: You're right! Thanks for clarifying, I will update this

p0mvn: Addressed

scripts/approximations/rational.py
p0mvn: Please see: https://github.com/osmosis-labs/osmosis/pull/3620#discussion_r1041255606

We are not solving but evaluating here. This is needed to compute the y coordinates for the final plot. We do this once the functions are already approximated and we have the coefficients

scripts/approximations/main.py
ValarDragon: I think this should be way higher, we should be able to see the accuracy diff with more points as well, no?

Perhaps lets set to like 10000?

p0mvn:
> The graph for errors is still TBD

In the latest commit [3cf279e](https://github.com/osmosis-labs/osmosis/pull/3620/commits/3cf279eaa19d7c7617a317118ad8e792530dfc07), I added the ability to print the max error for each approximation. The result looks something like this:

```
Max Error on [0, 1]
10 coordinates equally spaced on the X axis
3 polynomial terms and (3, 3) rational terms used for approximations.


Equispaced Poly: 0.014336181311823015
Chebyshev Poly: 0.009865548570433536
Chebyshev Rational: 7.428807818232741e-06
```

Please let me know if this is sufficient with regards to benchmarking errors

scripts/approximations/polynomial.py
p0mvn: Resolving this since seeing an ACK. The function name has been changed.

scripts/approximations/rational.py
p0mvn: Resolving based on discussion in: https://github.com/osmosis-labs/osmosis/pull/3620#discussion_r1041362613

scripts/approximations/main.py
p0mvn: Resolving based on ACK. The comment was corrected

scripts/approximations/approximations.py
```diff
@@ -0,0 +1,80 @@
...
+    The return value is a list of num_terms polynomial coefficients needed to get the returned y coordinates from returned x coordinates.
+    """
+    # Compute Chebyshev coordinates.
+    x_estimated, y_estimated = chebyshev.get_nodes(fn, x_start, x_end, num_terms)
+
+    # Construct a system of linear equations.
+    vandermonde_matrix = polynomial.construct_vandermonde_matrix(x_estimated)
+
+    # Solve the matrix to get the coefficients used in the final approximation polynomial.
+    coef = np.linalg.solve(np.array(vandermonde_matrix), y_estimated)
```
ValarDragon, line 40: Ahh, this is where we solve it

scripts/approximations/main.py
p0mvn: Went even higher with `100000000`:

```
Max Error on [0, 1]
100000000 coordinates equally spaced on the X axis
3 polynomial terms and (3, 3) rational terms used for approximations.


Equispaced Poly: 0.014420666050105169
Chebyshev Poly: 0.009865548570433536
Chebyshev Rational: 7.428807818232741e-06

**Comment by p0mvn** [comment]
Made the suggested change to perform apples to apples comparison by having:


[polynomial_num_terms = numerator_terms + denominator_terms](https://github.com/osmosis-labs/osmosis/pull/3620/commits/f173587ebdb931d0d4b178a4751c4ba6ed1b51e7)

**Comment by p0mvn** [comment]
Latest state:

image

Considering using 13 parameters Chebyshev Rational:

```
Max Error on [0, 1]
10000 coordinates equally spaced on the X axis
13 parameters used


Equispaced Poly: 1.34501742439960E-16
Chebyshev Poly: 8.21380541414887E-18
Chebyshev Rational: 1.41143741935000E-20
```

p0mvn:
13 parameter Chebyshev Rational Errors with coefficient truncations done to match `osmomath` precision of 36:

ValarDragon (approved):
Nice!

ValarDragon:
LGTM, just need an issue to measure truncating the solved coefficients before plotting/computing error. Happy to merge now and track that in an issue

p0mvn:
> LGTM, just need an issue to measure truncating the solved coefficients before plotting/computing error. Happy to merge now and track that in an issue

This should already be achieved here: https://github.com/osmosis-labs/osmosis/blob/e6784a365519112ef9f918434355f4ad0d6b3435/scripts/approximations/main.py#L179-L181

The 3rd plot in the PR description is made from that.

I'm in the process of writing Go code right now. In that PR, I will also hard code the coefficients I'm using in this script and replot the error graph

PR #3634, opened 2022-12-05
Ensure go v1.18 used in make build & make install

ValarDragon:

Go 1.19 had changes that are consensus breaking. Thank you to @jhernandezb for pointing this out. This applies his suggested patch to Makefiles to ensure that all binary builds are done with the same go minor release line.

## Brief Changelog

- Require go 1.18 installed for building the binary.

## Testing and Verifying

Try building the binary if you don't have go 1.18, or build being a success if you do have go 1.18.

I've tested both

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Misc Improvements
 
 * [#3611](https://github.com/osmosis-labs/osmosis/pull/3611) Introduce osmocli, to automate thousands of lines of CLI boilerplate
+* [#3634](https://github.com/osmosis-labs/osmosis/pull/3634) (Makefile) Ensure correct golang version in make build and make install. (Thank you @jhernandezb )
 
 ## v13.0.0
 
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ BUILDDIR ?= $(CURDIR)/build
 E2E_UPGRADE_VERSION := "v13"
 
 
+GO_MAJOR_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f1)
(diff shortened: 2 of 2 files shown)


faddat:
Alternative suggestion:

We should require go 1.19.  This way, there's no chance of those breaking changes cropping up because that's the latest release (no mixing of runtimes)

ValarDragon:
I'm worried about changing to go 1.19 right now so close to release. I think in the future we should do latest minor version, but this one was tested + has binaries out with 1.18

czarcas7ic (approved):
ACK

08d2:
If some code works when compiled with Go 1.18, and breaks when compiled with Go 1.19, this represents a serious problem with that code, which needs to be fixed. Is there a summary or description of the issue?

ValarDragon:
Oh woah, I didn't know about this! Thats really cool

ValarDragon:
> If some code works when compiled with Go 1.18, and breaks when compiled with Go 1.19, this represents a serious problem with that code that absolutely needs to be fixed. Is there a summary or description of the problem somewhere?

So go minor versions have been known to break things at the CGO boundary numerous times often as bug patches. Its been reported that there have been crashes / breakages due to this in at least two places:

* Sort behavior has been changed. (golang sort is unstable, so the unstability altered. This is what halted persistence)
* Gzip encoding edge cases were changed, which breaks hash compatibility of state sync gossip across versions

These are both will little testing in prod thus far

08d2:
> Sort behavior has been changed. (golang sort is unstable, so the unstability altered . . .)

It would be a serious problem if the sort behaviors expressed by the Go stdlib API(s) broke between versions. What do you mean by `golang sort` exactly?

> Gzip encoding edge cases were changed, which breaks hash compatibility of state sync gossip across versions

Er, what? Gzip is neither deterministic nor bijective. The gzip encoding/compression of some data X has arbitrarily many representations X.gz, and arbitrarily many representations of X.gz will decode/decompress to the same data X. If there is any code which assumes otherwise, that code is broken, isn't it?

PR #3635, opened 2022-12-05
fix: don't increment code ID in e2e before running

czarcas7ic:
Closes: #XXX

Roman had to disable the IBC rate limiter test due to an unknown bug. This fixes that bug and re-enables that e2e test.

The reason we incremented this before is because the contract was uploaded in the upgrade handler, and this was a prior workaround to test this within e2e itself. 

## Brief Changelog

- Removed code ID increment prior to contract upload
- Re-enabled IBC test

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -133,10 +133,6 @@ func copyFile(a, b string) error {
 }
 
 func (s *IntegrationTestSuite) TestIBCTokenTransferRateLimiting() {
-
-	// TODO: need to upload the contract in setup for this test to work.
-	s.T().Skip("Skipping IBC rate limiting tests")
-
 	if s.skipIBC {
 		s.T().Skip("Skipping IBC tests")
 	}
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_test.go
```diff
@@ -165,8 +161,7 @@ func (s *IntegrationTestSuite) TestIBCTokenTransferRateLimiting() {
 	fmt.Println(wd, projectDir)
 	err = copyFile(projectDir+"/x/ibc-rate-limit/bytecode/rate_limiter.wasm", wd+"/scripts/rate_limiter.wasm")
 	s.NoError(err)
-	// set LatestCodeId to 1 since we upload a contract in the upgrade handler for v13
-	chainA.LatestCodeId = 1
+
 	node.StoreWasmCode("rate_limiter.wasm", initialization.ValidatorWalletName)
 	chainA.LatestCodeId += 1
```
p0mvn, line 166: Just to make sure it's clear that this is being set to 1. Otherwise it's not immediately clear if we are incrementing something like 2 to 3
```suggestion
	chainA.LatestCodeId = 1
```

p0mvn (approved):

Issue #3594, opened 2022-12-01
x/protorev PR Breakdown


davidterpay:
## Background

The most up to date spec of `x/protorev` can be found [here](https://skip-protocol.notion.site/x-protorev-Spec-3ba5cb368f314120bc514eb2d0a6a0dd). 

As development of the `x/protorev` module is approaching a V1, Skip is ready to start breaking up the implementation into PRs that are modular and straightforward to review. The proposed outline for PRs is as follows:

- [x] 1. Proto-code + App Configuration + Scaffolding (*https://github.com/osmosis-labs/osmosis/pull/3587*)
- [x] 2. Basic Keeper + Epoch Hook (*https://github.com/osmosis-labs/osmosis/pull/3646*)
- [x] 3. Route Building (*https://github.com/osmosis-labs/osmosis/pull/3683*)
- [x] 4. Trade Execution + Statistics (*https://github.com/osmosis-labs/osmosis/pull/3709*)
- [x] 5. Post Handler + Governance Proposals + Developer Profit Sharing (*https://github.com/osmosis-labs/osmosis/pull/3806*)
- [x] 6. (Upgrade PR) Generalizable Route Building (*https://github.com/osmosis-labs/osmosis/pull/4000*)
- [x] 7. Query server + Message server + CLI (*https://github.com/osmosis-labs/osmosis/pull/4214*)
- [x] 8. Optimizations + testing + E2E testing (*https://github.com/osmosis-labs/osmosis/pull/4181*)

### 1. Proto-code + App Configuration + Scaffolding ✅
*Update: This PR has been merged–––> https://github.com/osmosis-labs/osmosis/pull/3587*

This PR includes all of the proto and basic app configuration code. Much of the boilerplate code for the module is also in this PR. Although this includes all of the proto code that will be used end to end, some of it will be implemented in subsequent PRs.

#### High Level Overview of Changes
- Init proto code for the module
- Hook up `x/protorev` to the app
- Genesis code + stateless validation
- Keeper interfaces (dependencies with gamm, epoch, etc.)
- Implementing the Msg types + stateless validation
- Governance proposals handlers + stateless validation
- TokenPairArbRoutes stateless validation (read more about this in the spec linked below)

#### Acceptance Criteria
This change added tests and can be verified as follows:

- *Added unit tests for stateless validation for the genesis state*
- *Added unit tests for stateless validation for Msgs*
- *Added unit tests for stateless validation of governance proposals*
---
[...]

PR #3646, opened 2022-12-07
Protorev Basic Keeper + Epoch Hook

davidterpay:

This PR includes all of the store keys + setters/getters needed to build cyclic arbitrage routes as well as the epoch hook which helps construct the stores. 

## Brief Changelog
- *Keys for keeper*
- *Setters and getters for TokenPairArbRoutes*
- *Setters and getters for Osmo Pools*
- *Setters and getters for Atom Pools*
- *Epoch hook runs after every week to update the highest liquidity pool pairings*

## Testing and Verifying
This PR will also include our testing environment that is configured through `keeper_test.go`. This test suite initializes user accounts for trading, creates over 30 pools, sets up token pair routes, and allows us to test all of the setters and getters for stateful information.

- *Added unit tests for OsmoPools*
- *Added unit tests for AtomPools*
- *Added unit tests for TokenPairArbRoutes*
- *Added unit tests for the Epoch Hook*

#### Keeper Updates
| TokenPairArbRoutes | TokenPairRoutes tracks cyclic arb routes that can be used to create a MultiHopSwap given two denoms | []byte{1} + []byte{inputDenom} +[]byte{outputDenom} | []byte{TokenPairArbRoutes} | KV |
| OsmoPools | List of highest liquidity pools for every token pair with Osmo | []byte{2} + []byte{tokenDenom} | []byte{poolID} | KV |
| AtomPools | List of highest liquidity pools for every token pair with Atom | []byte{3} + []byte{tokenDenom} | []byte{poolID} | KV |

**TokenPairArbRoutes**
Each hot route configured through the admin account must be stored in the keeper. Each Route will be indexed by an input and output denomination. The value will correspond to a TokenPairArbRoutes object which contains a list of lists which correspond to routes that may be taken.

**OsmoPools**
This tracks the highest liquidity pools for every single asset that Osmo is paired with on-chain. This is updated through the epoch hook and on genesis.

**AtomPools**
This tracks the highest liquidity pools for every single asset that Atom is paired with on-chain. This is updated through the epoch hook and on genesis.

**Epoch Hook**
The Epoch hook allows the module to update the highest liquidity pools and distribute developer profits (implemented in a later PR) after every week while incrementally tracking profits through the day epoch identifier. 

### Highest Liquidity Pools
One method of determining cyclic arbitrage opportunities is to use the highest liquidity pools paired with Atom and Osmo. While this calculation is done on genesis, the pools may restructure over time and new tokens may end up being traded heavily with Osmo or Atom. As such, it is necessary to update this over time so that the module’s logic in determining cyclic arbitrage opportunities is most optimal. Using the `AfterEpochEnd` hook in combination with the `week` epoch identifier, we are able to successfully update the pool information every week. At runtime, `UpdatePools` will be executed and all of the pools updated.

### DIFF
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -30,7 +30,8 @@ func (h EpochHooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, ep
 	return nil
 }
 
-// AfterEpochEnd is the epoch end hook.
+// AfterEpochEnd is the epoch end hook. The module will update all of the pools in the store that are
+// used for trading.
 func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
 	switch epochIdentifier {
 	case "week":
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -91,12 +91,12 @@ func (k Keeper) GetHighestLiquidityPools(ctx sdk.Context) (map[string]LiquidityP
 			}
 
 			// Check if there is a match with osmo
-			if otherDenom, match := types.CheckMatchAndReturnOther(tokenA.Denom, tokenB.Denom, types.OsmosisDenomination); match {
+			if otherDenom, match := types.CheckOsmoAtomDenomMatch(tokenA.Denom, tokenB.Denom, types.OsmosisDenomination); match {
 				k.updateHighestLiquidityPool(otherDenom, osmoPools, newPool)
 			}
 
 			// Check if there is a match with atom
-			if otherDenom, match := types.CheckMatchAndReturnOther(tokenA.Denom, tokenB.Denom, types.AtomDenomination); match {
+			if otherDenom, match := types.CheckOsmoAtomDenomMatch(tokenA.Denom, tokenB.Denom, types.AtomDenomination); match {
 				k.updateHighestLiquidityPool(otherDenom, atomPools, newPool)
 			}
 		}
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -45,6 +45,7 @@ func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epoch
 	return nil
 }
 
+// Update pools requests the highest liquidity pools from the gamm module and updates the pools in the store
 func (k Keeper) UpdatePools(ctx sdk.Context) error {
 	// Get the highest liquidity pools
 	osmoPools, atomPools, err := k.GetHighestLiquidityPools(ctx)
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -92,27 +92,26 @@ func (k Keeper) GetHighestLiquidityPools(ctx sdk.Context) (map[string]LiquidityP
 
 			// Check if there is a match with osmo
 			if otherDenom, match := types.CheckMatchAndReturnOther(tokenA.Denom, tokenB.Denom, types.OsmosisDenomination); match {
-				if currPool, ok := osmoPools[otherDenom]; !ok {
-					osmoPools[otherDenom] = newPool
-				} else {
-					if newPool.Liquidity.GT(currPool.Liquidity) {
-						osmoPools[otherDenom] = newPool
-					}
-				}
+				k.updateHighestLiquidityPool(otherDenom, osmoPools, newPool)
 			}
 
 			// Check if there is a match with atom
 			if otherDenom, match := types.CheckMatchAndReturnOther(tokenA.Denom, tokenB.Denom, types.AtomDenomination); match {
-				if currPool, ok := atomPools[otherDenom]; !ok {
-					atomPools[otherDenom] = newPool
-				} else {
-					if newPool.Liquidity.GT(currPool.Liquidity) {
-						atomPools[otherDenom] = newPool
-					}
-				}
+				k.updateHighestLiquidityPool(otherDenom, atomPools, newPool)
 			}
 		}
 	}
 
 	return osmoPools, atomPools, nil
+}
+
+// updateHighestLiquidityPool updates the pool with the highest liquidity for either osmo or atom
+func (k Keeper) updateHighestLiquidityPool(denom string, pool map[string]LiquidityPoolStruct, newPool LiquidityPoolStruct) {
+	if currPool, ok := pool[denom]; !ok {
+		pool[denom] = newPool
+	} else {
+		if newPool.Liquidity.GT(currPool.Liquidity) {
+			pool[denom] = newPool
+		}
+	}
 }
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -35,10 +35,6 @@ func (h EpochHooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, ep
 func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
 	switch epochIdentifier {
 	case "week":
-		// Reset the pools in the store
-		h.k.DeleteAllAtomPools(ctx)
-		h.k.DeleteAllOsmoPools(ctx)
-
 		// Update the pools in the store
 		return h.k.UpdatePools(ctx)
 	}
@@ -47,6 +43,10 @@ func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epoch
 
 // Update pools requests the highest liquidity pools from the gamm module and updates the pools in the store
 func (k Keeper) UpdatePools(ctx sdk.Context) error {
+	// Reset the pools in the store
+	k.DeleteAllAtomPools(ctx)
+	k.DeleteAllOsmoPools(ctx)
+
 	// Get the highest liquidity pools
 	osmoPools, atomPools, err := k.GetHighestLiquidityPools(ctx)
 	if err != nil {
--- a/x/protorev/keeper/epoch_hook.go
+++ b/x/protorev/keeper/epoch_hook.go
@@ -1,0 +1,116 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	epochstypes "github.com/osmosis-labs/osmosis/v13/x/epochs/types"
+	"github.com/osmosis-labs/osmosis/v13/x/protorev/types"
+)
+
+type EpochHooks struct {
+	k Keeper
+}
+
+// Struct used to track the pool with the highest liquidity
+type LiquidityPoolStruct struct {
+	Liquidity sdk.Int
+	PoolId    uint64
+}
+
+var (
+	_ epochstypes.EpochHooks = EpochHooks{}
+)
+
+func (k Keeper) EpochHooks() epochstypes.EpochHooks {
+	return EpochHooks{k}
+}
+
+// BeforeEpochStart is the epoch start hook.
+func (h EpochHooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	return nil
+}
+
+// AfterEpochEnd is the epoch end hook.
+func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	switch epochIdentifier {
+	case "week":
+		// Reset the pools in the store
+		h.k.DeleteAllAtomPools(ctx)
+		h.k.DeleteAllOsmoPools(ctx)
+
+		// Update the pools in the store
+		return h.k.UpdatePools(ctx)
+	}
+	return nil
+}
+
+func (k Keeper) UpdatePools(ctx sdk.Context) error {
+	// Get the highest liquidity pools
+	osmoPools, atomPools, err := k.GetHighestLiquidityPools(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Update the pools in the store
+	for token, poolInfo := range osmoPools {
+		k.SetOsmoPool(ctx, token, poolInfo.PoolId)
+	}
+	for token, poolInfo := range atomPools {
+		k.SetAtomPool(ctx, token, poolInfo.PoolId)
+	}
+
+	return nil
+}
+
+// GetHighestLiquidityPools returns the highest liquidity pools for pools that have Osmo or Atom
+// and Osmo/Atom
+func (k Keeper) GetHighestLiquidityPools(ctx sdk.Context) (map[string]LiquidityPoolStruct, map[string]LiquidityPoolStruct, error) {
+	// Get all pools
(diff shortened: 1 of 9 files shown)


x/protorev/keeper/epoch_hook.go
```diff
@@ -0,0 +1,117 @@
...
+
+// BeforeEpochStart is the epoch start hook.
+func (h EpochHooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	return nil
+}
+
+// AfterEpochEnd is the epoch end hook. The module will update all of the pools in the store that are
+// used for trading.
+func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	switch epochIdentifier {
```
davidterpay, line 36: This switch statement will get populated with more info as PRs are submitted.

x/protorev/keeper/epoch_hook.go
```diff
@@ -0,0 +1,117 @@
...
+				if currPool, ok := atomPools[otherDenom]; !ok {
+					atomPools[otherDenom] = newPool
+				} else {
+					if newPool.Liquidity.GT(currPool.Liquidity) {
+						atomPools[otherDenom] = newPool
+					}
+				}
+			}
+		}
+	}
```
stackman27, line 114: I think we can make this function shorter by just adding an extra function like `checkOsmoAtomDenomMatch`

x/protorev/keeper/epoch_hook.go
```diff
@@ -0,0 +1,117 @@
...
+		h.k.DeleteAllAtomPools(ctx)
+		h.k.DeleteAllOsmoPools(ctx)
+
+		// Update the pools in the store
+		return h.k.UpdatePools(ctx)
+	}
+	return nil
+}
+
+func (k Keeper) UpdatePools(ctx sdk.Context) error {
```
stackman27, line 48: Asking to learn: When we fix arb and perform swaps who'll be paying for swap fees? Would it be beneficial to route swap between high fee vs low fee pools?

x/protorev/keeper/protorev.go
```diff
@@ -0,0 +1,144 @@
...
+
+// SetAtomPool sets the pool id of the Atom pool for the given denom paired with Atom
+func (k Keeper) SetAtomPool(ctx sdk.Context, denom string, poolId uint64) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixAtomPools)
+	key := types.GetKeyPrefixAtomPool(denom)
+
+	store.Set(key, sdk.Uint64ToBigEndian(poolId))
+}
+
+// DeleteAllAtomPools deletes all the Atom pools
```
stackman27, line 135: nit: feel free to change this to anything else
```suggestion
// DeleteAllAtomPools deletes all the Atom pools from modules store for liquidity update
```

x/protorev/keeper/epoch_hook_test.go
```diff
@@ -0,0 +1,74 @@
...
+		// Module currently limited to two asset pools
+		// Instantiate asset and amounts for the pool
+		if len(pool.PoolAssets) == 2 {
+			pool.Asset1 = pool.PoolAssets[0].Token.Denom
+			pool.Amount1 = pool.PoolAssets[0].Token.Amount
+			pool.Asset2 = pool.PoolAssets[1].Token.Denom
+			pool.Amount2 = pool.PoolAssets[1].Token.Amount
+		}
+
+		if pool.Asset1 == types.OsmosisDenomination || pool.Asset2 == types.OsmosisDenomination || pool.Asset1 == types.AtomDenomination || pool.Asset2 == types.AtomDenomination {
```
stackman27, line 29: Shouldn't this be `Asset1 == osmo || Asset1 == atom && Asset2 == osmo || Asset2 == atom` I thought the `tokenInDenom` **and** `tokenOutDenom` had to be either osmo or atom

x/protorev/keeper/epoch_hook_test.go
```diff
@@ -0,0 +1,74 @@
...
+			// create a key that is a combination of asset1 and asset2 in alphabetical order
```
stackman27, line 30: Note: We started sorting denoms lexicographically in twap so might wanna see if it's relevant here

x/protorev/keeper/epoch_hook_test.go
```diff
@@ -0,0 +1,74 @@
...
+			key := fmt.Sprintf("%s-%s", pool.Asset1, pool.Asset2)
+			if pool.Asset1 > pool.Asset2 {
+				key = fmt.Sprintf("%s-%s", pool.Asset2, pool.Asset1)
+			}
+
+			if storedPool, ok := expectedToSee[key]; !ok {
+				expectedToSee[key] = pool
+			} else {
+				liquidity := pool.Amount1.Mul(pool.Amount2)
```
stackman27, line 39: Asking to learn: Why is liquidity Multiplied between two pools? Shouldn't we be adding?

stackman27 (commented):
Left few comments still reviewing `keeper_test.go`. It'd be really nice if we can break `keeper_test.go` into smaller files, maybe one for setup and one for test :taco:

x/protorev/keeper/epoch_hook.go
davidterpay: The module account will be minting the coins and paying for the swap fees. If swap fees end up consuming the entire arb, the module won't execute that trade (we simulate the trade before executing it). This will be more hashed out in PR #3 and #4 (https://github.com/osmosis-labs/osmosis/issues/3594). 

Building routes with low fee vs high fee pools is an interesting point. I think @NotJeremyLiu might have some more insight into that.

NotJeremyLiu: Module account pays for the swap (module account just calls MultiHopSwapExactAmountIn like a normal account would).

Determining the optimal input amount accounts for swap fees.

If liquidity, weight, and assets were equal on 2 pools, routing through lower fee pools would be beneficial.

x/protorev/keeper/epoch_hook_test.go
davidterpay: tldr the pools that we instantiate in `keeper_test.go` don't always contain either atom or osmo because we can sandwich a pool that has neither atom or osmo if we pair it with atom or osmo on either side. i.e. if we have pool juno/akash, we can make a route osmo/juno -> juno/akash -> akash/osmo (replace osmo with atom and it is also valid).

`TestEpochHook` is checking the pools that were created in`keeper_test.go` and seeing if any of the assets in the pool are atom or osmo. Once we have all the pools that contain atom, osmo or both, we get the pools that have the highest liquidity for every unique pair. The pools that end up being filtered here should be the same ones that the epoch sets in state.

davidterpay: I'll add more documentation on the tests so that its easier to read since `keeper_test.go` is a heavy read (will also look into breaking it up).

x/protorev/keeper/epoch_hook_test.go
davidterpay: The module assumes the constant product market maker model (x * y = k) where x and y are the pool reserves and k is can be seen as total liquidity in the pool. You can read more about it [here](https://github.com/runtimeverification/verified-smart-contracts/blob/uniswap/uniswap/x-y-k.pdf)

x/protorev/keeper/epoch_hook_test.go
davidterpay: pools are a custom struct that we made in `keeper_test.go` so we can't assume that they are sorted. However, we will keep this in mind for subsequent PRs (if we sort by pool asset anywhere, we will just assume its already done for us).

x/protorev/keeper/epoch_hook_test.go
stackman27: aah yes the cfmm equation!

x/protorev/keeper/epoch_hook.go
davidterpay: Are you talking about the function name or what it returns? We need some functionality to check whether any of the assets in a 2 asset pool are atom or osmo and then return the other denom in that pool that is neither atom nor osmo (used here and in `epoch_hook.go`).

davidterpay:
@stackman27 brings up a good point about `keeper_test.go` and I am curious to hear what others think. While initializing all of the pools, configs, and etc. makes it cleaner to write test case (less set up code through out the unit tests) because testing environment is unified, it makes it a bit more difficult to reason about when initially reviewing or looking at the code.

x/protorev/types/keys.go
```diff
@@ -10,3 +10,36 @@ const (
 	// RouterKey defines the module's message routing key
 	RouterKey = ModuleName
 )
+
+const (
+	prefixTokenPairRoutes = iota + 1
```
bpiv400, line 15: Why do we set this one to iota + 1 and don't set the others? 

Is this standard prefix naming convention?

x/protorev/keeper/epoch_hook.go
bpiv400: In the current form, it's just recomputing and storing the highest liquidity pool for each token x {osmo, atom}. 

This seems like most of the logic that would happen on an epoch to me. @davidterpay what's missing?

x/protorev/keeper/epoch_hook.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil
+}
+
+// AfterEpochEnd is the epoch end hook. The module will update all of the pools in the store that are
+// used for trading.
+func (h EpochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	switch epochIdentifier {
+	case "week":
+		// Reset the pools in the store
+		h.k.DeleteAllAtomPools(ctx)
```
bpiv400, line 39: Why have separate functions for delete and update? Would we ever call delete without updating or vice versa?

x/protorev/keeper/protorev.go
```diff
@@ -0,0 +1,144 @@
...
+// SetOsmoPool sets the pool id of the Osmo pool for the given denom paired with Osmo
+func (k Keeper) SetOsmoPool(ctx sdk.Context, denom string, poolId uint64) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixOsmoPools)
+	key := types.GetKeyPrefixOsmoPool(denom)
+
+	store.Set(key, sdk.Uint64ToBigEndian(poolId))
+}
+
+// DeleteAllOsmoPools deletes all the Osmo pools from modules store
+func (k Keeper) DeleteAllOsmoPools(ctx sdk.Context) {
```
bpiv400, line 103: Nit: DeleteAllOsmoPools and DeleteAllAtomPools are identical apart from switching a key. Would it make more sense to pass that key in as an additional arg? Or, if we prefer the readability of this approach, have a helper function that both call which takes in the key and does all of the logic. 

We're re-writing the code in both places.

x/protorev/keeper/protorev.go
```diff
@@ -0,0 +1,144 @@
...
+	iterator := sdk.KVStorePrefixIterator(store, types.KeyPrefixOsmoPools)
+
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		store.Delete(iterator.Key())
+	}
+}
+
+// GetAtomPool returns the pool id of the Atom pool for the given denom paired with Atom
+func (k Keeper) GetAtomPool(ctx sdk.Context, denom string) (uint64, error) {
```
bpiv400, line 114: Nit: Same comment about redundancy as above

x/protorev/keeper/protorev.go
```diff
@@ -0,0 +1,144 @@
...
+	if len(bz) == 0 {
+		return 0, fmt.Errorf("no atom pool for denom %s", denom)
+	}
+
+	poolId := sdk.BigEndianToUint64(bz)
+	return poolId, nil
+}
+
+// SetAtomPool sets the pool id of the Atom pool for the given denom paired with Atom
+func (k Keeper) SetAtomPool(ctx sdk.Context, denom string, poolId uint64) {
```
bpiv400, line 128: Nit: Same comment about redundancy as above

bpiv400 (approved):
Just a few nits. Well done chef

x/protorev/types/keys.go
davidterpay: Standard prefix naming convention, it increments it by one for every value that you add after.

x/protorev/keeper/epoch_hook.go
davidterpay: I was thinking that we reset all of the pools and have a clean slate when setting the pools in the `UpdatePools` function. I wasn't fully sure if pools ever get removed for whatever reason but it seems like thats an edge case I don't need to account for (so I will remove this deletion code and have it only update).

davidterpay: Per the comment above this one, the epoch hook will also be populated with functionality that distributes a portion of the developer fees to a developer account (which presumably will be skip's address once deployed). This requires us to track various info such as the number of days since genesis so that we can properly split the arbitrage revenue generated at any given moment in time.

x/protorev/keeper/protorev.go
davidterpay: I agree, I can have a wrapper function that both `DeleteAllOsmoPools` and `DeleteAllAtomPools` call. I do think having the function signatures remain as is makes the most sense though. However, I can update it so that the code is not duplicated and a key is just passed in.

x/protorev/keeper/protorev.go
davidterpay: going to refactor as per comment above

x/protorev/keeper/protorev.go
davidterpay: same comment as above.

x/protorev/keeper/epoch_hook.go
davidterpay: Given that we wouldn't be able to see a swap on a deprecated pool, the route building logic we have can potentially build a route with a stale/deprecated pool which is why we want to be able to delete all of the deprecated pools from the store before resetting it in state. 

In terms of the additional functionality for the epoch hook, it will also be populated with functionality that distributes a portion of the developer fees to a developer account (which presumably will be skip's address once deployed). This requires us to track various info such as the number of days since genesis so that we can properly split the arbitrage revenue generated at any given moment in time.

x/protorev/keeper/protorev.go
davidterpay: I think for getters and setters it makes sense to not wrap it around a single set/get since even logic like error message would have to be passed through the wrapper function

x/protorev/keeper/epoch_hook.go
stackman27: I was suggesting `checkOsmoAtomDenomMatch` as function name, but we can use a helper function here for logic and reduce code length

x/protorev/keeper/epoch_hook_test.go
stackman27: sounds good

x/protorev/keeper/epoch_hook.go
davidterpay: Oh I see. I agree, I'll make change.

Issue #3641, opened 2022-12-06
osmocli epic:


ValarDragon:
OsmoCli (#3611) roadmap of things to tackle to improve testing UX:
- [x] Add the ability to register a custom parser for a given field. (Lets us eliminate the rest of existing boilerplate / non-converted funcs)
- [x] Add some go tests within osmo-cli
- [x] Change the syntax from cmd.AddCommands to instead be osmocli.AddTx[M sdk.Msg](&TxDesc, msg M)
  - [x] Need to figure out how to register custom flagsets in this syntax 
  - [x] Lets us make a second method, for getting a proto-parser from desc, and then we can kill all the current slow CLI tests
- [ ] README
- [x] Split this into its own go.mod, so others can import

RusAkh:
hello @ValarDragon. I would like to collaborate here, is it ok with you?

PR #3647, opened 2022-12-07
Osmocli: Fix lines of code overhead + speed of CLI tests

ValarDragon:
Cref: #3641 

Makes writing CLI tests very fast, and few lines of code, testing exactly what you want to test

Txs currently done and queries done. Propogated to most of the CLI.

* I migrated everything in incentives, except the test for CmdRewardsEst, which if you investigate didn't actually test anything.
* Migrated everything in gamm. Removed a few of the 0 argument queries, since not really anything meaningful to test
* Deleted tests in pool-incentives, because they were never ran before.

Net `go test time` down from 4minutes to 3minutes 30 seconds. (I am confused though, because if you add up the time thats no longer in CLI tests from gamm, incentives, lockup, thats 76 seconds. So perhaps theres some degree of parallelism in tests somewhere? But also confusingly, sum of the remaining module test times appear to me to be roughly around 2 minutes, not sure how its still saying its 3.5 minutes. My guess would be theres something with it not counting time executed from cgo/cosmwasm contracts?)

## Brief Changelog

- Makes CLI tests effective and succinct

## Testing and Verifying

This change adds tests for all CLI commands it changes, that can be easily verified!
Note, I do also delete a few of the old cli_tests in lockup. We tested parsing of a larger subset of the messages before, but due to the standardization now applied, I only wrote one test per "class" of message we parse.

  - How is the feature or change documented? N/A for now, perhaps we should add some docs in Readme as later step in epic


RusAkh:
👍

RusAkh:
Here are PRs related to this PR:
`mint` module tests migration: https://github.com/osmosis-labs/osmosis/pull/3662
`epochs` module tests migration: https://github.com/osmosis-labs/osmosis/pull/3661

ValarDragon:
Per call, were good with merging this, as its just CLI affecting. (And unblocks other things)

PR #3648, opened 2022-12-07
feat: add tick spacing

czarcas7ic:
Closes: #XXX

Adds tick spacing parameter to creating new pools as well as logic to enforce tick spacing requirements to positions created.

This is getting merged into `adam/msgCreateConcentratedPool` since concentrated-liquidity-main is in code freeze at the moment.


x/concentrated-liquidity/model/msgs_test.go
```diff
@@ -26,42 +26,67 @@ func TestMsgCreateConcentratedPool(t *testing.T) {
...
+			},
+			expectPass: false,
+		},
+		{
+			name: "tick spacing less than minimum",
+			msg: clmodel.MsgCreateConcentratedPool{
+				Sender:      addr1,
+				Denom0:      "eth",
+				Denom1:      "usdc",
+				TickSpacing: 0,
```
p0mvn, line 79: Let's define these minimums and maximums as constants in the production code and import them here please. So that if they get updated, we don't need to update the tests.

Also, where is the current choice of min and max documented? How was it chosen?

x/concentrated-liquidity/model/msgs.go
```diff
@@ -23,22 +23,29 @@ func NewMsgCreateConcentratedPool(
...
-func (msg MsgCreateConcentratedPool) Route() string { return cltypes.RouterKey }
+func (msg MsgCreateConcentratedPool) Route() string { return types.RouterKey }
 func (msg MsgCreateConcentratedPool) Type() string  { return TypeMsgCreateConcentratedPool }
 func (msg MsgCreateConcentratedPool) ValidateBasic() error {
 	_, err := sdk.AccAddressFromBech32(msg.Sender)
 	if err != nil {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid sender address (%s)", err)
 	}
 
+	// TODO: Determine what we want the min and max tick range to be.
```
p0mvn, line 44: ahh I see, that's TBD

x/concentrated-liquidity/model/msgs.go
```diff
@@ -23,22 +23,29 @@ func NewMsgCreateConcentratedPool(
...
+	if !(msg.TickSpacing >= types.TickSpacingMinimum) || !(msg.TickSpacing <= types.TickSpacingMaximum) {
```
p0mvn, line 45: Let's use these `types.TickSpacingMinimum` in tests please

x/concentrated-liquidity/tick.go
```diff
@@ -81,15 +81,23 @@ func (k Keeper) SetTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64, tic
 // Also, lower tick must be less than upper tick.
 // Returns error if validation fails. Otherwise, nil.
 // TODO: test
-func validateTickRangeIsValid(lowerTick int64, upperTick int64) error {
-	// ensure types.MinTick <= lowerTick < types.MaxTick
+func validateTickRangeIsValid(tickSpacing uint64, lowerTick int64, upperTick int64) error {
```
p0mvn, line 84: Let's add tests to this function, please. If none exist, let's make an issue to test it

x/concentrated-liquidity/tick.go
```diff
@@ -81,15 +81,23 @@ func (k Keeper) SetTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64, tic
...
+	// Check if the lower and upper tick values are divisible by the tick spacing.
+	if lowerTick%int64(tickSpacing) != 0 || upperTick%int64(tickSpacing) != 0 {
```
p0mvn, line 86: Just wondering - what if tick spacing is say 3 or 6, will users have to think about multiples of these numbers when LPing? Is there a way to avoid this UX impediment?

x/concentrated-liquidity/lp_test.go
```diff
@@ -82,6 +84,17 @@ func (s *KeeperTestSuite) TestCreatePosition() {
...
 			expectedError:     errors.New("liquidityDelta calculated equals zero"),
 		},
+		"create a position with non default tick spacing (10) with ticks that fall into tick spacing requirements": {
+			lowerTick:       int64(84220),
+			upperTick:       int64(86130),
+			amount0Expected: sdk.NewInt(997568),
+			liquidityAmount: sdk.MustNewDecFromStr("1514719247.706987476085061790"),
+			tickSpacing:     10,
+		},
+		"error: attempt to create a position that does not fall into tick spacing requirement of that pool": {
```
p0mvn, line 94: Can we add a comment that default lower tick and default upper tick are not divisible by the tick spacing of 10. Wasn't immediately apparent to me why this would error.

p0mvn (commented):
Nice work.

2 questions:
- How does tick spacing affect swaps? I'm not seeing any changes to that. I would assume that  tick spacing helps us avoid searching over every tick in swaps. For example, when we look for the next initialized tick.
- Can this be made against `concentrated-liquidity-main`. This PR seems to be independent of `MsgCreateConcentratedPool`, and we shouldn't block it on swaprouter merge

x/concentrated-liquidity/tick.go
czarcas7ic: The solution is to make specific ticks (1,10,30,60,200) as uni does rather than a range

czarcas7ic:
Closing in favor of new PR

PR #3649, opened 2022-12-07
osmocli: add float parser, add test for ParseFieldFromArg

RusAkh:
## Changes
Test for `ParseFieldFromArg`
Added float parser
@ValarDragon

### DIFF
--- a/osmoutils/osmocli/parsers.go
+++ b/osmoutils/osmocli/parsers.go
@@ -191,6 +191,14 @@ func ParseFieldFromArg(fVal reflect.Value, fType reflect.StructField, arg string
 		}
 		fVal.SetInt(i)
 		return nil
+	case reflect.Float32, reflect.Float64:
+		typeStr := fType.Type.String()
+		f, err := ParseFloat(arg, typeStr)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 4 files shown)


osmoutils/osmocli/parsers.go
```diff
@@ -241,6 +249,14 @@ func ParseUint(arg string, fieldName string) (uint64, error) {
 	return v, nil
 }
 
+func ParseFloat(arg string, fieldName string) (float64, error) {
+	v, err := strconv.ParseFloat(arg, 64)
+	if err != nil {
+		return 0, fmt.Errorf("could not parse %s as uint for field %s: %w", arg, fieldName, err)
```
ValarDragon, line 255:
```suggestion
		return 0, fmt.Errorf("could not parse %s as float for field %s: %w", arg, fieldName, err)
```

ValarDragon:
Oh yikes, looks like this got closed because it was from a fork, when I merged the base PR in. If you want to make this as a branch from the osmosis repo, I can handle rebasing it to get it merged.

(Or if you'd like to, take this branch, and do `rebase -i main`, after you've merged in main. Then drop all commits from the base branch)

PR #3656, opened 2022-12-08
osmocli: add float parser, add test for `ParseFieldFromArg`

RusAkh:
- add float parser, add test for ParseFieldFromArg

Adds float parser to osmocli + test


ValarDragon (approved):
Nice!

PR #3662, opened 2022-12-08
mint: change tests 

RusAkh:
`x/mint` tests run using `osmocli`


faddat:
dude this is great!

@ValarDragon is it okay to merge into your branch?

RusAkh:
I think it is okay as we discussed that already 👍. Shall I merge?

ValarDragon:
I don't know that we can merge this easily unfortunately -- mint wasn't adjusted in the initial PR because mint's differences from everything else are inherited from the SDK.

So this is actually mint-CLI breaking

ValarDragon:
Hrmm, I think we should make a github issue for breaking mint CLI.

I want to do this, we just need to get changelog set, release in a minor version and communicate to integrators.

PR #3671, opened 2022-12-08
feat[CL]: tick spacing

czarcas7ic:
Closes: #XXX

As discussed offline, we decided to implement tick spacing directly into concentrated-liquiditiy-main so we are not blocked on msgs being completed.

This PR implements AuthorizedTickSpacing as a param. Once the CreateCLPool msg is created in a subsequent PR, the validate basic for that message will be checked there.

Also of note: we dont have to implement any extra logic for NextInitializedTick when implementing tick spacing, because we only cycle through initialized ticks. Because only ticks divisible by the tick spacing can be initialized, those ticks will never be initialized and therefore will never need to be cycled through!


x/concentrated-liquidity/genesis.go
```diff
@@ -6,11 +6,14 @@ import (
 	types "github.com/osmosis-labs/osmosis/v13/x/concentrated-liquidity/types"
 )
 
+// InitGenesis initializes the concentrated-liquidity module with default parameters
 func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
-
+	k.SetParams(ctx, types.DefaultParams())
```
p0mvn, line 11: Let's test this please

x/concentrated-liquidity/genesis.go
```diff
@@ -6,11 +6,14 @@ import (
...
 }
 
-// ExportGenesis returns the capability module's exported genesis.
+// ExportGenesis returns the concentrated-liquidity module's exported genesis.
 func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
-	return &types.GenesisState{}
+	return &types.GenesisState{
+		Params: k.GetParams(ctx),
```
p0mvn, line 17: and this

x/concentrated-liquidity/internal/model/pool.go
```diff
@@ -15,20 +15,24 @@ var (
 	_ types.ConcentratedPoolExtension = &Pool{}
 )
 
-func NewConcentratedLiquidityPool(poolId uint64, denom0, denom1 string, currSqrtPrice sdk.Dec, currTick sdk.Int) (Pool, error) {
+func NewConcentratedLiquidityPool(poolId uint64, denom0, denom1 string, currSqrtPrice sdk.Dec, tickSpacing uint64) (Pool, error) {
+	// Order the initial pool denoms so that denom0 is lexicographically smaller than denom1.
```
p0mvn, line 19: Do we have a test for this function? Let's make an issue if not please

x/concentrated-liquidity/internal/model/pool.go
```diff
@@ -15,20 +15,24 @@ var (
...
+
+	// Create a new pool struct with the specified parameters
 	pool := Pool{
 		// TODO: move gammtypes.NewPoolAddress(poolId) to swaproutertypes
 		Address:          gammtypes.NewPoolAddress(poolId).String(),
 		Id:               poolId,
-		CurrentSqrtPrice: currSqrtPrice,
-		CurrentTick:      currTick,
+		CurrentSqrtPrice: sdk.ZeroDec(),
+		CurrentTick:      sdk.ZeroInt(),
```
p0mvn, line 31: Does it make sense t initialize the current tick here? Zero is a valid value for a tick. To signify, that it is uninitialized, maybe we should keep it as nil int. Wdyt?

x/concentrated-liquidity/keeper_test.go
```diff
@@ -39,6 +41,13 @@ func (suite *KeeperTestSuite) SetupTest() {
 	suite.Setup()
 }
 
+// PrepareDefaultPool sets up an eth usdc concentrated liquidity pool with pool ID 1, tick spacing of 1, and no liquidity
+func (s *KeeperTestSuite) PrepareDefaultPool(ctx sdk.Context) types.ConcentratedPoolExtension {
```
p0mvn, line 45: nice

x/gamm/keeper/swap_test.go
```diff
@@ -184,7 +184,7 @@ func (suite *KeeperTestSuite) TestCalcOutAmtGivenIn() {
 				suite.NoError(err)
 				pool = poolExt.(swaproutertypes.PoolI)
 			} else if test.param.poolType == "concentratedliquidity" {
-				poolExt, err := suite.App.ConcentratedLiquidityKeeper.CreateNewConcentratedLiquidityPool(ctx, 1, "bar", "foo", sdk.MustNewDecFromStr("70.710678118654752440"), sdk.NewInt(85176))
+				poolExt, err := suite.App.ConcentratedLiquidityKeeper.CreateNewConcentratedLiquidityPool(ctx, 1, "bar", "foo", 1)
```
p0mvn, line 187: Can we use the new `PrepareDefaultPool` helper here?

x/concentrated-liquidity/types/params.go
```diff
@@ -0,0 +1,70 @@
...
+	ticks, ok := i.([]string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// Iterate over the slice of strings.
+	// For each string, attempt to convert it to an unsigned 64-bit integer.
+	// If the conversion fails, return an error.
+	for _, tick := range ticks {
+		_, err := strconv.ParseUint(tick, 10, 64)
```
p0mvn, line 63: Can we make these as `uint64` in the first place? Do we have to store these as strings?

x/concentrated-liquidity/types/params.go
```diff
@@ -0,0 +1,70 @@
...
+func NewParams(authorizedTickSpacing []string) Params {
+	return Params{
+		AuthorizedTickSpacing: authorizedTickSpacing,
+	}
+}
+
+// DefaultParams returns default concentrated-liquidity module parameters.
+func DefaultParams() Params {
+	return Params{
+		AuthorizedTickSpacing: []string{"1", "10", "60", "200"},
```
p0mvn, line 31: How were these values decided on? Should we have an issue tracking revisiting these?

x/concentrated-liquidity/swaps.go
```diff
@@ -23,28 +23,36 @@ type SwapState struct {
...
+
+	// Create a new concentrated liquidity pool with the given parameters.
 	pool := &model.Pool{
 		// TODO: move gammtypes.NewPoolAddress(poolId) to swaproutertypes
 		Address:          gammtypes.NewPoolAddress(poolId).String(),
 		Id:               poolId,
-		CurrentSqrtPrice: currSqrtPrice,
-		CurrentTick:      currTick,
+		CurrentSqrtPrice: sdk.ZeroDec(),
+		CurrentTick:      sdk.ZeroInt(),
```
p0mvn, line 48: ditto for deciding whether initializing `CurrentTick` to 0 makes sense

x/concentrated-liquidity/swaps.go
```diff
@@ -23,28 +23,36 @@ type SwapState struct {
...
 	}
+
+	// Create a new concentrated liquidity pool with the given parameters.
 	pool := &model.Pool{
 		// TODO: move gammtypes.NewPoolAddress(poolId) to swaproutertypes
 		Address:          gammtypes.NewPoolAddress(poolId).String(),
 		Id:               poolId,
-		CurrentSqrtPrice: currSqrtPrice,
-		CurrentTick:      currTick,
+		CurrentSqrtPrice: sdk.ZeroDec(),
```
p0mvn, line 47: We should also be using `NewConcentratedLiquidityPool` constructor here as opposed initializing the pool directly

czarcas7ic: We cant do this until we move the pool model outside of the internal package. This was supposed to be done in MsgCreatePool PR but I can do that now if you think I should

x/concentrated-liquidity/internal/model/pool.go
czarcas7ic: Unit test created

x/concentrated-liquidity/internal/model/pool.go
czarcas7ic: Discussed this offline, going to keep this at zero for now, since sdk numbers when marshaled will always default to zero if nil.

x/gamm/keeper/swap_test.go
czarcas7ic: Are you able to import functions from one test package into another?

x/concentrated-liquidity/types/params.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/3684

x/concentrated-liquidity/lp.go
```diff
@@ -184,3 +207,41 @@ func (k Keeper) sendCoinsBetweenPoolAndUser(ctx sdk.Context, denom0, denom1 stri
...
+func (k Keeper) isInitialPosition(initialSqrtPrice sdk.Dec, initialTick sdk.Int) bool {
+	if initialSqrtPrice.Equal(sdk.ZeroDec()) && initialTick.Equal(sdk.ZeroInt()) {
+		return true
+	}
+	return false
+}
+
+// createInitialPosition ensures that the first position created on this pool includes both asset0 and asset1
+// This is required so we can set the pool's sqrtPrice and calculate it's initial tick from this
+func (k Keeper) createInitialPosition(ctx sdk.Context, pool types.ConcentratedPoolExtension, amount0Desired, amount1Desired sdk.Int) error {
```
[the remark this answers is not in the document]
czarcas7ic: Added

x/concentrated-liquidity/lp_test.go
```diff
@@ -75,6 +77,17 @@ func (s *KeeperTestSuite) TestCreatePosition() {
...
 			expectedError:  errors.New("liquidityDelta calculated equals zero"),
 		},
+		"create a position with non default tick spacing (10) with ticks that fall into tick spacing requirements": {
+			lowerTick:       int64(84220),
+			upperTick:       int64(86130),
+			amount0Expected: sdk.NewInt(997568),
+			liquidityAmount: sdk.MustNewDecFromStr("1514719247.706987476085061790"),
+			tickSpacing:     10,
+		},
+		"error: attempt to use and upper and lower tick that are not divisible by tick spacing": {
```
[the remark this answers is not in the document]
czarcas7ic: Actually, I suppose it wouldn't hurt to include that case in the CreatePosition test as well, will add it

x/gamm/keeper/swap_test.go
p0mvn: WDYT about re-defining this helper on the `apptesting` suite and naming it `PrepareConcentratedPool`. I think it would also be helpful for swaprouter tests

x/concentrated-liquidity/swaps.go
czarcas7ic: Actually disregard, I was able to do this (thought there was going to be a circular dep but there wasn't)

x/concentrated-liquidity/internal/model/pool.go
```diff
@@ -81,8 +87,14 @@ func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom
 		return p.CurrentSqrtPrice.Power(2), nil
 	} else if p.Token1 == baseAssetDenom && p.Token0 == quoteAssetDenom { // if not, we calculate the reverse spot price by 1 / currentSqrtPrice^2
 		return sdk.NewDec(1).Quo(p.CurrentSqrtPrice.Power(2)), nil
+	} else if p.Token0 != baseAssetDenom && p.Token1 != baseAssetDenom {
```
p0mvn, line 90: Consider simplifying this to:

```go
func (p Pool) SpotPrice(ctx sdk.Context, baseAssetDenom string, quoteAssetDenom string) (sdk.Dec, error) {
        // validate base asset is in pool
       if baseAssetDenom != p.Token0 && baseAssetDenom != p.Token1 ) {
		return sdk.Dec{}, fmt.Errorf("base asset denom (%s) is not in pool with (%s, %s) pair, baseAssetDenom, p.Token0, p.Token1)
	}
        // validate quote asset is in pool
        if quoteAssetDenom != p.Token0 && quoteAssetDenom != p.Token1 ) {
		return sdk.Dec{}, fmt.Errorf("quote asset denom (%s) is not in pool with (%s, %s) pair, quoteAssetDenom, p.Token0, p.Token1)
	}

       if baseAssetdenom == p.Token0 {
             return p.CurrentSqrtPrice.Power(2), nil
       } 
       return sdk.NewDec(1).Quo(p.CurrentSqrtPrice.Power(2)), nil
}
``` 
```

File: x/concentrated-liquidity/pool.go [hunk 14]
```diff
@@ -54,3 +63,31 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
 	osmoutils.MustSet(store, key, poolModel)
 	return nil
 }
+
+// convertConcentratedToPoolInterface takes a types.ConcentratedPoolExtension and attempts to convert it to a
+// swaproutertypes.PoolI. If the conversion is successful, the converted value is returned. If the conversion fails,
+// an error is returned.
+func convertConcentratedToPoolInterface(concentratedPool types.ConcentratedPoolExtension) (swaproutertypes.PoolI, error) {
```
p0mvn, line 70: Nice! Is this tested?

x/concentrated-liquidity/pool.go
```diff
@@ -54,3 +63,31 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
...
+		return nil, fmt.Errorf("given pool does not implement CFMMPoolI, implements %T", pool)
+	}
+	// Return the converted value
+	return pool, nil
+}
+
+// convertPoolInterfaceToConcentrated takes a swaproutertypes.PoolI and attempts to convert it to a
+// types.ConcentratedPoolExtension. If the conversion is successful, the converted value is returned. If the conversion fails,
+// an error is returned.
+func convertPoolInterfaceToConcentrated(poolI swaproutertypes.PoolI) (types.ConcentratedPoolExtension, error) {
```
p0mvn, line 84: and this

x/concentrated-liquidity/swaps.go
```diff
@@ -23,34 +23,41 @@ type SwapState struct {
...
-		Address:          gammtypes.NewPoolAddress(poolId).String(),
-		Id:               poolId,
-		CurrentSqrtPrice: currSqrtPrice,
-		CurrentTick:      currTick,
-		Liquidity:        sdk.ZeroDec(),
-		Token0:           denom0,
-		Token1:           denom1,
+
+	// Create a new concentrated liquidity pool with the given parameters.
+	poolI, err := model.NewConcentratedLiquidityPool(poolId, denom0, denom1, tickSpacing)
```
p0mvn, line 44: In `NewConcentratedLiquidityPool`, do have to return a `PoolI`. If not, maybe we just return the "concentrated interface extension" directly so that we don't have to do conversions?

x/concentrated-liquidity/swaps.go
```diff
@@ -23,34 +23,41 @@ type SwapState struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	conentratedPool, err := convertPoolInterfaceToConcentrated(&poolI)
```
p0mvn, line 49: typo

x/concentrated-liquidity/swaps.go
czarcas7ic: The other pool types return PoolI, I think we should follow the same API. Lmk what you think.

czarcas7ic:
@p0mvn thanks for the review, good for another one!

x/concentrated-liquidity/swaps.go
p0mvn: `NewConcentratedLiquidityPool` returns a CL pool, not a pool interface.

So we should not need to cast in the first place because CL pool should implement the `ConcentratedPoolExtension`. However, it does not seem to do so which isn't right.

p0mvn: We should make sure that CL pool implements `ConcentratedPoolExtension`

p0mvn:
LGTM once this thread is resolved:

https://github.com/osmosis-labs/osmosis/pull/3671#discussion_r1045297209

x/concentrated-liquidity/swaps.go
czarcas7ic: It seems if I want to follow Stableswap and Balancer, the CreateNewConcentratedLiquidityPool in pool.go should actually return poolI instead of ConcentratedPoolExtension

p0mvn: I'm referring to `NewConcentratedLiquidityPool` and the cast that happens after

p0mvn: `NewConcentratedLiquidityPool` returns `model.Pool`. `model.Pool` implements both `swaprouter.PoolI` and `cltypes.ConcentratedPoolExtension`. So cast shouldn't' be needed

czarcas7ic: Check the most recent commit, it seems I was able to use it directly as long as it was used with a pointer. Lmk if this satisfies the requirement 

https://github.com/osmosis-labs/osmosis/pull/3671/commits/f9b57316e24ae6f374fa062d52849adad8b1fb19

p0mvn (approved):

PR #3676, opened 2022-12-09
feat(osmomath): BigDec power function with integer exponent, overflow tests at max spot price

p0mvn:
Progress towards: https://github.com/osmosis-labs/osmosis/issues/3013

This PR introduces a `PowerInteger` function defined on `BigDec` to compute the power with an integer exponent.
The implementation is copied over from [`sdk.Dec`](https://github.com/osmosis-labs/cosmos-sdk/blob/f6e51566a5d878c561bc73fa4f0852ad6e02aa3c/types/decimal.go#L466) and uses square and multiply algorithm.

All pre-existing tests are also copied from the SDK and pass. More tests around max spot price are added to ensure that there are no-overflows occurring in geometric twap.

This function will be used by the future `Power` function that takes a decimal exponent. In that function, we split integer and non-integer part of the exponent. The integer part will be computed using this `PowerInteger` function. The non-integer part will be computed using Chebyshev Rational Approximation with 13 parameters. The progress towards that can be seen in: https://github.com/osmosis-labs/osmosis/pull/3620

### DIFF
--- a/osmomath/decimal.go
+++ b/osmomath/decimal.go
@@ -972,3 +972,23 @@ func (x BigDec) CustomBaseLog(base BigDec) BigDec {
 
 	return y
 }
+
+// PowerInteger takes a given decimal to an integer power
+// and returns the result. Non-mutative. Uses square and multiply
+// algorithm for performing the calculation.
+func (d BigDec) PowerInteger(power uint64) BigDec {
+	if power == 0 {
+		return OneDec()
+	}
+	tmp := OneDec()
+
+	for i := power; i > 1; {
+		if i%2 != 0 {
+			tmp = tmp.Mul(d)
+		}
+		i /= 2
+		d = d.Mul(d)
+	}
+
+	return d.Mul(tmp)
+}
--- a/osmomath/decimal_test.go
+++ b/osmomath/decimal_test.go
@@ -1116,6 +1116,18 @@ func (s *decimalTestSuite) TestPowerInteger() {
 
 			expectedResult: NewBigDec(81),
 		},
+		"-3 ^ 50 = 717897987691852588770249": {
+			base:     NewBigDec(-3),
+			exponent: 50,
+
+			expectedResult: MustNewDecFromStr("717897987691852588770249"),
+		},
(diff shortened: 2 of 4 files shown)


osmomath/decimal.go
```diff
@@ -972,3 +972,23 @@ func (x BigDec) CustomBaseLog(base BigDec) BigDec {
 
 	return y
 }
+
+// PowerInteger takes a given decimal to an integer power
+// and returns the result. Non-mutative. Uses square and multiply
+// algorithm for performing the calculation.
+func (d BigDec) PowerInteger(power uint64) BigDec {
```
ValarDragon, line 979: Do we have mutative multiplication on bigdec?

If so, lets make a PowerIntegerMut, have this create a copy, and then run PowerIntegerMut on the copy

osmomath/decimal_test.go
```diff
@@ -1018,3 +1019,123 @@ func (s *decimalTestSuite) TestCustomBaseLog() {
...
+			exponent: 3,
+
+			expectedResult: NewBigDec(27),
+		},
+		"-3 ^ 4 = 81": {
+			base:     NewBigDec(-3),
+			exponent: 4,
+
+			expectedResult: NewBigDec(81),
+		},
```
ValarDragon, line 1118: Can we do one to an exponent like `50` (something longer, with both 1's and 0's)

ValarDragon (approved):
LGTM, minor comments.

Can you add a changelog entry?

osmomath/decimal.go
p0mvn: There is no mutative multiplication. This would require introducing `MulMut` and refactoring `Mul` to utilize that.

I'm going to implement that in a separate PR. As soon as the new PR is merged, I will rebase this and implement the requested change.

p0mvn: https://github.com/osmosis-labs/osmosis/pull/3678

osmomath/decimal_test.go
p0mvn: Tests added

PR #3683, opened 2022-12-10
ProtoRev: Route building

davidterpay:

This PR implements the logic of building cyclic arbitrage routes given a swap. There are two primary methods for route generation: Highest Liquidity Pools and Hot Routes.

## Brief Changelog

- *Build routes that start and end with osmo (Highest Liquidity Pools)*
- *Build routes that start and end with atom (Highest Liquidity Pools)*
- *Build routes that start and end with osmo or atom from the set of hot routes (Hot Routes)*

## Testing and Verifying

- *Added unit tests for atom routes*
- *Added unit tests for osmo routes*
- *Added unit tests for routes built with tokenPairArbRoutes*
- *Added unit tests for BuildRoutes (which is a wrapper to the above three)*

### Highest Liquidity Pools

There are two types of trades that must be considered:

1. **The assets in the trade are neither Atom nor Osmo**
    - In this case, `x/protorev` will sandwich the pool with either Osmo or Atom on the other end. For example, say that a swap occurs on the **Akash** ↔ **Juno** pool. There are four possible routes that can be taken (in order of trades made starting on the left and ending on the right).
        - (Osmo → Akash), (Akash → Juno), (Juno → Osmo)
        - (Osmo → Juno), (Juno → Akash), (Akash → Osmo)
        - (Atom → Akash), (Akash → Juno), Juno → Atom)
        - (Atom → Juno), (Juno → Akash), (Akash → Atom)
    - Capturing cyclic arbitrage opportunities happen in the opposite direction of the trade. Using the example above, we can cut down the routes to only two routes if we know that the user traded Akash → Juno. The routes in that case would be
        - (Osmo → Juno), (Juno → Akash), (Akash → Osmo)
        - (Atom → Juno), (Juno → Akash), (Akash → Atom)
2. **The assets in the trade include either Atom or Osmo or both**
- If the **************Osmo ↔ Atom**************  pool is swapped against, then the highest liquidity pool route building method does not produce any routes. Otherwise, if only one of OSMO or ATOM was swapped against in a single pool, then `x/protorev` will look for the opposite asset pools from what was traded in the pool. For example, say a swap has been executed on the  ************Osmo ↔ TokenXYZ************ pool, tendering OSMO and receiving TokenXYZ, the route generated would be:
    - (Atom → TokenXYZ), (TokenXYZ → Osmo), (Osmo → Atom)

In both cases, the route that is built will always sandwich the swap that was made. However, we allow for more flexibility in route generation as the highest liquidity method may not be optimal via Hot Routes.

### Hot Routes
Populated through the admin account (which will be implemented in a later PR), the module’s keeper holds a KV store that associates token pairs (for example, osmo/juno) to the routes that result in a high percentage of arbitrage profit on Osmosis (as determined by external analysis).

The purpose of storing Hot Routes is a recognition that the Highest Liquidity Pool method may not present the best arbitrage routes. As such, hot routes can be configured through governance to store additional routes that may be more effective at capturing arbitrage opportunities. Each hot route will store a placeholder (the value 0) for where the current swapped pool will fit into the trade. TokenPairArbRoutes are always going to be sorted in the direction of the trade i.e. if we see a swap of ***Atom -> Osmo*** we can immediately find a route that goes in the opposite direction by replacing the placeholder (0) with this pool id.

### DIFF
--- a/x/protorev/keeper/routes.go
+++ b/x/protorev/keeper/routes.go
@@ -65,9 +65,9 @@ func (k Keeper) BuildTokenPairRoutes(ctx sdk.Context, tokenIn, tokenOut string, 
 // BuildTradeInfoHotRoute constructs a cyclic arbitrage route given a hot route from the store and information about the swap that should be placed
 // in the hot route.
 func (k Keeper) BuildTradeInfoHotRoute(ctx sdk.Context, route *types.Route, tokenIn, tokenOut string, poolId uint64) (Route, error) {
-	newRoute := Route{}
+	newRoute := Route{Trades: make([]TradeInfo, len(route.Trades))}
 
-	for _, trade := range route.Trades {
+	for index, trade := range route.Trades {
 		var newTrade TradeInfo
 		// 0 is a placeholder for swaps that should be entered into the hot route
 		if trade.Pool == 0 {
@@ -96,7 +96,7 @@ func (k Keeper) BuildTradeInfoHotRoute(ctx sdk.Context, route *types.Route, toke
 			}
 		}
 
-		newRoute.Trades = append(newRoute.Trades, newTrade)
+		newRoute.Trades[index] = newTrade
 	}
 
 	if err := k.CheckValidHotRoute(newRoute); err != nil {
--- a/x/protorev/keeper/routes.go
+++ b/x/protorev/keeper/routes.go
@@ -16,11 +16,15 @@ type TradeInfo struct {
 	Pool        gammtypes.CFMMPoolI
 }
 
-// BuildRoutes builds all of the possible arbitrage routes given the given tokenIn, tokenOut and poolId that were used in the swap
-func (k Keeper) BuildRoutes(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) [][]TradeInfo {
-	routes := make([][]TradeInfo, 0)
+type Route struct {
+	Trades []TradeInfo
+}
 
-	// Check hot routes if enabled
+// BuildRoutes builds all of the possible arbitrage routes given the tokenIn, tokenOut and poolId that were used in the swap
+func (k Keeper) BuildRoutes(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) []Route {
+	routes := make([]Route, 0)
+
+	// Append hot routes if they exist
 	if tokenPairRoutes, err := k.BuildTokenPairRoutes(ctx, tokenIn, tokenOut, poolId); err == nil {
 		routes = append(routes, tokenPairRoutes...)
 	}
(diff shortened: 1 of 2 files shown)


x/protorev/keeper/routes.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	return routes, nil
+}
+
+// BuildOsmoRoute builds a cyclic arbitrage route that starts and ends with osmo given the tokenIn, tokenOut and poolId that were used in the swap
+func (k Keeper) BuildOsmoRoute(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) ([]TradeInfo, error) {
+	// Creating the first trade in the arb
+	entryPoolId, err := k.GetOsmoPool(ctx, tokenOut)
```
stackman27, line 101: Does this get the highest liquidity osmo pool?

x/protorev/keeper/routes_test.go
```diff
@@ -0,0 +1,249 @@
...
+					{PoolId: 10, InputDenom: "bitcoin", OutputDenom: types.OsmosisDenomination},
+				},
+			},
+		},
+		{
+			description: "Route exists for swap in Bitcoin and swap out ethereum",
+			inputDenom:  "bitcoin",
+			outputDenom: "ethereum",
+			poolID:      19,
+			expected: [][]TestRoute{
```
stackman27, line 57: Asking to learn: In what case does making trade in "opposite direction" reduce arbitrage? Is there an example available?

x/protorev/keeper/routes_test.go
```diff
@@ -0,0 +1,249 @@
...
+		{
+			description:   "Route exists for swap in Akash and swap out Atom",
+			swapIn:        "akash",
+			swapOut:       types.AtomDenomination,
+			poolId:        1,
+			expectedRoute: []TestRoute{{25, types.OsmosisDenomination, types.AtomDenomination}, {1, types.AtomDenomination, "akash"}, {7, "akash", types.OsmosisDenomination}},
+			hasRoute:      true,
+		},
+		{
+			description:   "Route exists for swap in Terra and swap out Atom (no mapping pool)",
```
stackman27, line 177: Shouldn't this description be "Route doesnot exist for swap in Terra and swap out Atom because the pool doesnot exist"?

stackman27 (commented):
Left few questions that i have about Routes. Also realized that this module doesnot have a `README.md` yet, i think it'll be nice to add one 🌮

ValarDragon (approved):
Logic LGTM, nice job!

I only consider de-duplicating BuildOsmoRoute & BuildAtomRoute as blocking

x/protorev/keeper/routes.go
```diff
@@ -0,0 +1,206 @@
...
+		OutputDenom: types.OsmosisDenomination,
+		SwapFee:     exitPool.GetSwapFee(ctx),
+		Pool:        exitPool,
+	}
+
+	return []TradeInfo{entryTrade, middleTrade, exitTrade}, nil
+}
+
+// BuildAtomRoute builds a cyclic arbitrage route that starts and ends with atom given the tokenIn, tokenOut and poolId that were used in the swap
+func (k Keeper) BuildAtomRoute(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) ([]TradeInfo, error) {
```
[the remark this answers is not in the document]
davidterpay: On it.

x/protorev/keeper/routes.go
davidterpay: yes, it gets the highest liquidity osmo pool that has osmo <-> tokenOut as pool assets

x/protorev/keeper/routes.go
```diff
@@ -0,0 +1,206 @@
...
+	OutputDenom string
+	SwapFee     sdk.Dec
+	Pool        gammtypes.CFMMPoolI
+}
+
+// BuildRoutes builds all of the possible arbitrage routes given the given tokenIn, tokenOut and poolId that were used in the swap
+func (k Keeper) BuildRoutes(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) [][]TradeInfo {
+	routes := make([][]TradeInfo, 0)
+
+	// Check hot routes if enabled
```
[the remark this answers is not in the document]
davidterpay: thats a typo, will fix

x/protorev/keeper/routes_test.go
davidterpay: yea you are correct. my comment is a little confusing without context. i'll switch it up to what you suggested.

x/protorev/keeper/routes.go
```diff
@@ -0,0 +1,206 @@
...
+
+type TradeInfo struct {
+	InputDenom  string
+	OutputDenom string
+	SwapFee     sdk.Dec
+	Pool        gammtypes.CFMMPoolI
+}
+
+// BuildRoutes builds all of the possible arbitrage routes given the given tokenIn, tokenOut and poolId that were used in the swap
+func (k Keeper) BuildRoutes(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) [][]TradeInfo {
```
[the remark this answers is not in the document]
davidterpay: Wrapped the trades into a Route struct. Updated in the latest commit.

x/protorev/keeper/routes.go
```diff
@@ -0,0 +1,206 @@
...
+				newTrade = TradeInfo{
+					InputDenom:  trade.TokenIn,
+					OutputDenom: trade.TokenOut,
+					SwapFee:     pool.GetSwapFee(ctx),
+					Pool:        pool,
+				}
+			}
+
+			newRoute = append(newRoute, newTrade)
+		}
```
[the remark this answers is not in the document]
davidterpay: Added new `BuildTradeInfoHotRoute` and `CheckValidHotRoute` helper functions in latest commit.

x/protorev/keeper/routes.go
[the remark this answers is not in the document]
davidterpay: Added a new `BuildTradeInfoRoute` helper function which abstracts the getter and swap so code is not duplicated.

x/protorev/keeper/routes_test.go
NotJeremyLiu: @davidterpay is correct.

@stackman27 if a regular user swaps OSMO -> JUNO on a single pool (pool 5 for this example), they are selling OSMO for JUNO, increasing the price of JUNO and decreasing the price of OSMO (relative prices) on pool 5. 

If you assume other OSMO/JUNO pools or routes are in sync w.r.t prices, the pool the user just swapped on now has cheap OSMO and expensive JUNO. The goal for the cyclic arbitrage is to find JUNO on a different pool that's now "cheap" compared to the JUNO on pool 5. They want to buy JUNO elsewhere, then sell JUNO on Pool 5 where it's expensive.

So when looking at pool 5, an arbitrager will always go the opposite direction as the original swap to capture the mispricing opportunity (swapping JUNO -> OSMO on pool 5).

Hope that makes sense, but ask more questions if not!

ValarDragon (approved):

Issue #3603, opened 2022-12-02
Downtime detector module


ValarDragon:
## Background

For several use cases, we need a module that can detect when the chain is recovering from downtime. We want to be able to efficiently know "Has it been $RECOVERY_PERIOD minutes since the chain has been down for $DOWNTIME_PERIOD", and expose this as a query to contracts.

So for instance, you'd want to know if it has been at least 10 minutes, since the chain was down for > 30 minutes. Since you assume in such an event that it may take ~10 minutes for price oracles to be arb'd to correct.

## Suggested Design

Theres a couple designs, such as:

* Iterating over block times from the last N blocks (with a heuristic filter based on average block time)
  - Implies bounds on recovery time
  - Linear iteration if heuristic is met
  - Requires encoding expected block time
* Restricting downtime period, and storing a state entry for last time a downtime of length $D occurred

Because this may be in the hot-loop of contracts, I think we need to go with the approach that has minimal query compute, which is the latter. So I'll explain that in more depth.

We restrict the $DOWNTIME_PERIOD options that you can query, to be say `30seconds, 1 min, 2 min, 3 min, 4 min, 5 min, 10 min, 20 min, 30 min, 40 min, 50 min, 1 hr, 1.5hr, 2 hr, 2.5 hr, 3 hr, 4 hr, 5 hr, 6 hr, 9hr, 12hr, 18hr, 24hr, 36hr, 48hr` 

In the downtime detector module, we store state entries for:

* Last blocks timestamp
* For each period, last time there was downtime

Then in every begin block:

* Store last blocks timestamp
* if time since last block timestamp >= 30 seconds, iterate through all $DOWNTIME_PERIODS and add a state entry for the current block time

Then our query for has it been $RECOVERY_PERIOD since $DOWNTIME_PERIOD, simply reads the state entry for that $DOWNTIME_PERIOD, and then checks if time difference between now and that block is > RECOVERY_PERIOD.

## Acceptance Criteria

- [x] Downtime detector module exists
- [x] Has query for `RECOVERY_SINCE_DOWNTIME_LENGTH`
- [x] Query is efficient

ValarDragon:
(Only remaining item for this issue is testing on localosmosis + getting CLI logs)

ValarDragon:
This has been completed

PR #3688, opened 2022-12-11
Add downtime detector module

ValarDragon:
Cref #3603 

Adds downtime detector module.

- Remaining elements at time of PR open
  - [x] Tests
    - [x] Keeper methods
    - [x] CLI
    - [x] Query
    - [x] types
  - [x] app.go wiring

## Brief Changelog

- Adds downtime detector module
  - Change downtime to an enum
  - Implement begin block logic
  - Implement initialization logic
  - Implement query 
  - Implement CLI logic (pending #3647 )

## Testing and Verifying

TODO


app/apptesting/test_suite.go
```diff
@@ -374,7 +374,7 @@ func (s *KeeperTestHelper) BuildTx(
 // StateNotAltered validates that app state is not altered. Fails if it is.
 func (s *KeeperTestHelper) StateNotAltered() {
 	oldState := s.App.ExportState(s.Ctx)
-	s.Commit()
+	s.App.Commit()
```
ValarDragon, line 377: This was needed to fix "QueryStateAltered" tests, because Commit would start a new block, which would cause a state write from downtime detector.

app/keepers/modules.go
```diff
@@ -28,6 +28,7 @@ import (
 	ica "github.com/cosmos/ibc-go/v3/modules/apps/27-interchain-accounts"
 
 	_ "github.com/osmosis-labs/osmosis/v13/client/docs/statik"
+	downtimemodule "github.com/osmosis-labs/osmosis/v13/x/downtime-detector/downtimedetector_module"
```
nicolaslara, line 31: nit: `downtime-detector/downtimedetector_module` is ugly. Can it be `downtime-detector/module`? or `sdkmodule`, or something similar?

proto/osmosis/downtime-detector/v1beta1/downtime_duration.proto
```diff
@@ -0,0 +1,38 @@
...
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/any.proto";
+import "cosmos_proto/cosmos.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/osmosis-labs/osmosis/v13/x/downtime-detector/types";
+
+enum Downtime {
```
nicolaslara, line 12: I noticed you're not using `iota`. Is that so that we don't accidentally change the internal representation by reordering these?

x/downtime-detector/abci.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func (k *Keeper) BeginBlock(ctx sdk.Context) {
+	curTime := ctx.BlockTime()
+	lastBlockTime, err := k.GetLastBlockTime(ctx)
+	if err != nil {
+		ctx.Logger().Error("Downtime-detector, could not get last block time, did initialization happen correctly. " + err.Error())
+	}
+	downtime := curTime.Sub(lastBlockTime)
+	if downtime > 30*time.Second {
```
nicolaslara, line 18: nit: shouldn't we just do `k.saveDowntimeUpdates(ctx, downtime)` here and have that function determine when to save the downtime or not ?

x/downtime-detector/abci.go
```diff
@@ -0,0 +1,33 @@
...
+		k.saveDowntimeUpdates(ctx, downtime)
+	}
+	k.StoreLastBlockTime(ctx, curTime)
+}
+
+func (k *Keeper) saveDowntimeUpdates(ctx sdk.Context, downtime time.Duration) {
```
nicolaslara, line 24: can we get a gdoc here?

x/superfluid/client/cli/query_test.go
```diff
@@ -90,12 +90,14 @@ func (s *QueryTestSuite) TestQueriesNeverAlterState() {
...
 			&types.ConnectedIntermediaryAccountResponse{},
 		},
-		{
-			"Query estimate sfs delegated amount by validator & denom",
-			"/osmosis.superfluid.Query/EstimateSuperfluidDelegatedAmountByValidatorDenom",
-			&types.EstimateSuperfluidDelegatedAmountByValidatorDenomRequest{ValidatorAddress: s.val.String(), Denom: "gamm/pool/1"},
-			&types.EstimateSuperfluidDelegatedAmountByValidatorDenomResponse{},
-		},
+		// need to adapt s.val.String() to have an intermediate account,
+		// else the response is nil and theres a panic internally.
```
nicolaslara, line 94: Is this related to the downtime detector?

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/CosmWasm/wasmd v0.29.2-osmo-v13
 	github.com/cosmos/cosmos-proto v1.0.0-alpha8
-	github.com/cosmos/cosmos-sdk v0.45.11
+	github.com/cosmos/cosmos-sdk v0.46.6
```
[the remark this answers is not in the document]
ValarDragon: Oh no, it doesn't actually update the SDK either (as its replaced by the replace tag lower)

Our proto tooling just changes this :/

proto/osmosis/downtime-detector/v1beta1/downtime_duration.proto
ValarDragon: Yeah exactly! (Plus this is a proto file, which I don't think supports iota?)

x/downtime-detector/abci.go
ValarDragon: Good point, will change to do that

x/superfluid/client/cli/query_test.go
ValarDragon: Only insofar as I had to change the `func (s *KeeperTestHelper) StateNotAltered() {` line in apptesting because it was designed incorrectly (as triggered by this module). But fixing that, led to showing this superfluid test was designed with an error as well.

app/modules.go
```diff
@@ -175,6 +179,7 @@ func orderBeginBlockers(allModuleNames []string) []string {
 	// IBChost came after staking, before superfluid.
 	// TODO: Come back and delete this line after testing the base change.
 	ord.Sequence(stakingtypes.ModuleName, ibchost.ModuleName, superfluidtypes.ModuleName)
+	// We leave downtime-detector un-constrained.
```
czarcas7ic, line 182: What do we mean by un-constrained here?

x/downtime-detector/genesis.go
```diff
@@ -0,0 +1,48 @@
...
+	k.setGenDowntimes(ctx, gen.Downtimes)
+}
+
+func (k *Keeper) setGenDowntimes(ctx sdk.Context, genDowntimes []types.GenesisDowntimeEntry) {
+	for _, downtime := range genDowntimes {
+		k.StoreLastDowntimeOfLength(ctx, downtime.Duration, downtime.LastDowntime)
+	}
+}
+
+// ExportGenesis returns the twap module's exported genesis.
```
czarcas7ic, line 23:
```suggestion
// ExportGenesis returns the downtime detector module's exported genesis.
```

x/downtime-detector/genesis.go
```diff
@@ -0,0 +1,48 @@
+package downtimedetector
```
czarcas7ic, line 1: I believe we still need to add tests for these methods for this module. Its okay if we don't address it here, but should make an issue if not.

Edit: Ah looks like you did this in keeper_test.go. Should we move them to a genesis_test.go file?

x/downtime-detector/types/constants.go
```diff
@@ -1,9 +1,81 @@
...
+
+	"github.com/tidwall/btree"
+)
+
 const (
 	ModuleName = "downtime-detector"
 	StoreKey   = ModuleName
-	RouterKey  = ModuleName
+	// must be alphanumeric
+	RouterKey = "downtimedetector"
```
czarcas7ic, line 15: I think we made concentratedliquidity just not use the "-" in the module name to prevent the diff here, might prevent some confusion down the road.

x/downtime-detector/types/keys.go
```diff
@@ -0,0 +1,15 @@
+package types
+
+import fmt "fmt"
+
+// There are few of these keys, so we don't concern ourselves with small key names.
+var (
+	lastBlockTimestampKey      = []byte("last_block_timestamp")
+	lastDowntimeOfLengthPrefix = "last_downtime_of_length/%s"
```
czarcas7ic, line 8: Was the %s supposed to be in the var itself?

czarcas7ic (approved):
Minor comments 👍

proto/osmosis/downtime-detector/v1beta1/downtime_duration.proto
nicolaslara: lol, right! didn't notice this was in the proto

x/downtime-detector/types/keys.go
nicolaslara: var itself?

ValarDragon: Its used in a fmt.Sprintf call later

czarcas7ic: Ah, I have never seen it used this way but it makes sense, thanks :)

app/modules.go
ValarDragon: It can go anywhere in the sequence of begin block runs

x/downtime-detector/types/constants.go
ValarDragon: good idea

x/downtime-detector/genesis.go
ValarDragon: did it

Issue #3655, opened 2022-12-08
Make a client.toml denom alias list


ValarDragon:
## Background

We should make client.toml have a map of denom aliases. e.g. `uatom -> ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2` (I also want a version of this trusted metadatra registry to get on-chain)

Then we can in our new osmocli parsing logic, check every parsed denom/coin, and find replace the human readable name to an ibc denom internally. (And do the same decoding on the response)

## Suggested Design

- Add a new field to client.toml. I don't remember how you do this, can it be done purely from the osmosis repo?
- Make a method to read it from `client.Context`
- In osmocli, use this when parsing strings, for field names that contain `denom`.

## Acceptance Criteria

- [ ] Can create CLI commands using `uatom` as a denom

RusAkh:
Why does it have to be in client.toml? It seems to me this file should contain info related to users. Maybe I do not understand something, could you please clarify it a little

czarcas7ic:
@pysel do you have any updates on this issue?

pysel:
@czarcas7ic https://github.com/osmosis-labs/osmosis/pull/3689#issuecomment-1433516636

PR #3689, opened 2022-12-11
Dynamically generated IBCAliases map

pysel:
- add assetlist.json
- add script to parse assetlist
- enable caching
- mod sum

Related to: https://github.com/osmosis-labs/osmosis/issues/3655
## Changes
This PR introduces a map: `denom -> ibc_denom`. assetlist.json is the same as https://github.com/osmosis-labs/assetlists/blob/main/osmosis-1/osmosis-1.assetlist.json. However, for some reason I could not import it, so I just simply added it to the project. 
## Documentation
not documented yet


ValarDragon:
Hrmm, I like this, except for having to vendor the assetlist in the repo :/

RusAkh:
> Hrmm, I like this, except for having to vendor the assetlist in the repo :/

yeah I also do not like that, but I could not think of a different approach, wdyt?

ValarDragon:
I also don't like the idea of introducing a security critical dependency to assetlists, under a different maintenance structure. Maybe we:
* ship with something hardcoded and trimmed
* make a command to generate from assetlist (and that script live downloads assetlist into your home repo)

RusAkh:
hello @ValarDragon. I found a way to read assetlist. This is kind of a "lol" way to do it: 
- read it from GitHub in python and create a file
- parse file in Golang
- then delete file from Golang 
If you are ok with this approach we can leave it like that, so we do not need to bring the whole `assetlist.json` to repo

p0mvn:
Hi @pysel . Is there anything that needs to be done here? AFAIU, the conclusion was to pursue an approach via smart contracts for solving this. If you are in agreement, can we close this?

pysel:
yeah, guess so

PR #3697, opened 2022-12-13
Add more file-awareness for tests

ValarDragon:

This PR makes:
go linter CI only run if:
- go mod/sum edited
- go file edited
- CI configs edited
- Makefile edited

md linter CI only run if
- go mod;sum edited
- markdown file edited
- Makefile edited

E2E tests (and simulation) only run if
- go mod;sum edited
- non-test go file edited
- anything in tests/** edited
- .wasm binary edited
- Makefile edited

Compile CI only run if:
- go mod;sum edited
- non-test go file edited
- .wasm binary edited
- Makefile edited

## Brief Changelog

Should hopefully speedup our average case CI time more.

### DIFF
--- a/.github/workflows/sim.yml
+++ b/.github/workflows/sim.yml
@@ -14,11 +14,23 @@ jobs:
       - uses: actions/setup-go@v2.1.4
         with:
           go-version: 1.18
-      - name: Display go version
-        run: go version
+      -
+        name: Get git diff
+        uses: technote-space/get-diff-action@v6.0.1
+        with:
+          PATTERNS: |
+            **/**.wasm
+            **/**!(test).go
+            tests/**/**.go
+            go.mod
+            go.sum
+            Makefile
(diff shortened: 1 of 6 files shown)


.github/workflows/sim.yml
```diff
@@ -14,11 +14,23 @@ jobs:
       - uses: actions/setup-go@v2.1.4
         with:
           go-version: 1.18
-      - name: Display go version
-        run: go version
+      -
```
nicolaslara, line 17: Is the newline intended here?

nicolaslara (approved):
LGTM

.github/workflows/sim.yml
ValarDragon: Actually we should be changing all the ymls/adding linters.

Nicco made a standard at one point to have a new line after every `-` line, for easier copy/paste

But we didn't enforce it, so it fell off

PR #3589, opened 2022-12-01
Ack callbacks for IBC transfers

nicolaslara:
## What is the purpose of the change

This PR introduces the notion of ack callbacks. 

## Brief Changelog

 
  - Any transfer packet can now include a callback contract in its memo
  - When an ack is received, the specified contract will receive a sudo message with the information about the ack

## Testing and Verifying

ToDo: add generic tests for this beyond the existing test in the [crosschain swaps branch](https://github.com/osmosis-labs/osmosis/tree/nicolas/crosschain-swaps-new)

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no)
  - How is the feature or change documented? (spec in the readme)

PR #3698, opened 2022-12-13
Crosschain Swaps contracts and tests

nicolaslara:
Closes: #XXX

Introduces the crosschain swaps contracts and tests.

Note: This currently includes the changes from #3589 and should be merged after that. 

## Brief Changelog

  - Contracts and tests for crosschain swaps added to the repo

## Testing and Verifying

 - There are comprehensive tests for the contracts

  - How is the feature or change documented? (specification in the module readme and the contract readme  )


nicolaslara:
Marking this as state compatible. The state breaking changes here are from #3589

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg, SudoMsg};
+use crate::state::{Config, CHANNEL_MAP, CONFIG, RECOVERY_STATES};
+use crate::{execute, sudo};
+
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:crosschain-swaps";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+/// Handling contract instantiation
+#[cfg_attr(not(feature = "imported"), entry_point)]
```
iboss-ptk, line 19: Nice, this looks more straightforward than "library" and not be confused with `cargo build --lib`

x/ibc-hooks/contracts/crosschain-swaps/src/msg.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Message type for `execute` entry_point
+#[cw_serde]
+pub enum ExecuteMsg {
+    /// Execute a swap and forward it to the receiver address on the specified ibc channel
+    OsmosisSwap {
+        input_coin: Coin,
+        output_denom: String,
+        receiver: Addr,
+        slippage: Slippage,
+        next_memo: Option<String>,
```
iboss-ptk, line 40: what is next_memo?

x/ibc-hooks/contracts/crosschain-swaps/src/msg.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// This should be an instance of the Osmosis swaprouter contract
+    pub swap_contract: String,
+    /// If set to true, the contract will add a callback request on the packet
+    /// so that it gets notified when an ack is received or if the packet timed
+    /// out. If set to false, any funds sent on a packet that fails after a swap
+    /// will be stuck in this contract.
+    ///
+    /// The information about the packet sender and recovery address is still
+    /// stored, so recovery could be possible after a contract upgrade.
+    pub track_ibc_sends: Option<bool>,
```
iboss-ptk, line 17: What are the scenarios that `track_ibc_sends` is not needed?

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
```diff
@@ -0,0 +1,73 @@
...
+) -> Result<Response, ContractError> {
+    // deps.api.debug(&format!(
+    //     "received ack for packet {channel:?} {sequence:?}: {ack:?}, {success:?}"
+    // ));
+    let response = Response::new()
+        .add_attribute("contract", "crosschain_swaps")
+        .add_attribute("action", "receive_ack");
+
+    // Check if there is an inflight packet for the received (channel, sequence)
+    let recovery = INFLIGHT_PACKETS.may_load(deps.storage, (&source_channel, sequence))?;
```
iboss-ptk, line 43: what does this (with type `IBCTransfer`) called recovery?

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
```diff
@@ -0,0 +1,73 @@
...
+    // Remove the in-flight packet
+    INFLIGHT_PACKETS.remove(deps.storage, (&source_channel, sequence));
+
+    if success {
+        // If the acc is successful, there is nothing else to do and the crosschain swap has been completed
+        return Ok(response.add_attribute("msg", "packet successfully delviered"));
+    }
+
+    // If the ack is a failure, we create a recovery for the original sender of the packet.
+    let recovery_addr = recovery.recovery_addr.clone();
```
iboss-ptk, line 57: This does not need clone, no?

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
```diff
@@ -0,0 +1,73 @@
...
+    RECOVERY_STATES.update(deps.storage, &recovery_addr, |recoveries| {
+        // Since the recovery state and the in-flight packet store the same
+        // data, we can just modify the status and store the object in the
+        // RECOVERY_STATES map.
+        recovery.status = Status::AckFailure;
+        let Some(mut recoveries) = recoveries else {
```
iboss-ptk, line 63: `let else` is nice :P

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -0,0 +1,115 @@
...
+            ack,
+            success,
+        } => sudo::receive_ack(deps, channel, sequence, ack, success),
+        SudoMsg::ReceiveTimeout {} => unimplemented!(),
+    }
+}
+
+#[cfg_attr(not(feature = "imported"), entry_point)]
+pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> Result<Response, ContractError> {
+    deps.api
```
iboss-ptk, line 108: Wanna keep `deps.api.debug` on? I guess there is no problem since on mainnet it's gonna be disabled anyway, but there are commented ones elsewhere, so I think, either remove them all or uncomment other ones

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,261 @@
...
+    deps.api.debug(&format!("executing swap and forward"));
+    let config = CONFIG.load(deps.storage)?;
+
+    // Message to swap tokens in the underlying swaprouter contract
+    let swap_msg = SwapRouterExecute::Swap {
+        input_coin: input_coin.clone(),
+        output_denom,
+        slippage,
+    };
+    let msg = wasm_execute(config.swap_contract, &swap_msg, vec![input_coin])?;
```
iboss-ptk, line 41: oh unaware of this `wasm_execute` util function, nice!

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,261 @@
...
+    if let Some(memo) = &next_memo {
+        validate_memo(memo)?;
+    }
+
+    // Check that there isn't anything stored in SWAP_REPLY_STATES. If there is,
+    // it means that the contract is already waiting for a reply and should not
+    // override the stored state. This should only happen if a contract we call
+    // calls back to this one. This is likely a malicious attempt modify the
+    // contract's state before it has replied.
+    if SWAP_REPLY_STATES.may_load(deps.storage)?.is_some() {
```
iboss-ptk, line 55: nice catch, so no stack lol I think this can be recommended pattern for passing data along replies, and maybe create some abstraction around it.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,261 @@
...
+            msg: format!("No data"),
+        })
+    };
+
+    // Parse underlying response from the chain
+    let parsed =
+        cw_utils::parse_execute_response_data(&b).map_err(|e| ContractError::FailedSwap {
+            msg: format!("failed to parse: {e}"),
+        })?;
+    let swap_response: SwapResponse = from_binary(&parsed.data.unwrap())?;
```
iboss-ptk, line 96: why don't we `.map_err(|e| ContractError::FailedSwap { .. })`  instead of `unwrap` here as well?

x/ibc-hooks/contracts/crosschain-swaps/tests/crosschain_swap_test.rs
```diff
@@ -0,0 +1,109 @@
...
+        receiver: Addr::unchecked("osmo1l4u56l7cvx8n0n6c7w650k02vz67qudjlcut89"),
+        failed_delivery: None,
+        next_memo: None,
+    };
+    let funds: &[Coin] = &[Coin::new(10000, "uosmo")];
+    println!("{}", serde_json_wasm::to_string(&msg).unwrap());
+    let _res = wasm.execute(&crosschain_address, &msg, funds, &sender);
+    //dbg!(&res);
+
+    // This test cannot be completed until we have ibc tests on osmosis testing.
```
iboss-ptk, line 75: noted!

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,261 @@
...
+        };
+
+        return Ok(response
+            .set_data(to_binary(&data)?)
+            .add_message(ibc_transfer));
+    }
+
+    // Store the ibc send information and the user's failed delivery preference
+    // so that it can be handled by the response
+    FORWARD_REPLY_STATES.save(
```
iboss-ptk, line 183: Don't we want to do the same check as `SWAP_REPLY_STATES` here as well?

iboss-ptk (commented):
Great work! The code overall is very clean & well documented. Added some nits, questions and minor suggestions!

x/ibc-hooks/contracts/crosschain-swaps/build-schema/schema.rs
```diff
@@ -0,0 +1,12 @@
...
+use crosschain_swaps::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
+
+fn main() {
+    write_api! {
+        name: "crosschain-swaps",
+        instantiate: InstantiateMsg,
+        query: QueryMsg,
+        execute: ExecuteMsg,
+    };
+}
```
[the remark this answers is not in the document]
nicolaslara: yeah, I'm experimenting with a slightly different structure here. Had a discussion about it in either GH or twitter, but can't find it anymore

x/ibc-hooks/contracts/swaprouter/src/contract.rs
```diff
@@ -0,0 +1,90 @@
...
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    // validate owner address and save to state
+    let owner = deps.api.addr_validate(&msg.owner)?;
+    let state = State { owner };
+    STATE.save(deps.storage, &state)?;
+
+    // return OK
+    Ok(Response::new().add_attribute("method", "instantiate"))
+    // .add_attribute("owner", owner.as_str()))
```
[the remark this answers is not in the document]
nicolaslara: it was supposed to be removed. I'll remove it

x/ibc-hooks/contracts/swaprouter/src/msg.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[cw_serde]
+pub enum Slippage {
+    MaxSlippagePercentage(Decimal),
+    MinOutputAmount(Uint128),
+}
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    TransferOwnership {
```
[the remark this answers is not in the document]
nicolaslara: will do

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
nicolaslara: Agreed! have an open issue for it here: https://github.com/CosmWasm/cw-template/issues/140

x/ibc-hooks/contracts/crosschain-swaps/src/msg.rs
nicolaslara: that's the memo to use in the forwarded packet. I'll document that.

x/ibc-hooks/contracts/crosschain-swaps/src/msg.rs
nicolaslara: If the chain doesn't have support for sending IBC acks back to the contract (like right now), then there is no way to know if the packet was received or not. So we skip tracking in the contract altogether.

We also remove the memo from the packet (as it would just be forwarded with the packet and have no effect unless the receiving chain handles it)

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
nicolaslara: the name is confusing. Since we're using the `INFLIGHT_PACKETS` and not `RECOVERY_STATE`. 

Here we want to check if the ack we received is for a packet we had sent. The name "recovery" is beacause we are going to add the same packet as a recovery state later on, but I can just rename or shadow it later.

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
nicolaslara: I need to clone it because the it gets "partially moved" to the closure and then I can't use recovery otherwise. 

But let me know if there's a better way to do this! I may be missing something

x/ibc-hooks/contracts/crosschain-swaps/tests/crosschain_swap_test.rs
nicolaslara: 😆

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
nicolaslara: Same as with the comments above: I have mixed feelings about that. 

I think it makes sense to keep some on and some commented out for future debugging. I also don't think that call has a performance impact on prod

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: lol yeah, it wasn't necessary here (it might be for more flexible contracts). 

I was also thinking about how to create an abstraction for this, but didn't have the time to go into it. It would be great to have standards and tooling for it!

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: parsed.data is an `Option`. I think we're garanteed to have data here, but it's probably better to do 

`let swap_response: SwapResponse = from_binary(&parsed.data.unwrap_or_default())?;`

anyway. That way if the data isn't there `from_binary` will fail with a nice error about not being able to decode SwapResponse

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: I don't think it's necessary here as that state does not depend on the call to another contract, so there's no one who can sneakily call this while it's in a dirty state (to be fair, I think that's true about the other one too as long as the underlying contract is our swaprouter).

But better safe than sorry. I'll add it.

x/ibc-hooks/contracts/swaprouter/tests/swap_test.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+    };
+}
+
+const INITIAL_AMOUNT: u128 = 1_000_000_000_000;
+
+fn test_swap_success_case(msg: ExecuteMsg, funds: &[Coin]) {
+    let (app, sender, _res) = setup_route_and_execute_swap(&msg, &funds);
+    //dbg!(res);
+    //println!("{:?}", String::from_utf8(to_vec(&msg).unwrap()).unwrap());
```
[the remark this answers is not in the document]
iboss-ptk: No strong opinion here as well

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
iboss-ptk: Oh yeah unwrap_or_default make sense. Generally I think if it's not obvious that why it's guaranteed to work, I would prefer expect over unwrap.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
iboss-ptk: Yeah feels safer to defense in depth

wasmbinding/stargate_whitelist.go
```diff
@@ -37,6 +37,8 @@ var stargateWhitelist sync.Map
 func init() {
 	// cosmos-sdk queries
 
+	setWhitelistedQuery("/cosmos.auth.v1beta1.Query/Account", &authtypes.QueryAccountResponse{})
```
nicolaslara, line 40: Note to self. Where did this come from? Do we need it?

iboss-ptk (approved):
LGTM

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: that makes sense! I changed it to unwrap_or_default(), but happy to change it to expect() if you have strong opinions

wasmbinding/stargate_whitelist.go
ValarDragon: Lets remove from this PR, and do in another

x/ibc-hooks/contracts/crosschain-swaps/Cargo.toml
```diff
@@ -0,0 +1,68 @@
+# This feature has been stabilized in 1.64. If using an earlier version (like in
+# rust-optimizer-arm:0.12.8) you should uncomment this like to be able to
+# compile the contract
```
ValarDragon, line 3: I generally think we should declare an MSRV of Rust 1.64 (two versions behind latest)

x/ibc-hooks/contracts/crosschain-swaps/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+
+[package]
+name = "crosschain-swaps"
+version = "0.1.0"
+authors = ["Nicolas Lara <nicolaslara@gmail.com>"]
+edition = "2021"
+rust-version  = "1.65.0"
+
+exclude = [
+  # Those files are rust-optimizer artifacts. You might want to commit them for convenience but they should not be part of the source code publication.
```
ValarDragon, line 15:
```suggestion
  # Those files are rust-optimizer artifacts. You might want to commit them for convenience but they should not be part of the crates publication.
```

x/ibc-hooks/contracts/crosschain-swaps/schema/crosschain-swaps.json
```diff
@@ -0,0 +1,270 @@
...
+          },
+          "denom": {
+            "type": "string"
+          }
+        }
+      },
+      "Decimal": {
+        "description": "A fixed-point decimal value with 18 fractional digits, i.e. Decimal(1_000_000_000_000_000_000) == 1.0\n\nThe greatest possible value that can be represented is 340282366920938463463.374607431768211455 (which is (2^128 - 1) / 10^18)",
+        "type": "string"
+      },
```
ValarDragon, line 119: wtf, this typedef isn't just imported from somewhere :/

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
```diff
@@ -0,0 +1,9 @@
+// Msg Reply IDs
+pub const SWAP_REPLY_ID: u64 = 1u64;
+pub const FORWARD_REPLY_ID: u64 = 2u64;
+
+// IBC timeout
+pub const PACKET_LIFETIME: u64 = 86400u64;
```
ValarDragon, line 6: Can we comment what this is / what units they are?

I assume seconds / 24 hours?

May want to consider elongating this

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
```diff
@@ -0,0 +1,9 @@
+// Msg Reply IDs
+pub const SWAP_REPLY_ID: u64 = 1u64;
+pub const FORWARD_REPLY_ID: u64 = 2u64;
```
ValarDragon, line 3: Should this be an enum?

x/ibc-hooks/contracts/swaprouter/build-schema/schema.rs
```diff
@@ -0,0 +1,24 @@
...
+    export_schema(&schema_for!(InstantiateMsg), &out_dir);
+    export_schema(&schema_for!(ExecuteMsg), &out_dir);
+    export_schema(&schema_for!(QueryMsg), &out_dir);
+
+    export_schema(&schema_for!(State), &out_dir);
+    export_schema(&schema_for!(SwapMsgReplyState), &out_dir);
+
+    export_schema(&schema_for!(GetOwnerResponse), &out_dir);
+    export_schema(&schema_for!(GetRouteResponse), &out_dir);
+}
```
ValarDragon, line 24: This entire folder feels silly :thinking:

x/ibc-hooks/contracts/swaprouter/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+# This feature has been stabilized in 1.64. If using an earlier version (like in
+# rust-optimizer-arm:0.12.8) you should uncomment this like to be able to
+# compile the contract
+#
+#cargo-features = ["workspace-inheritance"]
+
+[package]
+authors = ["Sunny Aggarwal <sunnya97@protonmail.ch>"]
```
ValarDragon, line 8: This still accurate?

ValarDragon:
Through the boilerplate files:
image
image

Happy to get these merged in a standalone PR if you'd like.

Going to try to review swaprouter first tmrw, (and recommend we get that merged in one PR), then cross-chain swaps second

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
iboss-ptk: `u64` is required for reply id

This might work:
```rs
#[repr(u64)]
pub enum Reply {
  Swap = 1,
  Forward = 2,
}
```

x/ibc-hooks/contracts/swaprouter/build-schema/schema.rs
iboss-ptk: we've got `write_api!` no? should be able to replace this with that

x/ibc-hooks/contracts/crosschain-swaps/Cargo.toml
nicolaslara: Why not latest? I find nightly annoying cause it changing so often means mental overhead, but since the code is compiled to wasm, any version should work. AFAIK there are no issues installing the versions to get the code to compile.

x/ibc-hooks/contracts/crosschain-swaps/schema/crosschain-swaps.json
nicolaslara: 🤷 I'm sure that's one of the many opinions Dan has about json schemas.

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
nicolaslara: what value would you suggest? a week?

x/ibc-hooks/contracts/swaprouter/build-schema/schema.rs
nicolaslara: yeah, this was written before `write_api!`. I can update it.

x/ibc-hooks/contracts/swaprouter/Cargo.toml
nicolaslara: Most of swaprouter was written by Sunny. I just added twap, transfer ownership and some cleanup. Maybe we should have a generic authors string? "osmosis team" or something

nicolaslara:
Not sure why GH is listing a bunch of the CL changes in the diff of this PR. May delete and recreate later.

x/ibc-hooks/contracts/crosschain-swaps/Cargo.toml
ValarDragon: I feel like it depends on how much of a library this is supposed to be. When I was an active rust dev, I often found it super annoying when dependencies would declare latest MSRV, when I wanted to provide longer term guarantees, or have things not break at every rust release. (Dependents would complain about latest MSRV)

(Also this is not remotely PR blocking, or even release blocking)

x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+``` json
+{"swap_contract": "osmo1thiscontract", "track_ibc_sends": false, "channels": [["cosmos", "channel-0"], ["juno", "channel-42"]]}
+```
+
+## Usage
+
+### Via IBC
+
+Assuming the current implementation of the wasm middleware on Osmosis, the memo
```
ValarDragon, line 28:
```suggestion
Assuming the current implementation of the wasm middleware on Osmosis v14 (`x/ibc-hooks/v0.0.5`), the memo
```

x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -0,0 +1,214 @@
...
+}}
+```
+
+Channels are determined by the prefixes specified in the contract during
+instantiation, so the user needs to provide a receiver with the supported
+prefix. In future (once ack/timeout tracking is supported by the chain), we
+could enable support for any channel by specifying it as a parameter.
+
+The `slippage` can be set to a percentage of the twap price (as shown above), or as
+the minimum amount of tokens expected to be received: `{"min_output_amount": "100"}`.
```
ValarDragon, line 53: I think we need some key adjustment for twap, e.g.
```
{
  "slippage": {
     "twap": {"window": "24h", "slippage_percent": "5"},
     "min amount out": {<however you specify amount>},
}
```

if twap window left blank infer some value. Always check min amount out if present

x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -0,0 +1,214 @@
+# Crosschain Swaps
```
ValarDragon, line 1: Nice job on this! I think we should link to the counterparty chain outpost contract in readme as well (once that has its own place, etc)

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
ValarDragon: (This comment still stands)

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
ValarDragon: yeah a week timeout sgtm

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -0,0 +1,116 @@
...
+pub fn sudo(deps: DepsMut, _env: Env, msg: SudoMsg) -> Result<Response, ContractError> {
+    match msg {
+        SudoMsg::IBCLifecycleComplete(IBCLifecycleComplete::IBCAck {
+            channel,
+            sequence,
+            ack,
+            success,
+        }) => sudo::receive_ack(deps, channel, sequence, ack, success),
+        SudoMsg::IBCLifecycleComplete(IBCLifecycleComplete::IBCTimeout { .. }) => {
+            unimplemented!()
```
ValarDragon, line 102: What does this unimplemented macro do here? If it returns an error, what deos a contract error do, does it block the timeout from happening?

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
ValarDragon: I think this is my ignorance of cosmwasm -- who is this entrypoint for? Synchronous calls from other contracts?

x/ibc-hooks/contracts/crosschain-swaps/src/ibc.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(
+    Clone,
+    PartialEq,
+    Eq,
+    ::prost::Message,
+    serde::Serialize,
+    serde::Deserialize,
+    schemars::JsonSchema,
+)]
+pub struct Height {
```
ValarDragon, line 12: Is it important this struct be called `Height`, or can it be called `IbcCounterpartyHeight` ?

(Its confusing as its not just the one u64 as you'd expect)

ValarDragon: oh is this derived from proto tooling?

If so we should add a `AUTO-GENERATED` comment at top of file

x/ibc-hooks/contracts/crosschain-swaps/src/sudo.rs
```diff
@@ -0,0 +1,73 @@
+use cosmwasm_std::{DepsMut, Response};
```
ValarDragon, line 1: I suggest naming this `lifecycle_complete.rs` or smth along those lines

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+/// It's objective is to trigger a swap between the supplied pairs
+///
+#[allow(clippy::too_many_arguments)]
+pub fn swap_and_forward(
+    deps: DepsMut,
+    block_time: Timestamp,
+    contract_addr: Addr,
+    input_coin: Coin,
+    output_denom: String,
+    slippage: Slippage,
```
ValarDragon, line 27: (minor nit, feel free to ignore)

I feel like ideally the aliases for this import would be:
`SwapRouter::Slippage`

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+        slippage,
+    };
+    let msg = wasm_execute(config.swap_contract, &swap_msg, vec![input_coin])?;
+
+    // Check that the received is valid and retrieve its channel
+    let (valid_channel, valid_receiver) = validate_receiver(deps.as_ref(), receiver)?;
+    // If there is a memo, check that it is valid
+    if let Some(memo) = &next_memo {
+        validate_memo(memo)?;
+    }
```
ValarDragon, line 48: Perhaps good to clarify what is the needed safety check? (and maybe a one sentence for why no callback?)

```suggestion
    if let Some(memo) = &next_memo {
        let json = parse_json(memo)?;
        ensure_no_callback(json)?;
    }
```

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+            forward_to: ForwardTo {
+                channel: valid_channel,
+                receiver: valid_receiver,
+                next_memo,
+                failed_delivery,
+            },
+        },
+    )?;
+
+    Ok(Response::new().add_submessage(SubMsg::reply_on_success(msg, SWAP_REPLY_ID)))
```
ValarDragon, line 76: What is this `add_submessage` doing?

x/ibc-hooks/contracts/crosschain-swaps/README.md
ValarDragon: Maybe we should also add some docs that slippage is coming from swaprouter?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+    // If there is a memo, check that it is valid
+    if let Some(memo) = &next_memo {
+        validate_memo(memo)?;
+    }
+
+    // Check that there isn't anything stored in SWAP_REPLY_STATES. If there is,
+    // it means that the contract is already waiting for a reply and should not
+    // override the stored state. This should only happen if a contract we call
+    // calls back to this one. This is likely a malicious attempt modify the
+    // contract's state before it has replied.
```
ValarDragon, line 54: Nice!

This reminds me I need to make a github issue for blocking this class of re-entrancy bugs via the VM being more opinionated in the callstack

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+    if SWAP_REPLY_STATES.may_load(deps.storage)?.is_some() {
```
ValarDragon, line 55: I was confused on first read by this data structure though -- this implies that there can only be one swap reply happening concurrently? (Rather than a map from an in-progress ID to reply state)

This makes sense though based on the concurrency model

ValarDragon: If this is true, lets rename the var to `SWAP_REPLY_STATE` ?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+    let SubMsgResult::Ok(SubMsgResponse { data: Some(b), .. }) = msg.result else {
+        return Err(ContractError::FailedSwap {
+            msg: format!("No data"),
+        })
+    };
+
+    // Parse underlying response from the chain
+    let parsed =
+        cw_utils::parse_execute_response_data(&b).map_err(|e| ContractError::FailedSwap {
+            msg: format!("failed to parse: {e}"),
```
ValarDragon, line 94: why is the error message "failed to parse" ?

ValarDragon, line 94:
```suggestion
            msg: format!("failed to parse swaprouter response: {e}"),
```

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+        })?
+    } else {
+        serde_json_wasm::from_str("{}").unwrap()
+    };
+
+    // If tracking callbacks, we want to include the callback key in the memo
+    // without otherwise modifying the provided one
+    let memo = match config.track_ibc_callbacks {
+        true => {
+            let serde_cw_value::Value::Map(mut m) = memo else { unreachable!() };
```
ValarDragon, line 117: Why don't we just define memo to be of the correct type on line 104?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+            Coin::new(
+                swap_response.amount.into(),
+                swap_response.token_out_denom.clone(),
+            )
+            .into(),
+        ),
+        sender: contract_addr.to_string(),
+        receiver: swap_msg_state.forward_to.receiver.clone().into(),
+        timeout_height: None,
+        timeout_timestamp: Some(ts.nanos()),
```
ValarDragon, line 153: oh wow, TIL this is nanoseconds (https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/tx.proto#L38-L40)

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    // Base response
+    let response = Response::new()
+        .add_attribute("status", "ibc_message_created")
+        .add_attribute("ibc_message", format!("{:?}", ibc_transfer));
+
+    if !config.track_ibc_callbacks || swap_msg_state.forward_to.failed_delivery.is_none() {
+        // If we're not tracking callbacks, or there isn't any recovery addres,
+        // then there's no need to listen to the response of the send.
+        // The response data
```
ValarDragon, line 165: Dangling comment?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+        let channel_id = swap_msg_state.forward_to.channel;
+        let to_address = swap_msg_state.forward_to.receiver;
+        let data = CrosschainSwapResponse {
+            msg: format!("Sent {amount}{denom} to {channel_id}/{to_address}"),
+        };
+
+        return Ok(response
+            .set_data(to_binary(&data)?)
+            .add_message(ibc_transfer));
+    }
```
ValarDragon, line 179: ngl, don't really understand why the data format for the response is the way it is.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+                next_memo,
+                failed_delivery,
+            },
+        },
+    )?;
+
+    Ok(Response::new().add_submessage(SubMsg::reply_on_success(msg, SWAP_REPLY_ID)))
+}
+
+pub fn handle_swap_reply(deps: DepsMut, msg: Reply) -> Result<Response, ContractError> {
```
ValarDragon, line 79: (annoying nit) Can we break some of this logic into smaller functions

ValarDragon: e.g. building response in case of no callback tracking, which seems like mostly boilerplate

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
ValarDragon: or what the add message is here

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+        timeout_timestamp: Some(ts.nanos()),
+        memo: memo_str,
+    };
+
+    // Base response
+    let response = Response::new()
+        .add_attribute("status", "ibc_message_created")
+        .add_attribute("ibc_message", format!("{:?}", ibc_transfer));
+
+    if !config.track_ibc_callbacks || swap_msg_state.forward_to.failed_delivery.is_none() {
```
ValarDragon, line 162: The respone flow / general expectations isn't super clear. Can we lay this out a bit more in the README, so callers can get a better sense of it?

ValarDragon: Perhaps just a response format section in readme? (or code doc somewhere, e.g. where its defined)

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+            amount: swap_response.amount.into(),
+            denom: swap_response.token_out_denom,
+            failed_delivery: swap_msg_state.forward_to.failed_delivery,
+        },
+    )?;
+
+    Ok(response.add_submessage(SubMsg::reply_on_success(ibc_transfer, FORWARD_REPLY_ID)))
+}
+
+use ::prost::Message; // Proveides ::decode() for MsgTransferResponse
```
ValarDragon, line 207: (nit) should this be at top of file?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+
+pub fn handle_forward_reply(deps: DepsMut, msg: Reply) -> Result<Response, ContractError> {
+    // Parse the result from the underlying chain call (IBC send)
+    let SubMsgResult::Ok(SubMsgResponse { data: Some(b), .. }) = msg.result else {
+        return Err(ContractError::FailedIBCTransfer { msg: format!("failed reply: {:?}", msg.result) })
+    };
```
ValarDragon, line 213: Sorry this is likely my cosmwasm ignorance, but I am a bit confused about the cosmwasm reply data type vs `Forward_Reply_Id`

Were assigning an ID for every reply that we are then supposed to forward, and each reply is communicated to the contract via cw?

I think this confusion would be mitigated by using `cosmwasm_std::Reply` as the type if your cool with that

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    // Similar consideration as the warning above. Is it safe for this to be an Item?
+    let ForwardMsgReplyState {
+        channel_id,
+        to_address,
+        amount,
+        denom,
+        failed_delivery,
+    } = FORWARD_REPLY_STATES.load(deps.storage)?;
+    FORWARD_REPLY_STATES.remove(deps.storage);
```
ValarDragon, line 231: What is the concern with being an item? Is there some mutable data in the struct?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+    // If a recovery address was provided, store sent IBC transfer so that it
+    // can later be recovered by that addr.
+    if let Some(Recovery { recovery_addr }) = failed_delivery {
+        let recovery = IBCTransfer {
+            recovery_addr,
+            channel_id: channel_id.clone(),
+            sequence: response.sequence,
+            amount,
+            denom: denom.clone(),
+            status: Status::Sent,
```
ValarDragon, line 242: What is this status? is it internal tracking?

ValarDragon: ah this entire IBC transfer is a this-contract data type.

Sorry for the same comment a bunch of times -- I feel like for things like this that juggle many things that could sensibly be from multiple packages, having the types be qualified (e.g. state::IBCTransfer) would make whats going on more obvious on first read

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -0,0 +1,271 @@
...
+        to_address,
+        amount,
+        denom,
+        failed_delivery,
+    } = FORWARD_REPLY_STATES.load(deps.storage)?;
+    FORWARD_REPLY_STATES.remove(deps.storage);
+
+    // If a recovery address was provided, store sent IBC transfer so that it
+    // can later be recovered by that addr.
+    if let Some(Recovery { recovery_addr }) = failed_delivery {
```
ValarDragon, line 235: I don't understand what is this `failed_delivery` argument were parsing? Seems like were marking this as in-flight, not failed?

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: Since Map is a variant, I would have to declare it using the variant's type: `BTreeMap<Value, Value>`, and then deal with more wrappings/unwrappings, I think. 

The only reason this is unreachable is because we run `validate_memo(memo)?;` on line 50 before the reply executes. I can look into passing the btree instead, but it doesn't seem as straightforward when I looked quickly now

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: yeah, I think I've incorrectly used seconds about 3 times and ended up debugging only to relearn this

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: clippy doesn't care. I like importing trait implementations close to where they're used (because one of my pet peeves with rust is not knowing where some impls come from like in this one)

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: ahh, that makes sense! I'm not a fan of reply ids either, but that's a wider cosmwasm discussion. 

Fixed.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: That's a dangling comment. I've removed it.

Above we're doing
```
    if FORWARD_REPLY_STATES.may_load(deps.storage)?.is_some() {
        return Err(ContractError::ContractLocked {
            msg: "Already waiting for a reply".to_string(),
        });
    }
```

to ensure that there's no "reentrency" (if a submessage would call this contract again). That is not strictly necessary, as the only submessages this contract sends are to swaprouter and ibc, which don't do that. I wasn't checking this before and there used to be a few comments discussing why it wasn't a problem but questioning if it was ok to leave it that way or to check defensively. 

After discussing it with Boss, we concluded it was better to be defensive, so now we check this and that comment isn't needed anymore. 

For background, the issue with it being an Item is that that's basically a singleton, so if we enter the contract again, then the value would be overriden. For contracts that expect this behaviour, we could use a Stack instead (felt silly to use a stack that I know is always gonna have depth 1)

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: qualified.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: failed_delivery comes from:

```
#[cw_serde]
pub struct ForwardTo {
    pub channel: String,
    pub receiver: Addr,
    pub next_memo: Option<String>,
    pub failed_delivery: Option<Recovery>,
}

#[cw_serde]
pub struct SwapMsgReplyState {
    pub swap_msg: SwapRouterExecute,
    pub contract_addr: Addr,
    pub block_time: Timestamp,
    pub forward_to: ForwardTo,
}
```

It represents what to do on a failed delivery (right now that's a recovery addr, but I can imagine it being more complex). I can see how it is a confusing name though. Maybe something like:

```
enum FailedIBCForwardAction {
   None,
   LocalRecovery { addr: Addr }
   ...
}
```
(and remembering to rename IBC so it doesn't serialize to `i_b_c` 😓 )

wasmbinding/stargate_whitelist.go
nicolaslara: done. Not sure what that was

x/ibc-hooks/contracts/crosschain-swaps/README.md
nicolaslara: yes! once they... exist 😓

x/ibc-hooks/contracts/crosschain-swaps/README.md
nicolaslara: I want to move Slippage (and some other utilities) to a package in the outpost contract repo.  Makes sense to add the window.

Not sure what you mean by: Always check min amount out if present

x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -0,0 +1,214 @@
...
+# Transaction broadcasting mode (sync|async|block)
+broadcast-mode = "block"
+```
+
+### Deploying the contracts
+
+Store the swaprouter code:
+
+``` sh
+> osmosisd-testnet tx wasm store ./bytecode/swaprouter.wasm --from owner  --gas auto --gas-prices 0.1uosmo --gas-adjustment 1.3 -y
```
nicolaslara, line 142: Note to self: I have better commands for this now. Cleanup and add the script I'm using to test this on the localrelayer tests.

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
nicolaslara: had to use a crate to get the value when passing it as a param, but it now is: 

```
use enum_repr::EnumRepr;

// Msg Reply IDs
#[EnumRepr(type = "u64")]
pub enum MsgReplyID {
    Swap = 1,
    Forward = 2,
}
```

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
nicolaslara: changed

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: I think that makes sense

nicolaslara: changed.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: Cleaned this up and split into multiple single-purpose functions. Added some docs.

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: Confio has been working on a cosmwasm framework ([sylvia](https://github.com/CosmWasm/sylvia/)). I've looked a but into it, but I don't think they have anything specifically for how the callstack is defined

x/ibc-hooks/contracts/crosschain-swaps/src/execute.rs
nicolaslara: That is correct. Renamed.

x/ibc-hooks/contracts/crosschain-swaps/src/ibc.rs
nicolaslara: This is not auto generated. Wrote it manually cause osmosis-std didn't implemented. I think we have it now. I'll check and may replace this.

nicolaslara: Double checked. We don't have them yet, but there's an open issue: https://github.com/osmosis-labs/osmosis-rust/issues/52

x/ibc-hooks/contracts/crosschain-swaps/src/consts.rs
ValarDragon: Cool cool!

x/ibc-hooks/contracts/crosschain-swaps/src/contract.rs
ValarDragon: TY for the detailed explanation! Makes sense

x/ibc-hooks/contracts/crosschain-swaps/README.md
nicolaslara: hmm, I guess that would work. But I changed it to an enum, so it can now be either `{"twap":{"slippage_percentage":"20","window_seconds":10}` or `{"min_output_amount": "100"}`.

ValarDragon (approved):
Down to merge to main and review more after merge

PR #3708, opened 2022-12-14
feat(osmomath): Exp2 function

p0mvn:
Closes: #XXX

This PR adds an `Exp2` function that computes $$2^{decimal}$$

The following property is used:
$$2^{decimal} = 2^{integer} * 2^{fractional}$$

The integer exponent result is computed using the existing `PowerInteger` function.

The fractional exponent result is computed using the 13-parameter Chebyshev rational approximation. The numerator coefficients are rounded down and the denominator coefficients are rounded up at the precision end. However, this does not guarantee that the approximated result is always smaller than the actual results (as determined by WolframAlpha).

The Python approximation script is modified to utilize the [production coefficients ](https://github.com/osmosis-labs/osmosis/blob/0cc8250f2795860ef2fe41896a0dc44ff68005dd/scripts/approximations/main.py#L194-L212) for estimating relative errors. Please see the results:
image

Additionally, `ErrTolerance` was refactored to use `sdk.Dec` for `AdditiveTolerance` for more accuracy in tests.

In a future PR, `Power` function with a decimal exponent can be added. This function will use `exp2` by utilizing the following property:
`x^y = exp2(y * log_2{x})`

## Testing and Verifying

Tests with links to WolframAlpha were added. The absolute and the relative errors are bounded.

## Open Questions

- Is it acceptable that we cannot guarantee rounding behavior?

  - How is the feature or change documented? godocs and approximation scripts


osmomath/exp2.go
```diff
@@ -0,0 +1,93 @@
...
+		MustNewDecFromStr("0.052724071627342653404436933178482287"),
+		MustNewDecFromStr("0.004760950735524957576233524801866342").Neg(),
+		MustNewDecFromStr("0.000267168475410566529819971616894193"),
+		MustNewDecFromStr("0.000008923715368802211181557353097439").Neg(),
+		MustNewDecFromStr("0.000000140277233177373698516010555916"),
+	}
+
+	// maxSupportedExponent = 2^10. The value is chosen by finding
+	// at which value some of the underlying internal functions overflow.
+	// If needed, exp2 can be reimplemented to allow for greater exponents.
```
p0mvn, line 32:
```suggestion
	// maxSupportedExponent = 2^10. The value is chosen by benchmarking
	// when the underlying internal functions overflow.
	// If needed in the future, Exp2 can be reimplemented to allow for greater exponents.
```

osmomath/exp2.go
```diff
@@ -0,0 +1,93 @@
...
+	maxSupportedExponent = MustNewDecFromStr("2").PowerInteger(9)
+)
+
+// Exp2 takes 2 to the power of a given non-negative decimal exponent.
```
p0mvn, line 36:
```suggestion
// Exp2 takes 2 to the power of a given non-negative decimal exponent
```

osmomath/exp2.go
```diff
@@ -0,0 +1,93 @@
...
+// and returns the result.
+// The computation is performed by using th following property:
+// 2^decimal_exp = 2^{integer_exp + fractional_exp} = 2^integer_exp * 2^fractional_exp
+// The max supported exponent is 2^10. If greater exponent is given, the function panics.
```
p0mvn, line 40:
```suggestion
// The max supported exponent is defined by the global maxSupportedExponent.
// If a greater exponent is given, the function panics.
```

osmomath/exp2_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/osmosis-labs/osmosis/v13/osmoutils"
+)
+
+var (
+	// minDecTolerance minimum tolerance for sdk.Dec, given its precision of 18.
+	minDecTolerance = sdk.MustNewDecFromStr("0.000000000000000001")
+)
+
+func TestExp2ChebyshevRationalApprox(t *testing.T) {
+
```
p0mvn, line 20:
```suggestion
```

osmomath/exp2_test.go
```diff
@@ -0,0 +1,298 @@
...
+			exponent:    osmomath.OneDec().Add(smallValue),
+			expectPanic: true,
+		},
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			osmoassert.ConditionalPanic(t, tc.expectPanic, func() {
+
```
p0mvn, line 155:
```suggestion
```

nicolaslara:
> Is it acceptable that we cannot guarantee rounding behavior?

I think as long as the error is within the an acceptable range (from the chart $[10^{-22}, 10^{-27}]$), then I think it doesn't matter if it's higher or lower than the actual results

p0mvn:
> > Is it acceptable that we cannot guarantee rounding behavior?
> 
> I think as long as the error is within the an acceptable range (from the chart [10−22,10−27]), then I think it doesn't matter if it's higher or lower than the actual results

There are some cases where the error is outside of the chart - `10^17`. For example:
https://github.com/osmosis-labs/osmosis/blob/79beaae72f54159a82c54a7872b6053fc8ffdc5a/osmomath/exp2_test.go#L112

I have not been able to explain why that is the case.

One guess is that approximated results in Go are being compared against WolframAlpha. The chart is constructed from a Python library where both the approximations and the actual results are using the same framework. I'm wondering if different computational platforms introduce additional errors.

Please let me know your thoughts.

p0mvn:
This is how the relative error plot looks like (very similar if not identical to the absolute):
Screenshot from 2022-12-15 14-53-00

From call, the concern was that relative and absolute plots should not look the same.

On a quick review, I do not see a bug in the way it is calculated. However, I'm going to investigate this better and get back with updates.

osmomath/exp2.go
```diff
@@ -0,0 +1,94 @@
...
+	// See scripts/approximations/main.py exponent_approximation_choice function for details.
+	denominatorCoefficients13Param = []BigDec{
+		OneDec(),
+		MustNewDecFromStr("0.341114724742545112949699755780593311").Neg(),
+		MustNewDecFromStr("0.052724071627342653404436933178482287"),
+		MustNewDecFromStr("0.004760950735524957576233524801866342").Neg(),
+		MustNewDecFromStr("0.000267168475410566529819971616894193"),
+		MustNewDecFromStr("0.000008923715368802211181557353097439").Neg(),
+		MustNewDecFromStr("0.000000140277233177373698516010555916"),
+	}
```
ValarDragon, line 28: Verified these were copied correctly

osmomath/exp2.go
```diff
@@ -0,0 +1,94 @@
...
+	p_x := denominatorCoefficients13Param[0].Clone()
+	x_exp_i := OneDec()
+	for i := 1; i < len(numeratorCoefficients13Param); i++ {
+		x_exp_i.MulMut(x)
+
+		h_x = h_x.Add(numeratorCoefficients13Param[i].Mul(x_exp_i))
+		p_x = p_x.Add(denominatorCoefficients13Param[i].Mul(x_exp_i))
+	}
+
+	return h_x.Quo(p_x)
```
ValarDragon, line 93: The Adds & Quo can be mutative!

osmomath/exp2_test.go
```diff
@@ -0,0 +1,296 @@
...
+)
+
+var (
+	// minDecTolerance minimum tolerance for sdk.Dec, given its precision of 18.
+	minDecTolerance = sdk.MustNewDecFromStr("0.000000000000000001")
+)
+
+func TestExp2ChebyshevRationalApprox(t *testing.T) {
+	smallValue := osmomath.MustNewDecFromStr("0.00001")
+	smallerValue := osmomath.MustNewDecFromStr("0.00000000000000000001")
```
ValarDragon, line 21: What is this smaller than & what is relation to minDecTolerance (sorry not sure if theres some important relation)

osmomath/exp2.go
```diff
@@ -0,0 +1,94 @@
...
+	if exponent.Abs().GT(maxSupportedExponent) {
+		panic(fmt.Sprintf("integer exponent %s is too large, max (%s)", exponent, maxSupportedExponent))
+	}
+	if exponent.IsNegative() {
+		panic(fmt.Sprintf("negative exponent %s is not supported", exponent))
+	}
+
+	integerExponentDec := exponent.TruncateDec()
+	integerExponent := integerExponentDec.TruncateInt()
+	integerResult := twoBigDec.PowerInteger(integerExponent.Uint64())
```
ValarDragon, line 53:
```suggestion
	// TODO: Switch the application of the integer result to just being a bit shift
	integerResult := twoBigDec.PowerInteger(integerExponent.Uint64())
```

ValarDragon:
Python and actual go code both LGTM, nice job! Still reviewing go tests

ValarDragon:
The plot difference LGTM

Were actually seeing what we'd expect to see between the absolute error vs relative error

The additive error on RHS side was larger than LHS. The relative error 'dampens' the RHS relative to the LHS, as exp only increasing on the range

osmomath/exp2_test.go
```diff
@@ -0,0 +1,296 @@
...
+	}{
+		"exp2(28.5)": {
+			exponent: osmomath.MustNewDecFromStr("28.5"),
+			// https://www.wolframalpha.com/input?i=2%5E%2828.5%29+45+digits
+			expectedResult: osmomath.MustNewDecFromStr("379625062.497006211556423566253288543343173698"),
+
+			errTolerance: osmoutils.ErrTolerance{
+				AdditiveTolerance:       minDecTolerance,
+				MultiplicativeTolerance: minDecTolerance,
+				RoundingDir:             osmomath.RoundUnconstrained, // TODO: confirm if this is acceptable.
```
ValarDragon, line 183: My answer all for all these "are they acceptable" is Yes!

Just need comments on the internal exp and overall exponent that

"The answer is correct up to a factor of 10^-22. Meaning that `actual = k * returned` for `k in [1 - 10^(-22), 1+ 10^22]`

ValarDragon: We can say 10^(-18) in godoc since thats what we test for, just add a note to ourself that we actually expect 10^22.

ValarDragon (approved):
Nice job!

osmomath/exp2.go
nicolaslara: This would mean we also have to manually do bounds checks, right?

ValarDragon: Shouldn't _have_ to, it should be implied by the max exponent check further up

osmomath/exp2.go
p0mvn: We don't yet have mutative version for `Add` and `Quo`, will make an issue

p0mvn: https://github.com/osmosis-labs/osmosis/issues/3773

osmomath/exp2_test.go
p0mvn: I used these to implement test cases closer to the boundaries.

The reason is that the errors tend to be large at the boundaries for non-Chebyshev approximations. This is just to make sure that, as expected, Chebyshev does not allow for larger errors close to boundaries. I will add a comment relaying this

osmomath/exp2.go
p0mvn: Great idea. Implemented it right away. Yes, this should be safe because of the check above

osmomath/exp2.go
```diff
@@ -48,16 +48,14 @@ func Exp2(exponent BigDec) BigDec {
...
 
-	fractionalExponent := exponent.Sub(integerExponentDec)
+	fractionalExponent := exponent.Sub(integerExponent)
 	fractionalResult := exp2ChebyshevRationalApprox(fractionalExponent)
 
-	result := integerResult.Mul(fractionalResult)
+	fractionalResult.i = fractionalResult.i.Lsh(fractionalResult.i, uint(integerExponent.TruncateInt().Uint64()))
 
-	return result
+	return fractionalResult
```
ValarDragon, line 58: Lets add a comment for the bitshift equivalency!

p0mvn: done

osmomath/exp2_test.go
p0mvn: godoc added

p0mvn:
> The plot difference LGTM
> 
> Were actually seeing what we'd expect to see between the absolute error vs relative error
> 
> The additive error on RHS side was larger than LHS. The relative error 'dampens' the RHS relative to the LHS, as exp only increasing on the range

That makes sense. Thanks for confirming

p0mvn:
All comments are now addressed. This should be ready for merge once CI passes.

PR #3709, opened 2022-12-14
ProtoRev: Trade Execution + Statistics

davidterpay:

This PR implements all of the logic surrounding finding the most profitable arbitrage opportunities given a set of routes, executing the most profitable arbitrage route, and storing profits in the module account. Additionally, the module tracks various statistics about how many trades have been executed, what pools had the most amount of arbitrage opportunities and more. Lastly, this PR refactors routes.go to use SwapAmountInRoutes instead of the custom route struct as it plugs into the gamm interface in a cleaner way.

Linking main issue for reference : https://github.com/osmosis-labs/osmosis/issues/3594.

## Brief Changelog
- *Calculating whether a cyclic arbitrage route is profitable*
- *Determining the most profitable cyclic arbitrage route from a set of routes given a swap*
- *Executing the most profitable cyclic arbitrage opportunity (if one exists)*
- *New setters, getters, and keys for statistics tracking*

## Testing and Verifying

- *Added unit tests for setters and getters for module statistics*
- *Added unit tests for FindMaxProfitRoute*
- *Added unit tests for CalculateRouteProfit*
- *Added unit tests for ExecuteTrade*
- *Added unit tests for ConvertProfits*

| NumberOfTrades | Tracks the number of trades protorev has executed | []byte{4} | []byte{numberOfTrades} | KV |
| ProfitsByDenom | Tracks the profits protorev has made | []byte{5} + []byte{tokenDenom} | []byte{sdk.Coin} | KV |
| TradesByRoute | Tracks the number of trades the module has executed on a given route | []byte{6} + []byte{route} | []byte{numberOfTrades} | KV |
| ProfitsByRoute | Tracks the profits the module has accumulated after trading on a given route | []byte{7} + []byte{route} | []byte{sdk.Coin} | KV |

**NumberOfTrades**
This will store the total number of arbitrage trades that `x/protorev` has executed since genesis. When finding a pool that has a cyclic arbitrage opportunity, the trade will be executed and this incremented.

**ProfitsByDenom**
`x/protorev` can execute cyclic arbitrage trades with either Osmosis or Atom. As such, this will store the profits that the module has earned from either denomination.

**TradesByRoute & ProfitsByRoute**
These metrics allows users and researchers to query the number of cyclic arbitrage trades that have been executed by `x/protorev` on an cyclic arbitrage route as well as all of the profits captured on that same route.

#### Trade Execution Logic
Now that we have a list of cyclic routes for each pool swapped by the user’s tx, we then determine if any of the routes are profitable. We determine this using a binary search algorithm that finds the amount of the asset to swap in that results in the most of that same asset out. We then calculate profits by taking the difference between the amount of asset out and amount of asset in. By iterating through the routes and storing the route, optimal input amount, and profit of the route with the highest profit > 0, we are left with the route and amount to execute the MultiHopSwap against.

The module mints the optimal input amount of the coin to swap in from the `bankkeeper` to the `x/protorev` module account, executes the MultiHopSwap by interacting with the `x/gamm` module, burns the optimal input amount of the coin minted to execute the MultiHopSwap, and sends subsequent profits to the module account.

- Note: `x/protorev` will exclusively execute MultiHopSwaps that originate and end in either OSMO or ATOM.

### DIFF
--- a/x/protorev/keeper/rebalance.go
+++ b/x/protorev/keeper/rebalance.go
@@ -11,7 +11,7 @@ import (
 // and returns the optimal route if there is one
 func (k Keeper) IterateRoutes(ctx sdk.Context, routes []gammtypes.SwapAmountInRoutes) (sdk.Coin, sdk.Int, gammtypes.SwapAmountInRoutes) {
 	var optimalRoute gammtypes.SwapAmountInRoutes
-	maxProfitInputCoin := sdk.NewCoin(types.OsmosisDenomination, sdk.ZeroInt())
+	var maxProfitInputCoin sdk.Coin
 	maxProfit := sdk.ZeroInt()
 
 	for _, route := range routes {
@@ -64,7 +64,8 @@ func (k Keeper) ConvertProfits(ctx sdk.Context, inputCoin sdk.Coin, profit sdk.I
 
 	// Calculate the amount of uosmo that we can get if we swapped the
 	// profited amount of the orignal asset through the highest uosmo liquidity pool
-	conversionTokenOut, err := conversionPool.CalcOutAmtGivenIn(ctx, sdk.NewCoins(sdk.NewCoin(inputCoin.Denom, profit)), types.OsmosisDenomination, conversionPool.GetSwapFee(ctx))
+	conversionTokenOut, err := conversionPool.CalcOutAmtGivenIn(ctx, sdk.NewCoins(sdk.NewCoin(inputCoin.Denom, profit)),
+		types.OsmosisDenomination, conversionPool.GetSwapFee(ctx))
 	if err != nil {
 		return profit, err
 	}
@@ -80,7 +81,7 @@ func (k Keeper) EstimateMultihopProfit(ctx sdk.Context, inputDenom string, amoun
 	tokenIn := sdk.NewCoin(inputDenom, amount)
 	amtOut, err := k.gammKeeper.MultihopEstimateOutGivenExactAmountIn(ctx, route, tokenIn)
 	if err != nil {
-		return sdk.NewCoin(types.OsmosisDenomination, sdk.ZeroInt()), sdk.ZeroInt(), err
+		return sdk.Coin{}, sdk.ZeroInt(), err
 	}
 	profit := amtOut.Sub(tokenIn.Amount)
 	return tokenIn, profit, nil
@@ -88,43 +89,41 @@ func (k Keeper) EstimateMultihopProfit(ctx sdk.Context, inputDenom string, amoun
 
 // FindMaxProfitRoute runs a binary search to find the max profit for a given route
 func (k Keeper) FindMaxProfitForRoute(ctx sdk.Context, route gammtypes.SwapAmountInRoutes, inputDenom string) (sdk.Coin, sdk.Int, error) {
-	left := 0
-	right := len(types.InputAmountList) - 1
-	midPosition := 0
+	var maxIterations = 20
+	var tokenIn sdk.Coin
+	var profit sdk.Int
+	currLeft := 0
+	currRight := len(types.InputAmountList) - 1
+	currMid := 0
 	iteration := 0
 
-	for left < right {
-		midPosition = (left + right) / 2
+	for currLeft < currRight && iteration < maxIterations {
+		iteration++
+
+		currMid = (currLeft + currRight) / 2
 
 		// Short circuit profit searching if there is an error in the GAMM module
-		_, profitMidPosition, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[midPosition], route)
+		tokenInMid, profitMid, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[currMid], route)
 		if err != nil {
-			return sdk.Coin{Denom: types.OsmosisDenomination, Amount: sdk.ZeroInt()}, sdk.ZeroInt(), err
+			return sdk.Coin{}, sdk.ZeroInt(), err
 		}
 
 		// Short circuit profit searching if there is an error in the GAMM module
-		_, profitMidPositionPlusOne, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[midPosition+1], route)
+		tokenInMidPlusOne, profitMidPlusOne, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[currMid+1], route)
 		if err != nil {
-			return sdk.Coin{Denom: types.OsmosisDenomination, Amount: sdk.ZeroInt()}, sdk.ZeroInt(), err
+			return sdk.Coin{}, sdk.ZeroInt(), err
 		}
 
 		// Reduce subspace to search for max profit
-		if profitMidPosition.LTE(profitMidPositionPlusOne) {
-			left = midPosition + 1
-		} else if profitMidPosition.GT(profitMidPositionPlusOne) {
-			right = midPosition
+		if profitMid.LTE(profitMidPlusOne) {
+			currLeft = currMid + 1
+			tokenIn = tokenInMidPlusOne
+			profit = profitMidPlusOne
+		} else {
+			currRight = currMid
+			tokenIn = tokenInMid
+			profit = profitMid
 		}
-
-		// Circuit breaker to prevent unbounded runtime
-		iteration += 1
-		if iteration >= 20 {
-			break
-		}
-	}
-
-	tokenIn, profit, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[left], route)
-	if err != nil {
-		return sdk.Coin{Denom: types.OsmosisDenomination, Amount: sdk.ZeroInt()}, sdk.ZeroInt(), err
 	}
 
 	return tokenIn, profit, nil
--- a/x/protorev/keeper/rebalance.go
+++ b/x/protorev/keeper/rebalance.go
@@ -92,28 +92,24 @@ func (k Keeper) FindMaxProfitForRoute(ctx sdk.Context, route gammtypes.SwapAmoun
 	tokenIn := sdk.Coin{}
 	profit := sdk.ZeroInt()
 
-	// 10 ^ 6 is multiplied to the input amount to convert to the base unit
-	million := sdk.NewInt(1_000_000)
 	curLeft := sdk.OneInt()
-	curMid := sdk.OneInt()
 	curRight := types.MaxInputAmount
-
 	iteration := 0
 
 	for curLeft.LT(curRight) && iteration < types.MaxIterations {
 		iteration++
 
-		curMid = (curLeft.Add(curRight)).Quo(sdk.NewInt(2))
+		curMid := (curLeft.Add(curRight)).Quo(sdk.NewInt(2))
 		curMidPlusOne := curMid.Add(sdk.OneInt())
 
 		// Short circuit profit searching if there is an error in the GAMM module
-		_, profitMid, err := k.EstimateMultihopProfit(ctx, inputDenom, curMid.Mul(million), route)
+		_, profitMid, err := k.EstimateMultihopProfit(ctx, inputDenom, curMid.Mul(types.StepSize), route)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


x/protorev/keeper/routes.go
```diff
@@ -62,125 +51,107 @@ func (k Keeper) BuildTokenPairRoutes(ctx sdk.Context, tokenIn, tokenOut string,
...
-	return k.BuildTradeInfoRoute(ctx, types.AtomDenomination, tokenIn, tokenOut, poolId, k.GetAtomPool)
+func (k Keeper) BuildAtomRoute(ctx sdk.Context, tokenIn, tokenOut string, poolId uint64) (gammtypes.SwapAmountInRoutes, error) {
+	return k.BuildRoute(ctx, types.AtomDenomination, tokenIn, tokenOut, poolId, k.GetAtomPool)
 }
 
-// BuildTradeInfoRoute constructs a cyclic arbitrage route that is starts/ends with swapDenom (atom or osmo) given the swap (tokenIn, tokenOut, poolId), and
+// BuildRoute constructs a cyclic arbitrage route that is starts/ends with swapDenom (atom or osmo) given the swap (tokenIn, tokenOut, poolId), and
 // a function that can get the poolId from the store given a (token, swapDenom) pair.
-func (k Keeper) BuildTradeInfoRoute(ctx sdk.Context, swapDenom, tokenIn, tokenOut string, poolId uint64, getPoolIDFromStore func(sdk.Context, string) (uint64, error)) (Route, error) {
+func (k Keeper) BuildRoute(ctx sdk.Context, swapDenom, tokenIn, tokenOut string, poolId uint64, getPoolIDFromStore func(sdk.Context, string) (uint64, error)) (gammtypes.SwapAmountInRoutes, error) {
```
bpiv400, line 112: Why do we pass the function in as a parameter instead of calling it from the context of BuildRoute?

bpiv400 (commented):
I didn't see the trade calculation logic

NotJeremyLiu:
> I didn't see the trade calculation logic

Just added rebalance.go and rebalance_test.go, has the trading calc logic

x/protorev/keeper/routes.go
davidterpay: we pass in a function because this way we can deduplicate the route creation for osmo and atom routes. The only difference between the two route generation methods is the way which pool Ids are retrieved (hence the `getPoolIDFromStore` param).

x/protorev/keeper/routes.go
```diff
@@ -9,20 +9,9 @@ import (
...
-	InputDenom  string
-	OutputDenom string
-	SwapFee     sdk.Dec
-	Pool        gammtypes.CFMMPoolI
-}
-
-type Route struct {
-	Trades []TradeInfo
-}
-
```
davidterpay, line 22: Per description in the PR, we deprecated the use of this struct in favor of `gammtypes.SwapAmountInRoutes` because we can directly plug this type into the gamm trading interface w/o additional code. As such, all routing logic is now plugging directly into the gammtype which reduces code complexity and improves general design. Outside of that, no logical changes were made to routes.go

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		// Circuit breaker to prevent unbounded runtime
+		iteration += 1
+		if iteration >= 20 {
+			break
+		}
+	}
+
+	tokenIn, profit, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[left], route)
```
bpiv400, line 125: This looks like some wasted computation

I think in all cases except when the iteration just stops because we've iterated 20 times already, this is just recalculating the profit that you've already calculated for midPositionPlusOne or midPosition (depending on whether profitMidPosition <= profitMidPositionPlusOne)

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+		// Short circuit profit searching if there is an error in the GAMM module
+		_, profitMidPositionPlusOne, err := k.EstimateMultihopProfit(ctx, inputDenom, types.InputAmountList[midPosition+1], route)
+		if err != nil {
+			return sdk.Coin{Denom: types.OsmosisDenomination, Amount: sdk.ZeroInt()}, sdk.ZeroInt(), err
+		}
+
+		// Reduce subspace to search for max profit
+		if profitMidPosition.LTE(profitMidPositionPlusOne) {
+			left = midPosition + 1
+		} else if profitMidPosition.GT(profitMidPositionPlusOne) {
```
bpiv400, line 114: just else? We should always make one of these updates

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// EstimateMultihopProfit estimates the profit for a given route
+// by estimating the amount out given the amount in for the first pool in the route
+// and then subtracting the amount in from the amount out to get the profit
+func (k Keeper) EstimateMultihopProfit(ctx sdk.Context, inputDenom string, amount sdk.Int, route gammtypes.SwapAmountInRoutes) (sdk.Coin, sdk.Int, error) {
+	tokenIn := sdk.NewCoin(inputDenom, amount)
+	amtOut, err := k.gammKeeper.MultihopEstimateOutGivenExactAmountIn(ctx, route, tokenIn)
+	if err != nil {
+		return sdk.NewCoin(types.OsmosisDenomination, sdk.ZeroInt()), sdk.ZeroInt(), err
```
bpiv400, line 83: Is 0 UOSMO the most sensible "null" value to return when erroring?

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Get the pool
+	conversionPool, err := k.gammKeeper.GetPoolAndPoke(ctx, conversionPoolID)
+	if err != nil {
+		return profit, err
+	}
+
+	// Calculate the amount of uosmo that we can get if we swapped the
+	// profited amount of the orignal asset through the highest uosmo liquidity pool
+	conversionTokenOut, err := conversionPool.CalcOutAmtGivenIn(ctx, sdk.NewCoins(sdk.NewCoin(inputCoin.Denom, profit)), types.OsmosisDenomination, conversionPool.GetSwapFee(ctx))
```
bpiv400, line 67: nit: break this into two lines

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+		// Reduce subspace to search for max profit
+		if profitMidPosition.LTE(profitMidPositionPlusOne) {
+			left = midPosition + 1
+		} else if profitMidPosition.GT(profitMidPositionPlusOne) {
+			right = midPosition
+		}
+
+		// Circuit breaker to prevent unbounded runtime
+		iteration += 1
+		if iteration >= 20 {
```
bpiv400, line 120: nit: This should be moved to the for loop: 

for left < right && iteration < 20

bpiv400: The "20" value here should be a variable and set somewhere that makes it more clear that it's a parameter and more accessible

bpiv400 (approved):

x/protorev/keeper/rebalance.go
NotJeremyLiu: Yeah let me make check a conditional here instead

NotJeremyLiu: Changed to set them with the profitMidPosition <= profitMidPositionPlusOne check

x/protorev/keeper/rebalance.go
NotJeremyLiu: With my latest commit I changed this to a variable at the top of the function. Should we think about making this a param that can be adjusted like hot routes? @davidterpay

NotJeremyLiu: The conclusion of this discussions is addressed in: 77d637131888dda0d65647784a5f71f95b970155

- Change from inputArray to InputMax
- Max iterations of 14
- Max OSMO/ATOM in 2 ^ 14

x/protorev/keeper/rebalance.go
stackman27: I think we can deduce this to sdk.Coin{} unless we want to specific osmo denom in error

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	gammtypes "github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+	"github.com/osmosis-labs/osmosis/v13/x/protorev/types"
+)
+
+// IterateRoutes checks the profitability of every single route that is passed in
+// and returns the optimal route if there is one
+func (k Keeper) IterateRoutes(ctx sdk.Context, routes []gammtypes.SwapAmountInRoutes) (sdk.Coin, sdk.Int, gammtypes.SwapAmountInRoutes) {
```
stackman27, line 12: Asking to learn: How does the system identifies routes? Do we have code for that logic yet?

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+	var optimalRoute gammtypes.SwapAmountInRoutes
+	maxProfitInputCoin := sdk.NewCoin(types.OsmosisDenomination, sdk.ZeroInt())
+	maxProfit := sdk.ZeroInt()
+
+	for _, route := range routes {
+		// Find the max profit for the route using the token out denom of the last pool in the route as the input token denom
+		inputCoin, profit, err := k.FindMaxProfitForRoute(ctx, route, route[2].TokenOutDenom)
+		if err != nil {
+			k.Logger(ctx).Error("Error finding max profit for route: ", err)
```
stackman27, line 21: Why not return err instead of logging error in context?

x/protorev/keeper/rebalance.go
```diff
@@ -0,0 +1,160 @@
...
+			continue
+		}
+
+		// Filter out routes that don't have any profit
+		if profit.LTE(sdk.ZeroInt()) {
+			continue
+		}
+
+		// If arb doesn't start and end with uosmo, then we convert the profit to uosmo, and compare profits in terms of uosmo
+		if inputCoin.Denom != types.OsmosisDenomination {
```
stackman27, line 31: Asking to learn: If the arb profit is really high wouldn't we create another arb position in balanced uosmo pool while converting? Is this expected?

stackman27 (commented):
Left few comments and questions. Still looking at tests :))

x/protorev/keeper/rebalance.go
NotJeremyLiu: @stackman27 So this logic isn't actually swapping the asset to uosmo (which could create an arb opportunity). This logic just called CalcOutAmtGivenIn and sees what the converted profit would be if it was osmo-denominated, and chooses the highest route/tokenIn that gives the most osmo-denominated profits (but still in the original denom)

x/protorev/keeper/rebalance.go
NotJeremyLiu: @stackman27 We wanted to both 1) be aware of any errors that occur when iterating through routes, but 2) if one route results in an error, not stop the loop of checking other routes that may be profitable.

We went with this logger / continue method to achieve the above. Do you have a better way to achieve the above / reasons why what we wanted to achieve is not ideal?

x/protorev/keeper/rebalance.go
NotJeremyLiu: @stackman27 We have two methods of creating routes, a highest liquidity method and a hot routes method.

Our ReadMe draft explains the methodology in the state transitions section: https://skip-protocol.notion.site/x-protorev-Spec-3ba5cb368f314120bc514eb2d0a6a0dd

The relevant code pushed for this is in epoch_hook.go and routes.go.

Highest liquidity methods finds the highest liquidity atom pools and osmo pools and uses them to be the first and last hop of a cycle, where we place the pool the user swapped in the middle of the cycle.

The hot routes method is uploaded by us based on external analysis of routes searchers use / have arb opportunities often.

stackman27: Gotchu this makes sense

PR #3711, opened 2022-12-14
refactor(osmoutils): use Dec for additive tolerance instead of Int

p0mvn:
Progress towards: #3540 

Suggesting to change additive tolerance to a Decimal to improve the experience when testing math functions that require high precision.


czarcas7ic (approved):

ValarDragon:
Needs to be backported to work w/ v13.x having osmomath in its own go.mod

PR #3712, opened 2022-12-14
refactor(osmomath): replace Power with PowerInteger

p0mvn:
Progress towards: #3540 

I missed the existence of `Power` when adding `PowerInteger`. As a result, these functions are identical.

I propose removing the old `Power` and replace all calls with `PowerInteger` since I am adding a new `Power` function that takes a decimal exponent.

**Note:**
- The removed `Power` is identical with `PowerInteger`
- The existing `PowerInteger` tests include and cover all removed test cases.


ValarDragon:
This LGTM, but it should be state compatible and backportable?

p0mvn:
> This LGTM, but it should be state compatible and backportable?

Agreed. Sorry I confused it with this `Pow` that is used in cfmm and erred on the side of caution:
https://github.com/osmosis-labs/osmosis/blob/d69dd69f0251c12077ca2bda8f085e389d58c9fd/osmomath/math.go#L57

p0mvn:
Labels are now changed

ValarDragon (approved):
LGTM!

PR #3719, opened 2022-12-14
Removing the second check as it breaks the yml file

nicolaslara:

Get the contract lints to run

## Brief Changelog

Removing the diff matching from the lints

## Testing and Verifying


.github/workflows/contracts.yml
```diff
@@ -25,9 +25,8 @@ jobs:
         with:
           PATTERNS: |
             **/**.rs
-            **/**.go
-            go.mod
```
ValarDragon, line 29: We should still have these two lines right?

nicolaslara: do we want to run contract tests if go changes? These aren't interacting with the chain at all

nicolaslara:
The "if" error makes sense now! see: [2d25c15](https://github.com/osmosis-labs/osmosis/pull/3719/commits/2d25c155c8a3a4fb036716b1f0dbc976fa78ebc8)

ValarDragon:
OHHHH

that makes sense now :sweat_smile:

.github/workflows/contracts.yml
ValarDragon: Go tests can via a query, no?

.github/workflows/contracts.yml
```diff
@@ -109,10 +87,6 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-      - uses: technote-space/get-diff-action@v6.0.1
-        with:
-          PATTERNS: |
-            **/**.rs
```
ValarDragon, line 115: Going to bring this back in subsequent commit

ValarDragon (approved):
TY!

.github/workflows/contracts.yml
nicolaslara: I don't think so. The rust tests are fetching the chain from git, so the version is specified: osmosis-testing-13.1.2 uses osmosis-13.1.1, for example. So even if "main" changes, the underlying rust test will not behave differently.

Now, this may not be what we want. We may want to be able to test the contracts against osmosis main so we are warned that we need to update them if something changes. It almost feels like we would want to test with both the version used by the contract *and* main.

@iboss-ptk Does it make sense to have a release of osmosis-testing that always points to main so we can do this?

Issue #3443, opened 2022-11-17
test(x/twap): split up TestComputeArithmeticTwap_ThreeAsset into arithmetic and geometric


p0mvn:
## Background

With the introduction of geometric twap, we split `computeArithmeticTWAP` into:
- `computeTWAP`
- `computeArithmeticTwap`
- `computeGeometricTwap`

`TestComputeArithmeticTwap_ThreeAsset` needs to be refactored to test all kinds of twaps and the 3 above functions

## Acceptance Criteria
- refactor the test to test both arithmetic and geometric twap whilte testing `computeTwap` function
- create a similar test for `computeArithmeticTwap`
- create a similar test for `computeGeometricTwap`
- reuse test cases and minimize redundancy while maximizing coverage

PR #3726, opened 2022-12-14
test(x/twap): 3-asset geometric logic tests

p0mvn:
Closes: #3443

Adding 3-asset geometric twap logic tests as per issues

### DIFF
--- a/x/twap/keeper_test.go
+++ b/x/twap/keeper_test.go
@@ -583,6 +583,7 @@ func newThreeAssetOneSidedRecord(time time.Time, accum sdk.Dec, useP0 bool) []ty
 	} else {
 		record.P1ArithmeticTwapAccumulator = accum
 	}
+	record.GeometricTwapAccumulator = accum
 	record.P0LastSpotPrice = sdk.ZeroDec()
 	record.P1LastSpotPrice = sdk.OneDec()
 	records := []types.TwapRecord{record, record, record}
--- a/x/twap/logic_test.go
+++ b/x/twap/logic_test.go
@@ -745,16 +745,6 @@ func TestComputeGeometricTwap(t *testing.T) {
 }
 
 func TestComputeArithmeticTwap_ThreeAsset_Arithmetic(t *testing.T) {
-	testThreeAssetCaseFromDeltas := func(startAccum, accumDiff sdk.Dec, timeDelta time.Duration, expectedTwap sdk.Dec) computeThreeAssetArithmeticTwapTestCase {
-		return computeThreeAssetArithmeticTwapTestCase{
-			newThreeAssetOneSidedRecord(baseTime, startAccum, true),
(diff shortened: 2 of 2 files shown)


x/twap/logic_test.go
```diff
@@ -1558,6 +1592,33 @@ func (s *TestSuite) TestAfterCreatePool() {
 	}
 }
 
+func (s *TestSuite) TestTwapLog() {
```
p0mvn, line 1595: Note: these are moved to group tests at the top of the file and helpers at the bottom. These are unchanged

x/twap/keeper_test.go
```diff
@@ -583,6 +583,7 @@ func newThreeAssetOneSidedRecord(time time.Time, accum sdk.Dec, useP0 bool) []ty
 	} else {
 		record.P1ArithmeticTwapAccumulator = accum
 	}
+	record.GeometricTwapAccumulator = accum
```
ValarDragon, line 586: odd that its the same accumulator set for arithmetic and geometric ones, but w/e

x/twap/logic_test.go
```diff
@@ -791,6 +780,51 @@ func TestComputeArithmeticTwap_ThreeAsset(t *testing.T) {
...
 
+func TestComputeArithmeticTwap_ThreeAsset_Geometric(t *testing.T) {
+	var (
+		five        = sdk.NewDec(5)
+		fiveFor3Sec = OneSec.MulInt64(3).Mul(twap.TwapLog(five))
+
+		ten          = five.MulInt64(2)
+		tenFor100Sec = OneSec.MulInt64(100).Mul(twap.TwapLog(ten))
+
+		errTolerance = sdk.MustNewDecFromStr("0.00000001")
```
ValarDragon, line 791: How is this chosen?

x/twap/logic_test.go
ValarDragon: Ok... , don't think this is an important code style thing though

x/twap/logic_test.go
ValarDragon: Can we make an issue to revisit this error tolerance in the lens of stableswap with high scaling factors?

ValarDragon (approved):
LGTM, but need to revisit the precision pre-release

x/twap/logic_test.go
p0mvn: This was chosen as the tightest allowed precision with the current `Pow` function. Should be possible to make tighter once the new `Power` function is merged.

Will make an issue

p0mvn: https://github.com/osmosis-labs/osmosis/issues/3770

p0mvn:
Issue created: https://github.com/osmosis-labs/osmosis/issues/3770

Going to merge with one approval since this is a test and < 100 LOC

Issue #3542, opened 2022-11-27
feat(app/upgrades): implement twap record migrations for geometric logic


p0mvn:
## Background

The existing twap records should be migrated in the upgrade handler so that the geometric twap accumulator is initialized to 0 for each record.

## Suggested Design

Add logic in v14 upgrade handler, iterating over all possible twap records and setting the geometric twap accumulator to 0

## Acceptance Criteria
- migrations added
- unit tested
- e2e test passses

hieuvubk:
@p0mvn Hi its blocked by pr [#3420 ](https://github.com/osmosis-labs/osmosis/pull/3420) right?

p0mvn:
@hieuvubk yessir. Just merged #3420 . This is also blocked on the v14 upgrade handler being present.

We have a CI job to generate it automatically. We would like to test it instead of adding the upgrade handler manually. However, the CI job is currently being triggered by the release:
https://github.com/osmosis-labs/osmosis/blob/f364fe4bbef43236acd13980d8794bfb4eeee19e/.github/workflows/auto-update-upgrade.yml#L8-L10

There is an issue to make it also run manually: https://github.com/osmosis-labs/osmosis/issues/3495.

If you have the capacity and interest in tacking #3495 to CI-generate the v14 upgrade handler and then adding the migrations described here, that would be amazing. Please let me know if you're willing to pick up #3495 and then this.

MasterPi-2124:
@p0mvn I am doing auto-generating upgrade handler on manual inputs from user as suggested on #3495. Here is my [workflow](https://github.com/MasterPi-2124/osmosis/actions/workflows/manually-update-upgrade.yml) with examples of generating `v14` upgrade handler, which leads to this [pull request](https://github.com/MasterPi-2124/osmosis/pull/2) and [result](https://github.com/MasterPi-2124/osmosis/commit/a0cfecafd5ea759bd241a9fff9d3f3cda13b986e).

MasterPi-2124:
About the release trigger, the workflow will be automatically running, and it will automatically take release tag as input. So you will only have to specific version on manual trigger.

I made this PR request: https://github.com/osmosis-labs/osmosis/pull/3577

Issue #3728, opened 2022-12-14
test(e2e): basic geometric twap test


p0mvn:
## Background

It would be nice to have a basic twap e2e test to ensure that the API works as expected and that the migrations are done correctly.

## Suggested Design

1. Create a pool with `1_000_000` `uosmo`, pool 2 with `2_000_000` `uatom`.

2. Query geometric twap. Base = uosmo, result = 2

3. Swap `1_000_000` `uatom` for `uosmo`.

4. Query geometric twap to now with start time after swap. Base = uosmo. Assert the result < 2

## Acceptance Criteria
- e2e test implemented per suggestion
- passes

p0mvn:
Blocked on #3514

PR #3729, opened 2022-12-14
feat(app/upgrades): twap record migration

p0mvn:
Closes: #3542

Migrating all existing twap records to initialize their geometric accumulator to 0.


p0mvn:
Upon more testing, I realized that the new fields are getting initialized to zero by default so migrations aren't needed.

I'm going to add a marshal/unmarshal test in a separate PR. Also, #3728 will help confirm that this is valid

PR #3735, opened 2022-12-15
[CL] (WIP) feat: Range Order

mattverse:
This PR includes feature of range orders for concentrated liquidity.

For now implementation is done in the following way:
- create separate message for range orders `RangePosition`
- When createRangePosition message is called from the msg server, normal `createPosition` process is done, followed by additional action of adding one more state entry for saving range positions
- Every time a swap has happened, we iterate over range positions to see if there are positions that needs to be closed:

**_this is the part I'm mostly concerned about, especially about overhead on calculation every swap. I'm seeing that this can be optimized in either two ways: 1. use better key indexing to minimize iteration.(I'm wondering if we can get computation to log by doing this?) 2. create a hook that gets triggered every time the pool current tick changes. This way closing range positions does not have to happen every swap, but only when current tick of the pool changes. Reviewrs, please let me know if these two options make sense or if there are alternatives for this_**
- If there are, we close position using same method as `withdrawPosition`

### DIFF
--- a/x/concentrated-liquidity/position.go
+++ b/x/concentrated-liquidity/position.go
@@ -2,6 +2,7 @@ package concentrated_liquidity
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/gogo/protobuf/proto"
 
 	"github.com/osmosis-labs/osmosis/v13/osmoutils"
 	"github.com/osmosis-labs/osmosis/v13/x/concentrated-liquidity/model"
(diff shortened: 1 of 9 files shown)


x/concentrated-liquidity/position.go
```diff
@@ -94,3 +95,54 @@ func (k Keeper) setPosition(ctx sdk.Context,
...
+		return err
+	}
+
+	currentTick := pool.GetCurrentTick()
+
+	iterator := k.RangePositionIterator(ctx, poolId)
+	defer iterator.Close()
+
+	// iterate over all range positions and withdraw expired positions
+	for ; iterator.Valid(); iterator.Next() {
```
p0mvn, line 129: These are the position iterations we are trying to avoid.

We seem to have 2 options:
- figure out the accumulator design Dev mentioned
- Implement a Uni-v3 style range order as our v1. This range order does not expire positions, making it only possible to finalize an order with a separate message

czarcas7ic: I am in favor of the latter for MVP, a user must pull out their position on their own instead of being automatically triggered.

For sure interested in working on the accumulator method for a future release, but dont think we should block on it (there are enough unknowns and testing left to keep us busy as it is imo)

PR #3737, opened 2022-12-15
Fix: Support HTTP requests for `FilteredPools`

mattverse:
cref: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto

Current `FilteredPools` does not support http requests, as grpc gateway cannot do `GET` requests for non primitive types that are repeated. 

This PR brings the solution using coins string, and then parsing the coins string manually with in the grp_query.go logic. 

## Brief Changelog
- Change `FilteredPools` `MinLiquidity` field from sdk.Coins struct to string

## Testing and Verifying
Passes existing test on the grpc method, also added new test cases for the helper method.


x/gamm/client/cli/query.go
```diff
@@ -286,21 +283,12 @@ $ %s query gamm pools-with-filter <min_liquidity> <pool_type>
...
-				min_liquidity = coins
+				min_liquidity = args[0]
 			} else {
-				coins, err := sdk.ParseCoinsNormalized(args[0])
-				if err != nil {
-					return status.Errorf(codes.InvalidArgument, "invalid token: %s", err.Error())
-				}
-
-				min_liquidity = coins
+				min_liquidity = args[0]
```
nicolaslara, line 291: does it make more sense to do:
```
			min_liquidity := args[0]
			var pool_type string
			if len(args) > 1 {
				pool_type = args[1]
			}
```

nicolaslara (approved):

p0mvn (approved):

jonator:
Do you have an example of how the query should be serialized?

mattverse:
@jonator are you asking for what an example request would look like? If so, it would look something like this
```
{
"min_liquidity": "100uosmo, 300uatom",
"pool_type": "Balancer"
}
```

ValarDragon (approved):
LGTM but needs API break changelog update

ValarDragon:
Changed labels, this is query breaking, but state compatible, as its not a query exposed to state machine. (If it was we couldn't backport)

jonator:
> @jonator are you asking for what an example request would look like? If so, it would look something like this
> 
> ```
> {
> "min_liquidity": "100uosmo, 300uatom",
> "pool_type": "Balancer"
> }
> ```

No I meant something like:
https://lcd-osmosis.keplr.app/osmosis/gamm/v1beta1/filtered_pools?pagination.limit=100&pooltype=stableswap

We can't send a body in a GET request

mattverse:
@jonator Ah I see, it could be done using something like https://lcd-osmosis.keplr.app/osmosis/gamm/v1beta1/filtered_pools?pagination.limit=100&min_liquidity=100uosmo&pool_type=Balancer

PR #3702, opened 2022-12-13
Swaprouter create pool logic-only migration

ValarDragon:
## What is the purpose of the change

Moves CreatePool logic from inside of gamm, to inside of swaprouter, maintaining the Gamm entrypoint. This as a component required wiring swaprouter.

Remaining work from #3638 , left as future TODO's:

* Add v14 upgrade logic for adding existing routes for gamm pools
* Migrate gamm pool creation fee to swaprouter (generally migrate the params)
* Remove Gamm CreatePool entrypoint
* Remove gamm pool ID tracking entirely. (Mostly a many-file refactor)
* Move the PoolCreation hook from gamm into swaprouter

## Brief Changelog

- Wire swaprouter into app
- 

## Testing and Verifying

This change is already covered by the existing gamm tests

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #3738, opened 2022-12-15
refactor(x/swaprouter): use fee from swaprouter's params

p0mvn:
Closes: #XXX

Earlier in #3702 , we performed swaprouter params migration. However, we kept using a custom x/gamm function to get the fee.

This PR changes the `CreatePool` logic to get the fee directly from `x/swaprouter` params.

## Testing

- `TestPoolCreationFee` has been moved to swaprouter for a better location. Passes


x/swaprouter/module/module.go
```diff
@@ -142,7 +143,7 @@ func (am AppModule) SimulatorGenesisState(simState *module.SimulationState, s *s
 	swaprouterGen := types.DefaultGenesis()
 	// change the pool creation fee denom from uosmo to stake
 	// TODO: uncomment this once simulation is enabled.
-	// swaprouterGen.Params.PoolCreationFee = sdk.NewCoins(swaproutersimulation.PoolCreationFee)
+	swaprouterGen.Params.PoolCreationFee = sdk.NewCoins(gammsimulation.PoolCreationFee)
```
czarcas7ic, line 146: Shouldn't this now read "swaproutersimulation.PoolCreationFee" since we are no longer reading it from gamm? If so, we should remove whatever var or const this is pointing to from gamm right?

Also, the above TODO can be removed now I believe.

x/swaprouter/types/expected_keepers.go
```diff
@@ -1,14 +0,0 @@
-package types
-
-import (
-	sdk "github.com/cosmos/cosmos-sdk/types"
-)
-
-// GammKeeper defines the expected interface needed for swaprouter module
-type GammKeeper interface {
-	SwapI
```
czarcas7ic, line 9: Oh, do we not need to specify SwapI in GammKeeper within expected_keepers anymore?

czarcas7ic (approved):

x/swaprouter/module/module.go
p0mvn: Removing TODO.

I will add package change task as a TODO in: https://github.com/osmosis-labs/osmosis/issues/3556

Currently, there is no swaprouter simulation package yet

x/swaprouter/types/expected_keepers.go
p0mvn: Gamm keeper implements `SwapI`, it is specified. The `GammKeeper` interface only existed to extend `SwapI`, adding a `GetPoolCreationFee`. This method is now removed so we can remove the interface extension too

ValarDragon (approved):

PR #3746, opened 2022-12-16
Make apply func if no err, out of gas panic, match std behavior

ValarDragon:
Closes: #2487

So it turns out that CacheCtx and Ctx bizarrely share the same gas meter. Source: https://github.com/cosmos/cosmos-sdk/blob/main/types/context.go#L289-L303 , and that GasMeter is a pointer. (Otherwise https://github.com/cosmos/cosmos-sdk/blob/main/types/context.go#L58 ctx,GasMeter().ConsumeGas() wouldn't work)

So this PR makes it so that if in apply func if no err block fails due to out of gas, it preserves the standard out of gas behavior flow. This is not an issue for beginblock / endblock code as they should not get panics for out of gas issues. Whereas for txs, right now you get a different error code / distinct flow right now.

This ideally would be handled upstream, but alas.

This is state breaking, as it effects tx execution codes.

## Brief Changelog

* [#3746](https://github.com/osmosis-labs/osmosis/pull/3746) Make ApplyFuncIfNoErr logic preserve panics for OutOfGas behavior.

## Testing and Verifying

This change added a test which shows:
- When gas limit not hit, its being updated correctly
- When we out of gas panic, we are:
  - still panicing with the expected value
  - underlying gas meter still updated

The QA requirement for this will be to test a superfluid delegate action with and without this change, in the out of gas behavior setting. (Namely that they should both return an error, just a different error code)

  - How is the feature or change documented? Code comment


osmoutils/cache_ctx.go
```diff
@@ -35,6 +43,19 @@ func ApplyFuncIfNoError(ctx sdk.Context, f func(ctx sdk.Context) error) (err err
...
 }
 
+// Frustratingly, this has to return the error descriptor, not an actual error itself
+// because the SDK errors here are not actually errors. (They don't implement error interface)
+func IsOutOfGasError(err any) (bool, string) {
+	switch e := err.(type) {
+	case types.ErrorOutOfGas:
+		return true, e.Descriptor
+	case types.ErrorGasOverflow:
+		return true, e.Descriptor
```
czarcas7ic, line 53: What about types.ErrorNegativeGasConsumed?

czarcas7ic (approved):
LGTM!

osmoutils/cache_ctx.go
ValarDragon: oh wow didn't know that existed, let me think about it

mattverse (approved):
LGTM once @czarcas7ic 's suggestion of negative error has been added!

osmoutils/cache_ctx.go
nicolaslara: hmmmmm, if only those were sum types and we could ensure the the whole space is covered... 🤔 (rust)

PR #3747, opened 2022-12-16
Make wasm hooks importable

nicolaslara:
This is a draft PR for experimenting with making wasm hooks importable. 

I had to copy a few functions from osmoutils, which makes me think we should consider also making osmoutils its own package.

As an alternative, we could use workspaces (https://go.dev/blog/get-familiar-with-workspaces) to avoid having to deal with the scattered go.mod/sum files.


nicolaslara:
> Can we make a separate PR for just moving the test folder, and get that merged in first?

Done: https://github.com/osmosis-labs/osmosis/pull/3803

nicolaslara:
For reference, this is how to get the version string if we want to avoid the replaces:

```
TZ=UTC git --no-pager show \
  --quiet \
  --abbrev=12 \
  --date='format-local:%Y%m%d%H%M%S' \
  --format="%cd-%h"
```

nicolaslara:
There are too many conflicts here. Closing in favor of #3850

PR #3766, opened 2022-12-16
refactor(wasmbindings): remove gamm and twap bindings

p0mvn:
Closes: #XXX

Removing bindings gamm and twap. These have been causing additional overhead during refactors while not having production use cases.

In production, everyone should be using `osmosis-std` instead.

Bank and tokenfactory bindings have use cases outside of Osmosis so these are kept for now.

An announcement about removal is to be made post-merge.

## Testing and Verifying


app/keepers/keepers.go
```diff
@@ -369,7 +369,6 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 	// if we want to allow any custom callbacks
 	supportedFeatures := "iterator,staking,stargate,osmosis,cosmwasm_1_1"
 
-	wasmOpts = append(owasm.RegisterCustomPlugins(appKeepers.GAMMKeeper, appKeepers.BankKeeper, appKeepers.TwapKeeper, appKeepers.TokenFactoryKeeper), wasmOpts...)
```
ValarDragon, line 372: bank definitely still needs to be there. TokenFactory is moving to general bindings as well, so I think we should keep that for now.

ACK for gamm + twap

nicolaslara (approved):
LGTM

p0mvn:
Planning to merge. No one has gotten back to me about using these on Telegram or Discord

czarcas7ic (approved):
ACK with the understanding that integrators will be without wasmbindings for swap messages if we do not implement the swaprouter wasmbindings by the next release

p0mvn:
The only concern raised by the integrators is having the following queries whitelisted:
https://github.com/osmosis-labs/osmosis/blob/a5bc6077a60f825ee95cec9be1f442d66a04cbfa/wasmbinding/stargate_whitelist.go#L87-L88

As a result, this should be good to be merged

PR #3767, opened 2022-12-16
test(x/swaprouter): TestCreatePool (merge)

p0mvn:
Closes: #XXX

Merging test for swaprouter's `CreatePool`

### DIFF
--- a/x/swaprouter/create_pool_test.go
+++ b/x/swaprouter/create_pool_test.go
@@ -2,14 +2,15 @@ package swaprouter_test
 
 import (
 	"fmt"
+	"reflect"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/osmosis-labs/osmosis/v13/app/apptesting"
 	"github.com/osmosis-labs/osmosis/v13/x/gamm/pool-models/balancer"
 	balancertypes "github.com/osmosis-labs/osmosis/v13/x/gamm/pool-models/balancer"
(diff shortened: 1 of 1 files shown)


x/swaprouter/create_pool_test.go
```diff
@@ -62,7 +63,7 @@ func (suite *KeeperTestSuite) TestPoolCreationFee() {
 		swaprouterKeeper := suite.App.SwapRouterKeeper
 
 		// set pool creation fee
-		swaprouterKeeper.SetParams(suite.Ctx, swaproutertypes.Params{
+		swaprouterKeeper.SetParams(suite.Ctx, types.Params{
```
czarcas7ic, line 66: semi-random aside that doesn't need to be actioned:

I feel like we should have a naming convention for types. IMO I like to call types imports by their name, even the types package of the module I am currently in. The only time I might not do that is if there is only one types imported. What do you think the standard should be?

czarcas7ic (approved):
Looks good 👍

x/swaprouter/create_pool_test.go
p0mvn: My personal convention has been just `types` within a module. E.g., if I'm in `swaprouter`, then swaprotuer's `types` would be `types` but `gamm`'s would be `gammtypes`.

The reason is twofold:
- SDK modules already follow this convention
- Easy to see whether I'm using types of the current module I'm in or an outside module

Happy to discuss separately to come to a consensus

p0mvn:
Merging since this is a test

PR #3768, opened 2022-12-16
Remove osmomath's dependency on osmoassert & gammtypes

ValarDragon:

Needed to split osmomath into its own go.mod

This basically involved making an exported var for `gammtypes.MaxSpotPrice`, and removing Conditional Panic cases. Only affects test files.

### DIFF
--- a/osmomath/export_test.go
+++ b/osmomath/export_test.go
@@ -1,6 +1,11 @@
 package osmomath
 
-import sdk "github.com/cosmos/cosmos-sdk/types"
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
 
 var (
 	EulersNumber = eulersNumber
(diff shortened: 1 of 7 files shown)


osmomath/export_test.go
```diff
@@ -1,6 +1,14 @@
...
 var (
 	EulersNumber = eulersNumber
 	TwoBigDec    = twoBigDec
 )
+
+// 2^128 - 1, needs to be the same as gammtypes.MaxSpotPrice
+// but we can't directly import that due to import cycles.
+// Hence we use the same var name, in hopes that if any change there happens,
+// this is caught via a CTRL+F
+var MaxSpotPrice = sdk.NewDec(2).Power(128).Sub(sdk.OneDec())
```
p0mvn, line 14: Would having separate constants or utils package help mitigate the import cycles?

osmomath/sigfig_round_test.go
```diff
@@ -77,16 +75,21 @@ func TestSigFigRound(t *testing.T) {
...
+				actualResult := SigFigRound(tc.decimal, tc.tenToSigFig)
 				require.Equal(
 					t,
 					tc.expectedResult,
 					actualResult,
 					fmt.Sprintf("test %d failed: expected value & actual value should be equal", i),
 				)
-			})
+			}
+			if expPanic {
```
p0mvn, line 88: Can we make an osmomath helper for this? Seems frequent enough to justify a helper

p0mvn (approved):
LGTM

osmomath/export_test.go
ValarDragon: It would have to be in its own independent go.mod, which seems pretty annoying for now

osmomath/sigfig_round_test.go
ValarDragon: good point, I'll just add it in export_test.go for now

Issue #3744, opened 2022-12-16
osmomath: move to its own go.mod


ValarDragon:
## Background

An issue to track that we should move osmomath to its own go.mod. Precursors / steps in this:

- [x] Move binary search from osmoutils to osmomath
- [x] Remove osmoassert dependencies
- [x] Wait for the power PRs to stabilize from Geometric TWAP (Currently blocked)
- [ ] Move osmomath into its own go.mod, that can be externally imported. Change 'main' go.mod to use this via local import.

PR #3771, opened 2022-12-16
Move osmomath into its own go.mod

ValarDragon:
Closes: #3744 (after we get a tag for this folder, that comes after this is merged)

Move osmomath into its own go.mod, but all our code set to use latest code in the folder.

## Brief Changelog

- Move osmomath into its own go.mod. To update, change your osmomath import to `github.com/osmosis-labs/osmosis/v13/osmomath` to `github.com/osmosis-labs/osmosis/osmomath`

## Testing and Verifying

All tests still pass


Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN apk add --no-cache \
 
 # Download go dependencies
 WORKDIR /osmosis
-COPY go.mod go.sum ./
```
ValarDragon, line 22: cc @niccoloraspa @p0mvn is there some more correct way I should be doing this?

AFAIU, we may actually be building against the wrong osmomath rn, since we can cache the osmomath copy. 

We need it here though, since go mod download looks for a local osmomath replace

osmomath/go.mod
```diff
@@ -0,0 +1,89 @@
+module github.com/osmosis-labs/osmosis/osmomath
```
p0mvn, line 1: Can we suffix this with a version?

I think not having a version might contribute to import cycles and other import problems in the future

Dockerfile
```diff
@@ -19,7 +19,8 @@ RUN apk add --no-cache \
...
+COPY go.work go.mod go.sum ./
+COPY osmomath/ ./osmomath
```
p0mvn, line 23: Is it possible to only copy `osmomath/go.mod` and `osmomath/go.sum` here?

This is so that updates to `osmomath` code do not trigger wget for wasvm. Having only `go.mod` will allow that layer to be cached

Dockerfile
p0mvn: Commented below with suggestions, rest LGTM on a first pass

osmomath/go.mod
ValarDragon: This is not 1.0 yet. We only need version suffixes once theres a 2.0

Dockerfile
ValarDragon: Let me try this, but I believe the answer is no. This is because the main go.mod replaces it with a local folder path, rather than downloading this folder from a normal upstream mirror.

Though we could move all these copies until after the wasmvm wget, I don't understand why we do the go mod download before the wget

osmomath/go.mod
nicolaslara: This opens up a discussion of the semantics of versioning for these submodules. What sort of changes are needed for a v2? is it any api breaking changes?

I think this depends on wether we have external users or not, but I'm almost leaning towards that projects importing this should target a specific commit in their go.mod. That way we don't have to worry about breaking changes for external users.

nicolaslara (approved):
Agree with moving the COPYs on the docker file and left a comment about versioning. But otherwise this LGTM

nicolaslara:
Note/warning: when main gets merged into this branch, you'll have issues with the Exp2 tests using osmoassert.

This will be solved by making osmoutils its own module too

ValarDragon:
Ok I tagged osmomath, and switched `make build`, `make install` and `dockerfile` to not use the go work, and instead depend on either tagged or hashpinned versions.

ValarDragon:
Should now work per the discussed plan! Going to merge

PR #3764, opened 2022-12-16
refactor(x/gamm, x/swaprouter): multihop and estimate swap move

p0mvn:
Closes: #XXX

## What is the purpose of the change

This PR moves multihop and estimate swaps keeper methods from x/gamm to x/swaprouter.

No messages or queries are broken.

Old messages and queries relying on the x/gamm keeper methods now directly call the respective methods in `x/swaprouter`.

## Testing and Verifying

All pre-existing tests have been moved to `x/swaprouter`. New tests added where applicable.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? yes
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

PR #3772, opened 2022-12-16
refactor(app/apptesting): PrepareBalancerPool helpers (simplifying swaprouter merge PR)

p0mvn:
Closes: #XXX

https://github.com/osmosis-labs/osmosis/pull/3764 needs `PrepareBalancerPool` helper that takes coins and pool params.

This PR introduces these helpers and slightly cleans them up.

- Introduced `PrapareCustomBalancerPool` that takes pool assets and pool params
- Introduced `PrepareCustomBalancerPoolFromCoins` that takes coins and pool params. Each coin is converted to asset with a weight of 1
- Removed `PrepareBalancerPoolWithPoolAssets` that takes pool assets only. There was only one use case. Replaced with `PrepareCustomBalancerPool instead
- Reused all existing helpers where applicable to reduce code duplication

For example, `PrepareBalancerPoolWithPoolAssets` was

### DIFF
--- a/app/apptesting/gamm.go
+++ b/app/apptesting/gamm.go
@@ -71,7 +71,10 @@ func (s *KeeperTestHelper) PrepareBalancerPoolWithCoinsAndWeights(coins sdk.Coin
 		poolAssets = append(poolAssets, poolAsset)
 	}
 
-	return s.PrepareBalancerPoolWithPoolAsset(poolAssets)
+	return s.PrepareCustomBalancerPool(poolAssets, balancer.PoolParams{
+		SwapFee: sdk.ZeroDec(),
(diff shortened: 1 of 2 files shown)


app/apptesting/gamm.go
```diff
@@ -128,33 +131,40 @@ func (s *KeeperTestHelper) PrepareImbalancedStableswapPool() uint64 {
 
 // PrepareBalancerPoolWithPoolParams sets up a Balancer pool with poolParams.
```
czarcas7ic, line 132: nit: can we add in the godoc that it uses default pool assets but custom poolParams?

czarcas7ic (approved):

Issue #3773, opened 2022-12-16
feat(osmomath): make mutative `Add` and `Quo`


p0mvn:
## Background

We need mutative `Add` and `Quo` defined on `osmomath.BigDec`

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/3708#discussion_r1050159218_

We alreadyhave mutative `MulMut: https://github.com/osmosis-labs/osmosis/pull/3678

## Suggested Design

Follow #3678 for an example on how to implement it with tests.

Modify `exp2` to utilize the mutatie versions as per: https://github.com/osmosis-labs/osmosis/pull/3708/files/79beaae72f54159a82c54a7872b6053fc8ffdc5a#r1050159218

## Acceptance Criteria
- `AddMut` added
- `QuoMut` added
- both tested
- implementation and tests similar to #3678 
- refactor `exp2`  as per: https://github.com/osmosis-labs/osmosis/pull/3708/files/79beaae72f54159a82c54a7872b6053fc8ffdc5a#r1050159218

PR #3779, opened 2022-12-17
osmomath: `AddMut` and `QuoMut`

RusAkh:
- mut add
- test add mut
- quo  mut
- test quo mut/ remove want from test struct

Closes: #3773 

mutative add and quo + tests


ValarDragon:
This isn't getting the speedup benefit of mutative arithmetic. Can you look at how we do it in our SDK fork for Decimal, and adapt it here in a similar style?

RusAkh:
@ValarDragon I pushed changes, not sure I understood what you mean tho, hope fixed correctly

osmomath/exp2.go
```diff
@@ -93,8 +93,8 @@ func exp2ChebyshevRationalApprox(x BigDec) BigDec {
 	for i := 1; i < len(numeratorCoefficients13Param); i++ {
 		x_exp_i.MulMut(x)
 
-		h_x = h_x.Add(numeratorCoefficients13Param[i].Mul(x_exp_i))
-		p_x = p_x.Add(denominatorCoefficients13Param[i].Mul(x_exp_i))
+		h_x.AddMut(numeratorCoefficients13Param[i].Mul(x_exp_i))
+		p_x.AddMut(denominatorCoefficients13Param[i].Mul(x_exp_i))
```
p0mvn, line 97: Please apply `QuoMut` at line 100 too

osmomath/decimal.go
```diff
@@ -332,6 +343,20 @@ func (d BigDec) Quo(d2 BigDec) BigDec {
 	return BigDec{chopped}
 }
 
+// mutative quotient
+func (d BigDec) QuoMut(d2 BigDec) BigDec {
+	// multiply precision twice
+	mul := new(big.Int).Mul(d.i, precisionReuse)
```
p0mvn, line 349: We should not allocate new int. One of the goals of the mutative versions is to gain speed up by avoiding reallocations and copying.

Please see:
https://github.com/osmosis-labs/cosmos-sdk/blob/48fe175f983925ff6202d492a363d2b2e3a62497/types/decimal.go#L343-L352

osmomath/exp2.go
RusAkh: applied, but why do we need it there? Does not a normal `Quo` suffice?

548207900.mdttt
```diff
@@ -0,0 +1,16 @@
+- mut add
```
p0mvn, line 1: Let's remove this

osmomath/exp2.go
p0mvn: There should be a a performance improvement with the mutative version

osmomath/decimal_test.go
```diff
@@ -38,6 +38,59 @@ func (s *decimalTestSuite) assertMutResult(expectedResult, startValue, mutativeR
...
+	for _, tc := range tests {
+		if !tc.expectingErr {
+			want := tc.testing.Quo(tc.quo)
+			tc.testing.QuoMut(tc.quo)
+
+			require.Equal(t, want, tc.testing)
+		} else {
+			save := tc.testing
+
+			require.Panics(t, func() { tc.testing.QuoMut(tc.quo) })
```
p0mvn, line 88: We have a helper for this here: https://github.com/osmosis-labs/osmosis/blob/62067092a726b59117345e11f3355e381a72bf3c/osmomath/export_test.go#L25

osmomath/decimal.go
```diff
@@ -332,6 +343,20 @@ func (d BigDec) Quo(d2 BigDec) BigDec {
 	return BigDec{chopped}
 }
 
+// mutative quotient
+func (d BigDec) QuoMut(d2 BigDec) BigDec {
```
p0mvn, line 347: Can you refactor `Quo` to look like:
https://github.com/osmosis-labs/osmosis/blob/62067092a726b59117345e11f3355e381a72bf3c/osmomath/decimal.go#L270-L276

osmomath/decimal.go
```diff
@@ -248,6 +248,17 @@ func (d BigDec) Add(d2 BigDec) BigDec {
 	return BigDec{res}
 }
 
+// mutative addition
+func (d BigDec) AddMut(d2 BigDec) BigDec {
```
p0mvn, line 252: Please refactor `Add`

osmomath/decimal_test.go
```diff
@@ -38,6 +38,59 @@ func (s *decimalTestSuite) assertMutResult(expectedResult, startValue, mutativeR
 	s.Require().Equal(nonMutativeStartValue, startValue)
 }
 
+func TestAddMut(t *testing.T) {
```
p0mvn, line 41: Since `Add` is supposed to call `AddMut`, can you please do a short test [like this](https://github.com/osmosis-labs/osmosis/blob/62067092a726b59117345e11f3355e381a72bf3c/osmomath/decimal_test.go#L1184-L1220) instead of duplicating test cases?

osmomath/decimal_test.go
```diff
@@ -38,6 +38,59 @@ func (s *decimalTestSuite) assertMutResult(expectedResult, startValue, mutativeR
...
+	}
+	for _, tc := range tests {
+		want := tc.testing.Add(tc.adding)
+		tc.testing.AddMut(tc.adding)
+
+		require.Equal(t, want, tc.testing)
+	}
+}
+
+func TestQuoMut(t *testing.T) {
```
p0mvn, line 63: Same for this test case

p0mvn (changes requested):

p0mvn (approved):
LGTM

ValarDragon (approved):
Awesome!

Issue #3239, opened 2022-11-04
[CL]:[Swap] Test CurrentSqrtPrice


czarcas7ic:
## Background

When running a SwapOutAmtGivenIn, we update the pool to show what the new currentSqrtPrice is after the swap. Sometimes, this takes multiple iterations to converge on the final value. We need to find a good way to test that the resulting currentSqrtPrice is appropriate and add it here to the table driven test.

https://github.com/osmosis-labs/osmosis/blob/dd6e2121430d236411802cfdc63973fc1db62ca3/x/concentrated-liquidity/pool_test.go#L490-L491

## Suggested Design

TBD

## Acceptance Criteria

TBD

PR #3783, opened 2022-12-18
test[CL]: CurrentSqrtPrice

czarcas7ic:
Closes: #3239 

This PR introduces a check in both Calc/Swap methods (for both In and Out) that the currentSqrtPrice we produce is what we expect. 

These checks took me down a bit of a rabbit hole to finally discover that the OOO was wrong on one of our liquidity checks, which is why the expected liquidity for a lot of the test cases has now changed. 

## Brief Changelog

  - Utilizes osmomath in Liquidity0, Liquidity1, and TickToSqrtPrice methods
  - Modfies Calc/Swap tests for both In and Out to utilize newer, more precise liquidity values as well as check that our final sqrtPrice is as intended (with corresponding wolframalpha links


x/concentrated-liquidity/internal/math/math.go
```diff
@@ -81,20 +95,21 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
+		denominator := liquidity.Add(product)
+		if denominator.GTE(liquidity) {
+			return liquidity.Mul(sqrtPriceCurrent).QuoRoundUp(denominator)
 		}
 	}
-	denominator := numerator.Quo(sqrtPriceCurrent).Add(amountRemaining)
-	sqrtPriceNext = numerator.QuoRoundUp(denominator)
-	return sqrtPriceNext
+	// if the product does overflow, use less precise formula
+	return liquidity.QuoRoundUp((liquidity.Quo(sqrtPriceCurrent).Add(amountRemaining)))
```
mattverse, line 112: Confirmed that there is no logic change here

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -81,20 +95,21 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
 
-	if product.Quo(amountRemaining).Equal(sqrtPriceCurrent) {
-		denominator := numerator.Add(product)
-		if denominator.GTE(numerator) {
-			numerator = numerator.Mul(sqrtPriceCurrent)
-			sqrtPriceNext = numerator.QuoRoundUp(denominator)
-			return sqrtPriceNext
+	product := amountRemaining.Mul(sqrtPriceCurrent)
+	// use precise formula if product doesn't overflow
+	// TODO: see if we even need to do this check
```
mattverse, line 104: Wdyt about creating an issue for this regarding adding testcase for testing product overflow?

x/concentrated-liquidity/internal/math/math_test.go
```diff
@@ -158,7 +158,7 @@ func (suite *ConcentratedMathTestSuite) TestCalcAmount0Delta() {
 			liquidity:       sdk.MustNewDecFromStr("1517882343.751510418088349649"), // we use the smaller liquidity between liq0 and liq1
 			sqrtPCurrent:    sdk.MustNewDecFromStr("70.710678118654752440"),         // 5000
 			sqrtPUpper:      sdk.MustNewDecFromStr("74.161984870956629487"),         // 5500
-			amount0Expected: "998976.618347426747968399",                            // TODO: should be 998976.618347426388356630
+			amount0Expected: "998976.618347426388356620",
```
mattverse, line 161: wow im surprised that we're getting so close to expected!

mattverse (approved):
LGTM. Overall very surprised that we got so close to reducing margin of error just by changing the Dec part for currentSqrtPrice! This is exciting!

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -12,9 +14,15 @@ func Liquidity0(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
-	product := sqrtPriceA.Mul(sqrtPriceB)
-	diff := sqrtPriceB.Sub(sqrtPriceA)
-	return amount.ToDec().Mul(product).Quo(diff)
+
+	// transform sqrtPriceA and sqrtPriceB to BigDec using osmomath for more final precision
```
p0mvn, line 18: Is precision loss significant enough to require this?

I'm surprised that basic math operations are causing a large precision loss

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -12,9 +14,15 @@ func Liquidity0(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+	sqrtPriceAOsmomath := osmomath.MustNewDecFromStr(sqrtPriceA.String())
+	sqrtPriceBOsmomath := osmomath.MustNewDecFromStr(sqrtPriceB.String())
```
p0mvn, line 20: We should do the conversions from `Dec` to `BigDec`

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -12,9 +14,15 @@ func Liquidity0(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+	product := sqrtPriceAOsmomath.Mul(sqrtPriceBOsmomath)
+	diff := sqrtPriceBOsmomath.Sub(sqrtPriceAOsmomath)
+
+	// conduct the liquidity calculation, then transform back to normal sdk.Dec
+	return osmomath.MustNewDecFromStr(amount.String()).Mul(product).Quo(diff).SDKDec()
```
p0mvn, line 25: We should not construct this from a string. There should be helpers to perform the conversion `Int` -> `Dec` -> `BigDec`

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -25,8 +33,14 @@ func Liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
-	diff := sqrtPriceB.Sub(sqrtPriceA)
-	return amount.ToDec().Quo(diff)
+
+	// transform sqrtPriceA and sqrtPriceB to BigDec using osmomath for more final precision
+	sqrtPriceAOsmomath := osmomath.MustNewDecFromStr(sqrtPriceA.String())
+	sqrtPriceBOsmomath := osmomath.MustNewDecFromStr(sqrtPriceB.String())
```
p0mvn, line 39: please change conversions

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -25,8 +33,14 @@ func Liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
...
+	diff := sqrtPriceBOsmomath.Sub(sqrtPriceAOsmomath)
+
+	// conduct the liquidity calculation, then transform back to normal sdk.Dec
+	return osmomath.MustNewDecFromStr(amount.String()).Quo(diff).SDKDec()
```
p0mvn, line 43: please change conversions

x/concentrated-liquidity/internal/math/math.go
p0mvn, line 112:
```suggestion
	return liquidity.QuoRoundUp(liquidity.Quo(sqrtPriceCurrent).Add(amountRemaining))
```

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -81,20 +95,21 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
+	}
 
-	if product.Quo(amountRemaining).Equal(sqrtPriceCurrent) {
-		denominator := numerator.Add(product)
-		if denominator.GTE(numerator) {
-			numerator = numerator.Mul(sqrtPriceCurrent)
-			sqrtPriceNext = numerator.QuoRoundUp(denominator)
-			return sqrtPriceNext
+	product := amountRemaining.Mul(sqrtPriceCurrent)
+	// use precise formula if product doesn't overflow
```
p0mvn, line 103: What is the precise formula?

Why would product overflow?

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -81,20 +95,21 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
+	if (product.Quo(amountRemaining)).Equal(sqrtPriceCurrent) {
+		denominator := liquidity.Add(product)
+		if denominator.GTE(liquidity) {
+			return liquidity.Mul(sqrtPriceCurrent).QuoRoundUp(denominator)
 		}
 	}
-	denominator := numerator.Quo(sqrtPriceCurrent).Add(amountRemaining)
-	sqrtPriceNext = numerator.QuoRoundUp(denominator)
-	return sqrtPriceNext
+	// if the product does overflow, use less precise formula
```
p0mvn, line 111: What do you mean by overflow? I think the definition is causing my confusion here

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -81,20 +95,21 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 // else
 // sqrtPriceNext = ((liquidity)) / (((liquidity) / (sqrtPriceCurrent)) + (amountRemaining))
 func GetNextSqrtPriceFromAmount0RoundingUp(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext sdk.Dec) {
-	numerator := liquidity
-	product := amountRemaining.Mul(sqrtPriceCurrent)
+	if amountRemaining.Equal(sdk.ZeroDec()) {
```
p0mvn, line 98: Can we remove pseudocode in the spec please and, instead, explain what that means?

x/concentrated-liquidity/internal/math/tick.go
```diff
@@ -28,15 +28,18 @@ func TickToSqrtPrice(tickIndex sdk.Int) (sqrtPrice sdk.Dec, err error) {
 	// If the tick index is positive, we can use the normal equation to calculate the square root price.
 	// However, if the tick index is negative, since we cannot take negative powers with the sdk,
 	// we need to take one over the original equation in order to make the power positive.
+	var sqrtPriceOsmoMath osmomath.BigDec
 	if tickIndex.GTE(sdk.ZeroInt()) {
-		sqrtPrice, err = sdk.NewDecWithPrec(10001, 4).Power(tickIndex.Uint64()).ApproxSqrt()
+		sqrtPriceOsmoMath, err = osmomath.MustNewDecFromStr("1.0001").Power(tickIndex.Uint64()).ApproxSqrt()
```
p0mvn, line 33: Let's initialize `osmomath.MustNewDecFromStr("1.0001")` as a package-global variable please

x/concentrated-liquidity/internal/math/tick.go
```diff
@@ -28,15 +28,18 @@ func TickToSqrtPrice(tickIndex sdk.Int) (sqrtPrice sdk.Dec, err error) {
...
 	} else {
-		sqrtPrice, err = sdk.OneDec().Quo(sdk.NewDecWithPrec(10001, 4).Power(tickIndex.Abs().Uint64())).ApproxSqrt()
+		sqrtPriceOsmoMath, err = osmomath.OneDec().Quo(osmomath.MustNewDecFromStr("1.0001").Power(tickIndex.Abs().Uint64())).ApproxSqrt()
```
p0mvn, line 35: and reuse it ere

x/concentrated-liquidity/internal/math/tick_test.go
```diff
@@ -13,22 +13,22 @@ func (suite *ConcentratedMathTestSuite) TestTickToSqrtPrice() {
 	}{
 		"positive tick index 1": {
 			tickIndex:         sdk.NewInt(85176),
-			sqrtPriceExpected: "70.710004849206351867", // 70.710004849206120647
+			sqrtPriceExpected: "70.710004849206120646",
```
p0mvn, line 16: The values seem slightly off judging by the old comment.

I think we should specify `sqrtPriceExpected` as given by Wolfram and define an "additive tolerance" for each test case.

x/concentrated-liquidity/internal/math/math.go
p0mvn: I think using `osmomath.BigDec` is useful in `TickToSqrtPrice` since it uses functions such as `Power`. However, I would be surprised if multiplications, subtractions and additions would be the cause of precision loss here

czarcas7ic: Screenshot 2022-12-19 at 10 34 01 AM

Reverting this back to what it was produces very minimal significance, I was just trying to match the exact result of wolframalpha. Do you think I should revert this to what is was (no using osmomath) and just accept the extremely small precision loss?

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Offline, agreed to revert liquidity0 and liquidity1 back to normal sdk.Decs due to minimal improvement in precision

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Offline, agreed to revert liquidity0 and liquidity1 back to normal sdk.Decs due to minimal improvement in precision

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Offline, agreed to revert liquidity0 and liquidity1 back to normal sdk.Decs due to minimal improvement in precision

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Offline, agreed to revert liquidity0 and liquidity1 back to normal sdk.Decs due to minimal improvement in precision

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Offline, agreed to revert liquidity0 and liquidity1 back to normal sdk.Decs due to minimal improvement in precision

x/concentrated-liquidity/internal/math/tick_test.go
czarcas7ic: old comment was actually off (rounded instead of truncated like we want).

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: After further tests, the precision loss is actually significant enough to warrant utilize BigDecs here. I will add them back and we can discuss further

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Having the formulas in the godoc has been really helpful for me to reference. I have referenced quite a few times.

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: This logic already existed prior to this PR. I was just adding comments from the uniswap code. The todo here is seeing if this less accurate formula to prevent overflows is needed in our context

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Precise formula is the top one, bottom one is less precise.

I don't know solidity math well enough to answer why their product might overflow

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/3791

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -12,9 +14,13 @@ func Liquidity0(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 	if sqrtPriceA.GT(sqrtPriceB) {
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
-	product := sqrtPriceA.Mul(sqrtPriceB)
-	diff := sqrtPriceB.Sub(sqrtPriceA)
-	return amount.ToDec().Mul(product).Quo(diff)
+	amountBigDec := osmomath.BigDecFromSDKDec(amount.ToDec())
```
p0mvn, line 17: Let's add a comment why we are making these conversions to `BigDec` please. That is, to avoid frequent precision loses for callers of this function

x/concentrated-liquidity/internal/math/math.go
p0mvn: It doesn't seem like a formula, more like a code block/pseudocode which shouldn't be in the spec.

The spec should abstract away the implementation details, allowing the developer to treat the function as a black box. In the current state, it only feels like additional complexity trying to visually QA the spec first before reading the function implementation. 

I agree that formulas are useful. However, if we are to leave them in the codebase, it should be clear what they are referring to exactly. I'm supportive of having them if they are clearly referencing the code lines and we don't have to include large chunks of pseudocode in the spec

x/concentrated-liquidity/internal/math/math.go
p0mvn: Let's not block this PR but create an issue to re-think this entire function and understand the math / whether we need the two kinds of computations

p0mvn (approved):
Pre-approving for when comments for `GetNextSqrtPriceFromAmount0RoundingUp` are addressed.

In this PR, I think it would be valuable to clean up the comments for that function.

In a future PR / issue, it would be valuable to revisit the calculations for that function, understand why we have two kinds and potentially simplify it

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/issues/3887

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Removed

Issue #3725, opened 2022-12-14
[CL]: Modfiy LocalOsmosis Script


czarcas7ic:
## Background

See this comment https://github.com/osmosis-labs/osmosis/pull/3707#discussion_r1048726159

We need to change the pool creation fee parameters in the LocalOsmosis script to change it via the swaprouter module.

Additionally, once we have solidified the CL module params, we need to ensure that none of these params need to get overridden via this LocalOsmosis script as well.

## Suggested Design

TBD

## Acceptance Criteria

TBD

PR #3784, opened 2022-12-18
chore[CL]: add swaprouter change to localosmosis

czarcas7ic:
Closes: #3725 

While we initialize the swaprouter already with uosmo as the default denom, it is good practice to ensure this gets overridden when making a new localnet in the case this unexpectedly changes.

This PR simply removes the commented out code that changed the default pool creation fee denom from the gamm module to where it now exists (the swaprouter module)


mattverse (approved):
LGTM

p0mvn:
Does it make sense to make this against `main`?

p0mvn (approved):

czarcas7ic:
@p0mvn I dont think we can move this to main until we actually remove the param from gamm to swaprouter. If you ack this and agree, I will just merge to cl-main

p0mvn:
The param has already been moved

czarcas7ic:
You are right! Closing this and opening a new one to main

Issue #3597, opened 2022-12-02
feat(client): change environment via CLI


p0mvn:
## Background

Currently, we have several environments:
- mainnet (under `$HOME/.osmosisd`)
- localosmosis (under `$HOME/.osmosisd-local`)

Separate folders for each are created for convenience to avoid having localosmosis overwriting mainnet client configurations.

However, to use a non-default config folder, users must provide a `--home` flag like:
```bash
osmosisd q tx {txhash} --home=$HOME/.osmosisd-local
```

We would like to avoid it and, instead, configure the environment permanently by setting it via a separate CLI command.

## Suggested Design

1. Modify `osmosisd` installation to create an environment file and source it in `.profile`

Modify [makefile installation step](https://github.com/osmosis-labs/osmosis/blob/1276d98bcc5cb45ad23423299f8261b004dd6dfb/Makefile#L92-L93) to also do the following:
- create a file under `$HOME/.osmosisd/env`
- set the following environment variable there `OSMOSISD_ENVIRONMENT=mainnet`
- source the path of the `env` to `.profile`: `. "$HOME/.osmosisd/env"`

Note that `cargo` works in this way.

2. According to the value of `OSMOSISD_ENVIRONMENT`, set the home dir

Set it here: https://github.com/osmosis-labs/osmosis/blob/1276d98bcc5cb45ad23423299f8261b004dd6dfb/cmd/osmosisd/cmd/root.go#L58

- mainnet => `$HOME/.osmosisd`
- localosmosis => `$HOME/.osmosisd-local`

3. Create a CLI command `osmosisd config set-env localosmosis`

This environment must update the value of `OSMOSID_ENVIRONMENT` at "$HOME/.osmosisd/env"`

## Acceptance Criteria
- switching between config folders via CLI is seamless
- new CLI command implemented
- environment sourced in `.profile`

ValarDragon:
This would be awesome!

nicolaslara:
I like this a lot! 

It would be great if we could also point to a custom home dir `osmosisd config set-env ~/devel/osmosis/v15-home`. 

Maybe we can give those directories names, but it may add complexity.

PR #3787, opened 2022-12-19
Cli change environment

hieuvubk:
Closes: #3597 

To avoid the user having to add the `--HOME` flag to each cli cmd, configure the environment permanently by setting it via a separate CLI command.

## Brief Changelog
- Create `.env` file in `$HOME/.osmosisd`. Only 2 values ​​should be set here (`mainnet` & `localosmosis`). Default value = `mainnet`
- HomeDir will depend on the value of env. `mainnet` ~ `$HOME/.osmosisd` .  `localosmosis` ~ `$HOME/.osmosisd-local`
- Create a cli command for updating env

## Testing and Verifying

cli update env & home dir successfully in my local


scripts/create-environment.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+
+echo "OSMOSISD_ENVIRONMENT=mainnet" > $HOME/.osmosisd/.env
```
p0mvn, line 4: have you tested what happens if `$HOME/.osmosisd/` does not exist?

cmd/osmosisd/cmd/root.go
```diff
@@ -85,6 +88,28 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
...
+		panic(err)
+	}
+
+	val := os.Getenv(EnvVariable)
+	if val == EnvMainnet {
+		return filepath.Join(userHomeDir, ".osmosisd")
+	} else if val == EnvLocalnet {
+		return filepath.Join(userHomeDir, ".osmosisd-local")
+	} else {
+		panic("invalid environment variable")
```
p0mvn, line 109: please format the value here

p0mvn (approved):
utACK. Great work. Thank you

I would like to test this before merge. However, if another reviewer tests this, I don't have to be blocking

scripts/create-environment.sh
hieuvubk: Yes we should check if there was an error in the previous install step

scripts/create-environment.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [ -d "$HOME/.osmosisd" ]; then
+  echo "OSMOSISD_ENVIRONMENT=mainnet" > $HOME/.osmosisd/.env
+fi
```
p0mvn, line 5: 1. What if .osmosisd exists but .osmosisd/.env does not?
2. If `"$HOME/.osmosisd"` does not exist, can we create one?

hieuvubk: This step runs after installing binary so `$HOME/.osmosisd` is required. If not, it can be considered a problem with the installation step.
Also, line 4 will create a `.env` file and fill it.

hieuvubk: If you have any suggestions pls lmk!

p0mvn: I see. Can we echo an error message if `"$HOME/.osmosisd"` does not exist? So that if this installation problem does happen, a user can know where to start debugging.

Also, would `echo "OSMOSISD_ENVIRONMENT=mainnet" > $HOME/.osmosisd/.env` append `"OSMOSISD_ENVIRONMENT=mainnet"` to `.env` every time we run `make build`? We should make sure that we don't append to the file redundantly on every build

ValarDragon: Hrmm, I'd rather we do this in the go code, similar to where we currently init the configs

I don't feel great about adding a separate process for this file being initialized

ValarDragon:
Quite excited for this!

Only note is can we get this file initialized at the same time / location other config files are already initialized

scripts/create-environment.sh
ValarDragon: I think so?

I always have to re-investigate every time where they get initialized, since actually multiple commands initialize the client.toml

cmd/osmosisd/cmd/root.go
```diff
@@ -85,6 +89,28 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
...
+		panic(err)
+	}
+
+	val := os.Getenv(EnvVariable)
+	if val == EnvMainnet {
+		return filepath.Join(userHomeDir, ".osmosisd")
+	} else if val == EnvLocalnet {
+		return filepath.Join(userHomeDir, ".osmosisd-local")
+	} else {
+		panic(fmt.Errorf("Invalid environment variable"))
```
ValarDragon, line 110: Wait the env val can be anything, we shouldn't restrict it. These are just the two common ones, no?

Like you can have a different filepath.

Also we need to ensure commands work with a different `--home` flag, and nothing in `.osmosisd`

hieuvubk:
Based on everyone's comments, I just:

- Initialize the environment at the same time as the config files

- Instead of restricting the env to only mainnet or testnet, we pass the path of the folders to the cli command

- Current limitation: `.osmosis` intermediary is required to save env

Pls lmk what u think @ValarDragon @p0mvn

ValarDragon:
`Current limitation: .osmosis intermediary is required to save env`

Assuming `.osmosis` is actually `.osmosisd`, sounds great to me!

p0mvn:
Hi @hieuvubk . 

Are you available to drive this PR to completion? If so, can you please get back to ValarDragon on the questions?

cmd/osmosisd/cmd/init.go
```diff
@@ -167,6 +169,18 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 
 			tmcfg.WriteConfigFile(filepath.Join(config.RootDir, "config", "config.toml"), config)
 
+			// create environment file in default node home
+			if defaultNodeHome == app.DefaultNodeHome {
```
p0mvn, line 173: q: what exactly does this check mean? I do not really understand.

Can you please add a comment explaining why we need it

cmd/osmosisd/cmd/root.go
```diff
@@ -85,6 +88,18 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 	return rootCmd, encodingConfig
 }
 
+func getHomeEnvironment() string {
+	envPath := filepath.Join(osmosis.DefaultNodeHome, ".env")
+
+	// Use default node home if can't get environment
+	err := godotenv.Load(envPath)
+	if err != nil {
+		return osmosis.DefaultNodeHome
```
p0mvn, line 97: I don't think we should be doing this. It will become confusing trying to figure out why this keeps giving me default when I expect something else

Makefile
```diff
@@ -100,7 +100,6 @@ build: BUILD_ARGS=-o $(BUILDDIR)/
 
 $(BUILD_TARGETS): check_version go.sum $(BUILDDIR)/
 	GOWORK=off go $@ -mod=readonly $(BUILD_FLAGS) $(BUILD_ARGS) ./...
-
```
p0mvn, line 103: What about this part of the spec in the issue?

"source the path of the env to .profile: . "$HOME/.osmosisd/env""

cmd/osmosisd/cmd/init.go
```diff
@@ -167,6 +169,18 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+				envFile, err := os.Create(filepath.Join(defaultNodeHome, ".env"))
```
p0mvn, line 174: Why are we using `defaultNodeHome` here and not `.osmosisd`?

What if I re-run init command while I'm in another environment?

We need to make sure that the `.env` is always saved in the same place independent of the environment

cmd/osmosisd/cmd/init.go
hieuvubk: I think we should remove this condition because caller already use `DefaultNodeHome` so no need to check more

Makefile
hieuvubk: Refer from [cmt](https://github.com/osmosis-labs/osmosis/pull/3787#issuecomment-1362194375), I move this step at the same time as the init chain

cmd/osmosisd/cmd/init.go
hieuvubk: The scenario would be, the first init will initialize the .`osmosisd` folder and the `.env` file (whether you initialize in `defaultNodeHome ` or not). 
From there, the `.env` is only be changed when using `ChangeEnvironmentCmd`. 
If we run-run init, the `.env` still save the last environment

p0mvn:
Heu @hieuvubk . e2e started failing here for some reason. Could you take a look please?

hieuvubk:
> Hey @hieuvubk . e2e started failing here for some reason. Could you take a look please?

Its seem from some fmt we put in

p0mvn:
> > Hey @hieuvubk . e2e started failing here for some reason. Could you take a look please?
> 
> Its seem from some fmt we put in The cmd response contains both params response & text msg Im gonna fix it rn

Thank you. This change LGTM, nice work!

Let's get this in and experiment with this over the next week with localosmosis.

p0mvn (approved):

p0mvn:
Tested locally with `mainnet` and `localosmosis` options. Works well for me, nice work @hieuvubk

PR #3793, opened 2022-12-19
test(x/twap): move three asset same record test to TestComputeTwap

p0mvn:
Closes: #XXX

The old three asset twap equal record test assumes additional validation from `computeTwap` function. Therefore, we should move the test case from being made against strategies to be against `computeTwap` function.

Additionally, we should add a test case against the arithmetic twap strategy where we end up having a time delta of zero, making it panic (Currently marked as TODO)


stackman27 (approved):
LGTM

PR #3795, opened 2022-12-19
chore: add swaprouter change to localosmosis

czarcas7ic:
Closes: #3725 

While we initialize the swaprouter already with uosmo as the default denom, it is good practice to ensure this gets overridden when making a new localnet in the case this unexpectedly changes.

This PR simply removes the commented out code that changed the default pool creation fee denom from the gamm module to where it now exists (the swaprouter module)

### DIFF
--- a/tests/localosmosis/scripts/setup.sh
+++ b/tests/localosmosis/scripts/setup.sh
@@ -55,8 +55,8 @@ edit_genesis () {
     dasel put string -f $GENESIS '.app_state.mint.params.mint_denom' "uosmo"
     dasel put string -f $GENESIS '.app_state.mint.params.epoch_identifier' "day"
 
-    # Update gamm module
-    dasel put string -f $GENESIS '.app_state.gamm.params.pool_creation_fee.[0].denom' "uosmo"
+    # Update swaprouter module
+    dasel put string -f $GENESIS '.app_state.swaprouter.params.pool_creation_fee.[0].denom' "uosmo"
 
     # Update txfee basedenom
     dasel put string -f $GENESIS '.app_state.txfees.basedenom' "uosmo"


p0mvn (approved):

tests/localosmosis/scripts/setup.sh
```diff
@@ -55,8 +55,8 @@ edit_genesis () {
     dasel put string -f $GENESIS '.app_state.mint.params.mint_denom' "uosmo"
     dasel put string -f $GENESIS '.app_state.mint.params.epoch_identifier' "day"
 
-    # Update gamm module
-    dasel put string -f $GENESIS '.app_state.gamm.params.pool_creation_fee.[0].denom' "uosmo"
+    # Update swaprouter module
+    dasel put string -f $GENESIS '.app_state.swaprouter.params.pool_creation_fee.[0].denom' "uosmo"
```
ValarDragon, line 59: Can we update both?

ValarDragon, line 59:
```suggestion
    # Update swaprouter module
    dasel put string -f $GENESIS '.app_state.swaprouter.params.pool_creation_fee.[0].denom' "uosmo"
    dasel put string -f $GENESIS '.app_state.gamm.params.pool_creation_fee.[0].denom' "uosmo"
```

ValarDragon (approved):

tests/localosmosis/scripts/setup.sh
czarcas7ic: I could be not tracking something, but isn't the param for pool creation fee under gamm getting removed in place of swaprouter? @p0mvn

p0mvn: It's still present only for migration purposes during the upgrade. We will remove it completely in the upgrade after.

AFAIU, changing it in gamm shouldn't make a difference

czarcas7ic: Right, I guess my point is, if we make this requested change, then when we do remove the param from gamm in the next upgrade, if we forget to remove it from here as well then localosmosis will fail to start. Basically no up side and only potential to make a mistake in the future.

ValarDragon: It helps if your using multi-version builds now, no? (e.g. accidentally have this file, or this produced genesis in v13)

Issue #3723, opened 2022-12-14
[CL]: Create Concentrated Pool: Check Denoms Exist On Chain


czarcas7ic:
## Background

See https://github.com/osmosis-labs/osmosis/pull/3687/#discussion_r1048716779

We should add a function to validate basic that checks if the denoms exist on chain https://github.com/osmosis-labs/osmosis/blob/da489c82c4b86fd3a6dc0ec1bb8eecee7629410d/x/concentrated-liquidity/model/msgs.go#L87-L90

Right now, we just check that the denoms are valid.

RusAkh:
Is this free to work on?

mattverse:
Closed by https://github.com/osmosis-labs/osmosis/pull/3804

PR #3804, opened 2022-12-20
[CL]Misc: Add Denom Check upon pool creation

mattverse:
Closes: #3723 

Adds denom check upon pool creation via msg server. 
The original design discussed in the issue was doing the validation using msg validation method, but this is found impossible, as we need to use bank keeper for the verification.

The denom check done here is namely the `GetDenomMetaData` check using Bank keeper, reviewers please confirm if this is the correct check we want to do here.


czarcas7ic (approved):
This logic LGTM! If nothing heard by EOD I will merge

PR #3806, opened 2022-12-20
Protorev: Post Handler + Proposal Handlers + Profit Sharing

davidterpay:
This PR implements the post handler, the proposal handlers that are responsible for setting the admin account, admin functionality, stateful changes depending on proposals, and developer profit sharing. This is PR 5/6 (https://github.com/osmosis-labs/osmosis/issues/3594).

## Brief Changelog
 
- *Implement the proposal handlers for `x/protorev`*
- *Create admin functionality to set hot routes, the developer account, and number of pools to iterate per tx*
- *Profit splitting after every successful trade execution for the developer account*
- *Implement the post handler*

## Testing and Verifying

- *Added unit tests for gas refunds*
- *Added unit tests for mock swap tx*
- *Added unit tests for extracting the correct pools from the tx*
- *Added unit tests for proposals*
- *Added unit tests for new setters and getters*
- *Added unit tests for updating developer fees*
- *Added unit tests for enabling or disabling the module (and affects in epoch and post handler)*

The `postHandler` extracts pools that were swapped in a transaction and determines if there is a cyclic arbitrage opportunity. If so, the handler will find an optimal route and execute it - rebalancing the pool and returning arbitrage profits to the module account. Additionally, the gas meter gets reset at the end of the post handler in order to refund users for computation done in the handler.

1. Extract all pools that were traded on in the transaction (`ExtractSwappedPools`) as well as the direction of the trade.
2. Create cyclic arbitrage routes for each of the swaps above (`BuildRoutes`)
3. For each feasible route, determine if there is a cyclic arbitrage opportunity (`IterateRoutes`)
    1. Determine the optimal amount to swap in and its respective profits via binary search over range of potential input amounts (`FindMaxProfitForRoute`)
    2. Compare profits of each route, keep the best route and input amount with the highest profit
4. If the best route and input amount has a profit > 0, execute the trade (`ExecuteTrade`) and rebalance the pools on-behalf of the chain through the `gammkeeper` (`MultiHopSwapExactAmountIn`)
5. Keep the profits in the module’s account for subsequent distribution.

#### Keeper Updates
| ProtoRevEnabled | Tracks whether the protorev module is enabled | []byte{7} | []byte{bool} | KV |
| AdminAccount | Tracks the admin account for protorev | []byte{8} | []byte{sdk.AccAddress} | KV |
| DeveloperAccount | Tracks the developer account for protorev | []byte{9} | []byte{sdk.AccAddress} | KV |
| DaysSinceGenesis | Tracks the number of days since the module was initialized. Used to track profits that can be withdrawn by the developer account | []byte{10} | []byte{uint} | KV |
| DeveloperFees | Tracks the profits that the developer account can withdraw | []byte{11} + []byte{tokenDenom} | []byte{sdk.Coin} | KV |
| MaxRoutesPerTx | Tracks the maximum number of routes that can be traversed per tx | []byte{12} | []byte{uint64} | KV |
| MaxRoutesPerBlock | Tracks the maximum number of routes that can be traversed per block | []byte{13} | []byte{uint64} | KV |
| RouteCountForBlock | Tracks the number of routes that have been traversed in the current block | []byte{14} | []byte{uint64} | KV |
| LatestBlockHeight | Tracks the latest recorded block height | []byte{15} | []byte{uint64} | KV |
| RouteWeights | Tracks the weights of the different routes | []byte{16} | []byte{uint64} | KV |

**ProtoRevEnabled**
`x/protorev` can be enabled or disabled through governance. As a proposal is a stateful change, we store whether the module is currently enabled or disabled in the module.

**AdminAccount**
The admin account is set through governance and has permissions to set hot routes and the developer account. On genesis, the admin account is nil.

**DeveloperAccount**
The developer account is set through a `NewMsgSetDeveloperAccount` tx. This is the account that will be able to withdraw a portion of the profits from `x/protorev` as specified by the Osmosis ↔ Skip proposal. Only the admin account has permission to make this message.

**DaysSinceGenesis**
`x/protorev` will distribute 20% of profits to skip in the year 1, 10% of profits in year 2, and 5% thereafter. To track how much profit can be distributed to the developer account at any given moment, we store the amount of days since genesis.

**DeveloperFees**
DeveloperFees tracks the total amount of profit that can be withdrawn by the developer account. These fees are sent to the developer account, if set, every week through the `epoch` hook. If unset, the funds are held in the module account.

**MaxRoutesPerTx**
MaxRoutesPerTx tracks the maximum number of routes that can be traversed in a given transaction. This is configurable (but bounded) by the admin account. We limit the number of routes per transaction so that all `x/protorev` execution is not limited to the top of the block.

**MaxRoutesPerBlock**
MaxRoutesPerBlock tracks the maximum number of routes that can be traversed in a given block. This is configurable (but bounded) by the admin account. We limit the number of routes per block so that the execution time of the `x/protorev` posthandler is reasonably bounded to ensure that block time remains as is.

**RouteCountForBlock**
RouteCountForBlock tracks the number of routes that have been traversed in the current block. Used to ensure that the module is not slowing down block speed.

**LatestBlockHeight**
LatestBlockHeight tracks the latest recorded block height. Used to track the number of routes that have been traversed in the current block.

**RouteWeights**
RouteWeights contains the weights of all of the different route types. Routes are broken up into different types based on the pool that is sandwiched in between the arbitrage route. This distinction is made and necessary because the execution time ranges fairly between the different route types.

#### Proposal Handlers
`x/protorev` implements two different governance proposals. 

**`SetProtoRevAdminAccountProposal`**
As the landscape of pools on Osmosis evolves, an admin account will be able to add and remove routes for `x/protorev` to check for cyclic arbitrage opportunities. Largely, the purpose of maintaining hot routes is to reduce the amount of computation that would otherwise be required to determine optimal paths at runtime. In addition, introducing a means of altering hot can allow external researchers to conduct meaningful analysis into the markets on-chain.

**`SetProtoRevEnabledProposal`**
This proposal type allows the chain to turn the module on or off. This is meant to be used as a fail safe in the case stakers and the chain decide to turn the module off. This might be used to halt the execution of trades in the case that the `x/gamm` module has significant upgrades that might produce unexpected behavior from the module.

#### Profit Sharing
Profits accumulated by the module will be partially distributed to the developers that built the module in accordance with the governance proposal that was passed: year 1 is 20% of profits, year 2 is 10%, and subsequent years is 5%.

In order to track how much profit the developers can withdraw at any given moment, the module tracks the number of days since genesis. This gets incremented in the epoch hook after every day. When a trade gets executed by the module, the module will determine how much of the profit from the trade the developers can receive by using `daysSinceGenesis` in a simple calculation. 

If the developer account is not set (which it is not on genesis), all funds are held in the module account. Once the admin account is set through a successful governance proposal, the developer address can be set and will start to automatically receive a share of profits every week through the epoch hook. The distribution of funds from the module account is done through `SendDeveloperFeesToDeveloperAccount`. Once the funds are distributed, the amount of profit developers can withdraw gets reset to 0 and profits will start to be accumulated and distributed on a week to week basis.


x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,121 @@
...
+		if err := protoRevDec.ProtoRevKeeper.ProtoRevTrade(cacheCtx, swap); err != nil {
+			tradeErr = err
+			break
+		}
+	}
+	// If there was no error, write the cache context to the main context
+	if tradeErr == nil {
+		write()
+		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
+	}
```
NotJeremyLiu, line 70: As of now, this current logic will track if an error occurs when attempting to execute a trade (this should never happen, if it occurs it means something is wrong with protorev logic), and if one occurs it doesn't write anything to the main ctx for that posthandler's run (even if a previous arb executed for the first pool in a user's swap).

We could however change this logic such that if an error occurs, we refresh our cacheCtx and continue going. This would allow for the case where a user swaps against 2 pools, the first pool results in an error, but we successfully arb the second. This code is below:

`		// If there was no error executing the trade, then write the cache context to the main context,
		// emit events, and create a new cache context with an infinite gas meter from the main context
		executed, err := protoRevDec.ProtoRevKeeper.ProtoRevTrade(cacheCtx, swap)
		if err == nil && executed {
			write()
			ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
			cacheCtx, write = ctx.CacheContext()
			cacheCtx = cacheCtx.WithGasMeter(sdk.NewInfiniteGasMeter())
		} else if err != nil {
			// If there was an error executing the trade, then create a
			// new cache ctx with an infinite gas meter from the main context
			cacheCtx, write = ctx.CacheContext()
			cacheCtx = cacheCtx.WithGasMeter(sdk.NewInfiniteGasMeter())
		}`

NotJeremyLiu: My preference is to keep the currently implemented logic because the error shouldn't happen. I would want to add a logger when `err != nil` though so that we could more easily identify if an error occurs while it's live.

x/protorev/genesis.go
```diff
@@ -17,6 +17,15 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 	// Init module parameters
 	k.SetParams(ctx, genState.Params)
 
+	// Init module state
+	k.SetProtoRevEnabled(ctx, genState.Params.Enabled)
+	k.SetDaysSinceGenesis(ctx, 0)
```
ValarDragon, line 22: Do we need to set days since genesis in the upgrade handler for protorev?

ValarDragon: EDIT: Ah this is 0, its since start of protorev

x/protorev/keeper/rebalance.go
```diff
@@ -153,5 +153,10 @@ func (k Keeper) ExecuteTrade(ctx sdk.Context, route swaproutertypes.SwapAmountIn
 		return err
 	}
 
+	// Update the developer fees
+	if err = k.UpdateDeveloperFees(ctx, inputCoin, tokenOutAmount); err != nil {
+		return err
+	}
```
ValarDragon, line 159: hrmm, in statistics its based on `	profit := tokenOutAmount.Sub(inputCoin.Amount)`

shouldn't that be whats passed in here as well?

ValarDragon: Ah I see its calculated again in the function, missed that.

x/protorev/client/cli/tx.go
```diff
@@ -0,0 +1,139 @@
...
+
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+
+	"github.com/osmosis-labs/osmosis/v13/x/protorev/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// CmdSetProtoRevAdminAccountProposal implements the command to submit a SetProtoRevAdminAccountProposal
+func CmdSetProtoRevAdminAccountProposal() *cobra.Command {
```
stackman27, line 21: We might not have to parse custom build message with new cli tools 

For reference: 
https://github.com/osmosis-labs/osmosis/blob/05a9fe4262a5fcc302e98b0ddf87d9f5f538b05e/x/concentrated-liquidity/client/cli/tx.go#L25-L51

x/protorev/keeper/developer_fees_test.go
```diff
@@ -0,0 +1,168 @@
+package keeper_test
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v13/app/apptesting"
+	"github.com/osmosis-labs/osmosis/v13/x/protorev/types"
+)
+
+// TestSendDeveloperFeesToDeveloperAccoutn tests the SendDeveloperFeesToDeveloperAccount function
```
stackman27, line 10:
```suggestion
// TestSendDeveloperFeesToDeveloperAccount tests the SendDeveloperFeesToDeveloperAccount function
```

x/protorev/keeper/developer_fees_test.go
```diff
@@ -0,0 +1,168 @@
...
+func (suite *KeeperTestSuite) TestSendDeveloperFeesToDeveloperAccount() {
+	cases := []struct {
+		description   string
+		malleate      func()
```
stackman27, line 14: What does malleate mean?

x/protorev/keeper/developer_fees_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}{
+		{
+			description:   "Send with unset developer account",
+			malleate:      func() {},
+			expectedErr:   true,
+			expectedCoins: sdk.NewCoins(),
+		},
+		{
+			description: "Send with set developer account",
+			malleate: func() {
```
stackman27, line 26: The structure of this test field looks very repetitive i.e. we're creating randomAcc, settingDevAccount, minting coins etc. Maybe we can use a helper setup function?

x/protorev/keeper/developer_fees_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for _, tc := range cases {
+		suite.Run(tc.description, func() {
+			suite.SetupTest()
+			tc.malleate()
+
+			err := suite.App.AppKeepers.ProtoRevKeeper.UpdateDeveloperFees(suite.Ctx, tc.inputCoin, tc.tokenOutAmount)
+			suite.Require().NoError(err)
+
+			developerFees, err := suite.App.AppKeepers.ProtoRevKeeper.GetDeveloperFees(suite.Ctx, tc.inputCoin.Denom)
```
stackman27, line 163: Is the `feesDenom` always going to either osmo or atom?

stackman27 (commented):
Apologies for the late review, got caught up with some other work. 

left few comments! still reviewing files past posthandler 🌮

x/protorev/keeper/developer_fees_test.go
davidterpay: Yup. `x/protorev` exclusively executes arbitrage trades using either Osmo or Atom, so the fees sent to the developer account will always be denominated in one of those two assets.

x/protorev/keeper/developer_fees_test.go
davidterpay: malleate i.e. applying some changes to state using a wrapper function.

davidterpay: might be more straightforward to just use alterState or something along those lines. will update that.

x/protorev/keeper/rebalance.go
davidterpay: I'll clean it up so the same profit obj is passed around instead of recalculating it.

x/protorev/client/cli/tx.go
davidterpay: This is a pretty neat update (great to know for the last PR I am going to make with all of the CLI cmds for `x/protorev`). I like it. I'll refactor to use the new cli tools.

x/protorev/keeper/developer_fees_test.go
davidterpay: added `pseudoExecuteTrade` as a wrapper to the functionality described above

x/protorev/keeper/developer_fees_test.go
davidterpay: updated to use `alterState`

x/protorev/keeper/rebalance.go
davidterpay: updated the PR so we use a single profit calc and opt for passing only denom instead of the input coin for both `UpdateStatistics` and `UpdateDeveloperFees`.

x/protorev/client/cli/tx.go
davidterpay: Per convo, I refactored the proposals to route directly to protorev instead of gov - mirroring the txfees module proposal handlers.

x/protorev/client/cli/tx.go
```diff
@@ -0,0 +1,154 @@
...
+	cmd.Flags().String(cli.FlagDeposit, "", "deposit of proposal")
+	flags.AddTxFlagsToCmd(cmd)
+	_ = cmd.MarkFlagRequired(cli.FlagTitle)
+	_ = cmd.MarkFlagRequired(cli.FlagDescription)
+
+	return cmd
+}
+
+// CmdSetProtoRevEnabledProposal implements the command to submit a SetProtoRevEnabledProposal
+func CmdSetProtoRevEnabledProposal() *cobra.Command {
```
bpiv400, line 92: Lots of duplicated code between the two RunE handlers. In fact it looks the same except for the proposal type that gets passed in. Maybe we should have a generator function that takes in a type and returns the function?

x/protorev/genesis.go
bpiv400, line 22: This name going to confuse future developers I assume, similar to how @ValarDragon was confused. Perhaps we rename it to something that indicates we mean since Protorev Genesis (e.g. k.SetDaysSinceModuleGenesis or k.SetDaysSinceProtorevGenesis)

x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,123 @@
...
+	gammtypes "github.com/osmosis-labs/osmosis/v13/x/gamm/types"
+)
+
+type SwapToBackrun struct {
+	PoolId        uint64
+	TokenOutDenom string
+	TokenInDenom  string
+}
+
+type ProtoRevDecorator struct {
```
bpiv400, line 15: Is ProtoRevDecorator implementing an Interface? If so, should we make that explicit? 

var _ Interface = Struct{} 

Even though it's unnecessary for the go compiler, it's a very common pattern in Tendermint. I don't know if it's widely used in Osmosis or in the SDK, but it's definitely helpful for helping your text editor find interface problems prior to compilation. No need to do this if it's uncommon.

x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,123 @@
...
+	// Get the max number of pools to iterate through
+	maxPoolsToIterate, err := protoRevDec.ProtoRevKeeper.GetMaxPools(cacheCtx)
+	if err != nil {
+		return next(ctx, tx, simulate)
+	}
+
+	// Find routes for every single pool that was swapped on (up to maxPoolsToIterate pools per tx)
+	swappedPools := ExtractSwappedPools(tx)
+	tradeErr := error(nil)
+	for index, swap := range swappedPools {
```
bpiv400, line 56: nit: I think it's cleaner to take the min of len(swappedPools) and maxPoolsToIterate and just iterate over that, so you can get rid of a break statement

x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,123 @@
...
+	} else {
+		ctx.Logger().Error("ProtoRevTrade failed with error", tradeErr)
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+// ProtoRevTrade wraps around the build routes, iterate routes, and execute trade functionality to execute an cyclic arbitrage trade
+// if it exists. It returns an error if there was an error executing the trade and a boolean if the trade was executed.
+func (k Keeper) ProtoRevTrade(ctx sdk.Context, swap SwapToBackrun) error {
```
bpiv400, line 79: It looks like we're not doing anything to limit the maximum number of routes iterated per block (i.e. to explicitly cap the extra amount of computation protorev can be responsible for in the worst case). 

Is this because that was hard to do or because we decided that was unnecessary for some reason?

bpiv400 (changes requested):

x/protorev/keeper/posthandler.go
NotJeremyLiu: Correct, as of now there are only local bounds (within each tx), nothing across the block.

We should be able to add an across block bound if we want to. 

The current decision to go to local bounding was because a cheap attack vector is to make a single tx that swaps against 100+ pools that busies up protorev, so local bounding is necessary (at min, so chose this over the across block bound method).

I'd be comfortable with a lenient across block bound that allows for up to 100 swaps (or some number that goes above some target average runtime), mainly would be concerned about attacks though rather than real defi activity for the time being. @bpiv400

x/protorev/keeper/posthandler.go
davidterpay: In this particular case, we are not implementing an interface. We did implement an interface in epoch_hooks.go which does include the one liner. For reference, the way the posthandler is implemented mirrors the same way the ante handler for other modules was designed (ex. https://github.com/osmosis-labs/osmosis/blob/2ec476bc12d2c8726c6850496ced274a5ba7ab49/x/txfees/keeper/feedecorator.go#L144).

x/protorev/keeper/posthandler.go
davidterpay: @bpiv400 we've updated the code such that we now have a maximum # of txs that can be processed per block. The counter tracking this will only be incremented if the tx had at least one swap msg within it. If the number of txs has exceeded the threshold, the posthandler does nothing. Additionally, we make this a configurable parameter that can be altered via the admin account (but still has an acceptable range so that the admin account doesn't introduce an attack vector).

davidterpay: update: we now constrain protorev by the number of routes traversed by block and tx.

x/protorev/genesis.go
```diff
@@ -14,8 +14,22 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 		panic(err)
 	}
 
-	// Init module parameters
+	// Init module state
 	k.SetParams(ctx, genState.Params)
+	k.SetProtoRevEnabled(ctx, genState.Params.Enabled)
+	k.SetDaysSinceModuleGenesis(ctx, 0)
+	k.SetLatestBlockHeight(ctx, uint64(ctx.BlockHeight()))
+	k.SetRouteCountForBlock(ctx, 0)
```
bpiv400, line 22: You should update the readme of the PR to include this since it's missing

x/protorev/keeper/protorev.go
```diff
@@ -135,3 +136,247 @@ func (k Keeper) DeleteAllEntriesForKeyPrefix(ctx sdk.Context, keyPrefix []byte)
...
+		// This should never happen as it is set to the default value on genesis
+		return 0, fmt.Errorf("max routes configuration has not been set in state")
+	}
+
+	res := sdk.BigEndianToUint64(bz)
+	return res, nil
+}
+
+// SetMaxRoutesPerTx sets the max number of routes that can be iterated per transaction
+func (k Keeper) SetMaxRoutesPerTx(ctx sdk.Context, maxRoutes uint64) error {
```
bpiv400, line 346: Is this really per Tx? It looks like this limit is applied per swap message, since the ante handler is called once per message.

x/protorev/keeper/protorev.go
```diff
@@ -135,3 +136,247 @@ func (k Keeper) DeleteAllEntriesForKeyPrefix(ctx sdk.Context, keyPrefix []byte)
...
+}
+
+// SetDeveloperAccount sets the developer account for protorev that will receive a portion of arbitrage profits
+func (k Keeper) SetDeveloperAccount(ctx sdk.Context, developerAccount sdk.AccAddress) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixDeveloperAccount)
+	store.Set(types.KeyPrefixDeveloperAccount, developerAccount.Bytes())
+}
+
+// GetMaxRoutesPerTx returns the max number of routes that can be iterated per transaction
+func (k Keeper) GetMaxRoutesPerTx(ctx sdk.Context) (uint64, error) {
```
bpiv400, line 333: Not documented in PR description

x/protorev/keeper/protorev.go
```diff
@@ -135,3 +136,247 @@ func (k Keeper) DeleteAllEntriesForKeyPrefix(ctx sdk.Context, keyPrefix []byte)
...
+
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixMaxRoutesPerTx)
+	bz := sdk.Uint64ToBigEndian(maxRoutes)
+	store.Set(types.KeyPrefixMaxRoutesPerTx, bz)
+
+	return nil
+}
+
+// GetMaxRoutesPerBlock returns the max number of routes that can be iterated per block
+func (k Keeper) GetMaxRoutesPerBlock(ctx sdk.Context) (uint64, error) {
```
bpiv400, line 359: Not documented in PR description

x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,189 @@
...
+	numberOfIterableRoutes, err := k.CalcNumberOfIterableRoutes(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Iterate and build arbitrage routes for each pool that was swapped on
+	for index := 0; index < len(swappedPools) && numberOfIterableRoutes > 0; index++ {
+		// Build the routes for the pool that was swapped on
+		routes := k.BuildRoutes(ctx, swappedPools[index].TokenInDenom, swappedPools[index].TokenOutDenom, swappedPools[index].PoolId)
+		numRoutes := uint64(len(routes))
```
bpiv400, line 108: nit: Be consistent with your naming. 

You have "numRoutes" and "numberOfIterableRoutes". Plus it's not clear what numRoutes represents. 

Maybe "numPossibleRoutes" and "numIterableRoutes" to make distinguishing easier.

bpiv400 (commented):
Few minor comments around naming and cleaning up the readme in light of the new keeper data.

x/protorev/keeper/protorev.go
davidterpay: resolved per convo. tldr is the antehandler runs per tx and we aggregate all the pools that were swapped on in the tx across all messages so MaxRoutesPerTx does constrain execution on a per tx scale.

x/protorev/keeper/protorev.go
davidterpay: updated

x/protorev/keeper/protorev.go
davidterpay: added to keeper updates

x/protorev/keeper/posthandler.go
```diff
@@ -0,0 +1,148 @@
...
+func NewProtoRevDecorator(protoRevDecorator Keeper) ProtoRevDecorator {
+	return ProtoRevDecorator{
+		ProtoRevKeeper: protoRevDecorator,
+	}
+}
+
+// This posthandler will first check if there were any swaps in the tx. If so, collect all of the pools, build three
+// pool routes for cyclic arbitrage, and then execute the optimal route if it exists.
+func (protoRevDec ProtoRevDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// Create a cache context to execute the posthandler such that
```
davidterpay, line 30: Do we want to add a check to ignore if we are in `checkTx`.

bpiv400 (dismissed):

x/protorev/keeper/routes.go
```diff
@@ -165,3 +198,55 @@ func (k Keeper) GetAndCheckPool(ctx sdk.Context, poolId uint64) (gammtypes.CFMMP
...
+	poolType, err := k.gammKeeper.GetPoolType(ctx, middlePool)
+	if err != nil {
+		return 0, err
+	}
+
+	switch poolType {
+	case swaproutertypes.Balancer:
+		return 1, nil
+	case swaproutertypes.Stableswap:
+		return 2, nil
```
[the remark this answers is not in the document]
davidterpay: Switching it over so that it is configurable, but we will still need to have some placeholder that can be used on genesis.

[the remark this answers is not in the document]
davidterpay: update: after running some tests, balancer routes take on average 1-2 ms and stable routes take around 4-5 ms. Setting the default to be 2 and 5 ms for the balancer and stable routes respectively.

x/protorev/types/genesis.go
```diff
@@ -19,7 +19,13 @@ var MaxInputAmount = sdk.NewInt(16_384)
 var StepSize = sdk.NewInt(1_000_000)
 
 // Max iterations for binary search (log2(16_384) = 14)
-const MaxIterations = 14
+const MaxIterations int = 14
+
+// Max number of routes that can be arbitraged per tx (default of 6)
```
stackman27, line 24: I'm trying to understand this, does this mean 6 routes each containing 3 pool routes? So it'd look something like [[route1][route2][route3]],  [[route1a][route2a][route3a]], [[route1b][route2b][route3b]] ... 6times in genesis

stackman27 (commented):
Left few comments, still catching up with changes that were made. But its looking solid :))

x/protorev/keeper/developer_fees_test.go
```diff
@@ -0,0 +1,171 @@
...
+			tc.alterState()
+
+			err := suite.App.AppKeepers.ProtoRevKeeper.SendDeveloperFeesToDeveloperAccount(suite.Ctx)
+			if tc.expectedErr {
+				suite.Require().Error(err)
+			} else {
+				suite.Require().NoError(err)
+			}
+
+			developerAccount, err := suite.App.AppKeepers.ProtoRevKeeper.GetDeveloperAccount(suite.Ctx)
```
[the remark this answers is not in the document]
davidterpay: good catch. going to wrap the err inside of `suite.Require().Error()` if we expect one.

x/protorev/keeper/developer_fees.go
```diff
@@ -0,0 +1,64 @@
...
+
+// UpdateDeveloperFees updates the fees that developers can withdraw from the module account
+func (k Keeper) UpdateDeveloperFees(ctx sdk.Context, denom string, profit sdk.Int) error {
+	daysSinceGenesis, err := k.GetDaysSinceModuleGenesis(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Calculate the developer fee
+	if daysSinceGenesis < 365 {
```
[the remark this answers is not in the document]
davidterpay: Consts does make more sense in this context. Will do that.

x/protorev/types/genesis.go
davidterpay: Yes exactly. But this is also dependent on the route that is taken. Routes that include a stable swap pool inheritely are more expensive to backrun. Because of this we count routes that include a stable swap pool with more weight. Currently the distribution is 2:5 for balancer : stable weights. The basic idea here is that `MaxIterableRoutesPerTx` and `MaxIterableRoutesPerBlock` closely follow the actual ms that we expect the module to take and route weights are the approximate amount of time it takes to actually traverse the route (in ms).

x/protorev/keeper/protorev.go
```diff
@@ -135,3 +136,284 @@ func (k Keeper) DeleteAllEntriesForKeyPrefix(ctx sdk.Context, keyPrefix []byte)
...
+	bz := []byte(strconv.FormatBool(enabled))
+	store.Set(types.KeyPrefixProtoRevEnabled, bz)
+}
+
+// GetRouteCountForBlock returns the number of routes that have been traversed in the current block
+func (k Keeper) GetRouteCountForBlock(ctx sdk.Context) (uint64, error) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixRouteCountForBlock)
+	bz := store.Get(types.KeyPrefixRouteCountForBlock)
+	if bz == nil {
+		// This should never happen as this is set to 0 on genesis
```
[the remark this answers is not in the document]
davidterpay: see above

PR #3811, opened 2022-12-21
fix[CL]: codec changes

czarcas7ic:
Closes: #XXX

Small codec change I left out. It is needed for pool queries


x/concentrated-liquidity/types/codec.go
```diff
@@ -10,8 +10,8 @@ import (
 
 func RegisterCodec(cdc *codec.LegacyAmino) {
 	cdc.RegisterInterface((*ConcentratedPoolExtension)(nil), nil)
-	cdc.RegisterConcrete(&MsgCreatePosition{}, "osmosis/cl-create-position", nil)
-	cdc.RegisterConcrete(&MsgWithdrawPosition{}, "osmosis/cl-withdraw-position", nil)
+	cdc.RegisterConcrete(&MsgCreatePosition{}, "osmosis/concentratedliquidity/create-position", nil)
```
mattverse, line 13: Can we make the amino codec as shortest as possible? Afaik we still have the word limit issue on ledger for amino codec msgs

x/concentrated-liquidity/model/codec.go
```diff
@@ -14,6 +16,12 @@ func RegisterCodec(cdc *codec.LegacyAmino) {
 }
 
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
+	registry.RegisterInterface(
+		"osmosis.concentratedliquidity.v1beta1.PoolI",
```
mattverse, line 20: I might be missing context, why not register as swapRouter PoolI as is?

mattverse (commented):
Left some minor questions and comments please take a look!

x/concentrated-liquidity/model/codec.go
czarcas7ic: Not quite sure what you mean? All I know is that PoolI was not registered in this module, so when we returned PoolI from pool queries, we were not able to unmarshal the response

x/concentrated-liquidity/types/codec.go
czarcas7ic: Was not aware of this, will change it, thanks!

x/concentrated-liquidity/model/codec.go
p0mvn: It think we should make this consistent with:
https://github.com/osmosis-labs/osmosis/blob/a1c13a6ebd92fdb46f2673f1702f5a430264964e/x/gamm/pool-models/balancer/codec.go#L24

czarcas7ic: Addressed

AlpinYukseloglu (approved):

x/concentrated-liquidity/model/codec.go
```diff
@@ -5,14 +5,22 @@ import (
...
+	cdc.RegisterConcrete(&Pool{}, "osmosis/cl-pool", nil)
+	cdc.RegisterConcrete(&MsgCreateConcentratedPool{}, "osmosis/cl-create-pool", nil)
 }
 
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
+	registry.RegisterInterface(
+		"osmosis.swaprouter.v1beta1.PoolI",
+		(*swaproutertypes.PoolI)(nil),
+		&Pool{},
+	)
```
p0mvn, line 22: Why do we marshal an interface though? I think this could have been avoided if we marshaled a pool struct directly

p0mvn: The reason we marshal an interface in gamm is that there are 2 implementations - balancer and stableswap. Here, we only have one pool model. So we should be able to marshal its concrete type directly

mattverse: @p0mvn wdym by marshalling a pool directly?

czarcas7ic: I'm pretty sure he means not marshaling into proto any and instead directly marshaling into the struct itself, but sometimes this isnt possible due to circular dependencies in the proto files

ValarDragon: Hrmm, it potentially makes sense to register this, if we wanted a swaprouter query for underlying structs as an interface tho

ValarDragon: But I do agree, that serialization into state should just be the concrete struct, and such a register would just be for a one-day hypothetical query

ValarDragon: FWIW, I don't really care if we serialize into state as concrete struct vs interface.

Concrete is slightly more space efficient, but I doubt this efficiency ever matters. Interface could plausibly give more ease of switching in the future, but if we didn't do an interface I don't think it'd be that hard to change in the future either.

x/concentrated-liquidity/model/codec.go
```diff
@@ -5,14 +5,22 @@ import (
...
 	"github.com/cosmos/cosmos-sdk/types/msgservice"
+
+	swaproutertypes "github.com/osmosis-labs/osmosis/v13/x/swaprouter/types"
 )
 
 func RegisterCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&Pool{}, "osmosis/concentratedliquidity/ConcentratedLiquidityPool", nil)
-	cdc.RegisterConcrete(&MsgCreateConcentratedPool{}, "osmosis/concentratedliquidity/create-concentrated-pool", nil)
+	cdc.RegisterConcrete(&Pool{}, "osmosis/cl-pool", nil)
+	cdc.RegisterConcrete(&MsgCreateConcentratedPool{}, "osmosis/cl-create-pool", nil)
```
ValarDragon, line 14: did we hit the 40 char limit before or smth?

ValarDragon (approved):

ValarDragon:
Agreed, I think we need this for generalized pool queries, even if we do concrete serialization within CL

x/concentrated-liquidity/model/codec.go
czarcas7ic: Matt was just telling me to be weary of the lengths of these, so I made it as short as possible to get the point across

Issue #3801, opened 2022-12-20
feat(x/twap): geometric twap CLI


p0mvn:
## Background

We need geometric twap CLI. It is currently blocked on: https://github.com/osmosis-labs/osmosis/pull/3696

## Suggested Design

`GetQueryTwapCommand` is refactored to take a string argument that represents a type "geometric" or "arithmetic" with shorthands of "g" and  "a" respectively. Any other value for type errors.

## Acceptance Criteria
- CLI is implemented

Issue #3541, opened 2022-11-27
test(twap): edge case tests for geometric twap


p0mvn:
## Background

Currently, most of the test cases for `computeGeometricTwap` use base values.

To be more confident about correctness, we should add more test cases with hand-picked custom values.

This is an open-ended task so someone having context on the state of geometric twap should be working on this.

## Acceptance Criteria
- custom tests added if applicable

p0mvn:
need to test this function: https://github.com/osmosis-labs/osmosis/blob/f9c3e06d5be473b66b92cbea4c05c369b0c793e0/x/twap/strategy.go#L39

where arithmeticMeanOfLogPrices is negative. It can happen when sp0 is < 1

p0mvn:
Revisit this comment while working on this issue: https://github.com/osmosis-labs/osmosis/pull/3809/files#r1053921196

p0mvn:
We need to test underflows.

Context:

        Do we also need to test min spot price, since this can get underflow? (or is that a future PR?)

_Originally posted by @ValarDragon in https://github.com/osmosis-labs/osmosis/pull/3809#discussion_r1054714992_

p0mvn:
TODO: perform manual CLI test on completion

p0mvn:
Everything here has been complete but the manual test. Tracking manual test here: https://github.com/osmosis-labs/osmosis/issues/3941

PR #3812, opened 2022-12-21
feat(x/twap): modify cli to add geometric option

p0mvn:
Closes: #3801 

Adds geometric twap CLI by implementing a new command with the same args as the old arithmetic command.

Did basic test checking that command was correctly wired up to the binary. I'm tracking proper manual localosmosis tests to be done separately in the future. This will be done once edge case tests are added: https://github.com/osmosis-labs/osmosis/issues/3541)


ValarDragon:
I'd rather add this as a flag than an arg

x/twap/client/cli/query.go
```diff
@@ -85,6 +119,30 @@ Example:
 	return cmd
 }
 
+// getQuoteDenomFromLiquidity gets the quote liquidity denom from the pool. In addition, validates that base denom
+// exists in the pool. Fails if not.
+func getQuoteDenomFromLiquidity(ctx context.Context, clientCtx client.Context, poolId uint64, baseDenom string) (string, error) {
```
p0mvn, line 124: note: this was extracted from pre-existing code. No additional changes were made

p0mvn:
> I'd rather add this as a flag than an arg

I made a separate command while trying to minimize code duplication. Felt like better UX. Please let me know what you think

ValarDragon (approved):
Nice

ValarDragon:
I worry about breaking the old CLI though, can we perhaps setup a viper alias?

p0mvn:
Alias added. Going to merge

PR #3814, opened 2022-12-21
chore: parallelize tests

faddat:

This PR parallelizes osmosis tests.

## Brief Changelog

add a call to t.Parallel() to every test, except for those related to wasm

## Testing and Verifying

```bash
time go test ./... -count=1 -parallel=$(nproc)
```

Intel 12600k, 64GB RAM:

Before:

real    2m50.129s
user    6m30.750s
sys     0m41.247s

After:

real    0m34.114s
user    4m28.669s
sys     0m30.082s

  - How is the feature or change documented? NA


Makefile
```diff
@@ -245,7 +245,7 @@ TEST_PACKAGES=./...
 
 test: test-unit test-build
 
-test-all: check test-race test-cover
```
ValarDragon, line 248: What even is check lol

faddat:
no, not really -- but I am not too surprised.  When I did this with the SDK, there were for sure some unforeseen consequences.  Let me poke around some. 

Happy new year!

ValarDragon:
I suggest we close this, and instead place go test speedup emphasis on GC reduction. ( https://github.com/osmosis-labs/osmosis/issues/3957 ) The observed local speedup is from skipping simulation tests running on save, which I think we should try to fix by making it either:

* quick by default
* behind some sort of build flag

PR #3788, opened 2022-12-19
Move osmoutils into own mod

nicolaslara:
## What is the purpose of the change
#3771 started moving some of our tools to their own modules. This PR expands that by moving osmoutils. 

Both of these are a requirement for https://github.com/osmosis-labs/osmosis/pull/3747 

## Brief Changelog

*(for example:)*
 
  - Move osmoutils into its own go.mod. To update, change your osmoutils import from `github.com/osmosis-labs/osmosis/v13/osmoutils` to `github.com/osmosis-labs/osmosis/osmoutils`
  - Moved osmoassert to osmoutils
  - Fix Exp2 tests to use assets from osmoutils

## Testing and Verifying

All tests still pass

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #3817, opened 2022-12-21
Move osmoassert into osmoutils, remove osmoutils twap dependence

ValarDragon:

Pre-requisite for moving osmoutils into its own go.mod


ValarDragon:
Merging as sub-component of #3788

Issue #2653, opened 2022-09-08
tests/e2e: revert logging all tx results


ValarDragon:
## Background

Not sure when it was added, but the e2e test suite is now logging all tx results. This should be reverted, as it massively crowds out the logs, and makes it hard to get relevant info at a glance. (Perhaps at most logging a tx hash by default, so you can quer)

If this is helpful for a particular debugging flow right now, we can add ways to improve that same debugging flows. (e.g. display results for last 3 txs)

p0mvn:
I think this has significantly improved the debugging experience. If these are not present, it is close to impossible to tell what exactly happened in the failing container / why did a certain CLI command fail.

In the long term,  we should instead pipe logs from each container to a file and preserve them as artifacts after running the CI job. This would remove the need for the extra logging. However, in the meantime, I think having more logs is better than having none

ValarDragon:
This is the log from the first TX I saw in the logs: https://pastebin.com/c94JZvwF

Its about 400 lines. Not sure which part of this is helpful for debugging in the normal case, beyond `txhash, stdout, stderr`. We can get the remainder of events & log from querying the txhash, but I don't see how this helps the standard case

p0mvn:
> Its about 400 lines. Not sure which part of this is helpful for debugging in the normal case, 

Yes, this is not useful at all for regular cases. However, the goal here is to log so that we can see the reasons for failure _when it fails_
 
> WDYT about only logging this for a tx that fails, but not all succesful txs?

This is exactly the problem - it is hard to tell when a transaction is successful and when it is not **in a general way**

Off the top of my head, the problems are:
- success is sometimes logged to stdout, other times in stderr
- Some CLI commands log "error" to the output when they are successful so it is non-trivial to do some regex/pattern matching.

I'm open to investigating identifying the tx fails more. However, from my previous attempts, there are quite a few edge cases to work around. I can try looking into this again after v12

PR #3822, opened 2022-12-22
Lower noise of e2e tests

ValarDragon:
Closes: #2653

Debugging E2E via CI is pretty unusable. It emits over 13000 log lines right now.

This PR:
* Turns the debug flag off
* Sets it s.t. in event of a tx / general action failing in E2E, we output stdout & stderr as was previously done.

I claim that this will massively reduce log output, and maintain the same debuggability.

## Brief Changelog

- Reduce E2E noisiness in CI

## Testing and Verifying

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,7 @@ test-e2e: e2e-setup test-e2e-ci
 # does not do any validation about the state of the Docker environment
 # As a result, avoid using this locally.
 test-e2e-ci:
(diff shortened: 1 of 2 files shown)


czarcas7ic (approved):
Agree with this change

p0mvn (approved):
There are cases when an actual error is produced neither to stdout, nor stderr. In such cases, if e2e happens to fail in CI but not locally, it becomes nearly impossible to debug without spitting out logs.

However, I guess we can turn on the debug flag temporarily and rerun the CI then

ValarDragon:
Feels good only 400 lines now :pray:

PR #3838, opened 2022-12-22
Upgrade IBC to v4.2.0

nicolaslara:

Update ibc to v4. This is required to make the wasm hooks reusable.

## Brief Changelog

  - Upgrade IBC to v4.2.0
  - (as a consequence) Upgrade wasm dependency to v0.30.0

## Testing and Verifying

All tests still pass

  - How is the feature or change documented? (not applicable )


osmoutils/ibc.go
```diff
@@ -3,11 +3,25 @@ package osmoutils
...
+	channeltypes "github.com/cosmos/ibc-go/v4/modules/core/04-channel/types"
+	ibcexported "github.com/cosmos/ibc-go/v4/modules/core/exported"
 )
 
+// NewStringErrorAcknowledgement returns a new instance of Acknowledgement using an Acknowledgement_Error
+// type in the Response field. These errors differ from the IBC errors in that we use custom error strings.
+// NOTE: Acknowledgements are written into state and thus, changes made to error strings included in packet acknowledgements
+// risk an app hash divergence when nodes in a network are running different patch versions of software.
+func NewStringErrorAcknowledgement(err string) channeltypes.Acknowledgement {
+	// ToDo: Do we want to do this or do we want to use the IBC errors and emit the string?
```
ValarDragon, line 16: I don;t understand the situation enough to have any opinions here

x/ibc-hooks/ibc_module.go
```diff
@@ -44,12 +44,12 @@ func (im IBCMiddleware) OnChanOpenInit(
 		hook.OnChanOpenInitBeforeHook(ctx, order, connectionHops, portID, channelID, channelCap, counterparty, version)
 	}
 
-	err := im.App.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, channelCap, counterparty, version)
+	finalVersion, err := im.App.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, channelCap, counterparty, version)
```
ValarDragon, line 47: Not sure what this final version is, but ACK that its wired right

x/ibc-rate-limit/testutil/chain.go
```diff
@@ -76,12 +75,8 @@ func SignAndDeliver(
 	)
 
 	// Simulate a sending a transaction and committing a block
-	app.BeginBlock(abci.RequestBeginBlock{Header: header})
 	gInfo, res, err := app.Deliver(txCfg.TxEncoder(), tx)
 
-	app.EndBlock(abci.RequestEndBlock{})
-	app.Commit()
```
ValarDragon, line 83: LGTM, though I don't see any relation to rest of the PR

ValarDragon (approved):
LGTM

osmoutils/ibc.go
damiannolan: is there any particular reason you guys have chosen to use this in favour of [channeltypes.NewErrorAcknowledgement](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/acknowledgement.go#L32)? 

Acknowledgements are written into the ibc store on `RecvPacket` so that the counterparty can verify their existence on `AcknowledgePacket`, and delete commitments (and in the case of ordered channels, increment an expected sequence).

Suppose an error string which is included in a packet acknowledgment is modified in a patch version of an osmosis release. Nodes running lets say `v14.0.0` and those running `v14.0.1` will produce different app hashes.

We made this change to `NewErrorAcknowledgement` in ibc-go/v4 as we deemed it unsafe as developers may unknowingly change a string value which could lead to the above. Also, to note, error values returned from protobuf codecs `UnmarshalJSON` produce non-deterministic output. So for example, when unmarshaling packet data `[]byte` to a well known type in `OnRecvPacket` its not safe to include the error in the ack. See for example we [omit the error in transfer acks](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L176-L179). I think you guys are using `encoding/json` in places here tho, which I hope is a bit more sane than the proto shenanigans.

x/ibc-rate-limit/testutil/chain.go
nicolaslara: This function overrides a function from the ibc testutils. They changed it so begin/end block must be called before and after, so if we keep this here all tests fail because of calling those twice

osmoutils/ibc.go
nicolaslara: ☝️ @ValarDragon This is the reason. I think it makes sense to change it, but I didn't want to do this without a discussion because it changes the behaviour (errors in ack vs errors in events).

I think a middle way is a good option, where we can add some information in the acks, but only from this packet and not from underlying errors.

I don't se why unmarshalling the error would lead to non-determinism as it's just a string (though I'm sure I'm wrong)

damiannolan: IIRC its because the `jsonpb` unmarshaling [uses maps](https://github.com/golang/protobuf/blob/master/jsonpb/decode.go#L290-L303). Ranging on these maps in the deserialisation process will be non deterministic and therefore, you may get a different key returned in the error msg for a particular struct value.

x/ibc-hooks/ibc_module.go
nicolaslara: `OnChanOpenInit` returns the version. I'm also not sure, but I think that is the modified version. I'm passing both to the After method in case it needs to act on either. We could also ignore it and if someone needs it they can use Override... just thought it was weird not to pass it down since we're passing err.

osmoutils/ibc.go
nicolaslara: that makes sense. And I think that aligns with only using string errors if they do not rely on errors from other parts of the stack. 

I like having data in acks (errors in this case, but the success could also have data) because it allows us to do communication in cases where events aren't available (cosmwasm). I think the Async-ICQ implementation does something similar. But you're right about the risk of halts if we're not careful when changing these. 

I should refactor the code so that it's very hard to accidentally introduce changes to the merkelized errors in minor versions. This change will probably need to happen both here and in the rate limit middleware/contract

PR #3839, opened 2022-12-22
Gamm stableswap improvements

ljah8:
This PR is created during the Informal Systems audit, after analysis of the existing specification and code inspection
Auditing is performed on commit hash: [7374795](https://github.com/osmosis-labs/osmosis/commit/7374795e0de22f3a291ca59c5faffa7851acf3bd)
This PR contains minor Stableswap code improvements within the GAMM module noticed during code inspection with the aim of finding issues.

## Brief Changelog
* *stableswap/pool.go*:
  - Part of the code used to build errors is moved to errors.go;
  - The `getLiquidityIndexMap` and `GetScalingFactorByLiquidityIndex` functions are replaced by the `GetScalingFactorByDenom` function to get the appropriate scaling factor faster and avoid creating a map. While implementing the new function, we didn't want to change logic, so it doesn't return an error or panic if no scaling factor is found.
  - `applyScalingFactorMultiplier` function is protected against unwanted overflow in case multiplier > 1
* Minor changes in the implementation of `BinarySearchBigDec`, `BinarySearch` (*osmoutils/binary_search.go*) and `joinPoolSharesInternal` (*stableswap/amm.go*) functions.
* Removed redundant error from `iterKCalculator` function that always returned nil.
* Added checking for potential division by zero in `Compare` and `CompareBigDec` functions.

## Testing and Verifying

* Added a unit test covering the `GetScalingFactorByDenom` function.
* Existing tests are modified in accordance with the new changes.


x/gamm/pool-models/stableswap/amm.go
```diff
@@ -38,10 +38,6 @@ func cfmmConstant(xReserve, yReserve osmomath.BigDec) osmomath.BigDec {
 // We use this version for calculations since the u
 // term in the full CFMM is constant.
 func cfmmConstantMultiNoV(xReserve, yReserve, wSumSquares osmomath.BigDec) osmomath.BigDec {
-	if !xReserve.IsPositive() || !yReserve.IsPositive() || wSumSquares.IsNegative() {
-		panic("invalid input: reserves must be positive")
-	}
```
AlpinYukseloglu, line 43: I think this check is worth keeping – this module deals with enough math where ensuring basic invariants are never broken (even if some checks are superfluous) is well worth it

AlpinYukseloglu (approved):
LGTM, thanks for the cleanup!

x/gamm/pool-models/stableswap/amm.go
ljah8: I understand, but cfmmConstantMultiNoV first calls cfmmConstantMultiNoVY which has the same checks at the beginning.

AlpinYukseloglu: Fair enough! I'm happy with removing then :)

ValarDragon:
LGTM! I had previously suggested backporting this.

Even though this looks like it should be fine, changing to no backport just to be abundandantly cautious with the division by zero change. (I'm _pretty_ sure it should be impossible to get a zero in current code)

PR #3844, opened 2022-12-23
test(x/twap): add randomized geometric twap test on a balancer pool

p0mvn:
Closes: #XXX

This PR adds a geometric twap to now randomized test against a balancer pool. It creates 1000 pools with randomized supply and executes twap over random duration for these pools.

The result is asserted to be within "0.0001" relative to spot price.

This test serves as a sanity check that we do not encounter any internal panics in math calculations

## Testing and Verifying

This change added tests.


x/twap/api_test.go
```diff
@@ -837,3 +841,70 @@ func (s *TestSuite) TestGetArithmeticTwapToNow_ThreeAsset() {
...
+					Weight: weightB,
+				},
+			}
+
+			s.PrepareCustomBalancerPool(assets, balancer.PoolParams{
+				SwapFee: sdk.ZeroDec(),
+				ExitFee: sdk.ZeroDec(),
+			})
+
+			//
```
stackman27, line 890: Were you going to add a comment here?

x/twap/api_test.go
```diff
@@ -837,3 +841,70 @@ func (s *TestSuite) TestGetArithmeticTwapToNow_ThreeAsset() {
...
+			oldTime := ctx.BlockTime().Add(1 * time.Millisecond)
```
stackman27, line 891: Why can't oldTime just be the current blockTime? Why add 1 * time.Millisecond?

x/twap/api_test.go
```diff
@@ -837,3 +841,70 @@ func (s *TestSuite) TestGetArithmeticTwapToNow_ThreeAsset() {
...
+
+			ctx = ctx.WithBlockTime(newTime)
+
+			spotPrice, err := app.GAMMKeeper.CalculateSpotPrice(ctx, 1, denom1, denom0)
+			s.Require().NoError(err)
+
+			twap, err := app.TwapKeeper.GetGeometricTwapToNow(ctx, 1, denom0, denom1, oldTime)
+			s.Require().NoError(err)
+
+			osmomath.ErrTolerance{
```
stackman27, line 902: Out of curiosity, why not check multiplicative tolerance too?

p0mvn: Additive just gives a better idea of how many precision points there are here for all test cases. The goal of this test is just to see that twap is roughly close to spot price. So any reasonable tolerance would suffice.

x/twap/api_test.go
p0mvn: Updated:

```
// We add 1ms to avoid always landing on the same block time
// In that case, the most recent spot price would be used
// instead of interpolation.
```

I think that answers your other question as well

x/twap/api_test.go
p0mvn: https://github.com/osmosis-labs/osmosis/pull/3844#discussion_r1056018157

x/twap/api_test.go
stackman27: sounds good

ValarDragon: Hrmm, I think the guarantee we want is actually the multiplicative tolerance though. Since spot prices can differ in orders of magnitude, this may be the wrong order of magnitude for checking many of the pools.

x/twap/api_test.go
```diff
@@ -837,3 +841,75 @@ func (s *TestSuite) TestGetArithmeticTwapToNow_ThreeAsset() {
...
+				AdditiveTolerance: sdk.NewDecWithPrec(1, 4),
+			}.CompareBigDec(
+				osmomath.BigDecFromSDKDec(spotPrice),
+				osmomath.BigDecFromSDKDec(twap),
+			)
```
czarcas7ic, line 912: Do you expect the additive tolerance to go both ways? My guess is yes, but I just wanted to be sure.

czarcas7ic (approved):
Despite me not having a ton of context on geo twap, the test was easy to follow. Nicely done 👍

x/twap/api_test.go
p0mvn: There are no guarantees on the direction of the tolerance because exp2 function does not guarantee it. However, log always rounds down.

So I would say it is more likely to be smaller than higher but I don't think we can claim that for all test cases

x/twap/api_test.go
p0mvn: Yeah but the largest spot price possible in this setup is `1_000_000_000 / 10_000  =  100_000` which should not cause an extremely high additive difference.

I think I will adjust the parameters to allow for even greater spot prices and switch to multiplicative tolerance

p0mvn:
Addressed comments.

The initial supply range of valid values has been increased. Multiplicative tolerance is now used. Also, reduced the number of retries to 200 only to avoid increasing CI time by over 2 seconds

PR #3845, opened 2022-12-23
refactor(x/twap): handle spot price error case in the context of geometric twap

p0mvn:
Closes: #XXX

This PR handles spot price errors in the context of geometric twap. When a spot price error occurs, twap module sets sp0 and sp1 to zero.

In geometric calculation logic, we take the log of sp0. Since logarithm of zero is undefined, we should be careful not to do it.

Therefore, we handle this edge case by not updating the geometric accumulator and setting last error time to the `newTime`

Additionally, it includes test cases for sig fig rounding and spot price that I use to facilitate my understanding of when errors are returned.

## Testing and Verifying

Added tests

### DIFF
--- a/x/twap/logic.go
+++ b/x/twap/logic.go
@@ -194,6 +194,14 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 	p1NewAccum := types.SpotPriceMulDuration(record.P1LastSpotPrice, timeDelta)
 	newRecord.P1ArithmeticTwapAccumulator = newRecord.P1ArithmeticTwapAccumulator.Add(p1NewAccum)
 
+	// If the last spot price is zero, then the logarithm is undefined.
+	// As a result, we cannot update the geometric accumulator.
+	// We set the last error time to be the new time, and return the record.
+	if record.P0LastSpotPrice.IsZero() {
+		newRecord.LastErrorTime = newTime
+		return newRecord
+	}
+
 	// logP0SpotPrice = log_{2}{P_0}
 	logP0SpotPrice := twapLog(record.P0LastSpotPrice)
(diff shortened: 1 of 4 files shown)


x/twap/logic.go
```diff
@@ -259,7 +267,12 @@ func computeTwap(startRecord types.TwapRecord, endRecord types.TwapRecord, quote
 }
 
 // twapLog returns the logarithm of the given spot price, base 2.
+// Panics if zero is given.
 func twapLog(price sdk.Dec) sdk.Dec {
+	if price.IsZero() {
+		panic("twap: cannot take logarithm of zero")
+	}
```
ValarDragon, line 274: This LGTM, but I feel like the mental overhead is better here if we make this return an error in such case instead. (We have to be careful that this panic is never called due to this being end block logic. Vs if its an error, we can just check that caller handles errors correctly, which is already the case w/ our errcheck linter)

ValarDragon (approved):

czarcas7ic (commented):
Not approving since I lack context, but just letting you know I reviewed and the logic all LGTM

x/twap/logic.go
p0mvn: I don't think this panic can ever be reached because we have an additional safeguard in its only caller:
https://github.com/osmosis-labs/osmosis/blob/a15947d8f16c25a750a092c2c82f95bf00dc3732/x/twap/logic.go#L200-L203

This is just a defense-in-depth sanity check.

The problem with returning an error is that it would require a large refactor to propagate this error to multiple callers of `recordWithUpdatedAccumulators`.

I'm going to keep this as is and merge it. However, please let me know if you have further concerns, happy to discuss and address in another PR if needed

PR #3846, opened 2022-12-23
refactor(x/twap): computeTwap edge cases and tests

p0mvn:
Progress towards #3541

This PR focuses on refactoring geometric `computeTwap` to account for edge cases and returning errors if any are found.

The following edge cases are tested here:
- We are able to calculate geometric twap given a max spot price * 100 years
- We are not able to calculate geometric twap with (1 / max spot price) * 100 years
   * this is due to losing precision and falling under `sdk.Dec`'s precision of 18
   * return error instead
- Currently, spot price of zero is set in twap on any spot price error. When that happens, the accumulators are not updated. As a result, we must handle this test case. Here, we validate that geometric twap still works under this condition

TODO:
- [ ] underflow test


p0mvn:
This was separated into other PRs and merged

PR #3853, opened 2022-12-24
Accumulator: `Set` and `Get`

AlpinYukseloglu:

This PR sets up the Accumulator API package and implements its core types and set/get functions. The remaining work is to implement the commented functions in `osmoutils/accum/accum.go`

## Brief Changelog

- Set up package under osmoutils
- Set up types and relevant proto
- Created test suite and added sanity tests

## Testing and Verifying

- The implemented logic is tested in `accum_test.go`


osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/cosmos/cosmos-sdk/store"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/gogo/protobuf/proto"
+)
+
+// We keep this object as a way to interface with the methods, even though
+// only the Accumulator inside is stored in state
+type AccumulatorObject struct {
+	// Store where accumulator is stored
+	Store store.KVStore
```
stackman27, line 16: Out of curiosity, but why do we need to know where the accumulator is stored? Can't we just store the key and reference that key? 

aah is it cause we usually store the key in keeper but we don't have that here?

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return AccumulatorObject{}, err
+	}
+
+	accum := AccumulatorObject{accum_store, accum_content.AccumName, accum_content.AccumValue}
+
+	return accum, nil
+}
+
+func setAccumulator(accum_store store.KVStore, accum AccumulatorObject, amt sdk.Dec) error {
```
stackman27, line 67: nit: 
Why pass in accum AccumulatorObject as func parameter, if we're only trying to get accum.Name

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+		return err
+	}
+
+	accum_store.Set(keybz, bz)
+
+	return nil
+}
+
+// TODO: consider making this increment the accumulator's value instead of overwriting it
+func (accum AccumulatorObject) UpdateAccumulator(amt sdk.Dec) {
```
stackman27, line 86: thoughts on making all the functions use this struct method design so that you don't have to pass `accum_store store.KVStore` as parameter all the time

stackman27 (commented):
Left few comments! :)) 

Is there going to be multiple PRs for the TODOs or is everything is going to be in this PR?

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+	// Accumulator's name (pulled from AccumulatorContent)
+	Name string
+
+	// Accumulator's current value (pulled from AccumulatorContent)
+	Value sdk.Dec
+}
+
+// Makes a new accumulator at store/accum/{accumulator_name}
+// returns an error if already exists / theres some overlapping keys
+func MakeAccumulator(accum_store store.KVStore, accum_name string) error {
```
p0mvn, line 27: let's name parameters in camel case for style consistency. For example, `accumStore`. Relevant to many other identifiers in the PR

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+	keybz := []byte(accum_name)
+	if accum_store.Has(keybz) {
+		return errors.New("Accumulator with given name already exists in store")
```
p0mvn, line 30: nit: Go does not perform redundant reallocations when casting is done directly via argument 
```suggestion
	if accum_store.Has([]byte(accum_name)) {
		return errors.New("Accumulator with given name already exists in store")
```

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+	if !accum_store.Has(keybz) {
+		return AccumulatorObject{}, errors.New(fmt.Sprintf("Accumulator name %s does not exist in store", accum_name))
+	}
+
+	var accum_content AccumulatorContent
+	bz := accum_store.Get(keybz)
+	err := proto.Unmarshal(bz, &accum_content)
+	if err != nil {
+		return AccumulatorObject{}, err
+	}
```
p0mvn, line 60: We can use `osmoutils.MustGet` instead

osmoutils/accum/accum.go
p0mvn, line 30: let's also format the name for ease of future debugging

osmoutils/accum/accum.go
p0mvn: I think the reason for passing in the store to `setAccumulator` is to share it between the following callers: `MakeAccumulator` function and `UpdateAccumulator` method.

`MakeAccumulator` is a function that takes the store as a parameter by design. As a result, if we want to share any logic between it and other methods, the shared function must also take in the store as a param.

osmoutils/accum/accum.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// We keep this object as a way to interface with the methods, even though
+// only the Accumulator inside is stored in state
+type AccumulatorObject struct {
+	// Store where accumulator is stored
+	Store store.KVStore
+
+	// Accumulator's name (pulled from AccumulatorContent)
+	Name string
```
p0mvn, line 19: We might want to lowercase these for better information hiding

proto/osmosis/accum/v1beta1/accum.proto
```diff
@@ -0,0 +1,30 @@
...
+  string accum_name = 1;
+  string accum_value = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message Record {
+  string num_shares = 1 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
```
p0mvn, line 19: Thoughts on making this decimal for future flexibility? If I'm not mistaken, even CL shares might be decimal as of right now

osmoutils/accum/accum.go
p0mvn, line 86: Can we distribute multiple rewarsd? E.g. `uosmo` and `uatom`. Maybe the accuulator should be `sdk.DecCoins` instead of `sdk.Dec`, thoughts?

osmoutils/accum/accum.go
AlpinYukseloglu: Agreed on this – was mainly following our spec and some written conventions from older issues/PRs but camel case does feel more appropriate

osmoutils/accum/accum.go
AlpinYukseloglu: Great point – I think `sdk.DecCoins` would be a little odd since the accumulator is not tracking coins (it's more like coins per share), but we could potentially support multiple accumulators per set of records so we could distribute different tokens to the same positions.

proto/osmosis/accum/v1beta1/accum.proto
AlpinYukseloglu: Good point! I was thinking of shares as fungible LP shares (which have to be ints), but since in CL the definition of a share is a bit more specific Dec makes a lot of sense here

osmoutils/accum/accum.go
AlpinYukseloglu: Well the store is passed into the make and get functions since we don't have the key here yes, but we keep it in this accumulator object so that the caller doesn't have to constantly pass in their store for every function.

i.e. intended workflow: call get function with store -> call every other function on the returned object, not having to worry about passing in the correct store for each

Note that only `AccumulatorContent` is marshaled and placed into the store – `AccumulatorObject` is simply a wrapper for convenience

osmoutils/accum/accum.go
AlpinYukseloglu: I see where you're coming from – is there a benefit to information hiding over the convenience (and potential utility) of being able to access e.g. the name/value of the accumulator in the caller? Having them public also simplifies testing a bit although that's a smaller issue.

osmoutils/accum/accum.go
AlpinYukseloglu: Actually the better change here would be to remove the store parameter as `AccumulatorObject` was updated to include the store for convenience. 

The intuition for passing the whole object in instead of necessary components in that case is 
1. fewer params and 
3. the accumulator object is the one that's being "set" by the function (i.e. multiple accumulators can have the same name as long as they are in different stores)

osmoutils/accum/accum.go
p0mvn: We can define exported getters for `Name` and `Value`.

I think that the main benefit of information hiding here is to prevent accidental modification. If users can only access these fields via getters, there is no way to mistakenly alter the accumulator value.

Also, I think that `Store` is better to be inaccessible from a security standpoint as well. One shouldn't really need to access the store outside of the internal accumulator logic

osmoutils/accum/accum.go
AlpinYukseloglu: Nevermind, came around on this!

osmoutils/accum/accum.go
AlpinYukseloglu: Agreed (and discussed on call) – done!

osmoutils/accum/accum.go
AlpinYukseloglu: Agreed, updated

p0mvn (approved):
LGTM. Nice work!

AlpinYukseloglu:
> Left few comments! :))
> 
> Is there going to be multiple PRs for the TODOs or is everything is going to be in this PR?

We'll be cleaning up the TODOs as we implement the rest of the package! Still converging on the details of the architecture (many of them are addressed in #3855 already)

PR #3854, opened 2022-12-24
fix(scripts): proto gen for osmoutils

p0mvn:
Closes: #XXX

Moving `osmoutils` to its own package has caused import paths changed. As a result, the proto generation script got broken

This PR fixes the problem and updates previously outdated protos


AlpinYukseloglu (approved):
Great fix!

scripts/protocgen.sh
```diff
@@ -25,6 +25,7 @@ cd ..
 #
 # Note: Proto files are suffixed with the current binary version.
 cp -r github.com/osmosis-labs/osmosis/v13/* ./
+cp -r github.com/osmosis-labs/osmosis/osmoutils ./
```
p0mvn, line 28: Should be this
```suggestion
cp -r github.com/osmosis-labs/osmosis/osmoutils ./osmoutils
```

PR #3855, opened 2022-12-24
feat(osmoutils): ClaimRewards and NewPosition foundations

p0mvn:
Closes: #XXX

Outlines initial logic for `NewPosition` and `ClaimRewards`. Makes other pass-through changes such as:

- change accumulators to `sdk.DecCoins`. This is done so that we can distribute and track rewards in multiple denominations
- changes `num_shares` to be a `Dec` instead of an `Int`. This allows for future flexibility to have non-int shares if needed. Also, simplifies calculations `ClaimRewards` since we can multiply `DecCoins` by a `Dec` directly
- remove `sendKeeper` from `ClaimRewards` API
   * it is counter-cohesive to pass in a keeper to an accumulator. Accumulator logic should just handle the computation logic while the caller can perform the bank send 
- remove `accum_name` from `AccumulatorContent`. There is no practical need for storing that in store
- implement store prefixes
   * a prefix for accumulators
   * a prefix for positions

## Testing and Verifying

- Added `TestNewPosition`
- Added `TestClaimRewards`

  - Does this pull request introduce a new feature or user-facing behavior changes? (no


osmoutils/accum/accum.go
```diff
@@ -83,16 +88,49 @@ func setAccumulator(accum_store store.KVStore, accum AccumulatorObject, amt sdk.
...
+// NewPosition creates a new position for the given address, with the given number of units
+// It takes a snapshot of the current accumulator value, and sets the position's initial value to that
+// The position is initialized with empty unclaimed rewards
+func (accum AccumulatorObject) NewPosition(addr sdk.AccAddress, numShareUnits sdk.Dec, options PositionOptions) {
+	position := Record{
+		NumShares:        numShareUnits,
+		InitAccumValue:   accum.Value,
+		UnclaimedRewards: sdk.NewDecCoins(),
+	}
+	osmoutils.MustSet(accum.Store, addr, &position)
```
AlpinYukseloglu, line 104: Is it okay to use the address directly as the key? Should we prefix it with "accumulator-" or something to avoid conflicts with other things that might store directly into the address key?

osmoutils/accum/prefix.go
```diff
@@ -0,0 +1,24 @@
...
+// accum module values to be stored in the KVStore.
+// Returns "accum/{key}" as bytes.
+func formatModulePrefixKey(key string) []byte {
+	return []byte(fmt.Sprintf("accum/%s", key))
+}
+
+// formatAccimPrefix returns the key prefix used
+// specifically for accumulator values in the KVStore.
+// Returns "accum/acc/{name}" as bytes.
+func formatAccumPrefixKey(name string) []byte {
```
p0mvn, line 15: Note: while we only discussed prefixing positions, I think it is also useful to prefix the accumulators themselves in case we ever want to iterate over accumulators

osmoutils/accum/accum.go
```diff
@@ -20,20 +20,21 @@ type AccumulatorObject struct {
...
+	if accumStore.Has(formatAccumPrefixKey(accumName)) {
 		return errors.New("Accumulator with given name already exists in store")
 	}
 
 	// New accumulator values start out at zero
 	// TODO: consider whether this should be a parameter instead of always zero
-	initAccumValue := sdk.ZeroDec()
+	initAccumValue := sdk.NewDecCoins()
 
 	newAccum := AccumulatorObject{accumStore, accumName, initAccumValue}
```
p0mvn, line 39: Note: I did not find a way to update the store here with an "accum" prefix. Instead, we must use the new "format..." helpers from `prefix.go` every time we want to interact with the store.

osmoutils/accum/accum.go
p0mvn: Please see: https://github.com/osmosis-labs/osmosis/pull/3855#discussion_r1056886112

and the `prefix.go` file for relevant updates

osmoutils/accum/prefix.go
```diff
@@ -0,0 +1,24 @@
+package accum
+
+import "fmt"
+
+// formatAccimPrefix returns the key prefix used for any
```
p0mvn, line 5:
```suggestion
// formatAccumPrefix returns the key prefix used for any
```

osmoutils/accum/prefix.go
```diff
@@ -0,0 +1,24 @@
...
+// accum module values to be stored in the KVStore.
+// Returns "accum/{key}" as bytes.
+func formatModulePrefixKey(key string) []byte {
+	return []byte(fmt.Sprintf("accum/%s", key))
+}
+
+// formatAccimPrefix returns the key prefix used
```
p0mvn, line 12:
```suggestion
// formatAccumPrefix returns the key prefix used
```

AlpinYukseloglu (approved):
LGTM once tests are added!

p0mvn:
Discussed offline. Going to merge as is to unblock further work. The tests for these methods are to be added in a subsequent PR.

osmoutils/accum/prefix.go
```diff
@@ -0,0 +1,30 @@
...
+// Returns "accum/acc/{name}" as bytes.
+func formatAccumPrefixKey(name string) []byte {
+	return formatModulePrefixKey(fmt.Sprintf("%s/%s", accumulatorPrefix, name))
+}
+
+// formatPositionPrefixKey returns the key prefix used
+// specifically for position values in the KVStore.
+// Returns "accum/pos/{accumName}/{address}" as bytes.
+func formatPositionPrefixKey(accumName, address string) []byte {
+	return formatAccumPrefixKey(fmt.Sprintf("%s/%s/%s", positionPrefix, accumName, address))
```
p0mvn, line 29: @AlpinYukseloglu  I realized that the position prefix should also include the accumulator name so that positions are not applied to all accumulators.

This was caught in tests and is now fixed

osmoutils/accum/export_test.go
```diff
@@ -1,15 +1,64 @@
...
+}
+
+// GetAllPositions returns all positions associated with the receiver accumulator.
+// Returns error if any database errors occur.
+// This function is currently used for testing purposes only.
+// If there is a need to use this function in production, it
+// can be moved to a non-test file.
+func (accum AccumulatorObject) GetAllPositions() ([]Record, error) {
+	return osmoutils.GatherValuesFromStorePrefix(accum.store, formatPositionPrefixKey(accum.name, ""), parseRecordFromBz)
+}
```
p0mvn, line 31: Note: this is a test helper for getting all of the positions of an accumulator

p0mvn: I ended up not using it but leaving it in case you need @AlpinYukseloglu . If not, we can remove in a subsequent PR

AlpinYukseloglu: This is super useful for testing edge cases on `AddToPosition` and `RemoveFromPosition` – thanks for adding it!

osmoutils/accum/prefix.go
AlpinYukseloglu: Great catch – I did find it odd that we only needed the address (and not the accumulator) to get a position, so this resolves my concerns there.

AlpinYukseloglu (approved):
Great work (and nice catch on the prefix bug)!

PR #3858, opened 2022-12-26
feat(osmoutils): accumulator position options

p0mvn:
Closes: #XXX

This PR implements an attempt to handle position `Options`.

It does so by adding position options as a nullable field to `Record` that represents a position.

As a result, we are able to set these options to nil for fees and incentives `Record`s that do not require auto-compounding.

Once there is a need for auto-compounding to be implemented, we can start creating `Record`s with non-nil `Options` containing inflation information.

- Pros: 
    * should not require migrations if we are to add new position types with auto-bonding later. This is provided that we set nil for the options for all current `Record`s
- Cons:
    * Additional complexity from managing the logic to validate that the positions options are nil for regular positions.
    * Risk of needing to do migrations after we have records with non-nil options in the production state. Then, if we were to add more option fields, we would also likely have to migrate the existing positions
        * (Need to re-confirm if this is a valid risk, likely not)

## Alternatives

It is also possible to solve this by implementing various `Record` types and serializing records as interfaces to state. Similar to how we have `balancer` and `stableswap` pools serialized into state in `x/gamm`

- Pros:
    - cleaner in terms of not having to do extra validations for nil options
    - no need for migrations because can register a new type / concrete implementation
- Cons:
    - Additional complexity to have to register amino and codec for every module that imports accum package
    - Likely to have a lot of code and field duplication between different implementations of `Record`s

Knowing the additional overhead that comes with serializing interfaces, I opted in for the options in this PR. However, still open to discussing potential; alternative approaches.

## Testing and Verifying


AlpinYukseloglu (approved):
LGTM! Discussed tradeoffs offline and agree with the direction in this PR

PR #3863, opened 2022-12-26
Enforce SDK 45 in the dependencies

nicolaslara:

Using the ibc-hooks on another projects fails if we have the sdk dependency as sdk 0.46.

These changes pin the versions to the latest commit with so that `go mod tidy` doesn't reset the dep to 0.45.

If we want to use this, we may want to do new releases for each package

## Brief Changelog

  - Pin dependencies for packages so that the cosmos sdk dependency is pinned to 0.45

## Testing and Verifying
All tests should pass


ValarDragon:
I think the downstream project needs to do a replace tag

(btw is there a way to get it to stop updating our minor version on `go mod tidy`)

nicolaslara:
> I think the downstream project needs to do a replace tag.

That would solve it, but I don't see why we should specify a version greater than the one we support. The only reason this dependency gets updated on `go mod tidy` is because it got updated by sub dependencies, but once I changed them all to 0.45 now they play well together. 

I find it odd for downstream to have to do a replace when they already specify `github.com/cosmos/cosmos-sdk v0.45.11`

> btw is there a way to get it to stop updating our minor version on go mod tidy, if so down to change this

I couldn't find how

ValarDragon:
oh interesting, this sounds good to me then, post -conflicts being fixed!

ValarDragon (approved):
Checked that go mod tidy doesn't update it locally!  LGTM post conflicts being fixed

PR #3879, opened 2022-12-29
Add sender to wasm hooks

nicolaslara:

WasmHooks has been calling all contracts using the same sender (the wasmhooks module).

To give contracts more flexibility we use an intermediary address based on the hash of the channel and sender addr

### Open Questions

 * Is this adding any extra risk? Are collisions possible (or at least more likely)? Should we try to prove this?

## Brief Changelog

 - The sender of wasmhooks is now an addres based of the hash of the channel id and the sender's addr

## Testing and Verifying

All tests pass

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -442,7 +442,7 @@ func (appKeepers *AppKeepers) WireICS20PreWasmKeeper(
 	bApp *baseapp.BaseApp,
 	hooksKeeper *ibchookskeeper.Keeper) {
 	// Setup the ICS4Wrapper used by the hooks middleware
-	wasmHooks := ibchooks.NewWasmHooks(hooksKeeper, nil) // The contract keeper needs to be set later
+	wasmHooks := ibchooks.NewWasmHooks(hooksKeeper, nil, appKeepers.BankKeeper, "osmo") // The contract keeper needs to be set later
 	appKeepers.Ics20WasmHooks = &wasmHooks
 	appKeepers.HooksICS4Wrapper = ibchooks.NewICS4Middleware(
 		appKeepers.IBCKeeper.ChannelKeeper,
--- a/x/ibc-hooks/wasm_hook.go
+++ b/x/ibc-hooks/wasm_hook.go
@@ -1,8 +1,10 @@
 package ibc_hooks
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 
 	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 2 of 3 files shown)


nicolaslara:
This is state breaking because of the errors. Not sure if it's worth backporting anyway. No one is using this feature rn

app/upgrades/v13/upgrade_test.go
```diff
@@ -55,8 +54,6 @@ func (suite *UpgradeTestSuite) TestUpgrade() {
 		{
 			"Test that the upgrade succeeds",
 			func() {
-				acc := suite.App.AccountKeeper.GetAccount(suite.Ctx, ibc_hooks.WasmHookModuleAccountAddr)
-				suite.App.AccountKeeper.RemoveAccount(suite.Ctx, acc)
```
ValarDragon, line 59: Can we comment these lines, as indication to future us that they existed?

app/keepers/keepers.go
```diff
@@ -442,7 +442,7 @@ func (appKeepers *AppKeepers) WireICS20PreWasmKeeper(
 	bApp *baseapp.BaseApp,
 	hooksKeeper *ibchookskeeper.Keeper) {
 	// Setup the ICS4Wrapper used by the hooks middleware
-	wasmHooks := ibchooks.NewWasmHooks(hooksKeeper, nil) // The contract keeper needs to be set later
+	wasmHooks := ibchooks.NewWasmHooks(hooksKeeper, nil, "osmo") // The contract keeper needs to be set later
```
ValarDragon, line 445: Its silly, but the way to get the constant from SDK properly is:
```
sdk.GetConfig().GetBech32AccountAddrPrefix()
```

x/ibc-hooks/wasm_hook.go
```diff
@@ -66,13 +68,22 @@ func (h WasmHooks) OnRecvPacketOverride(im IBCMiddleware, ctx sdk.Context, packe
 		return osmoutils.NewStringErrorAcknowledgement("error in wasmhook message validation")
 	}
 
-	// The funds sent on this packet need to be transferred to the wasm hooks module address/
-	// For this, we override the ICS20 packet's Receiver (essentially hijacking the funds for the module)
+	// Calculate the receiver / contract caller based on the packet's channel and sender
+	senderStr := fmt.Sprintf("%s/%s", packet.GetDestChannel(), data.GetSender())
+	senderHash32 := sha256.Sum256([]byte(senderStr))
```
ValarDragon, line 73: We need to ensure this is properly namespaced away from any possible user address, or module account address.

Lets do so, by not hashing directly like this, but by instead using https://github.com/cosmos/cosmos-sdk/blob/main/types/address/hash.go

We get the safety guarantee we need, by just using some `type` field we think won't get re-used, perhaps `ibc-memo-action` or smth

ValarDragon (commented):
LGTM aside from my comments + needs spec update. (This is an important change!)

app/upgrades/v13/upgrade_test.go
nicolaslara: Added with comments

app/keepers/keepers.go
nicolaslara: ahh this is really useful! I was looking into how to do this! (I thought the sdk constant would be updated, but that just gives me "stake")

nicolaslara: > sdk.GetConfig().GetBech32AccountAddrPrefix()

Will this work on any chain? If so, we can remove the param and get it inside the function, but maybe it's good to have it configurable anyway?

ValarDragon: this _should_ work on any chain.

I don't have a preference between arg or computed locally

ValarDragon (approved):
LGTM!

Issue #3808, opened 2022-12-20
test(x/swaprouter): add a test to validate that swap routes serialize correctly


p0mvn:
Add a test to validate that old and new swap routes are equal after proto-serializing them.

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/3764#discussion_r1053827780_

PR #3880, opened 2022-12-29
refactor(x/swaprouter, x/gamm): unify swap routes in swaprouter

p0mvn:
Closes: #3808

This PR removes the swap routes in `x/gamm` and refactors all legacy `x/gamm` swap messages and methods to utilize the new swap route structs from the `x/swaprouter` module.

This is done to reduce code duplication and redundant conversions between the different swap routes.

## Brief Changelog
- removed proto declarations for all `x/gamm `swap routes but [`SwapExactAmountInRoute`](https://github.com/osmosis-labs/osmosis/blob/a3afcda5ca7694ea40a4fa573db19ffdae2d0a3b/proto/osmosis/gamm/v1beta1/tx.proto#L86-L90)
   * It is kept for the serialization / deserialization test but is to be removed in the future.
- substituted usage of old `x/gamm` swap routes with the `x/swaprouter` swap routes.

## Testing and Verifying

https://github.com/osmosis-labs/osmosis/blob/a3afcda5ca7694ea40a4fa573db19ffdae2d0a3b/x/gamm/types/route_test.go#L16


p0mvn:
While this is not state-breaking, I added the label still to be more careful with this PR as it has an important API break.

proto/osmosis/gamm/v1beta1/tx.proto
```diff
@@ -3,16 +3,17 @@ package osmosis.gamm.v1beta1;
...
 
 service Msg {
   rpc JoinPool(MsgJoinPool) returns (MsgJoinPoolResponse);
   rpc ExitPool(MsgExitPool) returns (MsgExitPoolResponse);
   rpc SwapExactAmountIn(MsgSwapExactAmountIn)
-      returns (MsgSwapExactAmountInResponse);
+      returns (MsgSwapExactAmountInResponse) {}
   rpc SwapExactAmountOut(MsgSwapExactAmountOut)
-      returns (MsgSwapExactAmountOutResponse);
+      returns (MsgSwapExactAmountOutResponse) {}
```
czarcas7ic, line 16: Why this change?

czarcas7ic (approved):

proto/osmosis/gamm/v1beta1/tx.proto
p0mvn: was trying to deprecate those and failed to revert properly. Going to fix

x/gamm/types/route_test.go
```diff
@@ -0,0 +1,26 @@
...
+	swapRouterExactAmountInRoute := swaproutertypes.SwapAmountInRoute{}
+	gammExactAmountInRoute := gammtypes.SwapAmountInRoute{}
+
+	swapRouterBz, err := proto.Marshal(&swapRouterExactAmountInRoute)
+	require.NoError(t, err)
+	gammBz, err := proto.Marshal(&gammExactAmountInRoute)
+	require.NoError(t, err)
+
+	require.Equal(t, swapRouterBz, gammBz)
+}
```
mattverse, line 26: What's the reason we do not delete `SwapAmountInRoute` from gamm types completely? Seems that `SwapAmountInRoute` in gamm types have no actual usage and its not getting called anywhere

mattverse (commented):
Left one question / concerning point about gamm types. Other then that, looks LGTM

x/gamm/types/route_test.go
p0mvn: We can delete that in a subsequent PR. It is currently being used as proof that there should be no breakage

proto/osmosis/gamm/v1beta1/query.proto
```diff
@@ -282,7 +283,7 @@ message QuerySwapExactAmountOutRequest {
   // TODO: CHANGE THIS TO RESERVED IN A PATCH RELEASE
   string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
   uint64 pool_id = 2 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
-  repeated SwapAmountOutRoute routes = 3 [
+  repeated osmosis.swaprouter.v1beta1.SwapAmountOutRoute routes = 3 [
     (gogoproto.moretags) = "yaml:\"routes\"",
     (gogoproto.nullable) = false
   ];
```
ValarDragon, line 289: We should ask dan lynch and front-end how breaking this is.

Its not clear to me that its worth breaking this in the gamm proto, vs just doing a conversion in the proto boilerplate handling code.

ValarDragon: (I think it will break their compiled types, unless we can register a way for a type to get aliased or something in codegen)

x/gamm/types/route_test.go
ValarDragon: This doesn't guarantee things, we need non-empty entries for every field for this to actually be checked.

p0mvn: Missed adding values. Added now

mattverse: https://github.com/osmosis-labs/osmosis/issues/3921

proto/osmosis/gamm/v1beta1/query.proto
pyramation: as long as the interfaces/API don't change, it doesn't matter where the types live. Telescope will handle it.

from [what I can see here](https://github.com/osmosis-labs/osmosis/blob/f594d0d6614a46f4cbf5dfb6af71ace3828b41d9/proto/osmosis/swaprouter/v1beta1/swap_route.proto#L8-L18) it looks the the interface to the type doesn't change, so will be totally ok!

pyramation: also, even w/o telescope, from what I can tell (since I believe osmosis-fe uses another tool) — nothing should break. But, we should probably all keep in sync when this update rolls out to update FE libs and repos.

p0mvn: Thanks, @pyramation . I also spoke with Jon to make sure that there is no unexpected breakage for FE.

We will track this change and remind everyone to update before this is out.

This should be good to go

ValarDragon (approved):

ValarDragon:
Needs changelog entry for breaking changes, otherwise LGTM.

p0mvn:
Seems to be JSON serialization compatible based on added test here: [3a46711](https://github.com/osmosis-labs/osmosis/pull/3880/commits/3a4671110656937ccb3cc26cd8d8b5cbe44aac97)

ValarDragon:
oh awesome, thats great!

Issue #3644, opened 2022-12-06
test(x/gamm): InitializePool


p0mvn:
Should we add InitializePool tests to this PR?

_Originally posted by @czarcas7ic in https://github.com/osmosis-labs/osmosis/pull/3638#discussion_r1041362545_

@p0mvn to finish issue description

p0mvn:
Just an update on the issue. This method needs to be tested and spec updated: https://github.com/osmosis-labs/osmosis/blob/a1c13a6ebd92fdb46f2673f1702f5a430264964e/x/gamm/keeper/pool_service.go#L64

The PR should be made against `main` as this function has been merged.

PR #3885, opened 2022-12-30
intializePool test and spec updated

expertdicer:
Closes: #3644

Create a test for function ``initializePool`` in ``x/gamm`` module and update its spec.

## Brief Changelog

  -  Add ``TestInitializePool`` in ``pool_service_test.go``, which cover 2 cases: 

     -  initialize pool with default assets (pass), 
     - initialize pool with empty sender (not pass)

  - If expect test case to pass, we check: 
     - make sure new pool exists and has minted the correct number of pool shares, 
     - ``tokenDenomMetadata`` set correctly  
     - ``AfterPoolCreated`` hook works fine. 

## Testing and Verifying

https://github.com/notional-labs/osmosis/blob/35fe5ab4cb13d47d550a3387e0f39299a56e8019/x/gamm/keeper/pool_service_test.go#L225


x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +222,121 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+			name:        "initialize pool with default assets",
+			msg:         balancer.NewMsgCreateBalancerPool(testAccount, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor),
+			emptySender: false,
+			expectPass:  true,
+		},
+		{
+			name:        "initialize pool with empty sender",
+			msg:         balancer.NewMsgCreateBalancerPool(testAccount, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor),
+			emptySender: true,
+			expectPass:  false,
```
p0mvn, line 251: Can we please also have the following tests:
- stableswap pool - gets initialized
- any non-CFMM pool such as concentrated liquidity does not get initialized. You can find mocks in `tests/mocks` helpful for this

expertdicer: sure. Lemme check it

expertdicer: I don't really sure why you want a CL pool initialized here in a test of ``x/gamm``. I think it is better placed in ``x/concentrated-liquidity``. Do you agree ?

p0mvn: I think we want to test that if you give it a CL pool (or any non-CFMM pool / mock), then it is going to fail. This function should only execute without error if a CFMM pool is given to it. The pool that implements the following interface:
https://github.com/osmosis-labs/osmosis/blob/f2a497dfac7f41878ddbdcb2d1df4acbe8e2e9b6/x/gamm/types/pool.go#L15

expertdicer: Yep. If I use the function gamm.InitializePool for a mock CL pool. It will not fail.

p0mvn: Did you use `MockConcentratedPoolExtension`?

expertdicer: Yes. I did. You can see in the files changes

x/swaprouter/export_test.go
```diff
@@ -15,6 +15,10 @@ func (k Keeper) GetOsmoRoutedMultihopTotalSwapFee(ctx sdk.Context, route types.M
 	return k.getOsmoRoutedMultihopTotalSwapFee(ctx, route)
 }
 
+func (k *Keeper) FundCommunityPool(ctx sdk.Context, fee sdk.Coins, sender sdk.AccAddress) error {
+	return k.communityPoolKeeper.FundCommunityPool(ctx, fee, sender)
+}
```
mattverse, line 20: Is this export needed? (Don't we use the method directly from the Distr Keeper any ways?)

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,169 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		if test.usingMock {
+			ctrl := gomock.NewController(suite.T())
+			mockPool := mocks.NewMockConcentratedPoolExtension(ctrl)
+			mockPool.EXPECT().GetTotalShares().Return(pool.GetTotalShares()).AnyTimes()
+			mockPool.EXPECT().GetId().Return(poolId).AnyTimes()
+			err = swapModule.InitializePool(suite.Ctx, pool, sender)
+		} else {
+			err = swapModule.InitializePool(suite.Ctx, pool, sender)
+		}
+		// attempt to create a pool with the given NewMsgCreateBalancerPool message. After that,
```
mattverse, line 346:
```suggestion
		// attempt to create a pool with the given CreatePool message of the pool. After that,
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,169 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+				fmt.Sprintf("share token should be minted as %s initially", types.InitPoolSharesSupply.String()),
+			)
+
+			// get expected tokens in new pool and corresponding pool shares
+			expectedPoolTokens := sdk.Coins{}
+			for _, asset := range test.msg.InitialLiquidity() {
+				expectedPoolTokens = expectedPoolTokens.Add(asset)
+			}
+			expectedPoolShares := sdk.NewCoin(types.GetPoolShareDenom(pool.GetId()), types.InitPoolSharesSupply)
+
```
mattverse, line 366: should we also check if the expected Pool share and the expected pool token matches the actual tokens and shares of / in the pool?

x/swaprouter/export_test.go
expertdicer: Yes. It is redundant. I might need to  delete it in the later commit.

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,169 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		},
+		{
+			name:        "initialize stableswap pool with default assets",
+			msg:         stableswap.NewMsgCreateStableswapPool(testAccount, defaultStableSwapPoolParams, defaultStableSwapPoolAssets, defaultScalingFactor, ""),
+			emptySender: false,
+			expectPass:  true,
+		},
+		{
+			name:        "CL Pool given",
+			msg:         balancer.NewMsgCreateBalancerPool(testAccount, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor),
```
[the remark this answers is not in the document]
expertdicer: And this test DOES NOT fail. Which means something is wrong, right ?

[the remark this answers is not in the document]
expertdicer: The balancer message is used for the process previous the ``initializePool``, or else it will fail before it reaches the phase we want to test.

x/gamm/keeper/pool_service_test.go
expertdicer: I can do that!

x/gamm/keeper/pool_service_test.go
expertdicer: The correct behavior of the CL pool test case is to panic. I have updated the code

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,181 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+			expectPass:  true,
+		},
+		{
+			name:      "initialize a CL pool which cause panic",
+			msg:       balancer.NewMsgCreateBalancerPool(testAccount, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor),
+			usingMock: true,
+		},
+	}
+
+	for _, test := range tests {
```
p0mvn, line 281: This is to make sure that there are no problems stemming from `test` of the wrong iteration being used. This is a Go-specific problem
```suggestion
	for _, test := range tests {
	    test := test
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,181 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		suite.SetupTest()
```
p0mvn, line 282: Can you please use `suite.Run` similar to: https://github.com/osmosis-labs/osmosis/blob/f611e01c7fb9ab9d5348250cb0b65e08f3ec651b/x/concentrated-liquidity/pool_test.go#L74

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,181 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+		},
+		{
+			name:        "initialize stableswap pool with default assets",
+			msg:         stableswap.NewMsgCreateStableswapPool(testAccount, defaultStableSwapPoolParams, defaultStableSwapPoolAssets, defaultScalingFactor, ""),
+			emptySender: false,
+			expectPass:  true,
+		},
+		{
+			name:      "initialize a CL pool which cause panic",
+			msg:       balancer.NewMsgCreateBalancerPool(testAccount, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor),
```
p0mvn, line 276: using balancer `MsgCreatePool` is misleading. Can you try this instead: https://github.com/osmosis-labs/osmosis/blob/f611e01c7fb9ab9d5348250cb0b65e08f3ec651b/x/concentrated-liquidity/model/msgs.go#L24

If you use it, you might find it easier to move away from using mocks

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,181 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+				expectedPoolBalance := poolBalanceBefore.Add(test.msg.InitialLiquidity()...)
+				suite.Require().Equal(expectedPoolBalance, pool.GetTotalPoolLiquidity(suite.Ctx))
+
+			} else {
+				suite.Require().Error(err, "test: %v", test.name)
+			}
+		}
+		// attempt to create a pool with the given NewMsgCreateBalancerPool message. After that,
+		// this function calls the InitializePool function
+		// poolId, err := swaprouterKeeper.CreatePool(suite.Ctx, test.msg)
```
p0mvn, line 415: What's this?

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,181 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+					gaugeId, err := poolIncentivesKeeper.GetPoolGaugeId(suite.Ctx, poolId, lockableDuration)
+					suite.Require().NoError(err, "test: %v", test.name)
+
+					poolIdFromPoolIncentives, err := poolIncentivesKeeper.GetPoolIdFromGaugeId(suite.Ctx, gaugeId, lockableDuration)
+					suite.Require().NoError(err, "test: %v", test.name)
+					suite.Require().Equal(poolIdFromPoolIncentives, poolId)
+				}
+
+				// make sure total liquidity increase recored
+				expectedPoolBalance := poolBalanceBefore.Add(test.msg.InitialLiquidity()...)
```
p0mvn, line 406: Nice work on thorough validations here

x/gamm/keeper/pool_service_test.go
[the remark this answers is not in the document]
p0mvn: Could you check a similar suggestion here: https://github.com/osmosis-labs/osmosis/pull/3885#discussion_r1060958449

x/gamm/keeper/pool_service_test.go
expertdicer: If I use ``NewMsgCreateConcentratedPool``, I will still have to do a little trick because ``swaprouterKeeper.GetPoolModule`` only return if the pool id is registered which only happens in constructor process

expertdicer: Or instead of using ``NewMsgCreateConcentratedPool``, I can use a mock.

expertdicer: Instead of ``NewMsgCreateConcentratedPool``, I can use a mock. I think it is more appropriate to use a mock. Can you check it again @p0mvn ?

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,187 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+				for _, asset := range test.msg.InitialLiquidity() {
+					poolBalanceBefore = poolBalanceBefore.Add(sdk.NewCoin(asset.Denom, gammKeeper.GetDenomLiquidity(suite.Ctx, asset.Denom)))
+				}
+
+				// create pool process which happens previous the initializePool phase
+				initialPoolLiquidity := test.msg.InitialLiquidity()
+
+				numAssets := initialPoolLiquidity.Len()
+				suite.Require().Equal(numAssets < swaproutertypes.MinPoolAssets, false)
+				suite.Require().Equal(numAssets > swaproutertypes.MaxPoolAssets, false)
```
p0mvn, line 328:
```suggestion
				suite.Require().False(numAssets < swaproutertypes.MinPoolAssets, false)
				suite.Require().False(numAssets > swaproutertypes.MaxPoolAssets, false)
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +242,187 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+				swaprouterKeeper.SetPoolRoute(suite.Ctx, poolId, test.msg.GetPoolType())
+				err = osmoutils.CreateModuleAccount(suite.Ctx, suite.App.AccountKeeper, pool.GetAddress())
+				suite.Require().NoError(err, "test: %v", test.name)
+
+				poolAddr := pool.GetAddress()
+				err = bankKeeper.SendCoins(suite.Ctx, sender, poolAddr, initialPoolLiquidity)
+				suite.Require().NoError(err, "test: %v", test.name)
+
+				swapModule, err := swaprouterKeeper.GetPoolModule(suite.Ctx, poolId)
+				suite.Require().NoError(err, "test: %v", test.name)
```
p0mvn, line 349: I don't see why all this setup logic is needed.

We should be able to simply do the following:
- define `swaprouter.PoolI` as a filed on the test case
- for each test case., initialize each pool via their respective constructor such as 
    * balancer: https://github.com/osmosis-labs/osmosis/blob/6436e5200ac38a3f39160f12b5c9fda5e57df7bd/x/gamm/pool-models/balancer/pool.go#L43
    * stableswap: https://github.com/osmosis-labs/osmosis/blob/6436e5200ac38a3f39160f12b5c9fda5e57df7bd/x/gamm/pool-models/stableswap/pool.go#L25
    * CL: https://github.com/osmosis-labs/osmosis/blob/6436e5200ac38a3f39160f12b5c9fda5e57df7bd/x/concentrated-liquidity/model/pool.go#L20

Then, all this swap router-related boilerplate becomes obsolete and we can simply provide a pool struct to `InitializePool` method below.

Please see this similar test for inspiration: https://github.com/osmosis-labs/osmosis/blob/6436e5200ac38a3f39160f12b5c9fda5e57df7bd/x/concentrated-liquidity/pool_test.go#L78

expertdicer: I have remove all the setup logic and use the constructors to initialize pool. Hope this time it looks good to you. Can you check it again ? @p0mvn

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+	poolCreationFeeDecCoins := sdk.DecCoins{}
+	for _, coin := range params.PoolCreationFee {
+		poolCreationFeeDecCoins = poolCreationFeeDecCoins.Add(sdk.NewDecCoin(coin.Denom, coin.Amount))
+	}
+
+	_, err := balancer.NewBalancerPool(1, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor, time.Now())
+	require.NoError(suite.T(), err)
+
+	_, err = stableswap.NewStableswapPool(1, defaultStableSwapPoolParams, defaultStableSwapPoolAssets, defaultScalingFactor, "", defaultFutureGovernor)
+	require.NoError(suite.T(), err)
```
p0mvn, line 258: I don't think we need any of these since we create a pool for each test case

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+			osmoassert.ConditionalPanic(suite.T(), test.expectPanic, func() {
+				suite.SetupTest()
+
+				gammKeeper := suite.App.GAMMKeeper
+				bankKeeper := suite.App.BankKeeper
+				poolIncentivesKeeper := suite.App.PoolIncentivesKeeper
+
+				// fund sender test account
+				var sender sdk.AccAddress
+				sender = testAccount
```
p0mvn, line 325: nit:
```suggestion
				sender := testAccount
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+
+	_, err := balancer.NewBalancerPool(1, defaultPoolParams, defaultPoolAssets, defaultFutureGovernor, time.Now())
+	require.NoError(suite.T(), err)
+
+	_, err = stableswap.NewStableswapPool(1, defaultStableSwapPoolParams, defaultStableSwapPoolAssets, defaultScalingFactor, "", defaultFutureGovernor)
+	require.NoError(suite.T(), err)
+
+	tests := []struct {
+		name        string
+		pool        func() swaproutertypes.PoolI
```
p0mvn, line 262:
```suggestion
		createPool        func() swaproutertypes.PoolI
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+
+					// check to make sure new pool exists and has minted the correct number of pool shares
+					pool, err := gammKeeper.GetPoolAndPoke(suite.Ctx, defaultPoolId)
+					suite.Require().NoError(err, "test: %v", test.name)
+					suite.Require().Equal(types.InitPoolSharesSupply.String(), pool.GetTotalShares().String(),
+						fmt.Sprintf("share token should be minted as %s initially", types.InitPoolSharesSupply.String()),
+					)
+
+					// get expected tokens in new pool and corresponding pool shares
+					expectedPoolTokens := sdk.Coins{}
```
p0mvn, line 341:
```suggestion
					expectedPoolTokens := sdk.NewCoins()
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+				err = gammKeeper.InitializePool(suite.Ctx, test.pool(), sender)
+
+				if test.expectPass {
+					suite.Require().NoError(err, "test: %v", test.name)
+
+					// check to make sure new pool exists and has minted the correct number of pool shares
+					pool, err := gammKeeper.GetPoolAndPoke(suite.Ctx, defaultPoolId)
+					suite.Require().NoError(err, "test: %v", test.name)
+					suite.Require().Equal(types.InitPoolSharesSupply.String(), pool.GetTotalShares().String(),
+						fmt.Sprintf("share token should be minted as %s initially", types.InitPoolSharesSupply.String()),
```
p0mvn, line 337: nit: This auto-expands when you use %s format
```suggestion
						fmt.Sprintf("share token should be minted as %s initially", types.InitPoolSharesSupply),
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+					)
+
+					// get expected tokens in new pool and corresponding pool shares
+					expectedPoolTokens := sdk.Coins{}
+					for _, asset := range test.pool().GetTotalPoolLiquidity(suite.Ctx) {
```
p0mvn, line 342: When you call `test.pool()` again, it creates a completely new pool with a different id

I think instead, we should use the pool that was returned from `GetPoolAndPoke` above to make sure it was updated

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
+					// check AfterPoolCreated hook
+					for _, lockableDuration := range poolIncentivesKeeper.GetLockableDurations(suite.Ctx) {
+						gaugeId, err := poolIncentivesKeeper.GetPoolGaugeId(suite.Ctx, defaultPoolId, lockableDuration)
+						suite.Require().NoError(err, "test: %v", test.name)
+
+						poolIdFromPoolIncentives, err := poolIncentivesKeeper.GetPoolIdFromGaugeId(suite.Ctx, gaugeId, lockableDuration)
+						suite.Require().NoError(err, "test: %v", test.name)
+						suite.Require().Equal(poolIdFromPoolIncentives, defaultPoolId)
+					}
+
```
p0mvn, line 376: Please also add a check that `sender` had shares minted to them by checking their address's bank balance

p0mvn (commented):
Nice work. Very close to being ready for merge. Had a couple of suggestions. Please take a look

x/gamm/keeper/pool_service_test.go
expertdicer: I agree.

x/gamm/keeper/pool_service_test.go
expertdicer: I agree

x/gamm/keeper/pool_service_test.go
p0mvn: Sweet, please remove this then, and then we can merge the PR! Everything else LGMT

x/gamm/keeper/pool_service.go
```diff
@@ -62,6 +62,9 @@ func (k Keeper) CalculateSpotPrice(
 // - Sets bank metadata for the LP denom
 // - Records total liquidity increase
 // - Calls the AfterPoolCreated hook
+//
+// After the gammKeeper set pool to the state, this function calls the AfterPoolCreated hook
+// which is responsible for createing a gauge for each pool’s lockable duration
```
ValarDragon, line 67: Isn't this documented right above?

The fact that it creates gauges for each pool's lockable duration isn't the problem of the gamm module, thats the advantage of hooks! 

```suggestion
// - Calls the AfterPoolCreated hook (see app module wiring for what may be running on this)
```

p0mvn:
Hi @expertdicer . Any updates on this comment: https://github.com/osmosis-labs/osmosis/pull/3885#discussion_r1064257045 ?

x/gamm/keeper/pool_service_test.go
expertdicer: I have removed the redundant logic. Tks Roman

expertdicer:
> Hi @expertdicer . Any updates on this comment: [#3885 (comment)](https://github.com/osmosis-labs/osmosis/pull/3885#discussion_r1064257045) ?


Yes. I have updated the code. Pls check

x/gamm/keeper/pool_service.go
```diff
@@ -61,7 +61,7 @@ func (k Keeper) CalculateSpotPrice(
 // - Mints LP shares to the pool creator
 // - Sets bank metadata for the LP denom
 // - Records total liquidity increase
-// - Calls the AfterPoolCreated hook
+// - Calls the AfterPoolCreated hook (see app module wiring for what may be running on this)
```
p0mvn, line 64:
```suggestion
// - Calls the AfterPoolCreated hook
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -222,6 +241,147 @@ func (suite *KeeperTestSuite) TestCreateBalancerPool() {
...
 
+func (suite *KeeperTestSuite) TestInitializePool() {
+	params := suite.App.GAMMKeeper.GetParams(suite.Ctx)
+	testAccount := suite.TestAccs[0]
+
+	// get raw pool creation fee(s) as DecCoins
+	poolCreationFeeDecCoins := sdk.DecCoins{}
+	for _, coin := range params.PoolCreationFee {
+		poolCreationFeeDecCoins = poolCreationFeeDecCoins.Add(sdk.NewDecCoin(coin.Denom, coin.Amount))
+	}
```
p0mvn, line 252:
```suggestion
```

p0mvn (approved):
LGTM nice work!

p0mvn:
Hi @expertdicer . Can you please address the latest comments and resolve conflicts so that we can merge this?

expertdicer:
> Hi @expertdicer . Can you please address the latest comments and resolve conflicts so that we can merge this?
Yes. I have resolved all the conflicts

x/gamm/keeper/pool_service_test.go
```diff
@@ -30,13 +41,21 @@ var (
...
+	defaultAcctFunds sdk.Coins = sdk.NewCoins(
 		sdk.NewCoin("uosmo", sdk.NewInt(10000000000)),
 		sdk.NewCoin("foo", sdk.NewInt(10000000)),
 		sdk.NewCoin("bar", sdk.NewInt(10000000)),
 		sdk.NewCoin("baz", sdk.NewInt(10000000)),
 	)
+	ETH                       = "eth"
+	USDC                      = "usdc"
+	defaultTickSpacing        = uint64(1)
+	DefaultExponentAtPriceOne = sdk.NewInt(-4)
```
p0mvn, line 58: Why do we need this? Seems like a concentrated liquidity-specific context?

p0mvn: Nvm, I see why

p0mvn (approved):

PR #3888, opened 2022-12-30
fix: add negative to cli

czarcas7ic:
Closes: #XXX

Osmocli does not allow negatives. This is a workaround. This needs to get merged since we utilize this as its own gomod package, so local changes do not overwrite like before.


p0mvn (approved):
:sandwich:

PR #3893, opened 2022-12-30
feat(CL): concentrated liquidity fees POC CU-86776j1g7

p0mvn:
Closes: #XXX

## What is the purpose of the change

This PR implements concentrated liquidity fees by utilizing our newly built `accum` module for claiming rewards.

There are a few breaking changes to the `accum` package made. For example, we need the ability to create, add to, and remove from a position with a custom accumulator update. In the context of concentrated liquidity, this custom accumulator update signifies the computation of fee growth outside the position. Without this change, we would be distributing rewards linearly across all positions.

By providing a custom position accumulator update, we are able to propagate the fee growth within a range to it.

This PR will be split into multiple smaller PRs. Tests are to be added.

## Additional Notes

I moved away from the idea of defining additional "Trackers" for ticks inside the package. Instead, tracked the per-tick accumulators directly in [tick structs](https://github.com/osmosis-labs/osmosis/blob/f6d87d1a6acbe0a7095c5269cc1d75367e122390/proto/osmosis/concentrated-liquidity/tickInfo.proto#L25-L28).

I think I was able to mitigate the problem that we were discussing earlier. The one about accum module distributing rewards linearly with the number of shares without accounting for whether a position range was ever activated.

It is mitigated by introducing an [ability to provide custom updates](https://github.com/osmosis-labs/osmosis/blob/fdc432058d9620b2db47c36691a913e1c7f1088a/osmoutils/accum/accum.go#L96-L98) to the position's accumulators. This way, we should be able to [compute fee growth outside the range](https://github.com/osmosis-labs/osmosis/blob/fdc432058d9620b2db47c36691a913e1c7f1088a/x/concentrated-liquidity/fees.go#L64) and, then, i[nject it into the ](https://github.com/osmosis-labs/osmosis/blob/fdc432058d9620b2db47c36691a913e1c7f1088a/x/concentrated-liquidity/fees.go#L74)accum[ module's position](https://github.com/osmosis-labs/osmosis/blob/fdc432058d9620b2db47c36691a913e1c7f1088a/x/concentrated-liquidity/fees.go#L74).

This custom "fee growth outside" position accumulator can then be used directly by the accum module's [ClaimRewards function](https://github.com/osmosis-labs/osmosis/blob/f6d87d1a6acbe0a7095c5269cc1d75367e122390/osmoutils/accum/accum.go#L221) since it would be subtracted from pool's "fee growth global" to get position's "fee growth inside" and multiply it by the number of shares inside that position to get final result.

PR #3895, opened 2022-12-31
refactor(osmoutils): UpdateAccumulator does not overwrite; tests

p0mvn:
Closes: #XXX

This PR is extracted from https://github.com/osmosis-labs/osmosis/pull/3893

We should be updating the accumulator instead of overwriting its value. The accumulator is likely to always grow. As a result, this is a UX feature.

In the unlikely case when it is needed to reset the accumulator value, the user can still do so by providing negative input.

Additionally, added a `GetValue()` API on the accumulator.

Tests added.

### DIFF
--- a/osmoutils/accum/accum.go
+++ b/osmoutils/accum/accum.go
@@ -63,10 +63,12 @@ func setAccumulator(accum AccumulatorObject, amt sdk.DecCoins) {
 	osmoutils.MustSet(accum.store, formatAccumPrefixKey(accum.name), &newAccum)
 }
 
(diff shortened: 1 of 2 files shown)


AlpinYukseloglu (approved):
LGTM!

czarcas7ic (approved):

p0mvn:
> LGTM, although I'm still a little confused on where/when this function is intended to be used (DM'd offline)

Replied but answering here as well - it's to be called when we charge a fee on the token in during a CL swap

PR #3896, opened 2022-12-31
refactor(osmoutils): index position by string name instead of address

p0mvn:
Closes: #XXX

Index position by string name instead of the address. This is required for some use cases such as concentrated liquidity (CL).

In CL, we need to index positions by the owner, pool id, and ticks.

The flexibility to index by address is still provided since address can easily be converted to string.

## Testing and Verifying

### DIFF
--- a/osmoutils/accum/accum_helpers.go
+++ b/osmoutils/accum/accum_helpers.go
@@ -7,25 +7,25 @@ import (
 )
 
 // Creates a new position at accumulator's current value with a specific number of shares and unclaimed rewards
-func createNewPosition(name AccumulatorObject, index string, numShareUnits sdk.Dec, unclaimedRewards sdk.DecCoins, options *Options) {
+func createNewPosition(accum AccumulatorObject, index string, numShareUnits sdk.Dec, unclaimedRewards sdk.DecCoins, options *Options) {
 	position := Record{
 		NumShares:        numShareUnits,
-		InitAccumValue:   name.value,
+		InitAccumValue:   accum.value,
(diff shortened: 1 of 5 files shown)


osmoutils/accum/accum_helpers.go
```diff
@@ -7,20 +7,20 @@ import (
 )
 
 // Creates a new position at accumulator's current value with a specific number of shares and unclaimed rewards
-func createNewPosition(accum AccumulatorObject, addr sdk.AccAddress, numShareUnits sdk.Dec, unclaimedRewards sdk.DecCoins, options *Options) {
+func createNewPosition(name AccumulatorObject, index string, numShareUnits sdk.Dec, unclaimedRewards sdk.DecCoins, options *Options) {
```
AlpinYukseloglu, line 10: Trying to reconcile this renaming with the other functions and tests – here you call the `AccumulatorObject` `name`, whereas in `AddToPosition` and in tests you call the address string `name`. We should probably align these before merging

AlpinYukseloglu (approved):
Left one comment about naming inconsistency but other than that LGTM

osmoutils/accum/accum_helpers.go
czarcas7ic: Second this

czarcas7ic (approved):

osmoutils/accum/accum_helpers.go
p0mvn: Erroneous change. Fixed this and other name issues found while re-reviewing

PR #3897, opened 2022-12-31
refactor(osmoutils): accumulator ClaimRewards returns sdk.Coins

p0mvn:
Closes: #XXX

When we `ClaimRewards` from the accumulator store, we are likely to utilize them for sdk operations such as "bank send". All such operations utilize `sdk.Coins` instead of `sdk.DecCoins`.

Therefore, in terms of UX, it is beneficial to return `sdk.Coins` right away.

It is also helpful for reinvesting the truncation change into the newly created position after claiming the rewards so that this change isn't lost.

## Testing and Verifying

### DIFF
--- a/osmoutils/accum/accum.go
+++ b/osmoutils/accum/accum.go
@@ -178,12 +178,13 @@ func (accum AccumulatorObject) ClaimRewards(addr sdk.AccAddress) (sdk.Coins, err
 	totalRewards := getTotalRewards(accum, position)
 
 	// Return the integer coins to the user
-	// The remaining change is reinvested into the new position.
-	truncatedRewards, remainingChange := totalRewards.TruncateDecimal()
+	// The remaining change is thrown away.
+	// This is acceptable because we round in favour of the protocol.
+	truncatedRewards, _ := totalRewards.TruncateDecimal()
 
 	// Create a completely new position, with no rewards
 	// TODO: remove the position from state entirely if numShares = zero
-	createNewPosition(accum, addr, position.NumShares, remainingChange, position.Options)
+	createNewPosition(accum, addr, position.NumShares, sdk.NewDecCoins(), position.Options)
 
 	return truncatedRewards, nil
 }
--- a/osmoutils/accum/accum.go
+++ b/osmoutils/accum/accum.go
@@ -169,17 +169,21 @@ func (accum AccumulatorObject) GetPositionSize(addr sdk.AccAddress) (sdk.Dec, er
 // unclaimed rewards. The position's accumulator is also set to the current accumulator value.
 // Returns error if no position exists for the given address. Returns error if any
 // database errors occur.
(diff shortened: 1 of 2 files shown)


osmoutils/accum/accum_test.go
```diff
@@ -212,6 +212,18 @@ func (suite *AccumTestSuite) TestClaimRewards() {
 			sdk.NewDecCoinFromDec(denomTwo, sdk.OneDec())).MulDec(sdk.NewDec(3))
 	)
 
+	// single output convinience wrapper.
```
AlpinYukseloglu, line 215:
```suggestion
	// single output convenience wrapper.
```

osmoutils/accum/accum.go
```diff
@@ -169,17 +169,21 @@ func (accum AccumulatorObject) GetPositionSize(addr sdk.AccAddress) (sdk.Dec, er
...
 	totalRewards := getTotalRewards(accum, position)
 
+	// Return the integer coins to the user
+	// The remaining change is reinvested into the new position.
+	truncatedRewards, remainingChange := totalRewards.TruncateDecimal()
+
 	// Create a completely new position, with no rewards
 	// TODO: remove the position from state entirely if numShares = zero
-	createNewPosition(accum, addr, position.NumShares, sdk.NewDecCoins(), position.Options)
+	createNewPosition(accum, addr, position.NumShares, remainingChange, position.Options)
```
AlpinYukseloglu, line 186: Are you confident leaving `remainingChange` in the position is necessary here? My understanding is that this can at most ever  be a rounding error and feels like it would be prudent to just shave it off for security reasons (the most harm that can be done is 1 token i.e. `1 uosmo`, which is error we accept in many places in GAMM)

AlpinYukseloglu (approved):
Had one question but other than that LGTM

osmoutils/accum/accum.go
czarcas7ic: Im leaning towards Alpin's point on this one. I especially think of a world in which there is a bunch of accounts that claim rewards that never intend on creating a new position, so we just have state bloat that is worth nothing to the user.

osmoutils/accum/accum_test.go
```diff
@@ -212,6 +212,18 @@ func (suite *AccumTestSuite) TestClaimRewards() {
...
+	toCoins := func(decCoins sdk.DecCoins) sdk.Coins {
+		coins, _ := decCoins.TruncateDecimal()
+		return coins
+	}
+
+	// single output convinience wrapper.
```
czarcas7ic, line 221:
```suggestion
	// single output convenience wrapper.
```

osmoutils/accum/accum.go
p0mvn: I was questioning this as well. Going to change to throw away this DecCoins change

p0mvn: addressed

PR #3894, opened 2022-12-31
feat: concentrated liquidity incentives

czarcas7ic:
Closes: #XXX

## What is the purpose of the change

> Add a description of the overall background and high level changes that this PR introduces

*(E.g.: This pull request improves documation of area A by adding ....*

## Brief Changelog

*(for example:)*
 
  - *The metadata is stored in the blob store on job creation time as a persistent artifact*
  - *Deployments RPC transmits only the blob storage reference*
  - *Daemons retrieve the RPC data from the blob cache*

## Testing and Verifying

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*
  - *Manually verified the change by ...*

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #3898, opened 2022-12-31
[CL]: Join-time tracker sumtree

AlpinYukseloglu:
Pulled from #3886

This PR implements a join-time tracker sumtree that dynamically updates as positions are changed in a given pool. This allows us to get the amount of liquidity that joined a pool before, at, or after a given join time, allowing us to incentivize uptime.

Once we finalize our process for reward calculation post-#3894 & decide on whether we want to support multiple uptime tiers per incentive token, we can add a second sum to our sumtree to track accumulated rewards.

## Brief Changelog

- Change sumtree to accumulate sdk.Dec instead of sdk.Int
- Populate a sumtree tracking position join times for each pool every time a position is updated
- Thoroughly fuzz test new sumtree logic

## Testing and Verifying

All of the new logic is heavily tested in `position_test.go` (both by expanding existing tests & adding new robust fuzz tests)


x/concentrated-liquidity/incentives.go
```diff
@@ -0,0 +1,83 @@
...
+func (k Keeper) getSumtree(ctx sdk.Context, poolID uint64) sumtree.Tree {
+	return sumtree.NewTree(prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyAccumulationStore(poolID)), 10)
+}
+
+// Adds to a position's liquidity in Liquidity tree, setting its JoinTime to current block time.
+//
+// Since we want to reset a position's JoinTime every time they update their liquidity, any
+// liquidity update is equivalent to removing the old JoinTime node from the liquidity tree
+// and replacing it with a new one. Thus, this function simply takes in the old and new liquidity
+// amounts and leaves liquidity delta calculations to the caller.
```
czarcas7ic, line 26: Side note: this will require the user to claim any outstanding rewards prior to replacing the node from the liquidity tree right?

x/concentrated-liquidity/position_test.go
```diff
@@ -163,6 +189,11 @@ func (s *KeeperTestSuite) TestGetPosition() {
 			// Init suite for each test.
 			s.Setup()
 
+			// Update expectedPos time to current context's blocktime
+			if test.expectedPosition != nil {
+				test.expectedPosition.JoinTime = s.Ctx.BlockTime()
+			}
+
```
czarcas7ic, line 196: This got added but it appears nothing tests this logic

x/concentrated-liquidity/position_test.go
```diff
@@ -179,7 +210,113 @@ func (s *KeeperTestSuite) TestGetPosition() {
...
+	// Max lower and upper ticks
+	tickLowerBound := int64(100)
+	tickUpperBound := int64(100)
+
+	// Max time elapsed between positions (in seconds)
+	timeElapsedUpperBound := 1000
+
+	// We generate a new test account per position for robustness
+	testAccounts := osmoutils.CreateRandomAccounts(numPositions)
```
czarcas7ic, line 238: I feel like, for robustness, we should have accounts with multiple positions (this feels like the more complex logic imo)

x/concentrated-liquidity/position_test.go
```diff
@@ -179,7 +210,113 @@ func (s *KeeperTestSuite) TestGetPosition() {
...
+		// Half the time, override time elapsed and add to the same join time as the previous position (time elapsed = 0)
+		timeElapsed = timeElapsed * time.Duration(rand.Int()%2)
+
+		// Update current block time to reflect time elapsed
+		joinTime := s.Ctx.BlockTime().Add(timeElapsed)
+		s.Ctx = s.Ctx.WithBlockTime(joinTime)
+
+		// Generate random lower and upper ticks
+		lowerTick := (-1) * (rand.Int63() % tickLowerBound)
+		upperTick := rand.Int63() % tickUpperBound
```
czarcas7ic, line 269: I like what you did with this fuzz testing, but I think we might need to expand this to prevent every lower tick from being negative and every upper tick from being positive. Seems like we might be leaving some edge cases untested.

x/concentrated-liquidity/incentives.go
```diff
@@ -0,0 +1,83 @@
+package concentrated_liquidity
```
czarcas7ic, line 1: Are you planning to add tests for these?

czarcas7ic (commented):
Nice work, have a few questions and then will happily review a second time (didn't feel I should approve on first look since this is my first time reviewing sumtree stuff)

x/concentrated-liquidity/incentives.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Add new position time to JoinTime sumtree
+	k.getSumtree(ctx, poolId).Increase(types.KeyJoinTime(position.JoinTime), newLiquidity)
+
+	return nil
+}
+
+// Remove liquidity from `position`s join time without resetting position.JoinTime.
+// This is to allow for LPs to remove a portion of their liquidity without forfeiting their uptime
+// on their remaining liquidity.
```
p0mvn, line 48:
```suggestion
// on their remaining liquidity.
// The given liquidity must be positive. Fails with an error otherwise.
```

x/concentrated-liquidity/position.go
```diff
@@ -57,6 +57,18 @@ func (k Keeper) initOrUpdatePosition(
 	// TODO: consider deleting position if liquidity becomes zero
 
 	k.setPosition(ctx, poolId, owner, lowerTick, upperTick, position)
+
+	// Update pool's liquidity tree to reflect new position
+	if liquidityDelta.GT(sdk.ZeroDec()) {
```
p0mvn, line 62: nit: we can store `sdk.ZeroDec` as a global variable to avoid underlying reallocations every time

p0mvn: Is there an `IsPositive` helper? It might be helpful here

p0mvn (commented):
Overall logic LGTM, nice work!

I'm wondering whether migration is justified. Is there a workaround to avoid changing Int -> Dec in the accumulator proto?

I still need to review position test updates. Also, would love to see tests added for logic in `incentives.go`

AlpinYukseloglu:
Thanks for the comments all – I just realized that this sumtree logic needs another layer to track tick range for joins for it to be useful for incentives, so converting this to a draft while I flesh out what needs to be done on that front

PR #3899, opened 2023-01-01
Making osmoutils compile on chains that don't use our SDK fork

nicolaslara:

Osmoutils wouldn't compile on chains that didn't use osmosis' sdk fork because ClawbackVestingAccount is not available.

This ensures that that struct is not needed at compile time.

Doing this via a global variable might be a bit controversial, but I think it's the cleanest abstraction to avoid having to maintain several other packages and/or have implementors override this with replace/fork.

We could use this pattern in other reusable packages and it might even be usable for supporting different versions of the sdk or ibc. 

## Brief Changelog

*(for example:)*
 
  - ClawbackVestingAccount is no longer part of osmoutils and is now injected in app.go

## Testing and Verifying
all tests pass

  - Does this pull request introduce a new feature or user-facing behavior changes? (/ no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (/ no)
  - How is the feature or change documented? (not applicable )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [#3693](https://github.com/osmosis-labs/osmosis/pull/3693) Add `EstimateSwapExactAmountOut` query to stargate whitelist
 * [#3731](https://github.com/osmosis-labs/osmosis/pull/3731) BigDec Power functions with decimal exponent.
 * [#3847](https://github.com/osmosis-labs/osmosis/pull/3847) GeometricTwap and GeometricTwapToNow queries added to Stargate whitelist.
-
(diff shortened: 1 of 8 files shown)


ValarDragon:
Hrmm, I really don't like this being done from the osmosis app. (Otherwise the reflection is good w/ me)

I mainly just don't want complexity pushed to app.go.

Happy to merge for now though! Theres more high leverage places to do next, for unwinding the globals complexity

ValarDragon:
Another option we may want to evaluate in future, is compile time build flags

ValarDragon (approved):
Can you add a changelog entry?

nicolaslara:
Added the changelog entry. 

I experimented with reflection first here, but you can't conditionally import something just based on the name, and you'd need the struct to exist to get it to compile.

Compile time build flags would work if it works like rust, but I think we'd run into the same issue as with reflection

PR #3908, opened 2023-01-03
Use standard IBC errors

nicolaslara:
Refs: https://github.com/osmosis-labs/osmosis/pull/3838#discussion_r1056616909

Wasm hooks now uses the standard ibc v4 errors but emits the errors for the user to review.

## Brief Changelog

*(for example:)*
 
  - Added an utility function to osmoutils for emiting errors before returning an ibc ack error
  - Updated wasmhooks errors to use that function

## Testing and Verifying

Updated tests to check the errors properly (TODO)


x/ibc-hooks/types/errors.go
```diff
@@ -1,8 +1,14 @@
...
 
+import sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
 var (
-	ErrBadPacketMetadataMsg = "cannot unmarshal metadata: '%v'. %s"
 	ErrBadMetadataFormatMsg = "wasm metadata not properly formatted for: '%v'. %s"
 	ErrBadExecutionMsg      = "cannot execute contract: %v"
-	ErrBadResponse          = "cannot create response: %v"
+
+	ErrMsgValidation = sdkerrors.Register("wasm-hooks", 1, "error in wasmhook message validation")
```
mattverse, line 9: Just curious why use "wasm-hooks" instead of ModuleName? Also is it possible in this case to avoid using sdkerrors? AFAIK our goal in general is to refrain from using sdk errors

nicolaslara: > Just curious why use "wasm-hooks" instead of ModuleName? 

This error is specific to the validation of the wasm hook. The module could (and probably will in the future) have different hooks. That said, don't mind changing it.

> is it possible in this case to avoid using sdkerrors? AFAIK our goal in general is to refrain from using sdk errors

What's the alternative? Ibc internally expects an error with codespace and code id. (https://github.com/cosmos/ibc-go/blob/b40dbc646c7191aaee12d3a70243f9b6f3f80919/modules/core/04-channel/types/acknowledgement.go#L32-L42)

osmoutils/ibc.go
```diff
@@ -2,24 +2,29 @@ package osmoutils
...
-func NewStringErrorAcknowledgement(err string) channeltypes.Acknowledgement {
-	// ToDo: Do we want to do this or do we want to use the IBC errors and emit the string?
-
-	return channeltypes.Acknowledgement{
-		Response: &channeltypes.Acknowledgement_Error{
-			Error: err,
-		},
+// NewEmitErrorAcknowledgement creates a new error acknowledgement after having emitted an event with the
+// details of the error.
+func NewEmitErrorAcknowledgement(ctx sdk.Context, err error, errorContexts ...string) channeltypes.Acknowledgement {
```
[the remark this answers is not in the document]
nicolaslara: I don't think we will use more than one, but AFAIK that's the only way to pass 0 or 1 arguments. 

I like the abstraction of calling it without the string if it isn't needed.

Happy to make that a required arg and call it as `NewEmitErrorAcknowledgement(ctx, err, "")` when there isn't any description/context

x/ibc-hooks/types/errors.go
mattverse: > This error is specific to the validation of the wasm hook. The module could (and probably will in the future) have different hooks. That said, don't mind changing it.
Ah I see what you were trying to do. LGTM!

> What's the alternative? Ibc internally expects an error with codespace and code id. (https://github.com/cosmos/ibc-go/blob/b40dbc646c7191aaee12d3a70243f9b6f3f80919/modules/core/04-channel/types/acknowledgement.go#L32-L42)

Don't know much about ibc-go but is this the method that receives the error? https://github.com/cosmos/ibc-go/blob/b40dbc646c7191aaee12d3a70243f9b6f3f80919/modules/core/04-channel/types/acknowledgement.go#L32-L35 looks like it takes in go lang error. Or I assume there might be different components in ibc-go that accepts codespace and code-id. If so, please feel free to resolve this comment!

osmoutils/ibc.go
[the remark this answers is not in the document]
mattverse: Oh we have cases where we have to take in 0 arguments, and that's the reason why a variadic function was used here. Gotcha. Thanks for the explanation!

x/ibc-hooks/types/errors.go
nicolaslara: yeah, we can pass a golang errors (the sdk errors are golang errors under the hood). I guess I could define the error type using `sdkerrors.New` or manually as `Error{codespace: codespace, code: code, desc: desc}`, but not sure what the advantage of doing that would be.

Maybe we should have a standard way to do this across osmosis?

nicolaslara: @ValarDragon can you chime in here? I think we should leave these using sdkerrors. AFAICT, passing a normal error to IBC's `channeltypes.NewErrorAcknowledgement` will return the default "internalABCICode" (1). Having the code numbers is useful to determine what happened.

ValarDragon: Agreed, SDK error for this makes sense

(It doesn't for most other modules tho)

ValarDragon: Hrmm, actually no, I'm onboard with just doing normal golang errors. I don't think people are actually going to parse off these codespaces, vs just looking at events / explorers to diagnose error.

ValarDragon: We're generally trying to rethink codespacing in cosmos: https://github.com/cosmos/cosmos-sdk/issues/14070

Very few people actually use codespacing in their integrator logic. 

I'm not very opinionated on this, since its unlikely these errors are things we'd want to change in a patch release.

ValarDragon: I believe we have to start registration from 2 onwards though

osmoutils/ibc.go
ValarDragon, line 13: Why don't we just move this to ibc-hooks, then we can delete the entire ibc-hooks <> osmoutils dependency

x/ibc-hooks/types/errors.go
ValarDragon: Ah given that this is IBC-standard here, agreed with SDK registering these

osmoutils/ibc.go
```diff
@@ -2,24 +2,29 @@ package osmoutils
...
 	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			"ibc-acknowledgement-error",
+			attributes...,
+		),
+	})
+
+	return channeltypes.NewErrorAcknowledgement(err)
```
ValarDragon, line 27: cref https://github.com/cosmos/ibc-go/blob/ecb845d5e43f53decf48f8ed88c7847a9a4375cb/modules/core/04-channel/types/acknowledgement.go#L32

This makes sense to me!

ValarDragon (approved):
LGTM, we just need to increment codespace by one

osmoutils/ibc.go
nicolaslara: sounds good!

x/ibc-hooks/types/errors.go
nicolaslara: ok, so, in conclusion: keep sdk errors for this so it matches other IBC errors and move the first code to 2 (which was just an oversight on my part), right?

osmoutils/ibc.go
nicolaslara: ahh, no. Just realized it's in osmoutils because rate-limiting is also using it (we could have rate limiting import it from wasm hooks, but it seems dirtier)

x/ibc-hooks/types/errors.go
ValarDragon: yup!

osmoutils/ibc.go
ValarDragon: Didn't we want rate limiting to import ibc-hooks anyway? (This isn't release blocking at all either way though)

nicolaslara: yeah, that makes sense. We can change it later when rate limits is based off ibc hooks.

Issue #3889, opened 2022-12-30
bug(CL): make calcOutAmtGivenIn and calcInAmtGivenOut non-mutative


p0mvn:
## Background

Currently, `calcOutAmtGivenIn` and `calcInAmtGivenOut` are designed to be mutative.

The mutations stem from the fact that we must wrirt ticks to the store as we cross them.

## Suggested Design

Change `ctx` to `cacheCtx` inside both  `calcOutAmtGivenIn` and `calcInAmtGivenOut`  to make sure that these methods do not mutate state

From `calcOutAmtGivenIn` and `calcInAmtGivenOut` , return a `writeCtx` callback that makes the function mutative if called.

Reference `createPosition`'s example:
https://github.com/osmosis-labs/osmosis/blob/f2a497dfac7f41878ddbdcb2d1df4acbe8e2e9b6/x/concentrated-liquidity/lp.go#L46

## Acceptance Criteria
- refactor with the suggested design
- add tests to make sure that when `writeCtx` return is not called, there are no mutations. If it is called, the state is mutated

Issue #3890, opened 2022-12-30
bug(CL): write tick to state during swaps


p0mvn:
## Background

Currently, the ticks are not written back to state when they are updated during swaps.

## Suggested Design

We need to make sure that ticks are updated.

## Acceptance Criteria
- add tests upon fixing the problem. For example, you cannot drain the same liquidity multiple times

mattverse:
Do ticks not get updated via https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/swaps.go#L71?

p0mvn:
I don't think so. Ticks would have to be updates as we cross them since they have liquidity drained.

This would have to happen at a much lower level of abstraction:
- either here: https://github.com/osmosis-labs/osmosis/blob/f611e01c7fb9ab9d5348250cb0b65e08f3ec651b/x/concentrated-liquidity/tick.go#L46
- oe here: https://github.com/osmosis-labs/osmosis/blob/f611e01c7fb9ab9d5348250cb0b65e08f3ec651b/x/concentrated-liquidity/swaps.go#L312

mattverse:
Closing this issue as an outcome from investigation that the tick state does not need to get updated during swaps, instead the fee accumulators need to be updated, will post a separate issue / PR for this task

PR #3910, opened 2023-01-03
[CL] Change `calcOutAmtGivenIn` and `calcInAmtGivenOut` to be non-mutative

mattverse:
Closes: [#3889 ](https://github.com/osmosis-labs/osmosis/issues/3889)

This PR changes the two calc methods, namely `calcOutAmtGivenIn` and `calcInAmtGivenOut` to be non mutative methods, as previously they have been mutating ticks of the pool when upon crossing tick during iteration process of the calculation.

Also added existing checks to existing test cases that ensures that pool has not been changed after performing calculation method.

## Testing and Verifying

Changed existing test to cover changes made within this PR


x/concentrated-liquidity/swaps.go
```diff
@@ -210,6 +210,7 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 	priceLimit sdk.Dec,
 	poolId uint64,
 ) (tokenIn, tokenOut sdk.Coin, updatedTick sdk.Int, updatedLiquidity, updatedSqrtPrice sdk.Dec, err error) {
+	ctx, _ = ctx.CacheContext()
```
p0mvn, line 213: We should return the second return of `CacheContext` from the caller (`calcOutAmtGivenIn`) itself. The second return is a callback to write the cache context.

The reason for this requirement is that if `calcOutAmtGivenIn` is called from `swapOtAmtGivenIn`, we actually do want to write the cache context to storage.

mattverse: I also noticed that part in the original issue, but did not quite understand. 

What's the reason we want to use write cache context here? Do we not store in state anyways using the results from calling the calc method anyways within the swap method?(https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/swaps.go#L167)

czarcas7ic: I am not sure if this is what you are saying Roman, but shouldnt we pass this cache ctx as a return value and then write in the swap function that calls this? If that is what you are saying, I agree. Without doing this, despite us calling applySwap, the tick boundaries that got passed wont be updated.

mattverse: Gotcha, this makes sense

mattverse:
@p0mvn  @czarcas7ic Would it be possible to give a quick conceptACK on the direction we're going that I put in https://github.com/osmosis-labs/osmosis/pull/3910/commits/da1c910435964a2c36470a1ce500249590e3daca before I make the changes further on?

x/concentrated-liquidity/swaps.go
```diff
@@ -209,10 +209,11 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 	swapFee sdk.Dec,
 	priceLimit sdk.Dec,
 	poolId uint64,
-) (tokenIn, tokenOut sdk.Coin, updatedTick sdk.Int, updatedLiquidity, updatedSqrtPrice sdk.Dec, err error) {
+) (cacheCtx sdk.Context, tokenIn, tokenOut sdk.Coin, updatedTick sdk.Int, updatedLiquidity, updatedSqrtPrice sdk.Dec, err error) {
+	ctx, _ = ctx.CacheContext()
```
p0mvn, line 213:
```suggestion
	ctx, writeCtx := ctx.CacheContext()
```

x/concentrated-liquidity/swaps.go
```diff
@@ -335,7 +336,7 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 	tokenIn = sdk.NewCoin(tokenInMin.Denom, amt0)
 	tokenOut = sdk.NewCoin(tokenOutDenom, amt1)
 
-	return tokenIn, tokenOut, swapState.tick, swapState.liquidity, swapState.sqrtPrice, nil
+	return ctx, tokenIn, tokenOut, swapState.tick, swapState.liquidity, swapState.sqrtPrice, nil
```
p0mvn, line 339:
```suggestion
	return ctx, tokenIn, tokenOut, swapState.tick, swapState.liquidity, swapState.sqrtPrice, nil
```
```suggestion
	return writeCtx, tokenIn, tokenOut, swapState.tick, swapState.liquidity, swapState.sqrtPrice, nil
```

x/concentrated-liquidity/swaps.go
```diff
@@ -178,12 +178,12 @@ func (k Keeper) CalcOutAmtGivenIn(
 	tokenIn sdk.Coin,
 	tokenOutDenom string,
 	swapFee sdk.Dec,
-) (tokenOut sdk.Coin, err error) {
-	_, tokenOut, _, _, _, err = k.calcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, sdk.ZeroDec(), poolI.GetId())
+) (newCtx sdk.Context, tokenOut sdk.Coin, err error) {
```
p0mvn, line 181:
```suggestion
) (writeCache func(), tokenOut sdk.Coin, err error) {
```

p0mvn (commented):
This is what I meant in an earlier discussion

mattverse:
@p0mvn Thanks for the clarification. I'm not knowledgable of writeCtx() function that much, but AFAIU, writeCtx works in pair with the cached context returned by `ctx.CacheContext()`. That is, AFAIU, only returning and calling the writeCtx function would not have effect on the actual context being used. See https://github.com/osmosis-labs/osmosis/pull/3910/commits/17385583032115119609bf3b54cc9234b5ce321c#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R584-R593 for example. This was why I originally tried the approach of returning the context itself. Is there something I'm misunderstanding about how writeCtx works?

p0mvn:
> @p0mvn Thanks for the clarification. I'm not knowledgable of writeCtx() function that much, but AFAIU, writeCtx works in pair with the cached context returned by `ctx.CacheContext()`. That is, AFAIU, only returning and calling the writeCtx function would not have effect on the actual context being used. See [1738558#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R584-R593](https://github.com/osmosis-labs/osmosis/commit/17385583032115119609bf3b54cc9234b5ce321c#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R584-R593) for example. This was why I originally tried the approach of returning the context itself. Is there something I'm misunderstanding about how writeCtx works?

What are you trying to test in the linked example? It seems that you're trying to assert that the pool has been updated.

However, it would not be because we write pool data to store in `applySwap` function:
https://github.com/osmosis-labs/osmosis/blob/794e3b10a50e92c8a2e8b40042c7f93c6d628329/x/concentrated-liquidity/swaps.go#L146

In fact, currently, we do not write anything to store inside `calcOutAmtGivenIn` and `calcInAmtGivenOut`. However, we should be writing ticks: https://github.com/osmosis-labs/osmosis/issues/3890

This trick this `writeCtx` function is to make sure that we control when ticks are written to store once https://github.com/osmosis-labs/osmosis/issues/3890 is implemented

p0mvn:
> > > @p0mvn Thanks for the clarification. I'm not knowledgable of writeCtx() function that much, but AFAIU, writeCtx works in pair with the cached context returned by `ctx.CacheContext()`. That is, AFAIU, only returning and calling the writeCtx function would not have effect on the actual context being used. See [1738558#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R584-R593](https://github.com/osmosis-labs/osmosis/commit/17385583032115119609bf3b54cc9234b5ce321c#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R584-R593) for example. This was why I originally tried the approach of returning the context itself. Is there something I'm misunderstanding about how writeCtx works?
> > 
> > 
> > What are you trying to test in the linked example? It seems that you're trying to assert that the pool has been updated.
> > However, it would not be because we write pool data to store in `applySwap` function:
> > https://github.com/osmosis-labs/osmosis/blob/794e3b10a50e92c8a2e8b40042c7f93c6d628329/x/concentrated-liquidity/swaps.go#L146
> > 
> > In fact, currently, we do not write anything to store inside `calcOutAmtGivenIn` and `calcInAmtGivenOut`. However, we should be writing ticks: #3890
> > This trick this `writeCtx` function is to make sure that we control when ticks are written to store once #3890 is implemented
> 
> I see, then do you have suggestion upon testing on state change by calling the writeCtx function as linked in my original question? If not, would pass on the testings for this specific case until PR #3890 is implemented

I think it's fine to pass testing until #3890 is implemented. Could you please make sure that we track adding these tests in #3890?

mattverse:
@p0mvn made https://github.com/osmosis-labs/osmosis/issues/3992 to track this. Also this PR is r4r, please feel free to give reviews for this PR

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1000,6 +1014,16 @@ func (s *KeeperTestSuite) TestCalcAndSwapInAmtGivenOut() {
 				// check that liquidity is what we expected
 				expectedLiquidity := math.GetLiquidityFromAmounts(DefaultCurrSqrtPrice, lowerSqrtPrice, upperSqrtPrice, test.poolLiqAmount0, test.poolLiqAmount1)
 				s.Require().Equal(expectedLiquidity.String(), updatedLiquidity.String())
+
+				// // check that the pool has not been modified after performing calc
```
mattverse, line 1018:
```suggestion
				// check that the pool has not been modified after performing calc
```

x/concentrated-liquidity/swaps.go
```diff
@@ -138,10 +138,11 @@ func (k Keeper) SwapOutAmtGivenIn(
 	priceLimit sdk.Dec,
 	poolId uint64,
 ) (calcTokenIn, calcTokenOut sdk.Coin, currentTick sdk.Int, liquidity, sqrtPrice sdk.Dec, err error) {
-	tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, poolId)
+	writeCtx, tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcOutAmtGivenIn(ctx, tokenIn, tokenOutDenom, swapFee, priceLimit, poolId)
 	if err != nil {
 		return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
 	}
+	writeCtx()
```
p0mvn, line 145:
```suggestion

        // N.B. making the call below ensures that any mutations done inside calcOutAmtGivenIn
        // are written to store. If this call were skipped, calcOutAmtGivenIn would be non-mutative.
        // An example of a store write done in calcOutAmtGivenIn is updating ticks as we cross them.
	writeCtx()
```

x/concentrated-liquidity/swaps.go
```diff
@@ -159,10 +160,11 @@ func (k *Keeper) SwapInAmtGivenOut(
 	priceLimit sdk.Dec,
 	poolId uint64,
 ) (calcTokenIn, calcTokenOut sdk.Coin, currentTick sdk.Int, liquidity, sqrtPrice sdk.Dec, err error) {
-	tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcInAmtGivenOut(ctx, desiredTokenOut, tokenInDenom, swapFee, priceLimit, poolId)
+	writeCtx, tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcInAmtGivenOut(ctx, desiredTokenOut, tokenInDenom, swapFee, priceLimit, poolId)
 	if err != nil {
 		return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
 	}
+	writeCtx()
```
p0mvn, line 167:
```suggestion

        // N.B. making the call below ensures that any mutations done inside calcInAmtGivenOut
        // are written to store. If this call were skipped, calcInAmtGivenOut would be non-mutative.
        // An example of a store write done in calcInAmtGivenOut is updating ticks as we cross them.
	writeCtx()
```

x/concentrated-liquidity/swaps.go
```diff
@@ -203,16 +205,18 @@ func (k Keeper) CalcInAmtGivenOut(
 // calcOutAmtGivenIn calculates tokens to be swapped out given the provided amount and fee deducted. It also returns
 // what the updated tick, liquidity, and currentSqrtPrice for the pool would be after this swap.
 // Note this method is non-mutative, so the values returned by CalcOutAmtGivenIn do not get stored
+// Instead, we return writeCtx function so that the caller of this method can decide to write the cached ctx to store or not.
```
p0mvn, line 208: Do you mind adding spec for `calcInAmtGivenOut` as a drive by change in this PR please?

Just noticed it's missing

p0mvn (approved):
LGTM

p0mvn (approved):

p0mvn:
Going to merge this as discussed offline.

We can iterate on this further in https://github.com/osmosis-labs/osmosis/issues/3890 and https://github.com/osmosis-labs/osmosis/issues/3992. Thanks for making the issue

PR #3911, opened 2023-01-03
Integrate Packet Forward Middleware

nicolaslara:

Integrate the packet forward middleware

## Brief Changelog

- Integrated packet forward middleware. 
- Should this be in the changelog?

## Testing and Verifying
- All existing tests that use IBC pass (especially E2E)
- Should we add an custom E2E test for PFM?

  - How is the feature or change documented? (not applicable   )


app/keepers/keepers.go
```diff
@@ -492,10 +502,32 @@ func (appKeepers *AppKeepers) WireICS20PreWasmKeeper(
...
+
+	// Packet Forward Middleware
+	// Initialize packet forward middleware router
+	appKeepers.PacketForwardKeeper = packetforwardkeeper.NewKeeper(
+		appCodec,
+		appKeepers.keys[packetforwardtypes.StoreKey],
+		appKeepers.GetSubspace(packetforwardtypes.ModuleName),
+		appKeepers.TransferKeeper,
+		appKeepers.IBCKeeper.ChannelKeeper,
+		appKeepers.DistrKeeper,
```
ValarDragon, line 514: Uhh, why does it need this? Community pool?

tests/e2e/e2e_test.go
```diff
@@ -333,6 +334,83 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
...
+	chainA.LatestCodeId = int(nodeA.QueryLatestWasmCodeID())
+	nodeA.InstantiateWasmContract(
+		strconv.Itoa(chainA.LatestCodeId),
+		`{"count": 0}`,
+		initialization.ValidatorWalletName)
+
+	contracts, err := nodeA.QueryContractsFromId(chainA.LatestCodeId)
+	s.NoError(err)
+	s.Require().Len(contracts, 1, "Wrong number of contracts for the counter")
+	contractAddr := contracts[0]
```
ValarDragon, line 364: Lets pull this into its own method called `UploadAndInstantiateCounter` or smth.

tests/e2e/e2e_test.go
```diff
@@ -333,6 +334,83 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
...
+		balance, err := nodeA.QueryBalances(contractAddr)
+		s.Require().NoError(err)
+		if len(balance) == 0 {
+			return false
+		}
+		return balance[0].Amount.Int64() == transferAmount
+	},
+		1*time.Minute,
+		10*time.Millisecond,
+	)
```
ValarDragon, line 393: Note for future, we should have this go through a helper, and then be a one liner.

`s.CheckBalance(addr, expectedbalance)`

tests/e2e/e2e_test.go
```diff
@@ -333,6 +334,83 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
 	)
 }
 
+func (s *IntegrationTestSuite) TestPacketForwarding() {
```
ValarDragon, line 337: Can we add some code comments to this for what chain were doing an action on, and what chain were querying at different points in this?

app/keepers/keepers.go
nicolaslara: yeah, the fee can be configured via a param as a percentage of the send and it funds the community pool. 

Ideally this should go to stakers IMO, but the fee defaults to 0, so we can change that later

tests/e2e/e2e_test.go
nicolaslara: refactored

tests/e2e/e2e_test.go
nicolaslara: extracted it for reusability

tests/e2e/e2e_test.go
nicolaslara: Added comments

app/keepers/keepers.go
ValarDragon: RIP, this is bad module design imo, should be a middleware responsibility

nicolaslara:
@ValarDragon Updated to the newer version of PFM which solves the json encoding issue we discussed. This should be ready for re-review/merge

app/keepers/keepers.go
agouin: Can you please expand on this @ValarDragon ? Would love to understand your perspective so we can improve it. I'm not sure what you mean, as this is the keeper for the IBC middleware.

ValarDragon: I don't see why your transport layer for packet sending should have bips of fee bundled in at default. This is equivalent in design practice to saying our API for bank sends should take in a fee parameter as well, and have bips of all sends be captured.

You could in principle want this, but this can be pushed to wrappers of the main object, rather than being the main thing you import. (I'd prefer no capability of this in what we import, than having such a parameter forced)

ValarDragon: (NOTE: This is very minor on the overall work going into this. Awesome job on the module! And I think my first comment "bad module design imo" was too harsh for the severity of opinion actually held)

ValarDragon (approved):

tests/e2e/e2e_test.go
```diff
@@ -359,6 +384,57 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
...
+	senderBech32, err := ibchookskeeper.DeriveIntermediateSender("channel-0", validatorAddr, "osmo")
+	var response map[string]interface{}
+	s.Eventually(func() bool {
+		response, err = nodeA.QueryWasmSmart(contractAddr, fmt.Sprintf(`{"get_count": {"addr": "%s"}}`, senderBech32))
+		if err != nil {
+			return false
+		}
+		count := response["count"].(float64)
+		// check if denom contains "uosmo"
+		return err == nil && count == 0
```
ValarDragon, line 431: We need to handle this TODO right? IMO we should ensure exact osmo amount received, in order to be certain that our default parameter has no fee taken.

ValarDragon:
Needs changelog pre-merge

tests/e2e/e2e_test.go
nicolaslara: I think that comment is a leftover from previous code. The balance check is done above (line419) but it's not checking the denom, just the amount. I'll add the denom check.

tests/e2e/e2e_test.go
```diff
@@ -27,6 +32,30 @@ import (
...
+		// check that the amount is in one of the balances inside the balance list
+		for _, b := range balance {
+			if b.Denom == denom && b.Amount.Int64() == amount {
+				return true
+			}
+		}
+		return false
+	},
+		1*time.Minute,
+		10*time.Millisecond,
```
p0mvn, line 55: I think that CI might be flaky and blocking requests due to retrying this too frequently. I've run into similar issues before where AFAIR increasing this number helped.

I looked into the node logs: https://github.com/osmosis-labs/osmosis/suites/10828110105/artifacts/545247532 and they seem okay.  So I'm guessing it might be some setup problem

Issue #3798, opened 2022-12-19
[CL][Swap Router]: move poolIsActive() check from gamm to swaprouter


p0mvn:
TODO: consider moving this to swaprotuer

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/3764#discussion_r1051491467_

p0mvn:
Blocked on: https://github.com/osmosis-labs/osmosis/issues/3807 . We need to make sure that API is introduced to make sure that we don't miss adding this check to it.

mattverse:
@p0mvn  What's the reason we want to move that check to the swaprouter? Seems that the caller function(`SwapExactAmountIn`) is getting called in other places as well within the gamm module, I'm concerned that if we move that check to the swap router layer, any other methods calling the `SwapExactAmountIn` would not have the isActive check

p0mvn:
The reason is that it would be great to perform the `IsActive` check for CL pools as well. As a result, x/swaprouter might be a good shared location for doing so.

I think the issue you are referring to (ref: https://github.com/osmosis-labs/osmosis/issues/3130) is orthogonal toIsActive`,

No client other than swaprouter should call `SwapExactAmountIn` in `x/gamm`. https://github.com/osmosis-labs/osmosis/issues/3130 is to investigate how we can prevent these other clients from doing so.

The `IsActive` check here is to have a quick switch guard for us to turn of all pools quickly (including CL) if there is a security incident in progress given, assuming that #3130 is addressed

PR #3920, opened 2023-01-04
[CL][SwapRouter]: Move Inactive pool check to Swaprouter module

mattverse:
Closes: #3798

This PR adds inactive pool check to the swap router module, more specifically does a check before each swap if the pool is inactive or not. If not, errors out.

## Testing and Verifying

This PR changed existing test for inactive pool from using gamm module -> swaprouter module.

### DIFF
--- a/x/gamm/keeper/swap_test.go
+++ b/x/gamm/keeper/swap_test.go
@@ -505,8 +505,32 @@ func (suite *KeeperTestSuite) TestInactivePoolFreezeSwaps() {
 	for _, test := range testCases {
 		suite.Run(test.name, func() {
 			// Check swaps
-			_, swapInErr := gammKeeper.SwapExactAmountIn(suite.Ctx, suite.TestAccs[0], test.pool, testCoin, "bar", sdk.ZeroInt(), sdk.ZeroDec())
-			_, swapOutErr := gammKeeper.SwapExactAmountOut(suite.Ctx, suite.TestAccs[0], test.pool, "bar", sdk.NewInt(1000000000000000000), testCoin, sdk.ZeroDec())
+			_, swapInErr := suite.App.SwapRouterKeeper.RouteExactAmountIn(
+				suite.Ctx,
+				suite.TestAccs[0],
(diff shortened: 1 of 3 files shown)


x/gamm/keeper/swap_test.go
```diff
@@ -505,8 +505,32 @@ func (suite *KeeperTestSuite) TestInactivePoolFreezeSwaps() {
 	for _, test := range testCases {
 		suite.Run(test.name, func() {
 			// Check swaps
-			_, swapInErr := gammKeeper.SwapExactAmountIn(suite.Ctx, suite.TestAccs[0], test.pool, testCoin, "bar", sdk.ZeroInt(), sdk.ZeroDec())
-			_, swapOutErr := gammKeeper.SwapExactAmountOut(suite.Ctx, suite.TestAccs[0], test.pool, "bar", sdk.NewInt(1000000000000000000), testCoin, sdk.ZeroDec())
+			_, swapInErr := suite.App.SwapRouterKeeper.RouteExactAmountIn(
```
p0mvn, line 508: Wdyt about moving this whole test to swaprouter?

p0mvn (approved):
LGTM once test is moved

czarcas7ic (approved):

x/gamm/keeper/swap_test.go
mattverse: I would also prefer this way, but one major blocker in doing so is that right now we don't have access to be setting pools directly outside the module via `setPool`. But then, we come across the problem of not being able to set an inactive pool. Lmk if you have thoughts for work around for this

p0mvn: Makes sense. The only thing I can think of is unexporting the `setPool`. 

However, I think between the choice of unexporting the method and keeping this test in x/gamm, I would prefer the latter.

Let's keep the test here for now, and if we still find this awkward in the future, we can think about this deeper.

I'm going to merge the PR but let me know if you still would like to continue discussing this separately.

PR #3922, opened 2023-01-04
test(x/twap): geometric twap edge case tests

p0mvn:
Closes: #XXX

Adds geometric twap edge case tests. Defines min spot price as `sdk.SmallestDec()`

### Overflow test

The old test case setup was incorrect as it was taking the logarithm of the accumulator instead of just the price. To mimic the real environment, this test case is now incorporated into `TestComputeGeometricStrategyTwap` table-driven test with quote assets in both denomes.

With one of the denoms, there is a panic due to precision loss. This happens because we end up taking 1 / < extremely large value >, resulting in a value smaller than the smallest Dec possible.

### Underflow Test

This test case assumes that the spot price is the smallest dec possible. By taking a logarithm of the smallest value (0, 1) and multiplying it by 100 years, we confirm that our math functions do not accidentally cause an underflow.

The test is run in both denoms as quote assets.

### Zero Accumulator Differences

This unlikely case might happen if all spot prices in history had returned a spot price error, resulting in geometric accumulators for start and end records as well as their difference always being 0.

While this is unlikely, this test helps to ensure that we have accounted for all kinds of risky outcomes.

### DIFF
--- a/osmomath/binary_search.go
+++ b/osmomath/binary_search.go
@@ -107,7 +107,7 @@ func (e ErrTolerance) CompareBigDec(expected BigDec, actual BigDec) int {
 	comparisonSign := 0
 	if expected.GT(actual) {
 		comparisonSign = 1
-	} else {
+	} else if expected.LT(actual) {
 		comparisonSign = -1
 	}
 
--- a/x/twap/strategy_test.go
+++ b/x/twap/strategy_test.go
@@ -249,15 +249,12 @@ func (s *TestSuite) TestComputeGeometricStrategyTwap() {
 			expTwap:     gammtypes.MaxSpotPrice,
 		},
 
-		"expected precision loss test: - at spot price denom1 quote - panic": {
+		"expected precision loss test: - at spot price denom1 quote - return zero": {
 			startRecord: withSp0(baseRecord, gammtypes.MaxSpotPrice),
 			endRecord:   newOneSidedGeometricRecord(baseRecord.Time.Add(oneHundredYearsMin1MsDuration), overflowTestCaseAccumDiff),
 			quoteAsset:  denom1,
 
-			// TODO: analyze if panic is acceptable.
-			// Should be acceptable since underlying logic is
-			// only called from the query path.
-			expPanic: true,
+			expTwap: sdk.ZeroDec(),
 		},
 
 		"no underflow test: spot price is smallestpossible denom0 quote": {
@@ -274,15 +271,12 @@ func (s *TestSuite) TestComputeGeometricStrategyTwap() {
 			expTwap:     sdk.OneDec().Quo(gammtypes.MinSpotPrice),
 		},
 
-		"zero accum difference ": {
+		"zero accum difference - return zero": {
 			startRecord: newOneSidedGeometricRecord(baseTime, sdk.ZeroDec()),
 			endRecord:   newOneSidedGeometricRecord(baseRecord.Time.Add(time.Millisecond), sdk.ZeroDec()),
 			quoteAsset:  denom1,
 
-			// TODO: analyze if panic is acceptable.
-			// Should be acceptable since underlying logic is
-			// only called from the query path.
-			expPanic: true,
+			expTwap: sdk.ZeroDec(),
 		},
 	}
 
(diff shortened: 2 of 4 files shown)


x/twap/strategy_test.go
```diff
@@ -166,6 +162,34 @@ func (s *TestSuite) TestComputeArithmeticStrategyTwap() {
...
+		// Compute accumulator difference for the overflow test case by
+		// taking log base 2 of the max spot price
+		maxSpotPriceLogBase2 = twap.TwapLog(gammtypes.MaxSpotPrice)
+
+		oneHundredYearsInHours        int64 = 100 * 365 * 24
+		oneHundredYears                     = OneSec.MulInt64(60 * 60 * oneHundredYearsInHours)
+		oneHundredYearsMin1MsDuration       = time.Duration(oneHundredYearsInHours)*time.Hour - time.Millisecond
+
+		// Subtract 1ms from 100 years to assume that we interpolate.
+		oneHundredYearsMin1Ms = oneHundredYears.Sub(oneDec)
```
czarcas7ic, line 184: Is oneDec in this case 1ms or a second?

czarcas7ic (commented):
I'm lacking context to approve but the logic feels sound

x/twap/strategy_test.go
p0mvn: This should be in milliseconds. That is one dec = 1ms

x/twap/strategy_test.go
```diff
@@ -211,28 +235,69 @@ func (s *TestSuite) TestComputeGeometricStrategyTwap() {
...
-		// - large time delta
-		// - both
 
-		// TODO: hand calculated tests
+		"no overflow test: at max spot price denom0 quote - get max spot price": {
+			startRecord: withSp0(baseRecord, gammtypes.MaxSpotPrice),
+			endRecord:   newOneSidedGeometricRecord(baseRecord.Time.Add(oneHundredYearsMin1MsDuration), overflowTestCaseAccumDiff),
+			quoteAsset:  denom0,
+			expTwap:     gammtypes.MaxSpotPrice,
+		},
```
ValarDragon, line 250: LGTM

x/twap/strategy_test.go
```diff
@@ -211,28 +235,69 @@ func (s *TestSuite) TestComputeGeometricStrategyTwap() {
...
+		"expected precision loss test: - at spot price denom1 quote - panic": {
+			startRecord: withSp0(baseRecord, gammtypes.MaxSpotPrice),
+			endRecord:   newOneSidedGeometricRecord(baseRecord.Time.Add(oneHundredYearsMin1MsDuration), overflowTestCaseAccumDiff),
+			quoteAsset:  denom1,
+
+			// TODO: analyze if panic is acceptable.
+			// Should be acceptable since underlying logic is
+			// only called from the query path.
+			expPanic: true,
+		},
```
ValarDragon, line 261: I don't understand why this is panicking?

ValarDragon: Ah we discussed, its due to divide by 0 right here: https://github.com/osmosis-labs/osmosis/blob/5d360857e24cf1aa7c0d465fd9d16df6bb202357/x/twap/strategy.go#L53

Discussed to just add an if statement there

ValarDragon:
LGTM for merge once we add that if statement for divide by zero -> min spot price!

x/twap/strategy_test.go
p0mvn: Upon looking into this more, I think this might have been a panic caused by the test setup.

I updated the assertion of this test case to return zero, as we discussed, and it passes.

Given the current test cases, I don't think it is possible for that value to be zero. This gets executed only if the arithmetic mean of spot prices is negative which happens when spot prices are in range (0, 1). 

I will still add this check as a defense-in-depth but I can't think of a test case that would cover this logic

osmomath/binary_search.go
```diff
@@ -107,7 +107,7 @@ func (e ErrTolerance) CompareBigDec(expected BigDec, actual BigDec) int {
 	comparisonSign := 0
 	if expected.GT(actual) {
 		comparisonSign = 1
-	} else {
+	} else if expected.LT(actual) {
```
p0mvn, line 110: Note: I found out that logic here was broken for geometric tolerance

x/twap/strategy_test.go
p0mvn: Also, FYI that I'm working on reducing the number of these multiplicative inverses that can cause the precision loss here:
https://github.com/osmosis-labs/osmosis/pull/3924

osmomath/binary_search.go
ValarDragon: Ah your right

ValarDragon (approved):
Nice

PR #3924, opened 2023-01-04
refactor(x/twap): geometric twap reduce multiplicative inverses

p0mvn:
Addresses: https://github.com/osmosis-labs/osmosis/issues/3541#issuecomment-1360745153
Progress toward: https://github.com/osmosis-labs/osmosis/issues/3541

This is an optimization that can reduce the frequency of precision loss due to taking multiplicative inverses. 

Currently, we might have to do 2 redundant multiplicative inverses:
1. If the arithmetic mean of log prices is negative, making the power function exponent be negative
2. if the quote asset is asset 1.

Instead, we can detect all cases when the multiplicative inverse is needed and do it only once.

These cases are listed in the comments:

**Case 1**: exponent is negative and quoteAsset is asset 0.
- This means that we need to invert the result to get the true value of the geometric mean.

**Case 2**: exponent is positive and quoteAsset is asset 1.
- We need to use the following property: the geometric mean of reciprocals is reciprocal of the geometric mean.

## Testing and Verifying

### DIFF
--- a/x/twap/export_test.go
+++ b/x/twap/export_test.go
@@ -5,6 +5,7 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	"github.com/osmosis-labs/osmosis/osmomath"
 	"github.com/osmosis-labs/osmosis/v13/x/twap/types"
 )
 
@@ -91,8 +92,14 @@ func TwapLog(x sdk.Dec) sdk.Dec {
 	return twapLog(x)
 }
 
-func TwapPow(x sdk.Dec) sdk.Dec {
-	return twapPow(x)
+// twapPow exponentiates 2 to the given exponent.
+// Used as a test-helper for the power function used in geometric twap.
+func TwapPow(exponent sdk.Dec) sdk.Dec {
+	exp2 := osmomath.Exp2(osmomath.BigDecFromSDKDec(exponent.Abs()))
+	if exponent.IsNegative() {
+		return osmomath.OneDec().Quo(exp2).SDKDec()
+	}
+	return exp2.SDKDec()
 }
 
 func GetSpotPrices(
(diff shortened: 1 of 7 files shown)


x/twap/strategy.go
```diff
@@ -50,14 +50,25 @@ func (s *geometric) computeTwap(startRecord types.TwapRecord, endRecord types.Tw
...
+
+	// Case 1: exponent is negative and quoteAsset is asset 0.
+	// This means that we need to invert the result to get the true value of the geometric mean.
+	invertCase1 := isExponentNegative && isQuoteAsset0
+	// Case 2: exponent is positive and quoteAsset is asset 1.
+	// We need to use the following property: geometric mean of recprocals is reciprocal of geometric mean.
 	// https://proofwiki.org/wiki/Geometric_Mean_of_Reciprocals_is_Reciprocal_of_Geometric_Mean
-	if quoteAsset == startRecord.Asset1Denom {
-		result = sdk.OneDec().Quo(result)
+	invertCase2 := !isExponentNegative && !isQuoteAsset0
```
czarcas7ic, line 66: Only thing I can think of here is we are just checking if its not the quoteAsset, which SHOULD mean its the base asset as long as checks are done properly before this.

czarcas7ic (approved):

x/twap/export_test.go
```diff
@@ -91,8 +92,14 @@ func TwapLog(x sdk.Dec) sdk.Dec {
...
-func TwapPow(x sdk.Dec) sdk.Dec {
-	return twapPow(x)
+// twapPow exponentiates 2 to the given exponent.
+// Used as a test-helper for the power function used in geometric twap.
+func TwapPow(exponent sdk.Dec) sdk.Dec {
+	exp2 := osmomath.Exp2(osmomath.BigDecFromSDKDec(exponent.Abs()))
+	if exponent.IsNegative() {
+		return osmomath.OneDec().Quo(exp2).SDKDec()
+	}
+	return exp2.SDKDec()
```
ValarDragon, line 102: Confirmed that this is only used for building test cases.

PR #3946, opened 2023-01-07
Document manual test script

nicolaslara:
This just adds a bit of documentation to the helper script for doing manual testing of ibc hooks.


tests/ibc-hooks/test_hooks.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
+#
+# This script can be used to manually test the ibc hooks. It is meant as a guid and not to be run directly
```
mattverse, line 3:
```suggestion
# This script can be used to manually test the ibc hooks. It is meant as a guide and not to be run directly 
```

mattverse (approved):
🌮

Issue #3950, opened 2023-01-09
feat(osmoutils/accum): remove the accumulator from state


p0mvn:
## Background

Currently, we never remove the accumulator from state. We need to define when the accumulator is deleted to avoid polluting the state.

## Suggested Design

Address the following TODO:
https://github.com/osmosis-labs/osmosis/blob/3e26ec2e1bf66becfbf8d9b8606fce7dd3e0cbe5/osmoutils/accum/accum.go#L277

## Acceptance Criteria
- TODO addressed
- Tests added

PR #3961, opened 2023-01-09
feat(osmoutil/accum): Remove position Accumulator from state

mattverse:
Closes: #3950 

Removes the position accumulator from state when the numShares for the position is equal to zero. 

This check / action happens when the update Position, specifically `RemoveFromPosition` is happening.

## Brief Changelog
- Remove position accumulator from state when numshares is zero

## Testing and Verifying
Added new tests for this / changed existing test cases

### DIFF
--- a/osmoutils/accum/accum.go
+++ b/osmoutils/accum/accum.go
@@ -196,7 +196,7 @@ func (accum AccumulatorObject) RemoveFromPositionCustomAcc(name string, numShare
 
 	// Ensure not removing more shares than exist
 	if numSharesToRemove.GT(position.NumShares) {
-		return fmt.Errorf("Attempted to remove more shares  (%s) than exist in the position (%s)", numSharesToRemove, position.NumShares)
+		return fmt.Errorf("Attempted to remove more shares (%s) than exist in the position (%s)", numSharesToRemove, position.NumShares)
 	}
 
 	// Save current number of shares and unclaimed rewards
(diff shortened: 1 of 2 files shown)


osmoutils/accum/accum.go
```diff
@@ -206,6 +206,14 @@ func (accum AccumulatorObject) RemoveFromPositionCustomAcc(name string, numShare
 		return err
 	}
 
+	newNumShares := oldNumShares.Sub(numSharesToRemove)
+
+	// remove the position from state entirely if numShares = zero
+	if newNumShares.Equal(sdk.ZeroDec()) {
+		accum.DeletePosition(name)
```
p0mvn, line 213: We should be deleting it in `ClaimRewards`.

If we delete it here, some uncollected rewards end up being unaccounted for

osmoutils/accum/accum.go
```diff
@@ -239,6 +247,10 @@ func (accum AccumulatorObject) UpdatePositionCustomAcc(name string, numShares sd
 	return accum.AddToPositionCustomAcc(name, numShares, customAccumulatorValue)
 }
 
+func (accum AccumulatorObject) DeletePosition(name string) {
```
p0mvn, line 250: We should unexpert this because it is unsafe to delete a position anywhere other than `ClaimRewards`

osmoutils/accum/accum.go
mattverse: @p0mvn  But does any position.NumShares happen in ClaimRewards?

osmoutils/accum/accum.go
mattverse: agreed

osmoutils/accum/accum.go
p0mvn: > Are we going to enforce calling ClaimRewards when closing positions?

That's the plan

> what do you think about leaving this here,

We cannot delete position from state here because any rewards, including unclaimed, would just disappear

mattverse: > > Are we going to enforce calling ClaimRewards when closing positions?
> 
> That's the plan
> 
> > what do you think about leaving this here,
> 
> We cannot delete position from state here because any rewards, including unclaimed, would just disappear

If this is the plan, agreed on having this check on closePosition

mattverse:
@p0mvn Changed to deleting positions when claiming reward, and added test for it r4r

p0mvn (approved):
lgtm

PR #3962, opened 2023-01-09
fix(CL Swaps): incorrect rounding; redundant computations

p0mvn:
Closes: #XXX

These problems have been discovered while working on fees in https://github.com/osmosis-labs/osmosis/pull/3893

- Problem 1: incorrect rounding

In `math.go`, we are expecting a division with rounding down. However, `Quo` is used which does banker's rounding on the precision end. While this might seem like a small issue, it was causing panic with fees at the edge cases.

- Problem 2: redundant computations

There are some redundant re-computations in swap strategies. Note that this change is equivalent to the original. The only difference is removing one redundant computation. This change does not affect change modifications.

## Testing and Verifying

This change is already covered by existing tests. Values changed.


x/concentrated-liquidity/internal/math/math.go
```diff
@@ -113,7 +113,7 @@ func GetNextSqrtPriceFromAmount0RoundingUp(sqrtPriceCurrent, liquidity, amountRe
 // getNextSqrtPriceFromAmount1RoundingDown utilizes the current squareRootPrice, liquidity of denom1, and amount of denom1 that still needs
 // to be swapped in order to determine the next squareRootPrice
 func GetNextSqrtPriceFromAmount1RoundingDown(sqrtPriceCurrent, liquidity, amountRemaining sdk.Dec) (sqrtPriceNext sdk.Dec) {
-	return sqrtPriceCurrent.Add(amountRemaining.Quo(liquidity))
+	return sqrtPriceCurrent.Add(amountRemaining.QuoTruncate(liquidity))
```
mattverse, line 116: Does this mean that we want to truncate instead of doing baker's rounding?

p0mvn: Yes, that's correct.

Rounding down was the intended behavior given the method name.

Additionally, I've run into panics while testing fees stemming from the fact that we do not round down / truncate here.

x/concentrated-liquidity/internal/swapstrategy/swap_strategy_test.go
```diff
@@ -59,9 +59,9 @@ func (suite *StrategyTestSuite) TestComputeSwapState() {
 			// TODO: test case where it does affect.
 			sqrtPriceLimit:        sdk.MustNewDecFromStr("70.738956735309575810").Add(sdk.OneDec()), // 5003
 			zeroForOne:            false,
-			expectedSqrtPriceNext: "70.738348247484497718",
-			expectedAmountIn:      "42000000.000000000749226725",
-			expectedAmountOut:     "8396.714242162445246671",
+			expectedSqrtPriceNext: "70.738348247484497717",
+			expectedAmountIn:      "41999999.999999999231344381",
+			expectedAmountOut:     "8396.714242162444943332",
```
mattverse, line 64: This seems like changes coming from moving the `CalcAmount0Delta` in`ComputeSwapState`. Are these result changes safe?

x/concentrated-liquidity/internal/math/math.go
mattverse: ACK On this change

mattverse (approved):
ACK

PR #3970, opened 2023-01-10
fix(twap/upgrades): geometric twap accumulator initialization; testnet bug reproduction

p0mvn:
Closes: #XXX

In https://github.com/osmosis-labs/osmosis/issues/3542, it was concluded that we do not need to perform the migrations to initialize the geometric twap accumulator.

However, the testnet halted due to the geometric twap accumulator ending up uninitialized.

The reason why our e2e test did not catch this problem was because we neve swapped against a pool that was created pre-upgrade after the upgrade itself.

This PR reproduced the testnet halt bug and fixed it by introducing twap record migrations that were originally spearheaded  in: https://github.com/osmosis-labs/osmosis/pull/3729

## Brief Changelog

- reproduce testnet halt
- introduce twap migrations from: https://github.com/osmosis-labs/osmosis/pull/3729
- test more thoroughly and document

## Testing and Verifying

This change added e2e and migrations tests.

### DIFF
--- a/x/twap/migrate.go
+++ b/x/twap/migrate.go
@@ -19,7 +19,7 @@ func NewMigrator(keeper Keeper) Migrator {
 }
 
 func (m Migrator) Migrate1To2(ctx sdk.Context) error {
-	return m.keeper.initializeGeometricTwap(ctx)
+	return m.keeper.initializeGeometricTwapAcc(ctx, sdk.ZeroDec())
 }
 
 // MigrateExistingPools iterates through all pools and creates state entry for the twap module.
@@ -68,7 +68,7 @@ func (k Keeper) MigrateTwapRecordsToGeometric(ctx sdk.Context) error {
 	return nil
 }
 
-func (k Keeper) initializeGeometricTwap(ctx sdk.Context) error {
+func (k Keeper) initializeGeometricTwapAcc(ctx sdk.Context, value sdk.Dec) error {
 	// In ascending order by time.
 	historicalTimeIndexed, err := k.getAllHistoricalTimeIndexedTWAPs(ctx)
 	if err != nil {
@@ -82,17 +82,19 @@ func (k Keeper) initializeGeometricTwap(ctx sdk.Context) error {
 	// Since we are iterate over time-indexed records in ascending order,
 	// most recent record should also be updated correctly.
 	for i, record := range historicalTimeIndexed {
+		record := record
 		// Sanity check order.
 		if i > 0 {
 			previousRecord := historicalTimeIndexed[i-1]
 
+			isSamePool := previousRecord.PoolId == record.PoolId
 			isCorrectOrder := previousRecord.Time.Before(record.Time)
-			if !isCorrectOrder {
+			if !isCorrectOrder && !isSamePool {
 				return fmt.Errorf("error: historical twap records are not in ascending order, (%v), was after (%v)", previousRecord, record)
 			}
 		}
 
-		record.GeometricTwapAccumulator = sdk.ZeroDec()
+		record.GeometricTwapAccumulator = value
 		k.storeNewRecord(ctx, record)
 	}
 	return nil
--- a/x/twap/migrate.go
+++ b/x/twap/migrate.go
@@ -5,9 +5,6 @@ import (
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	"github.com/osmosis-labs/osmosis/osmoutils"
-	"github.com/osmosis-labs/osmosis/v13/x/twap/types"
 )
 
 type Migrator struct {
@@ -30,41 +27,6 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
 			return err
 		}
 	}
-	return nil
-}
-
-func (k Keeper) MigrateTwapRecordsToGeometric(ctx sdk.Context) error {
-
-	// types
-
-	allMostRecetRecords, err := k.getAllMostRecentRecords(ctx)
-	if err != nil {
-		return err
-	}
-
-	store := ctx.KVStore(k.storeKey)
-
-	for _, record := range allMostRecetRecords {
-		record := record
-		record.GeometricTwapAccumulator = sdk.ZeroDec()
-		key := types.FormatMostRecentTWAPKey(record.PoolId, record.Asset0Denom, record.Asset1Denom)
-		osmoutils.MustSet(store, key, &record)
-	}
-
-	// make available for GC
-	allMostRecetRecords = nil
-
-	allHistoricalRecords, err := k.getAllHistoricalPoolIndexedTWAPs(ctx)
-
-	for _, record := range allHistoricalRecords {
-		record := record
-		record.GeometricTwapAccumulator = sdk.ZeroDec()
-		k.storeHistoricalTWAP(ctx, record)
-	}
-
-	// make available for GC
-	allHistoricalRecords = nil
-
 	return nil
 }
 
--- a/x/twap/migrate.go
+++ b/x/twap/migrate.go
@@ -1,11 +1,26 @@
 package twap
 
 import (
+	"errors"
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/osmosis-labs/osmosis/osmoutils"
 	"github.com/osmosis-labs/osmosis/v13/x/twap/types"
 )
+
+type Migrator struct {
+	keeper Keeper
+}
+
+func NewMigrator(keeper Keeper) Migrator {
+	return Migrator{keeper: keeper}
+}
+
+func (m Migrator) Migrate1To2(ctx sdk.Context) error {
+	return m.keeper.initializeGeometricTwap(ctx)
+}
 
 // MigrateExistingPools iterates through all pools and creates state entry for the twap module.
 func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error {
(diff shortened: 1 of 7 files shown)


x/twap/migrate.go
```diff
@@ -14,3 +32,46 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
...
+	}
+
+	if len(historicalTimeIndexed) == 0 {
+		return errors.New("error: no historical twap records found")
+	}
+
+	// Since we are iterate over time-indexed records in ascending order,
+	// most recent record should also be updated correctly.
+	for i, record := range historicalTimeIndexed {
+		record := record
```
mattverse, line 58: why is this line needed?

x/twap/migrate_test.go
```diff
@@ -93,3 +93,189 @@ func (suite *TestSuite) TestTwapRecord_GeometricTwap_MarshalUnmarshal() {
 	suite.Require().False(originalRecord.GeometricTwapAccumulator.IsNil())
 	suite.Require().Equal(sdk.ZeroDec(), originalRecord.GeometricTwapAccumulator)
 }
+
+func (suite *TestSuite) TestInitializeGeometricTwap() {
+	var (
+		expectedGeometricAccumUpdate = sdk.OneDec()
```
mattverse, line 99: nit suggestion please feel free to resolve this / disregard this
```suggestion
		defaultGeometricAccumUpdateVal = sdk.OneDec()
```

x/twap/migrate_test.go
```diff
@@ -93,3 +93,189 @@ func (suite *TestSuite) TestTwapRecord_GeometricTwap_MarshalUnmarshal() {
...
+				threeAssetRecordWithPoolIdAndTime(basePoolId, baseTime),
+				threeAssetRecordWithPoolIdAndTime(basePoolId+1, baseTime.Add(time.Second))...),
+
+			expectedMostRecent: map[uint64][]types.TwapRecord{
+				baseRecord.PoolId:     updateGeometricAccum(threeAssetRecordWithPoolIdAndTime(basePoolId, baseTime)),
+				baseRecord.PoolId + 1: updateGeometricAccum(threeAssetRecordWithPoolIdAndTime(basePoolId+1, baseTime.Add(time.Second))),
+			},
+		},
+		"three pools, 2 at the same time, 1 is not, entries and different time for each": {
+			preExistingRecords: merge([][]types.TwapRecord{
```
mattverse, line 183: why not put in []types.TwapRecord directly instead of providing here [][] and using merge function?

x/twap/migrate_test.go
```diff
@@ -93,3 +93,189 @@ func (suite *TestSuite) TestTwapRecord_GeometricTwap_MarshalUnmarshal() {
...
+			suite.preSetRecords(tc.preExistingRecords)
+
+			// Grab all historical pool indexed twap records prior to initializing the
+			// geometric accumulator.
+			oldHistoricalPoolIndexed, err := k.GetAllHistoricalPoolIndexedTWAPs(ctx)
+			suite.Require().NoError(err)
+			// Initialize the geometric accumulator to get the expected values.
+			expectedHistoricalPoolIndexed := updateGeometricAccum(oldHistoricalPoolIndexed)
+
+			// Repeate for historical time indexed records.
```
mattverse, line 244:
```suggestion
			// Repeat for historical time indexed records.
```

x/twap/migrate.go
```diff
@@ -14,3 +32,46 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
...
+		// Sanity check order.
+		if i > 0 {
+			previousRecord := historicalTimeIndexed[i-1]
+
+			isInvalidOrder := previousRecord.Time.After(record.Time)
+			if isInvalidOrder {
+				return fmt.Errorf("error: historical twap records are not in ascending order, (%v), was after (%v)", previousRecord, record)
+			}
+		}
```
mattverse, line 67: Do we have proof that `getAllHistoricalTimeIndexedTWAPs` returns in correctly in ascending order? Looked into existing tests and could not find any relevent test

x/twap/migrate.go
```diff
@@ -14,3 +32,46 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
...
+
+// initializeGeometricTwapAcc initializes all existing twap records in state
+// with the given value. Returns nil on success, error otherwise.
+// Returns error if:
+// - fails to get historical time indexed records.
+// - does not find any time indexed records.
+// - the time indexed records returned from state are written back
+// in incorrect order. The order must be ascending by time so that
+// we update the "most recent" index with the right records correctly.
+func (k Keeper) initializeGeometricTwapAcc(ctx sdk.Context, value sdk.Dec) error {
```
mattverse, line 44: I think the current tests test this method indirectly by manually calling `updateGeometricAccum`, I think it would be better if we can test this method directly. Wdyt?

mattverse (commented):
Main logic looking good, left some questions and concerns about testing. Please take a look!

x/twap/migrate.go
```diff
@@ -1,9 +1,27 @@
...
+
+func NewMigrator(keeper Keeper) Migrator {
+	return Migrator{keeper: keeper}
+}
+
+// Migrate1To2GeometricTwapAcc migrates all existing twap records
+// by initializing their geometric twap accumulator to zero.
+// Return nil on success, error otherwise.
+func (m Migrator) Migrate1To2GeometricTwapAcc(ctx sdk.Context) error {
+	return m.keeper.initializeGeometricTwapAcc(ctx, sdk.ZeroDec())
```
nicolaslara, line 22: Shouldn't this be `sdk.OneDec()`? (the identity element of multiplication)

x/twap/migrate.go
```diff
@@ -14,3 +32,46 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
...
+		if i > 0 {
+			previousRecord := historicalTimeIndexed[i-1]
+
+			isInvalidOrder := previousRecord.Time.After(record.Time)
+			if isInvalidOrder {
+				return fmt.Errorf("error: historical twap records are not in ascending order, (%v), was after (%v)", previousRecord, record)
+			}
+		}
+
+		record.GeometricTwapAccumulator = value
```
nicolaslara, line 69: Can/should we check if `record.GeometricTwapAccumulator` already has a value and skip it if it does?

I'm thinking of a future minor version update from state export (like the issue we just had on 13.1.2) that may try to run this and would end up resetting all geometric accums

x/twap/migrate.go
p0mvn: No, it should be zero because we are computing the geometric time-weighted average of prices as an arithmetic time-weighted average of prices in log scale.

x/twap/migrate.go
p0mvn: This logic is meant to be run only once. I don't see why and how this would be re-run.

I would be supportive of adding this check. However, it is non-trivial for tests.

The reason is that we are already on v14.x code here. If the proto field exists, then it is initialized to zero.
We cannot set up the test in such a way where we have a v13.x (old) twap record so that its geometric twap accumulator is uninitialized.

x/twap/migrate_test.go
p0mvn: Well, each call to `threeAssetRecordWithPoolIdAndTime` returns three different []types.TwapRecord that must be merged into a single slice. What is the alternative? We could used `append` multiple times but it's not as clean

x/twap/twapmodule/module.go
```diff
@@ -79,6 +79,12 @@ type AppModule struct {
 
 func (am AppModule) RegisterServices(cfg module.Configurator) {
 	queryproto.RegisterQueryServer(cfg.QueryServer(), grpc.Querier{Q: twapclient.Querier{K: am.k}})
+
+	m := twap.NewMigrator(am.k)
+	err := cfg.RegisterMigration(types.ModuleName, 1, m.Migrate1To2GeometricTwapAcc)
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
p0mvn: What is the problem with doing the registration and increasing the consensus version?

I'm happy to change if there is a problem that I'm unaware of but, AFAIK, the approaches are identical

x/twap/migrate.go
```diff
@@ -14,3 +32,46 @@ func (k Keeper) MigrateExistingPools(ctx sdk.Context, latestPoolId uint64) error
...
+func (k Keeper) initializeGeometricTwapAcc(ctx sdk.Context, value sdk.Dec) error {
+	// In ascending order by time.
+	historicalTimeIndexed, err := k.getAllHistoricalTimeIndexedTWAPs(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(historicalTimeIndexed) == 0 {
+		return errors.New("error: no historical twap records found")
+	}
```
czarcas7ic, line 53: Is it appropriate for this to return an error? It seems like if no historical twap records are found, no logic needs to be run so it should just return nil

x/twap/types/keys.go
```diff
@@ -35,7 +35,7 @@ var (
 
 	// format is just pool id | denom1 | denom2
 	// made for getting most recent key
-	mostRecentTWAPsPrefix = mostRecentTWAPsNoSeparator + KeySeparator
+	MostRecentTWAPsPrefix = mostRecentTWAPsNoSeparator + KeySeparator
```
czarcas7ic, line 38: why do we export this now?

p0mvn:
Closing in favor of: https://github.com/osmosis-labs/osmosis/pull/3972

x/twap/migrate.go
nicolaslara: This makes sense. I think we should make sure that it doesn't get run again on an upgrade from export. I'm not sure if this is what would happen though. Just extrapolating from the issue with `initGenesis` on 13.1.2.

Issue #3958, opened 2023-01-09
test(CL Fees): test values for `GetFeeGrowthOutside`


mattverse:
## Background
Right now, testing expected values for `GetFeeGrowthOutside` is blocked by https://github.com/osmosis-labs/osmosis/issues/3956, as tick fee growth outside has not been properly initialized. 

We need to add test cases and checks for checking correct expected value once #3956 gets unblocked. 

## Acceptance Criteria

Test case for checks on correct expected value for `GetFeeGrowthOutside` has been properly added

PR #3979, opened 2023-01-11
test(CL Fees): test values for `GetFeeGrowthOutside`

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/3958

Adds and uncomments testing vectors for `GetFeeGrowthOutside`

### DIFF
--- a/x/concentrated-liquidity/fees_test.go
+++ b/x/concentrated-liquidity/fees_test.go
@@ -74,41 +74,44 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
 
 func (s *KeeperTestSuite) TestGetFeeGrowthOutside() {
 	type feeGrowthOutsideTest struct {
-		poolSetup           bool
-		tickSetup           bool
-		feeAccumulatorSetup bool
-		expectedError       bool
+		poolSetup                bool
+		tickSetup                bool
+		expectedFeeGrowthOutside sdk.DecCoins
+		invalidTick              bool
+		expectedError            bool
 	}
 
+	defaultAccumCoins := sdk.NewDecCoins(sdk.NewDecCoin("foo", sdk.NewInt(50)))
+	defaultPoolId := uint64(1)
+	defaultLiquidityForLowerTick := sdk.MustNewDecFromStr("10.0")
+	defaultLiquidityForUpperTick := sdk.MustNewDecFromStr("20.0")
+
+	defaultUpperTickIndex := int64(5)
+	defaultLowerTickIndex := int64(3)
+
 	tests := map[string]feeGrowthOutsideTest{
-		// TODO: uncomment this once tickInfo feeGrowthOutside logic has been implemented
-		"happy path": {
-			poolSetup:           true,
-			tickSetup:           true,
-			feeAccumulatorSetup: true,
-			expectedError:       false,
+		"fee growth outside with tick initialized": {
+			poolSetup:                true,
+			tickSetup:                true,
+			expectedFeeGrowthOutside: defaultAccumCoins,
+			expectedError:            false,
 		},
-		// "tick has not been initialized": {
-		// 	poolSetup:           true,
-		// 	tickSetup:           false,
-		// 	feeAccumulatorSetup: true,
-		// 	expectedError:       false,
-		// },
+		"tick has not been initialized": {
+			poolSetup:                true,
+			tickSetup:                false,
+			expectedFeeGrowthOutside: sdk.DecCoins(nil),
+			expectedError:            false,
+		},
 		"error: pool has not been setup": {
-			poolSetup:           false,
-			tickSetup:           false,
-			feeAccumulatorSetup: false,
-			expectedError:       true,
+			poolSetup:     false,
+			tickSetup:     false,
+			expectedError: true,
 		},
 	}
 
 	for name, tc := range tests {
 		s.Run(name, func() {
 			s.SetupTest()
-			defaultPoolId := uint64(1)
-			defaultLiquidityForTick := sdk.MustNewDecFromStr("10.0")
-			defaultUpperTickIndex := int64(5)
-			defaultLowerTickIndex := int64(3)
 
 			// if pool set up true, set up default pool
 			if tc.poolSetup {
@@ -117,12 +120,18 @@ func (s *KeeperTestSuite) TestGetFeeGrowthOutside() {
 
 			// if tick set up true, set upper and lower ticks to default values
 			if tc.tickSetup {
+				feeAccum, err := s.App.ConcentratedLiquidityKeeper.GetFeeAccumulator(s.Ctx, 1)
+				s.Require().NoError(err)
+
+				// manually update accumulator value to properly initialize tick's fee growth outside
+				feeAccum.UpdateAccumulator(defaultAccumCoins)
+
 				// first initialize upper tick
-				err := s.App.ConcentratedLiquidityKeeper.InitOrUpdateTick(
+				err = s.App.ConcentratedLiquidityKeeper.InitOrUpdateTick(
 					s.Ctx,
 					defaultPoolId,
 					defaultUpperTickIndex,
-					defaultLiquidityForTick,
+					defaultLiquidityForUpperTick,
 					true,
 				)
 				s.Require().NoError(err)
@@ -132,7 +141,7 @@ func (s *KeeperTestSuite) TestGetFeeGrowthOutside() {
 					s.Ctx,
 					defaultPoolId,
 					defaultLowerTickIndex,
-					defaultLiquidityForTick,
+					defaultLiquidityForLowerTick,
 					true,
 				)
 				s.Require().NoError(err)
@@ -146,7 +155,7 @@ func (s *KeeperTestSuite) TestGetFeeGrowthOutside() {
 				s.Require().NoError(err)
 
 				// check if returned fee growth outside has correct value
-				s.Require().Equal(feeGrowthOutside, sdk.DecCoins(nil))
+				s.Require().Equal(feeGrowthOutside, tc.expectedFeeGrowthOutside)
 			}
 		})
 	}
--- a/x/concentrated-liquidity/tick_test.go
+++ b/x/concentrated-liquidity/tick_test.go
@@ -248,6 +248,7 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
 
 			// Get the tick info for poolId 1
 			tickInfo, err := s.App.ConcentratedLiquidityKeeper.GetTickInfo(s.Ctx, 1, test.param.tickIndex)
+			s.Require().NoError(err)
 
 			// Ensure tick state contains any preexistingLiquidity (zero otherwise)
 			s.Require().Equal(preexistingLiquidity, tickInfo.LiquidityGross)


x/concentrated-liquidity/fees_test.go
```diff
@@ -74,41 +74,44 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
-		// TODO: uncomment this once tickInfo feeGrowthOutside logic has been implemented
-		"happy path": {
-			poolSetup:           true,
-			tickSetup:           true,
-			feeAccumulatorSetup: true,
-			expectedError:       false,
+		"fee growth outside with tick initialized": {
+			poolSetup:                true,
+			tickSetup:                true,
+			expectedFeeGrowthOutside: defaultAccumCoins,
```
p0mvn, line 96: We need to have more tests for various cases of what fee growth outside is.

For example, assume we have a position at ticks 10 and 20. As a swap is completed, we cross the ticks from left to right. Both ticks are crossed and the current tick stops at tick 25. As a result, both tick 10 and tick 20 are updated with the `feeGrowthGlobal`.

We need to simulate and test all the edge cases such as:
- crossing left to right
   *  the current price stays before the lower tick so none of the tick accumulators are updated
   *  the current price stays in-between the lower and upper tick so only lower tick is updated
   *  the current price stays after the upper tick so both lower and upper tick accumulators are updated
- crossing right to left
   * similar test cases to left to right but in the other direction.


I did a similar test for `collectFees` in https://github.com/osmosis-labs/osmosis/pull/3893
Test link:
https://github.com/osmosis-labs/osmosis/blob/e1e8a4624fdd844fdd8536de13ca8924a3c51de2/x/concentrated-liquidity/fees_test.go#L400

We need something similar but more focused on `feeGrowthOutside` specifically. I think it's important that we test all the edge cases at a lower level of abstraction. Otherwise, if we miss something, it will be close to impossible to debug at a higher level of abstraction (e.g. in swaps) so we will have to go back here.

mattverse:
@p0mvn  addressed comments, referenced the specific cases you linked to this case, but specifically focusing on testing the GetFeeGrowthOutside testings

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
-			feeAccumulatorSetup: true,
-			expectedError:       false,
-		},
-		// "tick has not been initialized": {
-		// 	poolSetup:           true,
-		// 	tickSetup:           false,
-		// 	feeAccumulatorSetup: true,
-		// 	expectedError:       false,
-		// },
+		"single swap left -> right: 2 ticks, one share - current price > upper tick": {
```
p0mvn, line 171: Could you please add an explanation comment similar to what you have for left-to-right tests cases?

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+			lowerTick:                 0,
+			upperTick:                 1,
+			currentTick:               2,
+			lowerTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(0))),
+			upperTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(10))),
+			globalFeeGrowth:           sdk.NewDecCoin(ETH, sdk.NewInt(10)),
+			expectedFeeGrowthOutside:  defaultAccumCoins,
+			expectedError:             false,
+		},
+		"single swap left -> right: 3 ticks, two shares - current price > upper tick": {
```
p0mvn, line 182:
```suggestion
		"single swap left -> right: 3 ticks, two shares - current tick > upper tick": {
```

x/concentrated-liquidity/fees_test.go
p0mvn, line 171:
```suggestion
		"single swap left -> right: 2 ticks, one share - current tick > upper tick": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+			lowerTick:                 0,
+			upperTick:                 2,
+			currentTick:               3,
+			lowerTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(0))),
+			upperTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(10))),
+			globalFeeGrowth:           sdk.NewDecCoin(ETH, sdk.NewInt(10)),
+			expectedFeeGrowthOutside:  defaultAccumCoins,
+			expectedError:             false,
+		},
+		"single swap left -> right: 2 ticks, one share - current price == upper tick": {
```
p0mvn, line 193:
```suggestion
		"single swap left -> right: 2 ticks, one share - current tick == lower tick": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+			poolSetup:                 true,
+			lowerTick:                 0,
+			upperTick:                 1,
+			currentTick:               0,
+			lowerTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(0))),
+			upperTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(10))),
+			globalFeeGrowth:           sdk.NewDecCoin(ETH, sdk.NewInt(10)),
+			expectedFeeGrowthOutside:  defaultAccumCoins,
+			expectedError:             false,
```
p0mvn, line 202: Suggest also having the following for left to right test cases:
- `current tick == upper tick`
- `current tick < lower tick`

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+		},
+		// imagine single swap over entire position
+		// crossing right > left and stopping at lower tick
+		// In this case, all fees must have been accrued inside the tick
+		// Since we track fees accrued below a tick, both upper and lower position
+		// ticks are zero
+		"single swap right -> left: 2 ticks, one share - current price == lower tick": {
```
p0mvn, line 209:
```suggestion
		"single swap right -> left: 2 ticks, one share - current tick == lower tick": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+			lowerTick:                 0,
+			upperTick:                 1,
+			currentTick:               0,
+			lowerTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(0))),
+			upperTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(10))),
+			globalFeeGrowth:           sdk.NewDecCoin(ETH, sdk.NewInt(10)),
+			expectedFeeGrowthOutside:  defaultAccumCoins,
+			expectedError:             false,
+		},
+		"single swap right -> left: 2 ticks, one share - current price < lower tick": {
```
p0mvn, line 220:
```suggestion
		"single swap right -> left: 2 ticks, one share - current tick < lower tick": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,111 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+			poolSetup:                 true,
+			lowerTick:                 0,
+			upperTick:                 1,
+			currentTick:               -1,
+			lowerTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(0))),
+			upperTickFeeGrowthOutside: sdk.NewDecCoins(sdk.NewDecCoin(ETH, sdk.NewInt(10))),
+			globalFeeGrowth:           sdk.NewDecCoin(ETH, sdk.NewInt(10)),
+			expectedFeeGrowthOutside:  defaultAccumCoins,
+			expectedError:             false,
+		},
```
p0mvn, line 230: suggest also having the following for right to left test cases:
- `current tick == upper tick`
- `current tick > upper tick`

p0mvn (commented):
Looks great. Requesting more test cases, lgtm otherwise

mattverse:
@p0mvn  I actually have a proposal regarding these test cases, the more I think about these cases and the test cases you mentioned(https://github.com/osmosis-labs/osmosis/pull/3979#discussion_r1067678790) as well as for the test cases`collectFees`, I feel like these test cases are closer to integration tests for specific edge cases and scenarios, more than unit testing. I would like to propose reverting these test cases back to `https://github.com/osmosis-labs/osmosis/pull/3979/commits/d6a2b4f1910bb42ed680ed574a635125513d678c` and then creating new test cases that mentions this and tests both `collectFees` and `GetFeeGrowthOutside`. What do you think?

p0mvn:
> @p0mvn I actually have a proposal regarding these test cases, the more I think about these cases and the test cases you mentioned([#3979 (comment)](https://github.com/osmosis-labs/osmosis/pull/3979#discussion_r1067678790)) as well as for the test cases`collectFees`, I feel like these test cases are closer to integration tests for specific edge cases and scenarios, more than unit testing. I would like to propose reverting these test cases back to [d6a2b4f](https://github.com/osmosis-labs/osmosis/commit/d6a2b4f1910bb42ed680ed574a635125513d678c) and then creating new test cases that includes this and tests both `collectFees` and `GetFeeGrowthOutside` with the scenarios you provided. What do you think?

I would still consider these to be unit tests. The unit test names are framed in terms of swaps because it is easier for us humans to reason about them. That's why it might feel like these are more functional in nature.

I don't really understand what exactly you're proposing with reverting and re-recreating test cases. I still think that we should have the requested test cases to validate that the internal calculations are done correctly at each level of abstraction (`collectFees`, `getFeeGrowthOutside` etc). As long we are tracking/planning to add them in the future, any approach is fine with me.

x/concentrated-liquidity/fees_test.go
mattverse: done!

mattverse:
> > @p0mvn I actually have a proposal regarding these test cases, the more I think about these cases and the test cases you mentioned([#3979 (comment)](https://github.com/osmosis-labs/osmosis/pull/3979#discussion_r1067678790)) as well as for the test cases`collectFees`, I feel like these test cases are closer to integration tests for specific edge cases and scenarios, more than unit testing. I would like to propose reverting these test cases back to [d6a2b4f](https://github.com/osmosis-labs/osmosis/commit/d6a2b4f1910bb42ed680ed574a635125513d678c) and then creating new test cases that includes this and tests both `collectFees` and `GetFeeGrowthOutside` with the scenarios you provided. What do you think?
> 
> I would still consider these to be unit tests. The unit test names are framed in terms of swaps because it is easier for us humans to reason about them. That's why it might feel like these are more functional in nature.
> 
> I don't really understand what exactly you're proposing with reverting and re-recreating test cases. I still think that we should have the requested test cases to validate that the internal calculations are done correctly at each level of abstraction (`collectFees`, `getFeeGrowthOutside` etc). As long we are tracking/planning to add them in the future, any approach is fine with me.

Hey @p0mvn! I don't have strong personal opinions regarding this case as well, I'm fine with either both ways. I think it's a  difference in point of view in which way is intuitive.

Decision made to stay as is, addressed your comments as well!

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,159 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
+		"single swap right -> left: 2 ticks, one share - current tick == upper tick": {
+			poolSetup:                 true,
+			lowerTick:                 0,
+			upperTick:                 1,
+			currentTick:               0,
```
p0mvn, line 250: It seems to me that here currentTick == lowerTick, not current tick == upper tick like the test case name says

x/concentrated-liquidity/fees_test.go
p0mvn: I think that `current tick > upper tick` is still missing

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,159 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
-		// 	feeAccumulatorSetup: true,
-		// 	expectedError:       false,
-		// },
+		// imagine single swap over entire position
+		// crossing left > right and stopping above upper tick
+		// In this case, only the upper tick accumulator must have
+		// been updated when crossed.
+		// Since we track fees accrued below a tick, upper tick is updated
+		// while lower tick is zero.
+		"single swap left -> right: 2 ticks, one share - current price > upper tick": {
```
p0mvn, line 175:
```suggestion
		"single swap left -> right: 2 ticks, one share - current tick > upper tick": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -144,67 +144,159 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
-					defaultLiquidityForTick,
-					true,
-				)
+				pool = s.PrepareConcentratedPool()
+
+				s.initializeTick(s.Ctx, tc.lowerTick, defaultInitialLiquidity, tc.lowerTickFeeGrowthOutside, false)
+				s.initializeTick(s.Ctx, tc.upperTick, defaultInitialLiquidity, tc.upperTickFeeGrowthOutside, true)
+				pool.SetCurrentTick(sdk.NewInt(tc.currentTick))
+				s.App.ConcentratedLiquidityKeeper.SetPool(s.Ctx, pool)
+				// poolId = pool.GetId()
```
AlpinYukseloglu, line 298:
```suggestion
```

AlpinYukseloglu (approved):
Would love to fuzz test fees across tick ranges and fee growths, but that might be better suited for a separate PR. This one LGTM!

x/concentrated-liquidity/fees_test.go
mattverse: good catch, fixed

x/concentrated-liquidity/fees_test.go
mattverse: added "single swap right -> left: 2 ticks, one share - current tick > upper tick":

czarcas7ic (approved):

Issue #3980, opened 2023-01-11
test(CL Fess): updateFeeAccumulatorPosition


p0mvn:
## Background

We need to test `updateFeeAccumulatorPosition`

## Suggested Design

Assume a base position with pool id, owner, lower tick, and upper tick

Some test vectors that should be considered relative to base:
- updating a position with a different pool id implies updating a different position
- updating a position with a different owner implies updating a different position
- updating a position with a  different lowe tick implies updating a different position
- updating a position with an upper tick implies updating a different position

Basically, we want to test that prefixing is working correctly for positions. We have a related test here:
https://github.com/osmosis-labs/osmosis/blob/a3f80b7b14ad43ba528878833d9d151d15c2a19f/x/concentrated-liquidity/fees_test.go#L103-L120

Also, test that positions accumulator and liquidity are updated correctly.

## Acceptance Criteria
- tests added
- suggested test vectors covered
- improve godoc

RusAkh:
hello @p0mvn is this free to pick up?

p0mvn:
Hi @RusAkh . Thanks for checking in. @mattverse and I are working on fees and have the context. For the speed of iteration, it's best if one of us drives this to completion

PR #3981, opened 2023-01-11
refactor(CL Fees): prefix positions by pool id, owner and ticks

p0mvn:
Closes: #XXX

Currently, our CL fees positions are prefixed by the owner. However, they should be prefixed by pool id, owner, lower and upper tick.

This PR changes the logic to apply the appropriate position prefixing.

Additionally, we should not allow re-initializing a position with non-zero liquidity to avoid accidentally overriding it. However, we do want to override a zero-liquidity position because a user might want to provide liquidity at the ticks where their old position used to be.

## Brief Changelog
- changed position prefixing
- added `HasPosition` accumulator method
- tested `HasPosition`
- refactored `initializeFeeAccumulatorPosition`
- refactor `initializeFeeAccumulatorPosition` test to account for the new prefix and behavior

## Testing and Verifying

This change added tests

### DIFF
--- a/x/concentrated-liquidity/fees.go
+++ b/x/concentrated-liquidity/fees.go
@@ -218,6 +218,7 @@ func calculateFeeGrowth(targetTick int64, feeGrowthOutside sdk.DecCoins, current
 
 // formatPositionAccumulatorKey formats the position accumulator key prefixed by pool id, owner, lower tick
 // and upper tick with a key separator in-between.
+// nolint: unused
 func formatPositionAccumulatorKey(poolId uint64, owner sdk.AccAddress, lowerTick, upperTick int64) string {
 	return strings.Join([]string{strconv.FormatUint(poolId, uintBase), owner.String(), strconv.FormatInt(lowerTick, uintBase), strconv.FormatInt(upperTick, uintBase)}, keySeparator)
 }
--- a/x/concentrated-liquidity/fees.go
+++ b/x/concentrated-liquidity/fees.go
@@ -1,6 +1,7 @@
 package concentrated_liquidity
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -10,9 +11,9 @@ import (
 )
 
 const (
-	feeAccumPrefix        = "fee"
-	feeAccumNameSeparator = "/"
-	uintBase              = 10
+	feeAccumPrefix = "fee"
+	keySeparator   = "/"
+	uintBase       = 10
 )
 
 var (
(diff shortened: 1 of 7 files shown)


x/concentrated-liquidity/fees.go
```diff
@@ -75,6 +98,7 @@ func (k Keeper) initializeFeeAccumulatorPosition(ctx sdk.Context, poolId uint64,
 
 // updateFeeAccumulatorPosition updates the owner's position
 // nolint: unused
+// TODO: test
```
p0mvn, line 101: https://github.com/osmosis-labs/osmosis/issues/3980

x/concentrated-liquidity/fees_test.go
```diff
@@ -10,61 +10,134 @@ import (
 	"github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types"
 )
 
+// fields used to identify a fee position.
+type positionIdentifiers struct {
```
p0mvn, line 14: note: this might be useful for: https://github.com/osmosis-labs/osmosis/issues/3980

x/concentrated-liquidity/fees.go
```diff
@@ -57,16 +58,38 @@ func (k Keeper) chargeFee(ctx sdk.Context, poolId uint64, feeUpdate sdk.DecCoin)
...
 	}
 
+	positionKey := formatPositionAccumulatorKey(poolId, owner, lowerTick, upperTick)
+
+	hasPosition, err := feeAccumulator.HasPosition(positionKey)
+	if err != nil {
+		return err
+	}
+
+	if hasPosition {
```
mattverse, line 80:
```suggestion
        // assure that existing position has zero liquidity
	if hasPosition {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -10,61 +10,134 @@ import (
...
-	tests := map[string]initFeeAccumTest{
-		"existing accumulator, new position": {
-			setPoolAccumulator:  true,
-			setExistingPosition: false,
-			expectedPass:        true,
+	tests := []initFeeAccumTest{
+		{
+			name:           "new zero liqudidity position",
+			liquidityDelta: sdk.ZeroDec(),
+			// Gets overwritten by the next test case
```
mattverse, line 74: nit: what do you think of making all test cases independent of each other and then resetting the testing environment every iteration of a test? IMO making an independent test case / environment for each table tests makes a safer / clearer environment to test

mattverse (approved):
LGTM 🌮

x/concentrated-liquidity/fees_test.go
p0mvn: The thing is, here specifically it is important to test the relationship between test cases. I've explored your suggestion and found it hard to implement without much code duplication. As a result, I concluded that it is better to have dependent test cases here.

For example, we must make sure that if we already created a position and it had liquidity, that position would not get overwritten. We also need to make sure that positions with different owners or ticks are completely distinct positions. By having the test cases share context, we can easily test this

mattverse: yeah makes sense, I also thought that we could be having the same pain point as this test somewhat has the property of story-driven tests, lets leave this as is !

PR #3874, opened 2022-12-28
feat: geometric tick spacing with additive ranges

czarcas7ic:
# Geometric Tick Spacing with Additive Ranges ADR

## Context

In Uniswap V3, discrete points (called ticks) are used when providing liquidity in a concentrated liquidity pool. The price [p] corresponding to a tick [t] is defined by the equation:

$$ p(i) = 1.0001^t $$

This results in a .01% difference between adjacent tick prices. However, this does not allow for control over the specific prices that the ticks correspond to. For example, if a user wants to make a limit order at the $17,100.50 price point, they would have to interact with either tick 97473 (corresponding to price $17,099.60) or tick 97474 (price $17101.30).

Since we know what range a pair will generally trade in, how do we go about providing more granularity at that range and provide a more optimal price range between ticks instead of the "one-size-fits-all" approach explained above?

## Decision

In Osmosis' implementation of concentrated liquidity, we will instead make use of geometric tick spacing with additive ranges. 

We start by defining a precision factor at a spot price of one ($k_{p1}$).

For instance, if $k_{p1} = -4$ , then each tick starting at 1 and ending at the first factor of 10 will represents a spot price increase of 0.0001. At this precision factor:
* $tick_0 = 1$ (tick 0 is always equal to 1 regardless of precision factor)
* $tick_1 = 1.0001$
* $tick_2 = 1.0002$
* $tick_3 = 1.0003$

This continues on until we reach a spot price of 10. At this point, since we have increased by a factor of 10, our $k_{current}$ increases from -4 to -3, and the ticks will increase as follows:
* $tick_{90000} = 10$
* $tick_{90001} = 10.001$
* $tick_{90002} = 10.002$

For spot prices less than a dollar, the precision factor decreases at every factor of 10. For example, with a $k_{p1}$ of -4:
* $tick_{-1} = 0.9999$
* $tick_{-2} = 0.9998$
* $tick_{-5001} = 0.4999$
* $tick_{-5002} = 0.4998$

With a $k_{p1}$ of -6:
* $tick_{-1} = 0.999999$
* $tick_{-2} = 0.999998$
* $tick_{-5001} = 0.994999$
* $tick_{-5002} = 0.994998$

## Formulas

After we define $k_{p1}$ (this is chosen by the user based on what precision they desire the asset pair to trade at), we can then calculate how many ticks must be crossed in order for k to be incremented ($kIncrementDistance$):

$$ kIncrementDistance = 9 * 10^{(-k_{p1})} $$

[...]

PR #3982, opened 2023-01-11
fix(CL): flaky TestChargeFee

p0mvn:
Closes: #XXX

The problem stems from the test relying on test case order while using maps. Map iterations are non-deterministic.

Fixed by switching to slice instead.


czarcas7ic:
It seems that tag might have broke the e2e test

p0mvn:
> It seems that tag might have broke the e2e test

Fixed now, updated wrong branch 😞

p0mvn:
This got in in #3874

PR #4010, opened 2023-01-13
Improved docs for crosschain swaps

nicolaslara:

Better docs for crosschain swaps

## Brief Changelog

 - Added testing details and a better spec in the readme
 - Updated the schemas

## Testing and Verifying
all tests pass

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (/ no)
  - How is the feature or change documented? (not applicable

### DIFF
--- a/x/ibc-hooks/contracts/crosschain-swaps/README.md
+++ b/x/ibc-hooks/contracts/crosschain-swaps/README.md
@@ -12,20 +12,19 @@ and forwarding to a different chain.
 To instantiate the contract, you need to specify the following parameters:
 
  * swap_contract: the swaprouter contract to be used
- * track_ibc_sends: true|false. Specifies if the contract should track the sent ibc packets for recovery. This should be false in v1
  * channels: a list of (bech32 prefix, channel_id) that the contract will allow. 
 
 ### Example instantiation message
 
 ``` json
-{"swap_contract": "osmo1thiscontract", "track_ibc_sends": false, "channels": [["cosmos", "channel-0"], ["juno", "channel-42"]]}
+{"swap_contract": "osmo1thiscontract", "channels": [["cosmos", "channel-0"], ["juno", "channel-42"]]}
 ```
 
 ## Usage
 
 ### Via IBC
 
-Assuming the current implementation of the wasm middleware on Osmosis v14 (`x/ibc-hooks/v0.0.5`), the memo
+Assuming the current implementation of the wasm middleware on Osmosis v14 (`x/ibc-hooks/v0.0.6`), the memo
 of an IBC transfer to do crosschain swaps would look as follows:
 
 ``` json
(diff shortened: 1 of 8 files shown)


x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -35,9 +34,9 @@ of an IBC transfer to do crosschain swaps would look as follows:
         "osmosis_swap": {
             "input_coin": {"denom":"token0","amount":"1000"}, 
             "output_denom":"token1",
-            "slippage":{"max_slippage_percentage":"5"},
+            "slippage":{"twap": {"slippage_percentage":"20", "window_seconds": 10}},
             "receiver":"juno1receiver",
-            "failed_delivery":null
+            "on_failed_delivery": "do_nothing",
```
ValarDragon, line 39: Presumably we also accept the key being missing, right?

x/ibc-hooks/contracts/crosschain-swaps/README.md
```diff
@@ -74,152 +79,240 @@ To use this contract for crosschain swaps, the following are needed:
...
-* Tokens on the non-osmosis testnet
-* A pool on the osmosis testnet between the IBC'd native asset of the
-  non-osmosis testnet and osmo
+Create an alias for chainA and chainB commands that will be used throughout the guide:
 
-### Localrelayer
+```bash
+alias chainA="osmosisd --node http://localhost:26657 --chain-id localosmosis-a"
+alias chainB="osmosisd --node http://localhost:36657 --chain-id localosmosis-b"
+```
```
ValarDragon, line 108: Wow good job on the docs, though def a bit painful we have to list this out all here.

Would be great to get into a standard localosmosis setup, or just directing people to a commented script that we actually use. (Rather than Readme + script)

ValarDragon (approved):

x/ibc-hooks/contracts/crosschain-swaps/README.md
nicolaslara: Not anymore. I can make it an option, but thought it was better to be explicit. Happy to change it either way

x/ibc-hooks/contracts/crosschain-swaps/README.md
nicolaslara: The script mostly works on its own, but there is some weirdness that make pool creation fail the first time and you have to kill it at that point and run it again. I didn't want to spend a lot of time debugging something that will run so seldom, so I just went with docs.

One downside of having a script is that people just trust it instead of going over the commands. If there's something wrong with the script it may not be caught

nicolaslara: oh, and... most people don't need to do this once the contracts are deployed (pools already exist)

x/ibc-hooks/contracts/crosschain-swaps/README.md
ValarDragon: Ah ok, that makes sense. Either is fine w/ me as well

PR #4035, opened 2023-01-16
Checks should run if the wasm changes

nicolaslara:

Checks should run if the wasm files change. This is good for safety, and also allows us to trigger them by modifying those files.

## Brief Changelog

CI checks now run when wasm files change

## Testing and Verifying
all tests pass


.github/workflows/contracts.yml
```diff
@@ -28,26 +28,10 @@ jobs:
       - uses: technote-space/get-diff-action@v6.0.1
         with:
           PATTERNS: |
+            **/bytecode/*.wasm
```
ValarDragon, line 31:
```suggestion
            **/**/*.wasm
```

ValarDragon (approved):

ValarDragon:
Err you need to bring back the wasm files =p

Issue #4028, opened 2023-01-16
[CL][CL Module]: CLI for `collectFees`


p0mvn:
## Background

We need to add CLI for collecting fees.

## Suggested Design

Utilize `osmocli` to implement:
https://github.com/osmosis-labs/osmosis/blob/0518bf608f512b5a7c953a8c910e16613e09821c/x/concentrated-liquidity/types/tx.pb.go#L267-L268

Should be similar to other messages within the module:
https://github.com/osmosis-labs/osmosis/blob/0518bf608f512b5a7c953a8c910e16613e09821c/x/concentrated-liquidity/client/cli/tx.go#L43-L51

## Acceptance Criteria
- CLI for collecting fees is implemented
- Basic manual test that it executes / runs, no need for more detailed tests

PR #4036, opened 2023-01-16
`collectFees` cli

hieuvubk:
Closes: #4028 

Refer from #4028 , I added cli for `collectFees` func

## Testing and Verifying

CLI can execute in my local 

### DIFF
--- a/x/concentrated-liquidity/client/cli/tx.go
+++ b/x/concentrated-liquidity/client/cli/tx.go
@@ -53,10 +53,10 @@ func NewWithdrawPositionCmd() (*osmocli.TxCliDesc, *types.MsgWithdrawPosition) {
 
 func NewCollectFeesCmd() (*osmocli.TxCliDesc, *types.MsgCollectFees) {
 	return &osmocli.TxCliDesc{
-		Use: "collect-fees [lower-tick] [upper-tick]",
-		Short: "collect fees from a liquidity position",
-		Example: "collect-fees [-69082] 69082 --pool-id 1 --from val --chain-id osmosis-1",
+		Use:                 "collect-fees [lower-tick] [upper-tick]",
+		Short:               "collect fees from a liquidity position",
+		Example:             "collect-fees [-69082] 69082 --pool-id 1 --from val --chain-id osmosis-1",
 		CustomFlagOverrides: poolIdFlagOverride,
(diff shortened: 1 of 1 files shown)


p0mvn (approved):
Amazing. Thank you

x/concentrated-liquidity/client/cli/tx.go
```diff
@@ -49,3 +50,13 @@ func NewWithdrawPositionCmd() (*osmocli.TxCliDesc, *types.MsgWithdrawPosition) {
 		Flags:               osmocli.FlagDesc{RequiredFlags: []*flag.FlagSet{FlagSetJustPoolId()}},
 	}, &types.MsgWithdrawPosition{}
 }
+
+func NewCollectFeesCmd() (*osmocli.TxCliDesc, *types.MsgCollectFees) {
+	return &osmocli.TxCliDesc{
+		Use:                 "collect-fees [lower-tick] [upper-tick]",
+		Short:               "collect fees from a liquidity position",
+		Example:             "collect-fees [-69082] 69082 --pool-id 1 --from val --chain-id osmosis-1",
```
stackman27, line 58: not sure if [] are necessary. I could be wrong
```suggestion
		Example:             "collect-fees -69082 69082 --pool-id 1 --from val --chain-id osmosis-1",
```

ValarDragon: It is for some dumb negative thing unfortunately.

Though I think having to list lower tick and upper tick is some not-great UX, and we will want to make a better CLI ux flow later

mattverse: Oh wow did not know that this was a thing that we need to do in order to use negatives in cli (and I think this is something important to have fixed, would folks mind if we create an issue for this to tackle whenever we have free capacity?)

mattverse (approved):
LGTM 🌮

x/concentrated-liquidity/client/cli/tx.go
RusAkh: I actually had the same problem and created and [issue](https://github.com/osmosis-labs/osmosis/issues/4041)

On my side, using [negative value] does not work when I run cli commands, maybe it is because I am on osx, not sure. I proposed some solutions in the issue, would be great to have some feedback!

PR #4051, opened 2023-01-18
feat[CL]: migrate GAMM shares to CL full range position

czarcas7ic:
Closes: #XXX

## Summary

The following PR implements the bare logic of migrating **unlocked** gamm shares from a balancer pool to a cl pool. The only logic missing from this which will be extracted in another PR is canonical links between a single balancer pool to a single cl pool. Right now, this simply takes in the destination cl pool ID. I forsee these canonical links being a gamm module parameter.

### Open questions:
* ~~How do we properly round up in the `CalcAmount0/1Delta` (`Ceil` was not acting as intended)? Is the newly introduced logic safe?~~
* ~~I had to replicate `GetMinAndMaxTicksFromExponentAtPriceOne` in both the internal cl package and in the cl home. What is a cleaner way to utilize this across modules?~~
* ~~Do we agree with the logic of replacing the `minTick` with a price of 0 instead of `.000000000001`?~~
  * ~~This seems like an easy "yes", but I think replacing the maxTick's price from the current max spot price of `100000000000000000000000000000000000000` to whatever the max BigDec is seems more questionable.~~

## Brief Changelog
 
  - Initialized `clKeeper` before `gammKeeper` due to new gamm dependency on `clKeeper`
  - Adds `MsgMigrateSharesToFullRangeConcentratedPosition` tx within the balancer proto
  - Changes the following cl functions from private to public methods due to gamm dependency
    - `CreatePosition`
    - `WithdrawPosition`
    - `GetPoolById`
  -  Changes `minTick's` price to return 0 instead of .000000000001
  - Adds `MigrateShares` msgServer message and respective tests


app/apptesting/gamm.go
```diff
@@ -16,6 +16,8 @@ var DefaultAcctFunds sdk.Coins = sdk.NewCoins(
 	sdk.NewCoin("foo", sdk.NewInt(10000000)),
 	sdk.NewCoin("bar", sdk.NewInt(10000000)),
 	sdk.NewCoin("baz", sdk.NewInt(10000000)),
+	sdk.NewCoin("eth", sdk.NewInt(10000000)),
```
mattverse, line 19: Nit: consider using ETH and USDC to consolidate with other testing denoms

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -63,8 +71,9 @@ func CalcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	// additionally, without rounding, there exists cases where the swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) for loop within
 	// the CalcOut/In functions never actually reach zero due to dust that would have never gotten counted towards the amount (numbers after the 10^6 place)
 	if roundUp {
-		return liq.Mul(diff).Quo(denom).Ceil()
+		return liq.Mul(diff).Quo(denom).RoundInt().ToDec()
```
mattverse, line 74: What was the reasoning behind ensuring safeness on this change?) I assumed ceiling would be safer then rounding)

x/concentrated-liquidity/internal/math/tick.go
```diff
@@ -50,14 +50,18 @@ func TickToSqrtPrice(tickIndex, exponentAtPriceOne sdk.Int) (price sdk.Dec, err
...
-	minTick, maxTick := GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne)
+	minTick, maxTick := getMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne)
 	if tickIndex.LT(sdk.NewInt(minTick)) {
 		return sdk.Dec{}, types.TickIndexMinimumError{MinTick: minTick}
 	}
 	if tickIndex.GT(sdk.NewInt(maxTick)) {
 		return sdk.Dec{}, types.TickIndexMaximumError{MaxTick: maxTick}
 	}
 
+	if tickIndex.Equal(sdk.NewInt(minTick)) {
```
mattverse, line 61: Does this mean that price is always zero at min tick?

x/concentrated-liquidity/internal/math/tick.go
```diff
@@ -180,9 +185,9 @@ func powTenBigDec(exponent sdk.Int) osmomath.BigDec {
 	return osmomath.OneDec().Quo(osmomath.NewBigDec(10).Power(osmomath.NewBigDec(exponent.Abs().Int64())))
 }
 
-// GetMinAndMaxTicksFromExponentAtPriceOne determines min and max ticks allowed for a given exponentAtPriceOne value
+// getMinAndMaxTicksFromExponentAtPriceOne determines min and max ticks allowed for a given exponentAtPriceOne value
```
mattverse, line 188: Isnt it safe to expose these get functions publically?

x/concentrated-liquidity/internal/math/tick.go
mattverse: Also noticed that we changed from using minsqrt price -> zero, what is the context for this?

x/concentrated-liquidity/internal/math/tick.go
```diff
@@ -50,14 +50,18 @@ func TickToSqrtPrice(tickIndex, exponentAtPriceOne sdk.Int) (price sdk.Dec, err
...
 		return sdk.Dec{}, types.TickIndexMinimumError{MinTick: minTick}
 	}
 	if tickIndex.GT(sdk.NewInt(maxTick)) {
 		return sdk.Dec{}, types.TickIndexMaximumError{MaxTick: maxTick}
 	}
 
+	if tickIndex.Equal(sdk.NewInt(minTick)) {
+		return sdk.ZeroDec(), nil
+	}
+
```
mattverse, line 64: We also need to add check for max tick here if thats the case(i suppose)

x/concentrated-liquidity/tick.go
```diff
@@ -9,11 +9,18 @@ import (
 	types "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types"
 )
 
+var (
+	sdkNineDec        = sdk.NewDec(9)
```
mattverse, line 13: These should be upstreamed😛

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +362,211 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+		fmt.Printf("userEthBalanceTransferredToClPool.Amount %v \n", userEthBalanceTransferredToClPool.Amount)
+		fmt.Printf("clPoolEthBalanceAfterMigration.Amount %v \n", clPoolEthBalanceAfterMigration.Amount)
+		fmt.Printf("userUsdcBalanceTransferredToClPool.Amount %v \n", userUsdcBalanceTransferredToClPool.Amount)
+		fmt.Printf("clPoolUsdcBalanceAfterMigration.Amount %v \n", clPoolUsdcBalanceAfterMigration.Amount)
+		fmt.Printf("poolUsdcBalanceInTolerance %v \n", poolUsdcBalanceInTolerance)
+		suite.Require().True(poolEthBalanceInTolerance)
+		suite.Require().True(poolUsdcBalanceInTolerance)
+		// suite.Require().Equal(userEthBalanceTransferredToClPool.String(), clPoolEthBalanceAfterMigration.String())
+		// suite.Require().Equal(userUsdcBalanceTransferredToClPool.String(), clPoolUsdcBalanceAfterMigration.String())
+
```
mattverse, line 535: Can we add checks for user's postion and balance before withdrawing?

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +362,211 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+		// Withdraw the user's concentrated liquidity position
+		minTick, maxTick := cl.GetMinAndMaxTicksFromExponentAtPriceOne(clPool.GetPrecisionFactorAtPriceOne())
```
mattverse, line 537: IMO doing all these checks without withdrawing feels like better unit testing process, as we're adding another secondary logic gate for the checks, instead of uniquely testing the SUT. It mught be because of a reason I do not know of or just a difference in perspective so feel free to ignore if you have different view points

x/gamm/pool-models/balancer/msgs.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	TypeMsgCreateBalancerPool = "create_balancer_pool"
+	TypeMsgMigrateShares      = "migrate_shares"
```
mattverse, line 13: Also a reminder to add this to codec

mattverse (commented):
Looks good! Left some comments on the main logics and testings

app/keepers/keepers.go
```diff
@@ -265,12 +265,19 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
...
+		appKeepers.GetSubspace(concentratedliquiditytypes.ModuleName),
+	)
+
 	gammKeeper := gammkeeper.NewKeeper(
 		appCodec, appKeepers.keys[gammtypes.StoreKey],
 		appKeepers.GetSubspace(gammtypes.ModuleName),
 		appKeepers.AccountKeeper,
 		// TODO: Add a mintcoins restriction
-		appKeepers.BankKeeper, appKeepers.DistrKeeper)
+		appKeepers.BankKeeper, appKeepers.DistrKeeper, appKeepers.ConcentratedLiquidityKeeper)
```
czarcas7ic, line 280: Note to reviewer: We now initialize the cl keeper before gamm since gamm calls on cl for share migration

proto/osmosis/gamm/pool-models/balancer/tx/tx.proto
```diff
@@ -29,3 +33,34 @@ message MsgCreateBalancerPool {
...
+message MsgMigrateSharesToFullRangeConcentratedPosition {
+  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
+
+  cosmos.base.v1beta1.Coin shares_to_migrate = 2 [
+    (gogoproto.moretags) = "yaml:\"shares_to_migrate\"",
+    (gogoproto.nullable) = false
+  ];
+  // temporary field, eventually gamm pool should be linked to cl pool
+  uint64 pool_id_entering = 3;
+}
```
czarcas7ic, line 47: Note to reviewer: Thinking of having some gamm module parameter that stores the canonical cl pool link to its respective gamm pool. At time of upgrade we can set some of the major pools, but other pools will need to be added via governance prop.

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: I was a little confused by how Ceil works. For a simplistic example, I would have a number like

10.999999

And using Ceil would return 10, which seems wrong.

x/concentrated-liquidity/internal/math/math.go
```diff
@@ -85,8 +99,9 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 	// additionally, without rounding, there exists cases where the swapState.amountSpecifiedRemaining.GT(sdk.ZeroDec()) for loop within
 	// the CalcOut/In functions never actually reach zero due to dust that would have never gotten counted towards the amount (numbers after the 10^6 place)
 	if roundUp {
-		return liq.Mul(diff).Ceil()
+		return liq.Mul(diff).RoundInt().ToDec()
 	}
```
czarcas7ic, line 89: Note to reviewer: Come to a conclusion on what the proper rounding strategy is (ceil does not act as I would expect it to)

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: This is still up for debate on how we want to treat this, but the logic is as follows:

Whatever the min tick is should be as close to 0 as possible and whatever the max tick is should be as close to inf as possible.

The problem we run into here with the minTick being .000000000001 is the sqrt of this is a larger number (.000001) and results in the user getting less shares than they should when transferring from a 0 to inf balancer position. In reality, the min tick should be 0 and the sqrt of 0 is 0. So this tries to replicate this logic.

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: I need help with a better way to implement this but the reason for this is as follows:

Other modules need to utilize this function, and having this in internal prevents this even if it is public. So what I did to circumnavigate this is make it private in internal, create a clone of it in the home of the cl module but make that public and utilize that public function in other modules. 

Any recommendation for a cleaner implementation would be appreciated.

x/concentrated-liquidity/tick.go
```diff
@@ -127,3 +134,21 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
...
+}
+
+// getMinAndMaxTicksFromExponentAtPriceOne determines min and max ticks allowed for a given exponentAtPriceOne value
+// This allows for a min spot price of 0.000000000000000001 and a max spot price of 100000000000000000000000000000000000000 for every exponentAtPriceOne value
+func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTick, maxTick int64) {
+	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(powTen(exponentAtPriceOne.Neg()))
+	minTick = sdkEighteenDec.Mul(geometricExponentIncrementDistanceInTicks).Neg().RoundInt64()
+	maxTick = sdkThirtyEightDec.Mul(geometricExponentIncrementDistanceInTicks).TruncateInt64()
+	return minTick, maxTick
+}
```
czarcas7ic: This is copied from the internal folder but made public for other modules to use. If there is a better way to do this please let me know.

x/gamm/pool-models/balancer/msgs.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4051/commits/406dae82d237eee2481d367e925af6e1c038ef0a

Thanks!

app/apptesting/gamm.go
czarcas7ic: Removed, was there for early testing

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: One of the current open question:

>* Do we agree with the logic of replacing the `minTick` with a price of 0 instead of `.000000000001`?
    * This seems like an easy "yes", but I think replacing the maxTick's price from the current max spot price of `100000000000000000000000000000000000000` to whatever the max BigDec is seems more questionable.

x/gamm/keeper/msg_server_test.go
czarcas7ic: I would agree with this, but I don't have a simple way right now of checking what cl liquidity corresponds to in terms of they underlying. Because of this, I need to withdraw to determine how much of the underlying the user is allotted.

czarcas7ic: After thinking about this for a little, I agree this test overstretched for no good reason. I will reduce it to what you suggested, thank you!

czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4051/commits/de1590c38fb8c5f801f5528677464682a79840e3

x/gamm/keeper/msg_server_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4051/commits/de1590c38fb8c5f801f5528677464682a79840e3

x/concentrated-liquidity/internal/math/math.go
AlpinYukseloglu: Do you have a branch with this test by any chance? `Ceil` seems to be functioning as intended on my end, and much of our rounding guardrails in GAMM rely on it.

In any case, `RoundInt` is not equivalent to rounding up, as it rounds down if the decimal is < .5

czarcas7ic: Since this is orthogonal to the PR now (am no longer using the round up functionality) I will revert this and make an issue to investigate

proto/osmosis/gamm/pool-models/balancer/tx/tx.proto
mattverse: Acknowledged for now!

x/concentrated-liquidity/internal/math/math.go
mattverse: Thanks! Please revert this before merging

x/concentrated-liquidity/internal/math/math.go
mattverse: Ditto here! Would it be possible to make an issue between the tradeoffs? Thanks!

x/concentrated-liquidity/internal/math/tick.go
mattverse: I see, thanks for the explanation. I think it would be nice to have more intense tests that especially touches this scenario, since it would result in better safeness imo! Doesnt have to be in this specific PR but can be a separate PR

mattverse: Would it also be possible to add in line comments from the context you have mentioned?

x/concentrated-liquidity/internal/math/tick.go
mattverse: > Other modules need to utilize this function, and having this in internal prevents this even if it is public.

Are we getting circular reference for this?

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Reverted https://github.com/osmosis-labs/osmosis/pull/4051/commits/6958a3d47f21635d1aa05858757f7ab18e3ae9b9

x/concentrated-liquidity/internal/math/math.go
czarcas7ic: Reverted https://github.com/osmosis-labs/osmosis/pull/4051/commits/6958a3d47f21635d1aa05858757f7ab18e3ae9b9

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: added in line comment explaining this https://github.com/osmosis-labs/osmosis/pull/4051/commits/ae60111c0c026afa84f24ddbaf66e591ef8cfe9d

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: Yes, normally this wouldn't be an issue but because it has to be used in the internal package, it has to be defined both within internal as well as outside. There has to be another way to do this that I am just not aware of though

p0mvn:
From reading the PR description, I have the following comments:

- exporting `getPoolById`
   * can we use an already exported `GetPool instead: https://github.com/osmosis-labs/osmosis/blob/62869348e43946f49c3886f5b1aa44d67ebe721a/x/concentrated-liquidity/pool.go#L36 ? The answer might be no because we want a CL extension
   * if we decide to unexport it, can we rename it to `GetConcentratedPoolI`. The reason is `GetPool` also gets it by id so this might be confusing

x/concentrated-liquidity/internal/math/tick.go
p0mvn: I know that I've agreed about this change during discussions. However, after looking at it the review, I'm concerned about returning 0 for the square root price. Price cannot be zero so this value is invalid, and I'm worried that it will cause us unforeseen problems and bugs in the future.

I prefer giving out fewer shares to users for now while preserving the current model with its correctness.

If you disagree, let's not block this. Just my 2 cents

x/concentrated-liquidity/pool.go
```diff
@@ -45,8 +45,8 @@ func (k Keeper) GetPool(ctx sdk.Context, poolId uint64) (poolmanagertypes.PoolI,
 	return poolI, nil
 }
 
-// getPoolById returns a concentratedPoolExtension that corresponds to the requested pool id. Returns error if pool id is not found.
-func (k Keeper) getPoolById(ctx sdk.Context, poolId uint64) (types.ConcentratedPoolExtension, error) {
+// GetPoolById returns a concentratedPoolExtension that corresponds to the requested pool id. Returns error if pool id is not found.
+func (k Keeper) GetPoolById(ctx sdk.Context, poolId uint64) (types.ConcentratedPoolExtension, error) {
```
p0mvn, line 49: Per my earlier comment,
- Is it possible to use `GetPool` instead
- If not, can we rename this to `GetConcentratedPoolI` because semantically `GetPool` also gets pool by id so current names might start feeling confusing in the future

x/gamm/keeper/msg_server.go
```diff
@@ -46,6 +47,11 @@ func (server msgServer) CreateBalancerPool(goCtx context.Context, msg *balancer.
 	return &balancer.MsgCreateBalancerPoolResponse{PoolID: poolId}, err
 }
 
+func (server msgServer) MigrateSharesToFullRangeConcentratedPosition(goCtx context.Context, msg *balancer.MsgMigrateSharesToFullRangeConcentratedPosition) (*balancer.MsgMigrateSharesToFullRangeConcentratedPositionResponse, error) {
```
p0mvn, line 50: Why do we need to methods for ` MigrateSharesToFullRangeConcentratedPosition` and `MigrateShares` here?

Looks like we could have just one

x/gamm/keeper/msg_server.go
```diff
@@ -309,3 +315,50 @@ func (server msgServer) ExitSwapShareAmountIn(goCtx context.Context, msg *types.
...
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+	}
+
+	exitCoins, err := server.keeper.ExitPool(ctx, sender, poolIdLeaving, msg.SharesToMigrate.Amount, sdk.NewCoins())
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+	}
+
+	pool, err := server.keeper.clKeeper.GetPoolById(ctx, msg.PoolIdEntering)
```
p0mvn, line 338: It might make sense to first try to get the CL pool before exiting the gamm pool in case the CL pool ends up being non-existent

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +361,158 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+				sharesToMigrateDenom:  defaultGammShares.Denom,
+				sharesToMigrateAmount: defaultGammShares.Amount.Quo(sdk.NewInt(2)),
+				poolIdEntering:        2,
+			},
+			sharesToCreate:             defaultGammShares.Amount.Mul(sdk.NewInt(2)),
+			expectedMigrateShareEvents: 1,
+			expectedMessageEvents:      3, // 1 exitPool, 1 createPosition, 1 migrateShares.
+		},
+		{
+			name: "error: attempt to migrate shares from non-existent pool",
```
p0mvn, line 423: we should have a test case trying to migrate to a pool that doesn't exist

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +361,158 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+
+		// Reset event counts to 0 by creating a new manager.
+		suite.Ctx = suite.Ctx.WithEventManager(sdk.NewEventManager())
+		suite.Require().Equal(0, len(suite.Ctx.EventManager().Events()))
+
+		// Migrate the user's gamm shares to a full range concentrated liquidity position
+		resp, err := msgServer.MigrateSharesToFullRangeConcentratedPosition(sdk.WrapSDKContext(suite.Ctx), msg)
+		if test.expectedErr != nil {
+			suite.Require().Error(err)
+			suite.Require().ErrorContains(err, test.expectedErr.Error())
```
p0mvn, line 485: Consider validating here that the gamm pool still has shares while CL pool does not have a position created

x/gamm/keeper/pool.go
```diff
@@ -295,3 +297,10 @@ func convertToCFMMPool(pool poolmanagertypes.PoolI) (types.CFMMPoolI, error) {
 	}
 	return cfmmPool, nil
 }
+
+func GetPoolIdFromSharesDenom(denom string) (uint64, error) {
```
p0mvn, line 301: does this have to be exported?

Also, please godoc it

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: Agreed with removing this logic, especially with Alpin's point on our call

https://github.com/osmosis-labs/osmosis/pull/4051/commits/3f4f84d93a1f5d16c4dd22177088668d569ee359

czarcas7ic: When removing this logic though, the following test's usdc balance is off by 100 units instead of just 1 

https://github.com/osmosis-labs/osmosis/blob/3f4f84d93a1f5d16c4dd22177088668d569ee359/x/gamm/keeper/msg_server_test.go#L386-L397

czarcas7ic: I think we need to look into this before merging this. Here is the updated tolerance for the test that is required when reverting this logic

https://github.com/osmosis-labs/osmosis/pull/4051/commits/e0e3a89b8b8a04aa0e5bdda4ffd0488c9c6a3950

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: Your idea from the call worked, thanks :)

https://github.com/osmosis-labs/osmosis/pull/4051/commits/297ca1c4d2bac22d0edaccfa1e92ea151cf8c18b

AlpinYukseloglu: For reviewer context, the suggestion on the call was to keep the internal function private and have a getter method exporting it to public at the CL package layer

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +361,160 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+		userEthBalanceTransferredToClPool := gammPoolEthBalancePostJoin.Sub(gammPoolEthBalancePostMigrate)
+		userUsdcBalanceTransferredToClPool := gammPoolUsdcBalancePostJoin.Sub(gammPoolUsdcBalancePostMigrate)
+
+		// Note cl pool balance after migration
+		clPoolEthBalanceAfterMigration := suite.App.BankKeeper.GetBalance(suite.Ctx, clPoolAddress, ETH)
+		clPoolUsdcBalanceAfterMigration := suite.App.BankKeeper.GetBalance(suite.Ctx, clPoolAddress, USDC)
+
+		// The balance in the cl pool should be equal to what the user previously had in the gamm pool
+		// This test is within 100 shares due to rounding that occurs from utilizing .000000000000000001 instead of 0
+		poolEthBalanceInTolerance := userEthBalanceTransferredToClPool.Amount.Sub(clPoolEthBalanceAfterMigration.Amount).LTE(sdk.NewInt(1))
```
AlpinYukseloglu, line 520: These checks should probably use `Compare`, and the tolerance should be extracted into a constant. If you only want to check tolerance in one direction, you can use the `RoundingDirection` field in `ErrTolerance` (this is how we check rounding behavior in stableswap)

Right now, it trivially passes if the result of the subtraction is negative

x/concentrated-liquidity/pool.go
czarcas7ic: Used GetPool and unexported getPoolById!

https://github.com/osmosis-labs/osmosis/pull/4051/commits/3b878d6a453dff3fdbbe471361d8381947aa8b17

x/gamm/keeper/msg_server.go
czarcas7ic: Done!

https://github.com/osmosis-labs/osmosis/pull/4051/commits/f0fccf345f46f4392c778accd153dacef96620e9

x/gamm/keeper/msg_server.go
czarcas7ic: Couldn't we say the same thing for CreateBalancerPool?

https://github.com/osmosis-labs/osmosis/blob/f0fccf345f46f4392c778accd153dacef96620e9/x/gamm/keeper/msg_server.go#L47-L50

I was just following previous design of extracting the primary logic in the msg Server, but I can change this if you think I should divert from previous design

x/gamm/keeper/msg_server_test.go
AlpinYukseloglu: Also it seems this check is comparing the user's transferred amount to the entire gamm pool's balance – this means it always passes trivially no?

x/gamm/keeper/msg_server_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4051/commits/30e8c1f1bf5506b37ea07553537ea6b31c2a7f43

x/gamm/keeper/msg_server_test.go
```diff
@@ -356,3 +362,171 @@ func (suite *KeeperTestSuite) TestExitPool_Events() {
...
+	defaultAccountFunds := sdk.NewCoins(sdk.NewCoin("eth", sdk.NewInt(200000000000)), sdk.NewCoin("usdc", sdk.NewInt(200000000000)))
+
+	type param struct {
+		sender                sdk.AccAddress
+		sharesToMigrateDenom  string
+		sharesToMigrateAmount sdk.Int
+		poolIdEntering        uint64
+	}
+
+	tests := []struct {
```
AlpinYukseloglu, line 379: We should check the exact state of the CL position that's created _somewhere_ (e.g. `expectedPosition`) even just as a sanity check so all our tests aren't relative to each other

x/gamm/keeper/pool.go
czarcas7ic: unexported and added godoc

https://github.com/osmosis-labs/osmosis/pull/4051/commits/7dfa7d09b58891e7f4598d85a572d8aeb0e6f87f

x/gamm/keeper/msg_server_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4051/commits/79009330eca4e9c173a63ee752de127812578a2e

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: Sorry, should have added the context, thanks!

x/gamm/keeper/msg_server_test.go
czarcas7ic: Addressed here https://github.com/osmosis-labs/osmosis/pull/4051/commits/65ed90e59267a06b1515671f21ffbae0cadb20ef

This check does not compare the user's transferred amount to the entire gamm pool balance though. It compares the cl pool balance to what was removed from the gamm pool after the migration, so it does not pass trivially

x/gamm/keeper/msg_server_test.go
czarcas7ic: Agreed, added!

https://github.com/osmosis-labs/osmosis/pull/4051/commits/9efd06c8b20c8039c61ba2b0a16cd8529dcc14de

x/concentrated-liquidity/internal/math/tick.go
czarcas7ic: Let me explain why this tick hack matters for amount1 but not amount0.

When adding print lines to the `msg_server_tests` we see the following for the first three tests:
```
exitCoins: 100000000000eth,100000000000usdc 
amount0: 100000000000 
amount1: 99999999900 

exitCoins: 50000000000eth,50000000000usdc 
amount0: 50000000000 
amount1: 49999999950 

exitCoins: 49999999999eth,49999999999usdc 
amount0: 49999999998 
amount1: 49999999949 
```

ExitCoins is from the ExitPool func from the gamm pool and amount0/amount1 is what actually gets calculated when we create a full range position in the cl pool. As we see, amount0 is always calculating a good number but amount1 is off. This is because the following:

When we run CalcAmount0Delta in cl, we use sqrtPriceCurrent (1) and sqrtPriceMax (10000000000000000000.000000000000000000). From there,
`diff = 9999999999999999999.000000000000000000`
`denominator = 10000000000000000000.000000000000000000`
`liquidity * diff / denominator = amount0 = 100000000000`

When we CalcAmount1Delta however, we us sqrtPriceCurrent (1) and sqrtPriceMin (0.000000001). From there,
`diff = 0.999999999000000000`
`liquidity * diff = amount1 = 99999999900`

AlpinYukseloglu (approved):
Discussed on call and this version generally LGTM - I think the precision issue is mostly fine as it is and can be improved in a future PR if there is ongoing debate about it

x/concentrated-liquidity/internal/math/tick.go
ValarDragon: So it makes sense to me that when we exit we get coins `C`, at ratio `R`, and when we rejoin we join coins `C'` at ratio `R'`, and `C'` makes full use of exactly one of the two coins.

What would make sense to me to happen, is to be single-asset joining the remaining `C - C'` coins.

czarcas7ic: Thanks Dev, this actually makes sense. I think right now I am going to leave the remaining dust in the account holders wallet instead of doing a second join for the sake of simplicity / gas. We can revisit later, but yeah this seems like a non issue now that you have explained it this way.

x/gamm/keeper/migrate.go
```diff
@@ -0,0 +1,204 @@
...
+		// Ensure records are sorted from lowest to highest balancer pool ID
+		if record.BalancerPoolId < lastBalancerPoolID {
+			return fmt.Errorf(
+				"Balancer pool ID #%d came after Balancer pool ID #%d.",
+				record.BalancerPoolId, lastBalancerPoolID,
+			)
+		}
+
+		// Ensure the provided balancerPoolId exists and that it is of type balancer
+		poolType, err := k.GetPoolType(ctx, record.BalancerPoolId)
```
p0mvn, line 121: Suggest changing `GetType` to be a method on the pool struct.

Then,
- get pool by id
- `GetType` on the pool

Having this defined on keeper doesn't feel right, especially considering that CL has only one type

x/gamm/keeper/msg_server.go
p0mvn: `CreateBalancerPool` exists because this what's defined by the message server interface:
https://github.com/osmosis-labs/osmosis/blob/df3853ff9886dad8604165f1fb43299d23caec2e/x/gamm/pool-models/balancer/tx.pb.go#L218-L220

Similar to `CreateStablesSwapPoll`. The reason we have `CreatePool` is to share logic between balancer and stableswap.

I'm not exactly sure why we would have redundant APIs with migrations

x/gamm/keeper/migrate.go
```diff
@@ -0,0 +1,204 @@
...
+	// Get the concentrated pool from the message and type cast it to ConcentratedPoolExtension.
+	poolI, err := k.clKeeper.GetPool(ctx, poolIdEntering)
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, err
+	}
+	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
+	if !ok {
+		// If the conversion fails, return an error.
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
+	}
```
p0mvn, line 47: If we define the `GetType` on the pool struct, we wouldn't have to typecast here which is an anti-pattern. Instead, we could check the pool type and error if it is not CL

x/concentrated-liquidity/tick.go
```diff
@@ -127,3 +127,9 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
 	}
 	return nil
 }
+
+// getMinAndMaxTicksFromExponentAtPriceOne determines min and max ticks allowed for a given exponentAtPriceOne value
```
p0mvn, line 131:
```suggestion
// GetMinAndMaxTicksFromExponentAtPriceOne determines min and max ticks allowed for a given exponentAtPriceOne value
```

p0mvn (approved):

x/gamm/keeper/migrate.go
czarcas7ic: I am implementing the GetType that you suggested, but we actually do need to type case here since we need to utilize GetPrecisionFactorAtPriceOne and GetToken0/1 (the type case is for utilizing specific functions rather than confirming its type)

x/gamm/keeper/migrate.go
czarcas7ic: Agreed, addressed here! 

https://github.com/osmosis-labs/osmosis/pull/4051/commits/a35fecb25da1660673532231fdc0d5ec821c2ad2

x/gamm/keeper/msg_server.go
czarcas7ic: Agreed, I misunderstood the purpose, changed!

https://github.com/osmosis-labs/osmosis/pull/4051/commits/b1008bc5da26c347b7969d4ecebe6c76c038a8ae

PR #4063, opened 2023-01-18
Simplified XC-swaps interface

nicolaslara:

Simplifies the outpost and crosschain swaps contracts by removing the need to specify the input denom (it can be inferred from the sent funds)

## Brief Changelog

 
  - Simplified contract interfaces

## Testing and Verifying
All tests pass


cosmwasm/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -51,25 +51,30 @@ pub fn execute(
     info: MessageInfo,
     msg: ExecuteMsg,
 ) -> Result<Response, ContractError> {
+    // IBC transfers support only one token at a time
     match msg {
         ExecuteMsg::OsmosisSwap {
-            input_coin,
+            swap_amount,
```
p0mvn, line 57: nit: consider just `amount`

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -24,7 +24,8 @@ pub fn swap_and_forward(
     deps: DepsMut,
     block_time: Timestamp,
     contract_addr: Addr,
-    input_coin: Coin,
+    swap_amount: u128,
+    received_coin: Coin,
```
p0mvn, line 28: why do we need both amount and coin?

cosmwasm/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -51,25 +51,30 @@ pub fn execute(
...
-            env.contract.address,
-            input_coin,
-            output_denom,
-            slippage,
-            receiver,
-            next_memo,
-            on_failed_delivery,
-        ),
+        } => {
+            let coin = cw_utils::one_coin(&info)?;
```
p0mvn, line 64: can we set `coin`'s amount to `swap_amount` and pass in just the `coin` to `swap_and_forward`?

p0mvn (approved):

cosmwasm/contracts/crosschain-swaps/src/contract.rs
nicolaslara: do you mean pass just the amount?

The reason for passing both is that the amounts may not be the same in the future. Specifically, when we start charging fees for the swap, we will want the amount of funds sent to the contract to be higher than the amount you want to swap

cosmwasm/contracts/crosschain-swaps/src/contract.rs
nicolaslara: I was worried this would be confused with the minAmountOut parameter of the slippage (and maybe later with other params), but maybe I'm worrying about nothing?

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -34,13 +35,16 @@ pub fn swap_and_forward(
     deps.api.debug(&format!("executing swap and forward"));
     let config = CONFIG.load(deps.storage)?;
 
+    // TODO: Here we can take the fees
+    let swap_coin = Coin::new(swap_amount, received_coin.denom);
```
p0mvn, line 39: I mean why would we not use `received_coin.amount` here? What is the reason for having `swap_amount` separately when `received_coin`'s amount is unused

cosmwasm/contracts/crosschain-swaps/src/contract.rs
p0mvn: https://github.com/osmosis-labs/osmosis/pull/4063/files#r1081482123

cosmwasm/contracts/crosschain-swaps/src/execute.rs
p0mvn: I'm down to not block on this and merge the PR if my proposal seems undesirable

nicolaslara: using `received_coin.amount` there is the same as just using `received_coin`. That would work right now. 

The reson to have it be different is that I'm planning changes to the contract that will require the amount you send to be slightly higher than the swapped amount. Specifically, fees. If we want to charge a fee for the transfer, the user needs to send swap_amount+fee_amount to the contract. So, if they want to swap 10osmo for juno (assuming 1:1 price and a fee of 1%) they will have to send 11 osmo to the contract so they pay 1 osmo in fees and get 10 juno back. To specify that, the send 11 osmo, but in the message include `"swap_amount": "10"`.

Since the fees may not be known in advance, I'm also planning a "cash-back" mechanism where they would be allowed to send a much higher amount without loss: send 100 osmo to the contract (with swap_amount=10), and get back: 10juno, 89osmo (where 1 osmo went to pay the fee). 

(In the last example the fee is specified as 1 osmo and not as a percentage. Still need to design the best strategy for fees)

PR #4038, opened 2023-01-17
e2e, concentrated-liquidity: create concentrated pool

RusAkh:
- rename CreatePool to CreateBalancerPool
- scaffold CreateConcentratedPool
- CreateConcentratedPool

Closes: #4037 (in addition: https://github.com/osmosis-labs/osmosis/issues/4030)

## Changes

This PR introduces a new function in `tests/e2e/configurer/chain/commands.go` which creates a concentrated pool.

Also, renames an existing `CreatePool` function to `CreateBalancerPool` to avoid ambiguity. 

## Testing and Verifying

Not yet tested 

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? n
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? n
  - How is the feature or change documented? n

PR #4068, opened 2023-01-19
e2e: setDenomMetadata

RusAkh:
## Brief Changelog

This PR adds a function to register denom metadata in genesis state of bank module in e2e tests.

## Problem that this PR solves 

https://github.com/osmosis-labs/osmosis/pull/4038 while working at this PR I have faced a problem that you cannot create a concentrated pool with any denom you want (what is weird, `gamm` does not have `DenomMetadata` check). Because of that, I was unable to add `CreateConcentratedPool` function to `e2e`, because tests would not pass (reason: using `UpgradeConfigurer` which does not register denoms other than `uosmo`). This PR is backport and introduces:

- A general function to add `DenomMetadata` to genesis state of bank module
- More denoms added to genesis state of bank module

### DIFF
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -67,6 +67,9 @@ Conceptually, we can split the e2e setup into 2 parts:
     configurations is hapenning in the background. The appropriate logic is selected
     depending on what the values of the arguments to `configurer.New(...)` are.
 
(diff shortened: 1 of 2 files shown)


p0mvn (approved):
Nice

czarcas7ic:
This needs to be done in LocalOsmosis as well. I realized this when trying to test some CL stuff.

Issue #4027, opened 2023-01-16
[Story][CL]: Simulator testing for concentrated liquidity


p0mvn:
## Background

> As an engineer, I would like to have randomized testing for concentrated liquidity, so that I can be more confident that the core logic works as expected under various situations

## Suggested Design

Implement a simulator for all concentrated liquidity messages, that would be similar with the gamm simulator( cref: https://github.com/osmosis-labs/osmosis/tree/main/x/gamm/simulation)

## Tasks

We want to have the following components tested via simulator

- [x] - Random Create Position Msg  https://github.com/osmosis-labs/osmosis/pull/4185
- [x] - Random Withdraw Position Msg https://github.com/osmosis-labs/osmosis/pull/4185
- [x] - Random create concentrated liquidity pool https://github.com/osmosis-labs/osmosis/pull/4094
- [ ] - RandomSwapExactAmountIn https://github.com/osmosis-labs/osmosis/pull/4241
- [ ] - RandomSwapExactAmountOut https://github.com/osmosis-labs/osmosis/pull/4241
- [x] - Random CollectFees  https://github.com/osmosis-labs/osmosis/pull/4185

We also want to add simulator testings for the pool manager module if possible

## Acceptance Criteria
- simulator is implemented for concentrated liquidity
- all messages are tested

p0mvn:
Hi @stackman27. Could you please break this down into tasks?

stackman27:
yup will do apologies for the delay

stackman27:
#4241 blocked by https://github.com/osmosis-labs/osmosis/issues/4366

p0mvn:
Addressed now

PR #4094, opened 2023-01-24
[CL: Simulator] Simulator for MsgCreateConcentratedPool 

stackman27:
Part of: https://github.com/osmosis-labs/osmosis/issues/4027

Add simulator testing to CL module messages starting with MsgCreateConcentratedPool

## Brief Changelog
n/a

## Testing and Verifying
n/a


stackman27:
Keeping this as Draft rn because with new CL sim-tests, some of gamm tests are failing. I am currently investigating this

x/concentrated-liquidity/model/pool.go
```diff
@@ -234,6 +234,7 @@ func (p *Pool) ApplySwap(newLiquidity sdk.Dec, newCurrentTick sdk.Int, newCurren
 	return nil
 }
 
+// TODO: finish this function
```
p0mvn, line 237: Good catch. Let's make an issue to track this, please

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	exponentAtPriceOne := sdk.NewInt(rand.Int63n(1+13) - 13)
+	authorizedTickSpacing := []uint64{1, 10, 60, 200}
+
+	// find an address with two or more distinct denoms in their wallet
+	sender, senderExists := sim.RandomSimAccountWithConstraint(createPoolRestriction(k, sim, ctx))
+	if !senderExists {
+		return nil, fmt.Errorf("no sender with two different denoms & pool creation fee exists")
+	}
+
+	// generate 3 coins, use 2 to create pool and 1 for fees. ["stake" denom - contiains invalid metadata]
```
p0mvn, line 31: ["stake" denom - contains invalid metadata] - what does this mean?

```suggestion
	// generate 3 coins, use 2 to create pool and 1 for fees. ["stake" denom - contains invalid metadata]
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	poolCoins, ok := sim.GetRandSubsetOfKDenoms(ctx, sender, 3)
```
p0mvn, line 32: Why are we getting subset of 3 denoms when `createPoolRestriction` returns accounts with only 2 coins?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) osmosimtypes.SimAccountConstraint {
+	return func(acc legacysimulationtype.Account) bool {
+		accCoins := sim.BankKeeper().SpendableCoins(ctx, acc.Address)
+		hasTwoCoins := len(accCoins) >= 2
+		hasPoolCreationFee := accCoins.AmountOf(PoolCreationFee.Denom).GT(PoolCreationFee.Amount)
+		return hasTwoCoins && hasPoolCreationFee
+	}
+}
+
+// getRandCLPool gets a concnerated liquidity pool with its pool denoms.
```
p0mvn, line 84: Let's try to avoid comitting comments if possible please

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+
+	// generate 3 coins, use 2 to create pool and 1 for fees. ["stake" denom - contiains invalid metadata]
+	poolCoins, ok := sim.GetRandSubsetOfKDenoms(ctx, sender, 3)
+	if !ok {
+		return nil, fmt.Errorf("provided sender with requested number of denoms does not exist")
+	}
+
+	// check if the sender has sufficient amount for fees
+	if poolCoins.Add(PoolCreationFee).IsAnyGT(sim.BankKeeper().SpendableCoins(ctx, sender.Address)) {
+		return nil, errors.New("chose an account / creation amount that didn't pass fee bar")
```
mattverse, line 39: what does the fee bar mean?

x/concentrated-liquidity/simulation/sim_msgs.go
mattverse: Hrmm, I think @stackman27 's original intent on this so that we can have 2 denoms for the pool, and then another denom for the pool creation fee, but as @p0mvn mentioned,  I think we should be having 2 rand subset here

mattverse (commented):
LGTM once the denom subset has been resolved

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: also the other reason was that "stake" denom fails this check. So i had to use 2 other gamm denoms here 
```
	_, denomExists := server.keeper.bankKeeper.GetDenomMetaData(ctx, msg.Denom0)
	if !denomExists {
		return nil, fmt.Errorf("received denom0 with invalid metadata: %s", msg.Denom0)
	}
```

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: I set the pool creation fee [here](https://github.com/osmosis-labs/osmosis/pull/4094/files#diff-f01e8795f0fb3ac50fd61ff5441aac0f1e4153629f16c05da9c3a8785c408dc9R18), which is the fee bar

x/concentrated-liquidity/model/pool.go
stackman27: just did https://github.com/osmosis-labs/osmosis/issues/4110

x/concentrated-liquidity/simulation/sim_msgs.go
czarcas7ic: By fee bar do you mean fee limit? I agree fee bar is not clear on meaning

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	clkeeper "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity"
+	clmodeltypes "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/model"
+	cltypes "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types"
+)
+
+var PoolCreationFee = sdk.NewInt64Coin("stake", 10_000_000)
+
+func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*clmodeltypes.MsgCreateConcentratedPool, error) {
+	// generate random values from -13 to 1 (accepted range: -12 to -1)
+	exponentAtPriceOne := sdk.NewInt(rand.Int63n(1+13) - 13)
```
czarcas7ic, line 22: Can we directly call the Max and Min exponentAtPriceOne limits and add/subtract to them? That way if we decide to change them in the future we still have the same logic apply?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	authorizedTickSpacing := []uint64{1, 10, 60, 200}
```
czarcas7ic, line 23: Same with authorized tick spacing, I believe there is a const somewhere that we can refer to.

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: sounds good i'll change it

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	// check if the sender has sufficient amount for fees
+	if poolCoins.Add(PoolCreationFee).IsAnyGT(sim.BankKeeper().SpendableCoins(ctx, sender.Address)) {
+		return nil, errors.New("chose an account / creation amount that didn't pass fee bar")
+	}
+
+	denom0 := poolCoins[0].Denom
+	denom1 := poolCoins[1].Denom
+
+	tickSpacing := authorizedTickSpacing[rand.Intn(len(authorizedTickSpacing))]
+	precisionFactorAtPriceOne := exponentAtPriceOne
```
czarcas7ic, line 46: nit: The conversion here is not necessary

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+	// get all pool positions for a specific pool (we can randomize which pool to use)
+	// - this will include, position lower_tick, upper_tick, liquidityAmt
+	// randomly select 1 pool position
+	// get random withdraw liquidity from [0 to existing liqudityAmt]
+	return &cltypes.MsgWithdrawPosition{}, nil
+}
+
+func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgCollectFees, error) {
+	return &cltypes.MsgCollectFees{}, nil
+}
```
czarcas7ic, line 68: Im confused, what do these functions do?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,119 @@
...
+
+func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) osmosimtypes.SimAccountConstraint {
```
czarcas7ic, line 70: godoc

x/gamm/simulation/sim_msgs.go
```diff
@@ -421,13 +422,9 @@ func getRandPool(k keeper.Keeper, sim *simtypes.SimCtx, ctx sdk.Context) (uint64
...
 
-	pool_id := uint64(simtypes.RandLTBound(sim, numPools) + 1)
-
-	pool := pools[pool_id-1]
-	if err != nil {
-		return 0, nil, sdk.NewCoin("denom", sdk.ZeroInt()), sdk.NewCoin("denom", sdk.ZeroInt()), []string{}, "", err
-	}
+	pool := pools[rand.Intn(numPools)]
 
+	pool_id := pool.GetId()
```
czarcas7ic, line 427: nit: no snake case

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: removed, it was for subsequent PRs

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: addressed

stackman27:
Resolved all your comments @czarcas7ic

stackman27:
@p0mvn @mattverse @czarcas7ic Thank you for the feedback! All comments should be resolved now :taco:

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -1,17 +1,68 @@
...
+)
+
+var PoolCreationFee = sdk.NewInt64Coin("stake", 10_000_000)
+
+func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*clmodeltypes.MsgCreateConcentratedPool, error) {
+	minExponentAtOneValue := cltypes.ExponentAtPriceOneMin.Int64()
+	maxExponentAtOneValue := cltypes.ExponentAtPriceOneMax.Int64()
+
+	// generate random values from -13 to 1 (current accepted range: -12 to -1)
+	exponentAtPriceOne := sdk.NewInt(minExponentAtOneValue + rand.Int63n(maxExponentAtOneValue-minExponentAtOneValue+1))
```
czarcas7ic, line 25: Im not quite understanding this. Why don't we just pick a random int between
`minExponentAtOneValue - 1` and `maxExponentAtOneValue + 2`?

czarcas7ic (commented):
I think we can merge once we get consensus on https://github.com/osmosis-labs/osmosis/pull/4094/files#r1087481800

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: oh yea forgot about that. It was randomizing(-12, -1). Fixed this to rand(-13 ,1)

stackman27:
Rebased and Ready for final check!

x/concentrated-liquidity/simulation/sim_msgs.go
mattverse: In that case, it seems that here we're taking the first two given coins 
```
denom0 := poolCoins[0].Denom
	denom1 := poolCoins[1].Denom
```

If the case is both 1 and 2 should we not have some sort of logic that verifies that it is not stake denom?

mattverse: Regarding point number 1, the reason it's happening is because of this line `https://github.com/osmosis-labs/osmosis/pull/4094/files#diff-f01e8795f0fb3ac50fd61ff5441aac0f1e4153629f16c05da9c3a8785c408dc9R64`. We can simply change that to be equal to 2 not greater than 2 wihtin the method

mattverse: IMO that should also go in the `createPoolRestriction` as it serves our purpose of createPoolRestrictions

stackman27: I kept it outside because it requires sender which we can only generate with constraint

mattverse: I was rather about doing the following 
```
// createPoolRestriction creates specific restriction for the creation of a pool.
func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) osmosimtypes.SimAccountConstraint {
	return func(acc legacysimulationtype.Account) bool {
		accCoins := sim.BankKeeper().SpendableCoins(ctx, acc.Address)
		hasTwoCoins := len(accCoins) == 2
		
		isStakeDenom := false
		for _, coin := range accCoins {
			if coin.Denom == "stake" {
				isStakeDenom = true
			}
		}
		hasPoolCreationFee := accCoins.AmountOf(PoolCreationFee.Denom).GT(PoolCreationFee.Amount)
		return hasTwoCoins && hasPoolCreationFee && !isStakeDenom
	}
}
```

Suggested this way because having another layer of extra code here (using GetRandSubsetOfKDenoms) seemed unnecessary when we're already calling a method that does this role

stackman27: the issue with this is that we have to have stake denom because that's being used as CreationFee, also any tokens being minted from accCoins contains "stake" denom, but in our case we cannot use stake to create Pool because of the "invalid metadata" problem

czarcas7ic: Can we look into registering stake in bank to get past the invalid metadata problem

stackman27: Yea we can do that by using `SetDenomMetaData` which we'll have to import from bankKeeper, which will break incentives, superfluid simulator. I didn't wanted this to be a big diff so i found this loophole

stackman27: tbh i think we should just rewrite the rest of module simulator to follow gamm structure

stackman27: I tried many things such as;
1. change feeDenom from "stake" to "uosmo" --> Error because: "uosmo" has invalid metadata + user doesnot contain the denom 
2. Set denom metadata before test --> Error: `SetDenomMetaData` cannot be added to simtypes.bankKeeper without breaking incentives, superfluid.  Currently, we can only access specific bankKeeper method in simulator, I created an issue to fix this https://github.com/osmosis-labs/osmosis/issues/4184

p0mvn (approved):
From our discussion on call, the consensus has been reached to merge this and address #4184 separately.

Thank you @stackman27

Issue #4096, opened 2023-01-25
[CL Fees]: Fee - Misc tracking issue


p0mvn:
## Background

This issue consists of misc TODOs that we should revisit before finalizing the fees user story.

## Suggested Design

- [x] index positions by strings in the accum module. Make sure that CL
  positions are indexed by pool, owner and tick range.
- [x] make sure accumulator module allows for prefixes that preserve data locality.
- [x] try moving fees logic into the internal package
- [x] figure out how to make calcs non-mutative
- [x] make sure fees are updated in the right places
- [x] document that if we withdraw all shares, the rewards are claimed and the
fee accumulator position is deleted. As a result, next time we add liqudity for the same
ticks, we need to re-initialzie the 
- [x] remove prints for swaps

## Acceptance Criteria
- all TODOs above are addressed

p0mvn:
Additional tracking:
- https://github.com/osmosis-labs/osmosis/pull/4249#discussion_r1100345811
- https://github.com/osmosis-labs/osmosis/pull/4249#discussion_r1100346638
- https://github.com/osmosis-labs/osmosis/pull/4249#discussion_r1100346863

p0mvn:
Confirmed that all TODOs are either addressed or obsolete

Issue #4088, opened 2023-01-24
feat(CL Fees): implement swap fees for calcIntAmtGivenOut


p0mvn:
## Background

We have a working prototype of swap fees implemented for ccalcOutAmtGivenIn. Now, we need the same for `calcIntAmtGivenOut` .

## Suggested Design

Aim to utilize the same design and helper functions as in `calcOutAmtGivenIn` to minimize code duplication.
Extend the CL python scripts to estimate similar test vectores.
Use the test case estimates to cover the new swap fee logic with tests

## Acceptance Criteria
- swap fees implemented for calcIntAmtGivenOut
- design similar to calcOutAmtGivenIN
- Pythons scripts are extended to estimate test cases
- tests added with coverage maximized

Issue #4086, opened 2023-01-24
test(CL Fees): updateFeeGrowthGlobal method


p0mvn:
## Background

The following method needs to be tested:
https://github.com/osmosis-labs/osmosis/blob/df3853ff9886dad8604165f1fb43299d23caec2e/x/concentrated-liquidity/swaps.go#L33

## Suggested Design

Add table driven test to maximize test coverage

## Acceptance Criteria
- test added
- coverage maximized

p0mvn:
It might be that this issue ends up being obsolete after implementing swap fees for cal out given in. Making it for tracking purposes

p0mvn:
Revisit after https://github.com/osmosis-labs/osmosis/issues/4088 is complete

PR #4097, opened 2023-01-25
feat/test(CL Swaps): Swap fees for amount out given in

p0mvn:
Closes: #XXX

This PR introduces the implementation for swap fees when swapping for amount out given token in.

The test vectors are estimated using the Python `sympy` scripts. The documentation for running the scripts is added.

There are some internal methods that are untested:
- https://github.com/osmosis-labs/osmosis/issues/4087
- https://github.com/osmosis-labs/osmosis/issues/4086

It is best to test them after swap fees in the other direction are complete (ref: https://github.com/osmosis-labs/osmosis/issues/4088). The reason is that we might refactor and remove these methods so it doesn't make sense to spend time adding detailed tests right now. The test coverage via script estimates is sufficient at this stage.

Lastly, there are some prints left in the codebase for development convenience. Removing them is tracked here: https://github.com/osmosis-labs/osmosis/issues/4096

For context, the original implementation was prototyped here: https://github.com/osmosis-labs/osmosis/pull/3893

## Brief Changelog

- implement swap fee for amount out given in
- implement tests with swap fee for all existing edge case tests
   * implement Python scripts to estimate the right values for them
 - refactor with better abstractions

## Testing and Verifying

This change added tests 

### DIFF
--- a/scripts/cl/common.py
+++ b/scripts/cl/common.py
@@ -8,7 +8,7 @@ precision = 30
 # When swapping token one for zero, sqrt_price_current <= sqrt_price_next.
 #
 # sqrt_price_next can be None. When it is None, that implies that the next sqrt
-# price must be calculated. In such a case, nex sqrt price depends on the
+# price must be calculated. In such a case, next sqrt price depends on the
 # remaining amount of token in to be swapped. This occurs for the last sqrt price
 # range in the collection of sqrt price ranges that represent a swap.
 #
--- a/scripts/cl/common.py
+++ b/scripts/cl/common.py
@@ -1,0 +1,39 @@
+import sympy as sp
+
+precision = 30
+
+# SqrtPriceRange represents a price range between the current and the next tick
+# as well as the liquidity in that price range.
+# When swapping token zero for one, sqrt_price_current >= sqrt_price_next.
+# When swapping token one for zero, sqrt_price_current <= sqrt_price_next.
+#
+# sqrt_price_next can be None. When it is None, that implies that the next sqrt
+# price must be calculated. In such a case, nex sqrt price depends on the
+# remaining amount of token in to be swapped. This occurs for the last sqrt price
+# range in the collection of sqrt price ranges that represent a swap.
+#
+# For example, I might have a swap of 100 ETH in from 5000 to 5001 with liquidity
+# X, from 5001 to 5002 with liquidity Y,
+# and from 5002 to UNKNOWN with liquidity Z. In this case, the UNKWNOWN
+# depends on how much ETH we have remaining after consuming liquidity X and Y.
+class SqrtPriceRange:
+  def __init__(self, sqrt_price_current: int, sqrt_price_next: int, liquidity: sp.Float):
+    self.sqrt_price_start = sp.sqrt(fixed_prec_dec(sqrt_price_current))
+    if sqrt_price_next is not None:
+        self.sqrt_price_next = sp.sqrt(fixed_prec_dec(sqrt_price_next))
+    else:
+       self.sqrt_price_next =  None
+    self.liquidity = liquidity
+
+def fixed_prec_dec(string: str) -> sp.Float:
+    """ Return an equivalent of a Python Decimal with fixed precision. 
+    """
+    return sp.Float(string, precision)
+
+def get_fee_amount_per_share(token_in: sp.Float, swap_fee: sp.Float, liquidity: sp.Float) -> sp.Float:
+    """ Returns the fee amount per share.
+    """
+    fee_charge_total = token_in * swap_fee
+    return fee_charge_total / liquidity
+
+zero = fixed_prec_dec("0")
--- a/scripts/cl/common.py
+++ b/scripts/cl/common.py
@@ -14,7 +14,7 @@ precision = 30
 #
 # For example, I might have a swap of 100 ETH in from 5000 to 5001 with liquidity
 # X, from 5001 to 5002 with liquidity Y,
-# and from 5002 to UNKNOWN with liquidity Z. In this case, the UNKWNOWN
+# and from 5002 to UNKNOWN with liquidity Z. In this case, the UNKNOWN
 # depends on how much ETH we have remaining after consuming liquidity X and Y.
 class SqrtPriceRange:
   def __init__(self, sqrt_price_current: int, sqrt_price_next: int, liquidity: sp.Float):
--- a/scripts/cl/main.py
+++ b/scripts/cl/main.py
@@ -1,0 +1,267 @@
+from typing import Tuple
+from common import *
+import zero_for_one as zfo
+import one_for_zero as ofz
+
+def estimate_test_case(tick_ranges: list[SqrtPriceRange], token_in_initial: sp.Float, swap_fee: sp.Float, is_zero_for_one: bool) -> Tuple[sp.Float, sp.Float]:
+    """ Estimates a calc concentrated liquidity test case.
+    
+    Given
+      - sqrt price range with the start sqrt price, next sqrt price and liquidity
+      - initial token in
+      - swap fee
+      - zero for one boolean flag
+    Estimates the final token out and the fee growth per share and prints it to stdout.
+    Also, estimates these and other values at each range and prints them to stdout.
+
+    Returns the total token out and the total fee growth per share.
+    """
+
+    token_in_consumed_total, token_out_total, fee_growth_per_share_total = zero, zero, zero
+
+    for i in range(len(tick_ranges)):
+        tick_range = tick_ranges[i]
+
+        # Normally, for the last swap range we swap until token in runs out
+        # As a result, the next sqrt price for that range calculated at runtime.
+        is_last_range = i == len(tick_ranges) - 1
+        # Except for the cases where we set price limit explicitly. Then, the
+        # last price range may have the upper sqrt price limit configured.
+        is_next_price_set = tick_range.sqrt_price_next != None 
+
+        is_with_next_sqrt_price = not is_last_range or is_next_price_set
+
+        if is_with_next_sqrt_price:
+            token_in_consumed, token_out, fee_growth_per_share = zero, zero, zero
+            if is_zero_for_one:
+                token_in_consumed, token_out, fee_growth_per_share = zfo.calc_test_case_with_next_sqrt_price(tick_range.liquidity, tick_range.sqrt_price_start, tick_range.sqrt_price_next, swap_fee)
+            else:
+                token_in_consumed, token_out, fee_growth_per_share = ofz.calc_test_case_with_next_sqrt_price(tick_range.liquidity, tick_range.sqrt_price_start, tick_range.sqrt_price_next, swap_fee)
+            
+            token_in_consumed_total += token_in_consumed
+            token_out_total += token_out
+            fee_growth_per_share_total += fee_growth_per_share
+
+        else:
+            token_in_remaining = token_in_initial - token_in_consumed_total
+
+            if token_in_remaining < zero:
+                raise Exception(F"token_in_remaining {token_in_remaining} is negative with token_in_initial {token_in_initial} and token_in_consumed_total {token_in_consumed_total}")
+
+            token_out, fee_growth_per_share = zero, zero
+            if is_zero_for_one:
+                _, token_out, fee_growth_per_share = zfo.calc_test_case(tick_range.liquidity, tick_range.sqrt_price_start, token_in_remaining, swap_fee)
+            else:
+                _, token_out, fee_growth_per_share = ofz.calc_test_case(tick_range.liquidity, tick_range.sqrt_price_start, token_in_remaining, swap_fee)
+
+            token_out_total += token_out
+            fee_growth_per_share_total += fee_growth_per_share
+        print("\n")
+        print(F"After processing range {i}")
+        print(F"current token_out_total: {token_out_total}")
+        print(F"current current fee_growth_per_share_total: {fee_growth_per_share_total}")
+        print("\n\n\n")
+
+    print("\n\n")
+    print("Final results:")
+    print("token_out_total: ", token_out_total)
+    print("fee_growth_per_share_total: ", fee_growth_per_share_total)
+
+    return token_out_total, fee_growth_per_share_total
+
+def validate_confirmed_results(token_out_total: sp.Float, fee_growth_per_share_total: sp.Float, expected_token_out_total: sp.Float, expected_fee_growth_per_share_total: sp.Float):
+    """Validates the results of a calc concentrated liquidity test case estimates.
+
+    This validation helper exists to make sure that subsequent changes to the script do not break test cases.
+    """
+
+    if sp.N(token_out_total, 18) != sp.N(expected_token_out_total, 18):
+        raise Exception(F"token_out_total {token_out_total} does not match expected_token_out_total {expected_token_out_total}")
+    
+    if sp.N(fee_growth_per_share_total, 18) != sp.N(expected_fee_growth_per_share_total, 18):
+        raise Exception(F"fee_growth_per_share_total {fee_growth_per_share_total} does not match expected_fee_growth_per_share_total {expected_fee_growth_per_share_total}")
+
+def estimate_single_position_within_one_tick_ofz():
+    """Estimates and prints the results of a calc concentrated liquidity test case with a single position within one tick
+    when swapping token one for token zero (ofz).
+
+     go test -timeout 30s -v -run TestKeeperTestSuite/TestCalcAndSwapOutAmtGivenIn/fee_1 github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity
+    """
+
+    is_zero_for_one = False
+    swap_fee = fixed_prec_dec("0.01")
+    token_in_initial = fixed_prec_dec("42000000")
+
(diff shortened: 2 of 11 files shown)


scripts/cl/common.py
```diff
@@ -0,0 +1,39 @@
...
+
+precision = 30
+
+# SqrtPriceRange represents a price range between the current and the next tick
+# as well as the liquidity in that price range.
+# When swapping token zero for one, sqrt_price_current >= sqrt_price_next.
+# When swapping token one for zero, sqrt_price_current <= sqrt_price_next.
+#
+# sqrt_price_next can be None. When it is None, that implies that the next sqrt
+# price must be calculated. In such a case, nex sqrt price depends on the
```
p0mvn, line 11:
```suggestion
# price must be calculated. In such a case, next sqrt price depends on the
```

x/concentrated-liquidity/swaps.go
```diff
@@ -301,21 +320,37 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 			return writeCtx, sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, fmt.Errorf("could not convert next tick (%v) to nextSqrtPrice", nextTick)
 		}
 
+		// Charge fee
+		feeOnFullAmountRemainingIn := swapState.amountSpecifiedRemaining.Mul(swapFee)
+
+		// TODO: remove this print.
+		fmt.Println("sqrtPrice before swap step", swapState.sqrtPrice)
```
p0mvn, line 327: Note: this and other removal TODOs are tracked in #4096

scripts/cl/main.py
```diff
@@ -0,0 +1,267 @@
...
+    ]
+
+    token_out_total, fee_growth_per_share_total = estimate_test_case(tick_ranges, token_in_initial, swap_fee, is_zero_for_one)
+
+    expected_token_out_total = fixed_prec_dec("64417624.9871649525380486017974")
+    expected_fee_growth_per_share_total = fixed_prec_dec("0.0000000849292577225588233432564611676")
+
+    validate_confirmed_results(token_out_total, fee_growth_per_share_total, expected_token_out_total, expected_fee_growth_per_share_total)
+
+def main():
```
p0mvn, line 244: Note: For scripts, I suggest reviewing this file in-detail and comparing it against Go test cases. However, I don't think there is value in reviewing other internal script calculations so feel free to skip

x/concentrated-liquidity/swaps.go
```diff
@@ -340,11 +375,22 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 				return writeCtx, sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
 			}
 		}
+
+		// update the swapState with the new sqrtPrice from the above swap
+		swapState.sqrtPrice = sqrtPrice
+		// we deduct the amount of tokens we input in the computeSwapStep above from the user's defined tokenIn amount
+		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(amountIn.Add(feeChargeTotal))
+		// we add the amount of tokens we received (amountOut) from the computeSwapStep above to the amountCalculated accumulator
+		swapState.amountCalculated = swapState.amountCalculated.Add(amountOut)
```
mattverse, line 384: Oh interesting, this looks like something we should have had prior to this PR as well, do we not have test cases for testing if these values were properly returned? If not, should we create an issue to track that?

mattverse (commented):
Left some minor comments, please take a look. Upon resolving / answered this LGTM.

(Note that I did not review the python code in detail as they were used for mathematic calculations)

x/concentrated-liquidity/swaps.go
p0mvn: Might be useful for debugging while we implement the swap in the other direction. Since the helpers are shared, changing them might break this kind of swap as well

p0mvn: I don't mind removing, I can just re-add them as needed so let me know if you strongly feel about this

x/concentrated-liquidity/fees.go
```diff
@@ -255,3 +255,45 @@ func calculateFeeGrowth(targetTick int64, feeGrowthOutside sdk.DecCoins, current
...
+// - nextSqrtPrice the sqrt price at which the swap step should end.
+// - sqrtPriceLimit the sqrt price corresponding to the sqrt of the price representing price impact protection.
+// - amountIn the amount of token in to be consumed during the swap step
+// - amountSpecifiedRemaining is the total remaining amount of token in that needs to be consumed to complete the swap.
+// - swapFee the swap fee to be charged.
+//
+// If swap fee is negative, it panics.
+// If swap fee is 0, returns 0. Otherwise, computes and returns the fee charge per step.
+// TODO: test this function.
+func computeFeeChargePerStep(currentSqrtPrice, nextSqrtPrice, sqrtPriceLimit, amountIn, amountSpecifiedRemaining, swapFee sdk.Dec) sdk.Dec {
```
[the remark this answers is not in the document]
p0mvn: per swap step. will change the name to reflect this

x/concentrated-liquidity/swaps.go
```diff
@@ -20,6 +20,24 @@ type SwapState struct {
...
+// As a result, the range from the end of position one to the beginning of position
+// two has no liquidity and can be skipped.
+// TODO: test
+func (ss *SwapState) updateFeeGrowthGlobal(feeChargeTotal sdk.Dec) {
+	if !ss.liquidity.IsZero() {
+		feeChargePerUnitOfLiquidity := feeChargeTotal.Quo(ss.liquidity)
+		// TODO: remove print.
+		fmt.Println("feeChargePerUnitOfLiquidity", feeChargePerUnitOfLiquidity)
+		ss.feeGrowthGlobal = ss.feeGrowthGlobal.Add(feeChargePerUnitOfLiquidity)
+	}
```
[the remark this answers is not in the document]
p0mvn: This might change as we implement the swap in the other direction. I don't think it makes sense to test these right now given the high likelihood of these methods being changed / removed in the next PR.

x/concentrated-liquidity/swaps.go
p0mvn: These are just moved from a few lines above. No changes have been made other than moving

x/concentrated-liquidity/swaps.go
p0mvn: removed now

x/concentrated-liquidity/swaps.go
p0mvn: Also, these are tracked here:
- https://github.com/osmosis-labs/osmosis/issues/4089
- https://github.com/osmosis-labs/osmosis/issues/4086

p0mvn: I moved this in the location as to reduce diff to make this less confusing

x/concentrated-liquidity/swaps.go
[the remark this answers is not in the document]
p0mvn: Also, these are tracked here:
- https://github.com/osmosis-labs/osmosis/issues/4089
- https://github.com/osmosis-labs/osmosis/issues/4086

x/concentrated-liquidity/fees.go
```diff
@@ -255,3 +255,45 @@ func calculateFeeGrowth(targetTick int64, feeGrowthOutside sdk.DecCoins, current
...
+	}
+
+	if swapFee.IsZero() {
+		return feeChargeTotal
+	}
+
+	// 1. The current tick does not have enough liqudity to fulfill the swap.
+	didReachNextSqrtPrice := nextSqrtPrice.Equal(currentSqrtPrice)
+	// 2. The next sqrt price was not reached due to price impact protection.
+	isPriceImpactProtection := currentSqrtPrice.Equal(sqrtPriceLimit)
```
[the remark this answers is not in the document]
p0mvn: Will get back to this

x/concentrated-liquidity/swaps.go
AlpinYukseloglu: You could also leave them as commented prints like we do elsewhere if needed

scripts/cl/common.py
```diff
@@ -0,0 +1,39 @@
...
+# When swapping token one for zero, sqrt_price_current <= sqrt_price_next.
+#
+# sqrt_price_next can be None. When it is None, that implies that the next sqrt
+# price must be calculated. In such a case, next sqrt price depends on the
+# remaining amount of token in to be swapped. This occurs for the last sqrt price
+# range in the collection of sqrt price ranges that represent a swap.
+#
+# For example, I might have a swap of 100 ETH in from 5000 to 5001 with liquidity
+# X, from 5001 to 5002 with liquidity Y,
+# and from 5002 to UNKNOWN with liquidity Z. In this case, the UNKWNOWN
```
czarcas7ic, line 17:
```suggestion
# and from 5002 to UNKNOWN with liquidity Z. In this case, the UNKNOWN
```

x/concentrated-liquidity/fees.go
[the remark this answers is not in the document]
czarcas7ic: Mmm, I feel like the way it currently is makes sense. The currentSqrtPrice will never exceed the sqrtPriceLimit AFAIK and when we hit it they are identical, so seems like this would be a logical boolean trigger for this.

x/concentrated-liquidity/swaps.go
```diff
@@ -20,6 +20,22 @@ type SwapState struct {
...
+}
+
+// updateFeeGrowthGlobal updates the swap state's fee growth global per unit of liquidity
+// when liquidity is positive.
+//
+// If the liquidity is zero, this is a no-op. This case may occur when there is no liquidity
+// between the ticks.This is possible when there are only 2 positions with no overlapping ranges.
+// As a result, the range from the end of position one to the beginning of position
+// two has no liquidity and can be skipped.
+// TODO: test
```
czarcas7ic, line 33: Is there an issue for this?

x/concentrated-liquidity/swaps.go
```diff
@@ -301,19 +318,25 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 			return writeCtx, sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, fmt.Errorf("could not convert next tick (%v) to nextSqrtPrice", nextTick)
 		}
 
+		// Charge fee
+		feeOnAmountRemainingIn := swapState.amountSpecifiedRemaining.Mul(swapFee)
+
```
czarcas7ic, line 323: This is probably right, but im not quite understanding the flow on first pass. 

We charge a fee, then subtract that fee from the amount remaining on the swap step (this makes sense to me)

What I dont understand is why we run a seperate function `computeFeeChargePerSwapStep` to calculate the `feeChargeTotal` for the global accumulator. Why would it not just be taken from `feeOnAmountRemainingIn`?

(Once again, I am sure this is right, but on first pass its not clear why)

x/concentrated-liquidity/swaps.go
p0mvn: yes ser: https://github.com/osmosis-labs/osmosis/issues/4086

x/concentrated-liquidity/swaps.go
p0mvn: Let's define what `swapState.amountSpecifiedRemaining` is.

It is the amount of token in remaining over all swap steps.

After performing the current swap step, the following cases are possible:
1.  we consumed all amount remaining
   * this is what we charge here: https://github.com/osmosis-labs/osmosis/blob/3df20182cae965f4c7d621ab947f4cee681becc1/x/concentrated-liquidity/fees.go#L292-L295
   * this essentially implies the difference between the original amount remaining and the one actually consumed. The difference between them is the fee
2. we did not consume all amount remaining
   * then we charge this: https://github.com/osmosis-labs/osmosis/blob/3df20182cae965f4c7d621ab947f4cee681becc1/x/concentrated-liquidity/fees.go#L290
   * this is the fee charged on the amount actually consumed during a swap step
3. Price impact protection makes us exit before consuming all amount remaining
* then we charge this (similar to option 2): https://github.com/osmosis-labs/osmosis/blob/3df20182cae965f4c7d621ab947f4cee681becc1/x/concentrated-liquidity/fees.go#L290
* the swap fee on the amount in actually consumed before price impact protection got trigerred

x/concentrated-liquidity/fees.go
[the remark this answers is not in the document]
p0mvn: Also, here's the breakdown of this logic:
https://github.com/osmosis-labs/osmosis/pull/4097#discussion_r1087470350

x/concentrated-liquidity/swaps.go
```diff
@@ -342,9 +365,13 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
...
 	}
 
+	if err := k.chargeFee(ctx, poolId, sdk.NewDecCoinFromDec(tokenInMin.Denom, swapState.feeGrowthGlobal)); err != nil {
+		return writeCtx, sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
+	}
+
 	// coin amounts require int values
 	// round amountIn up to avoid under charging
-	amt0 := tokenAmountInAfterFee.Sub(swapState.amountSpecifiedRemaining).RoundInt()
+	amt0 := tokenAmountInSpecified.Sub(swapState.amountSpecifiedRemaining).RoundInt()
```
AlpinYukseloglu, line 374: Just to clarify, these two are equivalent since we charge fees stepwise in `amountSpecifiedRemaining` instead of upfront in `tokenAmountInAfterFee` right?

AlpinYukseloglu (commented):
Overall LGTM except one comment about `chargeFee` making the calc function mutative – happy to approve after that is addressed. Also left a few other minor comments

x/concentrated-liquidity/swaps.go
```diff
@@ -342,9 +365,13 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 		}
 	}
 
+	if err := k.chargeFee(ctx, poolId, sdk.NewDecCoinFromDec(tokenInMin.Denom, swapState.feeGrowthGlobal)); err != nil {
```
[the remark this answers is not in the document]
mattverse: this doesn't since we are using cache ctx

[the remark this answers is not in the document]
p0mvn: That's correct, more context on how we actually make this be written to state:

https://github.com/osmosis-labs/osmosis/blob/4d9b21299a8e2942919a17d3a1f25335c36cd964/x/concentrated-liquidity/swaps.go#L146-L149

x/concentrated-liquidity/swaps.go
p0mvn: Yes, this should be just a name change

x/concentrated-liquidity/fees.go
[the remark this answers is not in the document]
p0mvn: @AlpinYukseloglu it should be equal here since we check for GT  prior in the execution flow and the swap all the way until `sqrtPriceLimit`:

https://github.com/osmosis-labs/osmosis/blob/4d9b21299a8e2942919a17d3a1f25335c36cd964/x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go#L34-L36

If price impact was triggered, `currentSqrtPrice` must equal `sqrtPriceLimit` here

x/concentrated-liquidity/swaps.go
```diff
@@ -301,19 +318,25 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 			return writeCtx, sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, fmt.Errorf("could not convert next tick (%v) to nextSqrtPrice", nextTick)
 		}
 
+		// Charge fee
+		feeOnAmountRemainingIn := swapState.amountSpecifiedRemaining.Mul(swapFee)
```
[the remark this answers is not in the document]
p0mvn: Done.

Rounding up at 10^-18 is done by doing `MulTruncate` and adding `SmallestDec`

x/concentrated-liquidity/swaps.go
mattverse: Awesome! This is something I did not previously know clearly before! Would we be able to document this context somewhere?

mattverse (approved):
Thanks for replying to all the comments! LGTM 🌮

x/concentrated-liquidity/swaps.go
czarcas7ic: This make sense to me, thanks! If this exact writeup could get documented somewhere that would be awesome

czarcas7ic (approved):
Approving since my only blocking question was answered, nicely done!

x/concentrated-liquidity/swaps.go
p0mvn: Docs added to `architecture.md`. The fee docs need to be refactored - quite outdated. I will come back to it after the swap in the other direction is done

Issue #4089, opened 2023-01-24
test(CL Fees): functional tests for CL fees


p0mvn:
## Background

We need functional tests for CL fees.

## Suggested Design

This should be a Go test withing a CL module that does the following:
- creates a pool with several positions from different accounts at various ranges. Maybe 1-3 positions from 3-5 accounts.
  * full range position
  * narrow range position
  * overlapping ranges
  * consecutive ranges
- performs 1-2 swaps from 3 accounts
- attempts to claim fees
- asserts that the right amount is claimed for each account

## Acceptance Criteria
- functional test added
- the correct amount claimed for each account

RusAkh:
https://github.com/osmosis-labs/osmosis/issues/4085 shall I close this then?

p0mvn:
> #4085 shall I close this then?

This is separate. This issue is related to testing the calculations at a lower level of abstraction while making sure that we don't withdraw more fees than needed.

In e2e terst, we want to do something similar but we don't have to make assertions about math being correct. It's more so - making sure nothing breaks and the chain keeps running

p0mvn:
I think we need to revisit our functional tests to determine why #4423 was uncaught.

PR #4108, opened 2023-01-26
(CL Test): Add bench test for crossTick Swapping

mattverse:
Closes: #XXX

We want to bench test the computational resource it takes to perform swaps that includes N amount of cross Ticks. 
This PR includes the small, medium and large bench test for the N amount of cross tick swapping.

Also added a small helper method in the keeper_test.go that initialzes four different positions that would later help us with the functional testing(cref #4089 )

## Brief Changelog
Add bench test for swaps with N times of crossing ticks.

## Testing and Verifying
This PR adds tests 


x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,104 @@
...
+	clmodel "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/model"
+)
+
+// benchmarkCrossTick performs bench test on the amount of computational resources it takes to cross ticks N times.
+// We achieve the goal of doing so initializing N amount of positions each with 1 * denom0, 1 * denom1.
+// This way performing a swap of N tokens would cause us to cross N amount of ticks, being able to bench `crossTick`.
+func benchmarkCrossTick(numCrossTick int, b *testing.B) {
+	b.StopTimer()
+
+	blockStartTime := time.Now().UTC()
```
p0mvn, line 26: let's not use `time.Now()`. Instead, set it to some deterministic value please

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,104 @@
...
+			fmt.Println(err.Error())
+			b.FailNow()
+		}
+	}
+
+	// test swapping and start bench testing
+	b.StartTimer()
+	tokenIn := sdk.NewCoin("usdc", sdk.NewInt(int64(numCrossTick)))
+	simapp.FundAccount(app.BankKeeper, ctx, addr, sdk.NewCoins(tokenIn))
+	_, _, currentTick, _, _, err = app.ConcentratedLiquidityKeeper.SwapOutAmtGivenIn(
```
p0mvn, line 79: For the benchmark to function correctly, we need to loop from i  [0, b.N)

ref: https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

p0mvn (commented):
Requesting changes for the looping part of benchmark.

Do you mind also submitting a [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) summary once it's ready please?

We have some docs here: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat

x/concentrated-liquidity/bench_test.go
mattverse: I reckned the current design would properly fit under doing a bench test because we're doing swaps that includes doing crossTick operation. i [0, b.N) 

Would it rather be more suitable if we did swap N times?

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,104 @@
...
+		apptesting.DefaultExponentAtPriceOne,
+		sdk.ZeroDec(),
+	))
+
+	// first we need to initialize n amount of ticks via adding position
+	// the goal is to initialize ticks with the minimum amount of liquidity(1 coins each) so that we can easily test crossing ticks
+	cl, err := app.ConcentratedLiquidityKeeper.GetPoolById(ctx, poolID)
+	currentTick := cl.GetCurrentTick()
+	startingTick := sdkrand.RandIntBetween(r, int(currentTick.Int64()), 5000)
+	spaceBetweenTicks := 100
```
czarcas7ic, line 49: So we are only ever testing tickSpacing of 1, but we are just spacing out the intervals in which we add liquidity to a tick? 

Unsure if this is ideal, but maybe it is sufficient for a benchmark. Would like other's thoughts

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// iterate over the amount of ticks we want to create
+	for i := 0; i < numCrossTick; i++ {
+		// fund 1 eth and 1 usdc
+		coins := sdk.NewCoins(sdk.NewCoin(apptesting.ETH, sdk.OneInt()), sdk.NewCoin(apptesting.USDC, sdk.OneInt()))
```
czarcas7ic, line 54: Since we are providing liquidity above the current price shouldn't we only need to fund one of the two denoms?

x/concentrated-liquidity/bench_test.go
czarcas7ic: From reading the link, it seems b.N is needed so the benchmark can run it as many times as it needs to get a proper average right?

ValarDragon: Depends what were benchmarking. We could imagine a two argument function for benchmarking, where its like "Benchmark swaps that cross `K` ticks. And then a top level benchmark that just runs it with k set to some standard value (e.g. "100")

In that mode, we'd do a swap b.N times

x/concentrated-liquidity/bench_test.go
mattverse: > but we are just spacing out the intervals in which we add liquidity to a tick?

wdym?

x/concentrated-liquidity/bench_test.go
mattverse: But then we should have liquidity for both sides or else we wouldn't have the ticks initialized properly to fulfill N times of swaps

x/concentrated-liquidity/bench_test.go
mattverse: Changed!

mattverse:
Updated so that we run swaps b.N times, 

The benchmarks are as the following:
BenchmarkSwapWithCrossTickTiny-8              12          99513264 ns/op
BenchmarkSwapWithCrossTickMedium-8            10         104463783 ns/op
BenchmarkSwapWithCrossTickLarge-8             10         106230875 ns/op

x/concentrated-liquidity/bench_test.go
mattverse: Going with the decision to run swaps b.N times

x/concentrated-liquidity/bench_test.go
czarcas7ic: spaceBetweenTicks is set to 100 here. tickSpacing is only ever set to 1 here. Im saying that we are never testing a tickSpacing value greater than 1 and instead are just utilizing spaceBetweenTicks which would potentially not act the same way.

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+
+	"github.com/osmosis-labs/osmosis/v14/app"
+	"github.com/osmosis-labs/osmosis/v14/app/apptesting"
+	sdkrand "github.com/osmosis-labs/osmosis/v14/simulation/simtypes/random"
+	clmodel "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/model"
+)
+
+// benchmarkCrossTick performs bench test on the amount of computational resources it takes to cross ticks N times.
+// We achieve the goal of doing so initializing N amount of positions each with 1 * denom0, 1 * denom1.
```
czarcas7ic, line 21:
```suggestion
// We achieve this goal by initializing N amount of positions, each with 1 * denom0, 1 * denom1.
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+// This way performing a swap of N tokens would cause us to cross N amount of ticks, being able to bench `crossTick`.
+func benchmarkCrossTick(numCrossTick int, b *testing.B) {
+	// b.StopTimer()
+
```
czarcas7ic, line 25:
```suggestion
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+	blockStartTime := time.Unix(2, 0)
+	app, cleanupFn := app.SetupTestingAppWithLevelDb(false)
+	defer cleanupFn()
+	ctx := app.BaseApp.NewContext(false, tmproto.Header{Height: 1, ChainID: "osmosis-1", Time: blockStartTime})
+	r := rand.New(rand.NewSource(10))
+	addr := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
+
+	// first create pool
+	_ = simapp.FundAccount(app.BankKeeper, ctx, addr, sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(1000000000))))
+	poolID, _ := app.PoolManagerKeeper.CreatePool(ctx, clmodel.NewMsgCreateConcentratedPool(
```
czarcas7ic, line 35: shouldnt we err check here?

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+		apptesting.ETH,
+		apptesting.USDC,
+		apptesting.DefaultTickSpacing,
+		apptesting.DefaultExponentAtPriceOne,
+		sdk.ZeroDec(),
+	))
+
+	// first we need to initialize n amount of ticks via adding position
+	// the goal is to initialize ticks with the minimum amount of liquidity(1 coins each) so that we can easily test crossing ticks
+	cl, err := app.ConcentratedLiquidityKeeper.GetPoolById(ctx, poolID)
```
czarcas7ic, line 46: same q about err check here

x/concentrated-liquidity/bench_test.go
czarcas7ic: Still have this question about spaceBetweenTicks

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+func BenchmarkSwapWithCrossTickMedium(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		benchmarkCrossTick(10, b)
+	}
+}
+func BenchmarkSwapWithCrossTickLarge(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		benchmarkCrossTick(10, b)
+	}
+}
```
czarcas7ic, line 109: These all have the save value of 10

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+// 	benchmarkCrossTick(10, b)
+// }
+
+// func BenchmarkCrossTickMedium(b *testing.B) {
+// 	benchmarkCrossTick(100, b)
+// }
+
+// func BenchmarkCrossTickLarge(b *testing.B) {
+// 	benchmarkCrossTick(1000, b)
+// }
```
czarcas7ic, line 121:
```suggestion
```

PR #4109, opened 2023-01-26
feat[CL]: base freeze API

czarcas7ic:

As part of the MVP for incentives, users must be able to freeze their concentrated liquidity positions for a certain period of time. This PR introduces a `frozenUntil` input throughout the CL module that makes positions with different freeze time unique and non fungible between one another.

For the remainder of this PR, I will be referring to FullPositions and Positions. By this, I mean:
* FullPosition: defined by poolID, owner, tick range, and frozenUntil timestamp.
* Position: defined by poolID, owner, and tick range.

## Brief Changelog

- Adds `frozenUntil` timestamp to the Position object proto
- Adds `frozenUntil` timestamps to the CreatePosition and WithdrawPosition proto messages
- Adds helper function `getAllPositionsWithVaryingFreezeTimes`. 
  - This allows a user to call on their positions without knowing their frozenUntil time.
- While the feeAccumulator is set for each unique FullPosition, claim fees is called on Position (which can describe a single position or slice of positions).
  - This prevents us from having to call claimFees for every single FullPosition 
- Updates all respective tests to include frozenUntil as part of the test


x/concentrated-liquidity/fees.go
```diff
@@ -189,46 +190,62 @@ func (k Keeper) collectFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress
...
 
+	// Once we have iterated through all the positions, we do a single bank send from the pool to the owner.
 	pool, err := k.getPoolById(ctx, poolId)
 	if err != nil {
 		return sdk.Coins{}, err
 	}
-
-	// distribute the fees from pool to the position owner.
-	if err := k.bankKeeper.SendCoins(ctx, pool.GetAddress(), owner, feesClaimed); err != nil {
+	if err := k.bankKeeper.SendCoins(ctx, pool.GetAddress(), owner, totalFeesClaimed); err != nil {
```
czarcas7ic, line 245: Note to reviewer: I decided to do a single bank send after iterating over all positions and claiming the rewards in order to reduce the overhead. Please lmk if you agree / can think of a way to do this with even less overhead.

x/concentrated-liquidity/fees_test.go
```diff
@@ -603,22 +620,23 @@ func (s *KeeperTestSuite) TestCollectFees() {
...
+			owner:       ownerWithValidPosition,
+			lowerTick:   0,
+			upperTick:   1,
+			frozenUntil: defaultFrozenUntil,
 
 			currentTick: 2,
 
 			isInvalidPoolIdGiven: true,
-			expectedError:        cltypes.PoolNotFoundError{PoolId: 2},
+			expectedError:        accum.AccumDoesNotExistError{AccumName: "fee/2"},
```
czarcas7ic, line 639: Note to reviewer: `Require.ErrorAs()` was not picking this up correctly. I find this happens a lot with that and the only one that actually picks up Errors correctly is `ErrorContains`. Basically the error that was occurring was actually that the accumulator did not exist (since that pool was never created), not a `PoolNotFoundError`

x/concentrated-liquidity/fees_test.go
```diff
@@ -651,7 +670,12 @@ func (s *KeeperTestSuite) TestCollectFees() {
 			clKeeper := s.App.ConcentratedLiquidityKeeper
 			ctx := s.Ctx
 
-			s.initializeFeeAccumulatorPositionWithLiquidity(ctx, validPoolId, tc.owner, tc.lowerTick, tc.upperTick, tc.initialLiquidity)
+			// Position needs to be set in the store before calling CollectFees.
+			// This is because we iterate through all positions instead of having to call each position with separate frozenUntil times.
+			position := model.Position{Liquidity: tc.initialLiquidity, FrozenUntil: tc.frozenUntil}
+			s.App.ConcentratedLiquidityKeeper.SetPosition(ctx, validPoolId, ownerWithValidPosition, tc.lowerTick, tc.upperTick, &position, tc.frozenUntil)
```
czarcas7ic, line 676: Note to reviewer: Without setting the position, we don't call initializePosition and therefore we wont have any positions in store. Without any positions in store, when we go to call `CollectFees` it wont have any positions to iterate over. 

There are no problems here, just wanted to explain why setting the position was suddenly necessary (now that we iterate over positions from the store when collecting fees)

AlpinYukseloglu (approved):
Core changes LGTM

x/concentrated-liquidity/types/keys.go
```diff
@@ -62,14 +66,17 @@ func KeyTickPrefix(poolId uint64) []byte {
 	return key
 }
 
+// KeyFullPosition uses pool Id, owner, lower tick, upper tick, and frozenUntil for keys
+func KeyFullPosition(poolId uint64, addr sdk.AccAddress, lowerTick, upperTick int64, frozenUntil time.Time) []byte {
+	frozenUntilKey := osmoutils.FormatTimeString(frozenUntil)
+	addrKey := address.MustLengthPrefix(addr.Bytes())
+	return []byte(fmt.Sprintf("%s%s%s%s%d%s%d%s%d%s%s", PositionPrefix, KeySeparator, addrKey, KeySeparator, poolId, KeySeparator, lowerTick, KeySeparator, upperTick, KeySeparator, frozenUntilKey))
```
mattverse, line 73: Is there future needs to have these values separated and parsed?

x/concentrated-liquidity/fees.go
```diff
@@ -189,6 +189,17 @@ func (k Keeper) collectFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress
...
+	// We retrieve all of them and claim fees for each one.
+	positions, err := k.getAllPositionsWithVaryingFreezeTimes(ctx, poolId, owner, lowerTick, upperTick)
+	if err != nil {
+		return sdk.Coins{}, err
+	}
+
+	if len(positions) == 0 {
+		return sdk.Coins{}, cltypes.PositionNotFoundError{PoolId: poolId, LowerTick: lowerTick, UpperTick: upperTick}
+	}
+
```
mattverse, line 202: So I reckon that these lines are for checking whether we have a valid position on the owner + lower tick + upper tick combination, do we not check this anyways in the code line below with the feeAccum.hasPosition?

x/concentrated-liquidity/fees_test.go
mattverse: Ditto for the question above, why do we need to set position in the first place when we have the has position check from collectFees?

mattverse (commented):
Overall LGTM, left some minor comments!

x/concentrated-liquidity/fees.go
czarcas7ic: Hey, great catch! 

So previously, this was needed because we were iterating over each FullPosition since I made it so that each FullPosition had a fee accumulator. Then, after our call, I realized we only needed a fee accumulator for Position, which then makes this check and the SetPosition requirement in the test not required.

x/concentrated-liquidity/types/keys.go
czarcas7ic: Does having the separator make some kind of meaningful overhead? It seems better to have it then to not have it, but I could be wrong (at least that was the conclusion I came to when I brought it up with Roman)

x/concentrated-liquidity/fees.go
czarcas7ic: Addressed here https://github.com/osmosis-labs/osmosis/pull/4109/commits/838c2ab9855677769da8374c5abbf1dcca93537d

x/concentrated-liquidity/fees_test.go
czarcas7ic: See this comment https://github.com/osmosis-labs/osmosis/pull/4109#discussion_r1089160947

Good catch :)

czarcas7ic:
Merging to unblock @AlpinYukseloglu

p0mvn (commented):
@czarcas7ic would it be possible to document the "full position" vs "position" context in a future PR please?

czarcas7ic:
@p0mvn added these definitions to the currently open PR here https://github.com/osmosis-labs/osmosis/pull/4129/commits/4c65e9830c94cfa06136725261e3a661a8693115

Issue #4099, opened 2023-01-25
[e2e][CL]: Collect Fees


pysel:
## Background

We need to add fee collection to `CL`'s e2e test.

## Suggested Design

Collect Fees in e2e test

## Acceptance Criteria

Design implmented

PR #4119, opened 2023-01-27
[e2e][CL]: Collect Fees

pysel:
Closes: #4099

This PR introduces:
* `CollectFees` method to `NodeConfig`
* Adds fee collection functionality to e2e test

## Testing and Verifying

e2e test

no

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -313,7 +313,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Track balances for address1 position1
 	addr1BalancesBefore := s.addrBalance(node, address1)
-	node.CollectFees(address1, "[-1200]", "400", poolID)
+	node.CollectFees(address1, fmt.Sprint(positionsAddress1[0].Position.PositionId))
 	addr1BalancesAfter := s.addrBalance(node, address1)
 
 	// Assert that the balance changed only for tokenIn (uosmo)
@@ -424,7 +424,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Track balances for address1 position1
 	addr1BalancesBefore = s.addrBalance(node, address1)
-	node.CollectFees(address1, "[-1200]", "400", poolID)
+	node.CollectFees(address1, fmt.Sprint(positionsAddress1[0].Position.PositionId))
 	addr1BalancesAfter = s.addrBalance(node, address1)
 
 	// Assert that the balance changed only for tokenIn (uosmo)
@@ -449,7 +449,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Track balance off address3 position2: check that position that has not been kicked out earned full rewards
 	addr3BalancesBefore := s.addrBalance(node, address3)
-	node.CollectFees(address3, fmt.Sprintf("[%d]", minTick), "1400", poolID)
+	node.CollectFees(address3, fmt.Sprint(positionsAddress3[1].Position.PositionId))
 	addr3BalancesAfter := s.addrBalance(node, address3)
 
 	// Calculate uncollected fees for address3 position2 earned from Swap 1
@@ -528,7 +528,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 		feeGrowthInsideAddress1Position1Last = feeGrowthGlobal_Swap1.Add(feeCharge_Swap2_Step1)
 	)
 	// Collect fees for address1 position1 to avoid overhead computations (swap2 already asserted fees are aggregated correctly from multiple swaps)
-	node.CollectFees(address1, "[-400]", "1200", poolID)
+	node.CollectFees(address1, fmt.Sprint(positionsAddress1[0].Position.PositionId))
 
 	// Perform a swap
 	node.SwapExactAmountIn(uionIn_Swap3, outMinAmt, fmt.Sprintf("%d", poolID), denom1, initialization.ValidatorWalletName)
@@ -541,7 +541,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Track balance of address1
 	addr1BalancesBefore = s.addrBalance(node, address1)
-	node.CollectFees(address1, "[-1200]", "400", poolID)
+	node.CollectFees(address1, fmt.Sprint(positionsAddress1[0].Position.PositionId))
 	addr1BalancesAfter = s.addrBalance(node, address1)
 
 	// Assert that the balance changed only for tokenIn (uion)
@@ -587,7 +587,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Track balance of address3
 	addr3BalancesBefore = s.addrBalance(node, address3)
-	node.CollectFees(address3, fmt.Sprintf("[%d]", minTick), "1400", poolID)
+	node.CollectFees(address3, fmt.Sprint(positionsAddress3[1].Position.PositionId))
 	addr3BalancesAfter = s.addrBalance(node, address3)
 
 	// Assert that the balance changed only for tokenIn (uion)
@@ -626,9 +626,9 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Assert that positions, which were not included in swaps, were not affected
 
-	// Address3 Position2: [-1600; -200]
+	// Address3 Position1: [-1600; -200]
 	addr3BalancesBefore = s.addrBalance(node, address3)
-	node.CollectFees(address3, "[-1600]", "[-200]", poolID)
+	node.CollectFees(address3, fmt.Sprint(positionsAddress3[0].Position.PositionId))
 	addr3BalancesAfter = s.addrBalance(node, address3)
 
 	// Assert that balances did not change for any token
@@ -636,7 +636,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Address2's only position: [2200; 3420]
 	addr2BalancesBefore := s.addrBalance(node, address2)
-	node.CollectFees(address2, "2200", fmt.Sprintf("%d", maxTick), poolID)
+	node.CollectFees(address2, fmt.Sprint(positionsAddress2[0].Position.PositionId))
 	addr2BalancesAfter := s.addrBalance(node, address2)
 
 	// Assert the balances did not change for every token
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -226,7 +226,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// Create 2 positions for address1: overlap together, overlap with 2 address3 positions
 	node.CreateConcentratedPosition(address1, "[-1200]", "400", fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
-	node.CreateConcentratedPosition(address1, "[-400]", "400", fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
+	node.CreateConcentratedPosition(address1, "[-400]", "1200", fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
 
 	// Create 1 position for address2: does not overlap with anything, ends at maximum
 	node.CreateConcentratedPosition(address2, "2200", fmt.Sprintf("%d", maxTick), fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -661,7 +661,7 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 
 	// address2: check removing some amount of liquidity
 	address2position1liquidityBefore := positionsAddress2[0].Position.Liquidity
-	node.WithdrawPosition(address2, defaultLiquidityRemoval, positionsAddress1[0].Position.PositionId)
+	node.WithdrawPosition(address2, defaultLiquidityRemoval, positionsAddress2[0].Position.PositionId)
 	// assert
 	positionsAddress2 = node.QueryConcentratedPositions(address2)
 	s.Require().Equal(address2position1liquidityBefore, positionsAddress2[0].Position.Liquidity.Add(sdk.MustNewDecFromStr(defaultLiquidityRemoval)))
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -225,7 +225,6 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 	address3 := node.CreateWalletAndFund("addr3", fundTokens)
 
 	// Create 2 positions for address1: overlap together, overlap with 2 address3 positions
-
 	node.CreateConcentratedPosition(address1, "[-1200]", "400", fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
 	node.CreateConcentratedPosition(address1, "[-400]", "400", fmt.Sprintf("10000000%s", denom0), fmt.Sprintf("10000000%s", denom1), 0, 0, poolID)
 
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -19,8 +19,6 @@ import (
 
 	ibchookskeeper "github.com/osmosis-labs/osmosis/x/ibc-hooks/keeper"
 
-	"github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/model"
-	"github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/types"
 	ibcratelimittypes "github.com/osmosis-labs/osmosis/v15/x/ibc-rate-limit/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -191,26 +189,24 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 	s.Require().NoError(err)
 
 	var (
-		denom0                    string = "uion"
-		denom1                    string = "uosmo"
-		tickSpacing               uint64 = 1
-		exponentAtPriceOne        int64  = -1
-		swapFee                          = "0.01"
+		denom0             string  = "uion"
+		denom1             string  = "uosmo"
+		tickSpacing        uint64  = 1
+		exponentAtPriceOne int64   = -1
+		swapFee                    = "0.01"
+		swapFeeDec         sdk.Dec = sdk.MustNewDecFromStr("0.01")
 	)
 
-	// helpers
-	var (
-		updatedPool = func(poolId uint64) types.ConcentratedPoolExtension {
-			concentratedPool, err := node.QueryConcentratedPool(poolId)
-			s.Require().NoError(err)
-			return concentratedPool
-		}
-	)
 	poolID := node.CreateConcentratedPool(initialization.ValidatorWalletName, denom0, denom1, tickSpacing, exponentAtPriceOne, swapFee)
 
-	concentratedPool := updatedPool(poolID)
+	concentratedPool := s.updatedPool(node, poolID)
 
-	// assert contents of the pool are valid
+	// Sanity check that pool initialized with valid parameters (the ones that we haven't explicitly specified)
+	s.Require().Equal(concentratedPool.GetCurrentTick(), sdk.ZeroInt())
+	s.Require().Equal(concentratedPool.GetCurrentSqrtPrice(), sdk.ZeroDec())
+	s.Require().Equal(concentratedPool.GetLiquidity(), sdk.ZeroDec())
+
+	// Assert contents of the pool are valid (that we explicitly specified)
 	s.Require().Equal(concentratedPool.GetId(), poolID)
 	s.Require().Equal(concentratedPool.GetToken0(), denom0)
 	s.Require().Equal(concentratedPool.GetToken1(), denom1)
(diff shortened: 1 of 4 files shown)


tests/e2e/configurer/chain/queries.go
```diff
@@ -134,6 +135,24 @@ func (n *NodeConfig) QueryBalances(address string) (sdk.Coins, error) {
 	return balancesResp.GetBalances(), nil
 }
 
+func (n *NodeConfig) QueryEstimateSwapExactAmountIn(sender string, poolId uint64, denom, swapRouteDenoms, swapRoutePoolIds string, chainId string) (sdk.Int, error) {
```
pysel, line 138: This query actually should not be here, but I implemented it at some point because I thought I needed it for e2e, I can leave it in this PR or create another one just so the work is not lost

p0mvn:
@pysel  the fixes to swap & fee bugs are merged now. Please feel free to continue on this test, would love to see this getting in soon

tests/e2e/e2e_test.go
```diff
@@ -186,65 +186,167 @@ func (s *IntegrationTestSuite) CheckBalance(node *chain.NodeConfig, addr, denom
 	)
 }
 
-func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
+func (s *IntegrationTestSuite) TestAAAConcentratedLiquidity() {
```
p0mvn, line 189:
```suggestion
func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
```

tests/e2e/e2e_test.go
```diff
@@ -186,65 +186,167 @@ func (s *IntegrationTestSuite) CheckBalance(node *chain.NodeConfig, addr, denom
...
-		tickSpacing               uint64 = 1
-		precisionFactorAtPriceOne int64  = -1
-		frozenUntil               int64  = time.Unix(86400, 0).Unix()
-		swapFee                          = "0.01"
+		// Pool parameters
+		denom0                    string  = "uion"
+		denom1                    string  = "uosmo"
+		tickSpacing               uint64  = 1
+		precisionFactorAtPriceOne int64   = -1
+		frozenUntil               int64   = time.Unix(86400, 0).Unix()
```
p0mvn, line 200: I think we don't use frozen until anymore

tests/e2e/e2e_test.go
```diff
@@ -186,65 +186,167 @@ func (s *IntegrationTestSuite) CheckBalance(node *chain.NodeConfig, addr, denom
...
+			if assertUosmo {
+				s.Require().True(balancesAfter.AmountOf("uosmo").Equal(balancesBefore.AmountOf("uosmo")))
+			}
+		}
+
+		// calculateFeeGrowthGlobal calculates fee growth global per unit of virtual liquidity based on swap parameters:
+		// amountIn - amount being swapped
+		// swapFee - pool's swap fee
+		// poolLiquidity - current pool liquidity
+		calculateFeeGrowthGlobal = func(amountIn, swapFee, poolLiquidity sdk.Dec) sdk.Dec {
```
p0mvn, line 245:
```suggestion
		calculateFeeGrowthGlobal = func(amountIn, swapFee, rangeLiquidity sdk.Dec) sdk.Dec {
```

tests/e2e/e2e_test.go
```diff
@@ -186,65 +186,167 @@ func (s *IntegrationTestSuite) CheckBalance(node *chain.NodeConfig, addr, denom
...
+			// Formula for finding uncollected fees in time range [t1; t2]:
+			// F_u = position_liquidity * (fee_growth_inside_t2 - fee_growth_inside_t1).
+			feesUncollected := positionLiquidity.Mul(feeGrowthInside.Sub(feeGrowthInsideLast))
+
+			return feesUncollected.TruncateDec()
+		}
+
+		// calculateSqrtPriceAtTick sqrtPrice at tickIndex
+		// View x/concentrated-liquidity/README.md -> ConcentratedLiquidity/Architecture/Ticks/Formulas for more details
+		calculateSqrtPriceAtTick = func(tickIndex, exponentAtPriceOne sdk.Int) sdk.Dec {
```
p0mvn, line 283: Can we import this helper form concentrated liquidity? If it's private, I propose exporting it and moving to `types`

tests/e2e/e2e_test.go
```diff
@@ -256,29 +358,410 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+		uosmoInDec_Swap1 = sdk.NewDec(3465198) // Minus swap fee = 3430546.02
+		uosmoIn_Swap1    = fmt.Sprintf("%suosmo", uosmoInDec_Swap1.String())
+	)
+	// Perform swap (not crossing initialized ticks)
+	node.SwapExactAmountIn(uosmoIn_Swap1, outMinAmt, fmt.Sprintf("%d", poolID), denom0, initialization.ValidatorWalletName)
+	// Calculate and track global fee growth for swap 1
+	feeGrowthGlobal.AddMut(calculateFeeGrowthGlobal(uosmoInDec_Swap1, swapFeeDec, concentratedPool.GetLiquidity()))
+
+	// Let the chain pick up the changes:
+	chainA.WaitForNumHeights(2)
```
p0mvn, line 405: Is this necessary? Swaps are executed with `-b=block` so changes should be propagated immediately. Let's remove delays that we don't need to avoid increasing CI times

tests/e2e/e2e_test.go
```diff
@@ -256,29 +358,410 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+		uosmoInDec_Swap2       = uosmoInDec_Swap2_NoFee.Quo(sdk.OneDec().Sub(swapFeeDec)).TruncateDec() // account for swap fee of 1%
+		uosmoIn_Swap2          = fmt.Sprintf("%suosmo", uosmoInDec_Swap2.String())
+
+		feeGrowthGlobal_Swap1 = feeGrowthGlobal.Clone()
+	)
+	// Perform a swap
+	node.SwapExactAmountIn(uosmoIn_Swap2, outMinAmt, fmt.Sprintf("%d", poolID), denom0, initialization.ValidatorWalletName)
+
+	// Let the chain pick up the changes:
+	chainA.WaitForNumHeights(2)
```
p0mvn, line 502: Do we need this?

tests/e2e/e2e_test.go
```diff
@@ -256,29 +358,410 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+		feeGrowthInsideAddress1Position1Last = feeGrowthGlobal_Swap1.Add(feeCharge_Swap2_Step1)
+	)
+	// Collect fees for address1 position1 to avoid overhead computations (swap2 already asserted fees are aggregated correctly from multiple swaps)
+	node.CollectFees(address1, "[-400]", "1200", poolID)
+
+	// Perform a swap
+	node.SwapExactAmountIn(uionIn_Swap3, outMinAmt, fmt.Sprintf("%d", poolID), denom1, initialization.ValidatorWalletName)
+
+	// Let the chain pick up the changes:
+	chainA.WaitForNumHeights(2)
```
p0mvn, line 650: Let's remove if possible

p0mvn (commented):
Great work. Some initial feedback, still need to review the calculations in-detail

go.work.sum
```diff
@@ -1,172 +1,2040 @@
+4d63.com/gochecknoglobals v0.1.0/go.mod h1:wfdC5ZjKSPr7CybKEcgJhUOgeAQW1+7WcyK8OvUilfo=
```
czarcas7ic, line 1: Copy this file from main and replace these changes (I think we need to add this to .gitignore, because I have accidentally almost commit this myself yesterday)

tests/e2e/e2e_test.go
```diff
@@ -192,59 +192,137 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+		// * amount of `e2e-default-feetoken` dropped by 1000 (default amount for fee per tx)
+		// * depending on `assertUosmo` and `assertUion` parameters, check that those balances have also not been changed
+		assertDefaultBalances = func(balancesBefore, balancesAfter sdk.Coins, assertUosmo, assertUion bool) {
+			s.Require().True(balancesAfter.AmountOf("stake").Equal(balancesBefore.AmountOf("stake")))
+			s.Require().True(balancesAfter.AmountOf("e2e-default-feetoken").Equal(balancesBefore.AmountOf("e2e-default-feetoken").Sub(defaultFeePerTx)))
+			if assertUion {
+				s.Require().True(balancesAfter.AmountOf("uion").Equal(balancesBefore.AmountOf("uion")))
+			}
+			if assertUosmo {
+				s.Require().True(balancesAfter.AmountOf("uosmo").Equal(balancesBefore.AmountOf("uosmo")))
```
czarcas7ic, line 234: nit: I would prefer a more verbose name like `assertUionBalanceIsConstant`, but just personal preference. Feel free to resolve with no further action.

tests/e2e/e2e_test.go
```diff
@@ -192,59 +192,137 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+
+		// Helper function for calculating uncollected fees since the time that feeGrowthInsideLast corresponds to
+		// positionLiquidity - current position liquidity
+		// feeGrowthBelow - fee growth below lower tick
+		// feeGrowthAbove - fee growth above upper tick
+		// feeGrowthInsideLast - amount of fee growth inside range at the time from which we want to calculate the amount of uncollected fees
+		// feeGrowthGlobal - variable for tracking global fee growth
+		calculateUncollectedFees = func(positionLiquidity, feeGrowthBelow, feeGrowthAbove, feeGrowthInsideLast sdk.Dec, feeGrowthGlobal sdk.Dec) sdk.Dec {
+			// Calculating fee growth inside range [-1200; 400]
+			feeGrowthInside := calculateFeeGrowthInside(feeGrowthGlobal, feeGrowthBelow, feeGrowthAbove)
```
czarcas7ic, line 268: nit: it feels like all these helper funcs (to include the ones above this hi-light) might be better off outside of the test itself. I think tests should be as sparse as possible to not detract from what it is testing, but just an opinion.

czarcas7ic (approved):
It feels like the CL e2e test could really use some logic abstraction to make it more understandable. With that being said, the logic seems right to me and don't want to block on making things neater. Would like an ACK from Roman and then we can merge. 

Thanks for the PR :)

tests/e2e/e2e_test.go
pysel: yeah, that will actually improve readability, thanks!

tests/e2e/e2e_test.go
pysel: hmm, makes sense, do you have any suggestions where I could move these to? i do not currently have any ideas, but this suggestion makes sense to me

tests/e2e/e2e_test.go
pysel: addressed, thanks again!

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	// * Uncollected fees from multiple swaps are correctly summed up and collected
+
+	// tickOffset is a tick index after the next initialized tick to which this swap needs to move the current price
+	tickOffset := sdk.NewInt(3)
+	sqrtPriceBeforeSwap = concentratedPool.GetCurrentSqrtPrice()
+	liquidityBeforeSwap = concentratedPool.GetLiquidity()
+	nextInitTick := sdk.NewInt(400) // address1 position1's upper tick
+
+	// Calculate sqrtPrice after and at the next initialized tick (upperTick of address1 position1 - 400)
+	// sqrtPriceAfterNextInitializedTick := calculateSqrtPriceAtTick(nextInitTick.Add(tickOffset), sdk.NewInt(precisionFactorAtPriceOne))
```
mattverse, line 444:
```suggestion
```

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	// deltaSqrtPriceAtNextInitializedTick = ΔsqrtP(400) - ΔsqrtP(currentTick)
+	deltaSqrtPriceAfterNextInitializedTick := sqrtPriceAfterNextInitializedTick.Sub(sqrtPriceAtNextInitializedTick)
+	deltaSqrtPriceAtNextInitializedTick := sqrtPriceAtNextInitializedTick.Sub(sqrtPriceBeforeSwap)
+
+	// Calculate the amount of osmo required to:
+	// * amountInToGetToTickAfterInitialized - move price from next initialized tick (400) to destination tick (400 + tickOffset)
+	// * amountInToGetToNextInitTick - move price from current tick to next initialized tick
+	// Formula is as follows:
+	// Δy = L * Δ(sqrtPrice)
+	amountInToGetToTickAfterInitialized := deltaSqrtPriceAfterNextInitializedTick.Mul(liquidityBeforeSwap.Sub(positionsAddress1[0].Liquidity))
```
mattverse, line 461: I think we should make all these calculations into functions, seeing that we are repeatedly doing this calculation + would have us with way better readability

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	// feeCharge = amountRemaining - amountOne
+	amountRemainingAfterStep1 := uosmoInDec_Swap2.Sub(amountInToGetToNextInitTick).Sub(feeCharge_Swap2_Step1)
+	feeCharge_Swap2_Step2 := amountRemainingAfterStep1.Sub(amountInToGetToTickAfterInitialized)
+
+	// per unit of virtual liquidity
+	feeCharge_Swap2_Step1.QuoMut(liquidityBeforeSwap)
+	feeCharge_Swap2_Step2.QuoMut(liquidityAfterSwap)
+
+	// Update feeGrowthGlobal
+	feeGrowthGlobal.AddMut(feeCharge_Swap2_Step1.Add(feeCharge_Swap2_Step2))
```
mattverse, line 507: what's the reason we manually do addition to fee growth global?

mattverse (commented):
The logic itself lgtm, I think we should definitely turn the frequently used calculations used in the method, since repeatitive .add .sub for each calculation defintiely takes up alot of unnecessary code lines

Also wondering if we have a way to turn the entire addition into a single function of test_cl_fees, since I feel like we are adding too much testing for a single method.

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+		outMinAmt       = "1"
+	)
+
+	// Swap 1
+	// Not crossing initialized ticks => performed in one swap step
+	// Swap affects 3 positions: both that address1 has and one of address3's positions
+	// Asserts that fees are correctly collected for non cross-tick swaps
+	var (
+		// Swap parameters
+		uosmoInDec_Swap1 = sdk.NewDec(3465198)
```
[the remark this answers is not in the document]
pysel: just a random number, the only constraint I had in mind when picking it was that it should not be too high to cross next initialized tick

tests/e2e/e2e_test.go
```diff
@@ -192,59 +192,137 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+			if assertUosmoBalanceIsConstant {
+				s.Require().True(balancesAfter.AmountOf("uosmo").Equal(balancesBefore.AmountOf("uosmo")))
+			}
+		}
+
+		// calculateFeeGrowthGlobal calculates fee growth global per unit of virtual liquidity based on swap parameters:
+		// amountIn - amount being swapped
+		// swapFee - pool's swap fee
+		// rangeLiquidity - current pool liquidity
+		calculateFeeGrowthGlobal = func(amountIn, swapFee, rangeLiquidity sdk.Dec) sdk.Dec {
```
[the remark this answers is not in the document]
pysel: I agree, that was initially called like that

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	concentratedPool = updatedPool(poolID)
+
+	liquidityAfterSwap := concentratedPool.GetLiquidity()
+	sqrtPriceAfterSwap := concentratedPool.GetCurrentSqrtPrice()
+
+	// Assert swaps don't change pool's liquidity amount
+	s.Require().Equal(liquidityAfterSwap, liquidityBeforeSwap)
+
+	// Assert current sqrt price
+	inAmountSubFee := uosmoInDec_Swap1.Mul(sdk.OneDec().Sub(swapFeeDec))
```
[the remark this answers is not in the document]
pysel: `inAmountSubFee` is the amount of tokens in actually being swapped. if we have a `subFee` = 1% and we are swapping, say, 100 uosmo, the amount of tokens that will be used throughout the swap is `100uosmo * (1 - 0.01) = 99uosmo`, the other 1uosmo will be consumed for fees.

I was wondering if the naming here is vague, do you think I shall change the variable name?

tests/e2e/e2e_test.go
```diff
@@ -256,29 +334,406 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	addr1BalancesBefore := addrBalance(address1)
+	node.CollectFees(address1, "[-1200]", "400", poolID)
+	addr1BalancesAfter := addrBalance(address1)
+
+	// Assert that the balance changed only for tokenIn (uosmo)
+	assertDefaultBalances(addr1BalancesBefore, addr1BalancesAfter, false, true)
+
+	// Assert Balances: Swap 1
+
+	// Calculate uncollected fees for address1 position1
```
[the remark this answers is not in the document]
pysel: [yes](https://github.com/osmosis-labs/osmosis/pull/4119/files/702a02bcae24040a6304111a07b84c260131f2ad#diff-7d1fc1e232b55677582b51e2623aae6d8c15b5f0960ac6100fc7d733cb795dacR421-R423), if that is what you mean

tests/e2e/e2e_test.go
pysel: I added this variable to track global fee growth after each swap, hence, at every swap I calculate and update this variable.

 Afaik, you cannot get global fee growth global by querying state or via some method on `ConcentratedPoolExtension` (which is called `concentratedPool` in this test), so I did not really see other options as to how I can track this variable other than manually calculating it on each swap

tests/e2e/e2e_test.go
pysel: It can be done, but only partially:

Swap2 and Swap3 use different formulas to calculate amounts needed:

Swap2 uses: `CalcAmount0Delta = liquidity * (sqrtPriceB - sqrtPriceA)) / (sqrtPriceB * sqrtPriceA`
Swap3 uses: `CalcAmount1Delta = liquidity * (sqrtPriceB - sqrtPriceA)`

I can make a function to calculate `liquidity * (sqrtPriceB - sqrtPriceA)` to use in both of these calculations, but do we want to do that?

p0mvn:
Hi @pysel . Can you please resolve merge conflicts here?

tests/e2e/e2e_test.go
```diff
@@ -193,93 +191,465 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	s.assertBalancesInvariants(addr3BalancesBefore, addr3BalancesAfter, true, true)
+
+	// Address2's only position: [2200; 3420]
+	addr2BalancesBefore := s.addrBalance(node, address2)
+	node.CollectFees(address2, "2200", fmt.Sprintf("%d", maxTick), poolID)
+	addr2BalancesAfter := s.addrBalance(node, address2)
+
+	// Assert the balances did not change for every token
+	s.assertBalancesInvariants(addr2BalancesBefore, addr2BalancesAfter, true, true)
+
```
mattverse, line 648: Just curious: Do you think it makes sense to add tests that try to collect fees when there are none or trying to collect fees from different ranges then provided positions?

mattverse (approved):
Nice work! LGTM

tests/e2e/e2e_test.go
```diff
@@ -193,93 +191,465 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	deltaSqrtPriceAfterNextInitializedTick := sqrtPriceAfterNextInitializedTick.Sub(sqrtPriceAtNextInitializedTick)
+	deltaSqrtPriceAtNextInitializedTick := sqrtPriceAtNextInitializedTick.Sub(sqrtPriceBeforeSwap)
+
+	// Calculate the amount of osmo required to:
+	// * amountInToGetToTickAfterInitialized - move price from next initialized tick (400) to destination tick (400 + tickOffset)
+	// * amountInToGetToNextInitTick - move price from current tick to next initialized tick
+	// Formula is as follows:
+	// Δy = L * Δ(sqrtPrice)
+	amountInToGetToTickAfterInitialized := deltaSqrtPriceAfterNextInitializedTick.Mul(liquidityBeforeSwap.Sub(positionsAddress1[0].Position.Liquidity))
+	amountInToGetToNextInitTick := deltaSqrtPriceAtNextInitializedTick.Mul(liquidityBeforeSwap)
```
stackman27, line 378: just wondering if we can put these calculation in helper function, since i've seen it elsewhere in code too. And it'll make it cleaner too :))

tests/e2e/e2e_test.go
```diff
@@ -193,93 +191,465 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
...
+	// Address2's only position: [2200; 3420]
+	addr2BalancesBefore := s.addrBalance(node, address2)
+	node.CollectFees(address2, "2200", fmt.Sprintf("%d", maxTick), poolID)
+	addr2BalancesAfter := s.addrBalance(node, address2)
+
+	// Assert the balances did not change for every token
+	s.assertBalancesInvariants(addr2BalancesBefore, addr2BalancesAfter, true, true)
+
+	// Withdraw Position
+
```
stackman27, line 650: just wondering: have we tested Swaps from Amount1 -> Amount0 or is this test solely focused on Amount0 -> Amount1 

```
CalcAmount0Delta = liquidity * ((sqrtPriceB - sqrtPriceA) / (sqrtPriceB * sqrtPriceA))
vs 
 calcAmount1Delta = liq * (sqrtPriceB - sqrtPriceA)

```

stackman27 (approved):
overall LGTM, just have few questions

x/concentrated-liquidity/export_internal.go
```diff
@@ -0,0 +1,11 @@
+package concentrated_liquidity
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/internal/math"
+)
+
+func TickToSqrtPrice(tickIndex, exponentAtPriceOne sdk.Int) (price sdk.Dec, err error) {
+	return math.TickToSqrtPrice(tickIndex, exponentAtPriceOne)
```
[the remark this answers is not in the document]
pysel: we have to export it cuz it is in `internal` package

tests/e2e/e2e_test.go
pysel: hmm, it is possible to put it into a helper function, but these calculations are calculated only once, so we might do that only for readability purposes, wdyt?

PS: swap 2 and swap 3 have similar calculations, but not same, so not possible to use the same function for both

tests/e2e/e2e_test.go
pysel: [yep](https://github.com/osmosis-labs/osmosis/pull/4119/files#diff-7d1fc1e232b55677582b51e2623aae6d8c15b5f0960ac6100fc7d733cb795dacR489)

tests/e2e/e2e_test.go
pysel: actually the first one: 
> collect fees when there are none

was [already there](https://github.com/osmosis-labs/osmosis/pull/4119/files#diff-7d1fc1e232b55677582b51e2623aae6d8c15b5f0960ac6100fc7d733cb795dacR627) in the `Collect Fees: Sanity Checks` section

tests/e2e/configurer/upgrade.go
```diff
@@ -143,7 +143,7 @@ func (uc *UpgradeConfigurer) CreatePreUpgradeState() error {
 	})
 
 	// test swap exact amount in for stable swap pool (only chainA)A
-	chainANode.SwapExactAmountIn("2000stake", "1", fmt.Sprintf("%d", config.PreUpgradeStableSwapPoolId), "uosmo", config.StableswapWallet)
+	chainANode.SwapExactAmountInLegacy("2000stake", "1", fmt.Sprintf("%d", config.PreUpgradeStableSwapPoolId), "uosmo", config.StableswapWallet)
```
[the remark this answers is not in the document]
pysel: yeah! thanks!

tests/e2e/e2e_test.go
pysel: hmmm @mattverse, we could add a test for asserting that trying to collect fees for position which is not yours is impossible, but that would horribly slow the test. 

the reason is that asserting that command has successfully passed is wrapped in `require.Eventually` and is capped by one minute. If we collect fees for a position that is not owned by a sender, we will currently have to wait for the whole minute to assert the failness

Other approach might be to inspect the logs of a container and check wether some logs were emitted that indicate failure of a command. I actually like this idea, because right now if a developer wants to run an e2e test, and something is failing inside, a person has to wait 1 minute more for the command to be declared failed, however, it might have failed, for example, after one second and just waited for 59 seconds more because `require.Eventually` said so. 

Let me know what you think on the other approach, it might improve dev-UX, but this is out of scope of this PR

PR #4126, opened 2023-01-28
CL: check denoms match when creating migrate link

czarcas7ic:
Closes: #XXX

This is a small patch to prevent migration links from being created between two pools that don't contain the same denoms. It also checks if the balancer pool in question has more than two assets. If it does, it throws an error.

Some tests had to be changed around, since the default balancer pool actually creates a four asset pool! 

## Brief Changelog

- Adds check for balancer pools with more than two assets
- Adds check for balancer pool assets to match cl pool assets
- Adds respective tests to ensure this logic is sound

### DIFF
--- a/x/gamm/keeper/migrate.go
+++ b/x/gamm/keeper/migrate.go
@@ -143,19 +143,25 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
 		}
 
 		// Ensure the balancer pools denoms are the same as the concentrated pool denoms
-		balancerPoolAssets := balancerPool.GetTotalPoolLiquidity(ctx)
+		// If clPoolID is 0, this signals a removal, so we skip this check.
+		if record.ClPoolId != 0 {
+			balancerPoolAssets := balancerPool.GetTotalPoolLiquidity(ctx)
 
-		// Type cast PoolI to ConcentratedPoolExtension
-		clPoolExt, ok := clPool.(cltypes.ConcentratedPoolExtension)
-		if !ok {
-			return fmt.Errorf("pool type (%T) cannot be cast to ConcentratedPoolExtension", clPool)
-		}
+			if len(balancerPoolAssets) != 2 {
+				return fmt.Errorf("Balancer pool ID #%d does not contain exactly 2 tokens", record.BalancerPoolId)
+			}
 
-		if balancerPoolAssets.AmountOf(clPoolExt.GetToken0()).IsZero() {
-			return fmt.Errorf("Balancer pool ID #%d does not contain token %s", record.BalancerPoolId, clPoolExt.GetToken0())
-		}
-		if balancerPoolAssets.AmountOf(clPoolExt.GetToken1()).IsZero() {
-			return fmt.Errorf("Balancer pool ID #%d does not contain token %s", record.BalancerPoolId, clPoolExt.GetToken1())
+			clPoolExt, ok := clPool.(cltypes.ConcentratedPoolExtension)
+			if !ok {
+				return fmt.Errorf("pool type (%T) cannot be cast to ConcentratedPoolExtension", clPool)
+			}
+
+			if balancerPoolAssets.AmountOf(clPoolExt.GetToken0()).IsZero() {
+				return fmt.Errorf("Balancer pool ID #%d does not contain token %s", record.BalancerPoolId, clPoolExt.GetToken0())
+			}
+			if balancerPoolAssets.AmountOf(clPoolExt.GetToken1()).IsZero() {
+				return fmt.Errorf("Balancer pool ID #%d does not contain token %s", record.BalancerPoolId, clPoolExt.GetToken1())
+			}
 		}
 
 		lastBalancerPoolID = record.BalancerPoolId
--- a/x/gamm/keeper/migrate.go
+++ b/x/gamm/keeper/migrate.go
@@ -82,6 +82,8 @@ func (k Keeper) SetMigrationInfo(ctx sdk.Context, migrationInfo types.MigrationR
 // validateRecords validates a list of BalancerToConcentratedPoolLink records to ensure that:
 // 1) there are no duplicates
 // 2) both the balancer and gamm pool IDs are valid
+// 3) the balancer pool has exactly two tokens
+// 4) the denoms of the tokens in the balancer pool match the denoms of the tokens in the gamm pool
 // It also reorders records from lowest to highest balancer pool ID if they are not provided in order already.
 func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConcentratedPoolLink) error {
 	lastBalancerPoolID := uint64(0)
--- a/x/gamm/keeper/migrate.go
+++ b/x/gamm/keeper/migrate.go
@@ -8,6 +8,7 @@ import (
 	cl "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity"
 	cltypes "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types"
 	"github.com/osmosis-labs/osmosis/v14/x/gamm/types"
+	poolmanagertypes "github.com/osmosis-labs/osmosis/v14/x/poolmanager/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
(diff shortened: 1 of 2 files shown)


x/gamm/keeper/migrate_test.go
```diff
@@ -308,6 +311,47 @@ func (suite *KeeperTestSuite) TestReplaceMigrationRecords() {
...
+					ClPoolId:       3,
+				},
+				{
+					BalancerPoolId: 2,
+					ClPoolId:       4,
+				},
+			},
+			overwriteBalancerDenom1: "uosmo",
+			expectErr:               true,
+		},
```
mattverse, line 343: would it be possible to change these test cases so that we have one link?

x/gamm/keeper/migrate_test.go
```diff
@@ -509,18 +575,73 @@ func (suite *KeeperTestSuite) TestUpdateMigrationRecords() {
...
+				balancerDenom1 = test.overwriteBalancerDenom1
+			}
+
 			// Our testing environment is as follows:
 			// Balancer pool IDs: 1, 2, 3, 4
 			// Concentrated pool IDs: 5, 6, 7, 8
-			suite.PrepareMultipleBalancerPools(4)
-			suite.PrepareMultipleConcentratedPools(4)
+			for i := 0; i < 4; i++ {
+				poolCoins := sdk.NewCoins(sdk.NewCoin(balancerDenom0, sdk.NewInt(1000000000)), sdk.NewCoin(balancerDenom1, sdk.NewInt(5000000000)))
```
mattverse, line 635: nit: consider having these coins as varaibles (and then maybe use them across the test cases above?)

mattverse (approved):
Main logic LGTM, left some minor nit comments to be potentially addressed before merging

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +119,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
 				return err
 			}
-			poolType = pool.GetType()
+			poolType = clPool.GetType()
 			if poolType.String() != "Concentrated" {
 				return fmt.Errorf("Concentrated pool ID #%d is not of type concentrated", record.ClPoolId)
 			}
 		}
 
+		// Ensure the balancer pools denoms are the same as the concentrated pool denoms
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4126/commits/f6fc40a51bf12bb2302fc3121831e29b265ba575

x/gamm/keeper/migrate_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4126/commits/25e6048ee13e479383cfef8217f0215673c3c20e

x/gamm/keeper/migrate_test.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4126/commits/447fc5de154cc1913e7e0ebd4742119f6161cbe3

czarcas7ic:
Good recommendations @mattverse :) 

All addressed!

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +121,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
+		// If clPoolID is 0, this signals a removal, so we skip this check.
+		if record.ClPoolId != 0 {
```
p0mvn, line 149: There is the same check `if record.ClPoolId != 0 {` right above. Can we move this code there?

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +121,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
 				return err
 			}
-			poolType = pool.GetType()
+			poolType = clPool.GetType()
 			if poolType.String() != "Concentrated" {
 				return fmt.Errorf("Concentrated pool ID #%d is not of type concentrated", record.ClPoolId)
 			}
 		}
 
+		// Ensure the balancer pools denoms are the same as the concentrated pool denoms
```
p0mvn, line 147: Do we have documentation for migration logic? Can we please add it somewhere? Preferably a section in `architecture.md`

p0mvn (approved):
Nice

p0mvn:
@czarcas7ic while reviewing, I noticed something unrelated to this PR but related to migrations and small enough that would be great to do it just here if you don't mind.

So I noticed that [here](https://github.com/osmosis-labs/osmosis/blob/7d239c73e32705f90c3433019056674705a3b096/x/gamm/keeper/migrate.go#L129) and [here](https://github.com/osmosis-labs/osmosis/blob/7d239c73e32705f90c3433019056674705a3b096/x/gamm/keeper/migrate.go#L142), we compare strings. However, we should be comparing against types instead. 

If the string value accidentally changes in the future, the whole logic can break. Using the types directly helps prevent that.

Please let me know if you can change it here. If not, we can cut an issue

x/gamm/keeper/migrate.go
ValarDragon: Was about to come and ask the same thing.

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +121,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
 		}
-		poolType := pool.GetType()
+		poolType := balancerPool.GetType()
 		if poolType.String() != "Balancer" {
 			return fmt.Errorf("Balancer pool ID #%d is not of type balancer", record.BalancerPoolId)
 		}
 
 		// Ensure the provided ClPoolId exists and that it is of type concentrated.
 		// If clPoolID is 0, this signals a removal, so we skip this check.
+		var clPool poolmanagertypes.PoolI
```
ValarDragon, line 135: Ah good change, separating the names here!

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +121,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
 			}
-			poolType = pool.GetType()
+			poolType = clPool.GetType()
 			if poolType.String() != "Concentrated" {
 				return fmt.Errorf("Concentrated pool ID #%d is not of type concentrated", record.ClPoolId)
 			}
 		}
 
+		// Ensure the balancer pools denoms are the same as the concentrated pool denoms
+		// If clPoolID is 0, this signals a removal, so we skip this check.
```
ValarDragon, line 148: Err, i'm not clear what clPool Id == 0 actually is doing here.

Perhaps lets have a block like
```
if clPoolId == 0 { // do stuff & exit }

// current logic
```

x/gamm/keeper/migrate.go
```diff
@@ -118,28 +121,51 @@ func (k Keeper) validateRecords(ctx sdk.Context, records []types.BalancerToConce
...
+		var clPool poolmanagertypes.PoolI
 		if record.ClPoolId != 0 {
-			pool, err = k.clKeeper.GetPool(ctx, record.ClPoolId)
+			clPool, err = k.clKeeper.GetPool(ctx, record.ClPoolId)
 			if err != nil {
 				return err
 			}
-			poolType = pool.GetType()
+			poolType = clPool.GetType()
 			if poolType.String() != "Concentrated" {
```
ValarDragon, line 142: Presumably we have an actual enum for this somewhere, right? Can we do that instead of string matching?

czarcas7ic: We do have an enum, but externally its less clear. This is quite clear that we want the pool to be of type Concentrated instead of type 3, which could mean anything.

x/gamm/keeper/migrate.go
czarcas7ic: We don't want validation to run when a CL pool is equal to 0. This is because when we submit a gov prop with a CL pool equal to zero, it indicates we want to remove the link between the CL and balancer pool.

I feel like what we do currently in this PR is equal to what you are saying. If clPoolId == 0, do stuff and exit (in this case, the stuff to do is nothing and we exit). I might be misunderstanding though.

x/gamm/keeper/migrate.go
czarcas7ic: Agreed, done here https://github.com/osmosis-labs/osmosis/pull/4126/commits/c93985b29cfb03cd2cf27abbfb7cae4e7eda3c3b

x/gamm/keeper/migrate.go
czarcas7ic: Addressed this here, where we now compare it to type instead of string

https://github.com/osmosis-labs/osmosis/pull/4126/commits/1844dcc1c66708cad48590e9cae9cbbdb1c25202

CC @p0mvn addressing your comment

x/gamm/keeper/migrate.go
czarcas7ic: I had added a blurb about this in GAMM here

https://github.com/osmosis-labs/osmosis/blob/1844dcc1c66708cad48590e9cae9cbbdb1c25202/x/gamm/README.md#L189-L191

I can go into more detail if this is desired

x/gamm/keeper/migrate.go
p0mvn: I propose to merge this PR and resolve this discussion separately unless any concerns.

I think changing this logic would require effort outside of the scope of work done here. So it is better to be addressed in a subsequent PR

PR #4129, opened 2023-01-28
feat[CL]: lock migration to frozen concentrated position

czarcas7ic:

We currently only have logic to migrate unlocked LP tokens to a full range CL position. This PR introduces logic slightly similar to the unpool logic, where we 

1. Superfluid undelegate if lock was superfluid delegated
2. Burn underlying synthetic OSMO
3. Break the lock

But now instead of creating new locks, we simply create full range CL position that is frozen for either the entire lock duration, or the remaining lock duration if it was in the process of unlocking.

## Brief Changelog

- Adds clKeeper dependence to the superfluid keeper
- Removes pool_id_entering proto field (is not automatically discerned from gov sanctioned link)
- Adds `MsgUnlockAndMigrateSharesToFullRangeConcentratedPosition`
  - This is used for locked LP tokens, whereas `MsgMigrateSharesToFullRangeConcentratedPosition` in gamm is only used on unlocked LP tokens
- Adds test cases for new `UnlockAndMigrate` logic 
- Adds event for `UnlockAndMigrate`

## Testing and Verifying

Various test cases added as described above.


proto/osmosis/gamm/pool-models/balancer/tx/tx.proto
```diff
@@ -42,8 +42,6 @@ message MsgMigrateSharesToFullRangeConcentratedPosition {
     (gogoproto.moretags) = "yaml:\"shares_to_migrate\"",
     (gogoproto.nullable) = false
   ];
-  // temporary field, eventually gamm pool should be linked to cl pool
-  uint64 pool_id_entering = 3;
```
czarcas7ic, line 46: Note to reviewer: No longer required since we now get this from the gov sanctioned link between pools.

x/gamm/keeper/migrate.go
```diff
@@ -13,56 +13,44 @@ import (
...
-		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, err
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, err
 	}
 
 	// Determine the max and min ticks for the concentrated pool we are migrating to.
 	minTick, maxTick := cl.GetMinAndMaxTicksFromExponentAtPriceOne(concentratedPool.GetPrecisionFactorAtPriceOne())
 
 	// Create a full range (min to max tick) concentrated liquidity position.
-	// TODO: Will need to implement lock breaking logic and add the corresponding freeze duration here.
 	amount0, amount1, liquidity, err = k.clKeeper.CreatePosition(ctx, poolIdEntering, sender, exitCoins.AmountOf(concentratedPool.GetToken0()), exitCoins.AmountOf(concentratedPool.GetToken1()), sdk.ZeroInt(), sdk.ZeroInt(), minTick, maxTick, time.Time{})
```
czarcas7ic, line 49: Note to reviewer: We freeze for the zero value of time here since only unlocked LP tokens can get migrated through this entry point.

x/gamm/keeper/migrate_test.go
```diff
@@ -98,25 +93,12 @@ func (suite *KeeperTestSuite) TestMigrate() {
...
-			param: param{
-				sender:                defaultAccount,
-				sharesToMigrateDenom:  defaultGammShares.Denom,
-				sharesToMigrateAmount: defaultGammShares.Amount,
-				poolIdEntering:        1,
-			},
-			sharesToCreate:         defaultGammShares.Amount,
-			setupPoolMigrationLink: true,
-			expectedErr:            types.InvalidPoolMigrationLinkError{PoolIdEntering: 1, CanonicalId: 2},
-		},
```
czarcas7ic, line 119: Note to reviewer: This test case is no longer possible due to user no longer being the one to provide link information

x/gamm/keeper/pool.go
```diff
@@ -297,12 +295,3 @@ func convertToCFMMPool(pool poolmanagertypes.PoolI) (types.CFMMPoolI, error) {
...
 }
-
-// getPoolIdFromSharesDenom takes in a string representing a pool share denom and extracts the pool ID.
-// It returns the pool ID as a uint64 and an error if the denom is invalid.
-func getPoolIdFromSharesDenom(denom string) (uint64, error) {
-	if !strings.HasPrefix(denom, "gamm/pool/") {
-		return 0, fmt.Errorf("invalid pool share denom %s", denom)
-	}
-	return strconv.ParseUint(denom[len("gamm/pool/"):], 10, 64)
-}
```
czarcas7ic, line 308: Note to reviewer: Removed since this is code duplication. Used the original func instead.

x/superfluid/keeper/migrate_test.go
```diff
@@ -0,0 +1,211 @@
...
+			// Run the unlock and migrate logic.
+			amount0, amount1, _, _, _, err := superfluidKeeper.UnlockAndMigrate(ctx, poolJoinAcc, lockID, poolShareOut)
+			suite.Require().NoError(err)
+
+			suite.AssertEventEmitted(ctx, gammtypes.TypeEvtPoolExited, 1)
+
+			// exitPool has rounding difference.
+			// We test if correct amt has been exited and frozen by comparing with rounding tolerance.
+			defaultErrorTolerance := osmomath.ErrTolerance{
+				AdditiveTolerance: sdk.NewDec(2),
```
mattverse, line 185: Do you have a brief number of how much the joinPoolAmt is here? (Asking to get a brief sense of the significance of the tolerance here)

x/superfluid/keeper/migrate_test.go
```diff
@@ -0,0 +1,211 @@
...
+				RoundingDir:       osmomath.RoundDown,
+			}
+			suite.Require().Equal(0, defaultErrorTolerance.Compare(joinPoolAmt.AmountOf("foo"), amount0))
+			suite.Require().Equal(0, defaultErrorTolerance.Compare(joinPoolAmt.AmountOf("stake"), amount1))
+
```
mattverse, line 190: Can we add a check that checks if the frozen unitl has been properly migrated?

x/superfluid/types/expected_keepers.go
```diff
@@ -47,6 +48,9 @@ type GammKeeper interface {
 	GetPoolAndPoke(ctx sdk.Context, poolId uint64) (gammtypes.CFMMPoolI, error)
 	GetPoolsAndPoke(ctx sdk.Context) (res []gammtypes.CFMMPoolI, err error)
 	ExitPool(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, shareInAmount sdk.Int, tokenOutMins sdk.Coins) (exitCoins sdk.Coins, err error)
+	GetMigrationInfo(ctx sdk.Context) gammtypes.MigrationRecords
+	Migrate(ctx sdk.Context, sender sdk.AccAddress, sharesToMigrate sdk.Coin) (amount0, amount1 sdk.Int, liquidity sdk.Dec, poolIdLeaving, poolIdEntering uint64, err error)
```
mattverse, line 52: Why do we need gamm keeper's Migrate method?

mattverse (commented):
Nice work! Thanks for all the detailed in line comments, was super helpful in reviewing this PR. Left some minor comments, please take a look.

x/superfluid/types/expected_keepers.go
czarcas7ic: We don't, good catch! Before I wanted to make superfluid the only entrypoint and simply skip the unlocking part if it wasn't locked, but it ended up being a bad design and I scrapped it but left this behind. Removed.

x/gamm/keeper/migrate.go
```diff
@@ -13,56 +13,44 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func (k Keeper) Migrate(ctx sdk.Context, sender sdk.AccAddress, sharesToMigrate sdk.Coin, poolIdEntering uint64) (amount0, amount1 sdk.Int, liquidity sdk.Dec, poolIdLeaving uint64, err error) {
+// Migrate migrates unlocked lp tokens from a balancer pool to a concentrated liquidity pool.
+// Fails if the lp tokens are locked (must utilize UnlockAndMigrate function in the superfluid module)
+func (k Keeper) Migrate(ctx sdk.Context, sender sdk.AccAddress, sharesToMigrate sdk.Coin) (amount0, amount1 sdk.Int, liquidity sdk.Dec, poolIdLeaving, poolIdEntering uint64, err error) {
```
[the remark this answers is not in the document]
czarcas7ic: Changed name as requested here https://github.com/osmosis-labs/osmosis/pull/4129/commits/36fb46c060e7274b5026b87b2d3632b38c9dffa4

x/gamm/keeper/migrate.go
```diff
@@ -13,56 +13,44 @@ import (
...
 	// Get the balancer poolId by parsing the gamm share denom.
-	poolIdLeaving, err = getPoolIdFromSharesDenom(sharesToMigrate.Denom)
-	if err != nil {
-		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, err
-	}
+	poolIdLeaving = types.MustGetPoolIdFromShareDenom(sharesToMigrate.Denom)
```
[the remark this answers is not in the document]
czarcas7ic: Do you think I should reintroduce the function I removed from this PR? I basically wanted to utilize a function that already existed but I realize here that we panic if the denom does not exist. Is this problematic? If so, I will reintroduce the function I had made previously for this.

x/gamm/keeper/migrate_test.go
```diff
@@ -162,7 +144,7 @@ func (suite *KeeperTestSuite) TestMigrate() {
 
 		// Migrate the user's gamm shares to a full range concentrated liquidity position
 		userBalancesBeforeMigration := suite.App.BankKeeper.GetAllBalances(suite.Ctx, test.param.sender)
-		amount0, amount1, _, _, err := keeper.Migrate(suite.Ctx, test.param.sender, sharesToMigrate, test.param.poolIdEntering)
+		amount0, amount1, _, _, _, err := keeper.Migrate(suite.Ctx, test.param.sender, sharesToMigrate)
```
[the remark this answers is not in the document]
czarcas7ic: Added check here https://github.com/osmosis-labs/osmosis/pull/4129/commits/e9412d780285a5fc6dab175e080ca0911f9aa739

x/gamm/keeper/migrate.go
```diff
@@ -206,3 +194,20 @@ func (k Keeper) UpdateMigrationRecords(ctx sdk.Context, records []types.Balancer
...
 	return nil
 }
+
+// CheckIfPoolMigrationLinkIsAuthorized checks if a governance sanctioned link exists between the provided balancer and concentrated pool.
+func (k Keeper) GetLinkedConcentratedPool(ctx sdk.Context, poolIdLeaving uint64) (poolIdEntering uint64, err error) {
+	migrationInfo := k.GetMigrationInfo(ctx)
+	matchFound := false
+	for _, info := range migrationInfo.BalancerToConcentratedPoolLinks {
+		if info.BalancerPoolId == poolIdLeaving {
+			poolIdEntering = info.ClPoolId
```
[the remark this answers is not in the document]
czarcas7ic: That is done in [this PR](https://github.com/osmosis-labs/osmosis/pull/4126/files) and I intend to merge it before this one. Good catch though

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,77 @@
...
+	// Get the concentrated pool from the provided ID and type cast it to ConcentratedPoolExtension.
+	poolI, err := k.clk.GetPool(ctx, poolIdEntering)
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, err
+	}
+	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
+	if !ok {
+		// If the conversion fails, return an error.
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
+	}
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4129/commits/a06774daf043b4f8e8803dc7ebb34a8be92761f6

x/gamm/keeper/migrate_test.go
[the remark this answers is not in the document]
czarcas7ic: And here in superfluid https://github.com/osmosis-labs/osmosis/pull/4129/commits/c4165de16b24c7c541bb83e5baa4257ce1c1b549

x/superfluid/keeper/migrate_test.go
czarcas7ic: What I ended up doing was checking that the expected position with a frozenUntil value was created https://github.com/osmosis-labs/osmosis/pull/4129/commits/5f0d2a048e2e2df35e8866eabf91cbcbf24210e6

We test `getExistingLockRemainingDuration` on its own, so we have a guarantee that whatever that frozenUntil value that gets returned is valid. Your point still stands though and it is now indirectly checked!

x/superfluid/keeper/migrate_test.go
czarcas7ic: Yeah, these values are equal to 5000 uosmo so the tolerance would be 2 uosmo compared to that. To be clear though (and this originally wasn't clear to me before) the user doesn't get shorted these 2 uosmo, they simply remain in the wallet and are free from any lock. Even with very large amounts this amounts to dust though.

x/gamm/keeper/migrate.go
```diff
@@ -206,3 +194,20 @@ func (k Keeper) UpdateMigrationRecords(ctx sdk.Context, records []types.Balancer
 	})
 	return nil
 }
+
+// CheckIfPoolMigrationLinkIsAuthorized checks if a governance sanctioned link exists between the provided balancer and concentrated pool.
+func (k Keeper) GetLinkedConcentratedPool(ctx sdk.Context, poolIdLeaving uint64) (poolIdEntering uint64, err error) {
```
p0mvn, line 199: Spec doesn't apply to method

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,80 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	cl "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity"
+	cltypes "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types"
+	gammtypes "github.com/osmosis-labs/osmosis/v14/x/gamm/types"
+)
+
+// UnlockAndMigrate unlocks a balancer pool lock, exits the pool and migrates the LP position to a full range concentrated liquidity position.
+// If the lock is also superfluid delegated, it will undelegate the superfluid position.
```
p0mvn, line 15:
```suggestion
// If the lock is superfluid delegated, it will undelegate the superfluid position.
```

p0mvn, line 15: Consider elaborating when errors are expected

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,80 @@
...
+	// Get the concentrated pool from the provided ID and type cast it to ConcentratedPoolExtension.
+	poolI, err := k.clk.GetPool(ctx, poolIdEntering)
+	if err != nil {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, time.Time{}, err
+	}
+	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
+	if !ok {
+		// If the conversion fails, return an error.
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, time.Time{}, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
+	}
```
p0mvn, line 35: We might want to have a helper in CL module that returns the desired type. This seems to be common enough to justify this helper and replace of all of the type casts

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Consistency check that lockID corresponds to sender, and contains correct LP shares.
+	// These are expected to be true by the caller, but good to double check
+	lock, err := k.validateLockForUnpool(ctx, sender, poolIdLeaving, lockId)
```
p0mvn, line 39: How do we compare the correct LP shares part? That is, how do we know that the lock corresponds to the right amount of shares?

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,80 @@
...
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, time.Time{}, err
+	}
+	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
+	if !ok {
+		// If the conversion fails, return an error.
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, time.Time{}, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
+	}
+
+	// Consistency check that lockID corresponds to sender, and contains correct LP shares.
+	// These are expected to be true by the caller, but good to double check
```
p0mvn, line 38: What does it mean? I looked at the caller and don't see the same validation being made anywhere

x/superfluid/keeper/migrate.go
p0mvn: I think we need a check and a test that `sharesToMigrate` and lock's shares are the same

x/superfluid/keeper/migrate_test.go
```diff
@@ -0,0 +1,224 @@
...
+			clPool := suite.PrepareCustomConcentratedPool(poolCreateAcc, defaultPoolAssets[0].Token.Denom, defaultPoolAssets[1].Token.Denom, 1, sdk.NewInt(-6), sdk.ZeroDec())
+
+			// Add a sanctioned link between the balancer and concentrated liquidity pool.
+			migrationRecord := gammtypes.MigrationRecords{BalancerToConcentratedPoolLinks: []gammtypes.BalancerToConcentratedPoolLink{
+				{BalancerPoolId: balancerPool.GetId(), ClPoolId: clPool.GetId()},
+			}}
+			gammKeeper.SetMigrationInfo(ctx, migrationRecord)
+
+			// The amount of coins we lock is the amount of LP tokens the account that joined the pool has.
+			coinsToLock := poolShareOut
```
p0mvn, line 133: Let's add a test where lock has a different number of shares locked from what we are trying to migrate

x/gamm/keeper/migrate.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4129/commits/4f62673f13b377e76a5feecccd511c09e4c08f9c

x/superfluid/types/expected_keepers.go
```diff
@@ -47,6 +48,8 @@ type GammKeeper interface {
 	GetPoolAndPoke(ctx sdk.Context, poolId uint64) (gammtypes.CFMMPoolI, error)
 	GetPoolsAndPoke(ctx sdk.Context) (res []gammtypes.CFMMPoolI, err error)
 	ExitPool(ctx sdk.Context, sender sdk.AccAddress, poolId uint64, shareInAmount sdk.Int, tokenOutMins sdk.Coins) (exitCoins sdk.Coins, err error)
+	GetMigrationInfo(ctx sdk.Context) gammtypes.MigrationRecords
+	GetLinkedConcentratedPool(ctx sdk.Context, poolIdLeaving uint64) (poolIdEntering uint64, err error)
```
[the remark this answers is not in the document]
czarcas7ic: Renamed https://github.com/osmosis-labs/osmosis/pull/4129/commits/786389c534947066e7fadd7e30a99005825ad662

x/gamm/keeper/migrate.go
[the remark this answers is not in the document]
czarcas7ic: Added https://github.com/osmosis-labs/osmosis/pull/4129/commits/46404a807264638c2071b92d4840714e739023a0

x/superfluid/keeper/migrate.go
czarcas7ic: Added https://github.com/osmosis-labs/osmosis/pull/4129/commits/8bdae4d6612b94284f3669241dd7814a5fae8945

x/superfluid/keeper/migrate.go
czarcas7ic: Added helper function in CL and implemented in gamm and superfluid https://github.com/osmosis-labs/osmosis/pull/4129/commits/7daae195de526ec1dcffc273516bccc83be9c560

x/superfluid/keeper/migrate.go
czarcas7ic: You are correct, it is not expected to be true by the caller. I utilized `validateLockForUnpool` from the unpool logic to prevent code duplication but the comment should have been changed. I changed the comment here, thanks! https://github.com/osmosis-labs/osmosis/pull/4129/commits/738b4a994d95e8cfc5c740d3fd965b2e6188505f

x/superfluid/keeper/migrate_test.go
czarcas7ic: Great point, I modified all the tests here to use varying percentages of the initial join amount! https://github.com/osmosis-labs/osmosis/pull/4129/commits/d961640126dbf52f7b779ae531a710723ecb92e3

x/superfluid/keeper/migrate.go
czarcas7ic: CL shares and balancer shares are not 1:1. With that being said, we test what the shares represent in terms of the underlying which is all that really matters IMO. Let me know if this addresses that concern.

czarcas7ic: Ohh I think I see the confusion now, the comment says "contains correct LP shares" but what it actually means is "contains correct denom LP shares".

czarcas7ic: Added this clarifying comment, lmk if that is sufficient for you: https://github.com/osmosis-labs/osmosis/pull/4129/commits/28f04163b4c7d1bccd2021d44dc209d4f79b9af6

x/gamm/keeper/migrate.go
```diff
@@ -13,56 +13,44 @@ import (
...
 	poolI, err := k.clKeeper.GetPool(ctx, poolIdEntering)
 	if err != nil {
-		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, err
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, err
 	}
 	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
 	if !ok {
 		// If the conversion fails, return an error.
-		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, fmt.Errorf("given pool does not implement ConcentratedPoolExtension, implements %T", poolI)
```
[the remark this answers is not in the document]
czarcas7ic: Agreed, did this with `CreateFullRangePosition` and `GetPoolFromPoolIdAndConvertToConcentrated`. While I could have actually put the `GetPoolFromPoolIdAndConvertToConcentrated` logic into `CreateFullRangePosition`, I want to have the poolId check in the very beginning to prevent overhead when running in invalid tx.

https://github.com/osmosis-labs/osmosis/pull/4129/commits/788065d31ce523e812cc679013c9dbde29ea6458

x/superfluid/keeper/migrate.go
p0mvn: I guess my main question is with regards to the number of shares locked being greater than the number of shares we migrate.

Say we migrate 4 shares but the lock contains 10. Then, we exit 4/10 shares. What happens to the other 6? Currently, it seems that they are left unlocked.

Maybe we should relock them?

Also, could you please document in the method spec and test the final behavior of this excess lock share amount

x/gamm/keeper/migrate.go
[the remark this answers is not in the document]
mattverse: Sweet thanks!

x/gamm/keeper/migrate.go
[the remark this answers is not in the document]
mattverse: Maybe it's a open question, but I personally see that unifying error / panic as much as possible, and letting the caller decide what to do with the error as a better design, so that we don't have multiple places that errors or panics.(Setters and getters from state are unevitable for panicing). But again, its an open ended question, I'll leave it up to you to decide on this :)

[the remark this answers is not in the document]
mattverse: I personally think unifying panics and errors as much as possible is a better design so that we don't have multiple places here and there that either errors or panics, so that the caller of this function can have a unified point of handling errors (the must getters and must setters from state is inevitable) . But again, I guess this is an open ended question, I'll leave it to you to decide :)

x/superfluid/keeper/migrate_test.go
mattverse: Sweet! Thanks for clarifying

mattverse (approved):
my main concerns and reviews have all been addressed! LGTM 🌮

x/gamm/keeper/migrate.go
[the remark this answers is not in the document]
czarcas7ic: Agreed! Addressed here https://github.com/osmosis-labs/osmosis/pull/4129/commits/d3bdaa11aa1768044f843bd148388a498601c318

x/superfluid/keeper/migrate.go
czarcas7ic: Woh, this is a great point. Thanks for pointing this out! Let me address it before we merge this

czarcas7ic: @p0mvn I believe the following commit solves the issue you found, thanks for pointing this out!

https://github.com/osmosis-labs/osmosis/pull/4129/commits/68c7991099c48746ef1fb25c640cb7b04e3328af

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,103 @@
...
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, 0, time.Time{}, err
+	}
+
+	// If shares to migrate is not specified, we migrate all shares.
+	if sharesToMigrate.IsZero() {
+		sharesToMigrate = gammSharesInLock
+	}
+
+	// Otherwise, we must ensure that the shares to migrate is less than or equal to the shares in the lock.
+	remainingGammShares := gammSharesInLock.Sub(sharesToMigrate)
```
p0mvn, line 63: nit: can we move this further down to line 86 where this is being used for the first time

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,103 @@
...
+	// If there are remaining gamm shares, we must re-lock them.
+	if !remainingGammShares.IsZero() {
+		newLock, err := k.lk.CreateLock(ctx, sender, sdk.NewCoins(remainingGammShares), freezeDuration)
+		newLockId = newLock.ID
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, 0, time.Time{}, err
+		}
+		// If the lock was unlocking, we begin the unlock from where it left off.
+		if preUnlockLock.IsUnlocking() {
+			_, err := k.lk.BeginForceUnlock(ctx, newLock.ID, newLock.Coins)
```
p0mvn, line 95: nit: we can call `beginUnlock` directly

x/superfluid/keeper/migrate.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// If there are remaining gamm shares, we must re-lock them.
+	if !remainingGammShares.IsZero() {
+		newLock, err := k.lk.CreateLock(ctx, sender, sdk.NewCoins(remainingGammShares), freezeDuration)
+		newLockId = newLock.ID
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, 0, 0, time.Time{}, err
+		}
+		// If the lock was unlocking, we begin the unlock from where it left off.
```
p0mvn, line 93: How is beginning to unlock where it left off guaranteed?

x/superfluid/keeper/migrate_test.go
```diff
@@ -0,0 +1,249 @@
...
+		},
+		{
+			"lock that is superfluid undelegating, not unlocking",
+			true,
+			true,
+			false,
+			sdk.MustNewDecFromStr("0.5"),
+		},
+		{
+			"lock that is superfluid undelegating, unlocking",
```
p0mvn, line 62: Is it possible to test the "where it left" off part somehow>?

p0mvn (approved):
LGTM if my questions have been considered

x/superfluid/keeper/migrate.go
czarcas7ic: I don't think we can, unless we want to export beginUnlock from the lockup module.

x/superfluid/keeper/migrate.go
czarcas7ic: We lock it above with freezeDuration which is calculated as the remaining time the lock had before breaking, but I agree this needs to be tested

czarcas7ic: Added test! https://github.com/osmosis-labs/osmosis/pull/4129/commits/04518037f42656afd74f111b82f9bdbfe96b541b

x/superfluid/keeper/migrate_test.go
czarcas7ic: Added test! https://github.com/osmosis-labs/osmosis/pull/4129/commits/04518037f42656afd74f111b82f9bdbfe96b541b

Issue #4134, opened 2023-01-29
[CL Incentives] Track relevant uptime accumulators in new positions


AlpinYukseloglu:
## Background

When a new position is created, we need to create records for each uptime accumulator they are eligible for.

## Suggested Design

- Upon position creation, check if `frozenUntil - curTime >= minUptime` for each supported uptime and, if so, create a record on that uptime's accumulator.

## Acceptance Criteria

- All new and existing tests pass

Issue #4137, opened 2023-01-30
[CL Incentives] Claim incentives for all uptimes


AlpinYukseloglu:
## Background

We should have one user-facing claim function for incentives. If needed, we can make a second one that allows the caller to specify the uptime accumulator being claimed from, but a "claim all" function is likely sufficient.

## Suggested Design

- `ClaimIncentivesForPosition` message that claims the outstanding incentives a position is entitled to for all supported uptimes in a pool
- `UptimeGrowthBelow` and `UptimeGrowthInside` functions that allow us to specify/understand how much the of the global incentives the claimer qualifies for.

## Acceptance Criteria

- All new and existing tests pass

PR #4138, opened 2023-01-30
[CL Incentives] Track relevant uptime accumulators in new positions

AlpinYukseloglu:
Closes: #4134

This PR implements CL positions tracking uptime accumulators when they are created. It is intended to only implement and test that new positions have corresponding records created for each uptime they are eligible for based on their freeze duration.

## Brief Changelog

- Added following logic: when a position is created, create a record for the respective accumulator of each uptime it qualifies for

## Testing and Verifying

- New functionality is tested in `incentives_test.go`
- Needed to add uptime accum initialization to `fee_test.go` so position creation would work (couldn't just use usual pool setup funcs as those would also initialize fee accumulators, breaking the tests)


x/concentrated-liquidity/fees_test.go
```diff
@@ -464,6 +463,9 @@ func (suite *KeeperTestSuite) TestGetInitialFeeGrowthOutsideForTick() {
 				suite.Require().NoError(err)
 
 				// Setup test position to make sure that tick is initialized
+				// We also set up uptime accums to ensure position creation works as intended
+				err = clKeeper.CreateUptimeAccumulators(ctx, validPoolId)
+				suite.Require().NoError(err)
```
czarcas7ic, line 468: This bool is a little confusing to me. We create the uptime accumulators regardless of whether this flag is set or not, since we call `CreateUptimeAccumulators` when we initialize the pool which gets called in create pool. 

I could be misunderstanding the purpose of this test vector, but if so would be great to clarify with name changes or comments.

x/concentrated-liquidity/incentives_test.go
```diff
@@ -0,0 +1,119 @@
...
+	for name, tc := range tests {
+		tc := tc
+		s.Run(name, func() {
+			s.SetupTest()
+			clKeeper := s.App.ConcentratedLiquidityKeeper
+
+			// system under test
+			var err error
+			if tc.initializePool {
+				err = clKeeper.CreateUptimeAccumulators(s.Ctx, tc.poolId)
```
czarcas7ic, line 59: should error check here since the err gets overwritten by GetUptimeAccumulators by the time we check it

x/concentrated-liquidity/incentives_test.go
```diff
@@ -0,0 +1,119 @@
...
+	// if the uptimes we support ever change.
+	curExpectedAccumValues := []sdk.DecCoins{}
+	for range types.SupportedUptimes {
+		curExpectedAccumValues = append(curExpectedAccumValues, cl.EmptyCoins)
+	}
+	s.Require().Equal(len(types.SupportedUptimes), len(curExpectedAccumValues))
+
+	type initUptimeAccumTest struct {
+		poolId              uint64
+		initializePool      bool
```
czarcas7ic, line 24: nit: maybe we should call this `initializePoolAccum` since we dont actually initialize a pool in the test.

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
 
 	// TODO: consider deleting position if liquidity becomes zero
 
+	// Create records for relevant uptime accumulators here.
```
czarcas7ic, line 66: Can you explain what happens to an uptime accumulator once the position is deleted? I just want to make sure it also gets deleted from state when that happens

czarcas7ic (approved):
Good job, have some minor comments that I would like to discuss but should not be blocking

x/concentrated-liquidity/position.go
AlpinYukseloglu: Discussed on call, created issue here: #4183

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Nice catch, naming was from an earlier version of tests. Changing!

x/concentrated-liquidity/fees_test.go
AlpinYukseloglu: This test does not initialize the whole pool state – it manually tries to set up everything except for the fee accumulator so it can isolate and test functionality around that. Now that uptime accumulators are a part of standard pool initialization, this manual setup process needs to include them as well.

In other words, since this PR introduces checks to position creation that requires uptime accums to be initialized, the previous manual pool creation is no longer sufficient and needs to include setup for uptime accumulators. cc @p0mvn

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Nice catch, fixed

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
...
+	uptimeAccumulators, err := k.getUptimeAccumulators(ctx, poolId)
+	if err != nil {
+		return err
+	}
+
+	for uptimeId, uptime := range types.SupportedUptimes {
+		if position.FrozenUntil.Sub(ctx.BlockTime()) >= uptime {
+			curUptimeAccum := uptimeAccumulators[uptimeId]
```
mattverse, line 74: Is the `uptimeAccumualtors` slice that we got via `getUptimeAccumulators` sorted? Worried that this is not sorted and we use the wrong index

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
...
+			if err != nil {
+				return err
+			}
+
+			if !recordExists {
+				err = curUptimeAccum.NewPosition(positionName, position.Liquidity, &accum.Options{})
+			} else if !liquidityDelta.IsNegative() {
+				err = curUptimeAccum.AddToPosition(positionName, liquidityDelta)
+			} else {
+				err = curUptimeAccum.RemoveFromPosition(positionName, liquidityDelta.Neg())
```
mattverse, line 89: why not use `UpdatePositionCustomAcc` here?

mattverse (commented):
Looks nice! Left some minor changes and questions, otherwise lgtm

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
...
+			// If a record does not exist for this uptime accumulator, create a new position.
+			// Otherwise, add to existing record.
+			positionName := string(types.KeyFullPosition(poolId, owner, lowerTick, upperTick, frozenUntil))
+			recordExists, err := curUptimeAccum.HasPosition(positionName)
+			if err != nil {
+				return err
+			}
+
+			if !recordExists {
+				err = curUptimeAccum.NewPosition(positionName, position.Liquidity, &accum.Options{})
```
p0mvn, line 85: nit: we can story empty options as a global variable

x/concentrated-liquidity/position.go
p0mvn: Agreed

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
 
 	// TODO: consider deleting position if liquidity becomes zero
 
+	// Create records for relevant uptime accumulators here.
+	uptimeAccumulators, err := k.getUptimeAccumulators(ctx, poolId)
+	if err != nil {
+		return err
+	}
+
+	for uptimeId, uptime := range types.SupportedUptimes {
```
p0mvn, line 72: nit: I was confused for a second why uptimes have ids
```suggestion
	for uptimeIndex, uptime := range types.SupportedUptimes {
```

x/concentrated-liquidity/position.go
```diff
@@ -62,6 +63,37 @@ func (k Keeper) initOrUpdatePosition(
...
+		if position.FrozenUntil.Sub(ctx.BlockTime()) >= uptime {
```
p0mvn, line 73: Can we have some tests around this check? I don't think we have any tests where we manipulate the block time

p0mvn: non-blocking: It might even be helpful to encapsulate all record creation logic in a separate function and test it

x/concentrated-liquidity/position.go
AlpinYukseloglu: This part of the code will be expanded on in #4137 (there's additional reward-related logic that needs to be added here for proper claiming – this PR is simply for initializing the accumulators in positions)

x/concentrated-liquidity/position.go
AlpinYukseloglu: For your first point, this could be useful to add but every test setup in `position_test.go` uses a different time (since it sets up to local time) so I believe tests for different uptimes are already tested

x/concentrated-liquidity/position.go
AlpinYukseloglu: Adding

x/concentrated-liquidity/position.go
AlpinYukseloglu: Updating

x/concentrated-liquidity/position_test.go
```diff
@@ -129,6 +129,22 @@ func (s *KeeperTestSuite) TestInitOrUpdatePosition() {
...
+			// If frozen for more than a specific uptime's period, the record should exist
+			for uptimeIndex, uptime := range types.SupportedUptimes {
+				recordExists, err := uptimeAccums[uptimeIndex].HasPosition(positionName)
+				s.Require().NoError(err)
+				if test.param.frozenUntil.Sub(s.Ctx.BlockTime()) >= uptime {
+					s.Require().True(recordExists)
+				} else {
+					s.Require().False(recordExists)
+				}
+			}
```
p0mvn, line 147: nit: It might even be better to hardcoded in a slice for which uptime we expect this to be true. Otherwise, if there is an error in the code, we can have it here left unnoticed

p0mvn (approved):

x/concentrated-liquidity/position_test.go
AlpinYukseloglu: I actually thought of this but I intentionally want to keep uptime-related tests agnostic to the number of uptimes we have so that if we change anything it doesn't break a bunch of things. Noted the concern though and will keep this in mind

Issue #4130, opened 2023-01-29
CLI for `GetPositionsByUserAddress`


RusAkh:
## Background
When https://github.com/osmosis-labs/osmosis/pull/4125#discussion_r1089869191 is merged, we will be able to query concentrated positions for a specific address. We need to add a cli query for that

## Suggested Design

Query added

## Acceptance Criteria

Query added

PR #4140, opened 2023-01-30
CLI: GetUserPositions

RusAkh:
- proto
- add query
- regenerated position.pb
- generated proto
- renames
- query for user positions
- move positions.proto back
- cl types
- positions import
- desc positions

Closes: #4130

This PR adds a cli command to query user positions by address

## Keep in mind
This PR moves `position.proto` from `model` to `types`. Reason: if we leave it in `model`, we cannot use it in `query.proto`, due to the import cycle error when the code is generated

no


proto/osmosis/concentrated-liquidity/position.proto
```diff
@@ -9,7 +9,7 @@ import "cosmos_proto/cosmos.proto";
 import "google/protobuf/timestamp.proto";
 import "gogoproto/gogo.proto";
 
-option go_package = "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/model";
+option go_package = "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidity/types";
```
RusAkh, line 12: Reason: if we leave this in `model`, we cannot use it in `query.proto`, because when the code is generated, there will be an import cycle error

PR #4143, opened 2023-01-30
ValSet Preference Wholistic Review

mattverse:
Closes: #XXX

A wholistic review of the val set preference module. 

This PR contains direct changes to variable name changes, method name changes and additional inline documentation and comments. 

Those that are further actions items and questions for @stackman27 is addressed with "// Matt" comment in the front


x/valset-pref/validator_set.go
```diff
@@ -44,10 +70,10 @@ func (k Keeper) SetValidatorSetPreference(ctx sdk.Context, delegator string, pre
 // For ex: delegate 10osmo with validator-set {ValA -> 0.5, ValB -> 0.3, ValC -> 0.2}
 // our delegate logic would attempt to delegate 5osmo to A , 2osmo to B, 3osmo to C
 func (k Keeper) DelegateToValidatorSet(ctx sdk.Context, delegatorAddr string, coin sdk.Coin) error {
-	// get the existingValSet if it exists, if not check existingStakingPosition and return it
+	// get valset formatted delegation either from existing val set prefernce or existing delegations
```
ValarDragon, line 73:
```suggestion
	// get valset formatted delegation either from existing val set preference or existing delegations
```

x/valset-pref/validator_set.go
```diff
@@ -62,10 +88,13 @@ func (k Keeper) DelegateToValidatorSet(ctx sdk.Context, delegatorAddr string, co
 			return err
 		}
 
+		// Matt: If we're truncating here, is it not true that we might be hitting the logic without
+		// fully delegating the given coin?
 		// tokenAmt takes the amount to delegate, calculated by {val_distribution_weight * tokenAmt}
 		tokenAmt := val.Weight.Mul(coin.Amount.ToDec()).TruncateInt()
```
ValarDragon, line 94: Great catch, yeah we need to make a running total, and delegate any surplus to one of the validators.

stackman27: how would i be storing the surplus amount here? for instance if delegateAmt = 1987.986, currently we just delegate 1987, how would i even delegate 0.986 from users Account since Delegate function takes an integer 

`Delegate(ctx sdk.Context, delAddr sdk.AccAddress, bondAmt sdk.Int, tokenSrc stakingtypes.BondStatus, validator stakingtypes.Validator, subtractAccount bool) (newShares sdk.Dec, err error)`

x/valset-pref/validator_set.go
```diff
@@ -62,10 +88,13 @@ func (k Keeper) DelegateToValidatorSet(ctx sdk.Context, delegatorAddr string, co
...
 
 		// TODO: What happens here if validator unbonding
+		// Matt: can we have test case for this to confirm this TODO?
```
stackman27, line 97: I will check the sdk as well about this

x/valset-pref/validator_set.go
```diff
@@ -117,6 +149,9 @@ func (k Keeper) UndelegateFromValidatorSet(ctx sdk.Context, delegatorAddr string
 			return err
 		}
 
+		// Matt: If we're truncating it here, is it not true that we have the possibility of not undelegating 100%
```
stackman27, line 152: I have the same question as delegate here, if we donot truncateInt we cannot how do you recommend getting SharesFromToken which expects an int. 

For all the rounding stuff, i truncated down to make sure we are under delegating/undelegating by 1-2 based denomination than over

x/valset-pref/validator_set.go
```diff
@@ -102,6 +131,9 @@ func (k Keeper) UndelegateFromValidatorSet(ctx sdk.Context, delegatorAddr string
 	}
 
 	// Handles rounding for ex: 9999.99999 = 10000.0
+	// Matt: Can we have test cases that handles this rounding?
```
stackman27, line 134: Yup for this, i think rounding is not really necessary. I removed and added test case

x/valset-pref/validator_set.go
```diff
@@ -170,49 +214,55 @@ func (k Keeper) PreformRedelegation(ctx sdk.Context, delegator sdk.AccAddress, e
 	}
 
 	// calculate the difference between two sets
-	var diffValSet []*valSet
+	var diffValSets []*valSet
+	// Matt: Does this mean that the diffAmount can be negative?
```
stackman27, line 218: yup, the diff amount is intended to contain negative numbers

x/valset-pref/validator_set.go
```diff
@@ -170,49 +214,55 @@ func (k Keeper) PreformRedelegation(ctx sdk.Context, delegator sdk.AccAddress, e
...
-		diffValSet = append(diffValSet, &diff_val)
+		diffValSets = append(diffValSets, &diff_val)
 	}
 
 	// Algorithm starts here
-	for _, diff_val := range diffValSet {
-		for diff_val.amount.TruncateDec().GT(sdk.NewDec(0)) {
-			source_val := diff_val.valAddr
-			target_val, idx := k.FindMin(diffValSet, source_val)
+	// Matt: we need more detailed documnetaiton on the algorithm here
```
stackman27, line 230: the documentation got a bit too large so i added it in `## Redelegate algorithm logic pseudocode` I can link it here

x/valset-pref/validator_set.go
```diff
@@ -170,49 +214,55 @@ func (k Keeper) PreformRedelegation(ctx sdk.Context, delegator sdk.AccAddress, e
...
+	// Matt: We're running in computation of O(n^3) here, two for loops here and then another iteration in the `FindMinAmtValSetExcept`
+	// What's the reason we'we need to find min for every iteration rather than simply using whats in the array in order?
```
stackman27, line 232: It's because after every redelegation there is going to be a new min. for instance 

diff_arr = [ValA: 10, ValB: 6, ValC: 4, ValD: -4, ValE: -4, ValF: -12]

the algorithm runs as follow 
```
loop through diff_arr (only the positive numbers): 
**** 1st iteration*****
    sourceVal = get the first validator Amount (valA in our case) 
    targetVal = Find the minimum amount in the entire diff_arr (valF in our case) 
    redelegate from valA --> valF (10) 
    update valA = 0, valF = -2 

***** 2nd iteration **** 
   sourceVal = get the 2nd validator Value (valB) 
   targetVal = Find the minimum amount in the entire diff_arr (ValD or ValE in our case) lets say we pickD 
   redelegate from valB --> valD (4) 
   update valB = 2, valD = 0 

..... We continue doing until all the validator values [valA..valF is 0, or if the minimum value we find is 0 because if minAmount is 0 there is no more redelegation possible) 

```

x/valset-pref/validator_set.go
```diff
@@ -156,6 +194,12 @@ func (k Keeper) PreformRedelegation(ctx sdk.Context, delegator sdk.AccAddress, e
 
 		tokenFromShares := validator.TokensFromShares(delegation.Shares)
 		existing_val, existing_val_zero_amount := k.GetValSetStruct(existingVals, tokenFromShares)
+		// Matt: Do we need to initiazlize them using this way in the first place?
+		// Here we're creating a copy of the val set struct with zero amount, then adding them to the other struct
+		// to compare the diffs.
+		// Is it not possible to not have this logic in the first place and then perform
+		// diffAmount := newVals.amount.Sub(newValSet[i].amount) right away?
+		// I feel like by using maps this problem can be solvable.
```
stackman27, line 202: yup tried this with map and it works perfectly as well :))

stackman27, line 214: the reason why i didn't use maps here is because we allow duplicate valAddrs as well which maps doesnot allow. For instance

if delegatorA with valSet [valA: 50osmo] wants to redelegate --> [valA: 20osmo, valB: 10osmo, valC: 20osmo]

x/valset-pref/msg_server_test.go
```diff
@@ -339,6 +355,15 @@ func (suite *KeeperTestSuite) TestUnDelegateFromValidatorSet() {
 			setExistingDelegations: true,
 			expectPass:             true,
 		},
+		{
+			name:           "Undelegate extreme amounts",
```
mattverse, line 359: Is this with the intention of indicating extreme amount in terms of truncation? Or the amount itself?

x/valset-pref/validator_set.go
mattverse: Did we ever get back to the necessity to use the min with the trade off of doing computation of O(n^3) vs just using the orders in array to make the computation O(n^2)?

x/valset-pref/validator_set.go
```diff
@@ -96,28 +137,36 @@ func (k Keeper) UndelegateFromValidatorSet(ctx sdk.Context, delegatorAddr string
...
+		// in the last valset iteration we dont calculate it from shares using decimals and trucation,
+		// we use whats remaining to get more accurate value
+		if len(existingSet.Preferences)-1 == i {
+			amountToUnDelegate = coin.Amount.Sub(totalUnDelAmt).ToDec().TruncateInt()
+		} else {
+			// Calculate the amount to undelegate based on the existing weights
+			amountToUnDelegate = val.Weight.Mul(tokenAmt).TruncateInt()
+			totalUnDelAmt = totalUnDelAmt.Add(amountToUnDelegate)
+		}
+
```
mattverse, line 165: test case for this needs to be added

mattverse (commented):
Other concerns and comments has all been addressed, the only part remaining is the redelegation complexity. Upon fixing redelegation complexity this should be good to merge

x/valset-pref/msg_server_test.go
stackman27: yup that is to check if there is any truncation issue

x/valset-pref/validator_set.go
stackman27: yes, i forgot to mention this but i did try without min, however i still have to iterate through the diffSet to find the first negative number. 

I can try to split up diffSet positive and negative numbers but that'll still require looping through both loops which can get complex.

x/valset-pref/validator_set.go
stackman27: yup added 2 that touches this logic (https://github.com/osmosis-labs/osmosis/pull/4143/commits/5c0b35f4e18968470c875431083a988e5a4e0a77#diff-e8d94c443b26e1cda354d150b30ce5542344099e0bbbb8e84e507d9df791daf7R359-R375)

x/valset-pref/msg_server_test.go
mattverse: sweet, noticed that you have clarified the test case name, thanks!

x/valset-pref/validator_set.go
mattverse: I still think this complexity is solvable by creating two arrays : negative array and positive array here https://github.com/osmosis-labs/osmosis/pull/4143/files#diff-4422f5e761e99a0e29d2ba7017ece4aa6b486b5ee24b18fb58fcbc9c26a8e7a6R239-R247 and then using those two arrays instead of having to iterate over them again

stackman27: Issue created for this so that we dont delay v15 https://github.com/osmosis-labs/osmosis/issues/4256

x/valset-pref/validator_set.go
```diff
@@ -170,53 +239,64 @@ func (k Keeper) PreformRedelegation(ctx sdk.Context, delegator sdk.AccAddress, e
...
-				break
-			}
-
-			validator_source, err := sdk.ValAddressFromBech32(source_val)
-			if err != nil {
-				return fmt.Errorf("source validator address not formatted")
+		diffValSets = append(diffValSets, &diff_val)
+	}
+
+	// Algorithm starts here, verbose explanation in README.md
```
stackman27, line 253: Algorithm has been tested in depth in msg_server_test.go, that's why i didn't add a keeper test for it

stackman27 (approved):
Approving this as per our discussion online regarding improving redelegation algorithm post v15

PR #4142, opened 2023-01-30
feat(osmoutils): add GatherValuesFromStorePrefixWithKeyParser function

p0mvn:
Closes: #XXX

## What is the purpose of the change

Completed in collaboration with @RusAkh 

There is a need for a helper function that allows parsing both keys and values when iterating a prefix.

For example, assume we want to get all positions by owner address.

Then, we would iterate the position index with the owner address formatted.

In state, positions have a key `<owner address><pool id><lower tick><upper tick>` and value `<liquidity>`

As a result, the value is not useful on its own and we would like to extract and parse additional data from key while iterating.

This PR adds a helper that allows achieving that. The helper takes a parse function with 2 parameters for key and a value and returns the final desired parsed result.

The pre-exisitng function that parses only value is refactored to reduce code duplication.

## Brief Changelog

- implement `GatherValuesFromStorePrefixWithKeyParse` and `gatherValuesFromIteratorWithKeyParser`
- test
- refactor ` gatherValuesFromIterator` and GatherValuesFromStorePrefix` to reduce code duplication
- `go mod tidy` and `go work sync`

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? yes
  - How is the feature or change documented? not applicable

PR #4145, opened 2023-01-30
refactor(CL): positions by user returns full position data

p0mvn:
Closes: #XXX

Suggested by @stackman27 

Completed in collaboration with @RusAkh 

Positions query currently does not return much important context since most of it is conveyed in the prefix key.

That is, the key contains:
- owner address
- pool id
- lower tick
- upper tick
- frozen until

While the value has only liquidity.

Previously, we did not have a store helper to extract these values from the key while iterating a store prefix.

In https://github.com/osmosis-labs/osmosis/pull/4142. we added such a helper.

This PR focuses on refactoring the `GetUserPositions` function to utilize the new helper to get more data from key when iterating a prefix and return it as part of the result.

Doing so required implementing a custom `ParseFullPositionFromBytes` function that parses both key and a value of a position.

## Testing and Verifying

Covered by existing and refactored test:

Need to add extra test for `ParseFullPositionFromBytes`

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (no


x/concentrated-liquidity/store.go
```diff
@@ -16,10 +19,82 @@ func (k Keeper) getAllPositionsWithVaryingFreezeTimes(ctx sdk.Context, poolId ui
...
 	err = proto.Unmarshal(bz, &position)
 	return position, err
 }
+
+// ParseFullPositionFromBytes parses a full position from key and value bytes.
+// Returns a struct containing the pool id, lower tick, upper tick, frozen until, and liquidity
+// associated with the position.
+// Returns an error if the key or value is not found.
+// Returns an error if fails to parse either.
+// TODO: test
```
p0mvn, line 39: TODO

RusAkh: @p0mvn done

x/concentrated-liquidity/store_test.go
```diff
@@ -75,3 +76,51 @@ func (s *KeeperTestSuite) TestGetAllPositionsWithVaryingFreezeTimes() {
...
+			key:          types.KeyFullPosition(defaultPoolId, defaultAddress, DefaultLowerTick, DefaultUpperTick, defaultFrozenUntil),
+			val:          []byte{},
+			expectingErr: true,
+		},
+		"empty key": {
+			key:          []byte{},
+			val:          cdc.MustMarshal(&model.Position{Liquidity: DefaultLiquidityAmt, FrozenUntil: defaultFrozenUntil}),
+			expectingErr: true,
+		},
+		"sufficient test case": {
```
p0mvn, line 100: Please also add tests where we format incorrect values for each of the fields. For example, pool id is not uint but a sequence of characters. As a result, it fails to be parsed

RusAkh: sounds sense

x/concentrated-liquidity/store_test.go
```diff
@@ -75,3 +77,77 @@ func (s *KeeperTestSuite) TestGetAllPositionsWithVaryingFreezeTimes() {
...
+			val:          cdc.MustMarshal(&model.Position{Liquidity: DefaultLiquidityAmt, FrozenUntil: defaultFrozenUntil}),
+			expectingErr: true,
+		},
+		"One key separator missing in key": {
+			key:          []byte(fmt.Sprintf("%s%s%s%d%s%d%s%d", []byte{0x02}, defaultAddress, "|", defaultPoolId, "|", DefaultLowerTick, "|", DefaultUpperTick)),
+			val:          cdc.MustMarshal(&model.Position{Liquidity: DefaultLiquidityAmt, FrozenUntil: defaultFrozenUntil}),
+			expectingErr: true,
+		},
+		"Wrong position prefix": {
+			key:          []byte(fmt.Sprintf("%s%s%s%s%d%s%d%s%d", []byte{0x01}, "|", defaultAddress, "|", defaultPoolId, "|", DefaultLowerTick, "|", DefaultUpperTick)),
```
p0mvn, line 117: Can we replicate this for each of pool id, lower tick, upper tick, and frozen until, please?

This is to ensure full coverage since we attempt to parse each of those fields and have an error check for it.

stackman27 (commented):
This is crazy!!  I never thought it was possible to do this without changing position.proto. Thank you so much @p0mvn

stackman27 (approved):
LGTM! Manually tested as well

x/concentrated-liquidity/store_test.go
RusAkh: hello @p0mvn. Changes addressed

x/concentrated-liquidity/store.go
```diff
@@ -16,10 +19,85 @@ func (k Keeper) getAllPositionsWithVaryingFreezeTimes(ctx sdk.Context, poolId ui
...
+	positionValue, err := ParsePositionFromBz(value)
+	if err != nil {
+		return types.FullPositionByOwnerResult{}, err
+	}
+
+	return types.FullPositionByOwnerResult{
+		PoolId:      poolId,
+		LowerTick:   lowerTick,
+		UpperTick:   upperTick,
+		FrozenUntil: frozenUntil,
```
RusAkh, line 100: What do you think about adding a check that `frozenUntil` which we get from the key is equal to `frozenUntil` in `positionValue`? Just a sanity check. I hesitate is we should add it, because it sounds a little bit redundant

@p0mvn

p0mvn: I think we only have liquidity in the value, there should be no `frozenUntil` as far as I know

RusAkh: yeah, my idea was to, like, make sure the position in the value corresponds with the key. And by comparing `frozenUntil` in both at least, we know up to some degree that it is the correct value.

RusAkh: and because if `frozenUntil`s are not the same in key and value, we can't be sure that the liquidity is correct in the value 
The reason I thought it is redundant: if a store provides a guarantee that there is absolutely no way for key with lets say `frozenUntil = 5` to return `frozenUntil = 10` value, then yes my suggestion is redundant

RusAkh: I hope that makes sense

p0mvn: Ah I see. I think we do not need to have that in both key and value.

I think we should merge this PR as well but investigate separately.

Having `frozenUntil` in both key and value is redundant. If we use it in key, we can also extract it from key.

CC: @czarcas7ic @AlpinYukseloglu

p0mvn:
Merging to unblock CLI and e2e work. Thanks for driving this to completion @RusAkh .

The comment about validating `frozenUntil` in key and value will be addressed separately. I raised this issue with the CL team

PR #4151, opened 2023-01-30
Allow explicit channels on XCS receiver

nicolaslara:

Allows overloading the received for users that want to explicitly specify which IBC channels and addr their XCS packets should be sent to

## Brief Changelog

 
  - Receiver now supports new syntax: `"ibc:channel-n/osmo1..."

## Testing and Verifying

Split the outpost test into two: one with the simplified receivers and one with the explicit ones. 


cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -1,11 +1,32 @@
 use cosmwasm_std::{Addr, Deps};
+use regex::Regex;
 
 use crate::{state::CHANNEL_MAP, ContractError};
 
-// Validate that the receiver address is a valid address for the destination chain.
-// This will prevent IBC transfers from failing after forwarding
-pub fn validate_receiver(deps: Deps, receiver: Addr) -> Result<(String, Addr), ContractError> {
-    let Ok((prefix, _, _)) = bech32::decode(receiver.as_str()) else {
+/// If the specified receiver is an explicity channel+addr, extract the parts
```
p0mvn, line 6:
```suggestion
/// If the specified receiver is an explicitly channel+addr, extract the parts
```

cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -1,11 +1,32 @@
...
+/// and use the strings as provided
+fn validate_explicit_receiver(receiver: &str) -> Result<(String, Addr), ContractError> {
+    let re = Regex::new(r"^ibc:(channel-\d+)/(.+)$").unwrap();
```
p0mvn, line 9: nit/q: can this be global to avoid initializing it every time this is run?

cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -1,11 +1,32 @@
...
+
+    let (channel, address) = (&captures[1], &captures[2]);
+    let Ok(_) = bech32::decode(address) else {
+        return Err(ContractError::InvalidReceiver { receiver: receiver.to_string() })
+    };
+
+    Ok((channel.to_string(), Addr::unchecked(address)))
+}
+
+/// If the specified received is not explicit, validate that the receiver
```
p0mvn, line 22:
```suggestion
/// If the specified receiver is not explicit, validate that the receiver
```

p0mvn (approved):
LGTM

cosmwasm/contracts/crosschain-swaps/src/checks.rs
ValarDragon: Yeah agreed, I think its important to have the regex compiled only once.

ValarDragon: Oh this might not be so simple: https://stackoverflow.com/questions/41503880/why-isnt-regex-a-wrapper-for-regexnew-to-offer-the-same-regex-matching-s

ValarDragon: Not seeing the regex! macro in latest docs, so maybe it just works as a global?

nicolaslara: It seems like the regex macro (1) didn't do what we want and (2) is gone: https://github.com/rust-lang/regex/blob/master/HACKING.md#the-regex-macro 

Using lazy_static! shouldn't help here, as we only call this once and IIUC the contract code gets loaded on each execution. 

I can replace the regex with manual checking. This is not hard and it's what I had before but I changed it to a regex to make the code shorter. it's basically: remove the `ibc:` prefix and check that it's there, split the string by `/` and check that the first part matches channel-x, use the rest as the denom (assuming addreses can't contain `/`)

nicolaslara: Found [this](https://news.ycombinator.com/item?id=27301320) which seems promising too, but it's not documented/standard, so we'd probably need to review the code.

ValarDragon: Cosmwasm tooling question -- if this doesn't exist, lets ask Boss to see if we can add it into beaker.

Can we use printlns to measure how much gas this call takes? This is really just a gas optimization question

ValarDragon: Lets make this a followup issue as gas optimization.

ValarDragon (approved):

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: Added https://github.com/osmosis-labs/osmosis/issues/4197 for this.

Issue #4149, opened 2023-01-30
(CL): Add `TickRangeLiquidityDepth` Query for Better FrontEnd Support


mattverse:
## Background

We want to have the following query to better support Front-end (cc. @jonator )

- Within a price range(or above and below the current price if given one hi or lo price value), return liquidity for each tick

Namely, `TickRangeLiquidityDepth`

## Suggested Design
- Create an iterator method that takes in start tick and end tick, use their prefixes and which then returns `sdk.Iterator`
- Create a secondary method using the iterator method created above to iterate over all the ticks within the range and then put them in a separate array
- Add protos for query that would look something like this
   - req: upper tick, lower tick (can be nil)
   - res: array of stuct that contains a) tick Index b) liquidity(sdk.Int)

## Blockers
We also need to think about how to handle full range positions. Iterating over **all** ticks would definitely be too much unnecessary computation. We need to think of a way to get the liquidity existing from tick A~ tick B without having to iterate overall ticks, and then we can use that method to return with longer ranges when we're further away from the current spot price

## Acceptance Criteria
- The query has been added along with necessary test cases
-  Iterator method and test cases has been implemented
- We want to double assure this is not a burden in terms of computation, a benchmark test for this query would be great

ValarDragon:
So what exactly is "liquidity depth"? Is it, for a price range `[p1, p2]` and current price `p`, how much needs to be bought/sold to get to `p1` and `p2` respectively?

jonator:
Yes but per individual tick, or a subset of ticks

The horizontal bars here, with vertical axis being price:

![image (5)](https://user-images.githubusercontent.com/4606373/216021020-714bf54c-0d1c-41e4-8f38-1ee9fa0f28cf.png)

PR #4167, opened 2023-01-31
(CL): Add `LiquidityDepthsForRange` Query

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/4149

This PR adds query `LiquidityDepthsForRange` where given lower and upper tick, we would return a array of struct that contains a) tick index b) tick liquidity net gross.

A notable point on this query is that we have maxRange of the ticks provided. That is, when the max Range is 10,000 and the diff between provided lower tick and upper tick is 20,000, this query would error out. This is designed like this to prevent heavy computations. It's now capped at 10,000, but open to discussions on what this cap should be.

(cc @jonator for further feedbacks on this PR as well)

## Brief Changelog
- Add query for `LiquidityDepthsForRange`

## Testing and Verifying
- Added tests for the iterator method that the querier would call for the query 


mattverse:
@jonator Do you see pagination necessary for this query? Asking since applying pagination to this would take some effort 
 (we dont have pagination support for ranges that are designated in cosmos sdk, would have to make a separate implementation for this pagination type)

x/concentrated-liquidity/grpc_query.go
```diff
@@ -102,3 +102,30 @@ func (q Querier) Params(goCtx context.Context, req *types.QueryParamsRequest) (*
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	return &types.QueryParamsResponse{Params: q.Keeper.GetParams(ctx)}, nil
 }
+
+// LiquidityDepthsForRange returns liquidity depths for the given range.
+func (q Querier) LiquidityDepthsForRange(goCtx context.Context, req *types.QueryLiquidityDepthsForRangeRequest) (*types.QueryLiquidityDepthsForRangeResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	maxRange := sdk.NewInt(10000)
```
czarcas7ic, line 109: if we decide to internally hard cap this I think it should live in a const somewhere

x/concentrated-liquidity/grpc_query.go
```diff
@@ -102,3 +102,30 @@ func (q Querier) Params(goCtx context.Context, req *types.QueryParamsRequest) (*
...
+
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	if req.LowerTick.GT(req.UpperTick) {
+		return nil, status.Error(codes.InvalidArgument, "lower tick is greater than upper tick")
+	}
```
czarcas7ic, line 117: We might have an error status already premade in errors.go (that also returns what the lower and upper tick that was provided for clarity)

x/concentrated-liquidity/grpc_query.go
```diff
@@ -102,3 +102,30 @@ func (q Querier) Params(goCtx context.Context, req *types.QueryParamsRequest) (*
...
+
+	if req.UpperTick.Sub(req.LowerTick).GT(maxRange) {
+		return nil, status.Error(codes.InvalidArgument, "price range given is too big")
```
czarcas7ic, line 120: Can we return max vs given range in the error message?

x/concentrated-liquidity/tick.go
```diff
@@ -142,3 +144,47 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
 func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTick, maxTick int64) {
 	return math.GetMinAndMaxTicksFromExponentAtPriceOneInternal(exponentAtPriceOne)
 }
+
+// getTickLiquidityDepthFromIterator uses the given lower tick and upper tick, iterates over ticks, creates and returns LiquidityDepth array.
+// LiquidityNet from the tick is used to indicate liquidity depths.
+func (k Keeper) GetTickLiquidityDepth(ctx sdk.Context, poolId uint64, lowerTick, upperTick int64) ([]types.LiquidityDepth, error) {
```
czarcas7ic, line 150: nit: Maybe `GetPerTickLiquidityDepthFromRange` might be a better, more descriptive name? If that is too long, maybe just `GetPerTickLiquidityDepth`?

x/concentrated-liquidity/tick_test.go
```diff
@@ -433,3 +435,119 @@ func (s *KeeperTestSuite) TestCrossTick() {
...
+		TickIndex:    sdk.NewInt(4),
+		LiquidityNet: sdk.NewDec(40),
+	}
+	tests := []struct {
+		name                   string
+		invalidPool            bool
+		expectedErr            bool
+		lowerTick              int64
+		upperTick              int64
+		exectedLiquidityDepths []types.LiquidityDepth
```
czarcas7ic, line 462:
```suggestion
		expectedLiquidityDepths []types.LiquidityDepth
```

x/concentrated-liquidity/tick_test.go
```diff
@@ -433,3 +435,119 @@ func (s *KeeperTestSuite) TestCrossTick() {
...
+			},
+		},
+		{
+			name:                   "tick that does not exist",
+			lowerTick:              10,
+			upperTick:              10,
+			exectedLiquidityDepths: []types.LiquidityDepth{},
+		},
+		{
+			name:        "Half range",
```
czarcas7ic, line 499: I think this name is wrong and should be something like "non-existent pool"

czarcas7ic (commented):
Looks good on first pass! Added some comments to be addressed, and not approving yet since I want to think over the design a little more.

jonator:
> @jonator Do you see pagination necessary for this query? Asking since applying pagination to this would take some effort (we dont have pagination support for ranges that are designated in cosmos sdk, would have to make a separate implementation for this type of pagination)

Pagination, via keys and perhaps a direction (for this custom case) would simplify frontends, however this is somewhat of an unconventional use of pagination so I could see it being excluded. For the frontend, I'm thinking we'd fetch the ticks in the direction the user is currently inspecting in the swap tool. Nothing stops them from entering very large numbers, so we could be making a lot of requests until the pool runs out of liquidity. The 10k limit should be a good balance between size vs number of requests.

jonator:
Q: Also, back to our sync call, how would the frontend provide the right tick IDs/numbers? How would the frontend know the ideal amount of ticks to request, are they the same across pools? Would it always be some number that doesn't cause too large of request bodies?

mattverse:
> > @jonator Do you see pagination necessary for this query? Asking since applying pagination to this would take some effort (we dont have pagination support for ranges that are designated in cosmos sdk, would have to make a separate implementation for this type of pagination)
> 
> Pagination, via keys and perhaps a direction (for this custom case) would simplify frontends, however this is somewhat of an unconventional use of pagination so I could see it being excluded. For the frontend, I'm thinking we'd fetch the ticks in the direction the user is currently inspecting in the swap tool. Nothing stops them from entering very large numbers, so we could be making a lot of requests until the pool runs out of liquidity. The 10k limit should be a good balance between size vs number of requests.

Sweet, then seems like we'll be excluding pagination as the scope of this query. Please let me know after trying out suff if we need pagination, we can always add them back in when needed

mattverse:
> Q: Also, back to our sync call, how would the frontend provide the right tick IDs/numbers? How would the frontend know the ideal amount of ticks to request, are they the same across pools? Would it always be some number that doesn't cause too large of request bodies?

Question regarding how to provide the right tick number: we would have a query to query the current tick of the pool. Most likely plus and minus around the current tick.

mattverse:
> Q: Also, back to our sync call, how would the frontend provide the right tick IDs/numbers? How would the frontend know the ideal amount of ticks to request, are they the same across pools? Would it always be some number that doesn't cause too large of request bodies?

Question regarding the ideal amount of ticks to request: So AFAIU, this query is going to be used for supporting swaps around spot prices. If that's the case I would say like at most 10~100, would get back to you with a better answer

x/concentrated-liquidity/tick_test.go
mattverse: Thanks for the catch! Forgot to push my commit of editing this, resolved in the most recent commit

x/concentrated-liquidity/tick_test.go
mattverse: Resolved in most recent commit

x/concentrated-liquidity/grpc_query.go
mattverse: defined a constant at the top of `grpc_query.go`, lmk how it works out

x/concentrated-liquidity/grpc_query.go
mattverse: Changed to the error suggested!

x/concentrated-liquidity/grpc_query.go
mattverse: Added a new err type to support this in the latest commit

x/concentrated-liquidity/types/query.pb.go
```diff
@@ -302,58 +303,221 @@ func (m *QueryParamsResponse) GetParams() Params {
...
+func (m *QueryLiquidityDepthsForRangeRequest) GetPagination() *query.PageRequest {
+	if m != nil {
+		return m.Pagination
+	}
+	return nil
+}
+
+type QueryLiquidityDepthsForRangeResponse struct {
+	LiquidityDepths []LiquidityDepth    `protobuf:"bytes,1,rep,name=liquidity_depths,json=liquidityDepths,proto3" json:"liquidity_depths"`
+	Pagination      *query.PageResponse `protobuf:"bytes,2,opt,name=pagination,proto3" json:"pagination,omitempty"`
```
jonator, line 363: What is this pagination object used for if we're not using pagination? Is this pagination within the same range?

RE:
https://github.com/osmosis-labs/osmosis/pull/4167#issuecomment-1411488545

mattverse:
@czarcas7ic Resolved all comments. Also, AFAIR, liquidity net from ticks is all we need to calculate swaps can you confirm if this is correct?

x/concentrated-liquidity/grpc_query.go
czarcas7ic: Oh I had meant constants.go https://github.com/osmosis-labs/osmosis/blob/d28c7412c8d16ed48a20f7b770185d238d83f12b/x/concentrated-liquidity/types/constants.go#L1

x/concentrated-liquidity/tick.go
```diff
@@ -142,3 +144,47 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
 func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTick, maxTick int64) {
 	return math.GetMinAndMaxTicksFromExponentAtPriceOneInternal(exponentAtPriceOne)
 }
+
+// getTickLiquidityDepthFromIterator uses the given lower tick and upper tick, iterates over ticks, creates and returns LiquidityDepth array.
```
czarcas7ic, line 148:
```suggestion
// GetPerTickLiquidityDepthFromRange uses the given lower tick and upper tick, iterates over ticks, creates and returns LiquidityDepth array.
```

x/concentrated-liquidity/tick.go
```diff
@@ -142,3 +144,47 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
...
+	}
+	store := ctx.KVStore(k.storeKey)
+	prefixBz := types.KeyTickPrefix(poolId)
+	prefixStore := prefix.NewStore(store, prefixBz)
+
+	lowerKey := types.TickIndexToBytes(lowerTick)
+	upperKey := types.TickIndexToBytes(upperTick)
+	iterator := prefixStore.Iterator(lowerKey, storetypes.InclusiveEndBytes(upperKey))
+
+	liquidityDepths := []types.LiquidityDepth{}
```
czarcas7ic, line 162: We should be able to pre-size this array right? If we don't, and we are iterating over the max, we will be resizing this many many times

czarcas7ic (approved):
This LGTM! Please take a look at the array sizing point I made in the most recent review though before merging.

x/concentrated-liquidity/types/query.pb.go
mattverse: this has been removed

x/concentrated-liquidity/tick_test.go
```diff
@@ -433,3 +435,119 @@ func (s *KeeperTestSuite) TestCrossTick() {
...
+	tests := []struct {
+		name                    string
+		invalidPool             bool
+		expectedErr             bool
+		lowerTick               int64
+		upperTick               int64
+		expectedLiquidityDepths []types.LiquidityDepth
+	}{
+		{
+			name:      "Full range",
```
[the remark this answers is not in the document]
mattverse: good point!

x/concentrated-liquidity/grpc_query.go
```diff
@@ -121,3 +125,33 @@ func (q Querier) Params(goCtx context.Context, req *types.QueryParamsRequest) (*
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	return &types.QueryParamsResponse{Params: q.Keeper.GetParams(ctx)}, nil
 }
+
+// LiquidityDepthsForRange returns liquidity depths for the given range.
+func (q Querier) LiquidityDepthsForRange(goCtx context.Context, req *types.QueryLiquidityDepthsForRangeRequest) (*types.QueryLiquidityDepthsForRangeResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// use constant pre-defined to limit range
+	maxRange := sdk.NewInt(liquidityDepthRangeQueryLimit)
```
p0mvn, line 134: nit: let's define this as a package-global variable to avoid re-allocations in the constructor every time the query is called

x/concentrated-liquidity/tick.go
```diff
@@ -142,3 +144,47 @@ func validateTickRangeIsValid(tickSpacing uint64, exponentAtPriceOne sdk.Int, lo
...
+
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
```
p0mvn, line 165: Can we use this store helper here instead?

https://github.com/osmosis-labs/osmosis/blob/ef1947527e612fbf9cd1532ca86c34fd342e9931/osmoutils/store_helper.go#L110

x/concentrated-liquidity/tick.go
mattverse: The problem is that we can't pre-size accurately enough, since we do not know how much value the iterator holds. 
A possible solution to your concern can be creating a pre-sized array that has a really really huge size. I don't know if this would be a better solution, let me know what you think!

x/concentrated-liquidity/grpc_query.go
mattverse: resolved in the most recent commit

x/concentrated-liquidity/tick.go
mattverse: We need the key of the iterator, not the values here, that's why I did not use any iterator helper in store, lmk if you have suggestions regarding this

jonator (approved):

Issue #4180, opened 2023-01-31
Panic Handling In MultihopEstimateOutGivenExactAmountIn borks Protorev


NotJeremyLiu:
## Background

Panics are caught when swapping (https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/swap.go#L39-L44) but not when estimating a swap (https://github.com/osmosis-labs/osmosis/blob/d28c7412c8d16ed48a20f7b770185d238d83f12b/x/poolmanager/router.go#L133). 

The specific panic that brought this to our attention for Protorev was in stableswap `solveCFMMBinarySearchMulti`: (https://github.com/osmosis-labs/osmosis/blob/d28c7412c8d16ed48a20f7b770185d238d83f12b/x/gamm/pool-models/stableswap/amm.go#L315), but generally any of these panics could likely cause an issue if `MultihopEstimateOutGivenExactAmountIn` does not catch the panics.

## Suggested Design

We are open to different forms of resolution, whether that may be core team adding panic catching in `MultihopEstimateOutGivenExactAmountIn` or us catching panics in Protorev, but we are waiting for core team decision on best way forward.

## Acceptance Criteria

All current protorev tests pass, and adding new tests to purposefully trigger the panics does not result in the actual panic to cause a halt.

ValarDragon:
I think we should definitely just catch this panic -- quite silly for estimate to panic haha

PR #4181, opened 2023-01-31
Protorev smarter logic and more testing

NotJeremyLiu:

This PR adds more testing to `x/protorev` and improves the arb calculation logic to be more efficient.

## Brief Changelog
 
  - Given a route, we now calculate if the minAmountIn is profitable as a pre-check. If minAmountIn is not profitable, then the route is not profitable and we don't run the full binary search algo. This will make Protorev take much less compute time in production compared to its previous design (which would run the full binary search even when a route has no-arb). We only increment the pool point counter if the trade is profitable, thereby saving us compute for swaps that create no backrunning opportunity.
  - Given a route, we now calculate if the maxAmountIn and maxAmountIn + 1 profit is increasing as a pre-check and increase the range if so. If the profit from maxInputIn is still increasing, this means that we need to increase the range to capture the max profit.
  - Adds tests that verifies the limit/point system works properly, >3 pool routes work, and other denom besides osmo/atom work.
  - Deprecate hot routes from being included in genesis state for simplicity. 
  - Add E2E testing that 
    - Checks that the module is correctly init on start
    - Checks that the module can correctly spot a back running opportunity after a swap
    - Checks that the module executes a profitable cyclic arbitrage trade
    - Checks that the module did not mint or alter the supplies from the bank
 - Add support for gamm swap msgs
 - Deprecate atom from the list of base denoms that are used on genesis for route creation

## Testing and Verifying

- All previous tests pass with new changes
- Added a test to ensure range extension works properly
- Added a test for swapExactAmountOut
- Added benchmark testing
- Added doomsday testing to ensure point system works properly and doesn't allow protorev to keep running
- Adds tests for generalized denoms
- Adds a test to ensure panic catching is handled properly. This currently FAILS. This is on purpose, this PR is not to be merged until https://github.com/osmosis-labs/osmosis/issues/4180 is resolved
- Add test for gamm msg swaps
- Add tests for E2E (as described above)


x/protorev/keeper/rebalance.go
```diff
@@ -94,28 +104,50 @@ func (k Keeper) EstimateMultihopProfit(ctx sdk.Context, inputDenom string, amoun
...
+	inputDenom := route.Route[route.Route.Length()-1].TokenOutDenom
+
+	// If a cyclic arb exists with an optimal amount in above our minimum amount in,
+	// then inputting the minimum amount in will result in a profit. So we check for that first.
+	// If there is no profit, then we can return early and not run the binary search.
+	_, minInProfit, err := k.EstimateMultihopProfit(ctx, inputDenom, curLeft.Mul(route.StepSize), route.Route)
+	if err != nil {
+		return sdk.Coin{}, sdk.ZeroInt(), err
+	} else if minInProfit.LTE(sdk.ZeroInt()) {
+		return sdk.Coin{}, sdk.ZeroInt(), fmt.Errorf("no profit for route %d", route.Route.PoolIds())
```
NotJeremyLiu, line 126: I imagine a majority of small swaps will result in no profits, so this will throw an error often. Feels kinda like we want to view no profit in a route as a non-error. maybe we log in some other way if we want to log?

NotJeremyLiu (commented):
Reviewed, left 1 comment about the changes made. Going to fix up the doomsday test given the new point refund logic.

NotJeremyLiu:
Another thing I noticed is that we still have a hardcoded denom for atom: https://github.com/osmosis-labs/osmosis/blob/e6c1ba6006a3ed2a414e14e1ad3e53b814d83f81/x/protorev/types/constants.go#L8

This is no longer used in protorev itself, but used throughout our test suite. Is there a way for us to move this easily to only be for test suite and not in any of the core module files (to avoid changing all the tests that reference it, but not have any unnecessary code in core module files)

x/protorev/keeper/rebalance.go
NotJeremyLiu: Resolved in ad5ea732f09ea71e7433825dba9f6523223c524d

davidterpay:
> Another thing I noticed is that we still have a hardcoded denom for atom:
> 
> https://github.com/osmosis-labs/osmosis/blob/e6c1ba6006a3ed2a414e14e1ad3e53b814d83f81/x/protorev/types/constants.go#L8
> 
> This is no longer used in protorev itself, but used throughout our test suite. Is there a way for us to move this easily to only be for test suite and not in any of the core module files (to avoid changing all the tests that reference it, but not have any unnecessary code in core module files).
> 
> I think this is a non-issue though

Update, I deprecated it from the module.

proto/osmosis/protorev/v1beta1/genesis.proto
```diff
@@ -11,6 +11,4 @@ option go_package = "github.com/osmosis-labs/osmosis/v14/x/protorev/types";
 message GenesisState {
   // Module Parameters
   Params params = 1 [ (gogoproto.nullable) = false ];
-  // Hot routes that are configured on genesis
-  repeated TokenPairArbRoutes token_pairs = 2 [ (gogoproto.nullable) = false ];
```
davidterpay, line 15: deprecating the use of hot routes in genesis since we will not be using it.

tests/e2e/initialization/config.go
```diff
@@ -52,24 +52,29 @@ const (
...
 	StakeIBCDenom       = "ibc/C053D637CCA2A2BA030E2C5EE1B28A16F71CCB0E45E8BE52766DC1B241B7787"
+	UstIBCDenom         = "ibc/BE1BB42D4BE3C30D50B68D7C41DB4DFCE9678E8EF8C539F6E6A9345048894FCC"
+	LuncIBCDenom        = "ibc/0EF15DF2F02480ADE0BB6E85D9EBB5DAEA2836D3860E9F97F9AADE4F57A31AA0"
 	MinGasPrice         = "0.000"
 	IbcSendAmount       = 3300000000
 	ValidatorWalletName = "val"
 	// chainA
 	ChainAID      = "osmo-test-a"
-	OsmoBalanceA  = 200000000000
+	OsmoBalanceA  = 20000000000000
```
davidterpay, line 62: osmo balance being increased so that the pools that protorev testing creates have enough funds

tests/e2e/initialization/config.go
```diff
@@ -52,24 +52,29 @@ const (
 	AtomDenom           = "uatom"
 	OsmoIBCDenom        = "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518"
 	StakeIBCDenom       = "ibc/C053D637CCA2A2BA030E2C5EE1B28A16F71CCB0E45E8BE52766DC1B241B7787"
+	UstIBCDenom         = "ibc/BE1BB42D4BE3C30D50B68D7C41DB4DFCE9678E8EF8C539F6E6A9345048894FCC"
```
p0mvn, line 55: Hi @davidterpay . Please manually backport just this file to `v14.x` branch.

Once merged, please ping me, and I will give you an image tag that will need to be put here: https://github.com/osmosis-labs/osmosis/blob/dd1d910d5a0fc4a81050690e7d0e84158faebc3d/tests/e2e/containers/config.go#L30

We do this to be able to build genesis and config files from v14.x to start up Osmosis containers of that version and perform the upgrade to the next (v15) version in your branch.

proto/osmosis/protorev/v1beta1/genesis.proto
ValarDragon: Hrm, I feel like it should be added back (in a subsequent non-release affecting PR), so that ExportGenesis -> ImportGenesis still works. (So state export based testnets for instance can use this)

The ideal is that for `state' = import(export(state))`, that `state'` is effectively the same as `state`.

x/protorev/genesis.go
```diff
@@ -42,32 +42,29 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
...
 		},
 	}
-	if err := k.SetBaseDenoms(ctx, baseDenomPriorities); err != nil {
+	if err := k.SetBaseDenoms(ctx, baseDenoms); err != nil {
 		panic(err)
 	}
 
-	// Update the pools on genesis
-	if err := k.UpdatePools(ctx); err != nil {
+	// Setting the admin account by default to a trusted address
```
ValarDragon, line 58: Can we document the source of this address? (Perhaps link to the commonwealth post once it exists)

davidterpay: Yup, will do. Just for reference this is an account we created and are storing on a ledger. We will use it to enable authz (which we have already test) for a different account so that the admin keys are safely stored.

x/protorev/keeper/epoch_hook_test.go
```diff
@@ -4,9 +4,28 @@ import (
...
+	b.ResetTimer()
+
+	// Setup the test suite
+	suite := new(KeeperTestSuite)
+	suite.SetT(&testing.T{})
+	suite.SetupTest()
+
+	for i := 0; i < b.N; i++ {
+		b.StartTimer()
+		suite.App.ProtoRevKeeper.UpdatePools(suite.Ctx)
```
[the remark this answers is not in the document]
davidterpay: Atm, the test suite sets up ~50 pools. I attached the benchmark results to the comment. if this scales linearly and we assume mainnet has around ~1000 pools, the execution time of `UpdatePools` should be around 5 ms in total.
Screenshot 2023-02-16 at 12 27 14 PM

tests/e2e/configurer/chain/queries.go
```diff
@@ -24,10 +24,177 @@ import (
...
+	if err != nil {
+		return false, err
+	}
+
+	// nolint: staticcheck
+	var response protorevtypes.QueryGetProtoRevEnabledResponse
+	err = util.Cdc.UnmarshalJSON(bz, &response)
+	require.NoError(n.t, err) // this error should not happen
+	return response.Enabled, nil
+}
```
ValarDragon, line 196: Wow we really need to add some significant E2E side automation if we have to be doing this for every query...

x/protorev/keeper/keeper_test.go
```diff
@@ -686,7 +785,64 @@ func (suite *KeeperTestSuite) setUpPools() {
...
+				sdk.NewCoin("usdc", sdk.NewInt(1000000000000000)),
+				sdk.NewCoin("usdt", sdk.NewInt(1000000000000000)),
+				sdk.NewCoin("busd", sdk.NewInt(2000000000000000)),
+			),
+			poolParams: stableswap.PoolParams{
+				SwapFee: sdk.NewDecWithPrec(1, 4),
+				ExitFee: sdk.NewDecWithPrec(0, 2),
+			},
+			scalingFactors: []uint64{1, 1, 1},
+		},
```
ValarDragon, line 800: Thanks for adding all these pools to whats protorev configured in the tests!

x/protorev/keeper/rebalance.go
```diff
@@ -9,40 +9,48 @@ import (
...
 	var maxProfitInputCoin sdk.Coin
 	maxProfit := sdk.ZeroInt()
 
-	for _, route := range routes {
-		// Find the max profit for the route using the token out denom of the last pool in the route as the input token denom
-		inputDenom := route[len(route)-1].TokenOutDenom
-		inputCoin, profit, err := k.FindMaxProfitForRoute(ctx, route, inputDenom)
-		if err != nil {
-			k.Logger(ctx).Error("Error finding max profit for route: ", err)
+	// Get the total number of pool points that can be consumed in this transaction
```
ValarDragon, line 17: Nice job on all the code comments in this file, quite easy to follow the logic :)

x/protorev/keeper/rebalance.go
```diff
@@ -94,28 +102,50 @@ func (k Keeper) EstimateMultihopProfit(ctx sdk.Context, inputDenom string, amoun
...
 
-	for curLeft.LT(curRight) && iteration < types.MaxIterations {
-		iteration++
+	// Input denom used for cyclic arbitrage
+	inputDenom := route.Route[route.Route.Length()-1].TokenOutDenom
+
+	// If a cyclic arb exists with an optimal amount in above our minimum amount in,
+	// then inputting the minimum amount in will result in a profit. So we check for that first.
+	// If there is no profit, then we can return early and not run the binary search.
+	_, minInProfit, err := k.EstimateMultihopProfit(ctx, inputDenom, curLeft.Mul(route.StepSize), route.Route)
```
ValarDragon, line 120: Nice!

ValarDragon (approved):
Nice job on this! LGTM to merge!

x/protorev/keeper/epoch_hook_test.go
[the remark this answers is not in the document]
ValarDragon: Ah awesome! Didn't realize this was the same setup test with everything else, that makes sense!

tests/e2e/configurer/chain/queries.go
ValarDragon: #4350

PR #4185, opened 2023-01-31
[CL: Simulator] Simulator for CreatePosition, WithdrawPosition, CollectFees

stackman27:
Part of: https://github.com/osmosis-labs/osmosis/issues/4027

Simulator for CreatePosition, WithdrawPosition, CollectFees

## Brief Changelog
n/a

## Testing and Verifying
sim tested


x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +193,59 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+
+	return concentratedPool, poolI, poolDenoms, err
+}
+
+// getRandomTickPositions returns random lowerTick and upperTick divisible by tickSpacing value.
+func getRandomTickPositions(minTick, maxTick int64, tickSpacing uint64) (int64, int64, error) {
+	lower_tick_disivisble := ((maxTick - minTick) / int64(tickSpacing)) + 1 // get random value between minTick, maxTick that is divisible by TickSpacing
+	if lower_tick_disivisble < 1 {
+		return 0, 0, fmt.Errorf("lower tick divisible by tickspacing not found")
+	}
```
czarcas7ic, line 233: Might want to rewrite / add more comments here, it took me a while to figure out how it was working.

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +61,128 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+
+	// pick a random position
+	randPosition := positions[rand.Intn(len(positions))]
+
+	// Randomize tick values from minTick to maxTick
+	minTick, maxTick := clkeeper.GetMinAndMaxTicksFromExponentAtPriceOne(clPool.GetPrecisionFactorAtPriceOne())
+
+	if randPosition.LowerTick < minTick || randPosition.UpperTick > maxTick {
+		return nil, fmt.Errorf("invalid ticks")
+	}
```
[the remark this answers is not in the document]
stackman27: yup that works too!

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +61,128 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	// get random user address with the pool denoms
+	sender, tokens, senderExists := sim.SelAddrWithDenoms(ctx, poolDenoms)
+	if !senderExists {
+		return nil, fmt.Errorf("no sender with denoms %s exists", poolDenoms)
+	}
+
+	// ensure that we always have 2 tokens
+	if len(tokens) < 2 {
+		return nil, fmt.Errorf("no pool denoms tokens")
+	}
```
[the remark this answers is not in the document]
stackman27: it returns a random subset of poolDenoms, so i had to add this assertion

stackman27:
Also wanted to post the sim results, i think its a fairly good outcome 
```
 "concentratedliquidity": {
  "CollectFees": {
   "failure": 28,
   "ok": 93
  },
  "CreateConcentratedPool": {
   "failure": 2,
   "ok": 121
  },
  "CreatePosition": {
   "failure": 29,
   "ok": 109
  },
  "WithdrawPosition": {
   "failure": 60,
   "ok": 43
  }
```

stackman27:
@czarcas7ic Addressed feedbacks. Ready for Re-Review

x/concentrated-liquidity/simulation/sim_msgs.go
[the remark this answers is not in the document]
czarcas7ic: Ahh it returns a random coin subset, gotcha

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +60,114 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+		return nil, err
+	}
+
+	// get random user address with the pool denoms
+	sender, tokens, senderExists := sim.SelAddrWithDenoms(ctx, poolDenoms)
+	if !senderExists {
+		return nil, fmt.Errorf("no sender with denoms %s exists", poolDenoms)
+	}
+
+	// ensure that we have atleast 2 tokens
```
czarcas7ic, line 76:
```suggestion
	// ensure that we have at least 2 tokens
```
Would maybe mention something about this check being necessary since SelAddrWithDenoms selects a random subset of denoms. 

Sidenote, is there not a function that selects and address with denoms and returns the ENTIRE denom subset? If so, I feel like thats what we would want to use here. If not, it shouldn't be hard to create.

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +60,114 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	if len(tokens) < 2 {
+		return nil, fmt.Errorf("not enough pool denoms tokens, require 2 got: %d", len(tokens))
+	}
+
+	// Randomize tick values from minTick to maxTick
```
czarcas7ic, line 81:
```suggestion
	// Retrieve minTick and maxTick from precision factor
``` 

nit: this step isn't actually randomizing

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +60,114 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	minTick, maxTick := clkeeper.GetMinAndMaxTicksFromExponentAtPriceOne(clPool.GetPrecisionFactorAtPriceOne())
+
+	// Randomize lowerTick and upperTick to create position
```
czarcas7ic, line 84:
```suggestion
	// Randomize lowerTick and upperTick from max values to create position
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -60,6 +60,114 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	}
+
+	if len(positions) == 0 {
+		return nil, fmt.Errorf("user does not have any position")
+	}
+
+	// pick a random position
+	randPosition := positions[rand.Intn(len(positions))]
+
+	withdrawAmount := randPosition.Liquidity.TruncateInt()
```
czarcas7ic, line 128: This is a good change from last time, but can we randomize the withdraw amount to be less than or equal to the total liquidity of the position?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +178,66 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
 		return hasTwoCoins && hasPoolCreationFee
 	}
 }
+
+// getRandCLPool gets a concnerated liquidity pool with its pool denoms.
```
czarcas7ic, line 182:
```suggestion
// getRandCLPool gets a concentrated liquidity pool with its pool denoms.
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +178,66 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+	// check if the pool exists
+	poolI, err := k.GetPool(ctx, pool_id)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Pool not found for id %d", pool_id)
+	}
+
+	concentratedPool, ok := poolI.(cltypes.ConcentratedPoolExtension)
+	if !ok {
+		return nil, nil, fmt.Errorf("interface conversion failed")
+	}
```
czarcas7ic, line 206: nit: should be able to GetPoolFromId to not have to do the conversion here.

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +178,66 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+
+	poolDenoms := []string{concentratedPool.GetToken0(), concentratedPool.GetToken1()}
+
+	return concentratedPool, poolDenoms, err
+}
+
+// getRandomTickPositions returns random lowerTick and upperTick divisible by tickSpacing value.
+func getRandomTickPositions(minTick, maxTick int64, tickSpacing uint64) (int64, int64, error) {
+	// Calculate the number of values between minTick and maxTick that are divisible by tickSpacing
+	lowerTickDivisible := ((maxTick - minTick) / int64(tickSpacing)) + 1
```
czarcas7ic, line 216: Can you explain this to me a little further, for some reason it just simply isn't clicking. Max tick is 10 and min tick is 5 with a tick spacing of 1. This returns a lowerTickDivisible of 6. Why 6?

czarcas7ic (approved):
Pre-approving, would like the below comments to be addressed. Would also like a second review if possible just to confirm.

If second review is not possible 24h from this comment and all comments are addressed we can merge. Nice work :)

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: I tried this before, but the withdrawAmt gets so large sometimes that it gets out of Int64 bounds. For ex: `256665256162150952146648277` int64() out of bound. But we accept this in sdk.Int 

I found a new way where i just withdraw a random % from 0-1 of total liquidity

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: I tried few stuff, and i think the closest efficient thing i go to was do is; 
```
userCoins := sim.BankKeeper().SpendableCoins(ctx, acc.Address)
checkSubset := poolDenoms.DenomsSubsetOf(userCoins)
if !checkSubset {
 error ("user doesnot contain pool denoms")
}
```

So the issue with this is that; 
1. poolDenoms is a denomString and doesnot contains full coinAmt as cannot retrieve that from CLPool
2. I created my own denom subset function that only checks if the account contains denom, but this requires multiple for loop and i'm afraid that it might slow down the simulator for such a simple task

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: I realized that this is a complex method. I found a simpler way :D 
we just keep generating random number from minTick to maxTick and take the first number divisible

stackman27: hopefully the new version is much clearer

simulation/simtypes/randutil.go
```diff
@@ -74,3 +76,17 @@ func (sim *SimCtx) RandSubsetCoins(coins sdk.Coins) sdk.Coins {
 	r := sim.GetSeededRand("random subset coins")
 	return sdkrand.RandSubsetCoins(r, coins)
 }
+
+// Random float point from 0-1
+func (sim *SimCtx) RandomWeight(maxVal sdk.Dec) (sdk.Dec, error) {
+	rand.Seed(time.Now().UnixNano())
```
p0mvn, line 82: Let's not use time.Now(). We should be able to replay all randomness from a daterministiv seed

simulation/simtypes/randutil.go
```diff
@@ -74,3 +76,17 @@ func (sim *SimCtx) RandSubsetCoins(coins sdk.Coins) sdk.Coins {
 	r := sim.GetSeededRand("random subset coins")
 	return sdkrand.RandSubsetCoins(r, coins)
 }
+
+// Random float point from 0-1
+func (sim *SimCtx) RandomWeight(maxVal sdk.Dec) (sdk.Dec, error) {
```
p0mvn, line 81: Also, we have a `RandomDecAmount` function. Why cannot we use that instead?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +61,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	}
+
+	return &cltypes.MsgCreatePosition{
+		PoolId:        clPool.GetId(),
+		Sender:        sender.Address.String(),
+		LowerTick:     lowerTick,
+		UpperTick:     upperTick,
+		TokenDesired0: tokens[0],
+		TokenDesired1: tokens[1],
+		// TODO: Randomize TokenMinAmount0 and TokenMinAmount1 in next iteration
```
p0mvn, line 99: Is there an issue for this?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +61,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+	if err != nil {
+		return nil, fmt.Errorf("position does not exist")
+	}
+
+	if len(positions) == 0 {
+		return nil, fmt.Errorf("user does not have any position")
+	}
+
+	// pick a random position
+	randPosition := positions[rand.Intn(len(positions))]
```
p0mvn, line 169: How's the seed for these rand operations set?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +185,54 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+	}
+
+	if lowerTick == upperTick {
+		return 0, 0, fmt.Errorf("lower tick and upper tick cannot be the same")
+	}
+
+	return lowerTick, upperTick, nil
+}
+
+//RandomTickDivisibility calculates a random number between minTick - maxTick (inclusive) that is divisible by tickSpacing
```
p0mvn, line 226:
```suggestion
// RandomTickDivisibility calculates a random number between minTick - maxTick (inclusive) that is divisible by tickSpacing
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +185,54 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+func RandomTickDivisibility(minTick int64, maxTick int64, tickSpacing uint64) (int64, error) {
+	if maxTick-minTick < int64(tickSpacing) {
+		return 0, fmt.Errorf("tick spacing is too large")
+	}
+
+	result := rand.Int63n(maxTick-minTick) + minTick
+	for result%int64(tickSpacing) != 0 {
+		result = rand.Int63n(maxTick-minTick) + minTick
+	}
```
p0mvn, line 235: I think it might be likely that we keep not hitting the desired number for some time due to randomly choosing a value for checking whether it it is divisible by `tickSpacing`

Consider the following algorithm that is more deterministic:
- estimate the number of intervals between `minTick` and `maxTick` that are divisible by `tickSpacing`
 * e.g. `minTick` is 100 and `maxTick` is 110, `tickSpacing` is 3. (110 - 100) // 3 = 3 
- randomly choose the number of such interval
 * e.g. 2
- multiply `tickSpacing` by such interval and add to `minTick`. 
 * 2 * 3 +100 = 106
- Iterate backwards to find the desired number
 * e.g takes one iteration to get to 105 that is divisible by 3

stackman27: 🤔  why do we have to do these extra steps; 
```
multiply tickSpacing by such interval and add to minTick.
2 * 3 +100 = 106
Iterate backwards to find the desired number
e.g takes one iteration to get to 105 that is divisible by 3
```

how about something simple like this to assure that the randomValue is always divisible by tickSpacing; 

```
randomVal := random(maxTick - minTick + tickSpacing) + minTick 
divisibleValue = randomVal - randomVal % tickSpacing 
for ex: 
randomValue := random(110 - 100 + 3) + 100 = 12 + 110  = 112
divisibleValue  := randomValue - randomValue % 3 = 12 - 12%3 = 12 - 0 = 12 
```

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: added, i thought this was set at setup but i added just incase

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: added https://github.com/osmosis-labs/osmosis/issues/4230

simulation/simtypes/randutil.go
stackman27: removed to use RandomDecAmount

simulation/simtypes/randutil.go
stackman27: removed to use RandomDecAmount

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: hmmm i took inspiration from your approach and found a similar approach. Please lmk what you think

for example; 
```
 minTick, maxTick, tickSpacing = 100, 110, 3
randomNumber = 103
// find the next multiple of tickSpacing that is >= randomNumber 
nextMultiple  = ((randomNumber + tickSpacing - 1) / tickSpacing) * tickSpacing
nextMultiple = ((103 + 3 - 1) / 3) * 3 = 35 * 3 = 105 
// check if nextMultiple is within [110, 110] 
```

tests/simulator/sim_test.go
```diff
@@ -25,7 +25,6 @@ func BenchmarkFullAppSimulation(b *testing.B) {
 	osmosim.FlagNumBlocksValue = 1000
 	osmosim.FlagBlockSizeValue = 200
 	osmosim.FlagCommitValue = true
-	osmosim.FlagVerboseValue = true
```
p0mvn, line 28: What does this flag do? Why removing it?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -18,6 +18,7 @@ import (
 var PoolCreationFee = sdk.NewInt64Coin("stake", 10_000_000)
 
 func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*clmodeltypes.MsgCreateConcentratedPool, error) {
+	rand := sim.GetSeededRand("select random seed")
```
p0mvn, line 21: The seed should already be configured at the simulator test level. Please use that one.

AFAIU, the rand manager that is seeded can be grabbed from here:
https://github.com/osmosis-labs/osmosis/blob/9d9806b4f36db9516fdb7e7419e4ab74eaf24997/simulation/simtypes/simctx.go#L20

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +62,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+		PoolId:          randPosition.PoolId,
+		Sender:          sender.Address.String(),
+		LowerTick:       randPosition.LowerTick,
+		UpperTick:       randPosition.UpperTick,
+		LiquidityAmount: withdrawAmountInt,
+	}, nil
+}
+
+func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgCollectFees, error) {
+	rand := sim.GetSeededRand("select random seed")
```
p0mvn, line 147: ditto

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +62,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+		TokenDesired0: tokens[0],
+		TokenDesired1: tokens[1],
+		// TODO: Randomize TokenMinAmount0 and TokenMinAmount1 in next iteration
+		TokenMinAmount0: sdk.NewInt(0),
+		TokenMinAmount1: sdk.NewInt(0),
+	}, nil
+}
+
+func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgWithdrawPosition, error) {
+	rand := sim.GetSeededRand("select random seed")
```
p0mvn, line 107: ditto

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +186,68 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	numPools := len(clPools)
+	if numPools == 0 {
+		return nil, nil, fmt.Errorf("no pools created")
+	}
+
+	randConcentratedPool := clPools[rand.Intn(numPools)]
```
p0mvn, line 203: Please use the seeded rand

x/concentrated-liquidity/simulation/sim_msgs.go
p0mvn: This is where the seed value is set: https://github.com/osmosis-labs/osmosis/blob/9d9806b4f36db9516fdb7e7419e4ab74eaf24997/tests/simulator/sim_test.go#L41

This is what should control all randomness in the simulator.

If that value is constant, I should get the same results deterministically. If I change it, only then I get different results.

By initializing new seeds here, we are introducing randomness independent of that global seed value so it becomes non-trivial to control

p0mvn (changes requested):

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: aah yea so it's getting random seed with "FlagSeedValue + string we pass"

stackman27: this is the function that uses FlagSeedValue and adds the string we pass(https://github.com/osmosis-labs/osmosis/blob/main/simulation/simtypes/simctx.go#L156) 

i think its fine to use string, i've changed string name to follow function name too

p0mvn: So what I'm afraid is that we're resetting the global seed here.

When I made a comment originally, I just wanted to make sure that we are using a simulator global-seed that is configured at the top-level of the test that I linked.

I think we can get the `rand` associated with that seed by doing:
```
simRand := sim.GetRand()

randInt := simRand.Intn()
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +62,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+
+	if len(positions) == 0 {
+		return nil, fmt.Errorf("user does not have any position")
+	}
+
+	// pick a random position
+	randPosition := positions[rand.Intn(len(positions))]
+
+	// get percentage amount from 1 to 100 to withdraw liquidity
+	randPerc := sdk.MustNewDecFromStr(fmt.Sprintf("%.2f", sim.RandomDecAmount(sdk.OneDec())))
```
p0mvn, line 133: nit: please double check but I think this should be sufficient for getting the desired value
```suggestion
	randPerc := sim.RandomDecAmount(sdk.OneDec())
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +62,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+		TokenDesired0: tokens[0],
+		TokenDesired1: tokens[1],
+		// TODO: Randomize TokenMinAmount0 and TokenMinAmount1 in next iteration
+		TokenMinAmount0: sdk.NewInt(0),
+		TokenMinAmount1: sdk.NewInt(0),
+	}, nil
+}
+
+func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgWithdrawPosition, error) {
+	rand := sim.GetSeededRand("select random position")
```
p0mvn, line 107: Suggest doing this in all cases because I don't think we should reuse the same domain separation across the messages
```suggestion
	rand := sim.GetRand()
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -61,6 +62,121 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
...
+		PoolId:          randPosition.PoolId,
+		Sender:          sender.Address.String(),
+		LowerTick:       randPosition.LowerTick,
+		UpperTick:       randPosition.UpperTick,
+		LiquidityAmount: withdrawAmountInt,
+	}, nil
+}
+
+func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgCollectFees, error) {
+	rand := sim.GetSeededRand("select random position")
```
p0mvn, line 147:
```suggestion
	rand := sim.GetRand()
```

x/concentrated-liquidity/simulation/sim_msgs.go
p0mvn, line 203: So this seems to be accessed from `math/rand`. Can you please remove that import and ensure that we are accessing all randomness from the simulator-specific rand manager?

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -70,3 +186,68 @@ func createPoolRestriction(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
...
+	if lowerTick == upperTick {
+		return 0, 0, fmt.Errorf("lower tick and upper tick cannot be the same")
+	}
+
+	return lowerTick, upperTick, nil
+}
+
+//RandomTickDivisibility calculates a random number between minTick - maxTick (inclusive) that is divisible by tickSpacing
+func RandomTickDivisibility(sim *osmosimtypes.SimCtx, minTick int64, maxTick int64, tickSpacing uint64) (int64, error) {
+	rand := sim.GetSeededRand("select random seed")
```
p0mvn, line 238:
```suggestion
	rand := sim.GetRand()
```

x/valset-pref/simulation/sim_msgs.go
```diff
@@ -172,10 +172,7 @@ func GetRandomValAndWeights(ctx sdk.Context, k valsetkeeper.Keeper, sim *osmosim
 			return nil, fmt.Errorf("No validator")
 		}
 
-		randValue, err := RandomWeight(remainingWeight)
-		if err != nil {
-			return nil, fmt.Errorf("Error with random weights")
-		}
+		randValue := sdk.MustNewDecFromStr(fmt.Sprintf("%.2f", sim.RandomDecAmount(remainingWeight)))
```
p0mvn, line 175: Similat to another comment, I think this should be fine but please confirm before committing
```suggestion
		randValue := sim.RandomDecAmount(remainingWeight)
```

p0mvn (approved):
Nice work. Only minor comments

x/concentrated-liquidity/simulation/sim_msgs.go
p0mvn: Confirmed this is working

x/concentrated-liquidity/simulation/sim_msgs.go
p0mvn: Discussed this on Slack, it looks like we shouldn't share seeds between messages to avoid an error in one function drastically changing randomness in another. So I think `GetRand` is preferable here because it doesn't use shared domain

x/valset-pref/simulation/sim_msgs.go
p0mvn: Confirmed this is working by changing the other suggestion and rerunning the simulator

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: makes sense! updated

p0mvn (approved):
Great work!

PR #4198, opened 2023-02-01
test(CL Fees): updateFeeGrowthGlobal

p0mvn:
Closes: #4087 and #4086

Implements tests for the new fee-related helper methods.


x/concentrated-liquidity/fees_test.go
```diff
@@ -951,3 +952,127 @@ func (s *KeeperTestSuite) TestUpdateFeeAccumulatorPosition() {
 	}
 
 }
+
+func (suite *KeeperTestSuite) TestComputeFeeChargePerSwapStep() {
+	var (
+		five              = sdk.NewDec(5)
+		belowDefaultPrice = DefaultCurrPrice.Sub(five)
+		aboveDefaultPrice = DefaultCurrPrice.Add(five)
```
czarcas7ic, line 960: Why don't we use `DefaultLowerPrice` and `DefaultUpperPrice` here?

x/concentrated-liquidity/fees_test.go
```diff
@@ -951,3 +952,127 @@ func (s *KeeperTestSuite) TestUpdateFeeAccumulatorPosition() {
...
+		"current sqrt price equals neither and currentSqrtPrice < nextTickSqrtPrice -> fee is the diff between amount specified remaining and amount in": {
+			currentSqrtPrice:         DefaultCurrSqrtPrice,
+			nextTickSqrtPrice:        sqrtAboveDefaultPrice,
+			sqrtPriceLimit:           sqrtAboveDefaultPrice,
+			amountIn:                 sdk.NewDec(101),
+			amountSpecifiedRemaining: sdk.NewDec(102),
+			swapFee:                  onePercentFee,
+
+			// 102 - 101 = 1
+			expectedFeeCharge: sdk.OneDec(),
```
czarcas7ic, line 1031: I don't know how I feel about having all test cases expect a zero or one fee. I would feel better having at least one example where the expected fee charge is larger.

x/concentrated-liquidity/fees_test.go
```diff
@@ -951,3 +952,127 @@ func (s *KeeperTestSuite) TestUpdateFeeAccumulatorPosition() {
...
+		"current sqrt price == sqrt price limit -> charge fee on amount in": {
+			currentSqrtPrice:         DefaultCurrSqrtPrice,
+			nextTickSqrtPrice:        sqrtAboveDefaultPrice,
+			sqrtPriceLimit:           DefaultCurrSqrtPrice,
+			amountIn:                 sdk.NewDec(100),
+			amountSpecifiedRemaining: five,
+			swapFee:                  onePercentFee,
+
+			expectedFeeCharge: sdk.OneDec(),
+		},
```
czarcas7ic, line 1000: Im a little confused here, if the currentPrice is equal to the limit and the next tick is upwards, shouldn't the swap never go through?

czarcas7ic (commented):
Looking to do one more pass once the below comments are addressed. Mainly just confused about the `sqrtPriceLimit` and how that comes into play with calculating fee charge per swap step.

x/concentrated-liquidity/fees_test.go
p0mvn: If we hit the price limit, we are still allowed to consume some `amountIn` during the swap. It would have been a larger amount if not the price limit. However, we still need to charge the fee on some `amountIn` consumed up until we hit the price limit

x/concentrated-liquidity/fees_test.go
p0mvn: This isn't a problem..

It's zero when the swap fee is zero. It's one due to how the input parameters are set up. The fee is still calculated to be one given the inputs in each test case.

For each branch in the function, there is a test covering it. I could have chosen a larger value, for example, `103` here to get the result of 2. The test coverage and test completeness would not change from the choice of the value.

Issue #4202, opened 2023-02-02
[accum][CL Incentives]: Track total shares for accumulators


AlpinYukseloglu:
## Background

Being able to retrieve the total number of shares for an accumulator is an important and necessary primitive for our use of the accum package for CL incentives. It is possible to implement this at a higher level of abstraction, but the most appropriate place to have it would be in the accumulator package.

## Suggested Design

- Add a `TotalShares` field to `AccumulatorContent` and update it under `initOrUpdatePosition`

## Acceptance Criteria

- All new and existing accum tests pass

PR #4204, opened 2023-02-02
[accum][CL Incentives]: Track total shares for accumulators

AlpinYukseloglu:
Closes: #4202

As described in the issue, tracking total shares for accumulators in our `accum` package is important functionality for us to have, especially in scenarios where we are juggling multiple/varying numbers of accumulators as with CL incentives.

## Brief Changelog

- Added `TotalShares` field to `AccumulatorContent` (stored in state)
- Implemented necessary update logic on all core position operations (new/add/remove)
- Implemented partially randomized testing to ensure coverage

## Testing and Verifying

- All new functionality is tested relatively thoroughly under `TestGetTotalShares` in `accum_test.go`
- To avoid having to refactor all accum test vectors, `TotalShares` is not tested/checked in past tests (which deal with raw accum objects not in store). The partial fuzz testing is intended to compensate for this by checking a broad range of operations.


osmoutils/accum/accum_test.go
```diff
@@ -1461,3 +1462,95 @@ func (suite *AccumTestSuite) TestSetPositionCustomAcc() {
 		})
 	}
 }
+
+// We run a series of partially random operations on two accumulators to ensure that total shares are properly tracked in state
+func (suite *AccumTestSuite) TestGetTotalShares() {
```
AlpinYukseloglu, line 1467: Note to reviewer: core changes are tested here

osmoutils/accum/accum_test.go
```diff
@@ -1461,3 +1462,95 @@ func (suite *AccumTestSuite) TestSetPositionCustomAcc() {
...
+		suite.Require().NoError(err)
+		if !positionExists {
+			err = curAccum.NewPosition(curAddr, baseAmt, nil)
+			suite.Require().NoError(err)
+		}
+
+		// We generate a random binary value (0 or 1) to determine
+		// whether we will add and/or remove liquidity this loop
+		addShares := sdk.NewDec(int64(rand.Int()) % 2)
+		removeShares := sdk.NewDec(int64(rand.Int()) % 2)
```
p0mvn, line 1522: We should be setting a deterministic seed for the randomness. Otherwise, it would be difficult to debug non-determinis if any ever occurs

p0mvn (approved):
LGTM. Only request is to use a seed in the test, please

osmoutils/accum/accum_test.go
AlpinYukseloglu: Good call!

mattverse (commented):
Minor comments, otherwise lgtm

osmoutils/accum/export_test.go
```diff
@@ -31,11 +31,15 @@ func (accum AccumulatorObject) GetAllPositions() ([]Record, error) {
 }
 
 // Creates an accumulator object for testing purposes
-func CreateRawAccumObject(store store.KVStore, name string, value sdk.DecCoins) AccumulatorObject {
+func CreateRawAccumObject(store store.KVStore, name string, value sdk.DecCoins, totalShares sdk.Dec) AccumulatorObject {
+	// We store an accumulator object in state even if unused in tests
+	// because position operations still require GetAccumulator to work
+	_ = MakeAccumulator(store, name)
```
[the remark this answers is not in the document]
AlpinYukseloglu: Since this PR makes position updates mutate accumulator state, it turns out that every call to this helper requires setting an accum in state.

I agree that changing the name would be better but every call to this function would require doing `MakeAccumulator` first, so having that in the helper avoids all that duplicate code

osmoutils/accum/accum.go
```diff
@@ -207,6 +226,13 @@ func (accum AccumulatorObject) RemoveFromPositionCustomAcc(name string, numShare
 	// Update user's position with new number of shares
 	initOrUpdatePosition(accum, customAccumulatorValue, name, oldNumShares.Sub(numSharesToRemove), unclaimedRewards, position.Options)
 
+	// Update total shares in accum (re-fetch accum from state to ensure it's up to date)
+	accum, err = GetAccumulator(accum.store, accum.name)
+	if err != nil {
+		return err
+	}
+	setAccumulator(accum, accum.value, accum.totalShares.Sub(numSharesToRemove))
+
```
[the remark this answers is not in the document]
AlpinYukseloglu: I believe we do since the caller can accidentally keep calling the method on the same `AccumulatorObject` without pulling the new one with updated `TotalShares` from state (and imo we shouldn't be relying on that happening if we can prevent the bug in the function)

mattverse (approved):
LGTM with the function name changes!

Issue #4147, opened 2023-01-30
Support Async ICQ (Interchain Queries)


daniel-farina:
## Background
 Interchain Queries (ICQ) as a new feature of the Inter-Blockchain Communication protocol (IBC), an application-layer protocol that enables general and standardized communication across blockchains. [Read more.](https://medium.com/@quasar.fi/introducing-interchain-queries-d6243e7b45cf)

Creating this issue to track the progress of supporting this in Osmosis. Quasar will launch on March 23rd so ideally this is supported on V15.

PR #4207, opened 2023-02-02
Integrate Async ICQ

nicolaslara:
Closes: #4147 

Integrates Async ICQ into osmosis (https://github.com/strangelove-ventures/async-icq)

## Brief Changelog

- Integrates async ICQ into osmosis

## Testing and Verifying

All tests pass


app/keepers/keepers.go
```diff
@@ -257,6 +262,26 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
 		AddRoute(ibctransfertypes.ModuleName, appKeepers.TransferStack)
 	// Note: the sealing is done after creating wasmd and wiring that up
 
+	// ICQ Keeper
+	icqKeeper := icqkeeper.NewKeeper(
+		appCodec,
+		appKeepers.keys[icqtypes.StoreKey],
+		appKeepers.GetSubspace(icqtypes.ModuleName),
+		appKeepers.IBCKeeper.ChannelKeeper, // may be replaced with middleware
+		appKeepers.IBCKeeper.ChannelKeeper,
```
ValarDragon, line 271: Can we add a comment for what these args are supposed to be? (Confused why theres two)

app/keepers/keepers.go
```diff
@@ -257,6 +262,26 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
...
+		&appKeepers.IBCKeeper.PortKeeper,
+		appKeepers.ScopedICQKeeper,
+		bApp,
```
ValarDragon, line 274: Uhhhh, why does this take in baseApp -- we may have to audit this and all dep updates carefully then. Its not safe to just give base app access to modules.

nicolaslara: it needs a Querier. I was wondering if there was a something more restrictive that we could pass there.

app/keepers/keepers.go
nicolaslara: Passing it twice is common on IBC modules where you want one with the default implementation and one that is being used to actually route packages through (if there's a middleware, you want to route packets through the middleware but may still need access to the default IBC implementation).

In this case, I don't think the second one is being used. So not sure why it's needed. And the first one is only being used to return the correct app version when implementing `GetAppVersion` (which is required by the interface)

app/keepers/keepers.go
ValarDragon: Gotcha, I think its worth making a wrapper struct on our side for now then

app/keepers/keepers.go
ValarDragon: Oh I see, I imagine this API / standard could be improved, but for another day. SGTM!

go.mod
```diff
@@ -26,6 +26,8 @@ require (
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.15.0
+	// Async ICQ branch: ibc-v4
+	github.com/strangelove-ventures/async-icq v0.0.0-20230116084035-5609e84dd443
```
ValarDragon, line 30: LGTM, can we add a release blocking TODO for this to be replaced with a tag?

ValarDragon (approved):
LGTM! Just needs changelog

go.mod
nicolaslara: Added https://github.com/osmosis-labs/osmosis/issues/4218. I'll ping them to add a tag

nicolaslara: also, how do I make a ticket release blocking?

ValarDragon: We can add it to the tracker issue for now

PR #4213, opened 2023-02-02
feat/fix(fees): fix fee claiming by updating position accumulator value to fee growth outside

p0mvn:
Closes: #XXX

Discovered by @czarcas7ic 

Previously, when we claimed the fees, we used the default claim logic that does not account for custom range positions.

With concentrated liquidity, accumulator positions track fee growth outside. As a result, a new method `ClaimRewardsCustomAcc` is introduced in the `accum` package in this PR.

Additionally, tests added for `CollectFees` to ensure that the position fee accumulator is not updated to the global fee growth value. 

To implement this test, a new method `GetPositionValue` was introduced in the `accum` package.

## Brief Changelog

- implement and test `ClaimRewardsCustomAcc` 
- fix the bug in `CollectFees` by calling the new claim function
- implement and test `GetPositionValue`
- use `GetPositionValue` to test that the position accumulator in `CollectFees` is not equal to the global fee growth accumulator.

## Testing and Verifying

This change added tests


x/concentrated-liquidity/fees.go
```diff
@@ -214,7 +214,7 @@ func (k Keeper) collectFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress
 	}
```
czarcas7ic, line 214: I believe the lines above this (210-213) need to be removed. We are properly accounting for the current fee growth outside when we use the ClaimRewardsCustomAcc func, so updating the accumulator prior that causes issues.

czarcas7ic:
Will describe in today's call why this needs further changes to completely work. Deleted my previous in line comments since the problem was involving more than what I just pointed out.

czarcas7ic:
Here is the commit I made on my branch with all the fixes https://github.com/osmosis-labs/osmosis/commit/b45eb46a5477dc55136868a12e4a1de404dd6748

p0mvn:
synched offline - pursuing another direction for fixes

Issue #1628, opened 2022-05-31
[tests/e2e]: simplify the logic for adding accounts to genesis sdk side and integrate into e2e


p0mvn:
**Background**

We should simplify the process of adding accounts to genesis on the SDK side to improve readability in e2e set up. This issue includes making additions to our SDK fork, updating the dependency in osmosis, and, eventually, adding it to our e2e set up.

Please follow this comment for more information: https://github.com/osmosis-labs/osmosis/pull/1625#discussion_r885131979_

**Acceptance Criteria**
- investigate an easier way to add accounts to genesis
- implement the improvement in Osmosis's SDK fork and unit test
- update the sdk dependency in Osmosis's `go.mod`
- refactor e2e set up to utilize the simpler method

p0mvn:
What needs to be done is the following:

- backport https://github.com/cosmos/cosmos-sdk/issues/12091 to our sdk fork
- update SDK dependency
- refactor e2e genesis logic to use the simplified helper

p0mvn:
Reopining for the fix to follow-up on the following CI failure: https://github.com/osmosis-labs/osmosis/actions/runs/4338982856/jobs/7576658435

PR #4217, opened 2023-02-03
[tests/e2e]: simplify the logic for adding accounts to genesis sdk side and integrate into e2e#1628

ThanhNhann:
Closes: #1628 

Update logic add account to genesis in e2e testing

## Testing and Verifying


tests/e2e/initialization/helper.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("Account %s already exists\nUse `append` flag to append account at existing address", accAddr)
+		}
+
+		genesisB := banktypes.GetGenesisStateFromAppState(cdc, appState)
+		for idx, acc := range genesisB.Balances {
+			if acc.Address != accAddr.String() {
+				continue
+			}
+
+			updatedCoins := acc.Coins.Add(coins...)
```
RusAkh, line 96: Wdyt: doesn't it make more sense to use `amountStr` as override parameter here and not to add it to existing coins? I think it makes more sense because when adding coins to existing account, you won't need to think about current balance and just specify the desired in `amountStr` parameter. This is just an idea, we will see if others like it

RusAkh (commented):
utACK, thanks!

tests/e2e/initialization/helper.go
ThanhNhann: Thank you for your suggestion, I think we should still keep this with precise logic, btw this is the draft version of this pr, in the future I will call this function from osmosis/cosmos-sdk in this link https://github.com/osmosis-labs/cosmos-sdk/blob/osmosis-main/x/auth/helpers/genaccounts.go

ThanhNhann: And can you explain more about your suggestion? I don't really understand all of it

RusAkh: I do not know if that is possible what I am suggesting, but what I mean is: do not add `coins` to account balance if that account exists, but override the current balance of this account with `coins` provided

lets say we have an account with `100uosmo`. If we are adding this account `AddGenesisAccount`, and provide `amountStr` as `120uosmo`, we end up with `220uosmo` because they are just added. My suggestion is this: if you provide `120uosmo` as `amountStr`, it essentially overrides the balance, so after this function, account that initially had `100uosmo` will have `120uosmo`. I think this way if a user wants a specific account to have some specific amount, they do not need to think about the current balance. Lmk if that makes sense

ThanhNhann: I think it does not override the balances instead add-in

RusAkh: yeah that is true I am just suggesting the override approach

p0mvn (approved):
Nice work, thank you!

PR #4219, opened 2023-02-03
XCS: management messages

nicolaslara:
Closes: #4156 

Adds management messages to XCS so the contract owner can modify the config. 

## Brief Changelog

 - Added management messages

## Testing and Verifying

Added tests for the new management messages. All old tests pass.


cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -1,7 +1,19 @@
...
 
-use crate::{state::CHANNEL_MAP, ContractError};
+use crate::{
+    state::{CHANNEL_MAP, CONFIG},
+    ContractError,
+};
+
+pub fn check_is_contract_owner(deps: Deps, sender: Addr) -> Result<(), ContractError> {
+    let config = CONFIG.load(deps.storage).unwrap();
+    if config.owner != sender {
```
ValarDragon, line 11: interesting, for some reason I thought owner was a native concept in cosmwasm, rather than just a contract variable

ValarDragon:
Can we update some docs for this?

TBH a little bit confused why its the XCS contract that maintains the channel registry

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: There's a native contract admin, that is allowed to migrate the contract, but if you want to use it for internal stuff you need to track it (you may want the internal owner and the admin to be different!)

ValarDragon: Ah I see, perhaps lets call this "governor", to help distinguish?

nicolaslara:
> TBH a little bit confused why its the XCS contract that maintains the channel registry

This was added for a few reasons:
 * To be able to send to a receiver without having to specify the channel
 * To be able to limit which channels were used. Sending to an address with a bad bech32 prefix would fail transactions at the IBC level and if that error happened, the tokens would be stuck on osmosis. This doesn't matter anymore, as we have IBC lifecycle callbacks, but we can still fail early instead of sending the tokens and having to deal with the callback
 * This should actually belong in the registry/router contract, but since the router is going to have to deal with unwinding and is a bit more complex, having it here for now allows us to launch XCS without having to wait for the registry/router.

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: I have a branch changing this to "governor" but the swaprouter contract (that's already on-chain)  uses "owner".

The cosmwasm internal one is called "admin"

ValarDragon: Are you planning to do the governor rename as a second PR?

nicolaslara: I was thinking of pushing it here, just thought I'd double check if we wanted this considering that swaprouter uses owner. I feel like they should use the same name. 

I can update it on swaprouter too, but then it will be different than the one on chain 😅

ValarDragon: I'm a fan of governor personally. the inconsistency between the two doesn't really bother me FWIW

cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -1,7 +1,19 @@
 use cosmwasm_std::{Addr, Deps};
 use regex::Regex;
 
-use crate::{state::CHANNEL_MAP, ContractError};
+use crate::{
+    state::{CHANNEL_MAP, CONFIG},
+    ContractError,
+};
+
+pub fn check_is_contract_owner(deps: Deps, sender: Addr) -> Result<(), ContractError> {
```
ValarDragon, line 9: Should we add a unit test for this function?

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -253,3 +253,236 @@ pub fn recover(deps: DepsMut, sender: Addr) -> Result<Response, ContractError> {
...
+}
+
+/// Remove a prefix from the registry
+pub fn remove_channel(
+    deps: DepsMut,
+    sender: Addr,
+    prefix: String,
+) -> Result<Response, ContractError> {
+    check_is_contract_owner(deps.as_ref(), sender)?;
+    CHANNEL_MAP.remove(deps.storage, &prefix);
```
ValarDragon, line 276: Hrmm, I feel a bit concerned about letting prefix be rebound. I feel like if its bound once, it can be disabled, but shouldn't be editable.

nicolaslara: I'm thinking of a case like luna, where a new channel is created for the same prefix

nicolaslara: btw, this function is deleting the prefix=>channel mapping. 

The set_channel msg does reset it: if it exists, it overrides the existing one. We could change that so that we can only set non-existing prefixes. 

I'm happy with that solution, even if I don't see a problem with having it be more flexible

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: ok! I'll push the governor change then

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: added

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: pushed

cosmwasm/contracts/crosschain-swaps/src/execute.rs
ValarDragon: I think rebinding should need more governance authority (since you can get rug assumptions or safety pulled out from under you)

Can we do Set New + Disable/Re-enable existing? (We can model Disabled as a field in the value type in the same map)

nicolaslara: Initially only governance can run these, but another option would be to have two governors so even if we transferred base ownershio to, say, a dao, only gov could rebind. 

I'll change it to set_new+disable/re-enable for now and we can discuss rebinding later.

ValarDragon: That makes sense!

ValarDragon (approved):

PR #4222, opened 2023-02-04
bug: multi swap-step and fee accum

czarcas7ic:

There were two bugs that were discovered when implementing the functional tests for fees:

1. We do not properly update the global fee accumulator when a swap has to go through more than one iteration
2. The fee accumulator itself needs a custom claim function in order to take into account the feeGrowthOutside

This PR fixes both of the above issues. I have left out the functional tests that helped discover this bug only to reduce the scope and will be introducing it in a subsequent PR.

## Brief Changelog

*(for example:)*
 

osmoutils/accum/accum_test.go
```diff
@@ -1418,18 +1472,6 @@ func (suite *AccumTestSuite) TestSetPositionCustomAcc() {
 			positionName:           validPositionName,
 			customAccumulatorValue: initialCoinsDenomOne,
 		},
-		"invalid update smaller than the initial value": {
```
czarcas7ic, line 1421: Note to reviewer: We no longer have this validation which is why these expectedErrors never happen

osmoutils/accum/accum_test.go
```diff
@@ -1418,18 +1472,6 @@ func (suite *AccumTestSuite) TestSetPositionCustomAcc() {
...
-			customAccumulatorValue: emptyCoins,
-
-			expectedError: accumPackage.NegativeAccDifferenceError{AccumulatorDifference: initialCoinsDenomOne},
-		},
-		"invalid update smaller than the initial value (non-empty custom value)": {
-			positionName:           validPositionName,
-			customAccumulatorValue: initialCoinsDenomOne.QuoDec(sdk.NewDec(2)),
-
-			expectedError: accumPackage.NegativeAccDifferenceError{AccumulatorDifference: initialCoinsDenomOne.QuoDec(sdk.NewDec(2))},
-		},
```
czarcas7ic, line 1432: Note to reviewer: We no longer have this validation which is why these expectedErrors never happen

PR #4223, opened 2023-02-04
Apply the Tendermint security patch

ValarDragon:

Tendermint applied a security patch as a patch release. The problem could notably increase a peers CPU usage.


mattverse (approved):
ACK

mattverse:
Solving merge conflicts seems unnecessary VS opening a new PR! 
replacing this PR with https://github.com/osmosis-labs/osmosis/pull/4383

Issue #3992, opened 2023-01-12
(CL test): Add test for ensuring `writeCtx` from calc methods correctly overwrites Context


mattverse:
## Background
In https://github.com/osmosis-labs/osmosis/pull/3910, we made the calc methods non-mutative, and instead used the methodology where we return the `writeCtx` function from the `ctx.CacheContext` method. 

Swap methods call this method to write the state in the context for ticks during swaps. 

This test was impossible to be added as a part of #3910 as it was blocked by #3890 at the point of this issue being written.

## Suggested Design

In`TestCalcAndSwapOutAmtGivenIn` and `TestCalcAndSwapInAmtGivenOut`, we're currently calling CalcInAmtGivenOutInternal, which returns writeCtx function. Currently the returned writeCtx function remains un-used, but the suggested design is using that function(calling that specific function) and then comparing state before calling that function and after calling that function. 

Any other methodologies for testing this is fine as well.

## Acceptance Criteria

- Tests for `writeCtx` functiton has been added
   - This should specifically test if calling the `writeCtx` function returned by the calc methods changes the ctx

PR #4236, opened 2023-02-06
(CL test): `writeCtx` tests for calc methods

mattverse:
Closes: #3992

This PR adds tests for the `writeCtx` being returned in the calc methods.
The `writeCtx` does not alter any pool state as we handle pool state changes in the `applySwap` method. Instead, it changes the fee accumulator state upon being called. The test case being added tests if calling `writeCtx` succesfully alters fee accum state whilst leaving the pool state as is. 

## Brief Changelog
Add test case for `writeCtx` being returned via calc methods.

## Testing and Verifying
This PR adds tests

### DIFF
--- a/x/concentrated-liquidity/swaps_test.go
+++ b/x/concentrated-liquidity/swaps_test.go
@@ -1692,6 +1692,175 @@ func (s *KeeperTestSuite) TestSwapExactAmountOut() {
 	}
 }
 
+// TestCalcOutAmtGivenInWriteCtx tests that writeCtx succesfully perfroms state changes as expected.
+// We expect writeCtx to only change fee accum state, since pool state change is not handled via writeCtx function.
+func (s *KeeperTestSuite) TestCalcOutAmtGivenInWriteCtx() {
+	// we only use fee cases here since write Ctx only takes effect in the fee accumulator
+	tests := make(map[string]SwapTest, len(swapOutGivenInFeeCases))
+
+	for name, test := range swapOutGivenInFeeCases {
+		tests[name] = test
+	}
+
+	for name, test := range tests {
+		test := test
+		s.Run(name, func() {
+			s.Setup()
+			s.FundAcc(s.TestAccs[0], sdk.NewCoins(sdk.NewCoin("eth", sdk.NewInt(10000000000000)), sdk.NewCoin("usdc", sdk.NewInt(1000000000000))))
+			s.FundAcc(s.TestAccs[1], sdk.NewCoins(sdk.NewCoin("eth", sdk.NewInt(10000000000000)), sdk.NewCoin("usdc", sdk.NewInt(1000000000000))))
+
+			// Create default CL pool
+			pool := s.PrepareConcentratedPool()
+
+			// add default position
+			s.SetupDefaultPosition(pool.GetId())
+
+			// add second position depending on the test
+			if !test.secondPositionLowerPrice.IsNil() {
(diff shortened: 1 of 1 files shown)


x/concentrated-liquidity/swaps_test.go
```diff
@@ -1692,6 +1692,175 @@ func (s *KeeperTestSuite) TestSwapExactAmountOut() {
...
+			s.Require().Equal(poolBeforeCalc.GetCurrentSqrtPrice(), poolAfterCalc.GetCurrentSqrtPrice())
+			s.Require().Equal(poolBeforeCalc.GetCurrentTick(), poolAfterCalc.GetCurrentTick())
+			s.Require().Equal(poolBeforeCalc.GetTotalShares(), poolAfterCalc.GetTotalShares())
+			s.Require().Equal(poolBeforeCalc.GetLiquidity(), poolAfterCalc.GetLiquidity())
+			s.Require().Equal(poolBeforeCalc.GetTickSpacing(), poolAfterCalc.GetTickSpacing())
+
+			feeAccum, err := s.App.ConcentratedLiquidityKeeper.GetFeeAccumulator(s.Ctx, 1)
+			s.Require().NoError(err)
+
+			feeAccumValue := feeAccum.GetValue()
```
p0mvn, line 1837: How feasible do you think it is to also check that the tick fee accumulator values are not updated?

p0mvn (approved):
Nice work. Super clean. One non-blocking question

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1692,6 +1692,175 @@ func (s *KeeperTestSuite) TestSwapExactAmountOut() {
 	}
 }
 
+// TestCalcOutAmtGivenInWriteCtx tests that writeCtx succesfully perfroms state changes as expected.
```
czarcas7ic, line 1695:
```suggestion
// TestCalcOutAmtGivenInWriteCtx tests that writeCtx successfully performs state changes as expected.
```

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1692,6 +1692,175 @@ func (s *KeeperTestSuite) TestSwapExactAmountOut() {
...
+					osmomath.BigDecFromSDKDec(feeAccum.GetValue().AmountOf(test.tokenIn.Denom)),
+				),
+			)
+		})
+	}
+}
+
+// TestCalcInAmtGivenOutWriteCtx tests that writeCtx succesfully perfroms state changes as expected.
+// We expect writeCtx to only change fee accum state, since pool state change is not handled via writeCtx function.
+func (s *KeeperTestSuite) TestCalcInAmtGivenOutWriteCtx() {
```
czarcas7ic, line 1782: Others might disagree, but it feels to me like this specific writeCtx test case could be merged between calc in and out since we are using the same boilerplate. Not a strong opinion though

czarcas7ic (approved):
Well done 👍

x/concentrated-liquidity/swaps_test.go
mattverse: I was also thinking about the same option in the beginning, but this is hard to achieve since what swap method does and write ctx does can not co-exist 😢

x/concentrated-liquidity/swaps_test.go
mattverse: To do this we first need to specify in each test case which specific tick index it crosses and use that to compare the before / after. 
I personally do not think it is necessary since we already check that fee accum did not change and it is impossible for tick's fee accum to change with out the fee accum changing. Lmk your thoughts

Issue #3320, opened 2022-11-10
Implement Minimum Gas Price (Proposal 354)


JohnnyWyles:
## Background

[Proposal 354](https://www.mintscan.io/osmosis/proposals/354) recently passed, adding this as a tracking issue

> Currently, Osmosis has no protocol-level minimum gas price. The reasoning for this is that the vast majority of OSMO liquidity is on Osmosis, making it difficult for new users to acquire elsewhere to carry out their first transaction. Allowing Osmosis users to submit transactions with a transaction fee of 0 has made it extremely easy for new users to onboard into Osmosis by enabling them to transact on Osmosis for the first time without having the native gas token.
> 
> At the same time, however, setting a minimum gas price of 0 leaves Osmosis vulnerable to denial of service attacks executed via the spamming of free transactions on the network. While thankfully Osmosis has not experienced a major DoS or DDoS event to date, incidences of recurring transaction spam have been increasing.
> 
> With a recent upgrade, Keplr now offers Osmosis users the ability to pay transaction fees on Osmosis in non-OSMO assets (for example, gas can now be paid in ATOM). While the ability to do this has existed on the Osmosis chain for quite some time, this wallet upgrade has opened the door to a large number of onboarding routes into Osmosis for all users on the main front-end for the very first time. Similarly, Osmosis’s Binance listing has made it much easier for new users to purchase OSMO off-chain to use as a transaction fee if necessary.
> 
> As such, the time is ripe to implement a minimum transaction fee on Osmosis. The purpose of this proposal is to signal to Osmosis Labs and other key Osmosis development contributors that governance mandates a minimum gas price of 0.0025 uOSMO be implemented at the protocol level. There is currently no governance adjustable parameter to implement this fee, so this proposal will be implemented in a subsequent software upgrade.
> 
> Further, to the extent that fees need to be priced via the Transaction Fee Module in another asset, the transaction fee module should be amended to utilize Osmosis Time Weighted Average Pricing over at least a 24 hour period instead of the currently utilized spot pricing to ensure that validators are fairly compensated no matter which asset is used for fees

## Suggested Design

* Implement a minimum gas price of 0.0025 uOSMO at the protocol level to be modifiable by governance.
* Replace spot pricing of alternative gas assets with a TWAP over at least 24 hours

ValarDragon:
ah rip, didn't realize it was approved to be at a 24 hour TWAP, I feel like thats not great. 

[...]

PR #4244, opened 2023-02-07
Consensus min gas fee of .0025 uosmo

ValarDragon:
Cref #3320 , does the raw osmo amount component, not the TWAP.

Implements a consensus min gas fee of .0025 uosmo / gas. The mempool policy is set to max(localMempoolPolicy, Consensus minimum)

This consensus minimum is not being exposed as a governance parameter right now, as it seems far more effective to get to adaptive, demand based fee markets (ala EIP-1559).

## Brief Changelog

- Implement a consensus min fee of .0025uosmo/gas as per Osmosis proposal 354 https://www.mintscan.io/osmosis/proposals/354 .

## Testing and Verifying

I've updated the existing test vectors to account for the consensus minimum, which includes adapting no fee and sub-consensus minimum test vectors to fail. Furthermore this PR refactors the test vector setup to be more succinct. (Removing extraneous fields, and using a for loop to combine Check and Deliver)

  - How is the feature or change documented? Yes


x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,92 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-		},
-		{
-			name:         "multiple fee coins - delivertx",
-			txFee:        sdk.NewCoins(sdk.NewInt64Coin(baseDenom, 1), sdk.NewInt64Coin(uion, 1)),
-			minGasPrices: sdk.NewDecCoins(),
-			gasRequested: 10000,
-			isCheckTx:    false,
-			expectPass:   false,
-			baseDenomGas: false,
-		},
```
ValarDragon, line 135: Moved to be deduplicated across check/deliver

x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,92 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-			txFee: sdk.NewCoins(sdk.NewInt64Coin(baseDenom, 1)),
-			minGasPrices: sdk.NewDecCoins(sdk.NewDecCoinFromDec(baseDenom,
-				sdk.MustNewDecFromStr("0.1"))),
-			gasRequested: 10000,
-			isCheckTx:    false,
-			expectPass:   true,
-			baseDenomGas: true,
+			isCheckTx:  true,
+			expectPass: true,
 		},
```
ValarDragon, line 89: Deduplicated above

x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,92 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-		},
-		{
-			name:         "no min gas price - delivertx",
-			txFee:        sdk.NewCoins(),
-			minGasPrices: sdk.NewDecCoins(),
-			gasRequested: 10000,
-			isCheckTx:    false,
-			expectPass:   true,
-			baseDenomGas: true,
-		},
```
ValarDragon, line 57: Changed to be consensus min gas failure tests.

x/txfees/keeper/feedecorator_test.go
```diff
@@ -226,11 +189,13 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-					moduleAddr := suite.App.AccountKeeper.GetModuleAddress(types.FeeCollectorName)
-					suite.Require().Equal(tc.txFee[0], suite.App.BankKeeper.GetBalance(suite.Ctx, moduleAddr, baseDenom), tc.name)
-				} else if !tc.txFee.IsZero() {
-					moduleAddr := suite.App.AccountKeeper.GetModuleAddress(types.NonNativeFeeCollectorName)
+				// ensure fee was collected
+				if !tc.txFee.IsZero() {
+					moduleName := types.FeeCollectorName
+					if tc.txFee[0].Denom != baseDenom {
+						moduleName = types.NonNativeFeeCollectorName
+					}
```
ValarDragon, line 203: Base denom gas was an unnecessary field, its role is automated with this if statement.

ValarDragon:
The failure in go tests is that our ibc-hooks tests use ibctesting. Ibctesting uses 0 fees in its tests, right here: https://github.com/cosmos/ibc-go/tree/v4.3.0

x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,98 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-			isCheckTx:    false,
-			expectPass:   true,
-			baseDenomGas: true,
-		},
+	}
+
+	tests := []testcase{}
+	txType := []string{"delivertx", "checktx"}
+	succesType := []string{"does", "doesn't"}
+	for isCheckTx := 0; isCheckTx <= 1; isCheckTx++ {
```
mattverse, line 42: nice way of deduplication, love it

x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,98 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-			minGasPrices: sdk.NewDecCoins(),
-			gasRequested: 10000,
-			isCheckTx:    false,
-			expectPass:   false,
-			baseDenomGas: false,
+			isCheckTx:  true,
+			expectPass: true,
 		},
 		{
 			name:         "mingasprice not containing basedenom gets treated as min gas price 0",
```
mattverse, line 111:
```suggestion
			name:         "min gas price not containing basedenom gets treated as min gas price 0",
```

x/txfees/keeper/feedecorator_test.go
```diff
@@ -19,137 +21,98 @@ func (suite *KeeperTestSuite) TestFeeDecorator() {
...
-			name:  "works with not enough fee in delivertx",
-			txFee: sdk.NewCoins(sdk.NewInt64Coin(baseDenom, 1)),
-			minGasPrices: sdk.NewDecCoins(sdk.NewDecCoinFromDec(baseDenom,
-				sdk.MustNewDecFromStr("0.1"))),
-			gasRequested: 10000,
-			isCheckTx:    false,
-			expectPass:   true,
-			baseDenomGas: true,
+			isCheckTx:  true,
+			expectPass: true,
```
mattverse, line 100: nit: I feel like the top two tests should be a part of array above, not cusTests

x/txfees/types/constants.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// ConsensusMinFee is a governance set parameter from prop 354 (https://www.mintscan.io/osmosis/proposals/354)
+// Its intended to be .0025 uosmo / gas
+var ConsensusMinFee sdk.Dec = sdk.NewDecWithPrec(25, 4)
```
mattverse, line 7: I was thinking of an approach using params so that the governance can revisit this if necessary in the future as well, I do think that constant should be good for now

mattverse (approved):
LGTM 🌮

x/txfees/types/constants.go
ValarDragon: I don't want to bother with params for now, since I would like to just get to EIP-1559 within 3 months

tests/e2e/containers/containers.go
```diff
@@ -62,7 +62,7 @@ func (m *Manager) ExecTxCmd(t *testing.T, chainId string, containerName string,
 // namely adding flags `--chain-id={chain-id} -b=block --yes --keyring-backend=test "--log_format=json"`,
 // and searching for `successStr`
 func (m *Manager) ExecTxCmdWithSuccessString(t *testing.T, chainId string, containerName string, command []string, successStr string) (bytes.Buffer, bytes.Buffer, error) {
-	allTxArgs := []string{fmt.Sprintf("--chain-id=%s", chainId), "-b=block", "--yes", "--keyring-backend=test", "--log_format=json"}
+	allTxArgs := []string{fmt.Sprintf("--chain-id=%s", chainId), "-b=block", "--yes", "--keyring-backend=test", "--log_format=json", "--fees=875uosmo"}
```
nicolaslara, line 65: This seems to be failing on E2E when executing https://github.com/osmosis-labs/osmosis/blob/nicolas/packet-forward-middleware/tests/e2e/configurer/upgrade.go#L140. 

It seems to me like this the user just doesn't have the funds to pay for fees.

tests/e2e/initialization/config.go
```diff
@@ -347,10 +355,35 @@ func updateMintGenesis(mintGenState *minttypes.GenesisState) {
...
+	pool1, err := balancer.NewBalancerPool(1, feePoolParams, feePoolAssets, "", time.Unix(0, 0))
+	if err != nil {
+		panic(err)
+	}
+	anyPool, err := types1.NewAnyWithValue(&pool1)
+	if err != nil {
+		panic(err)
+	}
+	gammGenState.Pools = []*types1.Any{anyPool}
+	gammGenState.NextPoolNumber = 2
```
p0mvn, line 386: Can we remove this and instead re-create this via CLI?

This interferes with pool IDs that we are creating via CLI

ValarDragon: Two questions:
- Why does it interfere?
- Where do it in CLI such that it comes first? It needs to be first so that other txs can pay fees, and we'd have to make distinct custom methods for paying fees to set this up.

p0mvn: 1. The reason is that we also create pre-upgrade pools as a result the pool is of 1 is already assigned after upgrade. It is possible to work around this by first making a separate PR to v14 genesis.

2. Makes sense. I will work around number 1 then and make the changes to v14 before main

p0mvn: Actually, for 2, it should be possible to create a pool with the fee via pool file, right?

ValarDragon: The pool file doesn't contain the tx fee. (Adding something custom for the tx fee for initialization isn't really a big deal. It just has to come first, if you point me to where I do that, I can add it in relatively easily)

tests/e2e/configurer/chain/commands.go
```diff
@@ -127,7 +128,7 @@ func (n *NodeConfig) SendIBCTransfer(from, recipient, amount, memo string) {
 
 	cmd := []string{"osmosisd", "tx", "ibc-transfer", "transfer", "transfer", "channel-0", recipient, amount, fmt.Sprintf("--from=%s", from), "--memo", memo}
 
-	_, _, err := n.containerManager.ExecTxCmdWithSuccessString(n.t, n.chainId, n.Name, cmd, "code: 0")
+	_, _, err := n.containerManager.ExecTxCmd(n.t, n.chainId, n.Name, cmd)
```
p0mvn, line 131: Why was this done?

tests/e2e/containers/containers.go
```diff
@@ -58,11 +60,25 @@ func (m *Manager) ExecTxCmd(t *testing.T, chainId string, containerName string,
...
+var txArgs = []string{"-b=block", "--yes", "--keyring-backend=test", "--log_format=json"}
+var txDefaultGasArgs = []string{"--gas=400000", "--fees=1000" + initialization.E2EFeeToken}
+
 // ExecTxCmdWithSuccessString Runs ExecCmd, with flags for txs added.
-// namely adding flags `--chain-id={chain-id} -b=block --yes --keyring-backend=test "--log_format=json"`,
+// namely adding flags `--chain-id={chain-id} -b=block --yes --keyring-backend=test "--log_format=json" --gas=400000`,
 // and searching for `successStr`
 func (m *Manager) ExecTxCmdWithSuccessString(t *testing.T, chainId string, containerName string, command []string, successStr string) (bytes.Buffer, bytes.Buffer, error) {
-	allTxArgs := []string{fmt.Sprintf("--chain-id=%s", chainId), "-b=block", "--yes", "--keyring-backend=test", "--log_format=json"}
+	allTxArgs := []string{fmt.Sprintf("--chain-id=%s", chainId)}
```
p0mvn, line 70: and this. I'm getting errors in bank send where the failure is stemming from some of these changes

tests/e2e/configurer/chain/commands.go
ValarDragon: This was a driveby change to get the execution stack to have one entry point. (At the time I did this line, I thought I'd do the gas if statement in this method, ended up not doing that)

So can be reverted, but looks identical to the ExecTxCmd as far as I can tell?

p0mvn: Yeah I think this is fine, was getting sidetracked by this but the change LGTM now.

I think I found the main problem. Should have the fix soon

p0mvn:
I think we can disable geometric TWAP as well. These are likely related

PR #4252, opened 2023-02-08
Add param to allow non lock owner to force unlock

iboss-ptk:

To allow an account to forcibly unlock any lock, including ones they do not own, if their address is on the whitelist. This functionality is needed for [transmigrator](https://github.com/osmosis-labs/transmigrator)

## Brief Changelog

  - Add `NonOwnerForceUnlockAddresses` to lockup params
  - By pass ownership check in `ForceUnlock` if  msg sender is in `NonOwnerForceUnlockAddresses`

## Testing and Verifying

  - Added unit test that validates if msg sender is in NonOwnerForceUnlockAddresses
    - they can force unlock their own locks 
    - they can force unlock their other's locks 


ValarDragon:
What is the precise permission we need to expose, this feels too permissive to expose

iboss-ptk:
@ValarDragon the contract needs to be able to:

unlock -> exit pool -> swap -> join pool -> lock

While retaining deposit period & needs to get everything unlocked locked back in some form.

The problem with existing approach is that if we authz ForceUnlock, `msg.Owner` will be the user not the contract. If we can get authz `MsgExec` sender somehow (which is the contract in this case), that might be better approach since it can be more restrictive. I'm not sure if that is possible.

mattverse:
feels like authz is a more preferrable way than using this method to me, @iboss-ptk  do you need help on further investigation if this is possible?

iboss-ptk:
@mattverse That would be helpful! I'm a bit lost in getting `MsgExec` sender info without modifying existing `MsgForceUnlock`. If there is a way for that, that would be ideal.

sunnya97:
@mattverse I don't see how authz would help here.  Cause normal users don't have ability to force unlock, so they can't authz the ability to do that either

sunnya97:
TODO:  Allow breaking of superfluid as well

p0mvn:
@iboss-ptk are we still planning to pursue this? If not, let's close.

PR #4262, opened 2023-02-08
fix(x/twap): geometric twap genesis validation

p0mvn:
Closes: #XXX

This broke state-exported testnet upgrade.

Geometric twap is allowed to be negative. This PR relaxes the validation and adds tests.


niccoloraspa:
Thanks for fixing It so quickly!

niccoloraspa:
Tested now. It works!

ValarDragon:
LGTM

Issue #4274, opened 2023-02-09
[CL Incentives] Implement `setIncentiveRecords` and `getIncentiveRecords`


AlpinYukseloglu:
## Background

Existing logic has tested behavior of incentive records without actually ever putting them in state. We should implement set/get functions that handle this behavior in a way that can be cleanly abstracted out of the CL module in the future.

## Suggested Design

- Key incentive records by pool ID, denom, and uptime
- Ideally scrap the `[]IncentiveRecords` field in pool proto and set/retrieve incentive records directly from state using pool ID (also generally lower surface area for bugs due to state falling out of sync)

## Acceptance Criteria

- All new and existing tests pass

PR #4283, opened 2023-02-10
[CL Incentives] Implement `setIncentiveRecords` and `getIncentiveRecords`

AlpinYukseloglu:
Closes: #4274

This PR implements a more portable version of state reads/writes relating to incentive records with the primary goal of making it simpler to abstract this logic into a separate module in the future. Specifically, it changes incentive records to work based off of pool Ids instead of as a field under pool objects.

## Brief Changelog

- Implement set/get functions for incentive records that do not depend on pool objects
- Remove the `PoolIncentives` field from CL pools

## Testing and Verifying

- The new functions/logic are tested in `incentives_test.go`


proto/osmosis/concentrated-liquidity/incentive_record.proto
```diff
@@ -10,26 +10,28 @@ option go_package = "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidit
 
 // IncentiveRecord represents an active perpetual incentive gauge for a pool
 message IncentiveRecord {
+  uint64 poolId = 1;
```
p0mvn, line 13: Why do we have `poolId` as part of the record if the key in state includes pool id as well?

proto/osmosis/concentrated-liquidity/incentive_record.proto
```diff
@@ -10,26 +10,28 @@ option go_package = "github.com/osmosis-labs/osmosis/v14/x/concentrated-liquidit
...
+
   // incentive_denom is the denom of the token being distributed as part of this
   // incentive record
-  string incentive_denom = 1;
+  string incentive_denom = 2;
```
p0mvn, line 17: Same with denom

proto/osmosis/concentrated-liquidity/incentive_record.proto
```diff
@@ -38,7 +40,7 @@ message IncentiveRecord {
   // min_uptime is the minimum uptime required for liquidity to qualify for this
   // incentive. It should be always be one of the supported uptimes in
   // types.SupportedUptimes
-  google.protobuf.Duration min_uptime = 5 [
+  google.protobuf.Duration min_uptime = 6 [
```
p0mvn, line 43: and min uptime

proto/osmosis/concentrated-liquidity/incentive_record.proto
p0mvn: We should be able to avoid duplicating state in both key and the value. It should be possible to extract these fields from the key by using:
https://github.com/osmosis-labs/osmosis/blob/316005aca4dcc3c01ca8249934b9907014ff55e8/osmoutils/store_helper.go#L110

AlpinYukseloglu: Discussed offline and came to a conclusion on a good approach tracked here: #4284

x/concentrated-liquidity/incentives.go
```diff
@@ -213,3 +218,44 @@ func calcAccruedIncentivesForAccum(ctx sdk.Context, accumUptime time.Duration, q
...
+	}
+}
+
+// GetIncentiveRecord gets the incentive record corresponding to the passed in values from store
+func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string, minUptime time.Duration) (types.IncentiveRecord, error) {
+	store := ctx.KVStore(k.storeKey)
+	incentiveStruct := types.IncentiveRecord{}
+	key := types.KeyIncentiveRecord(poolId, denom, minUptime)
+
+	found, err := osmoutils.Get(store, key, &incentiveStruct)
```
p0mvn, line 245: Does it make sense to use `MustGet` for simplicity because we aren't using the `found` flag for anything other than doing the error check?

I think `Get` and the `found` flag are useful when we need to use the flag for controlling the future execution flow. On the other hand, if we simply return an error when `!found`, we might as well use the `MustGet` instead that encapsulates the error check and panics

x/concentrated-liquidity/store.go
```diff
@@ -101,3 +102,71 @@ func ParseFullPositionFromBytes(key, value []byte) (types.FullPositionByOwnerRes
...
+
+	// These may include irrelevant parts of the prefix such as the module prefix.
+	incentiveRecordKeyComponents := strings.Split(keyStr, types.KeySeparator)
+
+	// We only care about the last 3 components, which are:
+	// - pool id
+	// - incentive denom
+	// - min uptime
+
+	relevantIncentiveKeyComponents := incentiveRecordKeyComponents[len(incentiveRecordKeyComponents)-3:]
```
p0mvn, line 140: non-blocking Q: what do you think about grabbing these components from the end?

I did something similar here: https://github.com/osmosis-labs/osmosis/blob/792375a8ae5cf1b8d5ef41ac3f27bb1394abc5f4/x/concentrated-liquidity/store.go#L59-L64

However, I'm afraid that is not very extensible because any time we want to add a new proto field, this logic will break.

We might want to refactor this later to grab these components relative to the start as opposed to the end. Let me know what you think

p0mvn (approved):
LGTM

x/concentrated-liquidity/incentives.go
```diff
@@ -213,3 +218,44 @@ func calcAccruedIncentivesForAccum(ctx sdk.Context, accumUptime time.Duration, q
 
 	return incentivesToAddToCurAccum, poolIncentiveRecords, nil
 }
+
+// nolint: unused
+// setIncentiveRecords sets the passed in incentive records in state
+func (k Keeper) setIncentiveRecord(ctx sdk.Context, incentiveRecord types.IncentiveRecord) {
+	// Potentially need to set the pool's incentiveRecord field as well
```
czarcas7ic, line 225: What do you mean by this comment?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -593,12 +599,75 @@ func (s *KeeperTestSuite) TestUpdateUptimeAccumulatorsToNow() {
...
+	clKeeper.SetIncentiveRecord(s.Ctx, incentiveRecordOne)
+	poolOneRecord, err := clKeeper.GetIncentiveRecord(s.Ctx, clPoolOne.GetId(), incentiveRecordOne.IncentiveDenom, incentiveRecordOne.MinUptime)
+	s.Require().NoError(err)
+	s.Require().Equal(incentiveRecordOne, poolOneRecord)
+	allRecordsPoolOne, err := clKeeper.GetAllIncentiveRecordsForPool(s.Ctx, clPoolOne.GetId())
+	s.Require().NoError(err)
+	s.Require().Equal([]types.IncentiveRecord{incentiveRecordOne}, allRecordsPoolOne)
+
+	// Ensure records for other pool remain unchanged
+	_, err = clKeeper.GetIncentiveRecord(s.Ctx, clPoolTwo.GetId(), incentiveRecordOne.IncentiveDenom, incentiveRecordOne.MinUptime)
```
czarcas7ic, line 646: nit: can we check the output of this is nil as well

x/concentrated-liquidity/incentives_test.go
```diff
@@ -593,12 +599,75 @@ func (s *KeeperTestSuite) TestUpdateUptimeAccumulatorsToNow() {
...
 			} else {
 				s.Require().Error(err)
 			}
 		})
 	}
 }
+
+// Note: we test that incentive records are properly deducted by emissions in `TestUpdateUptimeAccumulatorsToNow` above.
+// This test aims to cover the behavior of a series of state read/writes relating to incentive records.
+func (s *KeeperTestSuite) TestIncentiveRecordsSetAndGet() {
```
czarcas7ic, line 615: This test was super clear and easy to follow, nice job :)

x/concentrated-liquidity/store.go
```diff
@@ -101,3 +102,71 @@ func ParseFullPositionFromBytes(key, value []byte) (types.FullPositionByOwnerRes
...
+// ParseFullIncentiveRecordFromBz parses a IncentiveRecord from a byte array.
+// Returns a struct containing the denom and min uptime associated with the incentive record.
+// Returns an error if the byte array is empty.
+// Returns an error if fails to parse.
+func ParseIncentiveRecordFromBz(bz []byte) (incentiveRecord types.IncentiveRecord, err error) {
+	if len(bz) == 0 {
+		return types.IncentiveRecord{}, errors.New("incentive record not found")
+	}
+	err = proto.Unmarshal(bz, &incentiveRecord)
+	return incentiveRecord, err
```
czarcas7ic, line 115: shouldn't we catch the error here to ensure if an error occurs that the incentive record that gets returned is empty? (I assume this is done by Unmarshal itself but seems just nice to guarantee on our end)

czarcas7ic (approved):
👍

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Hmm I opted for `Get` because I wanted to avoid putting a panic in a getter (this also seems to be a design pattern we follow elsewhere) – did you mean using `MustGet` and then catching the panic to return an error?

x/concentrated-liquidity/store.go
AlpinYukseloglu: Unmarshal does handle this, but I agree that for readability it's better to catch the error. Added!

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Done!

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: This was from before I removed `incentiveRecord` as a pool field – deleted this comment now.

x/concentrated-liquidity/store.go
AlpinYukseloglu: I figure even if we add a proto field it is unlikely we will add it to the key (in fact if it's in the key we probably won't make it a proto field), so I don't foresee us breaking these unless we do a major refactor to pool/position protos.

This is since I'm assuming these fields will always be sufficient to identify a position (or incentive record in my case), so we will likely leave the key unchanged. Does this align with your thinking or did you mean something else?

x/concentrated-liquidity/incentives.go
p0mvn: I see - this sgtm. 

I was coming from the perspective of reducing the code boilerplate. However, your explanation justified keeping it as is

x/concentrated-liquidity/store.go
p0mvn: Yeah, that makes sense - agreed

p0mvn (approved):
Nice work. Merging since 2 approvals + all comments have been addressed

AlpinYukseloglu:
Reverting since the base branch is already quite large & difficult to review. Made new PR #4295 that we should hold off on merged until #4231 is reviewed and merged

Issue #4266, opened 2023-02-09
feat(dev-ux): go.mod update tool


p0mvn:
## Background

Currently, when we update one of:
- `osmoutils`
- `osmomath`
- `x/ibc-hooks`

our local tests pick up the changes from the modules right away. Howevever, the e2e test depends on the `go.mod` being correctly tagged with the latest version.

As a result, an engineer has to manually run the following command to update the root `go.mod`:
```
go get < module > @< commit hash >

go mod tidy
```

We need a way to automate this process.

## Suggested Design

The end goal of this is the following:

- detect if osmoutils, osmomath and x/ibc-hooks modules were updated in a PR
- if so, auto-open PR to your branch with global go.mod updated to the deps in your branch

I propose splitting this into the following steps / PRs:

1. Create a script that when run, does the following:
```bash
# detect current branch

go get github.com/osmosis-labs/osmosis/osmomath@< branch hash >
go get github.com/osmosis-labs/osmosis/osmoutils@< branch hash >
go get github.com/osmosis-labs/osmosis/x/ibc-hooks@< branch hash >

go mod tidy
```

Note that for this to work, the current branch must be pushed to remote.

2. Create a CI job that detects if any of `osmomath`, `osmoutils`, x/ibc-hooks` had any changes. 
   * If so, it runs the script and opens a PR against your branch with the changes from script 1 applied.

## Acceptance Criteria
- The process of updating `go.mod` is fully automated
- Successfully executes every time when an update to `osmomath`, ``osmoutils` or `x/ibc-hooks` is made
- Auto-opens a PR against developer's branch with the `go.mod` changes

PR #4291, opened 2023-02-11
go.mod auto update

pysel:
Closes: #4266

This PR introduces a new workflow that is being ran when PR is created with changes in one of:
- ibc-hooks
- osmoutils
- osmomath

## Brief Changelog

When one of those files are changed, a script is ran, which updates versions of ibc-hooks, osmoutils and osmomath in `go.mod` and creates a PR against a branch, that triggered this action.

## Testing and Verifying

https://github.com/osmosis-labs/osmosis/pull/4291#issuecomment-1459619762

no

### DIFF
--- a/.github/workflows/go-mod-auto-bump.yml
+++ b/.github/workflows/go-mod-auto-bump.yml
@@ -1,0 +1,34 @@
+# This workflow is used in order to automatically update go.mod in case one of:
+# * osmoutils
+# * osmomath
+# * ibc-hooks
+# is updated in local branch.
+name: Go Mod Auto Version Update
+on: [pull_request]
+
+jobs:
+  update_go_mod:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+        with: 
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '>=1.19.0'
+
+      - name: Run script 
+        run: bash ./scripts/update-go-mod.sh
+
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@v4
+        with: 
+          token: ${{ secrets.COMMIT_TO_BRANCH }}
+          base: ${{ github.base_ref }}
+          commit-message: go.mod update
+          delete-branch: true
+          branch: create-pull-request/upgrade-go-mod
+          title: go.mod update
+          body: Updated go.mod due to changes in osmoutils, osmomath or x/ibc-hooks on branch ${{ github.head_ref }}
--- a/.github/workflows/go-mod-auto-bump.yml
+++ b/.github/workflows/go-mod-auto-bump.yml
@@ -42,6 +42,6 @@ jobs:
           delete-branch: true
           branch: create-pull-request/upgrade-go-mod
           title: Update go.mod
-          body: Updated go.mod due to changes in osmoutils, osmomath or x/ibc-hooks 
+          body: Updated go.mod due to recent push to main with changes in osmoutils, osmomath or x/ibc-hooks 
           labels: |
             V:state/compatible/no_backport
--- a/.github/workflows/go-mod-auto-bump.yml
+++ b/.github/workflows/go-mod-auto-bump.yml
@@ -23,6 +23,7 @@ jobs:
         run: bash ./scripts/update-go-mod.sh
 
       - name: Create pull request
+        if: github.head_ref != 'create-pull-request/upgrade-go-mod' # This makes sure the PR is not created when this worflow is ran from PR's that this workflow created initially
         uses: peter-evans/create-pull-request@v4
         with: 
           token: ${{ secrets.COMMIT_TO_BRANCH }}
--- a/.github/workflows/go-mod-auto-bump.yml
+++ b/.github/workflows/go-mod-auto-bump.yml
@@ -4,7 +4,10 @@
 # * ibc-hooks
 # is updated in local branch.
 name: Go Mod Auto Version Update
-on: [pull_request]
+on: 
+  push:
+    branches:
+      - main
 
 jobs:
   update_go_mod:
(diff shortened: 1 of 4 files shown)


osmoutils/cache_ctx_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 var expectedOutOfGasError = types.ErrorOutOfGas{Descriptor: "my func"}
 
+// osmoutils change
```
pysel, line 12: triggers action

pysel:
PR created by this action: https://github.com/osmosis-labs/osmosis/pull/4537

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow is used in order to automatically update go.mod in case one of:
+# * osmoutils
+# * osmomath
+# * ibc-hooks
+# is updated in local branch.
+name: Go Mod Auto Version Update
+on: [pull_request]
```
nicolaslara, line 7: Is there a way to run this only when a PR gets merged? IIUC, this will create a PR to the branch that modified the dependencies. I think what we want is to create a new PR to main once something that modifies the dependencies gets merged. This way we know that the commit being tagged is a commit that has been merged to main.

pysel: interesting suggestion, I will look into that, thanks!

pysel:
hey @nicolaslara, I found a way to do that how you requested. Unfortunately, there is no `merged` (or something like that) sub event in `pull request` event for github workfows, however, this can be done by specifying `on.push`. By specifying `on.push`, that will also mean that it will run when PR is merged, so I guess that is exactly what we need.

More specifically, by specifying

```
  paths: 
     - osmoutils/**
     - osmomath/**
     - x/ibc-hooks/**
```
, it will only run on pushes that changed these folders.

The rest of the logic is as follows: take commit before push and the most recent after the push, see the `git diff` between them, and update `osmoutils`, `osmomath` and/or `ibc-hooks` in go.mod based on the diff. Then, creates a PR to main.

Limitations: it only works if you are pushing to `main`, if you create a PR to the branch, other than `main`, it won't trigger the workflow after it gets merged. However, I thought that usually if you have this situation, the branch you create a PR to will eventually be merged to `main`, hence, will eventually update go.mod. If a desired functionality is to make it possible to update go.mod even when PRs are merged to the branches other than `main`, I will do that

nicolaslara (approved):
This is great! We would still need to updating the dependencies on the branches to get the tests to pass, but I think that's expected (and probably as easy as pushing and then merging a PR)

ValarDragon:
hrmm, I think this is going to bloat our PR list, I wonder if we can commit directly.

pysel:
> hrmm, I think this is going to bloat our PR list, I wonder if we can commit directly, or comment with a github suggestion

yeah, that [can be done](https://github.com/osmosis-labs/osmosis/pull/4291/commits/c8d9465f4671da3e6799fa07a1f7da6c86c1e3f4).

It will directly push changes to `main` now, if a push action happened on `main`, that changes one of the specified folders

@nicolaslara, wdyt about this new design?

nicolaslara:
I like that. We can always verify the latest hash right before a release

.github/workflows/go-mod-auto-bump.yml
p0mvn: The reason why we wanted a PR to PR was because e2e oftentimes fails if you don't retag the dependencies.

For this reason, we might either want a PR to PR or a direct commit to PR with the changes applied

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -0,0 +1,40 @@
+# This workflow is used in order to automatically update go.mod in case one of:
+# * osmoutils
+# * osmomath
+# * ibc-hooks
+# is updated when a push to the main branch happened (including merged PRs)
```
p0mvn, line 5: I saw previous discussions but I still think that we need to figure out a way to commit directly to the developer's branch / PR.

If you make a breaking change to `osmoutils`, `osmomath`, or `x/ibc-hooks`, e2e CI fails unless you manually retag.

The biggest value add of this feature IMO is automating such manual retagging. Please let me know what you think

nicolaslara:
We could have the automatic commit solution on every branch, not only main.

pysel:
Hey @nicolaslara, @p0mvn. I changed the code according to your requests. Right now, the push will be made to any branch that changes any of those folders. 


Here is a [commit ](https://github.com/osmosis-labs/osmosis/pull/4291/commits/d46f1408b659d51ae6a3262bd0ee65c4647f3f93) with changes that was triggered by [commit that changed osmoutils](https://github.com/osmosis-labs/osmosis/pull/4291/commits/f45fc18f8e655f99262ac35d8448620b0a2c6706)

nicolaslara (approved):
LGTM

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -0,0 +1,45 @@
...
```
p0mvn, line 5: Not only `main` anymore, right? Can you please update this if so?

osmoutils/cache_ctx_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 var expectedOutOfGasError = types.ErrorOutOfGas{Descriptor: "my func"}
 
+// change
```
p0mvn, line 12: let's remove comment please

p0mvn (approved):
Great work @pysel 

Please resolve nits and rebase, and I think we can merge this!

.github/workflows/go-mod-auto-bump.yml
pysel: yeah, any branch now, gonna update right away

scripts/update-go-mod.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ $update_osmomath -eq 1 ]
+then 
+    go get github.com/osmosis-labs/osmosis/osmomath@$commit_after
+fi
+
+if [ $update_ibc_hooks -eq 1 ]
+then 
+    go get github.com/osmosis-labs/osmosis/x/ibc-hooks@$commit_after
+fi
```
p0mvn, line 45: Apologies for the breakage but we just merged another module separated in go.mod: https://github.com/osmosis-labs/osmosis/pull/4336

Would you mind quickly updating the script to include `epochs` as well, please?

pysel: no worries, it does not seem to be difficult to add

p0mvn (approved):
Nice, thank you!

PR #4299, opened 2023-02-13
(CL) feat: Wire Twap to Concentrated Liquidity

mattverse:

Wires Twap to Concentrated Liquidity. 
This PR also includes the process of wiring pool manager to the twap module, while previously the gamm keeper was directly wired to the twap keeper. 

## Brief Changelog
- Wire Twap to Concentrated Liquidity Pools

## Testing and Verifying
- This PR adds tests to new features


app/keepers/keepers.go
```diff
@@ -33,6 +33,10 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/upgrade"
 	upgradekeeper "github.com/cosmos/cosmos-sdk/x/upgrade/keeper"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	icq "github.com/strangelove-ventures/async-icq/v4"
+	icqtypes "github.com/strangelove-ventures/async-icq/v4/types"
```
czarcas7ic, line 38: What was the reasoning for reordering here? Seems like this PR should have nothing to do with icq

x/concentrated-liquidity/pool.go
```diff
@@ -101,6 +101,41 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
...
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	sqrtPrice := concentratedPool.GetCurrentSqrtPrice().Power(2)
+	// if sqrt price is zero, it means that two assets are with same ratio
+	// this we state that spot price is 1
+	if sqrtPrice.IsZero() {
+		return sdk.OneDec(), nil
+	}
```
czarcas7ic, line 130: This logic does not follow for me. Sqrt price of zero should not be possible, and even if it was, a sqrtPrice of zero should not equal a spot price of 1.

x/concentrated-liquidity/pool.go
```diff
@@ -101,6 +101,41 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
...
+	poolId uint64,
+	quoteAssetDenom string,
+	baseAssetDenom string,
+) (spotPrice sdk.Dec, err error) {
+	concentratedPool, err := k.getPoolById(ctx, poolId)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	sqrtPrice := concentratedPool.GetCurrentSqrtPrice().Power(2)
```
czarcas7ic, line 125: This is no longer a sqrtPrice, just a price since we are squaring it.

x/concentrated-liquidity/pool_test.go
```diff
@@ -199,3 +199,48 @@ func (s *KeeperTestSuite) TestPoolIToConcentratedPool() {
...
+	// set up default position to change spot price
+	s.SetupDefaultPosition(defaultPoolId)
+
+	spotPriceBaseUSDC, err := s.App.ConcentratedLiquidityKeeper.CalculateSpotPrice(s.Ctx, concentratedPool.GetId(), ETH, USDC)
+	s.Require().NoError(err)
+	s.Require().Equal(spotPriceBaseUSDC, DefaultCurrSqrtPrice.Power(2))
+
+	spotPriceBaseETH, err := s.App.ConcentratedLiquidityKeeper.CalculateSpotPrice(s.Ctx, concentratedPool.GetId(), USDC, ETH)
+	s.Require().NoError(err)
+	s.Require().Equal(spotPriceBaseETH, sdk.OneDec().Quo(spotPriceBaseUSDC))
```
czarcas7ic, line 241: For clarity, can we change `spotPriceBaseUSDC` here to `DefaultCurrSqrtPrice.Power(2)`?

x/concentrated-liquidity/pool_test.go
```diff
@@ -199,3 +199,48 @@ func (s *KeeperTestSuite) TestPoolIToConcentratedPool() {
...
+
+	// try getting spot price from a non-existent pool
+	_, err = s.App.ConcentratedLiquidityKeeper.CalculateSpotPrice(s.Ctx, concentratedPool.GetId()+1, USDC, ETH)
```
czarcas7ic, line 244: Lets check spotPrice isNil on error

x/gamm/keeper/pool_service_test.go
```diff
@@ -304,6 +304,9 @@ func (suite *KeeperTestSuite) TestInitializePool() {
 				senderBalBeforeNewPool := bankKeeper.GetAllBalances(suite.Ctx, sender)
 
 				// initializePool with a poolI
+				// initializePool shoould be called by pool manager in practice.
+				// We set pool route here to make sure hoks from InitializePool do not break
```
czarcas7ic, line 308:
```suggestion
				// We set pool route here to make sure hooks from InitializePool do not break
```

x/poolmanager/router_test.go
```diff
@@ -98,6 +98,147 @@ func (suite *KeeperTestSuite) TestGetPoolModule() {
...
+			baseAssetDenom:    "baz",
+			expectedSpotPrice: sdk.MustNewDecFromStr("0.99999998"),
+		},
+		"valid concentrated liquidity pool": {
+			preCreatePoolType: types.Concentrated,
+			poolId:            1,
+			quoteAssetDenom:   "usdc",
+			baseAssetDenom:    "eth",
+			expectedSpotPrice: sdk.OneDec(),
+		},
```
czarcas7ic, line 199: Ahh I see why you added that spot price logic earlier, because when we initialize a CL pool there are no assets and it defaults to a sqrtPrice of zero. 

The logic you added though still needs to be removed, and I suggest for this test we just add a logic gate to initialize a position if its a concentrated pool types so we can get a valid spot price

x/twap/logic_test.go
```diff
@@ -134,7 +135,7 @@ func (s *TestSuite) TestNewTwapRecord() {
 	}
 	for name, test := range tests {
 		s.Run(name, func() {
-			twapRecord, err := twap.NewTwapRecord(s.App.GAMMKeeper, s.Ctx, test.poolId, test.denom0, test.denom1)
+			twapRecord, err := twap.NewTwapRecord(s.App.PoolManagerKeeper, s.Ctx, test.poolId, test.denom0, test.denom1)
```
czarcas7ic, line 138: This test is initialized with a balancer pool only. I think we need to refactor this test to also include concentrated pool as well.

x/concentrated-liquidity/pool.go
czarcas7ic: See comment here https://github.com/osmosis-labs/osmosis/pull/4299#discussion_r1105982854

app/keepers/keepers.go
mattverse: This was caused due to fixing merge conflict then go fmt, unfortunately I do not know the exact reason why or how this works, but go fmt would automatically make this change since I have fixed merge conflict :(

x/poolmanager/router_test.go
mattverse: @czarcas7ic  When a pool is created and a position has been created with equal amounts of each asset, the current tick (sqrt price) would stay the same (=zero)

x/concentrated-liquidity/pool.go
mattverse: See comment here https://github.com/osmosis-labs/osmosis/pull/4299#discussion_r1107408872

x/concentrated-liquidity/pool.go
mattverse: https://github.com/osmosis-labs/osmosis/pull/4299/commits/1d1761106a70069bfd63ffe1427cbad0ebf6e9ca

x/concentrated-liquidity/pool_test.go
mattverse: https://github.com/osmosis-labs/osmosis/pull/4299/commits/1d1761106a70069bfd63ffe1427cbad0ebf6e9ca

x/concentrated-liquidity/pool_test.go
mattverse: https://github.com/osmosis-labs/osmosis/pull/4299/commits/1d1761106a70069bfd63ffe1427cbad0ebf6e9ca

x/twap/logic_test.go
mattverse: https://github.com/osmosis-labs/osmosis/pull/4299/commits/1d1761106a70069bfd63ffe1427cbad0ebf6e9ca

x/poolmanager/router_test.go
czarcas7ic: This should not be the case. My guess is that you potentially did not re-fetch the pool when re-querying the spot price, so you got the same spot price as when the pool was created (I know because I have made this mistake a couple of times now haha)

mattverse: Edited the logic so that on zero spot price it would return zero as is (could not completely eliminate the logic gate of checking if spotPrice is not zero since this would lead to zero division errors)

x/concentrated-liquidity/pool.go
```diff
@@ -101,6 +101,38 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
...
+	concentratedPool, err := k.getPoolById(ctx, poolId)
+	if err != nil {
+		return nil, err
+	}
+
+	denoms := []string{concentratedPool.GetToken0(), concentratedPool.GetToken1()}
+	return denoms, nil
+}
+
+func (k Keeper) CalculateSpotPrice(
```
p0mvn, line 114: Let's have table-driven test cases for this function, please. It is important to cover all edge cases here, including the rounding logic when we do `price = sdk.OneDec().Quo(price)` as `price` can become zero.

It would be great to separate the spot price function into a separate PR for faster merge and review cycle as well as to be able to review all edge cases in isolation.

Some of the edge cases around spot price erros can be referenced from gamm:
https://github.com/osmosis-labs/osmosis/blob/be95710eca5d4feb982c946b13716896a4e0f8f7/x/gamm/keeper/pool_service.go#L47-L53

mattverse: @p0mvn  Which case are you talking about when you are referring by the following statement? 
> the rounding logic when we do price = sdk.OneDec().Quo(price) as price can become zero.

I personally think that if the sqrt price is too big and `price` results in 0 due to 1/ <large_number> we should still be returning 0., which is how the current logic handles. Do you have any other suggestion for this?

mattverse: I could definitely split this logic out into a separate PR for faster iterations and easier reviews once the question above has been answered

p0mvn: I think we should standardize the API between x/gamm and x/concentrated-liquidity. Given that x/gamm returns `sdk.Dec{}` in that case, I think we should do the same here for compatibility reasons. 

Also, a price cannot be zero, so I think returning `sdk.Dec{}` is more correct.

What's most important is returning error when spot price is zero because twap directly depends on an error being returned in such a case [here](https://github.com/osmosis-labs/osmosis/blob/106ac2058f126997e7f73e549390f065b7b02b49/x/twap/logic.go#L53)

x/concentrated-liquidity/pool.go
```diff
@@ -101,6 +101,44 @@ func (k Keeper) setPool(ctx sdk.Context, pool types.ConcentratedPoolExtension) e
...
+	}
+	if quoteAssetDenom == concentratedPool.GetToken1() {
+		price = sdk.OneDec().Quo(price)
+	}
+
+	if price.GT(types.MaxSpotPrice) || price.LT(types.MinSpotPrice) {
+		return sdk.Dec{}, types.PriceBoundError{ProvidedPrice: price, MinSpotPrice: types.MinSpotPrice, MaxSpotPrice: types.MaxSpotPrice}
+	} else if !price.IsPositive() {
+		return sdk.Dec{}, types.SpotPriceNegativeError{ProvidedPrice: price}
+	}
```
p0mvn, line 137: I think this case is unreachable because if price is smaller than min spot price, it will hit the first condition

mattverse:
Closing this PR as this was replaced with PRs that were separated out

PR #4306, opened 2023-02-13
Prevent adding to a finished gauge

ValarDragon:

@JeremyParish69 pointed out that there is currently a bug, where it is possible to add more rewards to an already finished gauge.

This PR fixes that as a bug vector.

Furthermore it also fixes a bug in `IsFinishedGauge` and `IsActiveGauge`, which is only used in `InitGenesis` logic

## Brief Changelog

* Prevent `AddToGauge` from succeeding on a Finished gauge, from which funds can never get distributed.
* Fix an InitGenesis bug around ActiveGauge vs FinishedGauge classification

## Testing and Verifying

This change adds a unit test to ensure it works.


p0mvn (approved):
LGTM once the simulator is passing

PR #4317, opened 2023-02-14
fix(e2e): fix e2e test for consensus min gas fee #4244

p0mvn:
Closes: #XXX

This relates to #4244

- updating chain init image needed for upgrade testing
- fixing e2e chain init image upload in ci to Dockerhub: https://github.com/osmosis-labs/osmosis/pull/4319
- updating pre-upgrade pool id used in various tests 
- setting the config for the hermes relayer to account for the new fee logic
- fixing `SendIBC` to exclude fee token from the balance check
- adding docs

### DIFF
--- a/tests/e2e/scripts/hermes_bootstrap.sh
+++ b/tests/e2e/scripts/hermes_bootstrap.sh
@@ -42,7 +42,8 @@ account_prefix = 'osmo'
 key_name = 'val01-osmosis-a'
 store_prefix = 'ibc'
 max_gas = 6000000
-gas_price = { price = 0.000, denom = 'uosmo' }
+default_gas = 400000
+gas_price = { price = 0.0025, denom = 'e2e-default-feetoken' }
 gas_adjustment = 1.0
 clock_drift = '1m' # to accomdate docker containers
 trusting_period = '239seconds'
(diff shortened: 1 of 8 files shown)


tests/e2e/scripts/hermes_bootstrap.sh
```diff
@@ -57,7 +58,8 @@ account_prefix = 'osmo'
 key_name = 'val01-osmosis-b'
 store_prefix = 'ibc'
 max_gas = 6000000
-gas_price = { price = 0.000, denom = 'uosmo' }
+default_gas = 400000
```
ValarDragon, line 61: Why are we putting default gas? Won't we get out of gas errors routinely?

p0mvn: AFAIU, this is what the relayer assumes for calculating fees when estimation fails ref: https://github.com/informalsystems/hermes/issues/1457

We use the value of 40k in the [tx arguments](https://github.com/osmosis-labs/osmosis/blob/b7d5fbddecf1df4743de3088c0a945b971f3a3e2/tests/e2e/containers/containers.go#L29) as well so I just reused it.

p0mvn: Btw, the only remaining problem is the geometric twap failing for reasons I cannot currently explain. Please let me know if something immediately comes to mind as the solution.

I'm going to keep investigating in the meantime

PR #4320, opened 2023-02-14
feat: crosschain registry contract

czarcas7ic:
Closes: #4437 

While crosschain-swaps is currently working, creating the memo to successfully execute the routing and swaps requires knowledge of:
* The XCS contract address on the destination chain
* The specific channels these funds must be routed through in order to unwrap into the desired native denom
* What underlying asset the ibc denom correlates to

This contract acts as a helper contract to crosschain-swaps in order to abstract this complexity away by:
* Creating a registry for contract alias <> contract address
* Creating a registry for chain pair <> channel id
* Creating a registry for channel ID on a source chain <> destination chain
* Creating a helper function to unwrap a transfer message to its underlying native denomination


cosmwasm/contracts/crosschain-registry/src/helpers.rs
```diff
@@ -0,0 +1,76 @@
...
+        "ustars".to_string(),
+        "osmo".to_string(),
+        "ibc/987C17B11ABC2B20019178ACE62929FE9840202CE79498E29FE8E5CB02B7C0A4".to_string(),
+    )?;
+
+    Ok(deps)
+}
+
+pub fn make_chain_channel_key(source_chain: &str, destination_chain: &str) -> String {
+    format!("{}|{}", source_chain, destination_chain)
```
nicolaslara, line 71: you can use composite keys in the Map and it will take care of turning the keys into strings internally. For example, from the xcs contract: `Map<(&str, u64), ibc::IBCTransfer>`

nicolaslara: You can also create a type for the key, but then it has to implement the PrimaryKey trait which is not hard, but not trivial either. Most of the times I think it's enough to use a tuple. ref: https://github.com/CosmWasm/cw-storage-plus#key-types

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,5 @@
+use cw_storage_plus::Map;
+
+pub const CONTRACT_MAP: Map<&str, String> = Map::new("contract_map");
+pub const CHAIN_CHANNEL_MAP: Map<&str, String> = Map::new("chain_channel_map");
+pub const ASSET_MAP: Map<&str, String> = Map::new("asset_map");
```
nicolaslara, line 5: I find it useful to add comments / docs to states to explain what the keys and values represent.

Oh, and in rust `//` is a comment and `///` is a doc (a comment that rustdoc will understand when generating doc pages)

cosmwasm/contracts/crosschain-registry/src/helpers.rs
czarcas7ic: I am not quite understanding the proposed alternative. How do I concatenate two values to be used as a single key without formatting as I did above?

czarcas7ic: Nicolas explained this further on a call and I implemented the suggestion!

cosmwasm/contracts/crosschain-registry/src/msg.rs
```diff
@@ -0,0 +1,139 @@
...
+    GetAddressFromAlias { contract_alias: String },
+
+    #[returns(GetChainToChainChannelLinkResponse)]
+    GetChainToChainChannelLink {
+        source_chain: String,
+        destination_chain: String,
+    },
+
+    #[returns(crate::helpers::QueryDenomTraceResponse)]
+    GetDenomTrace { hash: String },
```
nicolaslara, line 126: Is the idea here that we "reexport" the denom trace query so that other contracts can use it without having to define the prost structs?

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,25 @@
...
+
+// CONTRACT_ALIAS_MAP is a map from a contract alias to a contract address
+pub const CONTRACT_ALIAS_MAP: Map<&str, String> = Map::new("contract_alias_map");
+
+// CHAIN_TO_CHAIN_CHANNEL_MAP is a map from a source<>destination chain pair to its respective channel id
+pub const CHAIN_TO_CHAIN_CHANNEL_MAP: Map<(&str, &str), String> =
+    Map::new("chain_to_chain_channel_map");
+
+// CHANNEL_TO_CHAIN_CHAIN_MAP is a map from a channel id on a source chain to its respective destination chain
+pub const CHANNEL_TO_CHAIN_CHAIN_MAP: Map<(&str, &str), String> =
```
nicolaslara, line 14: can we rename to `CHANNEL_ON_CHAON_CHAIN_MAP` ?

"to" makes it seem like we are mapping a channel to a chain

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,25 @@
...
+    Map::new("channel_to_chain_chain_map");
+
+// NATIVE_DENOM_TO_IBC_DENOM_MAP is a map from a native cosmos chain denom to its respective osmosis ibc denom
+pub const NATIVE_DENOM_TO_IBC_DENOM_MAP: Map<&str, String> = Map::new("osmosis_denom_map");
```
nicolaslara, line 18: assuming a direct ibc transfer, right? This is the one that can be calculated via:
```rust 
let chain_channel = CHANNEL_TO_CHAIN_CHAIN_MAP.load((chain_name, "osmosis"))?;
sha256sum(format!("transfer/{chain_channel}/{denom}"))
```
right?

czarcas7ic: We had previously wrote up in the design doc that this was a required registry, but I believe you are right (if you change load in your example above to take both the chain channel and the chain name)

cosmwasm/contracts/crosschain-registry/src/msg.rs
czarcas7ic: Yes! Was hoping this would just be the one stop shop for all info/queries CXS requires

cosmwasm/contracts/crosschain-registry/src/state.rs
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4320/commits/740d4bdd8cb18a6dea53f9469988c97aa0a30d7b

cosmwasm/contracts/crosschain-registry/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
nicolaslara, line 1: remove this file

cosmwasm/contracts/crosschain-registry/.gitignore
```diff
@@ -0,0 +1,16 @@
+# Build results
```
nicolaslara, line 1: remove this one too (we have these in the global gitignore)

cosmwasm/contracts/crosschain-registry/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+cosmwasm-std =  { workspace = true }
+cosmwasm-storage = { workspace = true }
+cw-storage-plus = { workspace = true }
+cw2 = { workspace = true }
+schemars = { workspace = true }
+osmosis-std = { workspace = true }
+serde = { workspace = true }
+thiserror = { workspace = true }
+serde-json-wasm = { workspace = true }
+serde-cw-value = { workspace = true }
```
nicolaslara, line 57: do we need all of these dependencies? Can we remove the ones we don't need?

cosmwasm/contracts/crosschain-registry/src/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+
+#[cfg_attr(not(feature = "imported"), entry_point)]
+pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    deps.api.debug(&format!("executing  query: {msg:?}"));
+    match msg {
+        QueryMsg::GetAddressFromAlias { contract_alias } => {
+            to_binary(&CONTRACT_ALIAS_MAP.load(deps.storage, &contract_alias)?)
+        }
+
+        QueryMsg::GetDestinationChainFromSourceChainViaChannel {
```
nicolaslara, line 62: very clear name! :)

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -0,0 +1,851 @@
+use crate::helpers::*;
+use crate::state::{CHAIN_TO_CHAIN_CHANNEL_MAP, CHANNEL_ON_CHAIN_CHAIN_MAP, CONTRACT_ALIAS_MAP};
+use cosmwasm_std::{Deps, DepsMut, Response, StdError};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::ContractError;
+
+// Enum to represent the operation to be performed
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
```
nicolaslara, line 10: you can rename this with `#[cw_serde]` which basically translates to the same derive.  

You can also add `#[serde(untagged)]` so that the name of the enum is not used when serializing. That is, ContractAliasInput becomes: `{"contract_alias_input": {"operation": "set", ...}}` instead of `{"contract_alias_input": {"operation": {"operation": "set", ...}}}`

nicolaslara: I like the operation pattern a lot, btw

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -0,0 +1,851 @@
...
+// Set, change, or remove a contract alias
+pub fn contract_alias_operations(
+    deps: DepsMut,
+    operations: Vec<ContractAliasInput>,
+) -> Result<Response, ContractError> {
+    let response = Response::new();
+    for operation in operations {
+        match operation.operation {
+            Operation::Set => {
+                if CONTRACT_ALIAS_MAP.has(deps.storage, &operation.alias) {
```
nicolaslara, line 37: we will probably want separate functions for each of these when we add gov handling

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -0,0 +1,851 @@
...
+                        source_chain: operation.source_chain.clone(),
+                        destination_chain: operation.destination_chain.clone(),
+                    })?;
+                if let Some(new_channel_id) = operation.new_channel_id {
+                    CHAIN_TO_CHAIN_CHANNEL_MAP.save(
+                        deps.storage,
+                        (&operation.source_chain, &operation.destination_chain),
+                        &new_channel_id,
+                    )?;
+                    CHANNEL_ON_CHAIN_CHAIN_MAP
```
nicolaslara, line 158: No need to remove. Save already overrides.

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,22 @@
...
+    Map::new("chain_to_chain_channel_map");
+
+// CHANNEL_TO_CHAIN_CHAIN_MAP is a map from a channel id on a source chain to its respective destination chain
+pub const CHANNEL_ON_CHAIN_CHAIN_MAP: Map<(&str, &str), String> =
+    Map::new("channel_to_chain_chain_map");
+
+pub const STATE: Item<State> = Item::new("state");
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct State {
```
nicolaslara, line 20: nit: I like calling these `Config`. Everything is state

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,22 @@
+use cosmwasm_std::Addr;
+use cw_storage_plus::{Item, Map};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+// CONTRACT_ALIAS_MAP is a map from a contract alias to a contract address
+pub const CONTRACT_ALIAS_MAP: Map<&str, String> = Map::new("contract_alias_map");
+
+// CHAIN_TO_CHAIN_CHANNEL_MAP is a map from a source<>destination chain pair to its respective channel id
+pub const CHAIN_TO_CHAIN_CHANNEL_MAP: Map<(&str, &str), String> =
```
nicolaslara, line 10: Did you experiment with using `Chain` and `ChannelId` here? If they implement `FromStr` we could also call create them by calling `"osmosis".into()`

cosmwasm/contracts/crosschain-registry/src/contract.rs
czarcas7ic: Slightly verbose but Ill take it for the clarity 😅

cosmwasm/contracts/crosschain-registry/src/execute.rs
czarcas7ic: How is that possible if the key changes? For instance, lets say the old key was channel-0<>osmosis which points to cosmos. If we want to change it to channel-1<>osmosis points to cosmos, if we don't remove the old key we will have two entries pointing to cosmos AFAI can tell. Lmk if I am misunderstanding.

cosmwasm/contracts/crosschain-registry/src/execute.rs
czarcas7ic: Thanks! Addressed the first comment here https://github.com/osmosis-labs/osmosis/pull/4320/commits/97be4979580e1b4dc2472366e3f4c34212461f0c

cosmwasm/contracts/crosschain-registry/src/state.rs
czarcas7ic: Agreed, changed! https://github.com/osmosis-labs/osmosis/pull/4320/commits/97be4979580e1b4dc2472366e3f4c34212461f0c

cosmwasm/contracts/crosschain-registry/src/state.rs
czarcas7ic: I did, and I am running into some issues (specifically stuck at having to implement a PrimaryKey to Chain in order to use load)

cosmwasm/contracts/crosschain-registry/Cargo.toml
czarcas7ic: Removed the ones I think were not needed https://github.com/osmosis-labs/osmosis/pull/4320/commits/ff227cffe240a9c7bd52660439a162e203ca2ea2

Is there a better way to do this other than just removing, saving, and seeing if there are any errors? 😅

cosmwasm/contracts/crosschain-registry/src/execute.rs
czarcas7ic: Discussed offline with Nicolas, if we decide to do this it can be in a separate PR. Please unresolve if this changes.

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: Ahh you're right. It's not possible.

cosmwasm/contracts/crosschain-registry/src/state.rs
nicolaslara: Ahh then there's no point. It's not like it adds so much clarity or we need many methods on that type

cosmwasm/contracts/crosschain-registry/Cargo.toml
nicolaslara: There are external tools for this. You could use [cargo udeps](https://github.com/est31/cargo-udeps) or, my new favourite because of the name, [cargo machete](https://blog.benj.me/2022/04/27/cargo-machete/)

tests/ibc-hooks/ibc_middleware_test.go
```diff
@@ -618,14 +639,178 @@ func (suite *HooksTestSuite) SetupCrosschainSwaps(chainName Chain) (sdk.AccAddre
...
+	registryAddr, token0CB, token0CBA, CBAPath := suite.SetupCrosschainRegistry(ChainA)
+
+	// Set the registry address to an alias
+	contractAlias := "osmosis_registry_contract"
+	suite.setContractAlias(registryAddr, contractAlias, ChainA)
+
+	// Retrieve the registry address from the alias
+	contractAddressFromAliasQuery := fmt.Sprintf(`{"get_address_from_alias": {"contract_alias": "%s"}}`, contractAlias)
+	contractAddressFromAliasQueryResponse := suite.chainA.QueryContract(&suite.Suite, registryAddr, []byte(contractAddressFromAliasQuery))
+	suite.Require().Equal(fmt.Sprintf("\"%s\"", registryAddr.String()), contractAddressFromAliasQueryResponse)
```
czarcas7ic, line 787: Note to reviewer: I have no idea why the address is returning with an extra set of quotes and I have no idea how to get rid of them in a non hacky way

nicolaslara: I'll review in more detail in a bit, but I think this is because we are json-serializing the string before returning it

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+    Map::new("chain_to_chain_channel_map");
+
+// CHANNEL_TO_CHAIN_CHAIN_MAP is a map from a channel id on a source chain to its respective destination chain
+pub const CHANNEL_ON_CHAIN_CHAIN_MAP: Map<(&str, &str), String> =
+    Map::new("channel_to_chain_chain_map");
```
nicolaslara, line 14: These strings are the storage prefix. We're not storing a lot of data here so it may not matter that they're long, but something to keep in mind for contracts with a lot of data.

An alternaative is to create an enum or a const that maps to short strings.

tests/ibc-hooks/ibc_middleware_test.go
nicolaslara: Right, so this is what we expect. Cosmwasm queries results as json encoded. We may want to create a struct so that they're return inside an object `{"contract_address": "..."}`

czarcas7ic: Got it, fixed here! https://github.com/osmosis-labs/osmosis/pull/4320/commits/64a70672fd9576013e35d625dff4615b3a799640

cosmwasm/contracts/crosschain-registry/src/state.rs
czarcas7ic: I think this is what you were describing!

https://github.com/osmosis-labs/osmosis/pull/4320/commits/d109bd04ceff1de8c67b4a5048ab3793ee3cad5e

nicolaslara:
@czarcas7ic should we normalize all inputs here (all denoms should be uppercase, all chain and channel names lowercase).

Thought about it when I was (temporarily) using the chain name as a bech32 prefix and it was invalid (no mixed case allowed).

nicolaslara:
We're probably going to need a registry mapping chains and their bech32 prefixes

nicolaslara (approved):
I think this is good to merge. We can improve on it on subsequent PRs.

What's left:
 * Permissions for managing the registries
 * Allowing other chains as receivers when unwrapping tokens
     * Guessing receiver chain based on prefix?
 * Use these registries in XCS (v2)
 * Docs

Issue #4303, opened 2023-02-13
Add IBC rate limits in the v15 upgrade


nicolaslara:

JohnnyWyles:
Approved by Governance: https://www.mintscan.io/osmosis/proposals/427

PR #4340, opened 2023-02-16
Added rate limits in upgrade

nicolaslara:
Closes: #4303 

This adds the rate limits in the v15 upgrade for when https://www.mintscan.io/osmosis/proposals/427 passes.

## Brief Changelog

 * Add rate limits in v15

## Testing and Verifying

 - [x] The upgrade has tests.
 - [x] The upgrade has E2E tests
 - [ ] This should also be tested on a testnet with state export. 


ValarDragon:
RIP goimports error. Do you have your editor set to format on save?

nicolaslara:
> RIP goimports error. Do you have your editor set to format on save?

I do, but it doesn't work when I fix conflicts and some other weird situations like that.


@ValarDragon  but!! this is not ready. The E2E tests are passing because I deactivated them after failing to get them to work (rip).

I asked Nicco to test the upgrade with a state export to make sure the issue is with E2E and not with the upgrade

nicolaslara:
That change makes a lot of sense! I thought that was just initialization. Thanks for the find!

nicolaslara:
hmm, this is still failing when there contracts exist on v14. I'll try to test this again tomorrow on a real node and see what happens

p0mvn:
Latest updates:
- wired ibc-rate-limit as SDK module to have init and export genesis logic
- changed querier to be code-gen compatible and fixed CLI
-  fixed bug with not getting params correctly due to not having a pointer receiver on `ParamSetPairs`: https://github.com/osmosis-labs/osmosis/blob/af89301de5270d9bb3f4b6e39c12d407ca331270/x/ibc-rate-limit/types/params.go#L44
- fixed bug with overriding contract parameter due to `ibc-rate-limit` consensus version being set to 1, leading to overwrite during the upgrade. Fixed with: https://github.com/osmosis-labs/osmosis/blob/af89301de5270d9bb3f4b6e39c12d407ca331270/app/upgrades/v15/upgrades.go#L56

We manually checked the presence of the contract parameter and rate limits post-upgrade. However, we should add a test in `e2e_test.go` doing that for us. I think this is the only remaining TODO here

tests/e2e/configurer/chain/chain.go
```diff
@@ -43,6 +53,8 @@ const (
 	waitUntilRepeatPauseTime = 2 * time.Second
 	// waitUntilrepeatMax is the maximum number of times to repeat the wait until condition.
 	waitUntilrepeatMax = 60
+
+	proposalStatusPassed = "PROPOSAL_STATUS_PASSED"
```
nicolaslara, line 57: This seems unrelated to this PR. Do we want it here?

p0mvn: This was used to ensure that the param change proposal passed before we tried to get the param itself. This could help to be more confident that the proposal not passing was the reason for the parameter not updating. However, that check seems to be removed now.

I'm cool with not blocking on this and just deleting `prposalStatusPassed`

p0mvn: Ahh, actually it's still being used for checking that the proposal passed here: https://github.com/osmosis-labs/osmosis/blob/376fccde469c3c4168127e48ee9beda44cee7647/tests/e2e/configurer/chain/chain.go#L285

So I think we can keep this

tests/e2e/e2e_test.go
```diff
@@ -445,23 +444,29 @@ func (s *IntegrationTestSuite) TestSuperfluidVoting() {
...
+	s.Require().NoError(err)
+
+	// Need to json unparshal the params because they are stored including quotes
+	paramA := nodeA.QueryParams(ibcratelimittypes.ModuleName, string(ibcratelimittypes.KeyContractAddress))
+	json.Unmarshal([]byte(paramA), &paramA)
+	paramB := nodeB.QueryParams(ibcratelimittypes.ModuleName, string(ibcratelimittypes.KeyContractAddress))
+	json.Unmarshal([]byte(paramB), &paramB)
+
+	// When upgrading to v15, we want to make sure that the rate limits have been set.
+	_, err = nodeA.QueryWasmSmartArray(paramA, `{"get_quotas": {"channel_id": "any", "denom": "ibc/E6931F78057F7CC5DA0FD6CEF82FF39373A6E0452BF1FD76910B93292CF356C1"}}`)
```
p0mvn, line 468: Do we want to check that the values (or at least some of them) are set correctly?

tests/e2e/configurer/chain/chain.go
nicolaslara: great!

tests/e2e/e2e_test.go
nicolaslara: good call! I'll check the last one (like on the test). If any of the other ones fail, it should panic before even executing that.

nicolaslara: actually, that query already fetches the latest value. if the upgrade failed that should be empty (`ibc/E6931F78057F7CC5DA0FD6CEF82FF39373A6E0452BF1FD76910B93292CF356C1` doesn't exist).  Added a length check

tests/e2e/e2e_test.go
```diff
@@ -445,23 +444,30 @@ func (s *IntegrationTestSuite) TestSuperfluidVoting() {
...
+	// Need to json unparshal the params because they are stored including quotes
+	paramA := nodeA.QueryParams(ibcratelimittypes.ModuleName, string(ibcratelimittypes.KeyContractAddress))
+	json.Unmarshal([]byte(paramA), &paramA)
+	paramB := nodeB.QueryParams(ibcratelimittypes.ModuleName, string(ibcratelimittypes.KeyContractAddress))
+	json.Unmarshal([]byte(paramB), &paramB)
+
+	// When upgrading to v15, we want to make sure that the rate limits have been set.
+	quotas, err := nodeA.QueryWasmSmartArray(paramA, `{"get_quotas": {"channel_id": "any", "denom": "ibc/E6931F78057F7CC5DA0FD6CEF82FF39373A6E0452BF1FD76910B93292CF356C1"}}`)
+	s.Require().Len(quotas, 2)
+	s.Require().NoError(err)
```
p0mvn, line 470: Should we repeat this check for both chains A and B?

The reason is that for chain A we overwrite the quotas in `TestIBCTokenTransferRateLimiting` so making this check against `nodeA` only does not validate the correctness of the upgrade

p0mvn (approved):
LGTM

PR #4341, opened 2023-02-16
Min consensus fee arithmetic twap debugging

p0mvn:
Closes: #XXX


p0mvn:
in favor of: https://github.com/osmosis-labs/osmosis/pull/4359

PR #4354, opened 2023-02-16
ProtoRev: Documentation, stateless msg validation + binary search edge case

davidterpay:
Most of the LOC here is due to the module specification (documentation) of `x/protorev`.

In order to have better trading and routing execution, we decided to move the `StepSize` to be specific to each hot route as oppose to having it be global to each token pair. For example, if we have a token pair like (Osmo, Atom) with 2 different cyclic arbitrage routes such as

- USDC -–> **ATOM**, **ATOM** ––> **OSMO**, **OSMO** --> USDC
- AKASH ––> JUNO, JUNO ––> **ATOM**, **ATOM** ––> **OSMO**, **OSMO** ––> AKASH

the underlying decimal points of USDC and AKASH may differ so having the same `StepSize` in the binary search may lead to worse execution. As such, we proposed and implemented the proto to include `StepSize` per cyclic arbitrage route so that `StepSize` is more closely coupled with the actual route and underlying number of decimal points of the arbitrage denomination.

Additionally, we add a few more stateless checks to a few msg types so no panics arise for nil pointers. Lastly, we add the `x/protorev` spec doc to the module.

## Brief Changelog
  - *Update the proto for HotRoutes to include `StepSize` on each cyclic arbitrage route*
  - *Deal with an edge case where the binary search only moves towards the maximum arbitrage input amount*
  - *Better stateless validation for a few msg types*
  - *Documentation for the module*

## Testing and Verifying
This change is already covered by existing tests.


x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+
+The same line of reasoning exists for Atom. `x/protorev` will attempt to find the highest liquidity pool between (Atom, Akash) and (Atom, Juno). If these pools exist, they will be added to the list of routes that can be simulated later in the pipeline. If not, the route is discarded.
+
+In both cases, the route that is built will always surround the pool of the original swap that was made. However, we allow for more flexibility in route generation as the highest liquidity method may not be optimal, hence the additional of hot routes.
+
+### Hot Route Method
+
+Populated through the admin account, the module’s keeper holds a KV store that associates token pairs (for example, osmo/juno) to the routes that result in a high percentage of arbitrage profit on Osmosis (as determined by external analysis).
+
+The purpose of storing Hot Routes is a recognition that the Highest Liquidity Pool method may not present the best arbitrage routes. As such, hot routes can be configured by the admin account to store additional routes that may be more effective at capturing arbitrage opportunities. Each hot route will store a placeholder for where the current swapped pool will fit into the trade.
```
p0mvn, line 267: Asking to learn: where can I read more about the trade-offs between the highest liquidity pool and the hot routes methods?

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+
+## Execution Guardrails
+
+`x/protorev` is bounded and limited in the number of trades the module can execute per block. The purpose of doing so is to ensure that the current block time does not substantially change and that the module does not introduce a new attack vector. 
+
+Execution is currently limited in the following ways
+
+1. The binary search method for finding input amounts is bounded by some number of iterations.
+2. The number of routes that can be traversed in a given transaction is bounded by some number.
+3. The number of routes that can be traversed in a given block is bounded by some number.
```
p0mvn, line 335: Asking to learn: how are these numbers decided on?

p0mvn (approved):
Incredible work on the documentation!

I'm happy to merge this once @stackman27 gives a second ACK.

Some of my comments might require changes, for example, including more fields in genesis. However, please address these comments in a separate PR. This is to merge this one and unblock testing that @niccoloraspa is planning on doing

stackman27 (approved):
LGTM just one question 

With stepSide inside route proto, do we check to see **each** route matches the stepSize swapAmount now?

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+
+### BaseDenoms
+
+BaseDenoms are the denominations that are used to build the highest liquidity routes. This will be configurable by the admin account, but will always maintain at least `uosmo` as a base denom. A base denom just means the denomination that will be used to start and end a cyclic arbitrage route. Base denoms can be added on as needed basis. 
+
+***NOTE***: BaseDenoms do have a priority that is directly tied down to the order in the list of base denoms that are used i.e. BaseDenoms that are closer to the front of the list will likely be simulated and executed more often than those later in the list. This is done by design so that we can prioritize certain denoms over others in order to simulate and execute the most profitable trades.
```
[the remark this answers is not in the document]
davidterpay: Priority for a denomination is vaguely tied down to how frequently we might see an arbitrage opportunity for that denomination. In practice, we will be conducting more research on the metrics we might want to use to determine which denominations should have priority over others.

Two of the more basic metrics might be something like hit rate - how often a denomination creates an arbitrage opportunity for every time we see it in a swap - and fill rate - how profitable each swap was by denomination.

More generally though this is an open ended problem. We wanted a way to be able to prioritize certain denominations over others in the case where we can only simulate a handful of routes. For example, lets say `osmo` ends up being the most profitable arbitrage denomination with `atom` being a distant second. In the case where we can only simulate and execute one route close to the end of the block because we have already exhausted all of the compute - pool points - in previous transcations, we would choose to simulate `osmo` because it might have a higher probability of generating a profitable route.

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+
+### NumberOfTrades
+
+This will store the total number of arbitrage trades that `x/protorev` has executed since genesis. This gets incremented every time the module executes a trade.
+
+### ProfitsByDenom
+
+This will store the profits `x/protorev` has accumulated for a given denom.
```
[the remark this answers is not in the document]
davidterpay: We figured it might be easier for the community to analyze protorev and its various metrics instead of having to have some off-chain infrastructure that can parse out this data. Added more for ease of use than anything.

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+| TokenPairArbRoutes | TokenPairRoutes tracks cyclic arb routes that can be used to create a MultiHopSwap given two denoms | []byte{1} + []byte{inputDenom} +[]byte{outputDenom} | []byte{TokenPairArbRoutes} | KV |
+| DenomPairToPool | Tracks the pool ids of the highest liquidity pools matched with a given denom[]byte{2} | []byte{2} + []byte{baseDenom} + []byte{denomToMatch} | []byte{poolID} | KV |
+| BaseDenoms | Tracks all of the base denominations that will be used to construct arbitrage routes | []byte{3} | []byte{[]BaseDenoms{}} | KV |
+| NumberOfTrades | Tracks the number of trades protorev has executed | []byte{4} | []byte{numberOfTrades} | KV |
+| ProfitsByDenom | Tracks the profits protorev has made | []byte{5} + []byte{tokenDenom} | []byte{sdk.Coin} | KV |
+| TradesByRoute | Tracks the number of trades the module has executed on a given route | []byte{6} + []byte{route} | []byte{numberOfTrades} | KV |
+| ProfitsByRoute | Tracks the profits the module has accumulated after trading on a given route | []byte{7} + []byte{route} | []byte{sdk.Coin} | KV |
+| ProtoRevEnabled | Tracks whether the protorev module is enabled | []byte{8} | []byte{bool} | KV |
```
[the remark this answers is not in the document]
davidterpay: This is something we recently rediscovered. We'll deprecate the use of the state entry in favor of the parameter in the follow up PR.

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+BaseDenoms
+
+- Osmosis
+- Atom
+
+Lets say the `postHandler` receives a transaction that contains a swap of **Juno** —> **Akash** on pool **4**. In this case, the module will attempt to create three-pool route where a base denomination is on either side of the route. For example, a route that it might create is
+
+- Osmosis —> Akash (on pool 1), Akash —> Juno (on pool 4), Juno —> Osmosis (on pool 2)
+
+It does so by finding the highest liquidity pool between (Osmosis, Akash) —> pool 1 and the highest liquidity pool between (Osmosis, Juno) —> pool 2. If there is no highest liquidity pool pair between (Osmosis, Juno) or (Osmosis, Akash), no route will be generated.
```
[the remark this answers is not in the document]
davidterpay: We only track the highest liquidity pools for certain denomination i.e. the base denominations. If the denomination has at least one pool, we will only consider the pool if it is paired with any one of the base denominations.

x/protorev/protorev.md
davidterpay: This is something we're actively researching. There isn't much online but maybe @NotJeremyLiu can follow up with some papers or insight on this.

x/protorev/protorev.md
```diff
@@ -0,0 +1,697 @@
...
+	// The weight of a balancer pool
+	BalancerWeight uint64 `protobuf:"varint,2,opt,name=balancer_weight,json=balancerWeight,proto3" json:"balancer_weight,omitempty"`
+	// The weight of a concentrated pool
+	ConcentratedWeight uint64 `protobuf:"varint,3,opt,name=concentrated_weight,json=concentratedWeight,proto3" json:"concentrated_weight,omitempty"`
+}
+```
+
+### GenesisState
+
+There is only one configurable parameter for the genesis state —> whether protorev is enabled or not.
```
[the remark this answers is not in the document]
davidterpay: We'll add this in the follow up PR. Good catch.

x/protorev/protorev.md
davidterpay: @NotJeremyLiu

davidterpay: > LGTM just one question
> 
> With stepSide inside route proto, do we check to see **each** route matches the stepSize amount now?

Yup! The step size now matches directly with each route.

davidterpay:
> 

Yup! Every step size now corresponds to a given route.

x/protorev/protorev.md
davidterpay: TLDR: We choose some number of iterations to bound by doing benchmark testing.

p0mvn:
Thanks for the replies @davidterpay and great work! 

Going to merge with the understanding that some comments are going to be addressed in a follow-up.

These include:
- extended genesis
- removing state entry for "enabled" while only keeping the param

Issue #4277, opened 2023-02-09
[CL Incentives]: Sync global uptime accumulators upon tick initialization


AlpinYukseloglu:
## Background

Once liquidity changes are properly reflected in global uptime accumulators (#4113), we then need to ensure that tick-level tracking of these global accums is implemented. The first step for this is to ensure new ticks are initialized to the correct value.

## Suggested Design

- Initialize uptime trackers of new ticks to the current value of global uptime accums
- Ensure that global accums are synced to curTime before setting values for the new tick

## Acceptance Criteria

- All new and existing tests should pass

PR #4355, opened 2023-02-17
[CL Incentives]: Sync global uptime accumulators upon tick initialization

AlpinYukseloglu:
Closes: #4277

This PR implements uptime tracker setup for tick initialization.

## Brief Changelog

- Sync global uptime accums to now prior to tick init
- Initialize uptime tracker values to current global uptime accum values

## Testing and Verifying

- New functionality is tested in `tick_test.go`
- Added new test helper to `incentives_test.go` to keep tests readable


AlpinYukseloglu:
Converted to draft as I found a much cleaner approach when working on tick crossing logic. Will open PR once I make the changes.

x/concentrated-liquidity/tick_test.go
```diff
@@ -342,28 +372,34 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
 			// System under test
-			tickInfo, err := s.App.ConcentratedLiquidityKeeper.GetTickInfo(s.Ctx, test.poolToGet, test.tickToGet)
+			tickInfo, err := clKeeper.GetTickInfo(s.Ctx, test.poolToGet, test.tickToGet)
 			if test.expectedErr != nil {
 				s.Require().Error(err)
 				s.Require().ErrorAs(err, &test.expectedErr)
 				s.Require().Equal(model.TickInfo{}, tickInfo)
 			} else {
 				s.Require().NoError(err)
+				clPool, err = clKeeper.GetPoolById(s.Ctx, validPoolId)
```
stackman27, line 402: nit: dont think this is being used anywhere
```suggestion
				
```

x/concentrated-liquidity/tick.go
```diff
@@ -94,7 +94,23 @@ func (k Keeper) getTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64) (ti
...
 		}
 
-		return model.TickInfo{LiquidityGross: sdk.ZeroDec(), LiquidityNet: sdk.ZeroDec(), FeeGrowthOutside: initialFeeGrowthOutside}, nil
+		// Sync global uptime accumulators to ensure the uptime tracker init values are up to date.
+		if err := k.updateUptimeAccumulatorsToNow(ctx, poolId); err != nil {
+			return tickStruct, err
+		}
+
+		// Initialize uptime trackers for the new tick to current global uptime accum values.
+		valuesToAdd, err := k.getInitialUptimeGrowthOutsidesForTick(ctx, poolId, tickIndex)
```
stackman27, line 103: Curious:  why did you pick `updateUptimeAccumulatorsToNow` over `getUptimeAccumulatorValues`? could there be a case where we can get have valuesToAdd for tickIndex > currentTick?

x/concentrated-liquidity/tick.go
```diff
@@ -94,7 +94,23 @@ func (k Keeper) getTickInfo(ctx sdk.Context, poolId uint64, tickIndex int64) (ti
 			return tickStruct, err
 		}
 
-		return model.TickInfo{LiquidityGross: sdk.ZeroDec(), LiquidityNet: sdk.ZeroDec(), FeeGrowthOutside: initialFeeGrowthOutside}, nil
+		// Sync global uptime accumulators to ensure the uptime tracker init values are up to date.
+		if err := k.updateUptimeAccumulatorsToNow(ctx, poolId); err != nil {
```
mattverse, line 98: I think we should think about when we want to run `updateUptimeAccumulatorsToNow`, doing this every time we call `getTickInfo` definitely feels like an overhead. Suggested way is creating a secondary method `getTickInfoAndUpdateAccumulators` and use this method for the necessary places. If this is not possible, we should at lest update this method name since its not a `get` method any more

AlpinYukseloglu: This code only executes when the tick is being initialized (note that it's under `if !found`), not on every call

czarcas7ic: Unresolving as I have a follow on question:

As Matt said, shouldn't this not be named "get" since it has the potential to actually initialize the tick accum? Maybe we can check if !found prior, and if it isn't then we call and `initTickInfo` which rolls into a `getTickInfo`. Otherwise we just call `getTickInfo`.

p0mvn, line 98: Why do we need to sync uptime accumulators if this is the very first initialization that I'm assuming should happen on the same block with uptime accumulator creation?

x/concentrated-liquidity/tick_test.go
```diff
@@ -298,35 +298,65 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
...
+			tickToGet:                preInitializedTickIndex,
+			preInitUptimeAccumValues: expectedUptimes.varyingTokensMultiDenom,
+			// Note that FeeGrowthOutside is not updated, but UptimeGrowthOutsides are.
+			// We expect uptime trackers to be initialized to zero since tick > active tick
+			expectedTickInfo: model.TickInfo{LiquidityGross: DefaultLiquidityAmt, LiquidityNet: DefaultLiquidityAmt.Neg(), UptimeTrackers: emptyUptimeTrackers},
+		},
+		{
+			name:                     "Get tick info on existing pool and existing tick with nonzero global uptime accums",
+			poolToGet:                validPoolId,
+			tickToGet:                preInitializedTickIndex - 3,
```
czarcas7ic, line 333: nit: maybe make a var with a name explaining that this tick is <= active tick, since -3 feels random / not explained

czarcas7ic (approved):
Approving but would like to resolve the comment on the `get` func semantics

x/concentrated-liquidity/tick_test.go
```diff
@@ -298,35 +298,65 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
...
+			// Note that FeeGrowthOutside is not updated, but UptimeGrowthOutsides are.
+			// We expect uptime trackers to be initialized to global accums since tick <= active tick
+			expectedTickInfo: model.TickInfo{LiquidityGross: sdk.ZeroDec(), LiquidityNet: sdk.ZeroDec(), FeeGrowthOutside: sdk.NewDecCoins(oneEth), UptimeTrackers: varyingTokensAndDenoms},
+		},
+		{
+			name:                     "Get tick info for active tick on existing pool with existing tick",
+			poolToGet:                validPoolId,
+			tickToGet:                DefaultCurrTick.Int64(),
+			preInitUptimeAccumValues: expectedUptimes.varyingTokensMultiDenom,
+			// Uptime trackers are set to global since tickToGet <= current tick
```
p0mvn, line 344:
```suggestion
			// both fee growth and uptime trackers are set to global since tickToGet <= current tick
```

x/concentrated-liquidity/tick_test.go
```diff
@@ -298,35 +298,65 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
...
+			// Note that FeeGrowthOutside is not updated, but UptimeGrowthOutsides are.
+			// We expect uptime trackers to be initialized to zero since tick > active tick
+			expectedTickInfo: model.TickInfo{LiquidityGross: DefaultLiquidityAmt, LiquidityNet: DefaultLiquidityAmt.Neg(), UptimeTrackers: emptyUptimeTrackers},
+		},
+		{
+			name:                     "Get tick info on existing pool and existing tick with nonzero global uptime accums",
+			poolToGet:                validPoolId,
+			tickToGet:                preInitializedTickIndex - 3,
+			preInitUptimeAccumValues: expectedUptimes.varyingTokensMultiDenom,
+			// Note that FeeGrowthOutside is not updated, but UptimeGrowthOutsides are.
```
p0mvn, line 335: aren't they both updated?

p0mvn: They should be updated under the same conditions as far as I can tell

x/concentrated-liquidity/tick_test.go
```diff
@@ -298,35 +298,65 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
...
+			expectedTickInfo: model.TickInfo{LiquidityGross: DefaultLiquidityAmt, LiquidityNet: DefaultLiquidityAmt.Neg(), UptimeTrackers: emptyUptimeTrackers},
+		},
+		{
+			name:                     "Get tick info on existing pool and existing tick with init but zero global uptime accums",
+			poolToGet:                validPoolId,
+			tickToGet:                preInitializedTickIndex,
+			preInitUptimeAccumValues: expectedUptimes.varyingTokensMultiDenom,
+			// Note that FeeGrowthOutside is not updated, but UptimeGrowthOutsides are.
+			// We expect uptime trackers to be initialized to zero since tick > active tick
+			expectedTickInfo: model.TickInfo{LiquidityGross: DefaultLiquidityAmt, LiquidityNet: DefaultLiquidityAmt.Neg(), UptimeTrackers: emptyUptimeTrackers},
```
p0mvn, line 328: What do you mean by updated?

My definition of updated is when we set them to non-empty coins.

According to my interpretation, both are not updated

p0mvn (approved):

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: I agree with your interpretation, the comment was incorrect: https://github.com/osmosis-labs/osmosis/pull/4355#discussion_r1116134261

Issue #4348, opened 2023-02-16
Initialize Async ICQ whitelist


nicolaslara:
## Background

The Async ICQ whitelist is controlled by a param. We should initialize it to the same as stargate whiltelist queries for cosmwasm in the upgrade.

## Suggested Design

Initialize Async ICQ param

## Acceptance Criteria

All tests pass

PR #4358, opened 2023-02-17
Nicolas/async icq whitelist

nicolaslara:
Closes: #4348 

Initializes the Async ICQ whitelist to the same as the cosmwasm stargate whitelist

## Brief Changelog

 - Initialize the asyncicq whitelist

## Testing and Verifying
Added tests for the upgrade

### DIFF
--- a/app/upgrades/v15/upgrades.go
+++ b/app/upgrades/v15/upgrades.go
@@ -4,6 +4,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	icqkeeper "github.com/strangelove-ventures/async-icq/v4/keeper"
+	icqtypes "github.com/strangelove-ventures/async-icq/v4/types"
 	packetforwardtypes "github.com/strangelove-ventures/packet-forward-middleware/v4/router/types"
 
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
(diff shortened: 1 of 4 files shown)


app/upgrades/v15/upgrades.go
```diff
@@ -18,6 +20,113 @@ import (
...
+	"/cosmos.staking.v1beta1.Query/Validator",
+
+	//osmosis queries
+
+	//epochs
+	"/osmosis.epochs.v1beta1.Query/EpochInfos",
+	"/osmosis.epochs.v1beta1.Query/CurrentEpoch",
+
+	//gamm
+	"/osmosis.gamm.v1beta1.Query/NumPools",
```
p0mvn, line 65: Maybe we should not whitelist it as it is deprecated?

app/upgrades/v15/upgrades.go
```diff
@@ -18,6 +20,113 @@ import (
...
+	"/osmosis.gamm.v1beta1.Query/PoolParams",
+	"/osmosis.gamm.v1beta1.Query/TotalPoolLiquidity",
+	"/osmosis.gamm.v1beta1.Query/TotalShares",
+	"/osmosis.gamm.v1beta1.Query/CalcJoinPoolShares",
+	"/osmosis.gamm.v1beta1.Query/CalcExitPoolCoinsFromShares",
+	"/osmosis.gamm.v1beta1.Query/CalcJoinPoolNoSwapShares",
+	"/osmosis.gamm.v1beta1.Query/PoolType",
+	"/osmosis.gamm.v2.Query/SpotPrice",
+	"/osmosis.gamm.v1beta1.Query/EstimateSwapExactAmountIn",
+	"/osmosis.gamm.v1beta1.Query/EstimateSwapExactAmountOut",
```
p0mvn, line 77: Same as these

app/upgrades/v15/upgrades.go
```diff
@@ -18,6 +20,113 @@ import (
...
+	"/osmosis.poolmanager.v1beta1.Query/EstimateSwapExactAmountOut",
+
+	//txfees
+	"/osmosis.txfees.v1beta1.Query/FeeTokens",
+	"/osmosis.txfees.v1beta1.Query/DenomSpotPrice",
+	"/osmosis.txfees.v1beta1.Query/DenomPoolId",
+	"/osmosis.txfees.v1beta1.Query/BaseDenom",
+
+	//tokenfactory
+	"/osmosis.tokenfactory.v1beta1.Query/params",
```
p0mvn, line 116: I know this is from stargate whitelist but is it actually correct to be lowercase?

For example, here it's upper case: https://github.com/osmosis-labs/osmosis/blob/9f1e7ef8920afac2015f3e0aef4094fd31b8de4b/x/tokenfactory/types/query.pb.go#L386

p0mvn (approved):

app/upgrades/v15/upgrades.go
nicolaslara: I would guess it isn't. I'll test it and change it in both places if it idn't.

app/upgrades/v15/upgrades.go
nicolaslara: Yeah, I think it makes sense to remove the deprecated queries. Is it just NumPools and EstimateSwapExactAmountIn/Out. Are there any replacements for those?

app/upgrades/v15/upgrades.go
```diff
@@ -18,6 +20,113 @@ import (
 	poolmanagertypes "github.com/osmosis-labs/osmosis/v14/x/poolmanager/types"
 )
 
+// an array of 10 strings containing the numbers from 1 to 10:
+
+// Using the same queries as stargate_whitelist.go
+var whitelistedQueries = []string{
+	// cosmos-sdk queries
+
```
mattverse, line 28: Im wondering if there is a way to use the querypath in stargate whitelist directly? (since the stargatewhitelist is a map, would it be possible to directly load the keys of the map?)

mattverse (approved):
Nit question about using stargate whitelist directly, otherwise lgtm

app/upgrades/v15/upgrades.go
p0mvn: Yes, the same queries but in the `x/poolmanager` module

app/upgrades/v15/upgrades.go
nicolaslara: yeah, I wanted to to make it dynamic first, but then got concerned that there might be issues later if the Map changes (and someone wants to recreate the chain by running the upgrade)

app/upgrades/v15/upgrades.go
p0mvn: I think it does matter as there is a pure Go string comparison:
https://github.com/strangelove-ventures/async-icq/blob/6f0b046511d24936943d44c890b9e392dcd32d50/types/keys.go#L29

As discussed offline, going to make a PR to this branch with the change

p0mvn:
Merging. Going to make casing and poolmanger changes in a follow-up

Issue #4342, opened 2023-02-16
fix(e2e) fix TestArithmeticTWAP


p0mvn:
## Background

Arithmetic TWAP test started failing for unknown reasons in https://github.com/osmosis-labs/osmosis/pull/4317

We disabled it.

## Suggested Design

Investigate the source of the problem and disable

## Acceptance Criteria
- unskip TestArithmeticTWAP
- test passes

PR #4359, opened 2023-02-17
fix(x/twap): incorrect time delta due to nanoseconds in time

p0mvn:
Closes: #4342

This PR fixes twap jitter that is caused by nanoseconds being accounted for in the time delta calculations.

In twap, we assume the time delta in milliseconds. However, the block time is given in nanoseconds. As a result, there were some cases such as:
```go
timeDelta := endRecord.Time.Sub(startRecord.Time)
```
where if `startRecord.Time` contained just a few nanoseconds, it would cause the timeDelta to fall by 1 ms below the expected value, causing a larger denominator in arithmetic mean computations. This would lead the final twap being off in the 10^-4-10^-2 digit.

To fix this, a function `types.CanonicalTimeMs` is introduced that converts the time stored in state into unix milliseconds with monotonic clock stripped away. This is consistent with how [CometBFT handles time](https://github.com/osmosis-labs/tendermint/blob/167fa738a379147f333b86f539bc03f597b978e5/libs/time/time.go#L17).

## Brief Changelog

- observed random failures in e2e twap tests in #4244 
- debugged with various logs added and manual Python recomputation of every step
- observed problems related to nanoseconds when computing time delta
- reproduced in unit tests
- refactored to strip away nanoseconds with the use of  `types.CanonicalTimeMs` function

## Testing and Verifying

This change added tests and is covered by e2e

### DIFF
--- a/x/twap/logic.go
+++ b/x/twap/logic.go
@@ -183,18 +183,15 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 	}
 	newRecord := record
 	timeDelta := types.CanonicalTimeMs(newTime) - types.CanonicalTimeMs(record.Time)
-	fmt.Println("timeDelta ", timeDelta)
 	newRecord.Time = newTime
 
 	// record.LastSpotPrice is the last spot price from the block the record was created in,
 	// thus it is treated as the effective spot price until the new time.
 	// (As there was no change until at or after this time)
 	p0NewAccum := types.SpotPriceMulDuration(record.P0LastSpotPrice, timeDelta)
-	fmt.Println("p0NewAccum ", p0NewAccum)
 	newRecord.P0ArithmeticTwapAccumulator = newRecord.P0ArithmeticTwapAccumulator.Add(p0NewAccum)
 
 	p1NewAccum := types.SpotPriceMulDuration(record.P1LastSpotPrice, timeDelta)
-	fmt.Println("p1NewAccum ", p1NewAccum)
 	newRecord.P1ArithmeticTwapAccumulator = newRecord.P1ArithmeticTwapAccumulator.Add(p1NewAccum)
 
 	// If the last spot price is zero, then the logarithm is undefined.
@@ -220,9 +217,6 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 // If for the record obtained, r.Time == r.LastErrorTime, this will also hold for the interpolated record.
 func (k Keeper) getInterpolatedRecord(ctx sdk.Context, poolId uint64, t time.Time, assetA, assetB string) (types.TwapRecord, error) {
 	record, err := k.getRecordAtOrBeforeTime(ctx, poolId, t, assetA, assetB)
-	fmt.Println("time ", t.UnixMilli())
-	fmt.Println("recordAtOrBeforeTime ", record)
-	fmt.Println("AtOrBeforeTime ", record.Time.UnixMilli())
 	if err != nil {
 		return types.TwapRecord{}, err
 	}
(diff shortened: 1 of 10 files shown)


x/twap/types/utils.go
```diff
@@ -37,16 +37,19 @@ func GetAllUniqueDenomPairs(denoms []string) []DenomPair {
 // SpotPriceMulDuration returns the spot price multiplied by the time delta,
 // that is the spot price between the current and last TWAP record.
 // A single second accounts for 1_000_000_000 when converted to int64.
-func SpotPriceMulDuration(sp sdk.Dec, timeDelta time.Duration) sdk.Dec {
+func SpotPriceMulDuration(sp sdk.Dec, timeDeltaMs int64) sdk.Dec {
+	return sp.MulInt64(timeDeltaMs)
+}
+
+func SpotPriceMulDurationOld(sp sdk.Dec, timeDelta time.Duration) sdk.Dec {
```
czarcas7ic, line 44: should we add a godoc explaining why Old is still retained / its purpose? I don't see it used anywhere but I might have missed it.

czarcas7ic (approved):
LGTM 👍

x/twap/types/utils.go
ValarDragon: unsure of why we would use this either

p0mvn: Kept it for testing and missed it during final review, going to remove now

p0mvn: removed

x/twap/types/utils.go
```diff
@@ -62,6 +60,12 @@ func LexicographicalOrderDenoms(denom0, denom1 string) (string, string, error) {
 	return denom0, denom1, nil
 }
 
+// CanonicalTimeMs returns the canonical time in milleseconds used for twap
```
mattverse, line 63:
```suggestion
// CanonicalTimeMs returns the canonical time in milliseconds used for twap
```

x/twap/types/utils.go
```diff
@@ -62,6 +60,12 @@ func LexicographicalOrderDenoms(denom0, denom1 string) (string, string, error) {
...
+// math computations in UTC. Removes any monotonic clock reading prior to conversion to ms.
+func CanonicalTimeMs(twapTime time.Time) int64 {
```
mattverse, line 65: just curious, why is it called "canonical" time?

x/twap/logic.go
```diff
@@ -182,7 +182,7 @@ func recordWithUpdatedAccumulators(record types.TwapRecord, newTime time.Time) t
 		return record
 	}
 	newRecord := record
-	timeDelta := newTime.Sub(record.Time)
+	timeDelta := types.CanonicalTimeMs(newTime) - types.CanonicalTimeMs(record.Time)
```
mattverse, line 185: can we add a one line comment above here on why we use canonical time for the time delta?

mattverse (approved):
Left some nit comments, otherwise LGTM. 

Also curious about whether this change does not cause conflicts with existing records

x/twap/types/utils.go
p0mvn: Canonical with respect to block times. It strips away any mononic part from the clock and returns the time in UTC,

This should already be done given that we use `ctx.BlockTime()`. However, this is useful for tests since we might not be setting up the time correctly in tests

x/twap/logic.go
p0mvn: Since this is called in multiple places, I elaborated in the method spec instead:
[937d148](https://github.com/osmosis-labs/osmosis/pull/4359/commits/937d148fcf078eb3e81dbd35a606104ff9eb88c4)

ValarDragon (approved):

Issue #4195, opened 2023-02-01
Add out-of-gas recovery logic to Wasm Hooks


nicolaslara:
## Background

Relayers batch IBC messages in an SDK multi-message. If one of these messages contains a wasm-hook execution (or lifecycle callback) it will execute a contract. Executing a contract means the expected gas of an IBC message execution is no longer "constant", but may vary depending on the contract's implementation. As a consequence, calling a contract with excesive gas consumption may fail the whole TX batch. 

This can lead to a temporary DoS for message transfers and puts the onus on the relayers to manage it. We should ensure that either:

 * There is a hard-cap on the gas that a wasm-hooks contract can use
 * Messages fail independently (https://github.com/cosmos/cosmos-sdk/issues/13911) 
 * Gas can be adequately predicted

or a combination of the above.

## Suggested Design

There are still open design questions here. My opinion is that improving multi-message and ensuring that relayers can deal with rogue packets is the right solution here,

## Acceptance Criteria

 - Only offending transactions should fail and others shouldn't be affected
 - There is no expense risk for relayers on high gas

nicolaslara:
@ValarDragon Added this ticket to track the out-of-gas issue. Can you add your notes from the call here when you get a chance?

nicolaslara:
This is an example from the ibc codebase on how to use a custom gas meter for the contract calls: https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go#L451

PR #4382, opened 2023-02-21
Tokenfactory: Add Before send hooks

mattverse:
Adds BeforeSend hooks to token factory.

Background: We want to enable calls to cosmwasm contracts before send of a token factory denom.

We have registered `BeforeSend` hooks to the cosmos sdk a while ago, this PR implements the BeforeSend hook to the token factory module. 

Whenever the before send hook in token factory is called, we check in state if a contract address has been registered for the  specific denom being sent. If so, the contract is called before send via hooks. 

## Brief Changelog
- Add BeforeSend Hooks to Token factory

## Testing and Verifying
This change adds tests


proto/osmosis/tokenfactory/v1beta1/tx.proto
```diff
@@ -63,6 +66,8 @@ message MsgBurn {
     (gogoproto.moretags) = "yaml:\"amount\"",
     (gogoproto.nullable) = false
   ];
+  string burnFromAddress = 3
```
nicolaslara, line 69: this is not related to the send hooks, right? Do we need these mint/burn changes?

mattverse: good catch! Srry meant to address this in a different PR must have been pushed together. Fixed

x/tokenfactory/keeper/before_send.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+
+			msgBz, err := json.Marshal(msg)
+			if err != nil {
+				return err
+			}
+
+			em := sdk.NewEventManager()
+
+			_, err = h.wasmkeeper.Sudo(ctx.WithEventManager(em), cwAddr, msgBz)
```
nicolaslara, line 107: We probably want to think about gas increase expectations for bank sends with tokenfactory denoms as the contract can consume unbounded gas

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -25,16 +34,35 @@ func TestKeeperTestSuite(t *testing.T) {
...
+
 func (suite *KeeperTestSuite) SetupTest() {
 	suite.Setup()
 	// Fund every TestAcc with two denoms, one of which is the denom creation fee
 	fundAccsAmount := sdk.NewCoins(sdk.NewCoin(types.DefaultParams().DenomCreationFee[0].Denom, types.DefaultParams().DenomCreationFee[0].Amount.MulRaw(100)), sdk.NewCoin(apptesting.SecondaryDenom, apptesting.SecondaryAmount))
 	for _, acc := range suite.TestAccs {
 		suite.FundAcc(acc, fundAccsAmount)
 	}
-
+	suite.contractKeeper = wasmkeeper.NewPermissionedKeeper(suite.App.WasmKeeper, SudoAuthorizationPolicy{})
```
nicolaslara, line 62: `NewGovPermissionKeeper` should already do what SudoAuthorizationPolicy is doing

nicolaslara (approved):
LGTM with the exception of the comments I left

x/tokenfactory/keeper/before_send.go
mattverse: Do you think its an possible option to gate usage of gas here?

x/tokenfactory/keeper/keeper_test.go
mattverse: sweet, replaced this with `NewGovPermissionKeeper`

x/tokenfactory/keeper/before_send.go
nicolaslara: yeah, we've been discussing something similar for wasm-hooks: (https://github.com/osmosis-labs/osmosis/issues/4195). For the ibc case, I think non-atomic multimsg is enough, but for send hooks I think we may want to have a hard-cap on how much gas those hooks are allowed to take (something like: https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go#L451) and have users submit with that limit. 

There is no risk to the chain here though, it's more of a UX problem. If a contract (by malice or incompetence) has an infinite loop or just very high gas usage, users may need to submit their tx with increasing gas limits

mattverse: I see I see, sounds fine for now for the UX tradeoff of gas problem, going ahead and merging this for now

PR #4383, opened 2023-02-21
Apply tendermint security patch

mattverse:
Bumps tendermint version to v0.34.25 and apply the recent security patch

## Brief Changelog
- Bump tendermint version to v0.34.25

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,10 +48,12 @@ require (
 	github.com/Djarvur/go-err113 v0.1.0 // indirect
 	github.com/alingse/asasalint v0.0.11 // indirect
 	github.com/benbjohnson/clock v1.3.0 // indirect
+	github.com/btcsuite/btcd/btcec/v2 v2.3.2 // indirect
 	github.com/cosmos/gogoproto v1.4.3 // indirect
 	github.com/cosmos/iavl v0.19.4 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/strangelove-ventures/async-icq/v4 v4.0.0-rc0
 	github.com/strangelove-ventures/packet-forward-middleware/v4 v4.0.4
 	github.com/stretchr/testify v1.8.1
-	github.com/tendermint/tendermint v0.34.24
+	github.com/tendermint/tendermint v0.34.25
```
nicolaslara, line 34: doesn't this also need to change in the replace?

```
github.com/tendermint/tendermint => github.com/informalsystems/tendermint v0.34.24 // replace with  c0.34.25
```

nicolaslara: Also noticed that CI is not passing for that tag, but it seems to be broken on most tags/branches too

mattverse: yes, thanks for pointing out! Seems that this is because of the tedermint version mismatches between our cosmos-sdk and Osmosis repo, investigating this further

mattverse:
Blocked on https://github.com/osmosis-labs/cosmos-sdk/pull/420

mattverse:
planning to reopen this PR once sdk side blocker has been solved

PR #4385, opened 2023-02-22
fix(CL): swap in given out fees and tests

p0mvn:
Closes: #XXX

The fee for the swap-in given out was still charged incorrectly. This PR fixes the fee charge as well as the tests

### DIFF
--- a/x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go
+++ b/x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go
@@ -83,7 +83,7 @@ func (s oneForZeroStrategy) ComputeSwapStep(sqrtPriceCurrent, sqrtPriceNextTick,
 	} else {
 		// amountZero is amountIn.
 		// TODO: multiplication with rounding up at precision end.
-		feeChargeTotal = amountZero.Mul(s.swapFee)
+		feeChargeTotal = amountZero.Mul(s.swapFee).Quo(sdk.OneDec().Sub(s.swapFee))
 	}
 
 	return sqrtPriceNext, amountOne, amountZero, feeChargeTotal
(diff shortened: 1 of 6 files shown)


x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go
```diff
@@ -83,7 +83,7 @@ func (s oneForZeroStrategy) ComputeSwapStep(sqrtPriceCurrent, sqrtPriceNextTick,
 	} else {
 		// amountZero is amountIn.
 		// TODO: multiplication with rounding up at precision end.
```
mattverse, line 85: Do we still need this?

x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go
```diff
@@ -83,7 +83,7 @@ func (s oneForZeroStrategy) ComputeSwapStep(sqrtPriceCurrent, sqrtPriceNextTick,
...
-		feeChargeTotal = amountZero.Mul(s.swapFee)
+		feeChargeTotal = amountZero.Mul(s.swapFee).Quo(sdk.OneDec().Sub(s.swapFee))
```
mattverse, line 86: what's the reason behind having to divide it by `1-s.swapFee` not `swapFee` itself?

p0mvn: We want to charge the fee on the amount in after fee.

By doing `amountZero / (1 - fee)` we get the amount in after fee. By multiplying by swap fee then, we get the desired amount.

For example, assume `amountZero` is 10, `swapFee` is 0.1

If we did `10 * 0.1`, we would get 0.1

However, by doing `10 * / 0.9 * 0.1`, we get `0.1111` which is the fee charged on the full amount needed

x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go
p0mvn: Yes, we do

czarcas7ic (approved):
This LGTM!

p0mvn:
Going to merge this. I'm working on compute swap step refactor so we can iterate further if any issues occur

PR #4388, opened 2023-02-22
Set proper wasm bytecode size

nicolaslara:
https://github.com/osmosis-labs/osmosis/pull/4174 changed the maximum bytecode size to 2MB. I didn't realize at the time that the max size for wasm upload via proposals was already set to 3MB, so that change actually "decreesed" the maximum allowed bytecode size for mainnet (since most contracts are uploaded via proposals) and only increased it for the testnet and allowed uploaders.

This change sets the wasm bytecode size to 3MB for both proposals and allowed wallets. 

## Brief Changelog

All stored contract codes have a maximum of 3MB now

## Testing and Verifying
all tests pass

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes /)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes /)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Misc Improvements
   * [#4131](https://github.com/osmosis-labs/osmosis/pull/4141) Add GatherValuesFromStorePrefixWithKeyParser function to osmoutils.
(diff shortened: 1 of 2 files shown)


p0mvn (approved):

ValarDragon:
Oh interesting, didn't realize they had a distinction between gov upload size and main tx upload size

nicolaslara:
me neither 😅 . Noticed it when I was looking into how small XCS needs to be to be uploaded before the upgrade. Turns out... we can just upload it now if it's via a proposal

PR #4384, opened 2023-02-22
migrate stXXX/XXX constant product pool -> stableswap pool

asalzmann:
## What is the purpose of the change

Migrate pools 833, 817, and 810 from constant product -> stable swap.

## Brief Changelog

* add upgrade handler code
* test migration function
* test basic LP functionality (swap, withdraw after migration)

## Testing and Verifying

This change added tests and can be verified as follows:

- Added a unittest `TestMigrateBalancerToStablePools` to verify liquidity stays constant before/after the migration
- Added e2e tests
    - setup in `CreatePreUpgradeState` by LPing to the 3 pools
    - add `TestMigrateBalancerToStable` e2e test which swaps in and exits each pool

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? yes
  - How is the feature or change documented? `CHANGELOG.md`

PR #4393, opened 2023-02-22
Setup e2e test for Stride pool migration

p0mvn:
Closes: #XXX

This PR sets up e2e for Stride's pool migration in #4384 

It does so by modifying genesis to create all pools up until and including the last pool id of the Stride's osmo - stOsmo pool.

It then adds a swap test to make sure that we can continue swapping after running the upgrade handler with migrations.

The test is currently disabled as we need to backport this genesis change to v14 and update the chain init image here: https://hub.docker.com/r/osmolabs/osmosis-e2e-init-chain

However, this has been successfully tested via current branch e2e test.

## Brief Changelog
- update `poolmanager` genesis to contain the mapping from pool id to pool type
   * this is needed to set up the routes correctly in e2e
- create pools in genesis up until and including the last Stride's pool
- update pool balances in genesis
- update twap records for genesis pools
- create a test that swaps against pools made from genesis and test from current branch configurer.


x/poolmanager/types/keys.go
```diff
@@ -31,9 +30,6 @@ func FormatModuleRouteKey(poolId uint64) []byte {
 // ParseModuleRouteFromBz parses the raw bytes into ModuleRoute.
 // Returns error if fails to parse or if the bytes are empty.
 func ParseModuleRouteFromBz(bz []byte) (ModuleRoute, error) {
-	if len(bz) == 0 {
```
p0mvn, line 34: Note: bz of length zero is expected because when we serialize `ModuleRoute` that has `PoolType` `Balancer` that is equal to `zero`, the serialization ends up being empty bytes.

proto/osmosis/poolmanager/v1beta1/genesis.proto
```diff
@@ -24,4 +25,6 @@ message GenesisState {
   uint64 next_pool_id = 1;
   // params is the container of poolmanager parameters.
   Params params = 2 [ (gogoproto.nullable) = false ];
+  // pool_routes is the container of the mappins from pool id to pool type.
```
stackman27, line 28: nit 
```suggestion
  // pool_routes is the container of the mappings from pool id to pool type.
```

x/poolmanager/router.go
```diff
@@ -20,6 +20,8 @@ func (k Keeper) RouteExactAmountIn(
 	tokenIn sdk.Coin,
 	tokenOutMinAmount sdk.Int,
 ) (tokenOutAmount sdk.Int, err error) {
+	ctx.Logger().Error("SWAP ARRIVED")
```
stackman27, line 23: Is this just a debugging log?

x/poolmanager/create_pool.go
```diff
@@ -141,3 +142,29 @@ func (k Keeper) GetPoolModule(ctx sdk.Context, poolId uint64) (types.SwapI, erro
...
+func (k Keeper) getAllPoolRoutes(ctx sdk.Context) []types.ModuleRoute {
+	store := ctx.KVStore(k.storeKey)
+	moduleRoutes, err := osmoutils.GatherValuesFromStorePrefixWithKeyParser(store, types.SwapModuleRouterPrefix, parsePoolRouteWithKey)
+	if err != nil {
+		panic(err)
+	}
+	return moduleRoutes
+}
+
+// / parsePoolRouteWithKey parses pool route by grabbing the pool id from key
```
stackman27, line 156:
```suggestion
// parsePoolRouteWithKey parses pool route by grabbing the pool id from key
```

tests/e2e/initialization/config.go
```diff
@@ -368,36 +413,47 @@ func updateTxfeesGenesis(txfeesGenState *txfeestypes.GenesisState) {
...
+			// test the migration all the way up to the largest pool id
+			// of StOSMO_OSMOPoolId.
+			pool = setupPool(poolId, OsmoDenom, AtomDenom)
+		}
+		gammGenState.Pools = append(gammGenState.Pools, pool)
 	}
-	gammGenState.Pools = []*types1.Any{anyPool}
-	gammGenState.NextPoolNumber = 2
+
+	// Note that we set the next pool numbe as 1 greater than the latest created pool.
```
stackman27, line 438: nit
```suggestion
	// Note that we set the next pool number as 1 greater than the latest created pool.
```

stackman27 (approved):
overall logic LGTM, left some nit comments

x/poolmanager/router.go
p0mvn: yes, removed, ty !

x/poolmanager/keeper_test.go
```diff
@@ -17,7 +17,19 @@ type KeeperTestSuite struct {
 
 const testExpectedPoolId = 3
 
-var testPoolCreationFee = sdk.Coins{sdk.NewInt64Coin(sdk.DefaultBondDenom, 1000_000_000)}
+var (
+	testPoolCreationFee = sdk.Coins{sdk.NewInt64Coin(sdk.DefaultBondDenom, 1000_000_000)}
+	testPoolRoute       = []types.ModuleRoute{
```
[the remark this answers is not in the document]
p0mvn: having CL type here doesn't really matter for test coverage that much here since it's just testing that the genesis updates correctly.

I left it out to minimize merge conflicts because this needs to be backported to v14.x that doesn't have CL type

p0mvn:
Marked this as state-breaking since modified some state entries in the new `poolmanager` module.

Also, would like to give this another round of eyes before v15

x/poolmanager/keeper_test.go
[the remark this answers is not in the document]
stackman27: yup makes sense

PR #4396, opened 2023-02-22
test(e2e): upgrade e2e init image for Stride migration

p0mvn:
Closes: #XXX

Updating previous (V14) e2e chain init Docker image to apply genesis changes from #4393 needed for #4384

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -1121,10 +1121,6 @@ func (s *IntegrationTestSuite) TestGeometricTWAP() {
 // This test is to be re-enabled for upgrade once the upgrade handler logic is added and
 // the balancer pool genesis is backported to v14.
 func (s *IntegrationTestSuite) TestStridePoolMigration() {
-	if !s.skipUpgrade {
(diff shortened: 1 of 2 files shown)


tests/e2e/e2e_test.go
```diff
@@ -1121,10 +1121,6 @@ func (s *IntegrationTestSuite) TestGeometricTWAP() {
 // This test is to be re-enabled for upgrade once the upgrade handler logic is added and
 // the balancer pool genesis is backported to v14.
 func (s *IntegrationTestSuite) TestStridePoolMigration() {
-	if !s.skipUpgrade {
```
stackman27, line 1124: Why did we remove this logic again?

stackman27 (approved):
LGTM.

tests/e2e/e2e_test.go
p0mvn: We disabled this previously because the v14 genesis was not updated. As a result, this test would fail when run with upgrade testing enabled.

Now, that we updated the v14 genesis generation via the `osmolabs/osmosis-e2e-init-chain` script, this test can be enabled with the upgrade

Issue #4278, opened 2023-02-09
[CL Incentives]: Sync global uptime accumulators upon tick crossing


AlpinYukseloglu:
## Background

Once tick initialization logic is set up for tick-level accum uptime trackers, we need to implement the necessary logic for accommodating tick crossing.

## Suggested Design

- Upon tick crossing, we should subtract the uptime tracker values of the tick we are entering from the tick we came from, as implemented here: https://github.com/osmosis-labs/osmosis/pull/4187/commits/c318fa036c22ff7cdc96aab28d0560e02009839f

## Acceptance Criteria

- All new and existing tests pass

PR #4399, opened 2023-02-23
[CL Incentives]: Sync global uptime accumulators upon tick crossing

AlpinYukseloglu:
Closes: #4278

This PR implements uptime-related logic for tick crossing. Specifically, it "flips" the tick-level uptime accum values upon crossing.

## Brief Changelog
 
- Implement logic for flipping uptime trackers when a tick is crossed

## Testing and Verifying

- New logic is relatively well tested in `tick_test.go` (including cases for technically undefined behavior such as crossing into non-adjacent ticks)


x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+			poolToGet:               validPoolId,
+			preInitializedTickIndex: preInitializedTickIndex,
+			tickToGet:               preInitializedTickIndex,
+			additiveFee:             defaultAdditiveFee,
+			// Global uptime accums remain unchanged after tick init
+			initGlobalUptimeAccumValues: expectedUptimes.twoHundredTokensMultiDenom,
+			globalUptimeAccumDelta:      expectedUptimes.hundredTokensMultiDenom,
+			// We expect new uptime trackers to be new global - init global
+			// This is because we init them to twoHundredTokensMultiDenom and then add hundredTokensMultiDenom,
+			// so when we cross the tick and "flip" it, we expect threeHundredTokensMultiDenom - twoHundredTokensMultiDenom
```
stackman27, line 441: Does "flip it" mean one side of the tick is cross so the liquidity is flipped to one token?

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+			expectedUptimeTrackers:       wrapUptimeTrackers(expectedUptimes.hundredTokensMultiDenom),
+			expectedLiquidityDelta:       DefaultLiquidityAmt.Neg(),
+			expectedTickFeeGrowthOutside: DefaultFeeAccumCoins.Add(defaultAdditiveFee),
+		},
+		{
+			name:                         "Get tick info on existing pool and existing tick below current tick (nil uptime trackers)",
```
stackman27, line 447: shouldn't this name be "get tickInfo on existing pool and existing tick above current tick(nil uptime tracker)" 

i'm going off of this logic 
```
// uptimeGrowthOutside =
// { uptimeGrowthGlobal current tick >= tick }
// { 0                  current tick <  tick }
//
```

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
 			poolToGet:                    validPoolId,
+			preInitializedTickIndex:      preInitializedTickIndex,
 			tickToGet:                    preInitializedTickIndex,
 			additiveFee:                  defaultAdditiveFee,
+			expectedUptimeTrackers:       emptyUptimeTrackers,
+			expectedLiquidityDelta:       DefaultLiquidityAmt.Neg(),
+			expectedTickFeeGrowthOutside: DefaultFeeAccumCoins.Add(defaultAdditiveFee),
+		},
+		{
+			name:                    "Get tick info on existing pool and existing tick above current tick (nonzero uptime trackers)",
```
stackman27, line 457: same thing as above comment, if existingTick > currentTick, shouldn't we automatically expect nil uptime growth outside the ticks?

x/concentrated-liquidity/tick.go
```diff
@@ -63,13 +63,31 @@ func (k Keeper) crossTick(ctx sdk.Context, poolId uint64, tickIndex int64, swapS
...
+	// Update global accums to now before uptime outside changes
+	if err := k.updateUptimeAccumulatorsToNow(ctx, poolId); err != nil {
+		return sdk.Dec{}, err
+	}
+
+	// For each supported uptime, subtract tick's uptime growth outside from the respective uptime accumulator
+	// This is functionally equivalent to "flipping" the trackers once the tick is crossed
+	updatedUptimeTrackers := tickInfo.UptimeTrackers
+	for uptimeId, uptimeAccum := range uptimeAccums {
+		updatedUptimeTrackers[uptimeId].UptimeGrowthOutside = uptimeAccum.GetValue().Sub(updatedUptimeTrackers[uptimeId].UptimeGrowthOutside)
```
stackman27, line 88: does `updatedUptimeTrackers[uptimeId].UptimeGrowthOutside` hold uptime below lowerTick and above upperTick? I'm just trying to fit `secondsPerLiquidityInside = secondsPerLiquidity - secondsPerLiquidityBelow(lowerTick) - secondsPerLiquidityAbove(upperTick).` from the paper into what we're using here

stackman27 (commented):
left few comments

x/concentrated-liquidity/tick.go
stackman27: oooh are we just calculating one side of the equation (bolded part) , as described here

For example, if tickCurrent < i, then **secondsPerLiquidityBelow(i) = secondsPerLiquidity - secondsPerLiquidityOutside(i)**; if tickCurrent >= i, then secondsPerLiquidityBelow(i) = secondsPerLiquidityOutside(i).

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: It means switching a tick's uptime growth "outside" from `growth above the tick` to `growth below the tick` (or vice versa)

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: Ah good point, naming should be other way around

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: Fixed

x/concentrated-liquidity/tick.go
AlpinYukseloglu: Yep! SecondsPerLiquidityInside functionality is implemented in this PR: https://github.com/osmosis-labs/osmosis/pull/4407

x/concentrated-liquidity/incentives_test.go
```diff
@@ -83,31 +83,34 @@ var (
...
-		twoHundredTokensMultiDenom: []sdk.DecCoins{},
-		varyingTokensSingleDenom:   []sdk.DecCoins{},
-		varyingTokensMultiDenom:    []sdk.DecCoins{},
+		emptyExpectedAccumValues:     []sdk.DecCoins{},
+		hundredTokensSingleDenom:     []sdk.DecCoins{},
+		hundredTokensMultiDenom:      []sdk.DecCoins{},
+		twoHundredTokensMultiDenom:   []sdk.DecCoins{},
+		threeHundredTokensMultiDenom: []sdk.DecCoins{},
+		varyingTokensSingleDenom:     []sdk.DecCoins{},
+		varyingTokensMultiDenom:      []sdk.DecCoins{},
```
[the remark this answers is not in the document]
AlpinYukseloglu: This is to ensure that tests don't break if/when we decide to change the supported uptimes (specifically the number of supported uptimes). Everything is regenerated at runtime to stay compatible.

Once we finalize the specific uptimes we're supporting, we can replace this function with hard-coded values.

x/concentrated-liquidity/tick.go
```diff
@@ -63,13 +63,31 @@ func (k Keeper) crossTick(ctx sdk.Context, poolId uint64, tickIndex int64, swapS
...
-	tickInfo.FeeGrowthOutside = accum.GetValue().Add(swapStateFeeGrowth).Sub(tickInfo.FeeGrowthOutside)
+	tickInfo.FeeGrowthOutside = feeAccum.GetValue().Add(swapStateFeeGrowth).Sub(tickInfo.FeeGrowthOutside)
+
+	uptimeAccums, err := k.getUptimeAccumulators(ctx, poolId)
+	if err != nil {
+		return sdk.Dec{}, err
+	}
+
+	// Update global accums to now before uptime outside changes
+	if err := k.updateUptimeAccumulatorsToNow(ctx, poolId); err != nil {
```
p0mvn, line 80: Why do we need to update them as we cross them?

p0mvn: Shouldn't we be able to update them once per block?

AlpinYukseloglu: My thinking was that the update logic for the target tick's uptime accumulators will involve the global accum values, so we need to ensure they are up to date. Are you suggesting this isn't necessary?

I do see a different issue though – we should be updating the accums prior to getting them instead of after, will fix.

p0mvn: So over a single swap, we might be crossing multiple ticks. From my understanding, we do not need to update the uptime accumulators every tick because that is idempotent within a swap. As a result, calling this update function as we cross every tick might be redundant.

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+		{
+			name:                         "Get tick info on existing pool and existing tick above current tick (nil uptime trackers)",
 			poolToGet:                    validPoolId,
+			preInitializedTickIndex:      preInitializedTickIndex,
 			tickToGet:                    preInitializedTickIndex,
 			additiveFee:                  defaultAdditiveFee,
+			expectedUptimeTrackers:       emptyUptimeTrackers,
+			expectedLiquidityDelta:       DefaultLiquidityAmt.Neg(),
+			expectedTickFeeGrowthOutside: DefaultFeeAccumCoins.Add(defaultAdditiveFee),
+		},
```
p0mvn, line 455: I know this existed before but I think we can merge these test cases into a single one because if a tick is found, we just return it without updates. The positions of the current tick relative to the active tick is irrelevant, and having two test cases only makes it more confusing

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+		{
+			name:                    "Get tick info on existing pool and existing tick below current tick (nonzero uptime trackers)",
```
p0mvn, line 457: Here, the tick did not exist so the name is misleading

p0mvn: I also think that we can remove "on existing pool" part in the names of most test cases and just have one simple test case with the name "non existent pool - error".

Having it adds a cognitive overhead for every test case but this is just a  simple top-level check that is unimportant relative to other test case details

x/concentrated-liquidity/tick.go
p0mvn: Outside of the scope of this PR but - why do we also need to call `updateUptimeAccumulatorsToNow` in `getTickInfo`?

It feels that this update needs to happen at another level of abstraction or in another location

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+		globalUptimeAccumDelta       []sdk.DecCoins
+		expectedUptimeTrackers       []model.UptimeTracker
 		additiveFee                  sdk.DecCoin
 		expectedLiquidityDelta       sdk.Dec
 		expectedTickFeeGrowthOutside sdk.DecCoins
 		expectedErr                  bool
 	}{
 		{
-			name:                         "Get tick info on existing pool and existing tick",
+			name:                    "Get tick info on existing pool and existing tick below current tick (nonzero uptime trackers)",
```
p0mvn, line 431: nit: let's remove on "existing pool" part from every test case name, please. It adds cognitive overhead while being unimportant context relative to other test case details because pool existence check happens once at the top level

x/concentrated-liquidity/tick_test.go
p0mvn, line 457: Isn't it ("zero initial uptime") trackers. Should the test case name reflect that?

x/concentrated-liquidity/tick_test.go
```diff
@@ -409,40 +409,100 @@ func (s *KeeperTestSuite) TestGetTickInfo() {
...
+			globalUptimeAccumDelta:      expectedUptimes.hundredTokensMultiDenom,
+			// We expect new uptime trackers to be new global - init global
+			// This is because we init them to twoHundredTokensMultiDenom and then add hundredTokensMultiDenom,
+			// so when we cross the tick and "flip" it, we expect threeHundredTokensMultiDenom - twoHundredTokensMultiDenom
+			expectedUptimeTrackers:       wrapUptimeTrackers(expectedUptimes.hundredTokensMultiDenom),
+			expectedLiquidityDelta:       DefaultLiquidityAmt.Neg(),
+			expectedTickFeeGrowthOutside: DefaultFeeAccumCoins.Add(defaultAdditiveFee),
+		},
+		{
+			name:                         "Get tick info on existing pool and existing tick above current tick (nil uptime trackers)",
```
p0mvn, line 447: Does it make sense also having a test case where non-zero initial uptime trackers and above current tick?

p0mvn (approved):
LGTM pending minor suggestions and the discussion on calling `updateUptimeAccumulatorsToNow` in `crossTick` and `getTickInfo`.

If that makes it easier for making progress, please feel free to merge this and move the discussion on Slack

x/concentrated-liquidity/tick.go
AlpinYukseloglu: > So over a single swap, we might be crossing multiple ticks. From my understanding, we do not need to update the uptime accumulators every tick because that is idempotent within a swap. As a result, calling this update function as we cross every tick might be redundant.

Good point – we can add a check so it only updates if `LastUpdateTime < current time`. Made an issue here (#4422) to not block this PR (and because it could break some downstream PRs)

AlpinYukseloglu: > Outside of the scope of this PR but - why do we also need to call `updateUptimeAccumulatorsToNow` in `getTickInfo`?
> 
> It feels that this update needs to happen at another level of abstraction or in another location

We update there simply because `getInitialUptimeGrowthOutsidesForTick` uses global accum values (so they need to be up to date). The level of abstraction does seem a little off, but the function does include accumulator init logic more broadly so it might be worth discussing a small refactor or rename

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: This is already implicitly covered by case right below your comment, but I just added an explicit case for clarity anyway.

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: Agreed, done

x/concentrated-liquidity/tick_test.go
AlpinYukseloglu: Hmm not sure I follow – the uptime trackers are nonzero in this test case no?

Issue #4279, opened 2023-02-09
[CL Incentives] Implement `GetUptimeGrowthInsideRange` and `GetUptimeGrowthOutsideRange` for arbitrary tick ranges


AlpinYukseloglu:
## Background

Implement `GetUptimeGrowthInsideRange` and `GetUptimeGrowthOutsideRange` functions that allow us to specify/understand how much the of the global incentives the claimer qualifies for.

## Suggested Design

- Implement `GetUptimeGrowthInsideRange` using uptime trackers (which contain uptime growth outside a tick)
- `GetUptimeGrowthOutsideRange` = global - `GetUptimeGrowthInsideRange`

## Acceptance Criteria

- All new and existing tests pass

PR #4407, opened 2023-02-23
[CL Incentives] Implement `GetUptimeGrowthInsideRange` and `GetUptimeGrowthOutsideRange` for arbitrary tick ranges

AlpinYukseloglu:
Closes: #4279 

This PR implements `GetUptimeGrowthInsideRange` and `GetUptimeGrowthOutsideRange` for CL incentives.

## Brief Changelog

- Add `SubDecCoinArray` helper to osmoutils (since we can't natively subtract _arrays_ of `DecCoins`)
- Implement `GetUptimeGrowthInside` and `GetUptimeGrowthOutside`

## Testing and Verifying

- New functions are thoroughly tested in `incentives_test.go`


osmoutils/coin_helper.go
```diff
@@ -25,3 +29,16 @@ func MinCoins(coinsA sdk.Coins, coinsB sdk.Coins) sdk.Coins {
...
+		return []sdk.DecCoins{}, fmt.Errorf("DecCoin arrays must be of equal length to be subtracted")
+	}
+
+	finalDecCoinArray := []sdk.DecCoins{}
+	for i := range decCoinsArrayA {
+		finalDecCoinArray = append(finalDecCoinArray, decCoinsArrayA[i].Sub(decCoinsArrayB[i]))
+	}
+
+	return finalDecCoinArray, nil
+}
```
czarcas7ic, line 44: Can we add a godoc? Also, do you think we should:

* sort the two coinArrays prior to subtracting them
* check that the denom at each index matches
* add tests for this helper

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,60 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+	if err != nil {
+		return []sdk.DecCoins{}, err
+	}
+
+	// Calculate uptime growth between lower and upper ticks
+	if currentTick < lowerTick {
+		// If current tick is below range, we subtract uptime growth of upper tick from that of lower tick
+		return osmoutils.SubDecCoinArrays(getUptimeTrackerValues(lowerTickInfo.UptimeTrackers), getUptimeTrackerValues(upperTickInfo.UptimeTrackers))
+	} else if currentTick < upperTick {
+		// If current tick is within range, we subtract uptime growth of lower and upper tick from global growth
```
stackman27, line 314: i think it'll be nice to mention that this logic also handles currentTick == lowerTick case

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,60 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+	}
+	upperTickInfo, err := k.getTickInfo(ctx, poolId, upperTick)
+	if err != nil {
+		return []sdk.DecCoins{}, err
+	}
+
+	// Calculate uptime growth between lower and upper ticks
+	if currentTick < lowerTick {
+		// If current tick is below range, we subtract uptime growth of upper tick from that of lower tick
+		return osmoutils.SubDecCoinArrays(getUptimeTrackerValues(lowerTickInfo.UptimeTrackers), getUptimeTrackerValues(upperTickInfo.UptimeTrackers))
```
czarcas7ic, line 312: this reads like it subtracts lower from upper but the comment says upper from lower

stackman27 (approved):
LGTM very straightforward and easy to review. just have few questions :D

osmoutils/coin_helper.go
AlpinYukseloglu: Good point, will add a godoc.

As for your other points:
> sort the two coinArrays prior to subtracting them

I don't think there's a clear dimension along which we can sort them at this level of abstraction. It would also make the function much less generally useful imo. All uptime-related calls for this function are already sorted by construction so this wouldn't make a difference on that front.

> check that the denom at each index matches

This func actually doesn't deal with denoms at all – each index of the array is a `DecCoins` so we defer to `DecCoins`'s `Sub` function which already handles all denom-related edge cases.

> add tests for this helper

I figured this was a basic enough function that's already implicitly tested where it's used, but I agree it would be good practice to add tests for this (and the other helpers in this file)

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Hmm the function does `first param - second param` which in this case is `lower - upper` (i.e. "subtract upper from lower", as the comment says) – how would you suggest we reword it?

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Agreed, would be good specify that "within range" means [lower, upper) i.e. inclusive of lower. Will add to godoc.

x/concentrated-liquidity/incentives_test.go
```diff
@@ -870,3 +874,655 @@ func (s *KeeperTestSuite) TestGetInitialUptimeGrowthOutsidesForTick() {
...
+		// current tick above range
+
+		"current tick > upper tick, nonzero uptime growth inside": {
+			poolSetup:                    true,
+			lowerTick:                    0,
+			upperTick:                    1,
+			currentTick:                  2,
+			lowerTickUptimeGrowthOutside: uptimeHelper.hundredTokensMultiDenom,
+			upperTickUptimeGrowthOutside: uptimeHelper.twoHundredTokensMultiDenom,
+			globalUptimeGrowth:           uptimeHelper.threeHundredTokensMultiDenom,
```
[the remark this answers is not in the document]
AlpinYukseloglu: It's not part of the calculation but it's important that everything functions as intended even with different global accum amounts

x/concentrated-liquidity/incentives_test.go
```diff
@@ -870,3 +874,655 @@ func (s *KeeperTestSuite) TestGetInitialUptimeGrowthOutsidesForTick() {
...
+			expectedError:              false,
+		},
+		"current tick > upper tick, zero uptime growth inside (nonempty trackers)": {
+			poolSetup:                    true,
+			lowerTick:                    0,
+			upperTick:                    1,
+			currentTick:                  2,
+			lowerTickUptimeGrowthOutside: uptimeHelper.hundredTokensMultiDenom,
+			upperTickUptimeGrowthOutside: uptimeHelper.hundredTokensMultiDenom,
+			globalUptimeGrowth:           uptimeHelper.threeHundredTokensMultiDenom,
```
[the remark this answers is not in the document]
AlpinYukseloglu: As per my other comment the global growth does not matter here but the params in this case are technically a valid state!

x/concentrated-liquidity/incentives_test.go
```diff
@@ -870,3 +874,655 @@ func (s *KeeperTestSuite) TestGetInitialUptimeGrowthOutsidesForTick() {
...
+
+		// current tick below range
+
+		"current tick < lower tick, nonzero uptime growth inside": {
+			poolSetup:                    true,
+			lowerTick:                    0,
+			upperTick:                    1,
+			currentTick:                  -1,
+			lowerTickUptimeGrowthOutside: uptimeHelper.twoHundredTokensMultiDenom,
+			upperTickUptimeGrowthOutside: uptimeHelper.hundredTokensMultiDenom,
```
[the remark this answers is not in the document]
AlpinYukseloglu: Since if curTick is below the range, both the lower and upper ticks are "facing away" from the current tick (meaning that `UptimeGrowthOutside` represents uptime growth _above_ the tick). Thus, lowerTick's `UptimeGrowthOutside` must include upperTick's `UptimeGrowthOutside` and therefore we can assert in this case that:

`lowerTick.UptimeGrowthOutside >= upperTick.UptimeGrowthOutside`

x/concentrated-liquidity/incentives.go
czarcas7ic: Ah, you are right, I misread the comment

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,62 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+		globalMinusUpper, err := osmoutils.SubDecCoinArrays(globalUptimeValues, getUptimeTrackerValues(upperTickInfo.UptimeTrackers))
+		if err != nil {
+			return []sdk.DecCoins{}, err
+		}
+
+		return osmoutils.SubDecCoinArrays(globalMinusUpper, getUptimeTrackerValues(lowerTickInfo.UptimeTrackers))
+	} else {
+		// If current tick is above range, we subtract uptime growth of lower tick from that of upper tick
+		return osmoutils.SubDecCoinArrays(getUptimeTrackerValues(upperTickInfo.UptimeTrackers), getUptimeTrackerValues(lowerTickInfo.UptimeTrackers))
+	}
```
p0mvn, line 326: This calculation model looks quite different from fees where we do the following:
1. calculate fee growth above upper tick
2. calculate fee growth below lower tick

This gets us fee growth outside.

Subtract fee growth outside from global fee growth to get fee growth inside.

From my understanding, we should be able to compute the uptime growth using a similar model.

Is it possible to use a similar strategy to reduce the complexity stemming from various strategies when a similar one could be used?

p0mvn: With fees, we also get fee growth outside first to then compute fee growth inside by subtracting it from global. With incentives, it's the opposite which introduces additional complexity from having to ponder why this is the case.

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,62 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+	if err != nil {
+		return []sdk.DecCoins{}, err
+	}
+
+	// Calculate uptime growth between lower and upper ticks
+	// Note that we regard "within range" to mean [lowerTick, upperTick),
+	// inclusive of lowerTick and exclusive of upperTick.
+	if currentTick < lowerTick {
+		// If current tick is below range, we subtract uptime growth of upper tick from that of lower tick
+		return osmoutils.SubDecCoinArrays(getUptimeTrackerValues(lowerTickInfo.UptimeTrackers), getUptimeTrackerValues(upperTickInfo.UptimeTrackers))
```
p0mvn, line 314: This branch LGTM

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,62 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+	} else if currentTick < upperTick {
+		// If current tick is within range, we subtract uptime growth of lower and upper tick from global growth
+		globalMinusUpper, err := osmoutils.SubDecCoinArrays(globalUptimeValues, getUptimeTrackerValues(upperTickInfo.UptimeTrackers))
+		if err != nil {
+			return []sdk.DecCoins{}, err
+		}
+
+		return osmoutils.SubDecCoinArrays(globalMinusUpper, getUptimeTrackerValues(lowerTickInfo.UptimeTrackers))
```
p0mvn, line 322: This branch LGTM

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,62 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
...
+	} else {
+		// If current tick is above range, we subtract uptime growth of lower tick from that of upper tick
+		return osmoutils.SubDecCoinArrays(getUptimeTrackerValues(upperTickInfo.UptimeTrackers), getUptimeTrackerValues(lowerTickInfo.UptimeTrackers))
```
p0mvn, line 325: LGTM

x/concentrated-liquidity/incentives.go
p0mvn: To be clear, I don't mind having this approach for computing growth inside and outside.

What I'm advocating for is consistency between fee growth and incentives uptime tracker growth

p0mvn (commented):
Reviewed logic. 

Would like to come to an agreement on different ways for computing growth between fees and incentives.

I think the strategy used should be as close as possible if not the same to reduce the complexity

x/concentrated-liquidity/incentives.go
```diff
@@ -270,3 +281,62 @@ func (k Keeper) GetIncentiveRecord(ctx sdk.Context, poolId uint64, denom string,
 func (k Keeper) GetAllIncentiveRecordsForPool(ctx sdk.Context, poolId uint64) ([]types.IncentiveRecord, error) {
 	return osmoutils.GatherValuesFromStorePrefixWithKeyParser(ctx.KVStore(k.storeKey), types.KeyPoolIncentiveRecords(poolId), ParseFullIncentiveRecordFromBz)
 }
+
+// GetUptimeGrowthInsideRange returns the uptime growth within the given tick range for all supported uptimes
+func (k Keeper) GetUptimeGrowthInsideRange(ctx sdk.Context, poolId uint64, lowerTick int64, upperTick int64) ([]sdk.DecCoins, error) {
```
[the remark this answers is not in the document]
AlpinYukseloglu: Mainly since we plan to move a large part of incentive logic out of the CL module, so there's potential utility for it in the future. My understanding is that it's good practice to export non-mutative getters if we expect they might have cross-module utility – do you foresee any issues with this?

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Discussed on call that this was not blocking but worth coming back to: https://github.com/osmosis-labs/osmosis/issues/4453

x/concentrated-liquidity/incentives.go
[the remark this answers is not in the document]
AlpinYukseloglu: Nice catch, fixed

p0mvn (approved):

Issue #4305, opened 2023-02-13
CL: Fixed fee levels


czarcas7ic:
## Background

We want a sanctioned list of fee levels for CL pool creation

## Suggested Design

TBD

## Acceptance Criteria

TBD

PR #4449, opened 2023-02-27
feat[CL]: add swap fee whitelist

czarcas7ic:
Closes: #4305 

The current behavior of CL swap fees is similar to other pools, allowing any swap fee from 0-100 to be used when creating a pool. However, for concentrated liquidity, we propose implementing a whitelist of swap fees that pools can be created with. This approach is intended to prevent confusion regarding the most profitable pool for swapping, which could result in fragmented liquidity and a poor user experience. By implementing a whitelist of swap fees, we can ensure a streamlined and intuitive user experience while maximizing the benefits of concentrated liquidity.

It is still an open question to determine what we actually want to launch with for authorized fee levels.

## Brief Changelog
  - `authorized_swap_fees` added to CL params
  - initial authorized swap fees are set to:
    - 0%, .01%, .03%, .05%, .3% and 1%

## Testing and Verifying

Initialize pool test was added to verify pools cannot be created with a swap fee that is not set in the params.


x/concentrated-liquidity/incentives_test.go
```diff
@@ -119,7 +119,7 @@ func getExpectedUptimes() ExpectedUptimes {
 func wrapUptimeTrackers(accumValues []sdk.DecCoins) []model.UptimeTracker {
 	wrappedUptimeTrackers := []model.UptimeTracker{}
 	for _, accumValue := range accumValues {
-		wrappedUptimeTrackers = append(wrappedUptimeTrackers, model.UptimeTracker{accumValue})
+		wrappedUptimeTrackers = append(wrappedUptimeTrackers, model.UptimeTracker{UptimeGrowthOutside: accumValue})
```
czarcas7ic, line 122: Drive by change exposed by linter

x/concentrated-liquidity/types/params.go
```diff
@@ -18,9 +20,10 @@ func ParamKeyTable() paramtypes.KeyTable {
 	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
 }
 
-func NewParams(authorizedTickSpacing []uint64) Params {
+func NewParams(authorizedTickSpacing []uint64, authorizedSwapFees []sdk.Dec) Params {
 	return Params{
 		AuthorizedTickSpacing: authorizedTickSpacing,
+		AuthorizedSwapFees:    authorizedSwapFees,
```
p0mvn, line 26: Are we planning to have these be controlled via governance? Currently, we directly hard-code them. If we are going to be set on certain values pre-launch that are going to remain constant, I don't think there is a need for SDK params. We can keep them as Go-defined constants

x/concentrated-liquidity/genesis_test.go
```diff
@@ -20,8 +20,10 @@ import (
 var (
 	testGenesisPools = []model.Pool{}
 	testGenesis      = types.GenesisState{
-		Params: types.Params{AuthorizedTickSpacing: []uint64{1, 10, 50}},
-		Pools:  []*codectypes.Any{},
+		Params: types.Params{
+			AuthorizedTickSpacing: []uint64{1, 10, 50},
+			AuthorizedSwapFees:    []sdk.Dec{sdk.MustNewDecFromStr("0.0001"), sdk.MustNewDecFromStr("0.0003"), sdk.MustNewDecFromStr("0.0005")}},
```
p0mvn, line 25: Same comment about genesis - why do we export it if the expectation is to have them be as constants?

x/concentrated-liquidity/pool.go
```diff
@@ -140,12 +146,23 @@ func (k Keeper) GetPoolFromPoolIdAndConvertToConcentrated(ctx sdk.Context, poolI
...
 		if tickSpacing == authorizedTick {
 			return true
 		}
 	}
 	return false
 }
+
+// validateSwapFee returns true if the given swap fee is one of the authorized swap fees set in the
+// params. False otherwise.
+func (k Keeper) validateSwapFee(ctx sdk.Context, params types.Params, swapFee sdk.Dec) bool {
```
p0mvn, line 161: Consider a quick direct unit test

p0mvn (approved):

x/concentrated-liquidity/types/constants.go
```diff
@@ -35,4 +35,5 @@ var (
 	// Supported uptimes preset to 1 min, 1 hr, 1D, 1W
 	SupportedUptimes      = []time.Duration{time.Minute, time.Hour, time.Hour * 24, time.Hour * 24 * 7}
 	AuthorizedTickSpacing = []uint64{1, 10, 60, 200}
+	AuthorizedSwapFees    = []sdk.Dec{sdk.ZeroDec(), sdk.MustNewDecFromStr("0.0001"), sdk.MustNewDecFromStr("0.0003"), sdk.MustNewDecFromStr("0.0005"), sdk.MustNewDecFromStr("0.003"), sdk.MustNewDecFromStr("0.01")}
```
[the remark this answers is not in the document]
czarcas7ic: 1 percent feels pretty high already, and if we really needed higher I think gov should agree and set it

x/concentrated-liquidity/genesis_test.go
czarcas7ic: @p0mvn do you think I am good to make a v16 upgrade folder so I can move the constants there?

x/concentrated-liquidity/types/params.go
czarcas7ic: Or should I just hardcode them here like mint does and then its not needed in the upgrade handler at all

https://github.com/osmosis-labs/osmosis/blob/641a9a8ef0ea4320e21f513b788a3dd91f325cb2/x/mint/types/params.go#L53-L70

x/concentrated-liquidity/pool.go
czarcas7ic: Added! https://github.com/osmosis-labs/osmosis/pull/4449/commits/c2573fbb4756aa2a3f55bb3fe427d7fad2c5868f

x/concentrated-liquidity/genesis_test.go
czarcas7ic: Ended up directly adding it to the default setting as mint does it https://github.com/osmosis-labs/osmosis/pull/4449/commits/02c67d25065d953315a3b4de4f88e99b41742865

x/concentrated-liquidity/types/params.go
p0mvn: This works!

x/concentrated-liquidity/types/constants.go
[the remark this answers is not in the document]
p0mvn: That sgtm as long as these are governance controlled params and not just constants

x/concentrated-liquidity/genesis_test.go
p0mvn: > do you think I am good to make a v16 upgrade folder so I can move the constants there?

Let's have this in the v15 upgrade handler please until we update e2e to perform an upgrade from v15 to v16.

I will port all the v16 logic from v15 handler as part of the post-release tasks.

p0mvn (approved):
Let's make sure that we set the new params in the v15 upgrade handler and have a comment please:
```
// TODO: move to v16 upgrade handler
```

p0mvn:
Correcting my earlier request with regards to upgrade handler params. Since DefaultParams() were added to DefaultGenesis() , they will be applied in the upgrade handler anyways:
https://github.com/osmosis-labs/cosmos-sdk/blob/5c0181df6e93b72e7dacde8e08d6a59d316c732f/types/module/module.go#L479

The above function is run in every upgrade handler for modules where the consensus version is 1. And it is 1 for CL:
https://github.com/osmosis-labs/osmosis/blob/c2573fbb4756aa2a3f55bb3fe427d7fad2c5868f/x/concentrated-liquidity/clmodule/module.go#L144

So no action is needed with regards to the upgrade handler parameter setting

Issue #4443, opened 2023-02-27
Add Client support for Concentrated Pool Migration 


mattverse:
## Background

https://github.com/osmosis-labs/osmosis/pull/4051 implemented governance support for pool migration of Balancer pools to Concentrated Liquidity pools. 

However, #4051 did not implement client side gov support for Pool migrations in the gamm module.

## Suggested Design

- Add Gamm module proposalHandlers for gov module.
- Add CLI support for migration proposals

PR #4451, opened 2023-02-28
CL: Add Client support for migration proposals, wire it to gov module

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/4443

Adds Client support for migration proposals and wires it to the gov module in app.go

## Brief Changelog
Add Client support for migration proposals for Concentrated Liquidity

### DIFF
--- a/x/gamm/client/cli/tx.go
+++ b/x/gamm/client/cli/tx.go
@@ -17,7 +17,10 @@ import (
 	poolmanagertypes "github.com/osmosis-labs/osmosis/v15/x/poolmanager/types"
 
 	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/tx"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	govcli "github.com/cosmos/cosmos-sdk/x/gov/client/cli"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 )
 
 func NewTxCmd() *cobra.Command {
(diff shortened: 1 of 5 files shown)


x/gamm/client/cli/tx.go
```diff
@@ -178,6 +181,115 @@ func NewStableSwapAdjustScalingFactorsCmd() *cobra.Command {
...
+	}
+
+	cmd.Flags().String(govcli.FlagTitle, "", "title of proposal")
+	cmd.Flags().String(govcli.FlagDescription, "", "description of proposal")
+	cmd.Flags().String(govcli.FlagDeposit, "", "deposit of proposal")
+	cmd.Flags().String(FlagMigrationRecords, "", "The migration records array")
+
+	return cmd
+}
+
```
czarcas7ic, line 292: Why are we not using osmocli for these?

mattverse: @czarcas7ic  Can osmocli handle specific flags and logics inside the RunE function as well?

p0mvn: It should be possible to add custom flag parsers:
https://github.com/osmosis-labs/osmosis/blob/9743eaae0925eed56ff5e7305935a54266220c1c/x/poolmanager/client/cli/tx.go#L43-L45

p0mvn (commented):
LGTM

Would like to know if osmocli can be integrated before approving

mattverse:
@czarcas7ic  @p0mvn changed to use osmocli and tried figuring out as much as I could referencing other code, but I have some questions:
1. how do I convert `osmocli.TxCliDesc` object to cobra.Command object? The [govClient Proposal handler](https://github.com/osmosis-labs/osmosis/blob/aad013223a/x/gamm/client/proposal_handler.go#L11) takes in cobra.Command, so we would still need a way to use cobra.Command object.
2. Do flags get automatically mapped to fields? Ex) https://github.com/osmosis-labs/osmosis/blob/aad013223afcd934a7cd84fad24a1f1103a96b34/x/gamm/client/cli/flags.go#L125 Here I am setting a flag for deposit, and in the message field, we have a field named "Initial Deposit" How do I map the flag to the actual field value?
3. How do I manually set a field value for the message? In `govtypes.MsgSubmitProposal`, we have a field named `IsExpedited`, where previously we would just set the field to false and then broadcast the tx. How do I achieve the same goal using osmo cli?

x/gamm/client/cli/tx.go
mattverse: Resolving this due to issues mentioned below

x/gamm/client/cli/tx.go
```diff
@@ -178,6 +181,115 @@ func NewStableSwapAdjustScalingFactorsCmd() *cobra.Command {
...
+// NewCmdSubmitUpdateMigrationRecordsProposal implements a command handler for replace migration records proposal
```
czarcas7ic, line 239:
```suggestion
// NewCmdSubmitUpdateMigrationRecordsProposal implements a command handler for update migration records proposal
```

czarcas7ic (approved):
This looks good, can you add a link here to the issue created for changing this to osmocli before we merge?

p0mvn (approved):
LGTM

x/gamm/client/cli/tx.go
mattverse: @czarcas7ic thanks for flagging! Here is the issue that I have created to track this https://github.com/osmosis-labs/osmosis/issues/4672

Issue #4454, opened 2023-02-28
[CL Incentives] Implement `initOrUpdatePositionUptime`


AlpinYukseloglu:
## Background

Currently, when position records are added to an uptime accumulator, they checkpoint the global accumulator value. This over-distributes incentives, as each position should only be entitled to incentives that were accrued within their tick range.

## Suggested Design

- Abstract all incentives-related position update logic into a helper for readability and proper testing:
   - If position doesn't exist (i.e. is new/being initialized), checkpoint `UptimeGrowthInside`
   - If position already exists, move up accum by `UptimeGrowthOutside` then update to new `UptimeGrowthInside`

## Acceptance Criteria

- New functionality is well-tested, and all existing tests pass

PR #4455, opened 2023-02-28
[CL Incentives] Implement proper accumulator checkpointing for positions `initOrUpdatePositionUptime`

AlpinYukseloglu:
Closes: #4454

Currently, when position records are added to an uptime accumulator, they checkpoint the global accumulator value. This over-distributes incentives, as each position should only be entitled to incentives that were accrued within their tick range.

## Brief Changelog

- Abstract all incentives-related position update logic into a helper for readability and proper testing:
   - If position doesn't exist (i.e. is new/being initialized), checkpoint `UptimeGrowthInside`
   - If position already exists, move up accum by `UptimeGrowthOutside` then update to new `UptimeGrowthInside`

## Testing and Verifying

- `initOrUpdatePositionUptime` is thoroughly tested in `incentive_test.go`
- Tests in `position_test.go` is updated to reflect new logic
 

x/concentrated-liquidity/incentives.go
```diff
@@ -340,3 +340,76 @@ func (k Keeper) GetUptimeGrowthOutsideRange(ctx sdk.Context, poolId uint64, lowe
...
+
+	globalUptimeGrowthInsideRange, err := k.GetUptimeGrowthInsideRange(ctx, poolId, lowerTick, upperTick)
+	if err != nil {
+		return err
+	}
+
+	globalUptimeGrowthOutsideRange, err := k.GetUptimeGrowthOutsideRange(ctx, poolId, lowerTick, upperTick)
+	if err != nil {
+		return err
+	}
```
czarcas7ic, line 368: nit: we can call growthInside here, and growthOutside only needs to be called if the record does exist below.

x/concentrated-liquidity/lp.go
```diff
@@ -125,6 +125,7 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAd
 	}
 
 	// Check if position is still frozen
+	// TODO: consider replacing this check with ClaimIncentives and distributing rewards back into the accumulator if BlockTime < frozenUntil
```
czarcas7ic, line 128: why?

czarcas7ic (approved):
This looks good on first pass! I would like to review it again, or get another pair of eyes on this before merge.

x/concentrated-liquidity/incentives.go
```diff
@@ -340,3 +340,76 @@ func (k Keeper) GetUptimeGrowthOutsideRange(ctx sdk.Context, poolId uint64, lowe
...
+		if position.FrozenUntil.Sub(ctx.BlockTime()) >= uptime {
+			curUptimeAccum := uptimeAccumulators[uptimeIndex]
+
+			// If a record does not exist for this uptime accumulator, create a new position.
+			// Otherwise, add to existing record.
+			positionName := string(types.KeyFullPosition(poolId, owner, lowerTick, upperTick, frozenUntil))
+			recordExists, err := curUptimeAccum.HasPosition(positionName)
+			if err != nil {
+				return err
+			}
```
p0mvn, line 387: We can move this outside of the loop. This would also be aligned with suggestion here: https://github.com/osmosis-labs/osmosis/pull/4455/files#r1120998514 to reduce the number of recomputations

x/concentrated-liquidity/incentives.go
```diff
@@ -340,3 +340,76 @@ func (k Keeper) GetUptimeGrowthOutsideRange(ctx sdk.Context, poolId uint64, lowe
 
 	return osmoutils.SubDecCoinArrays(globalUptimeValues, uptimeGrowthInside)
 }
+
+// initOrUpdatePositionUptime either adds or updates records for all uptime accumulators `position` qualifies for
+func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, position *model.Position, owner sdk.AccAddress, lowerTick, upperTick int64, liquidityDelta sdk.Dec, frozenUntil time.Time) error {
```
p0mvn, line 345: I know this was discussed before but can we separate init with update? Maybe in a post-completion refactor?

Currently, we are bundling two separate behaviors together, increasing the complexity and a likelihood of missing something small.

For ref, with fees, we init the position separately here: https://github.com/osmosis-labs/osmosis/blob/03f2a989a5b603ad9b34b90294bcbb7c7dae86cc/x/concentrated-liquidity/lp.go#L72-L74

and then doing the update here assuming that it is already initialized: https://github.com/osmosis-labs/osmosis/blob/03f2a989a5b603ad9b34b90294bcbb7c7dae86cc/x/concentrated-liquidity/lp.go#L225-L227

x/concentrated-liquidity/lp.go
AlpinYukseloglu: #4456

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: I believe that after #4468 is merged and positions are keyed by `joinTime` and `freezeDuration` (instead of just `frozenUntil`, we will be able to safely remove the ability to add to a frozen position, as the only case where it would be relevant becomes when someone tries to deposit to the same `freezeDuration` multiple times in the same block (see #4482).

Since this is not functionality we necessarily need to support (and actually disallowing it would serve as a great spam resistance vector), I think it will be appropriate to simplify this function to only handle initialization.

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Well, growthOutside returns the growth outside for all uptime accumulators so including it in the loop would lead to more calls right?

Also, after this change: #4482, this code won't require a call to growthOutside at all so this part will be removed anyway.

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: If you mean the position key computation then I agree – moved outside loop

stackman27 (approved):
LGTM, will refactor the functions once #4468 gets merged

Issue #4458, opened 2023-02-28
Reduce the size of ``crosschain-swaps`` and ``outpost`` contracts


expertdicer:
## Background

The current code of ``crosschain-swaps`` and ``outpost`` contracts use an import of ``swaprouter`` in cargo.toml, which causes the size of wasm bytecode of ``crosschain-swaps`` and ``outpost`` to increase heavily. Currently, the size of ``crosschain-swaps`` is more than 1mb, which is too large to store code on Osmosis testnet (osmo-test-4), makes the process of testing it impossible.

## Suggested Design

Remove the import of ``swaprouter`` in both ``crosschain-swaps`` and ``outpost`` contracts and write a new packages contains ``msg`` that all 3 contracts use.

Remove theses: 
https://github.com/osmosis-labs/osmosis/blob/ccc07c3d1831e84dbbf35fb842ff3973f08e912f/cosmwasm/contracts/crosschain-swaps/Cargo.toml#L62

https://github.com/osmosis-labs/osmosis/blob/ccc07c3d1831e84dbbf35fb842ff3973f08e912f/cosmwasm/contracts/outpost/Cargo.toml#L59

## Acceptance Criteria
- All new and existing tests pass

nicolaslara:
Thanks for reporting this and for the exploration work! Closing this in favour of [#4494](https://github.com/osmosis-labs/osmosis/issues/4494) (since that lead to a larger reduction the the size of the bytecode)

PR #4459, opened 2023-02-28
reduce wasm bytecode size

expertdicer:
Closes: #4458 

This pull request reduces the size of ``crosschain-swaps`` and ``outpost`` contracts by removing the import of ``swaprouter`` contract.

## Brief Changelog

  - Remove ``swaprouter`` router in ``crosschain-swaps`` and ``outpost`` contracts
  - Create a package name ``osmosis-swap`` contains msgs that all 3 contract uses.

## Testing and Verifying


nicolaslara:
This is a good idea. We should replace the bytecode for the x86 bytecode to get CI to pass though.

cosmwasm/optimizer_arm64.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
nicolaslara, line 1: Can we remove this file?

cosmwasm/packages/osmosis-swap/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
nicolaslara, line 1: we can also remove this one,

cosmwasm/packages/osmosis-swap/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+cosmwasm-schema =  { workspace = true }
+cosmwasm-std =  { workspace = true }
+cosmwasm-storage = { workspace = true }
+cw-storage-plus = { workspace = true }
+cw2 = { workspace = true }
+schemars = { workspace = true }
+osmosis-std = { workspace = true }
+serde = { workspace = true }
+thiserror = { workspace = true }
```
nicolaslara, line 24: This package doesn't need all these dependencies. Let's remove the ones we don't need. Off the top of my head: storage(-plus), thiserror, cw2, schemars, I think

cosmwasm/packages/osmosis-swap/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dev-dependencies]
+osmosis-testing = { workspace = true }
```
nicolaslara, line 27: Also no need for this since there aren't any tests.

cosmwasm/packages/osmosis-swap/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod swaprouter;
```
nicolaslara, line 1: Adding the comment here cause I can't comment on swaprouter.rs because it is a move.

I think we shouldn't move all of the contents of swaprouter::msg here. For example, the instantiate msg and queries and query responses aren't used by cross chain swaps or outpost, so we can leave those in `swaprouter`

nicolaslara (changes requested):
Left a few comments with change requests, but generally looks good. Thanks for this!

cosmwasm/optimizer_arm64.sh
expertdicer: Sure

cosmwasm/packages/osmosis-swap/src/lib.rs
expertdicer: I think the best practice here is to leave all swaprouter::msg in one place, so that we don't have to look at different places.

nicolaslara:
Thanks for the changes! I see that CI is still not passing. How are you compiling the contracts? To ensure it passes it should be: https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/contracts.yml#L40-L43

cosmwasm/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(not(feature = "imported"))]
+#[cfg(not(feature = "library"))]
```
nicolaslara, line 1: I noticed you changed these. In the Cargo.toml files we're not defining the "library" feature but defining it as "imported" instead (https://github.com/osmosis-labs/osmosis/blob/main/cosmwasm/contracts/crosschain-swaps/Cargo.toml#L29). This is because many things are already called "library" and it can lead to confiusion.

This is not standard at the moment, but I have an issue to change it: https://github.com/CosmWasm/cw-template/issues/140

expertdicer: Oh. I see

expertdicer:
> Thanks for the changes! I see that CI is still not passing. How are you compiling the contracts? To ensure it passes it should be: https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/contracts.yml#L40-L43

I'm running on MacOs so I will need to use ``workspace-optimizer-arm64`` instead, which does not have 0.12.10 tag release. It only has 0.12.8 and 0.12.11.

ValarDragon:
Nice, thanks for making this!

Can we sanity check that binary size went down from this though? Its surprising to me that the Rust compiler wouldn't optimize this, if we're importing all of this from a subpackage

expertdicer:
> Nice, thanks for making this!
> 
> Can we sanity check that binary size went down from this though? Its surprising to me that the Rust compiler wouldn't optimize this, if we're importing all of this from a subpackage

It actually just decrease 15 bytes. Guess I will have to look for another way to reduce it.

ValarDragon:
Up to @nicolaslara if this code structure is preferrable, first pass doesn't really seem like it is to me.

For some other code size reductions, perhaps in another PR try removing the `regex` with some more basic string checking?

expertdicer:
Yeah. I will give it a try.

PR #4461, opened 2023-02-28
added RL sliding windows and osmo RLs

nicolaslara:

We had discussed adding sliding windows to rate limits to avoid allowing double the max at the time where the RL falls over. This wasn't included in the original code that was going to accompany the proposal, but adding it here now. 

## Brief Changelog

- Added sliding windows
- Added RLs for uosmo

## Testing and Verifying
all tests pass


app/upgrades/v15/upgrades.go
```diff
@@ -131,67 +131,94 @@ func setRateLimits(ctx sdk.Context, accountKeeper *authkeeper.AccountKeeper, rat
 	contractKeeper := wasmkeeper.NewGovPermissionKeeper(wasmKeeper)
 
 	paths := []string{
+		`{"add_path": {"channel_id": "any", "denom": "uosmo",
+          "quotas":
+            [
+              {"name":"ATOM-DAY-1","duration":86400,"send_recv":[30,30]},
+              {"name":"ATOM-DAY-2","duration":129600,"send_recv":[30,30]},
+              {"name":"ATOM-WEEK-1","duration":604800,"send_recv":[60,60]},
+              {"name":"ATOM-WEEK-2","duration":907200,"send_recv":[60,60]}
```
p0mvn, line 140: Shouldn't name be UOSMO-DAY1 etc?

app/upgrades/v15/upgrades.go
```diff
@@ -131,67 +131,94 @@ func setRateLimits(ctx sdk.Context, accountKeeper *authkeeper.AccountKeeper, rat
...
+            ]
+          }}`,
 		`{"add_path": {"channel_id": "any", "denom": "ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2",
           "quotas":
             [
-              {"name":"ATOM-DAY","duration":86400,"send_recv":[30,30]},
-              {"name":"ATOM-WEEK","duration":604800,"send_recv":[60,60]}
+              {"name":"ATOM-DAY-1","duration":86400,"send_recv":[30,30]},
+              {"name":"ATOM-DAY-2","duration":129600,"send_recv":[30,30]},
```
p0mvn, line 147: Asking to learn: can you please elaborate on the sliding window method and how this change enabled it?

p0mvn:
Also, what's the decision about adding the rate limits for ion?

p0mvn:
Please add a changelog entry for this

app/upgrades/v15/upgrades.go
nicolaslara: Yes!  copy/paste mistake

app/upgrades/v15/upgrades.go
nicolaslara: yes. If we have a limit of 30% every 24h, for example, that means that at the seams (when one period ends and the other begins) you could actually transfer 60%, 30% in the block before, and then 30% in the block after.

To avoid this, we stagger the windows by half the time, so if you tried to do the above, the second quota would prevent it.

nicolaslara:
> Also, what's the decision about adding the rate limits for ion?

I don't think we have one. I think osmo can be considered as implied by https://commonwealth.im/osmosis/discussion/8543-set-ibc-rate-limits as it was listed in all the charts, but prob not the same fore ion?

p0mvn (approved):

PR #4467, opened 2023-02-28
chore: update osmoutils, osmomath and x/ibc-hooks tags on v15.x

p0mvn:
Closes: #XXX

Updating in-repo package dependency tags for v15 release


nicolaslara (approved):
Reviewed the tags. They're in the v15.x branch and look good.

PR #4478, opened 2023-03-02
Add gas meter limit to protorev

ValarDragon:

Were using an infinite gas meter in the post handler: https://github.com/osmosis-labs/osmosis/blob/main/x/protorev/keeper/posthandler.go#L39

This PR changes it to have an upperbound, in case of an accidental uncaught infinite loop. We want it to be another defense in depth against accidental non-halting behavior. (Any infinite loop with a state read would eventually halt with this)

## Testing and Verifying


NotJeremyLiu (approved):
lgtm

PR #4501, opened 2023-03-05
feat[CL]: add claimable fees query

czarcas7ic:
Closes: #XXX

Frontend is requesting a query to check a position's claimable fees.

## Brief Changelog

- Adds a non-mutative query to check claimable fees

## Testing and Verifying

- Modifies CollectFees test to also check that the query matches what is expected


go.work.sum
```diff
@@ -1,172 +1,2037 @@
+4d63.com/gochecknoglobals v0.1.0/go.mod h1:wfdC5ZjKSPr7CybKEcgJhUOgeAQW1+7WcyK8OvUilfo=
```
p0mvn, line 1: Are all these changes needed?

proto/osmosis/concentrated-liquidity/pool-model/query.proto
```diff
@@ -136,3 +139,18 @@ message LiquidityDepth {
     (gogoproto.nullable) = false
   ];
 }
+
+// ===================== MsgQueryClaimableFees
+message QueryClaimableFeesRequest {
```
p0mvn, line 144: Looks like all CL queries are not generated correctly. The goal is to use `make run-querygen` for code-gening them.

Maybe we do this in a separate PR?

x/concentrated-liquidity/fees.go
```diff
@@ -253,6 +253,51 @@ func (k Keeper) collectFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress
 	return feesClaimed, nil
 }
 
+// queryClaimableFees queries the fee accumulator for the position given by pool id, owner, lower tick and upper tick.
+// It returns the outstanding fees that can be claimed by the owner.
+// Returns error if:
+// - pool with the given id does not exist
+// - position given by pool id, owner, lower tick and upper tick does not exist
+// - other internal database or math errors.
+func (k Keeper) queryClaimableFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, lowerTick int64, upperTick int64) (sdk.Coins, error) {
```
p0mvn, line 262: Let's have a basic test making sure that this query does not modify state please

p0mvn (approved):

x/concentrated-liquidity/fees.go
```diff
@@ -253,6 +253,51 @@ func (k Keeper) collectFees(ctx sdk.Context, poolId uint64, owner sdk.AccAddress
...
+	}
+
+	positionKey := formatFeePositionAccumulatorKey(poolId, owner, lowerTick, upperTick)
+
+	hasPosition, err := feeAccumulator.HasPosition(positionKey)
+	if err != nil {
+		return nil, err
+	}
+
+	if !hasPosition {
```
mattverse, line 276: This is an open ended question, but Im not sure if we should error here. 
For the bank balance query for example, it does not error upon querying for an account that does not exist in state, it returns empty response. But again, this is an open-ended question, please feel free to resolve this if you otherwise disagree

mattverse (approved):
+1 on @p0mvn 's comment on test for ensuring tests for state changes, otherwise lgtm

x/concentrated-liquidity/fees.go
ValarDragon: I vote for lets flag this in an issue, and start a discussion in our shared-SDK team slack channel for whats the standard we should have for clients.

go.work.sum
czarcas7ic: @p0mvn I have no idea why this is showing as a diff, its exactly the same as what is on main

x/concentrated-liquidity/fees.go
czarcas7ic: Does the test I wrote not cover this? If the query mutated state, the amount we collect after calling the query would be different than expected

proto/osmosis/concentrated-liquidity/pool-model/query.proto
czarcas7ic: Created separate issue for this https://github.com/osmosis-labs/osmosis/issues/4520

x/concentrated-liquidity/fees.go
czarcas7ic: Issue created and asked in the SDK channel https://github.com/osmosis-labs/osmosis/issues/4521

x/concentrated-liquidity/fees.go
p0mvn: What I was thinking of is having an explicit check that the position fee accumulator stays the same after the claim. I think this would be a tighter check than checking the collected values.

czarcas7ic: Sounds good, I implemented this here https://github.com/osmosis-labs/osmosis/pull/4501/commits/5aacf7b8c0c8f8613055a93c6c2737e2786274ff

I had to export GetPosition in order to compare the init values, since that is what gets modified when we claim rewards. Additionally, to work with error cases (without making it hacky), I had to differentiate with a MustGetPosition vs GetPosition. Previously, GetPosition acted as a MustGet. So I replaced all current values of GetPosition with MustGet, and utilized the regular GetPosition for tests.

p0mvn (approved):

Issue #4423, opened 2023-02-27
fix(CL Swaps): bug in swap in given out in related to zero for one logic


p0mvn:
## Background

The way we currently whether we perform swap zero for one vs one for zero in in given out is currently broken:
https://github.com/osmosis-labs/osmosis/blob/eb3ea4808ffca3169eda02284f08e893feaddcb6/x/concentrated-liquidity/swaps.go#L413

## Suggested Design

The logic above should read:
```go
zeroForOne := desiredTokenOut.Denom == asset1
```

The bug is also present here:
https://github.com/osmosis-labs/osmosis/blob/eb3ea4808ffca3169eda02284f08e893feaddcb6/x/concentrated-liquidity/swaps.go#L116

## Acceptance Criteria
- fix the bug
- update all tests

PR #4500, opened 2023-03-04
refactor/test(cl/swapstrategy): separate compute swap steps for out given in and in given out

p0mvn:
Closes: #4427

## What is the purpose of the change

This is progress towards: https://github.com/osmosis-labs/osmosis/issues/4424

Currently, our swap logic is needlessly complex because we have a single method for computing swaps out given in and in given out. In fact, merging the implementations was incorrect because the rounding behavior and swap direction are the opposite for in given out from out given in. While working on this refactor, a bug with given out was found #4423 that still needs to be fixed.

There are various edge cases between the two methods. As a result, it makes sense to split the internal calculations implementation into separate methods - `ComputeSwapStepOutGivenIn` and `ComputeSwapStepInGivenOut`.

Separating the methods allows us to improve readability, testability, and documentation.

In this PR, tests for each method were added, including the inverse tests. Additionally, method specs are customized for each compute swap step.

**Important note:** `ComputeSwapStepLegacy` is still kept and used for swap in given out to avoid breaking many swap tests that are tedious to fix. To limit the scope, these tests will be update in #4423, `ComputeSwapStepLegacy` removed and  `ComputeSwapStepInGivenOut` called in its place instead.

## Brief Changelog

- implement and test `ComputeSwapStepOutGivenIn` 
- implement and test ``ComputeSwapStepInGivenOut`
- implement inverse tests
- update documentation and method specs
- rewire swap out given in in `swaps.go` to call `ComputeSwapStepOutGivenIn`
- rename `ComputeSwapStep` to `ComputeSwapStepLegacy` and keep using it in swap in given out to avoid breaking tests

## Testing and Verifying

- Added tests for each compute swap step
- Added inverse tests
- Removed outdated tests with low utility

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? via method specs

Issue #4475, opened 2023-03-01
[CL]: SwapInverseTest errTolerance


czarcas7ic:
## Background

Previously for the SwapIn/OutAmtGivenOut/In inverse tests, we checked if the user and pool balances after the swap were EXACTLY the same:

https://github.com/osmosis-labs/osmosis/blob/03f2a989a5b603ad9b34b90294bcbb7c7dae86cc/x/concentrated-liquidity/swaps_test.go#L2079-L2083

This passed because we weren't actually doing bank sends on the expected method. In a recent PR, bank sends were added, and now this test only passes when adding a error tolerance.

https://github.com/osmosis-labs/osmosis/blob/bb91cf67af7631fcab43bf5a33955d45f16fb171/x/concentrated-liquidity/swaps_test.go#L2079-L2087

## Suggested Design

* Prior to addressing this, investigate the potential use of BigDec for swaps
  * If we use BigDecs, the diff here should be minimal if not non-existent and this can we swapped back
  * If we dont use BigDec, check if the tolerance used here is acceptable

PR #4503, opened 2023-03-06
fix/test(CL Swaps): in given out

p0mvn:
Closes: #4423

This PR is a follow-up to: https://github.com/osmosis-labs/osmosis/pull/4500

It fixes two problems with swap in given out:

1. Determining whether a swap is zero for one or one for zero

2. Using the correct swap step implementation that correctly computes next square root price by swapping the amount out in the appropriate direction 

Note that the inverse relationship test will be updated later in: https://github.com/osmosis-labs/osmosis/issues/4475

## Brief Changelog

- switch swap in given out to use the correct compute swap step logic from #4500
- remove legacy `ComputeSwapStepLegacy`
- update logic for detecting "zeroForOne" vs "oneForZero" swaps
- update all test cases

## Testing and Verifying

The tests are updated to reflect the new changes.

For each relevant test updated, there is a short Python script in comments that can be used to validate the expected numbers


mattverse (approved):
ACK on the logic changes, I did not go through the process of verifying testing vectors that have been changed. I'm thinking whether we should be looking into using Uniswap's cl testings, AFAIR, they have a simple test setup we can use to get the expected results given set of testing vectors

czarcas7ic (approved):

x/concentrated-liquidity/swaps.go
```diff
@@ -536,8 +536,8 @@ func (k Keeper) calcInAmtGivenOut(
 
 	// coin amounts require int values
 	// round amountIn up to avoid under charging
-	amt0 := swapState.amountCalculated.TruncateInt()
-	amt1 := desiredTokenOut.Amount.ToDec().Sub(swapState.amountSpecifiedRemaining).RoundInt()
+	amt0 := swapState.amountCalculated.Ceil().TruncateInt()
+	amt1 := desiredTokenOut.Amount.ToDec().Sub(swapState.amountSpecifiedRemaining).TruncateInt()
```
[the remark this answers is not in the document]
p0mvn: As a general rule, we should be truncating token out and rounding up token in. Fees should also be always rounded up

p0mvn:
Closed and reopened to re-run CI

x/concentrated-liquidity/swaps.go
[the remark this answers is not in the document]
AlpinYukseloglu: Yeah the rule of thumb should be "round in whichever direction is in the pool's favor" – not sure if this warrants a doc but we should definitely have it written down somewhere

x/concentrated-liquidity/swaps_test.go
```diff
@@ -624,36 +643,52 @@ var (
 		//          5000
 		//  4545 -----|----- 5500
 		//  4545 -----|----- 5500
-		"two positions within one tick: usdc -> eth": {
-			tokenOut:                 sdk.NewCoin("usdc", sdk.NewInt(42000000)),
+		"two positions within one tick: eth (in) -> usdc (out) zfo": {
```
AlpinYukseloglu, line 646: nit: put `zfo` and `ofz` in square brackets or somehow separate them to make it clear they're intentional (thought they were typos at first)

x/concentrated-liquidity/swaps_test.go
```diff
@@ -624,36 +643,52 @@ var (
...
 			expectedLowerTickFeeGrowth: DefaultFeeAccumCoins,
 			expectedUpperTickFeeGrowth: DefaultFeeAccumCoins,
 			// two positions with same liquidity entered
 			poolLiqAmount0: sdk.NewInt(1000000).MulRaw(2),
 			poolLiqAmount1: sdk.NewInt(5000000000).MulRaw(2),
 		},
-		"two positions within one tick: eth -> usdc": {
-			tokenOut:                 sdk.NewCoin("eth", sdk.NewInt(13370)),
+		"two positions within one tick: usdc (in) -> eth (out) ofz": {
+			tokenOut:                 sdk.NewCoin("eth", sdk.NewInt(8398)),
```
AlpinYukseloglu, line 673: Out of curiosity, what was the reasoning behind changing the `tokenOut` amount for these tests instead of keeping it the same and just changing the expected input value?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1268,10 +1716,10 @@ func (s *KeeperTestSuite) TestCalcAndSwapInAmtGivenOut() {
 				s.Require().NoError(err)
 
 				// check that tokenIn, tokenOut, tick, and sqrtPrice from CalcOut are all what we expected
+				s.Require().Equal(test.expectedSqrtPrice, sqrtPrice)
```
AlpinYukseloglu, line 1719: Out of curiosity, what prompted moving this check higher up?

AlpinYukseloglu (approved):
Logic LGTM, as do tests at a high level. Note that I did not locally recompute the test vectors to confirm expected values.

x/concentrated-liquidity/swaps_test.go
p0mvn: I tried matching them with out given in tests. If you take a look at the corresponding test cases there, you will see that they almost match with some negligible variance.

This was helpful for benchmarking against the initial stages of the refactor

x/concentrated-liquidity/swaps_test.go
p0mvn: Done

x/concentrated-liquidity/swaps_test.go
p0mvn: Sqrt price is calculated first. If it si wrong, the following values would be incorrect as well.

p0mvn:
Thanks for the reviews, everyone. As there are no more action items in this PR, I'm planning on merging it once CI passes

Issue #4482, opened 2023-03-02
[CL Incentives] Remove ability to add to frozen positions


AlpinYukseloglu:
## Background

Now that we key positions by `joinTime` and `freezeDuration`, it is no longer possible for people to add to an existing position. While we might still want to special-case non-frozen positions to save state, we should at the very least remove unnecessary code complexity from updates for uptime accumulators.

## Suggested Design

- `initOrUpdatePositionUptime` -> `initPositionUptime`
- Return error if record exists and last pool update is `Before` current blocktime
   - This is to ensure that users can create multiple txs in the same block (although not sure why we wouldn't just error here)

## Acceptance Criteria

- New and existing tests pass

stackman27:
Completed #https://github.com/osmosis-labs/osmosis/pull/4508

PR #4508, opened 2023-03-06
[CL Incentives] Remove ability to add to frozen positions

stackman27:
[Closes: #4482](https://github.com/osmosis-labs/osmosis/issues/4482)

Now that we key positions by joinTime and freezeDuration, it is no longer possible for people to add to an existing position.

## Brief Changelog
-> users cannot update their existing position, instead we create new position

## Testing and Verifying
remove unwanted tests


AlpinYukseloglu (commented):
Good work, left one comment about a logic-related issue

x/concentrated-liquidity/incentives.go
```diff
@@ -381,37 +368,14 @@ func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, posit
 		if freezeDuration >= uptime {
 			curUptimeAccum := uptimeAccumulators[uptimeIndex]
 
-			// If a record does not exist for this uptime accumulator, create a new position.
-			// Otherwise, add to existing record.
-			recordExists, err := curUptimeAccum.HasPosition(positionName)
```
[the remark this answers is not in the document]
stackman27: yup just added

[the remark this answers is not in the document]
stackman27: hrmm now that i'm thinking about this, i think it makes sense to refactor the entire position function `initOrUpdatePosition` -> `initPosition` and remove update. What do you think?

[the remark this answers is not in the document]
stackman27: upon further inspection i think returning an error if !recordExist is not a good idea rn as this will break lots of other key functions. For instance, `updatePosition` while calling `withdrawPosition` fails because the position already exist but that we should call `updatePosition` regardless  and not call `initUptime` 

i could refactor the entire flow of init  and update position, but i think thats out of scope for this PR. Will create an issue to track this, meanwhile i propose we keep the logic as is without throwing an error

x/concentrated-liquidity/position.go
```diff
@@ -68,7 +68,7 @@ func (k Keeper) initOrUpdatePosition(
 
 	// TODO: consider deleting position if liquidity becomes zero
 
-	err = k.initOrUpdatePositionUptime(ctx, poolId, position, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration)
+	err = k.initPositionUptime(ctx, poolId, position, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration)
```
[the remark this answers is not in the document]
stackman27: i dont understand how can "position be allowed to have liquidity increase for same freeze duration", if there is any change in liquidity we update JoinTime thus creating new position

x/concentrated-liquidity/incentives_test.go
```diff
@@ -1530,9 +1530,9 @@ func (s *KeeperTestSuite) TestGetUptimeGrowthOutsideRange() {
 	}
 }
 
-func (s *KeeperTestSuite) TestInitOrUpdatePositionUptime() {
+func (s *KeeperTestSuite) TestinitPositionUptime() {
 	uptimeHelper := getExpectedUptimes()
-
+	defaultJoinTime := s.Ctx.BlockTime()
```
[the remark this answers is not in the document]
stackman27: yup i can add that

x/concentrated-liquidity/incentives.go
```diff
@@ -396,22 +381,9 @@ func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, posit
 				if err != nil {
 					return err
 				}
-			} else {
```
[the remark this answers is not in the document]
stackman27: so returning error here was breaking lots of other functions such as create/withdraw position as we discussed on slack 

For example: we call `initPositionUptime` through `updatePosition` which is called on withdrawPosition. therefore `initPositionUptime` gets called on same record whenever we withdrawLiquidity which will error here but we dont want that, i'm planning on creating an issue to refactor that flow outside of this PR

x/concentrated-liquidity/incentives.go
```diff
@@ -387,31 +372,18 @@ func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, posit
...
-				// Note that even though "unclaimed rewards" accrue in the accumulator prior to freezeDuration, since position withdrawal
-				// and incentive collection are only allowed when current time is past freezeDuration these rewards are not accessible until then.
-				err = curUptimeAccum.UpdatePositionCustomAcc(positionName, liquidityDelta, globalUptimeGrowthInsideRange[uptimeIndex])
-				if err != nil {
-					return err
-				}
+			// Since the position should only be entitled to uptime growth within its range, we checkpoint globalUptimeGrowthInsideRange as
+			// its accumulator's init value. During the claiming (or, equivalently, position updating) process, we ensure that incentives are
+			// not overpaid.
+			err = curUptimeAccum.NewPositionCustomAcc(positionName, liquidityDelta, globalUptimeGrowthInsideRange[uptimeIndex], emptyOptions)
```
stackman27, line 382: Note: want to make sure liquidityDelta is the value we want to pass here. I was thinking about passing sdk.ZeroDec but i assumed we want to store the user position liquidity amt in the accumulator

x/concentrated-liquidity/incentives.go
```diff
@@ -343,16 +343,8 @@ func (k Keeper) GetUptimeGrowthOutsideRange(ctx sdk.Context, poolId uint64, lowe
...
-func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, position *model.Position, owner sdk.AccAddress, lowerTick, upperTick int64, liquidityDelta sdk.Dec, joinTime time.Time, freezeDuration time.Duration) error {
-	// We update accumulators _prior_ to any position-related updates to ensure
-	// past rewards aren't distributed to new liquidity. We also update pool's
-	// LastLiquidityUpdate here.
-	err := k.updateUptimeAccumulatorsToNow(ctx, poolId)
-	if err != nil {
-		return err
-	}
-
+// initPositionUptime either adds or updates records for all uptime accumulators `position` qualifies for.
```
czarcas7ic, line 346: Update godoc in regards to updating records

x/concentrated-liquidity/incentives.go
```diff
@@ -387,31 +372,18 @@ func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, posit
 			if err != nil {
 				return err
 			}
+			if recordExists {
```
czarcas7ic, line 375: Above this line it says

```
// If a record does not exist for this uptime accumulator, create a new position.
// Otherwise, add to existing record.
```

This is no longer the case

x/concentrated-liquidity/incentives.go
```diff
@@ -364,20 +356,13 @@ func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, posit
 		return err
 	}
 
-	globalUptimeGrowthOutsideRange, err := k.GetUptimeGrowthOutsideRange(ctx, poolId, lowerTick, upperTick)
-	if err != nil {
-		return err
-	}
-
 	// Loop through uptime accums for all supported uptimes on the pool and init or update position's records
```
czarcas7ic, line 359: Change comment

czarcas7ic (commented):
Will finish review soon, just wanted to get these nits out there while I finished a few other tasks

x/concentrated-liquidity/position_test.go
```diff
@@ -225,13 +225,10 @@ func (s *KeeperTestSuite) TestInitOrUpdatePosition() {
...
 
 				// Position-related checks
-
 				// If frozen for more than a specific uptime's period, the record should exist
 				recordExists, err := newUptimeAccums[uptimeIndex].HasPosition(positionName)
 				s.Require().NoError(err)
-				if test.param.freezeDuration >= uptime {
-					s.Require().True(recordExists)
-
+				if test.param.freezeDuration >= uptime && recordExists {
```
p0mvn, line 231: Should we remove `s.Require().False(recordExists)` below as well?

x/concentrated-liquidity/position_test.go
```diff
@@ -191,17 +189,19 @@ func (s *KeeperTestSuite) TestInitOrUpdatePosition() {
...
 			positionInfo, err = s.App.ConcentratedLiquidityKeeper.GetPosition(s.Ctx, validPoolId, s.TestAccs[0], test.param.lowerTick, test.param.upperTick, test.param.joinTime, test.param.freezeDuration)
 			s.Require().NoError(err)
-			fmt.Println(positionInfo)
-			// Check that the initialized or updated position matches our expectation
-			s.Require().Equal(test.expectedLiquidity, positionInfo.Liquidity)
+
+			if !test.positionExists {
+				// Check that the initialized or updated position matches our expectation
+				s.Require().Equal(test.expectedLiquidity, positionInfo.Liquidity)
+			}
```
p0mvn, line 204: I don't think this check should be here, it should be the same as before `s.Require().Equal(test.expectedLiquidity, positionInfo.Liquidity)` because this is a validation for when we are expecting `InitOrUpdatePosition` to be successful

x/concentrated-liquidity/lp_test.go
```diff
@@ -275,11 +275,11 @@ func (s *KeeperTestSuite) TestCreatePosition() {
 			s.Require().Equal(userBalancePrePositionCreation.Sub(sdk.NewCoins(sdk.NewCoin(ETH, asset0), (sdk.NewCoin(USDC, asset1)))).String(), userBalancePostPositionCreation.String())
 			s.Require().Equal(poolBalancePrePositionCreation.Add(sdk.NewCoin(ETH, asset0), (sdk.NewCoin(USDC, asset1))).String(), poolBalancePostPositionCreation.String())
 
-			hasPosition := clKeeper.HasFullPosition(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, defaultJoinTime, tc.freezeDuration)
+			hasPosition := clKeeper.HasFullPosition(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, DefaultJoinTime, tc.freezeDuration)
 			s.Require().True(hasPosition)
 
 			// Check position state
-			s.validatePositionUpdate(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, defaultJoinTime, tc.freezeDuration, tc.liquidityAmount)
+			s.validatePositionUpdate(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, DefaultJoinTime, tc.freezeDuration, tc.liquidityAmount)
```
p0mvn, line 282: I don't think we have assertions around uptime accumulators being updated the way we expect them to be - only on initialization. Please add tests

x/concentrated-liquidity/lp.go
```diff
@@ -75,6 +75,11 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		if err := k.initializeFeeAccumulatorPosition(cacheCtx, poolId, owner, lowerTick, upperTick); err != nil {
 			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
 		}
+
+		err = k.initPositionUptime(ctx, poolId, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration)
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, err
+		}
```
p0mvn, line 82: It seems that the semantics of position being "initial" is different between incentives and fees.

As a result, I don't think we should be adding the call to `initPositionUptime` here.

Based on the description of the issue, what I think should happen is the following:
- outside of this position length branch, check if `hasFullPosition` returns true. 
    - If yes, check if it is still frozen.
       * if yes, return error
       * if not, continue by adding to it
    - if not, call `initPositionUptime`

stackman27: I thought the point of `initPositionUptime` here was that we expect there to be no full position initialized in that specific tick range with the owner. It it does then we create a new one with different joinTime. 

In what cases would `hasFullPosition` if there is no `InitPositionUptime` initialized?

stackman27: ooh i see, its cause `getAllPositionsWithVaryingFreezeTimes` doesnot return FullPosition it just returns position based on lower, upper ticks, owner, poolId. So, we might run into a case where at that joinTime is not initialized but it is initialized at that tick range

x/concentrated-liquidity/position_test.go
stackman27: ah yes readded!

x/concentrated-liquidity/lp.go
```diff
@@ -77,6 +77,13 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		}
 	}
 
+	if !k.hasFullPosition(ctx, poolId, owner, lowerTick, upperTick, joinTime, freezeDuration) {
```
stackman27, line 80: @p0mvn not sure why we have to check if the position is frozen or not as you suggested?

x/concentrated-liquidity/lp_test.go
stackman27: added!

x/concentrated-liquidity/lp.go
stackman27: talked to @AlpinYukseloglu  about this and here's a quick tldr is 

"currently we donot support 0sec uptime therefore positions are not frozen from 0sec-60sec (min supported uptime is 1min). I think in future we are planning on adding 0sec uptime so that all positions are frozen to begin with" 

for the sake of this PR I added frozen check and made a note to remove it once we add 0sec supported uptime

AlpinYukseloglu (approved):
LGTM, good work. Left a few minor comments

x/concentrated-liquidity/position.go
```diff
@@ -150,3 +145,9 @@ func (k Keeper) CreateFullRangePosition(ctx sdk.Context, concentratedPool types.
 
 	return amount0, amount1, liquidity, joinTime, nil
 }
+
+// IsPositionStillFrozen checks if (joinTime + freezeDuration) is more than (currentBlockTime). If yes, position is still frozen.
+// Note: JoinTime is set to currentBlockTime when a user creates or updates position.
+func (k Keeper) IsPositionStillFrozen(ctx sdk.Context, joinTime time.Time, freezeDuration time.Duration) bool {
```
AlpinYukseloglu, line 151: Where is this currently used?

x/concentrated-liquidity/lp.go
```diff
@@ -77,6 +77,15 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
 	}
 
+	hasFullPosition := k.hasFullPosition(ctx, poolId, owner, lowerTick, upperTick, joinTime, freezeDuration)
+
+	if !hasFullPosition {
+		err = k.initPositionUptime(ctx, poolId, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration)
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
+		}
+	}
```
p0mvn, line 87: Why do we not error in the else case of this if it should not be possible to have the position with the same key due to `joinTime`?

stackman27: i didn't error for the <1min uptime case where one could modify position from 0sec to 60sec, but since we added 1nano second uptime, i'll add error

stackman27:
Thank you @AlpinYukseloglu  and @roman for comments. Resolved all your comments!

x/concentrated-liquidity/types/errors.go
```diff
@@ -231,6 +231,23 @@ func (e InvalidSwapFeeError) Error() string {
...
 
+type PositionStillFrozenError struct {
+	FreezeDuration time.Duration
+}
+
+func (e PositionStillFrozenError) Error() string {
+	return fmt.Sprintf("position is still under freeze duration %s", e.FreezeDuration)
+}
+
+type PositionUptimeAlreadyExistsError struct {
```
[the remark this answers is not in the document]
stackman27: yup changed

x/concentrated-liquidity/lp.go
```diff
@@ -80,6 +80,16 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
 	k.getNextPositionIdAndIncrement(ctx)
 
+	hasFullPosition := k.hasFullPosition(ctx, poolId, owner, lowerTick, upperTick, joinTime, freezeDuration)
+	if !hasFullPosition {
+		err = k.initPositionUptime(ctx, poolId, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration)
+		if err != nil {
+			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
+		}
+	} else {
+		return sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, types.PositionNotFoundError{PoolId: poolId, LowerTick: lowerTick, UpperTick: upperTick, JoinTime: joinTime, FreezeDuration: freezeDuration}
```
p0mvn, line 90: Is there a test covering this?

Also, this should not be a `PositionNotFoundError`. We should have a new one for attempting to add to an existing position

stackman27: yup changed to PositionAlreadyExist Error and is covered here: https://github.com/osmosis-labs/osmosis/pull/4508/files#diff-91569dd5b56ddbfc9450f57de47bc13061e1daceaba154cab15b635d98912d19R1716-R1773

p0mvn: The linked test seems to cover another function.

We should have a direct test for `createPosition` in `lp_test.go` for this. I think the commented-out test that was removed can be converted to cover this

stackman27: converted commented out test to cover this 
https://github.com/osmosis-labs/osmosis/pull/4508/files#diff-8a43021241fc488af68dd6cdd7c762a85476f3aa8e44248784e9265c9ef0bbdfR799

x/concentrated-liquidity/lp.go
```diff
@@ -130,7 +140,7 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
 	}
 
 	// If the position is still frozen, claim and forfeit any accrued incentives for the position.
-	isPositionFrozen := position.JoinTime.Add(position.FreezeDuration).After(ctx.BlockTime())
+	isPositionFrozen := k.IsPositionStillFrozen(ctx, position.JoinTime, position.FreezeDuration)
```
AlpinYukseloglu, line 143: Now that we're removing freezeDuration, we should be able to remove this check and just call claimAllIncentivesForPosition. It might make sense to rebase this PR to the latest CL PR that has these changes.

AlpinYukseloglu:
Merging this understanding that it will cause some minor conflicts in #4724 to not block this small change on that large PR

Issue #4401, opened 2023-02-23
[CL Incentives] Implement incentives wiring from governance-triggered emissions


AlpinYukseloglu:
## Background

For OSMO incentives, we need to ensure that governance is able to use the same pipes it currently uses to also distribute incentives to CL pools. This will require adding the necessary wiring to route OSMO to the necessary pools to be distributed as incentives.

## Suggested Design

- This is still a WIP, but one option for implementing this could be to create a new `IncentiveRecord` for all incentivized pools each week with the new amount (and therefore rate). While this approach would work, there is likely a better way to implement this.

## Acceptance Criteria

- New and existing tests pass

PR #4526, opened 2023-03-07
[CL Incentives] Implement incentives wiring from governance-triggered emissions 

stackman27:
[Closes: #4401](https://github.com/osmosis-labs/osmosis/issues/4401)

For OSMO incentives, we need to ensure that governance is able to use the same pipes it currently uses to also distribute incentives to CL pools. This will require adding the necessary wiring to route OSMO to the necessary pools to be distributed as incentives.

Design choice to simplify reviews
```
CreateCLPool -> Call AfterPoolCreation hook -> CreateGauge for CLPool
At then end of each epoch -> getPoolIdFromGaugeId -> check PoolId is CLPool -> If yes createIncentiveRecord (by calling createIncentive) -> otherwise run distribution like usual
```

## Brief Changelog
- added gov triggered any token incentives to distribute to necessary pools

## Testing and Verifying
- Added gov_test and did simple CI testing 


x/concentrated-liquidity/client/cli/tx.go
```diff
@@ -27,7 +37,7 @@ var poolIdFlagOverride = map[string]string{
 
 func NewCreateConcentratedPoolCmd() (*osmocli.TxCliDesc, *clmodel.MsgCreateConcentratedPool) {
 	return &osmocli.TxCliDesc{
-		Use:     "create-concentrated-pool [denom-0] [denom-1] [tick-spacing] [swap-fee]",
+		Use:     "create-concentrated-pool [denom-0] [denom-1] [tick-spacing] [precision-factor-At-PriceOne] [swap-fee]",
```
AlpinYukseloglu, line 40: Feels odd that this was missed if it was necessary – what triggered adding it?

stackman27: not sure why it was left out, the cli expected 5 args regardless

AlpinYukseloglu (commented):
Discussed offline that after further review it would make much more sense to tie incentives to the existing `update-pool-incentives` proposal, as in any case the approach in this PR would require us to wire up the mint module directly to CL, which would:
1. Introduce meaningful overhead and a number of unknown unknown 
2. Significantly increase the surface area of attack for CL as a module

x/concentrated-liquidity/clmodule/module.go
```diff
@@ -155,12 +154,12 @@ func (am AppModule) GenerateGenesisState(simState *module.SimulationState, s *si
 
 func (am AppModule) Actions() []simtypes.Action {
 	return []simtypes.Action{
-		simtypes.NewMsgBasedAction("CreateConcentratedPool", am.keeper, simulation.RandomMsgCreateConcentratedPool),
```
stackman27, line 158: planning on working on simulator after incentives, therefore disabling rn so that it doesnot conflict with CL hooks

x/poolmanager/keeper.go
```diff
@@ -106,3 +107,29 @@ func (k Keeper) SetNextPoolId(ctx sdk.Context, poolId uint64) {
...
+		return nil, err
+	}
+
+	pool, poolErr := swapModule.GetPool(ctx, poolId)
+	if poolErr != nil {
+		return nil, poolErr
+	}
+
+	return pool, nil
+}
```
p0mvn, line 124: There's already a `RoutePool` method that does exactly the same. Let's use that to avoid duplication

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,78 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+		if isCLPoolGauge {
+			// since we're only doing osmo incentive we can assume there to be only 1 coin therefore coins[0]
+			incentiveRecord := cltypes.IncentiveRecord{
+				PoolId:           pool.GetId(),
+				IncentiveDenom:   gauge.Coins[0].Denom,
+				IncentiveCreator: k.ak.GetModuleAddress(types.ModuleName),
+				RemainingAmount:  sdk.Dec(gauge.Coins[0].Amount),
+				// calculate amount of tokens to emit per second
+				// for ex: 10000tokens to be distributed over 1day epoch will be 1000 tokens ÷ 86,400 seconds ≈ 0.0116 tokens per second
+				EmissionRate: sdk.NewDecFromInt(gauge.Coins[0].Amount).Quo(sdk.NewDecFromInt(sdk.NewInt(int64(currEpoch.Duration.Seconds())))),
```
AlpinYukseloglu, line 395: Discussed offline to convert `sdk.NewDecFromInt(sdk.NewInt(` -> `sdk.NewDec(` and `.Quo` -> `.QuoTruncate`

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,78 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+		incParams := k.GetParams(ctx).DistrEpochIdentifier
+		currEpoch := k.ek.GetEpochInfo(ctx, incParams)
+
+		isCLPoolGauge, pool := k.IsPoolValidAndIsCLPoolGauge(ctx, gauge.Id, currEpoch.Duration)
+		// only want to run this logic if the gaugeId is associated with CL PoolId
+		if isCLPoolGauge {
+			// since we're only doing osmo incentive we can assume there to be only 1 coin therefore coins[0]
+			incentiveRecord := cltypes.IncentiveRecord{
+				PoolId:           pool.GetId(),
+				IncentiveDenom:   gauge.Coins[0].Denom,
```
AlpinYukseloglu, line 390: Discussed offline to create incentive records for each denom and handle the DoS vector in a separate PR (will create issue for scaling gas costs on adding to gauges)

x/poolmanager/create_pool.go
```diff
@@ -86,6 +86,11 @@ func (k Keeper) CreatePool(ctx sdk.Context, msg types.CreatePoolMsg) (uint64, er
 	}
 
 	emitCreatePoolEvents(ctx, poolId, msg)
+	// TODO: Remove gamm hooks and only use poolmanager hook for all types of poolType
+	// only call the hook through poolmanager if the pool created is CL
+	if msg.GetPoolType() == types.Concentrated {
+		k.hooks.AfterPoolCreated(ctx, sender, pool.GetId())
+	}
```
p0mvn, line 93: Please make an issue for this

x/pool-incentives/keeper/keeper.go
```diff
@@ -58,19 +63,24 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", "x/"+types.ModuleName)
 }
 
+// CreatePoolGauges checks if the poolId is CLPoolType, if it is create one gauge,
+// otherwise create multiple gauges based on lockableDurations.
 func (k Keeper) CreatePoolGauges(ctx sdk.Context, poolId uint64) error {
-	// Create the same number of gaugeges as there are LockableDurations
-	for _, lockableDuration := range k.GetLockableDurations(ctx) {
-		gaugeId, err := k.incentivesKeeper.CreateGauge(
+	pool, err := k.poolmanagerKeeper.RoutePool(ctx, poolId)
```
p0mvn, line 69: nit: suggest renaming this to `GetPool`, `RoutePool` does not make sense anymore in this context

x/pool-incentives/keeper/keeper.go
```diff
@@ -80,7 +90,33 @@ func (k Keeper) CreatePoolGauges(ctx sdk.Context, poolId uint64) error {
...
+				true,
+				k.accountKeeper.GetModuleAddress(types.ModuleName),
+				sdk.Coins{},
+				lockuptypes.QueryCondition{
+					LockQueryType: lockuptypes.ByDuration,
+					Denom:         gammtypes.GetPoolShareDenom(poolId),
+					Duration:      lockableDuration,
+					Timestamp:     time.Time{},
+				},
+				// QUESTION: Should we set the startTime as the epoch start time that the modules share or the current block time?
```
p0mvn, line 110: I think we can remove this question?

x/pool-incentives/keeper/keeper.go
```diff
@@ -58,19 +63,24 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+		return err
+	}
+	isCLPool := pool.GetType() == poolmanagertypes.Concentrated
+	if isCLPool {
+		gaugeIdCL, err := k.incentivesKeeper.CreateGauge(
 			ctx,
 			true,
 			k.accountKeeper.GetModuleAddress(types.ModuleName),
 			sdk.Coins{},
+			// dummy variable so that the existing logic doesnot break
```
p0mvn, line 80:
```suggestion
			// dummy variable so that the existing logic does not break
```

p0mvn, line 80: Please elaborate how it breaks in the comment and why we use these value specifically

x/pool-incentives/keeper/keeper_test.go
```diff
@@ -71,3 +71,24 @@ func (suite *KeeperTestSuite) TestCreateBalancerPoolGauges() {
...
+	keeper := suite.App.PoolIncentivesKeeper
+
+	for i := 0; i < 3; i++ {
+		clPool := suite.PrepareConcentratedPool()
+
+		incParams := suite.App.IncentivesKeeper.GetParams(suite.Ctx).DistrEpochIdentifier
+		currEpoch := suite.App.EpochsKeeper.GetEpochInfo(suite.Ctx, incParams)
+
+		// Same amount of gauges as lockableDurations must be created for every pool created.
+		gaugeId, err := keeper.GetPoolGaugeId(suite.Ctx, clPool.GetId(), currEpoch.Duration)
```
p0mvn, line 87: Please also add an assertion about the value of the gauge id

x/concentrated-liquidity/clmodule/module.go
```diff
@@ -155,12 +154,12 @@ func (am AppModule) GenerateGenesisState(simState *module.SimulationState, s *si
...
-		simtypes.NewMsgBasedAction("CreatePosition", am.keeper, simulation.RandMsgCreatePosition),
+		// simtypes.NewMsgBasedAction("CreateConcentratedPool", am.keeper, simulation.RandomMsgCreateConcentratedPool),
+		// simtypes.NewMsgBasedAction("CreatePosition", am.keeper, simulation.RandMsgCreatePosition),
 		//simtypes.NewMsgBasedAction("CLSwapExactAmountIn", am.keeper, simulation.RandomSwapExactAmountIn),
-		//simtypes.NewMsgBasedAction("CLSwapExactAmountOut", am.keeper, simulation.RandomSwapExactAmountOut),
-		simtypes.NewMsgBasedAction("WithdrawPosition", am.keeper, simulation.RandMsgWithdrawPosition),
-		simtypes.NewMsgBasedAction("CollectFees", am.keeper, simulation.RandMsgCollectFees),
-		simtypes.NewMsgBasedAction("CollectIncentives", am.keeper, simulation.RandMsgCollectIncentives),
+		// //simtypes.NewMsgBasedAction("CLSwapExactAmountOut", am.keeper, simulation.RandomSwapExactAmountOut),
```
AlpinYukseloglu, line 160:
```suggestion
		// simtypes.NewMsgBasedAction("CLSwapExactAmountOut", am.keeper, simulation.RandomSwapExactAmountOut),
```

x/pool-incentives/types/expected_keepers.go
```diff
@@ -41,3 +47,14 @@ type DistrKeeper interface {
 	SetFeePool(ctx sdk.Context, feePool distrtypes.FeePool)
 	FundCommunityPool(ctx sdk.Context, amount sdk.Coins, sender sdk.AccAddress) error
 }
+type ClKeeper interface {
```
p0mvn, line 50: nit: let's be verbose and call it as `ConcentratedLiquidityKeeper` please for clarity

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
 	}
 }
 
+func (suite *KeeperTestSuite) TestDistributeToCLPools() {
```
p0mvn, line 96:
```suggestion
func (suite *KeeperTestSuite) TestDistributeToConcentratedLiquidityPools() {
```

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+
+		"valid case: multiple gaugeId and poolId": {
+			numPools:              3,
+			expectErr:             false,
+			expectedDistributions: sdk.Coins{fifteenKRewardCoins},
+		},
+	}
+
+	for _, tc := range tests {
+		// setup test
```
p0mvn, line 124: suite.Run(...)

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+
+			// get the gaugeId corresponding to the CL pool
+			gaugeId, err := suite.App.PoolIncentivesKeeper.GetPoolGaugeId(suite.Ctx, clPool.GetId(), currEpoch.Duration)
+			suite.Require().NoError(err)
+
+			// get the gauge from the gaudeId
+			gauge, err := suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, gaugeId)
+			suite.Require().NoError(err)
+
+			gauge.Coins = sdk.Coins{sdk.NewInt64Coin(defaultRewardDenom, 5000)}
```
p0mvn, line 150: nit: please aim to use coin constructor as it does additional validation that oftentimes proves to be useful

x/incentives/keeper/distribute.go
```diff
@@ -196,26 +199,28 @@ func (d *distributionInfo) addLockRewards(owner string, rewards sdk.Coins) error
...
+		err := k.bk.SendCoinsFromModuleToManyAccounts(
+			ctx,
+			types.ModuleName,
+			distrs.idToDecodedAddr,
+			distrs.idToDistrCoins)
+		if err != nil {
+			return err
+		}
+		ctx.Logger().Debug("Finished sending, now creating liquidity add events")
+		for id := 0; id < numIDs; id++ {
```
AlpinYukseloglu, line 213: Out of curiosity, why would we emit an event for each ID?

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+					IncentiveCreator: k.ak.GetModuleAddress(types.ModuleName),
+					RemainingAmount:  sdk.Dec(coin.Amount),
+					// calculate amount of tokens to emit per second
+					// for ex: 10000tokens to be distributed over 1day epoch will be 1000 tokens ÷ 86,400 seconds ≈ 0.01157 tokens per second (truncated)
+					EmissionRate: sdk.NewDecFromInt(coin.Amount).QuoTruncate(sdk.NewDec(int64(currEpoch.Duration.Seconds()))),
+					StartTime:    gauge.GetStartTime(),
+					MinUptime:    currEpoch.Duration,
+				}
+
+				coinsToDistribute, err := k.distributeCLIncentiveInternal(ctx, incentiveRecord, gauge)
```
p0mvn, line 401: nit: let's be verbose and call this `distributeConcentratedLiquidityInternal`. While it makes sense to us working on the feature right now, anyone onboarding on the codebase in the future will likely have no idea what CL stands for

x/incentives/keeper/distribute.go
```diff
@@ -258,6 +263,32 @@ func (k Keeper) distributeSyntheticInternal(
...
 }
 
+// distributeCLIncentiveInternal runs the distribution logic for CL pools only. It creates new incentive record with osmo incentives
+// and distributes all the tokens to the dedicated pool
+func (k Keeper) distributeCLIncentiveInternal(ctx sdk.Context, incentiveRecord cltypes.IncentiveRecord, gauge types.Gauge) (sdk.Coins, error) {
+	incentiveRecord, err := k.clk.CreateIncentive(ctx,
+		incentiveRecord.PoolId,
+		incentiveRecord.IncentiveCreator,
+		incentiveRecord.IncentiveDenom,
+		sdk.Int(incentiveRecord.RemainingAmount),
```
AlpinYukseloglu, line 273:
```suggestion
		incentiveRecord.RemainingAmount.TruncateInt(),
```

x/incentives/keeper/distribute.go
```diff
@@ -258,6 +263,32 @@ func (k Keeper) distributeSyntheticInternal(
...
+		incentiveRecord.EmissionRate,
+		incentiveRecord.StartTime,
+		incentiveRecord.MinUptime,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	incentiveCoins := sdk.NewCoins(sdk.Coin{Denom: appparams.BaseCoinUnit, Amount: sdk.Int(incentiveRecord.RemainingAmount)})
```
AlpinYukseloglu, line 282:
```suggestion
	incentiveCoins := sdk.NewCoins(sdk.NewCoin(Denom: appparams.BaseCoinUnit, Amount: incentiveRecord.RemainingAmount.TruncateInt()))
```

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+			for _, coin := range gauge.Coins {
+				// since we're only doing osmo incentive we can assume there to be only 1 coin therefore coins[0]
+				incentiveRecord := cltypes.IncentiveRecord{
+					PoolId:           pool.GetId(),
+					IncentiveDenom:   coin.Denom,
+					IncentiveCreator: k.ak.GetModuleAddress(types.ModuleName),
+					RemainingAmount:  sdk.Dec(coin.Amount),
+					// calculate amount of tokens to emit per second
+					// for ex: 10000tokens to be distributed over 1day epoch will be 1000 tokens ÷ 86,400 seconds ≈ 0.01157 tokens per second (truncated)
+					EmissionRate: sdk.NewDecFromInt(coin.Amount).QuoTruncate(sdk.NewDec(int64(currEpoch.Duration.Seconds()))),
```
p0mvn, line 396: I'm a bit concerned on returing `float64` here. While casting `flaot64` to `int64` is deterministic, using floats is not in general. So I would prefer if we avoided using any method that returns floats

Can we instead return milliseconds and do a manual conversion to seconds?

x/incentives/keeper/distribute.go
p0mvn: Also, unsure what "internal" suffix means - consider dropping it

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
-		// send based on synthetic lockup coins if it's distributing to synthetic lockups
-		var gaugeDistributedCoins sdk.Coins
-		var err error
-		if lockuptypes.IsSyntheticDenom(gauge.DistributeTo.Denom) {
-			gaugeDistributedCoins, err = k.distributeSyntheticInternal(ctx, gauge, filteredLocks, &distrInfo)
+		// get pool Id corresponding to the gaugeId
+		incParams := k.GetParams(ctx).DistrEpochIdentifier
+		currEpoch := k.ek.GetEpochInfo(ctx, incParams)
+
+		isCLPoolGauge, pool := k.IsValidClGauge(ctx, gauge.Id, currEpoch.Duration)
```
p0mvn, line 384: nit: `IsValidConcentratedLiquidityGauge`

x/incentives/keeper/distribute.go
p0mvn, line 282:
```suggestion
	incentiveCoins := sdk.NewCoins(sdk.NewCoin(Denom: appparams.BaseCoinUnit, Amount: sdk.Int(incentiveRecord.RemainingAmount)))
```

p0mvn (commented):
Great work. Only nits so far

I would like to do another pass before approving

x/pool-incentives/keeper/keeper.go
```diff
@@ -58,19 +63,24 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 			lockuptypes.QueryCondition{
-				LockQueryType: lockuptypes.ByDuration,
-				Denom:         gammtypes.GetPoolShareDenom(poolId),
-				Duration:      lockableDuration,
-				Timestamp:     time.Time{},
+				LockQueryType: lockuptypes.ByTime,
```
AlpinYukseloglu, line 82: Would it be possible to leave `QueryCondition` fully blank? e.g. `lockuptypes.QueryCondition{}`

x/pool-incentives/keeper/keeper.go
AlpinYukseloglu, line 80:
```suggestion
			// dummy variable so that the existing logic does not break
```

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+			for _, gauge := range gauges {
+				incParams := suite.App.IncentivesKeeper.GetParams(suite.Ctx).DistrEpochIdentifier
+				currEpoch := suite.App.EpochsKeeper.GetEpochInfo(suite.Ctx, incParams)
+
+				// get poolId from GaugeId
+				poolId, err := suite.App.PoolIncentivesKeeper.GetPoolIdFromGaugeId(suite.Ctx, gauge.GetId(), currEpoch.Duration)
+				suite.Require().NoError(err)
+
+				// GetIncentiveRecord to see if pools recieved incentives properly
+				incentiveRecord, err := suite.App.ConcentratedLiquidityKeeper.GetIncentiveRecord(suite.Ctx, poolId, defaultRewardDenom, currEpoch.Duration, suite.App.AccountKeeper.GetModuleAddress(types.ModuleName))
```
AlpinYukseloglu, line 176: We should check to ensure the full record is correct (i.e. check all fields, not just remaining incentives)

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+
+			gauge.Coins = sdk.Coins{sdk.NewInt64Coin(defaultRewardDenom, 5000)}
+
+			gauges = append(gauges, *gauge)
+		}
+
+		// Distribute tokens from the gauge
+		_, err := suite.App.IncentivesKeeper.Distribute(suite.Ctx, gauges)
+		if tc.expectErr {
+			suite.Require().Error(err)
```
AlpinYukseloglu, line 158: We should check that a record wasn't created and that assets weren't transferred to the pool balance here.

Also, we can return in this if statement and take the rest of the test out of the `else` block for better readability

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+	fifteenKRewardCoins := sdk.NewInt64Coin(defaultRewardDenom, 15000)
+
+	coinsToMint := sdk.Coins{sdk.NewInt64Coin(defaultRewardDenom, 1000000)}
+
+	tests := map[string]struct {
+		// setup
+		numPools int
+
+		// expected
+		expectErr             bool
```
AlpinYukseloglu, line 107: Are there any error case we can test here?

x/poolmanager/types/hooks.go
```diff
@@ -0,0 +1,24 @@
...
+
+type PoolManagerHooks interface {
+	// AfterPoolCreated is called after CreatePool
+	AfterPoolCreated(ctx sdk.Context, sender sdk.AccAddress, poolId uint64)
+}
+
+var _ PoolManagerHooks = MultiPoolManagerhooks{}
+
+// combine multiple PoolManager hooks, all hook functions are run in array sequence.
+type MultiPoolManagerhooks []PoolManagerHooks
```
AlpinYukseloglu, line 13: What's the reason for having multiple combined hooks here?

AlpinYukseloglu (commented):
The core logic LGTM, but left a number of comments around testing and some nits/code hygiene comments that I'd like to be addressed before approving

x/incentives/types/expected_keepers.go
```diff
@@ -43,3 +45,25 @@ type CommunityPoolKeeper interface {
 type TxFeesKeeper interface {
 	GetBaseDenom(ctx sdk.Context) (denom string, err error)
 }
+
+type ClKeeper interface {
+	CreateIncentive(ctx sdk.Context, poolId uint64, sender sdk.AccAddress, incentiveDenom string, incentiveAmount sdk.Int, emissionRate sdk.Dec, startTime time.Time, minUptime time.Duration) (cltypes.IncentiveRecord, error)
+	GetPoolFromPoolIdAndConvertToConcentrated(ctx sdk.Context, poolId uint64) (cltypes.ConcentratedPoolExtension, error)
```
mattverse, line 51: Unused method
```suggestion
```

x/pool-incentives/types/expected_keepers.go
```diff
@@ -41,3 +47,14 @@ type DistrKeeper interface {
...
+	GetPoolFromPoolIdAndConvertToConcentrated(ctx sdk.Context, poolId uint64) (cltypes.ConcentratedPoolExtension, error)
```
mattverse, line 51: Unused method
```suggestion
```

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +375,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+
 	k.hooks.AfterEpochDistribution(ctx)
 
 	k.checkFinishDistribution(ctx, gauges)
 	return totalDistributedCoins, nil
 }
 
+// IsValidClGauge checks if a specific gaugeId is is associated with a poolId. It also checks if the poolType is CL pool.
+// If yes return true, otherwise false.
+func (k Keeper) IsValidClGauge(ctx sdk.Context, gaugeId uint64, duration time.Duration) (bool, poolmanagertypes.PoolI) {
```
mattverse, line 438: Can we have unit tests for this method please?

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+			expectedDistributions: sdk.Coins{fiveKRewardCoins},
+		},
+
+		"valid case: multiple gaugeId and poolId": {
+			numPools:              3,
+			expectErr:             false,
+			expectedDistributions: sdk.Coins{fifteenKRewardCoins},
+		},
+	}
+
```
mattverse, line 122: we also need test cases that tests mixture of providing test variables of non cl gauge + cl gague

x/incentives/keeper/distribute.go
```diff
@@ -258,6 +263,32 @@ func (k Keeper) distributeSyntheticInternal(
 	return k.distributeInternal(ctx, gauge, sortedAndTrimmedQualifiedLocks, distrInfo)
 }
 
+// distributeCLIncentiveInternal runs the distribution logic for CL pools only. It creates new incentive record with osmo incentives
+// and distributes all the tokens to the dedicated pool
+func (k Keeper) distributeCLIncentiveInternal(ctx sdk.Context, incentiveRecord cltypes.IncentiveRecord, gauge types.Gauge) (sdk.Coins, error) {
```
mattverse, line 268: can we also add unit tests for this method where under different conditions we check that
1. incentive was created
2. gauge has been correctly updated

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +93,97 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+			gauge, err := suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, gaugeId)
+			suite.Require().NoError(err)
+
+			gauge.Coins = sdk.Coins{sdk.NewInt64Coin(defaultRewardDenom, 5000)}
+
+			gauges = append(gauges, *gauge)
+		}
+
+		// Distribute tokens from the gauge
+		_, err := suite.App.IncentivesKeeper.Distribute(suite.Ctx, gauges)
```
mattverse, line 156: Can we also add checks for `totalDistributedCoins` returned from Distribute?

mattverse (commented):
Looking good, leaving some minor comments

x/pool-incentives/keeper/keeper_test.go
stackman27: added!

x/incentives/keeper/distribute.go
stackman27: I followed the current naming standard for "distirbuteInternal"

x/pool-incentives/keeper/keeper.go
stackman27: unfortunately, i cannot, not without modifying `CreateGauge` function because it is default to Query `lockuptypes.ByDuration`

x/incentives/keeper/distribute.go
stackman27: it's for tokens sent from moduletoUserAccounts, therefore at every successful transfer an event is emitted

x/incentives/keeper/distribute_test.go
stackman27: there's really one thing i can think of that we can check for invalid case; 
1. Amount cannot be <=0 here because `for _, coins := gauge.Coins` will only run for position coins 
2. Same thing ^ goes for emissionRate because we use coinAmt to calculate it 
3. I can check minUptime but it gets hardcoded as epoch value when pool gets created aka gauge creation 

the only thing to check would be if  starttime < currentBlockTime

x/poolmanager/types/hooks.go
stackman27: it's so that we can have multiple hook recievers for each hook. For example: Gamm has 2 hook recievers poolIncentive and twap. At the end of a specific function in Gamm we can trigger something in both poolIncentives & twap

x/incentives/keeper/distribute.go
stackman27: these checks has already been added as a part of TestDistributeToConcentratedLiquidityPools

x/incentives/keeper/distribute_test.go
stackman27: hmm, i tried testing this but even with non-cl gauge the distribute functions doesnot error and works as expected. This was to make things backwards compatible

x/incentives/keeper/distribute.go
stackman27: can you please give me an example of a potential security risk here? should sdk.Dec be able to handle it

AlpinYukseloglu (approved):
Test changes LGTM (note that I was only able to take a cursory glance since my last review)

x/poolmanager/create_pool.go
stackman27: https://github.com/osmosis-labs/osmosis/issues/4763

x/incentives/keeper/distribute.go
p0mvn: Floats are non-deterministic so if someone refactors this and accidentally uses the return of `currEpoch.Duration.Seconds()` in math operations, this will lead to an app hash.

Some examples and breakdown of non-determinism cases with floats: https://shaderfun.com/2020/10/25/understanding-determinism-part-1-intro-and-floating-points/

I suggest minimizing the use of APIs and types that can lead to non-determinism

x/pool-incentives/keeper/keeper.go
p0mvn: Please add an issue to revisit and refactor `CreateGauge` and the hack here

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +373,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+	if err != nil {
+		return false, nil
+	}
+
+	pool, err := k.pmk.GetPool(ctx, poolId)
+	if err != nil {
+		return false, nil
+	}
+
+	return pool.GetType() == poolmanagertypes.Concentrated, pool
```
p0mvn, line 447: The second return of `pool` does not fit the abstraction of `IsValidConcentratedLiquidityGauge`

Consider getting the pool from the caller, passing in the `poolType` as a parameter and returning a single boolean

x/incentives/keeper/distribute_test.go
```diff
@@ -345,3 +484,40 @@ func (suite *KeeperTestSuite) TestNoLockNonPerpetualGaugeDistribution() {
...
+	tests := []struct {
+		name              string
+		gaugeId           uint64
+		expectedGaugeType poolmanagertypes.PoolType
+		expectErr         bool
+	}{
+		{
+			name:              "Valid gaugeId and poolType",
+			gaugeId:           1,
+			expectedGaugeType: poolmanagertypes.Concentrated,
```
p0mvn, line 498: Would like to see a test with non-CL pool

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +373,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
 
 	locksByDenomCache := make(map[string][]lockuptypes.PeriodLock)
 	totalDistributedCoins := sdk.Coins{}
+	var gaugeDistributedCoins, gaugeDistributedCoinsCL sdk.Coins
```
p0mvn, line 376: I don't understand why we need to differentiate between these. Can we keep `totalDistributedCoins` only, please?

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +95,142 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+
+				// get the gaugeId corresponding to the CL pool
+				gaugeId, err := suite.App.PoolIncentivesKeeper.GetPoolGaugeId(suite.Ctx, poolId, currentEpoch.Duration)
+				suite.Require().NoError(err)
+
+				// get the gauge from the gaudeId
+				gauge, err := suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, gaugeId)
+				suite.Require().NoError(err)
+
+				gauge.Coins = sdk.NewCoins(sdk.NewCoin(defaultRewardDenom, sdk.NewInt(5000)))
```
p0mvn, line 169: Do we have a case where a gauge contains multiple coins? This should be added before merge IMO

x/incentives/keeper/distribute.go
```diff
@@ -344,32 +373,80 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+		// get pool Id corresponding to the gaugeId
+		incParams := k.GetParams(ctx).DistrEpochIdentifier
+		currEpoch := k.ek.GetEpochInfo(ctx, incParams)
+
+		isCLPoolGauge, pool := k.IsValidConcentratedLiquidityGauge(ctx, gauge.Id, currEpoch.Duration)
+		// only want to run this logic if the gaugeId is associated with CL PoolId
+		if isCLPoolGauge {
+			for _, coin := range gauge.Coins {
+				// since we're only doing osmo incentive we can assume there to be only 1 coin therefore coins[0]
+				incentiveRecord := cltypes.IncentiveRecord{
```
p0mvn, line 387: Why is the incentive record created here? It seems that we only use it as a structure to pass in the parameters, and a new incentive record is created by the `CreateIncentive` function.

I suggest passing in the parameters directly without relying on the `IncentiveRecord`. This is a CL-related abstraction that I think makes sense to keep encapsulated within the CL-module unless necessary . 

The goal is to reduce complexity (e.g. having it here has led me to spend 5 mins investigating why it is here)

x/incentives/keeper/distribute.go
```diff
@@ -196,26 +198,28 @@ func (d *distributionInfo) addLockRewards(owner string, rewards sdk.Coins) error
...
 	numIDs := len(distrs.idToDecodedAddr)
-	ctx.Logger().Debug(fmt.Sprintf("Beginning distribution to %d users", numIDs))
-	err := k.bk.SendCoinsFromModuleToManyAccounts(
-		ctx,
-		types.ModuleName,
-		distrs.idToDecodedAddr,
-		distrs.idToDistrCoins)
-	if err != nil {
-		return err
+	if numIDs > 0 {
```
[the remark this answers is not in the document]
stackman27: i feel like this check was missed so i added it. We always expect the number of ids to distribute tokens to be > 0

x/incentives/keeper/distribute.go
stackman27: i actually changed it to `GetValidConcentratedLiquidityGauge` and return Pool, lmk if that makes sense

x/incentives/keeper/distribute.go
stackman27: i mainly added it so that we can only return the amount distributed in CL, but fine to remove it rn since we're not doing anything with that value

x/incentives/types/expected_keepers.go
```diff
@@ -43,3 +45,25 @@ type CommunityPoolKeeper interface {
 type TxFeesKeeper interface {
 	GetBaseDenom(ctx sdk.Context) (denom string, err error)
 }
+
+type ClKeeper interface {
```
p0mvn, line 49:
```suggestion
type ConcentratedLiquidityKeeper interface {
```

x/pool-incentives/keeper/keeper.go
```diff
@@ -58,29 +63,60 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", "x/"+types.ModuleName)
 }
 
+// CreatePoolGauges checks if the poolId is CLPoolType, if it is create one gauge,
```
p0mvn, line 66:
```suggestion
// CreatePoolGauges checks if the poolId is CLPoolType, if it is create one gauge,
```
```suggestion
// CreatePoolGauges checks if the poolId is Concentrated PoolType, if it is create one gauge,
```

x/pool-incentives/keeper/keeper.go
```diff
@@ -58,29 +63,60 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+// otherwise create multiple gauges based on lockableDurations.
 func (k Keeper) CreatePoolGauges(ctx sdk.Context, poolId uint64) error {
```
p0mvn, line 68: Please add a test for this method for both CL and non-CL pool

x/pool-incentives/keeper/keeper_test.go
```diff
@@ -71,3 +71,25 @@ func (suite *KeeperTestSuite) TestCreateBalancerPoolGauges() {
 		suite.Equal(lockableDurations[2], gauge.DistributeTo.Duration)
 	}
 }
+
+func (suite *KeeperTestSuite) TestCreateCLPoolGauges() {
```
p0mvn, line 75:
```suggestion
func (suite *KeeperTestSuite) TestCreateCLPoolGauges() {
```
```suggestion
func (suite *KeeperTestSuite) TestCreateConcentratePoolGauges() {
```

x/poolmanager/create_pool.go
p0mvn, line 93: I know that we are addressing this in a separate issue but, for now, there should be a test covering that for CL pools this hook gets executed. For non-CL pools, it does not

x/poolmanager/types/hooks.go
```diff
@@ -0,0 +1,24 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+type PoolManagerHooks interface {
+	// AfterPoolCreated is called after CreatePool
+	AfterPoolCreated(ctx sdk.Context, sender sdk.AccAddress, poolId uint64)
+}
```
p0mvn, line 8: We should not need this. There are already `PoolCreationListeners` in `listeners.go` that we can use instead

p0mvn:
After thinking more about extensibility, testability, and how this plays into the `x/cosmwasmpool`, I think that we should have two separate hooks for each of `x/concentrated-liquidity` and `x/gamm`.

That is `AfterCFMMCreatedPool` and  `AfterConcentratedCreatedPool`. Then we can have a separate [`CreateGauges` method](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L68) instead of [checking for pool type here](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L73-L74).

This architecture will make it so that the logic is more testable since we can test each method and flow separately. It will also make it more secure since we have a separate execution flow for each type.

It will also allow us to integrate any other pool type such as `x/cosmwasmpool` without depending on this logic and requiring to add more of these `if pooltype is coswasm => not create gauges`

Additionally, this will allow us to avoid hacks like [this one](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/poolmanager/create_pool.go#L89-L93) where we have to check for pool type in multiple places - `x/poolmanager`, `x/pool-incentives` and `x/incentives`

This request is also related to https://github.com/osmosis-labs/osmosis/issues/4665 . To integrate CL into `x/twap`, we will have to separate the hooks. This goes in line with this suggestion.

Due to how many other fronts this change is affecting,  as well as the number of modules it is affecting, I think this should be addressed in this PR.

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3362,3 +3362,35 @@ func (s *KeeperTestSuite) TestFindUptimeIndex() {
 		})
 	}
 }
+
+func (s *KeeperTestSuite) TestCreateIncentivesNew() {
```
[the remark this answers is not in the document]
stackman27: this was just a dummy test to check the parsing error, it's safe to remove it

x/incentives/keeper/distribute.go
```diff
@@ -343,33 +371,82 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
 	for _, gauge := range gauges {
-		filteredLocks := k.getDistributeToBaseLocks(ctx, gauge, locksByDenomCache)
-		// send based on synthetic lockup coins if it's distributing to synthetic lockups
 		var gaugeDistributedCoins sdk.Coins
-		var err error
-		if lockuptypes.IsSyntheticDenom(gauge.DistributeTo.Denom) {
-			gaugeDistributedCoins, err = k.distributeSyntheticInternal(ctx, gauge, filteredLocks, &distrInfo)
+		clPool := k.GetValidConcentratedLiquidityGauge(ctx, gauge.Id, currentEpoch.Duration)
+		// only want to run this logic if the gaugeId is associated with CL PoolId
+		if clPool != nil {
```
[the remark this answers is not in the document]
stackman27: yea i can do that the only thing is we donot want to error in `GetValidConcentratedLiquidityGauge` because a gauge doesnot always have to be associated with a poolId

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +97,164 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+					for _, coin := range gauge.Coins {
+						// get poolId from GaugeId
+						poolId, err := suite.App.PoolIncentivesKeeper.GetPoolIdFromGaugeId(suite.Ctx, gauge.GetId(), currentEpoch.Duration)
+						suite.Require().NoError(err)
+
+						// check that incentive record wasn't created
+						_, err = suite.App.ConcentratedLiquidityKeeper.GetIncentiveRecord(suite.Ctx, poolId, coin.Denom, currentEpoch.Duration, suite.App.AccountKeeper.GetModuleAddress(types.ModuleName))
+						suite.Require().Error(err)
+					}
+				}
```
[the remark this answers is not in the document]
stackman27: added

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +97,164 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+			} else {
+				suite.Require().NoError(err)
+
+				// check if module amount got deducted correctly
+				balance := suite.App.BankKeeper.GetAllBalances(suite.Ctx, suite.App.AccountKeeper.GetModuleAddress(types.ModuleName))
+				actualbalanceAfterDistribution := coinsToMint.AmountOf(defaultRewardDenom).Sub(balance.AmountOf(defaultRewardDenom))
+
+				if tc.poolType == poolmanagertypes.Concentrated {
+					suite.Require().Equal(tc.expectedDistributions.AmountOf(defaultRewardDenom), actualbalanceAfterDistribution)
+				}
```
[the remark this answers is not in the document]
stackman27: changed

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +97,164 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+
+				for _, gauge := range gauges {
+					for _, coin := range gauge.Coins {
```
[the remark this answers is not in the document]
stackman27: yup added a clause that checks `suite.Require().NotEqual(len(gauges), 0)`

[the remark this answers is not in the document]
stackman27: Also in current setup, len(gauge) ==0 will only happen if it's a balancer pool gauge, since we donot create a pool if it's anything other than type concentrated

stackman27:
> After thinking more about extensibility, testability, and how this plays into the `x/cosmwasmpool`, I think that we should have two separate hooks for each of `x/concentrated-liquidity` and `x/gamm`.
> 
> That is `AfterCFMMCreatedPool` and `AfterConcentratedCreatedPool`. Then we can have a separate [`CreateGauges` method](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L68) instead of [checking for pool type here](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L73-L74).
> 

running AfterPoolCreate for every module (gamm, cl, cosmwasm) seperately makes sense, but i donot think we can actually we can createGauge from other modules except pool-incentives. Because of the following reasons; 
1. it uses its own KV store to store the created gauge information 
2. the minted tokens is stored in pool-incentives modules address which then gets transferred to incentives module and to lock addresses 
3.  This will break existing flow of gauge distribution which was not expected from this PR

x/concentrated-liquidity/clmodule/module.go
czarcas7ic: Please link issue here when created so we can resolve

stackman27: yup added https://github.com/osmosis-labs/osmosis/issues/4753

x/pool-incentives/types/expected_keepers.go
```diff
@@ -41,3 +46,11 @@ type DistrKeeper interface {
...
 	FundCommunityPool(ctx sdk.Context, amount sdk.Coins, sender sdk.AccAddress) error
 }
+
+type EpochKeeper interface {
+	GetEpochInfo(ctx sdk.Context, identifier string) epochstypes.EpochInfo
+}
+
+type GAMMKeeper interface {
+	GetPoolType(ctx sdk.Context, poolId uint64) (poolmanagertypes.PoolType, error)
+}
```
p0mvn, line 56: Why is this a gamm keeper? Shouldn't this be poolmanager keeper?

stackman27: actually we dont even use this, i'll remove it

x/pool-incentives/keeper/keeper.go
p0mvn: @stackman27 was there an issue created?

x/incentives/keeper/distribute.go
```diff
@@ -364,12 +419,29 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+// If yes return true, otherwise false.
+func (k Keeper) GetValidConcentratedLiquidityGauge(ctx sdk.Context, gaugeId uint64, duration time.Duration) poolmanagertypes.PoolI {
+	poolId, err := k.pik.GetPoolIdFromGaugeId(ctx, gaugeId, duration)
+	if err != nil {
+		return nil
+	}
+
+	pool, err := k.pmk.GetPool(ctx, poolId)
+	if err != nil {
+		return nil
```
p0mvn, line 439: There should be a second error return, and this should return error, not nil

x/incentives/keeper/distribute.go
```diff
@@ -364,12 +419,29 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
 	k.checkFinishDistribution(ctx, gauges)
 	return totalDistributedCoins, nil
 }
 
+// GetValidConcentratedLiquidityGauge checks if a specific gaugeId is is associated with a poolId. It also checks if the poolType is CL pool.
+// If yes return true, otherwise false.
+func (k Keeper) GetValidConcentratedLiquidityGauge(ctx sdk.Context, gaugeId uint64, duration time.Duration) poolmanagertypes.PoolI {
+	poolId, err := k.pik.GetPoolIdFromGaugeId(ctx, gaugeId, duration)
+	if err != nil {
+		return nil
```
p0mvn, line 434: Error should be returned here

x/incentives/keeper/distribute.go
```diff
@@ -343,19 +371,46 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
 	for _, gauge := range gauges {
-		filteredLocks := k.getDistributeToBaseLocks(ctx, gauge, locksByDenomCache)
-		// send based on synthetic lockup coins if it's distributing to synthetic lockups
 		var gaugeDistributedCoins sdk.Coins
-		var err error
-		if lockuptypes.IsSyntheticDenom(gauge.DistributeTo.Denom) {
-			gaugeDistributedCoins, err = k.distributeSyntheticInternal(ctx, gauge, filteredLocks, &distrInfo)
+		pool := k.GetValidConcentratedLiquidityGauge(ctx, gauge.Id, currentEpoch.Duration)
+		// only want to run this logic if the gaugeId is associated with CL PoolId
+		if pool != nil && pool.GetType() == poolmanagertypes.Concentrated {
```
p0mvn, line 384:
```suggestion
		if pool.GetType() == poolmanagertypes.Concentrated {
```

x/pool-incentives/keeper/keeper.go
stackman27: yup forgot to tag it https://github.com/osmosis-labs/osmosis/issues/4833

x/incentives/keeper/distribute.go
```diff
@@ -343,19 +371,46 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
+	currentEpoch := k.ek.GetEpochInfo(ctx, incentiveParams)
+
 	for _, gauge := range gauges {
-		filteredLocks := k.getDistributeToBaseLocks(ctx, gauge, locksByDenomCache)
-		// send based on synthetic lockup coins if it's distributing to synthetic lockups
 		var gaugeDistributedCoins sdk.Coins
-		var err error
-		if lockuptypes.IsSyntheticDenom(gauge.DistributeTo.Denom) {
-			gaugeDistributedCoins, err = k.distributeSyntheticInternal(ctx, gauge, filteredLocks, &distrInfo)
+		pool := k.GetValidConcentratedLiquidityGauge(ctx, gauge.Id, currentEpoch.Duration)
```
p0mvn, line 382: From previous comments, let's do an error check here

p0mvn:
> > After thinking more about extensibility, testability, and how this plays into the `x/cosmwasmpool`, I think that we should have two separate hooks for each of `x/concentrated-liquidity` and `x/gamm`.
> > That is `AfterCFMMCreatedPool` and `AfterConcentratedCreatedPool`. Then we can have a separate [`CreateGauges` method](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L68) instead of [checking for pool type here](https://github.com/osmosis-labs/osmosis/blob/9565b36dc622dd3dd4536a63a0449fde3fee18df/x/pool-incentives/keeper/keeper.go#L73-L74).
> 
> running AfterPoolCreate for every module (gamm, cl, cosmwasm) seperately makes sense, but i donot think we can actually `createGauge` from other modules except pool-incentives. Because of the following reasons;
> 
> 1. it uses its own KV store to store the created gauge information, that is tied up to pool-incentive module
> 2. the minted tokens is stored in pool-incentives modules address which then gets transferred to incentives module and to lock addresses. If we were to createGauge via different module we'd the modules would have to store the minted tokens
> 3. This will break existing flow of gauge distribution which was not expected from this PR

Discussed offline - will work on this in a separate PR after the twap logic from https://github.com/osmosis-labs/osmosis/pull/4831/files is in since it affects the same hooks that are used here

p0mvn (approved):
LGTM once the remaining comments are addressed 🚀

x/incentives/keeper/distribute.go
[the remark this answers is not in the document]
czarcas7ic: I currently do not see this added table driven test case

x/incentives/keeper/distribute.go
```diff
@@ -258,6 +261,31 @@ func (k Keeper) distributeSyntheticInternal(
 	return k.distributeInternal(ctx, gauge, sortedAndTrimmedQualifiedLocks, distrInfo)
 }
 
+// distributeConcentratedLiquidityInternal runs the distribution logic for CL pools only. It creates new incentive record with osmo incentives
+// and distributes all the tokens to the dedicated pool
+func (k Keeper) distributeConcentratedLiquidityInternal(ctx sdk.Context, poolId uint64, sender sdk.AccAddress, incentiveDenom string, incentiveAmount sdk.Int, emissionRate sdk.Dec, startTime time.Time, minUptime time.Duration, gauge types.Gauge) (sdk.Coins, error) {
```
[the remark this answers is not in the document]
czarcas7ic: I do not see a test added for this function

x/incentives/keeper/distribute.go
```diff
@@ -343,19 +371,46 @@ func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, er
...
-		// send based on synthetic lockup coins if it's distributing to synthetic lockups
 		var gaugeDistributedCoins sdk.Coins
-		var err error
-		if lockuptypes.IsSyntheticDenom(gauge.DistributeTo.Denom) {
-			gaugeDistributedCoins, err = k.distributeSyntheticInternal(ctx, gauge, filteredLocks, &distrInfo)
+		pool, isCLPool := k.GetValidConcentratedLiquidityGauge(ctx, gauge.Id, currentEpoch.Duration)
+		// only want to run this logic if the gaugeId is associated with CL PoolId
+		if isCLPool {
+			for _, coin := range gauge.Coins {
+				emissionRate := sdk.NewDecFromInt(coin.Amount).QuoTruncate(sdk.NewDec(currentEpoch.Duration.Milliseconds()).QuoInt(sdk.NewInt(1000)))
```
czarcas7ic, line 386: I think an in line comment explaining the calculation for the emission rate would be appropriate here.

x/incentives/keeper/distribute_test.go
czarcas7ic: Right, but shouldn't that be a test case still? Showing that a test case that mixes cl and non cl gauge does not error and works as expected?

x/incentives/keeper/distribute_test.go
```diff
@@ -92,6 +96,201 @@ func (suite *KeeperTestSuite) TestDistribute() {
...
+			expectedDistributions: sdk.NewCoins(sdk.NewCoin(defaultRewardDenom, sdk.NewInt(3000))),
+			lockExist:             true,
+			expectErr:             false, // we do not expect error because we run the gauge distribution to lock logic
+		},
+		"invalid case: attempt to createIncentiveRecord with starttime < currentBlockTime": {
+			numPools:       1,
+			poolType:       poolmanagertypes.Concentrated,
+			gaugeCoins:     sdk.NewCoins(sdk.NewCoin(defaultRewardDenom, sdk.NewInt(5000))),
+			gaugeStartTime: defaultGaugeStartTime.Add(-5 * time.Hour),
+			expectErr:      true,
```
czarcas7ic, line 175: Since we are taking a boolean error, did you confirm that the error we are getting in the test case is the one you are expecting?

x/pool-incentives/keeper/keeper.go
czarcas7ic: I did not see this get added

czarcas7ic (approved):

x/incentives/keeper/distribute_test.go
stackman27: yup i ended up adding this test case here https://github.com/osmosis-labs/osmosis/pull/4526/files#diff-d5967ddef40ae47ef4140b9c47026cac85b85d9d04173b271c6c23dd182abdbcR153

x/incentives/keeper/distribute_test.go
stackman27: yup i confirmed it before using bool

x/concentrated-liquidity/store_test.go
```diff
@@ -139,3 +140,38 @@ func (s *KeeperTestSuite) TestParseFullTickFromBytes() {
 		})
 	}
 }
+
+// TestParseIncentiveRecordFromBytes_KeySeparatorInAddress validates that parsing
+// succeeds even if the address contains the key separator. This is ensured
+// by base32 encoding of the key separator.
+func (s *KeeperTestSuite) TestParseIncentiveRecordFromBytes_KeySeparatorInAddress() {
+	s.T().Skip()
```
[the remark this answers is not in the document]
stackman27: fixed! sorry i missed this

x/incentives/keeper/distribute_test.go
```diff
@@ -587,3 +587,105 @@ func (suite *KeeperTestSuite) TestGetValidConcentratedLiquidityGauge() {
...
+			sender:          suite.TestAccs[0],
+			incentiveDenom:  defaultRewardDenom,
+			incentiveAmount: sdk.NewInt(200),
+			emissionRate:    sdk.NewDec(2),
+			startTime:       defaultGaugeStartTime,
+			minUptime:       time.Hour * 24,
+			gauges:          []perpGaugeDesc{defaultGauge, doubleLengthGauge},
+
+			expectedError: true,
+		},
```
czarcas7ic, line 657: I might be missing it, but what makes this an invalid incentive record

stackman27: it has to do with the starttime, here the startime is before blockTime. I'll make it a little more verbose

Issue #4527, opened 2023-03-07
CLI: list environments


p0mvn:
## Background

Recently we merged a quality of life improvement: https://github.com/osmosis-labs/osmosis/pull/3787

Upon using it more today, I realized that it would be useful to have an additional command for listing all available environments.

## Suggested Design

Create a command `list-env` that functions similar to `get-env` but prints all available environments

## Acceptance Criteria
- new CLI added
- prints all environments

p0mvn:
@hieuvubk @ThanhNhann folks, flagging this issue in case there is interest in picking it up

czarcas7ic:
If picking up this issue, please reach out to @p0mvn if the context in this issue is not clear enough.

Side note, should add docs on how to properly utilize this somewhere.

PR #4528, opened 2023-03-07
add list-env

DongLieu:

CLI: list environments #4527

## Testing and Verifying
golangci-lint run --fix

### DIFF
--- a/cmd/osmosisd/cmd/change_environment.go
+++ b/cmd/osmosisd/cmd/change_environment.go
@@ -105,3 +105,40 @@ func environmentNameToPath(environmentName string) (string, error) {
 		return osmosisdPath, nil
 	}
 }
+func PrintAllEnvironmentCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list-env ",
+		Short: "listing all available environments.",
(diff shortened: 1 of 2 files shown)


cmd/osmosisd/cmd/change_environment.go
```diff
@@ -105,3 +105,40 @@ func environmentNameToPath(environmentName string) (string, error) {
...
+			fmt.Println("Environment name: ", EnvLocalnet)
+			fmt.Println("Environment path: ", path)
+
+			environment := getHomeEnvironment()
+			path, err = environmentNameToPath(environment)
+			if err != nil {
+				return err
+			}
+			fmt.Println("Environment name: ", environment)
+			fmt.Println("Environment path: ", path)
```
p0mvn, line 138: For custom path, currently, there is no easy way to list them all as they all are created under `$HOME`. I think we should change that and make it be created under `$HOME/.osmosisd-custom/<custom path>`. Then, to list the available custom environments, we simply iterate the  `$HOME/.osmosisd-custom/` folder printing all of its directory contents

DongLieu:
I just corrected it with a new commit, please check if it's correct @p0mvn

PR #4529, opened 2023-03-07
Remove build from docker compose state export

mattverse:
Removes build from the docker-compose file being used for state export.
The reason build is to be removed is since we already specified image to be ran in. 
Having secondary command for build would result in docker container pulling the image and then build and overwrite the container into the binary state it is in, not from the image that has been pulled

### DIFF
--- a/tests/localosmosis/state_export/docker-compose.yml
+++ b/tests/localosmosis/state_export/docker-compose.yml
@@ -4,12 +4,6 @@ services:
 
   osmosisd:
     image: local:osmosis
-    build:
-      context: ../../../
-      dockerfile: Dockerfile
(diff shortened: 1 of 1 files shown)


niccoloraspa:
Hey @mattverse , the `build` step is necessary as you wouldn't normally have the `local:osmosis` image in your local docker registry.

Without it, how could you start localosmosis?

mattverse:
Initial status we need the build, but when this script gets used for upgrades, we then face problem as the build overwrites the provided image. Suggested way here is to instead of deleting the build lines, separate out docker compose files for each usage 1) starting local osmosis 2) doing upgrades using images

PR #4530, opened 2023-03-07
Skip consensus min fee checks during simulate

ValarDragon:
A problem cited by people on the testnet, was that the consensus min fee violated their expectations around simulate txs.

Namely that they should be able to simulate execution with 0 fees, in order to determine the required amount of fees.

## Brief Changelog

- Skip consensus min fee check during tx simulation

## Testing and Verifying

Added a simulate test vector.

Raising an SDK issue for simulation tx guarantees

### DIFF
--- a/tests/e2e/scripts/hermes_bootstrap.sh
+++ b/tests/e2e/scripts/hermes_bootstrap.sh
@@ -43,7 +43,7 @@ account_prefix = 'osmo'
 key_name = 'val01-osmosis-a'
 store_prefix = 'ibc'
 max_gas = 6000000
-gas_multiplier = 1.1
+gas_multiplier = 1.5
 default_gas = 400000
 gas_price = { price = 0.0025, denom = 'e2e-default-feetoken' }
 clock_drift = '1m' # to accomdate docker containers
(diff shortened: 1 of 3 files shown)


p0mvn (approved):

p0mvn (changes requested):
Please DNM.

I think increasing `gas_multiplier` to make e2e pass wasn't exactly right here. Going to check more

p0mvn:
Conclusion on why adjustng `gas-adjustment` worked:

Previously, the simulation would fail when Hermes attempted to simulate a tx. As a result, Hermes used a `default_value` from its config that satisfied the actual non-simulated tx. 

Refs:
- https://hermes.informal.systems/documentation/configuration/description.html?highlight=gas_multiplier#description-of-the-parameters
- https://hermes.informal.systems/advanced/troubleshooting/cross-comp-config.html?highlight=gas_multiplier#out-of-gas

Now, that we fixed simulation, Hermes is able to run the simulation logic without errors and receiving a `gasUsed` estimate. However, that estimates ends up being lower that actually needed due to known variability in gas estimation. As a result, a higher gas adjustment ends up being necessary in e2e

p0mvn (approved):

Issue #4426, opened 2023-02-27
refactor(CL Swaps): remove redundant swap API: SwapInAmtGivenOut


p0mvn:
## Background

Similar to https://github.com/osmosis-labs/osmosis/issues/4425 but for the swap in the other direction.

## Acceptance Criteria
- method removed
- logic ported to another public API
- tests updated

p0mvn:
Complete in: https://github.com/osmosis-labs/osmosis/pull/4533

PR #4533, opened 2023-03-08
refactor/fix(CL Swaps): inverse tests; in given out API reduction

p0mvn:
Closes: #4475 , #4426

This PR focuses on fixing the inverse in given out tests. To be able to fix them, I had to overlap with #4426 (where the redundant swap API was to be removed) and addressed it here.

The core of the fix is to allow an additive tolerance of 1 when restrictive multiplicative tolerance fails. There are some cases where our test vector values are small while multiplicative error is also insignificant (`0.01%`). As a result, an difference of just 1 makes the multiplicative check fail.

The validation is now relaxed by `validateAmountsWithTolerance` test function.

There are also a few drive-by changes to address code smells. Namely:
- add and test `GetPriceLimit` function for getting the price limit depending on swap direction
- return reusable errors and use them in assertions for certain test vectors
- add a few test vectors to swaps
   * user does not have enough balance for swap for both in given out and out given in
 - remove `applySwap` function and move its contents into `updatePoolForSwap`
     * reason: modules should be deep, and it does not make sense to have 2 separate semantically same functions that mutate the pool in different ways. There should be one that applies all mutation.

## Additional tests

- [x] test `updatePoolForSwap`
- [x] test covering error edge case where amount < 1 computed - out given in
 * was unable to set up a similar in given out test case, concluding that this test case is not reproducible

## Testing and Verifying

Already covered by tests, fixed previously failing tests and added new


x/concentrated-liquidity/swaps_test.go
```diff
@@ -2110,10 +2153,20 @@ func (s *KeeperTestSuite) TestSwapExactAmountOut() {
...
+		},
+		{
+			name: "not enough balance",
+			param: param{
+				tokenOut:         sdk.NewCoin(ETH, sdk.NewInt(13370)),
+				tokenInDenom:     USDC,
+				tokenInMaxAmount: sdk.NewInt(66891663).Sub(differenceFromMax.Mul(sdk.NewInt(2))),
+				expectedTokenIn:  sdk.NewInt(66891663),
+			},
+			expectedErr: &types.InsufficientUserBalanceError{},
```
czarcas7ic, line 2169: Same point on pool balance error branch

x/concentrated-liquidity/swaps_test.go
```diff
@@ -2528,23 +2580,115 @@ func (s *KeeperTestSuite) TestInverseRelationshipSwapInAmtGivenOut() {
...
+
+			// Default pool values are initialized to one.
+			pool.ApplySwap(sdk.OneDec(), sdk.OneInt(), sdk.OneDec())
+
+			err := concentratedLiquidityKeeper.UpdatePoolForSwap(suite.Ctx, pool, sender, tc.tokenIn, tc.tokenOut, tc.newCurrentTick, tc.newLiquidity, tc.newSqrtPrice)
+
+			if tc.expectError != nil {
+				suite.Require().Error(err)
+				suite.Require().ErrorAs(err, &tc.expectError)
+				return
```
czarcas7ic, line 2653: Shouldn't we be checking the pool does not change in the error case?

czarcas7ic (approved):
Well done 👍

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1987,7 +1998,27 @@ func (s *KeeperTestSuite) TestSwapExactAmountIn() {
...
+			name: "not enough balance",
+			param: param{
+				tokenIn:           sdk.NewCoin(USDC, sdk.NewInt(42000000)),
+				tokenOutDenom:     ETH,
+				tokenOutMinAmount: types.MinSpotPrice.RoundInt(),
+				expectedTokenOut:  sdk.NewInt(8396),
+				underFundBy:       sdk.OneInt(),
+			},
+			expectedErr: &types.InsufficientUserBalanceError{},
+		},
```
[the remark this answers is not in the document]
p0mvn: I was unable to set up insufficient pool balance but I think it makes sense to keep the check as defense in depth

x/concentrated-liquidity/internal/swapstrategy/swap_strategy.go
```diff
@@ -78,3 +82,16 @@ func New(zeroForOne bool, sqrtPriceLimit sdk.Dec, storeKey sdk.StoreKey, swapFee
 	}
 	return &oneForZeroStrategy{sqrtPriceLimit: sqrtPriceLimit, storeKey: storeKey, swapFee: swapFee}
 }
+
+// GetPriceLimit returns the price limit based on which
+// token is being swapped in.
+// If zero in for one out, the price is decreasing.
+// Therefore, min spot price is the limit.
+// If one in for zero out, the price is increasing.
+// Therefore, max spot price is the limit.
```
AlpinYukseloglu, line 91:
```suggestion
// GetPriceLimit returns the price limit based on which token is being swapped in.
// If zero in for one out, the price is decreasing. Therefore, min spot price is the limit.
// If one in for zero out, the price is increasing. Therefore, max spot price is the limit.
```

x/concentrated-liquidity/model/pool.go
```diff
@@ -238,11 +238,10 @@ func (p Pool) isCurrentTickInRange(lowerTick, upperTick int64) bool {
 
 // ApplySwap state of pool after swap.
 // It specifically overwrites the pool's liquidity, curr tick and the curr sqrt price
-func (p *Pool) ApplySwap(newLiquidity sdk.Dec, newCurrentTick sdk.Int, newCurrentSqrtPrice sdk.Dec) error {
+func (p *Pool) ApplySwap(newLiquidity sdk.Dec, newCurrentTick sdk.Int, newCurrentSqrtPrice sdk.Dec) {
```
AlpinYukseloglu, line 241: Should we add a validity check here or are we assuming the parameters are correctly passed in? e.g. return err if liq < 0 etc.

x/concentrated-liquidity/swaps.go
```diff
@@ -535,7 +522,8 @@ func (k Keeper) calcInAmtGivenOut(
 	}
 
 	// coin amounts require int values
-	// truncate amountIn up to avoid under charging
+	// Round amount in up to avoid under charging the user.
+	// Round amount out down to avoid over charging the pool.
 	amt0 := swapState.amountCalculated.Ceil().TruncateInt()
 	amt1 := desiredTokenOut.Amount.ToDec().Sub(swapState.amountSpecifiedRemaining).TruncateInt()
```
AlpinYukseloglu, line 528: Shouldn't this rounding behavior be dependent on `zeroForOne`? i.e. if we're swapping from asset 1 to asset 0 we would expect the opposite rounding behavior right?

x/concentrated-liquidity/swaps_test.go
```diff
@@ -55,6 +56,15 @@ type SwapTest struct {
 }
 
 var (
+	// Allow 0.01% margin of error.
+	multiplicativeTolerance = osmomath.ErrTolerance{
+		MultiplicativeTolerance: sdk.MustNewDecFromStr("0.0001"),
+	}
+	// Allow additive margin equal of 1. It may stem from round up token in and rounding down
+	// tokenOut in favor of the pool.
+	oneAdditiveTolerance = osmomath.ErrTolerance{
```
AlpinYukseloglu, line 65: Does this tolerance not cause issues when we're dealing with large numbers? I believe our balancer and stableswap rounding guarantees are just purely `10^-8` multiplicative

AlpinYukseloglu (approved):
High level logic LGTM (note that I didn't dig too deep into specific test cases). Left a couple comments around rounding behavior that I'd like to be addressed before merging

x/concentrated-liquidity/swaps.go
```diff
@@ -145,51 +121,62 @@ func (k Keeper) SwapExactAmountOut(
...
 	priceLimit sdk.Dec,
-	poolId uint64,
 ) (calcTokenIn, calcTokenOut sdk.Coin, currentTick sdk.Int, liquidity, sqrtPrice sdk.Dec, err error) {
-	writeCtx, tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcInAmtGivenOut(ctx, desiredTokenOut, tokenInDenom, swapFee, priceLimit, poolId)
+	writeCtx, tokenIn, tokenOut, newCurrentTick, newLiquidity, newSqrtPrice, err := k.calcInAmtGivenOut(ctx, desiredTokenOut, tokenInDenom, swapFee, priceLimit, pool.GetId())
 	if err != nil {
 		return sdk.Coin{}, sdk.Coin{}, sdk.Int{}, sdk.Dec{}, sdk.Dec{}, err
 	}
 
+	// check that the tokenOut calculated is both valid and less than specified limit
```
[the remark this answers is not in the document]
p0mvn: This seems outdated. There is no such comment there anymore

x/concentrated-liquidity/swaps.go
```diff
@@ -52,33 +52,24 @@ func (k Keeper) SwapExactAmountIn(
...
+	tokenIn, tokenOut, _, _, _, err := k.swapOutAmtGivenIn(ctx, sender, pool, tokenIn, tokenOutDenom, swapFee, priceLimit)
 	if err != nil {
 		return sdk.Int{}, err
 	}
-
-	// check that the tokenOut calculated is both valid and less than specified limit
 	tokenOutAmount = tokenOut.Amount
-	if !tokenOutAmount.IsPositive() {
-		return sdk.Int{}, fmt.Errorf("token amount must be positive: got %v", tokenOutAmount)
-	}
```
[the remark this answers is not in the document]
p0mvn: Yes, this is ensured by `swapOutAmtGivenIn(`

x/concentrated-liquidity/swaps.go
p0mvn: zeroForOne means that zero is token in and one is token out.

We always want to round up token in and round down token out.

We already guarantee that no matter what the value of zeroForOne is, we always round up token in and round down token out

x/concentrated-liquidity/model/pool.go
p0mvn: Agree that it doesn't hurt to add validation but this is outside of the scope of this PR so I will make an issue

p0mvn: https://github.com/osmosis-labs/osmosis/issues/4744

x/concentrated-liquidity/swaps_test.go
p0mvn: We actually first check multiplicative tolerance which is 10^-4.

If the multiplicative tolerance check fails because we operate on small values, then we fall back to an additive tolerance of 1 which we assume is fine due to rounding.

See this for details: 
https://github.com/osmosis-labs/osmosis/blob/a522eea21804350b620f91d9b6c282ac12c7d89e/x/concentrated-liquidity/swaps_test.go#L2731-L2744

x/concentrated-liquidity/swaps_test.go
p0mvn: [b76748b](https://github.com/osmosis-labs/osmosis/pull/4533/commits/b76748be5ad31e7079425cc991124258263f4114)

p0mvn:
Addressed all comments. Planning on merging upon CI passing

PR #4549, opened 2023-03-09
added single pool query

nicolaslara:
Closes: #4481 

Estimate swap queries can't be executed via LCD endpoints due to an issue with grpc gateway that prevents (repeated?) nested structures from being specified as query params.

This adds simple queries for estimating the price when using a single pool. This also simplifies querying in most cases (where users just care about one pool)

## Brief Changelog

 * Added two new queries:
     *  /osmosis.poolmanager.v1beta1.Query/EstimateSinglePoolSwapExactAmountOut
     * /osmosis.poolmanager.v1beta1.Query/EstimateSinglePoolSwapExactAmountIn
     

## Testing and Verifying

Added tests that check that the results of those queries is the same as the underlying implementations
All other tests pass

  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes /)


p0mvn (approved):

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -45,8 +59,8 @@ message ParamsResponse { Params params = 1 [ (gogoproto.nullable) = false ]; }
 
 //=============================== EstimateSwapExactAmountIn
 message EstimateSwapExactAmountInRequest {
-  // TODO: CHANGE THIS TO RESERVED IN A PATCH RELEASE
-  string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
+  reserved 1;
+  reserved "sender";
```
ValarDragon, line 63: Is this query in the stargate whitelist? If so, then seems important to reason about if this would cause a state break on mainnet. My first pass thought is no, but I don't really know.

ValarDragon (approved):
LGTM, but feels like we should try to get this backported / into a patch release

ValarDragon:
In a second non-state compatible PR, lets add these new queries to the whitelist?

proto/osmosis/poolmanager/v1beta1/query.proto
nicolaslara: It is. It's also potentially used by contracts. 

One place that may be state breaking is https://github.com/osmosis-labs/osmosis/pull/4549/files#diff-417b3334c7fd0aefd80aafa70e8feb77a81586dc63b266105aa25bad2ab09a2fL53. It was (unnecessarily) checking that the sender wasn't empty. So a call with an empty sender would've failed before and would pass now.

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -27,13 +27,27 @@ service Query {
         "/osmosis/gamm/v1beta1/{pool_id}/estimate/swap_exact_amount_in";
   }
 
+  rpc EstimateSinglePoolSwapExactAmountIn(
+      EstimateSinglePoolSwapExactAmountInRequest)
+      returns (EstimateSwapExactAmountInResponse) {
+    option (google.api.http).get = "/osmosis/gamm/v1beta1/{pool_id}/estimate/"
```
mattverse, line 33:
```suggestion
    option (google.api.http).get = "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/"
```

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -27,13 +27,27 @@ service Query {
         "/osmosis/gamm/v1beta1/{pool_id}/estimate/swap_exact_amount_in";
```
mattverse, line 27:
```suggestion
        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/swap_exact_amount_in";
```

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -27,13 +27,27 @@ service Query {
...
+      returns (EstimateSwapExactAmountInResponse) {
+    option (google.api.http).get = "/osmosis/gamm/v1beta1/{pool_id}/estimate/"
+                                   "single_pool_swap_exact_amount_in";
+  }
+
   // Estimates swap amount in given out.
   rpc EstimateSwapExactAmountOut(EstimateSwapExactAmountOutRequest)
       returns (EstimateSwapExactAmountOutResponse) {
     option (google.api.http).get =
         "/osmosis/gamm/v1beta1/{pool_id}/estimate/swap_exact_amount_out";
```
mattverse, line 41:
```suggestion
        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/swap_exact_amount_out";
```

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -27,13 +27,27 @@ service Query {
...
   }
 
+  rpc EstimateSinglePoolSwapExactAmountOut(
+      EstimateSinglePoolSwapExactAmountOutRequest)
+      returns (EstimateSwapExactAmountOutResponse) {
+    option (google.api.http).get = "/osmosis/gamm/v1beta1/{pool_id}/estimate/"
```
mattverse, line 47:
```suggestion
    option (google.api.http).get = "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/"
```

proto/osmosis/poolmanager/v1beta1/query.proto
nicolaslara: ohh good catch on all of these! committing them

x/poolmanager/client/cli/query.go
```diff
@@ -76,7 +76,6 @@ func EstimateSwapExactAmountInParseArgs(args []string, fs *flag.FlagSet) (proto.
 	}
 
 	return &queryproto.EstimateSwapExactAmountInRequest{
```
mattverse, line 78: the query cli still takes in sender as argument, reminder to fix the use, short, long for the cmd method!

x/poolmanager/client/cli/query.go
```diff
@@ -95,7 +94,6 @@ func EstimateSwapExactAmountOutParseArgs(args []string, fs *flag.FlagSet) (proto
 	}
```
mattverse, line 95: can we also add cli for the newly added queries?

x/poolmanager/client/cli/query.go
nicolaslara: Doesn't it get automatically added by https://github.com/osmosis-labs/osmosis/blob/95de2fe591cf4b46149052e842b52f6016853938/proto/osmosis/poolmanager/v1beta1/query.yml#L21 ? 🤔

mattverse: Don't quite know, rn I'm testing this on an actual environment, seems like it still has the field

mattverse: I think we manually have to fix `GetCmdEstimateSwapExactAmountIn`

proto/osmosis/poolmanager/v1beta1/query.proto
mattverse: good catch, this leads me to think that we might want to add queries included in stargate queries to e2e testing, since we can mistakenly merge a query PR and have it state breaking

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -24,14 +24,30 @@ service Query {
...
-        "/osmosis/gamm/v1beta1/{pool_id}/estimate/swap_exact_amount_in";
+        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/swap_exact_amount_in";
+  }
+
+  rpc EstimateSinglePoolSwapExactAmountIn(
+      EstimateSinglePoolSwapExactAmountInRequest)
+      returns (EstimateSwapExactAmountInResponse) {
+    option (google.api.http).get =
+        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate/"
+        "single_pool_swap_exact_amount_in";
```
mattverse, line 35: oof just noticed that this is automatically done by linter, should not be divided by lines, or else would not work

nicolaslara: I assumed it automatically concatenated adjacent strings (some other languages do that). Is that not the case? it seems like the proto generation works

x/poolmanager/client/cli/query.go
nicolaslara: added

nicolaslara: I see what you meant now. Should be fixed in the latest commit

proto/osmosis/poolmanager/v1beta1/query.proto
nicolaslara: changed to one line just in case

mattverse: thought so as well, but I don't think proto compilers support this tho 😭

x/poolmanager/client/cli/cli_test.go
```diff
@@ -74,9 +74,9 @@ func TestNewSwapExactAmountOutCmd(t *testing.T) {
 	desc, _ := cli.NewSwapExactAmountOutCmd()
 	tcs := map[string]osmocli.TxCliTestCase[*types.MsgSwapExactAmountOut]{
 		"swap exact amount out": {
-			Cmd: "10stake 20 --swap-route-pool-ids=1 --swap-route-denoms=node0token --from=" + testAddresses[0].String(),
+			Cmd: "10stake 20 --swap-route-pool-ids=1 --swap-route-denoms=node0token",
 			ExpectedMsg: &types.MsgSwapExactAmountOut{
-				Sender:           testAddresses[0].String(),
+				Sender:           "",
```
mattverse, line 79: why are we removing Sender from `SwapExactAmountOutCmd`? Shouldn;t `SwapExactAmountOutCmd` still have the sender?

mattverse (approved):
Just one question I have in the change in the cli test, otherwise LGTM

x/poolmanager/client/cli/cli_test.go
nicolaslara: yes! I just removed it from the wrong place (search and replace)

nicolaslara:
Since this is not gonna be state compatible anyway, added the stargate whitelist queries here.

mattverse:
> Should we maybe backport this without the sender change so that thosse queries are available in a minor release? I think that's not really necessary, as these can still be queried from typescript via the rpc endpoints

I don't think it should be necessary as well, it has been in the current state for years, I personally think it can wait

PR #4567, opened 2023-03-10
ProtoRev: Adding Txs to CLI

davidterpay:
In this PR, I add the remaining admin transactions to the CLI. Additionally, I add one modification to `MsgSetMaxPoolPointsPer(Tx/Block)` so that the max pool points per tx cannot be set to something higher than the max pool points per block and vice versa.

## Brief Changelog
  - *Added all protorev txs to the cli*
  - *Added an additional check to `SetMaxPoolPointsPerTx` and `SetMaxPoolPointsPerBlock`*
  - *Update the protorev spec*

## Testing and Verifying
This change is already covered by existing tests. Additionally, I tested all of the CLI commands locally with accounts that were designated as the admin account and those that were not, with various params (correct and incorrect), and more. The CLI commands function as expected.

Updated the `x/protorev` spec to include all of the updates made in this PR.

### DIFF
--- a/x/protorev/client/cli/utils.go
+++ b/x/protorev/client/cli/utils.go
@@ -21,7 +21,7 @@ type Trade struct {
 
 type ArbRoutes struct {
 	Trades   []Trade `json:"trades"`
-	StepSize sdk.Int `json:"step_size"`
+	StepSize uint64  `json:"step_size"`
 }
 
 type hotRoutesInput struct {
@@ -30,11 +30,9 @@ type hotRoutesInput struct {
 	ArbRoutes []ArbRoutes `json:"arb_routes"`
 }
 
-type createHotRoutesInput struct {
-	HotRoutes []hotRoutesInput
-}
+type createArbRoutesInput []hotRoutesInput
 
-type XCreateHotRoutesInputs createHotRoutesInput
+type XCreateHotRoutesInputs hotRoutesInput
 
 type XCreateHotRoutesExceptions struct {
 	XCreateHotRoutesInputs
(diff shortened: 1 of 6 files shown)


x/protorev/client/cli/utils.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+
+			current.ArbRoutes = append(current.ArbRoutes, currentArbRoute)
+		}
+
+		tokenPairArbRoutes = append(tokenPairArbRoutes, current)
+	}
+
+	return tokenPairArbRoutes
+}
```
davidterpay, line 95: Not sure if there is a more optimal way of reading in the json file and building the `TokenPairArbRoutes` since the only thing that is problematic is dealing with the StepSize which is an `sdk.Int`.

x/protorev/keeper/msg_server.go
```diff
@@ -71,6 +71,15 @@ func (m MsgServer) SetMaxPoolPointsPerTx(c context.Context, msg *types.MsgSetMax
 		return nil, err
 	}
 
+	maxPointsPerBlock, err := m.k.GetMaxPointsPerBlock(ctx)
```
p0mvn, line 74: Non-blocking but just FYI for future contributions that it is beneficial to separate state-breaking changes from non-state-breaking so that we can release non-state-breaks sooner.

We have docs for this here:
https://github.com/osmosis-labs/osmosis/blob/d945e6230d042d26c700717b2373cd6d5976e48b/CONTRIBUTING.md?plain=1#L365

If we only had CLI changes in this PR, we could release them in the v15 line. Since there are message server changes, this will have to wait till v16.

x/protorev/client/cli/utils.go
p0mvn: The only other example I know of parsing a file withing our cli is the pool file:
https://github.com/osmosis-labs/osmosis/blob/3cd13a674a7647a4b3f48fad0c1fa7091a66b167/x/gamm/client/cli/parse.go#L27-L39

This implementation seems to be aligned with that. If parsing files becomes a common theme, we can better abstract it in the future

p0mvn (approved):
utACK

Thanks for this change and for testing it out locally

PR #4573, opened 2023-03-11
refactor(CL): position state - only store sdk.Dec instead of position's struct

p0mvn:
Closes: #4593

This change is needed to productive implement genesis for CL.

Since the only value for position state entry is liquidity, we can serialize `sdk.DecProto` into state as opposed to defining a separate struct. The bulk of the data can be stored in key.

Also, moving query logic into its own package to avoid import cycles.

## Brief Changelog

- change position state value to `sdk.DecProto`
- queries in their own package

## Testing and Verifying


p0mvn:
This is still active, reopened in: https://github.com/osmosis-labs/osmosis/pull/4602

Issue #4515, opened 2023-03-06
[CL Incentives] Key incentive records by owner address to prevent DoS/trolling


AlpinYukseloglu:
## Background

We currently key incentive records by `poolId`, `denom`, and `minUptime`. While this is technically sufficient and robust, it allows for any user to dust each uptime accumulator with a denom and essentially prevent that denom from being used for incentives.

This became much bigger of an issue when we constrained the supported uptimes, since it becomes trivially easy to do this.

## Suggested Design

- I suggest we add `incentiveOwner` to the end of the incentive record key. This prevents anyone from being blocked from creating incentives due to existing incentives on the pool. This will require updating incentive record getters and setters.
- To ensure incentive retrieval remains simple, I suggest we move `minUptime` in front of `denom` in the key and then implement a `GetAllIncentiveRecordsForUptime` function that gathers all the incentives for a given `poolId | minUptime` from state

## Acceptance Criteria

- All new and existing tests pass

PR #4581, opened 2023-03-12
[CL Incentives] Key incentive records by owner address to prevent DoS/trolling

AlpinYukseloglu:
Closes: #4515

This PR adds the incentive creator's address to the key of each incentive record. This is to prevent the DoS vector where someone could create incentives for a specific denom on all supported uptimes with dust amounts and thus make it impossible for others to emit incentives on that denom.

## Brief Changelog

- Add incentive creator address to incentive record key
- Also add it to the `IncentiveRecord` helper struct for convenience and backwards compatibility

## Testing and Verifying

Thanks to the second change above & already robust incentive record tests, existing tests cover the changes made after minor updates.


stackman27 (approved):
This LGTM! very straightforward

x/concentrated-liquidity/types/keys.go
```diff
@@ -94,8 +94,9 @@ func KeyPool(poolId uint64) []byte {
 	return []byte(fmt.Sprintf("%s%d", PoolPrefix, poolId))
 }
 
-func KeyIncentiveRecord(poolId uint64, denom string, minUptime time.Duration) []byte {
-	return []byte(fmt.Sprintf("%s%s%d%s%s%s%d", IncentivePrefix, KeySeparator, poolId, KeySeparator, denom, KeySeparator, uint64(minUptime)))
+func KeyIncentiveRecord(poolId uint64, denom string, minUptime time.Duration, addr sdk.AccAddress) []byte {
+	addrKey := address.MustLengthPrefix(addr.Bytes())
```
p0mvn, line 98: I noticed similar formatting while working on positions and ended up removing them due to the edge cases that stem from the prefix: https://github.com/osmosis-labs/osmosis/pull/4602

Why do we need it? If we are surrounding an address with a key separator, there is no need to know the length.

The length should only be needed when we are unsure how to deserialize the address from the store. Given that we have key separator placeholders around, we should be able to write the address directly to the store

p0mvn (approved):

Issue #4577, opened 2023-03-11
fix: auto upgrade handler PR is not linted and does not have state compatibility lables applied


p0mvn:
## Background

Ref: https://github.com/osmosis-labs/osmosis/pull/4575

The PR was automatically created but it was not listed and did not have state compatibility labels applied

## Suggested Design

We need to make sure that the problems above are mitigated.

## Acceptance Criteria
- apply `V:state/breaking` label to upgrade handler PR
- the PR is created without lint issues

PR #4588, opened 2023-03-13
Fix linting issues in auto upgrade handler generation

pysel:
Closes: #4577

fixes linting issues in auto upgrade handler generation


czarcas7ic (approved):
utACK

p0mvn (approved):

PR #4589, opened 2023-03-13
Format Osmosis

pysel:
`go fmt ./...`


pysel:
formatted files were proto-generated. By changing them, `check-proto` action fails: going to close

Issue #4591, opened 2023-03-13
feat: implement position and tick genesis for concentrated liquidity 


p0mvn:
## Background

The current genesis only has pools. We should add all useful state into it.

## Suggested Design

This issue focuses on position and tick genesis only

## Acceptance Criteria
- position and tick genesis added
- tested

PR #4628, opened 2023-03-14
feat: parse full tick information from key and value

p0mvn:
Progress toward: #4591

This pull request introduces the addition of the ParseFullTickFromBytes function and its related helper functions to the Osmosis codebase. These functions provide a way to parse tick information from byte slices, improving the handling of tick data in the system.

This new function is helpful for getting all of the tick-specific data to be exported in genesis.

The following function has been added in this PR:

- `ParseFullTickFromBytes`: Parses a FullTick struct from key and value byte slices.
- `ParseTickFromBz`: Parses a TickInfo struct from a byte slice using protobuf Unmarshal.

The following have been modified:
- `keyTickPrefixByPoolIdPrealloc`: Generates a tick prefix key for a given pool by concatenating the TickPrefix, KeySeparator, and the big-endian representation of the pool id. Takes an int value for the number of bytes to pre-allocate.
- `KeyTick`: Generates a tick key for a given pool and tick index by concatenating the tick prefix key, KeySeparator, and the tick index bytes.
- `KeyTickPrefixByPoolId`: Generates a tick prefix key for a given pool by calling the keyTickPrefixByPoolIdPrealloc function with the appropriate pre-allocated memory size.

These functions will improve the overall handling of tick data and ensure more efficient parsing and storage of tick information.

Additionally, this PR moves genesis logic to `types/genesis` to avoid import cycles.

## Brief Changelog

- move genesis to `types/genesis` to avoid import cycles
- add `ParseFullTickFromBytes` function and tess
- add helper functions 
- add/improve specs
- added custom error types for tests

## Testing and Verifying

This change added tests.


x/concentrated-liquidity/store_test.go
```diff
@@ -184,3 +205,113 @@ func (s *KeeperTestSuite) TestParseFullPositionFromBytes() {
...
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		s.Run(name, func() {
+
+			fullTick, err := cl.ParseFullTickFromBytes(tc.key, tc.val)
+			if tc.expectedErr != nil {
+				s.Require().Error(err)
+				s.Require().ErrorIs(err, tc.expectedErr)
```
czarcas7ic, line 307: Can we change this to ErrorContains? ErrorIs has weird behavior that have cause false negatives in the past.

x/concentrated-liquidity/tick_test.go
```diff
@@ -811,3 +822,108 @@ func (s *KeeperTestSuite) TestValidateTickRangeIsValid() {
...
+			if len(test.expectedTicksOverwrite) > 0 {
+				expectedTicks = test.expectedTicksOverwrite
+			}
+
+			// System Under Test
+			ticks, err := s.App.ConcentratedLiquidityKeeper.GetAllInitializedTicksForPool(s.Ctx, defaultPoolId)
+
+			if test.expectedError != nil {
+				s.Require().Error(err)
+				s.Require().ErrorContains(err, test.expectedError.Error())
```
czarcas7ic, line 918: I don't see any error cases. Can we either add error cases or remove this conditional?

czarcas7ic (approved):
🫡

x/concentrated-liquidity/store_test.go
p0mvn: What were the problems?

`ErrorIs` is the correct way in this case because we want to check the presence of the `tc.expectedErr` in `err`'s chain.

`ErrorContains` is less preferred because it requires hardcoding an error as a string. So that if an error message changes, we have to update all tests as well. On the other hand, with `ErrorIs` check, the tests can continue working despite a change in the message

x/concentrated-liquidity/store_test.go
```diff
@@ -184,3 +205,113 @@ func (s *KeeperTestSuite) TestParseFullPositionFromBytes() {
...
+	var (
+		cdc = s.App.AppCodec()
+
+		formatFullKey = func(tickPrefix []byte, poolIdBytes []byte, tickIndexBytes []byte) []byte {
+			key := make([]byte, 0)
+			key = append(key, tickPrefix...)
+			key = append(key, poolIdBytes...)
+			key = append(key, tickIndexBytes...)
+			return key
+		}
```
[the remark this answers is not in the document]
p0mvn: I would prefer not to move it outside of the test because there are helpers in `types` that allow to format a position's key correctly such as [`types.KeyTick`](https://github.com/osmosis-labs/osmosis/blob/cb7fc937c4bd9262fa3d637d139f9cf94b8c1565/x/concentrated-liquidity/types/keys.go#L73-L78)

The reason why we have this `formatFullKey` helper here is that I want to be testing injecting the wrong prefix. There should not be a need for that outside of these tests

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/CosmWasm/wasmd v0.30.0
 	github.com/cosmos/cosmos-proto v1.0.0-alpha8
-	github.com/cosmos/cosmos-sdk v0.46.11
+	github.com/cosmos/cosmos-sdk v0.47.0
```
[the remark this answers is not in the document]
p0mvn: This is changed automatically when a new version is created. We use a replace directive so this update has no effect

x/concentrated-liquidity/tick_test.go
p0mvn: I remove the error condition because it is non-trivial to trigger an error case for this function.

Given how extensively the underlying helpers are tested, I think the effort to trigger an error case here is not justified. Let me know if any objections

x/concentrated-liquidity/store_test.go
czarcas7ic: The problem was that the ErrorIs check was passing despite the error being a completely different error. This has occurred for me on more than one occasion. I can specifically point it out the next time I occur it.

p0mvn: This might happen if slices are used as error fields. Since slices reference different buffers under the hood, the equivalence check would fail. From my experience, this frequently happens with byte slices, which I would convert to strings instead.

AlpinYukseloglu (approved):

AlpinYukseloglu:
High level logic LGTM

x/concentrated-liquidity/store_test.go
p0mvn: Going to merge but happy to continue this discussion separately and revisit if needed

Issue #4517, opened 2023-03-06
[CL Incentives] Scale gas cost for new incentives by number of existing incentives


AlpinYukseloglu:
## Background

The current implementation of incentive creation allows for a DoS vector where a user can spam new incentives with dust amounts. Since our reward calculation is linear in the number of incentive records, this exposes the chain to being forced to run unbounded compute.

To combat this, we should scale gas costs for creating incentives by the number of existing incentive records.

## Suggested Design

- I propose we add a base gas fee amount and charge `baseGas * len(existingIncentiveRecordsForUptime)` on each incentive record add
- Implement getting incentive records by uptime (move up minUptime in key and make function)
- Add the following constant to `concentrated-liquidity/types/constants.go`:
```
BaseGasFeeForNewIncentive = 10_000
```
- Add the following logic right before `setIncentiveRecord` in `createIncentive`:
```
// Get all incentive records for uptime
existingRecordsForUptime, err := k.GetAllIncentiveRecordsForUptime(ctx, poolId, minUptime)

// Fixed gas consumption per swap to prevent spam
ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForNewIncentive) * len(existingRecordsForUptime), "balancer swap computation")
```

## Acceptance Criteria

- All new and existing tests pass

PR #4638, opened 2023-03-16
[CL Incentives]: Scale gas cost for new incentives by number of existing incentives

AlpinYukseloglu:
Closes: #4517

This PR implements scaling gas costs for creating incentives records.

## Brief Changelog

- Charge additional gas costs for new incentive records with the following equation: (# existing records for uptime) * (some base amount). This makes it so that the first record has no additional cost and the rest scale linearly.
- Implement two helpers: `GetAllIncentiveRecordsByUptime` and `findUptimeIndex`
- Replace `minUptime` in incentive record key with `uptimeIndex` and move up to right after the `poolId`. This was to aid with implementing `GetAllIncentiveRecordsByUptime` and in general because using `minUptime` directly causes edge cases when gathering values from store.

## Testing and Verifying

- All new logic is tested in `incentives_test.go`


x/concentrated-liquidity/incentives.go
```diff
@@ -592,8 +645,20 @@ func (k Keeper) createIncentive(ctx sdk.Context, poolId uint64, sender sdk.AccAd
...
 	}
 
+	// Get all incentive records for uptime
+	existingRecordsForUptime, err := k.GetAllIncentiveRecordsForUptime(ctx, poolId, minUptime)
+	if err != nil {
+		return types.IncentiveRecord{}, err
+	}
+
+	// Fixed gas consumption per swap to prevent spam
+	ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForNewIncentive*len(existingRecordsForUptime)), "balancer swap computation")
```
p0mvn, line 655:
```suggestion
	ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForNewIncentive*len(existingRecordsForUptime)), "cl incentive creation")
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -2944,6 +2977,10 @@ func (s *KeeperTestSuite) TestCreateIncentive() {
 			s.Require().Equal(tc.recordToSet, recordInState)
 			s.Require().Equal(tc.recordToSet, incentiveRecord)
 
+			// Ensure that at least the minimum amount of gas was charged (based on number of existing incentives for current uptime)
+			gasConsumed := s.Ctx.GasMeter().GasConsumed() - existingGasConsumed
+			s.Require().True(gasConsumed >= tc.minimumGasConsumed)
```
p0mvn, line 2982: Any reason you chose >= over an equality check?

p0mvn (approved):
LGTM nice work. 

I focused on logic and that test cases aim to sufficiently cover the edge cases but did not review the test boilerplate in detail

x/concentrated-liquidity/incentives.go
```diff
@@ -592,8 +645,20 @@ func (k Keeper) createIncentive(ctx sdk.Context, poolId uint64, sender sdk.AccAd
...
+
```
czarcas7ic, line 656: On the first record, will this consume zero gas *total* or zero gas *extra*?

stackman27 (commented):
overall LGTM, just wanted to clarify something before approving!

czarcas7ic (approved):

x/concentrated-liquidity/incentives.go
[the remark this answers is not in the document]
stackman27: Resolved offline. Maybe we can add this comment here; 
```
we charge for # of uptime records of all kinds, so it implicitly covers DoS subvectors for any specific field
e.g. different start times, different emission rates, different remaining amounts etc.
All implicitly covered since the bottleneck is the specific uptime we’re scaling the gas cost on
```

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Zero gas extra, as `ConsumeGas` is purely additive to total consumed gas

x/concentrated-liquidity/incentives.go
```diff
@@ -230,32 +235,63 @@ func calcAccruedIncentivesForAccum(ctx sdk.Context, accumUptime time.Duration, q
 	return incentivesToAddToCurAccum, poolIncentiveRecords, nil
 }
 
+// findUptimeIndex finds the uptime index for the passed in min uptime.
+// Returns error if uptime index cannot be found.
+func findUptimeIndex(uptime time.Duration) (int, error) {
+	index := slices.IndexFunc(types.SupportedUptimes, func(e time.Duration) bool { return e == uptime })
```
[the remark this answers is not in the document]
AlpinYukseloglu: Thanks for double checking – I had done a sanity check but good to have another pair of eyes on it.

x/concentrated-liquidity/types/keys.go
```diff
@@ -90,11 +90,15 @@ func KeyPool(poolId uint64) []byte {
 	return []byte(fmt.Sprintf("%s%d", PoolPrefix, poolId))
 }
 
-func KeyIncentiveRecord(poolId uint64, denom string, minUptime time.Duration, addr sdk.AccAddress) []byte {
+func KeyIncentiveRecord(poolId uint64, minUptimeIndex int, denom string, addr sdk.AccAddress) []byte {
 	addrKey := address.MustLengthPrefix(addr.Bytes())
-	return []byte(fmt.Sprintf("%s%s%d%s%s%s%d%s%s", IncentivePrefix, KeySeparator, poolId, KeySeparator, denom, KeySeparator, uint64(minUptime), KeySeparator, addrKey))
+	return []byte(fmt.Sprintf("%s%s%d%s%d%s%s%s%s", IncentivePrefix, KeySeparator, poolId, KeySeparator, minUptimeIndex, KeySeparator, denom, KeySeparator, addrKey))
```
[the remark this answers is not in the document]
AlpinYukseloglu: Mainly since when we serialize minUptimeValue into bytes, the store iterator seems to not be able to tell the difference between the full number and its first two bytes (as a toy example to demonstrate, it might only read the first two characters of `10,000` and therefore conclude that `110 > 10,000` since `11 > 10`).

Using the index lets us get around this because we are guaranteed to have a small number of supported indices (specifically, a lower number than what we can fit into two bytes). It's likely possible to do a workaround to make it work with `minUptimeValue`, but this approach seemed both simpler and in line with our convention of using IDs in keys.

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Yes – there seems to be varying amounts of gas charged based on what the underlying function does, so it seemed cleaner to do a >= check and just ensure that the test cases eclipse any dust amount of gas that's charged instead of hard-coding how much gas is currently being charged by each call.

Issue #4649, opened 2023-03-17
[CL][Key Refactor]: Append id to the end of the existing position key


p0mvn:
## Background

As part of an incremental refactor toward a final design using only positions, we would like to append the id to the end of the existing position key

## Suggested Design

Modify the existing position key

## Acceptance Criteria
- the key is modified
- tests are updated where needed

p0mvn:
@czarcas7ic 

Realized that we should do this first for position id only: https://github.com/osmosis-labs/osmosis/issues/4626

Please make sure it's done before starting this as it will simplify future steps.

Issue #4647, opened 2023-03-17
[CL] Refactor positions state to key by id


p0mvn:
## Background

Currently, we have the following structure of the position key:
https://github.com/osmosis-labs/osmosis/blob/c9a496ab0e0d374c3021d6ab6e9e7393e414b4fb/x/concentrated-liquidity/types/keys.go#L112-L116

We would like to switch to ID management and position keys in the state.

This will lead us to having additional store indexes for:
- `pool id -> position ids`
- `owner | pool id | lower tick | upper tick -> position ids`
- `position id  -> position struct`

This would address all of the following problems:
- Mars contracts needing to have position id to claim fees on behalf of users
- Having a single owner in key breaks this ability
- Queries 
   - for positions by user
   - for positions by pool id
   - Problem: having the current key structure does not allow us to have one or the other query as iteration depends on key order
- Future NFT contract becomes trivial to implement
    - With the current design, there is a problem in that when we transfer a position via an external NFT contract, internally on-chain we still have the position belonging to the original owner
- Remove freeze id as a parameter
    - This refactor will also simplify the ability to remove freeze id as a parameter

## Suggested Design

```[tasklist]
### Tasks
- [ ] #4648
- [ ] #4649
- [ ] #4650
- [ ] #4651
```

## Acceptance Criteria
- positions are keyed by IDs in state

czarcas7ic:
All subtasks complete! Closing

Issue #4626, opened 2023-03-14
fix(CL): avoid store key formatting with `fmt.Sprintf`


p0mvn:
## Background

Using `fmt.Sprintf` to format storage keys may not be the best approach for several reasons:

- Performance: `fmt.Sprintf` is slower than other key formatting alternatives, such as using `strings.Builder` or simple string concatenation. When working with storage keys that require high performance, such as in databases or caching systems, using `fmt.Sprintf` may introduce unnecessary overhead.

- Readability: When formatting complex storage keys, using `fmt.Sprintf` can lead to less readable code due to the use of formatting specifiers. It might be harder to understand the final key structure, which can be problematic when debugging or maintaining the code.

- Error-prone: It's easy to make mistakes when using `fmt.Sprintf` with multiple formatting specifiers, such as providing the wrong data types or an incorrect number of arguments. This can result in subtle bugs that are difficult to identify and fix.

- Inefficient memory usage: `fmt.Sprintf` creates a new string object, which can lead to increased memory allocations and garbage collection pressure in performance-critical applications.

Instead, consider using alternative methods for formatting storage keys:

Simple string concatenation: If you are working with a small number of strings, using the + operator for concatenation is a more efficient and straightforward option:
```go
key := "user:" + userID + ":email"
```

strings.Builder: For more complex key structures or when concatenating a larger number of strings, using strings.Builder is an efficient and readable approach:
```go
var builder strings.Builder
builder.WriteString("user:")
builder.WriteString(userID)
builder.WriteString(":email")
key := builder.String()
```

By choosing a more suitable method for formatting storage keys, you can improve the performance, readability, and maintainability of your code.

## Suggested Design

Follow this format:
https://github.com/osmosis-labs/osmosis/blob/222d66bdf2dd8277cb81505e1a0120619619c8df/x/concentrated-liquidity/types/keys.go#L63-L65

To further improve the above, consider pre-allocating the byte buffer

## Acceptance Criteria
- key format in the concentrated liquidity module is refactored
- efficient approaches sed
- byte buffer pre-allocated

p0mvn:
We also format addresses as bytes that have 2x amplification because `fmt.Sprintf` uses 2 characters for every byte, leading to 2 bytes for every byte in the original address.

DongLieu:
can i handle it?

PR #4661, opened 2023-03-18
feat[CL]: utilize position id in position key

czarcas7ic:
Closes: #4649 

This PR utilizing the recently added position ID in the `FullPosition` key. This is part of a larger refactor, which is tracked in https://github.com/osmosis-labs/osmosis/issues/4647.

## Brief Changelog

- Adds position ID to `Position` proto struct
- Adds position ID to `MsgWithdrawPosition`
- Initializes `NextPositionId` to 1 if `NextPositionId` is not set in genesis
- `CreatePosition` (and therefore position migration) now returns position Id

## Testing and Verifying

Tests updated to reflect new parameter


x/concentrated-liquidity/types/keys.go
```diff
@@ -113,10 +116,30 @@ func keyTickPrefixByPoolIdPrealloc(poolId uint64, preAllocBytes int) []byte {
...
 }
 
-// KeyFullPosition uses pool Id, owner, lower tick, upper tick, joinTime and freezeDuration for keys
-func KeyFullPosition(poolId uint64, addr sdk.AccAddress, lowerTick, upperTick int64, joinTime time.Time, freezeDuration time.Duration) []byte {
+// KeyFullPosition uses pool Id, owner, lower tick, upper tick, joinTime, freezeDuration, and positionId for keys
+func KeyFullPosition(poolId uint64, addr sdk.AccAddress, lowerTick, upperTick int64, joinTime time.Time, freezeDuration time.Duration, positionId uint64) []byte {
 	joinTimeKey := osmoutils.FormatTimeString(joinTime)
-	return []byte(fmt.Sprintf("%s%s%x%s%d%s%d%s%d%s%s%s%d", PositionPrefix, KeySeparator, addr.Bytes(), KeySeparator, poolId, KeySeparator, lowerTick, KeySeparator, upperTick, KeySeparator, joinTimeKey, KeySeparator, uint64(freezeDuration)))
+
+	var builder strings.Builder
```
p0mvn, line 123: Consider using `bytes.Buffer` instead of `strings.Builder`.

When you need to serialize data consisting of strings, byte slices, integers, and time in Go, it is more efficient to use bytes.Buffer. The reason is that bytes.Buffer can handle both strings and byte slices directly, whereas strings.Builder is designed to work only with strings.

Using bytes.Buffer allows you to avoid unnecessary conversions between strings and byte slices, which can lead to better performance and reduced memory overhead.

x/concentrated-liquidity/types/keys.go
```diff
@@ -113,10 +116,30 @@ func keyTickPrefixByPoolIdPrealloc(poolId uint64, preAllocBytes int) []byte {
...
+	builder.Grow(len(PositionPrefix) + 9*len(KeySeparator) + len(addr.Bytes())*2 + 5*20) // pre-allocating the buffer
+
+	builder.Write(PositionPrefix)
+	builder.WriteString(KeySeparator)
+	builder.WriteString(hex.EncodeToString(addr.Bytes()))
+	builder.WriteString(KeySeparator)
+	builder.WriteString(strconv.FormatUint(poolId, 10))
```
p0mvn, line 130: This is how we can serialize integers in a space-efficient way per my earlier `bytes.Buffer` suggestion:

```go
	// Create a buffer to store the bytes
	var buf bytes.Buffer

	// Serialize the uint64 value using PutUvarint
	n := binary.PutUvarint(buf.Bytes()[:binary.MaxVarintLen64], value)

	// Truncate the buffer to the actual number of bytes used by the encoded value
	buf.Truncate(n)
```

x/concentrated-liquidity/types/keys.go
p0mvn: Also, I know my issue https://github.com/osmosis-labs/osmosis/issues/4626 suggested `strings.Builder` but I thought about this only upon seeing this change so apologies for re-requesting a change I earlier suggested

x/concentrated-liquidity/keeper_test.go
```diff
@@ -111,13 +111,21 @@ func (s *KeeperTestSuite) SetupOverlappingRangePositionAcc(poolId uint64, owner
...
 	s.Require().True(newPositionLiquidity.GTE(sdk.ZeroDec()))
 }
 
+// validatePositionsUpdate validates that positions with the given parameters have expectedRemainingLiquidity left.
+func (s *KeeperTestSuite) validatePositionsUpdate(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, lowerTick int64, upperTick int64, expectedRemainingLiquidity sdk.Dec) {
+	newPositionsLiquidity, err := s.App.ConcentratedLiquidityKeeper.GetPositionsLiquidity(ctx, poolId, owner, lowerTick, upperTick)
+	s.Require().NoError(err)
+	s.Require().Equal(expectedRemainingLiquidity.String(), newPositionsLiquidity.String())
+	s.Require().True(newPositionsLiquidity.GTE(sdk.ZeroDec()))
+}
```
p0mvn, line 127: Why do we need this method and not use the `validatePositionUpdate` by providing an id?

x/concentrated-liquidity/position.go
```diff
@@ -95,6 +97,21 @@ func (k Keeper) GetPositionLiquidity(ctx sdk.Context, poolId uint64, owner sdk.A
...
+		return sdk.Dec{}, err
+	}
+
+	totalLiquidity := sdk.ZeroDec()
+	for _, liquidity := range liquidityArray {
+		totalLiquidity = totalLiquidity.Add(liquidity)
+	}
+
+	return totalLiquidity, nil
+}
```
p0mvn, line 113: I don't think we should be adding this method to be used in tests only.

Since we treat positions with different IDs as separate, there should be no need for knowing liquidity across all positions within the same tick range.

Additionally, at the moment, it seems to me that the test helper `validatePositionsUpdate` where this function is called should be removed, and `validatePositionUpdate` used instead.

x/concentrated-liquidity/lp_test.go
```diff
@@ -276,11 +280,11 @@ func (s *KeeperTestSuite) TestCreatePosition() {
 			s.Require().Equal(userBalancePrePositionCreation.Sub(sdk.NewCoins(sdk.NewCoin(ETH, asset0), (sdk.NewCoin(USDC, asset1)))).String(), userBalancePostPositionCreation.String())
 			s.Require().Equal(poolBalancePrePositionCreation.Add(sdk.NewCoin(ETH, asset0), (sdk.NewCoin(USDC, asset1))).String(), poolBalancePostPositionCreation.String())
 
-			hasPosition := clKeeper.HasFullPosition(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, defaultJoinTime, tc.freezeDuration)
+			hasPosition := clKeeper.HasFullPosition(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, defaultJoinTime, tc.freezeDuration, tc.positionId)
 			s.Require().True(hasPosition)
 
 			// Check position state
-			s.validatePositionUpdate(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, defaultJoinTime, tc.freezeDuration, tc.liquidityAmount)
+			s.validatePositionsUpdate(s.Ctx, tc.poolId, s.TestAccs[0], tc.lowerTick, tc.upperTick, tc.liquidityAmount)
```
p0mvn, line 287: Why can't we continue using `validatePositionUpdate`? It looks like this change has led to introducing a getter in production code

p0mvn (commented):
Great work. Overall, LGTM.

I would like to discuss the requested key changes and the new GetPositionsLiqudity getter before approving

x/concentrated-liquidity/genesis.go
```diff
@@ -13,6 +13,10 @@ import (
 // InitGenesis initializes the concentrated-liquidity module with the provided genesis state.
 func (k Keeper) InitGenesis(ctx sdk.Context, genState genesis.GenesisState) {
 	k.SetParams(ctx, genState.Params)
+	// Initialize next position id to 1 if it is not set.
+	if genState.NextPositionId == 0 {
+		genState.NextPositionId = 1
+	}
```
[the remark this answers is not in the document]
czarcas7ic: Idk why I didn't do this to begin with, thanks 😅

https://github.com/osmosis-labs/osmosis/pull/4661/commits/824bd2f6e26bbaa0d4078de9c55a07f4564e338a

x/concentrated-liquidity/types/keys.go
```diff
@@ -113,10 +116,30 @@ func keyTickPrefixByPoolIdPrealloc(poolId uint64, preAllocBytes int) []byte {
...
+func KeyFullPosition(poolId uint64, addr sdk.AccAddress, lowerTick, upperTick int64, joinTime time.Time, freezeDuration time.Duration, positionId uint64) []byte {
 	joinTimeKey := osmoutils.FormatTimeString(joinTime)
-	return []byte(fmt.Sprintf("%s%s%x%s%d%s%d%s%d%s%s%s%d", PositionPrefix, KeySeparator, addr.Bytes(), KeySeparator, poolId, KeySeparator, lowerTick, KeySeparator, upperTick, KeySeparator, joinTimeKey, KeySeparator, uint64(freezeDuration)))
+
+	var builder strings.Builder
+	builder.Grow(len(PositionPrefix) + 9*len(KeySeparator) + len(addr.Bytes())*2 + 5*20) // pre-allocating the buffer
+
+	builder.Write(PositionPrefix)
+	builder.WriteString(KeySeparator)
+	builder.WriteString(hex.EncodeToString(addr.Bytes()))
```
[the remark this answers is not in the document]
czarcas7ic: Thanks for explaining this in detail!

x/concentrated-liquidity/lp_test.go
czarcas7ic: the test case `liquidityAmount` was combining the position 1 liquidity and position 2 liquidity because we previously merged this into one single position. Because of this, we validate all positions with the same key (minus ID). The alternative would be to just validate the second position, but this feels wrong.

Also, what is wrong with have a getter in production code?

czarcas7ic: I guess what I could do is just each position individually for the test case that has two positions, but just felt less clean

tests/e2e/e2e_test.go
```diff
@@ -288,35 +288,35 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 	// Assert removing some liquidity
 	// address1: check removing some amount of liquidity
 	address1position1liquidityBefore := positionsAddress1[0].Position.Liquidity
-	node.WithdrawPosition(address1, "[-1200]", "400", defaultLiquidityRemoval, poolID, positionsAddress1[0].Position.JoinTime, positionsAddress1[0].Position.FreezeDuration.String())
+	node.WithdrawPosition(address1, "[-1200]", "400", defaultLiquidityRemoval, poolID, positionsAddress1[0].Position.JoinTime, positionsAddress1[0].Position.FreezeDuration.String(), 1)
```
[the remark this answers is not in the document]
czarcas7ic: Done here https://github.com/osmosis-labs/osmosis/pull/4661/commits/5214e1d42b0b699f377315897e8d25b43234f025

x/concentrated-liquidity/keeper_test.go
czarcas7ic: Removed logic here https://github.com/osmosis-labs/osmosis/pull/4661/commits/c906adae1baa2cf19a090ec602b4e8e61fa6db9d

x/concentrated-liquidity/lp_test.go
czarcas7ic: I ended up just checking the second position created for that single test case https://github.com/osmosis-labs/osmosis/pull/4661/commits/c906adae1baa2cf19a090ec602b4e8e61fa6db9d

p0mvn:
Can we merge this? This is blocking some of the events work I'm working on for the data team

x/concentrated-liquidity/lp.go
```diff
@@ -56,54 +56,53 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
...
 	if len(positions) == 0 {
 		if err := k.initializeFeeAccumulatorPosition(cacheCtx, poolId, owner, lowerTick, upperTick); err != nil {
-			return sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
 		}
 	}
 
-	// TODO: Utilize the position ID returned from this function in the position key.
-	k.getNextPositionIdAndIncrement(ctx)
+	positionId := k.getNextPositionIdAndIncrement(ctx)
```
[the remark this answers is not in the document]
czarcas7ic: Correct, in default genesis!

x/concentrated-liquidity/incentives_test.go
```diff
@@ -1821,7 +1821,7 @@ func (s *KeeperTestSuite) TestInitOrUpdatePositionUptime() {
 
 			// --- System under test ---
 
-			err := s.App.ConcentratedLiquidityKeeper.InitOrUpdatePositionUptime(s.Ctx, clPool.GetId(), test.positionLiquidity, s.TestAccs[0], test.lowerTick.tickIndex, test.upperTick.tickIndex, test.positionLiquidity, DefaultJoinTime, DefaultFreezeDuration)
+			err := s.App.ConcentratedLiquidityKeeper.InitOrUpdatePositionUptime(s.Ctx, clPool.GetId(), test.positionLiquidity, s.TestAccs[0], test.lowerTick.tickIndex, test.upperTick.tickIndex, test.positionLiquidity, DefaultJoinTime, DefaultFreezeDuration, 1)
```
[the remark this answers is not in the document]
czarcas7ic: Funny thing is I actually this set as a var, but the idea of a "default" unique position ID felt weird... (its not like a param) so I removed it.

We can re-engage this after some of the following changes

Issue #4660, opened 2023-03-17
fix: go mod auto-update to commit to custom branch when dependency module is updated


p0mvn:
## Background

In the following PR, I made changes to `osmoutils`:
https://github.com/osmosis-labs/osmosis/pull/4659

The autoupdate job ran successfully but did not make a commit to my branch:
https://github.com/osmosis-labs/osmosis/actions/runs/4452408410/jobs/7820019622?pr=4659

## Suggested Design

The CI should auto-commit to my branch when I change one of the dependency modules in the repository

## Acceptance Criteria
- auto-commits to custom branch

p0mvn:
@pysel are you available to look into this please?

p0mvn:
Also, noticed the following problem:

![image](https://user-images.githubusercontent.com/34196718/226152205-0f778d72-7579-4228-a307-4ce4621b1695.png)

When the bot commits, the CI is not run so you have to close the PR and reopen it. We should make sure that closing-reopening isn't necessary

p0mvn:
Additionally, the bot seems to be comitting at the wrong time breaking the PR: 

[40986dd](https://github.com/osmosis-labs/osmosis/pull/4659/commits/40986ddaec8410d191509d128faf3fd5d673eaa5)

Relevant PR: https://github.com/osmosis-labs/osmosis/pull/4659

pysel:
> Additionally, the bot seems to be comitting at the wrong time breaking the PR:
> 
> [40986dd](https://github.com/osmosis-labs/osmosis/pull/4659/commits/40986ddaec8410d191509d128faf3fd5d673eaa5)
> 
> Relevant PR: #4659

hey @p0mvn, could you please elaborate on this? I do not really get this

p0mvn:
So I manually fixed `go.mod` in my PR #4659, and the bot committed over it with the wrong change. The commit that broke it is linked

The way I think it currently works is that when a commit to any branch is made within a repository, this workflow runs.

Therefore, when I merge a PR that contains `osmoutils` changes to `main`, this action will make a commit to all open PRs

This is not desired.

The goal of this workflow is to commit directly to your custom branch when you update `osmoutils` or any other dependency module.

See our discussion here: https://github.com/osmosis-labs/osmosis/pull/4666#discussion_r1141407194

pysel:
[...]

PR #4666, opened 2023-03-19
ci: go mod auto-update runs on push to curr

p0mvn:
Closes: #4660

Currently, `go.mod` auto-update runs on push to every branch.

Instead, it should be running on push to the current branch.

## Testing and Verifying

### DIFF
--- a/.github/workflows/go-mod-auto-bump.yml
+++ b/.github/workflows/go-mod-auto-bump.yml
@@ -8,6 +8,8 @@
 name: Go Mod Auto Version Update
 on: 
   push:
+    branches:
+    - ${{ github.ref }}
     paths: 
       - osmoutils/**
       - osmomath/**


.github/workflows/go-mod-auto-bump.yml
```diff
@@ -8,6 +8,8 @@
 name: Go Mod Auto Version Update
 on: 
   push:
+    branches:
+    - ${{ github.ref }}
```
pysel, line 12: I think with this syntax the workflow will run only when the push to `github.ref` is made. However, this variable is only accessible when the workflow is running, so, I suspect, that when GH Actions check if the condition for this workflow is met, it will see `github.ref` as a blank string, because as I said, this variable is only accessible when workflow is running. I will check if that is indeed like that.

Even if my suspicion is wrong, I do not really see why we are changing this. I do not see what is wrong with the "push to any branch" condition. Basically, the logic of this workflow is: when any branch `X` is changed, update go mod on this branch, and push the changes to `X` (aka `github.ref`)

pysel: if I am right, I will check out other ways of how this can be done properly

pysel: yes, this workflow will never run with these changes

p0mvn: Why is that a problem?

This should be checked once a push is made to the current branch. A push triggers the workflow run where `github.ref` is accessible

p0mvn: I understand now. ChatGPT was confusing me here:

image

PR #4667, opened 2023-03-19
ci: manual job for updating in-repo go mods; ci to warn if files changed

pysel:
Closes: #4660 

## Description 

This PR introduces a manual CI job to update `go.mod` for `osmomath`, `osmoutils`, `x/ibc-hooks` and `x/epochs` if changed.

The changes introduced in this PR include:

* Manual CI that runs a bash script for updating go mods
* CI that comments in PRs with instructions for running if files are updated

Tested manually on a fork


.github/workflows/go-mod-auto-bump.yml
```diff
@@ -33,14 +33,33 @@ jobs:
         run: |
           echo "CURRENT_BRANCH=$(echo ${GITHUB_REF} | sed 's|refs/heads/||')" >> $GITHUB_OUTPUT
 
-      - name: Run script 
+      # This job fixes a bug, due to which the push to any branch, ex: person-A/name, triggered this workflow on branch person-b/name,
```
pysel, line 36:
```suggestion
      # This job fixes a bug, due to which the push to any branch, for example: to person-A/name, triggered this workflow on branch person-b/name,
```

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -33,14 +33,33 @@ jobs:
...
+      # and updated the go.mod on branch person-A/name with commit from branch person-b/name
```
pysel, line 37:
```suggestion
      # and updated the go.mod on branch person-B/name with commit from branch person-A/name
```

pysel:
hey @p0mvn. A quick update on this:

There is a problem with this workflow when a push with changes is a `force` push. Sometimes, if a force push removes the latest commit in PR, and sets the newly pushed commit as the latest one, this github workflow will fail, because `github.event.before` (which is used to find the commit before the push), will point to the commit that was deleted when force-pushing

We could solve that by giving the ability to the developer to run this workflow manually, however, I think this completely ruins the point of this automation.

Another idea is to run the updating script ALWAYS when a `force` push happens, updating all 4 go.mod folders, but I am also not a fan of this idea.

The third idea I have, is to change the `commit before` in this workflow to be the latest commit on `main` branch. This will remove the need to figure out ways to deal with force pushes, and it seems to me that it will also be logically fine.

wdyt?

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -33,14 +33,29 @@ jobs:
         run: |
           echo "CURRENT_BRANCH=$(echo ${GITHUB_REF} | sed 's|refs/heads/||')" >> $GITHUB_OUTPUT
 
-      - name: Run script 
+      # This step fixes a bug, due to which the push to any branch, for example: to person-A/name, triggered this workflow on branch person-b/name,
```
p0mvn, line 36: nit: can we put this before "Setup Go"

.github/workflows/go-mod-auto-bump.yml
```diff
@@ -33,14 +33,29 @@ jobs:
...
+          CURRENT_BRANCH=${{ steps.extract_branch.outputs.CURRENT_BRANCH }}
+          COMMIT_EXISTS_ON=$(echo $(git branch --contains ${{ github.event.after}} ) | grep $CURRENT_BRANCH)
+
+          if [[ "$COMMIT_EXISTS_ON" == *"$CURRENT_BRANCH"* ]];
+          then
+            echo "VALID=1" >> $GITHUB_OUTPUT
+          fi
+          
+      - name: Run script
+        if: ${{ steps.valid_branch.outputs.VALID }} == 1
```
p0mvn, line 50: So assume I'm on branch `roman/A` and you are on `ruslan/B`

Then, I modify `osmoutils` on my branch. Then, this workflow will run for both `roman/A` and `ruslan/B`. However, it will not continue past this step on `ruslan/B`. Is my understanding correct?

I'm wondering if there is a way to not run this job on other branches at all to avoid having a queue in CI.

Let me know what you think

p0mvn:
> hey @p0mvn. A quick update on this:
> 
> There is a problem with this workflow when a push with changes is a `force` push. Sometimes, if a force push removes the latest commit in PR, and sets the newly pushed commit as the latest one, this github workflow will fail, because `github.event.before` (which is used to find the commit before the push), will point to the commit that was deleted when force-pushing
> 
> We could solve that by giving the ability to the developer to run this workflow manually, however, I think this completely ruins the point of this automation.
> 
> Another idea is to run the updating script ALWAYS when a `force` push happens, updating all 4 go.mod folders, but I am also not a fan of this idea.
> 
> The third idea I have, is to change the `commit before` in this workflow to be the latest commit on `main` branch. This will remove the need to figure out ways to deal with force pushes, and it seems to me that it will also be logically fine.
> 
> wdyt?

Thanks for the breakdown of the options.

> We could solve that by giving the ability to the developer to run this workflow manually

This makes sense to me as a temporary solution

> Another idea is to run the updating script ALWAYS when a `force` push happens, updating all 4 go.mod folders, but I am also not a fan of this idea.

What do you mean by all `go.mod` folders? Do you mean that If I make a change to `osmoutils`, it will update `go.mod` for all of `osmoutils`, `osmomath` etc?

This seems fine to me. It is actually preferable that you are always depending on the modules directly in your branch.

> The third idea I have, is to change the `commit before` in this workflow to be the latest commit on `main` branch. This will remove the need to figure out ways to deal with force pushes, and it seems to me that it will also be logically fine.

This is not what we would like to have. In this case, I can update `osmoutils` and have `e2e` CI fail because I'm not depending on the `osmoutils` change from my PR

pysel:
> What do you mean by all go.mod folders? Do you mean that If I make a change to osmoutils, it will update go.mod for all of osmoutils, osmomath etc?

Yes, but only when a force push happens, to avoid checking `git diff` with the latest commit sha, which could have been deleted by force pushing 

> This is not what we would like to have. In this case, I can update osmoutils and have e2e CI fail because I'm not depending on the osmoutils change from my PR

Hmm, I do not really understand this. `Commit After` will still be the latest commit on your local branch, by changing `commit before` to latest on `main`, we will basically check `git diff` between your latest branch state and the latest main state. 

In comparison, right now we are checking `git diff` between the `latest commit - 1` and `latest commit`, where `latest commit - 1` is `commit before` and `latest commit` is `commit after`

[This](https://github.com/osmosis-labs/osmosis/pull/4667/files#diff-0c9ab0ab8a9efd1037add345b0edbadd454fe1637ee9972fe4fd0c7798e749deR52-R53) is what I mean by `commit before` and `commit after`, by the way

CC: @p0mvn

.github/workflows/go-mod-auto-bump.yml
pysel: Yes, that is right. However, as mentioned [here](https://github.com/osmosis-labs/osmosis/issues/4660#issuecomment-1475526794), I haven't seen such behavior by changing, for example, `osmoutils` on my local branch. I checked the `Action` tab, which showed me, that this workflow was running only for my branch. 

A more detailed explanation you may find in the attached link above

pysel: let me know what you think on removing `main` from this workflow triggers

pysel:
I tested this workflow on my fork and locally in this branch. Seems fine:
* pushes to separate branches do not intervene with each other
* problem with force pushes solved
* CI does not block from the auto push
* PR merges to main do not trigger the workflow

CHANGELOG.md
```diff
@@ -123,6 +123,7 @@ and control over token transfers.
   * [#5356](https://github.com/osmosis-labs/osmosis/pull/5356) Fix wrong restHandler for ReplaceMigrationRecordsProposal
   * [#5020](https://github.com/osmosis-labs/osmosis/pull/5020) Add gas config to the client.toml
   * [#5105](https://github.com/osmosis-labs/osmosis/pull/5105) Lint stableswap in the same manner as all of Osmosis
+  * [#4667](https://github.com/osmosis-labs/osmosis/pull/4667) Adds fixed go.mod auto-update workflow
```
p0mvn, line 126:
```suggestion
```

p0mvn (approved):
@pysel great work on this.

I made some changes to make this manual instead of automatic:
https://github.com/osmosis-labs/osmosis/pull/4667/commits/e3dbdaa207a9b5608c22e71314833266b1337cd2

Also added a job that would comment where to manually run the CI whenever one of the modules updated.

Did manual test here: https://github.com/p0mvn/osmosis/pull/4

Let's merge and test this out

PR #4671, opened 2023-03-20
Use Go routines in CL go client

mattverse:
Closes: #XXX

Change existing process of creating positions in the go client for CLs to use go routines for faster performance. 

## Brief Changelog
- Use Go Routines in CL Go client for creating positions

## Testing and Verifying
Tested manually in local environment

### DIFF
--- a/tests/cl-go-client/main.go
+++ b/tests/cl-go-client/main.go
@@ -107,9 +107,8 @@ func main() {
 			amt0, amt1, liquidity := createPosition(igniteClient, expectedPoolId, accountName, lowerTick, upperTick, tokenDesired0, tokenDesired1, defaultMinAmount, defaultMinAmount)
 			log.Println("created position: amt0", amt0, "amt1", amt1, "liquidity", liquidity)
 		}(i)
-
-		wg.Wait()
 	}
+	wg.Wait()
 }
 
 func createPool(igniteClient cosmosclient.Client, accountName string) uint64 {
--- a/tests/cl-go-client/main.go
+++ b/tests/cl-go-client/main.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"math/rand"
 	"strings"
+	"sync"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
@@ -40,7 +41,6 @@ var (
 )
 
 func main() {
-
 	ctx := context.Background()
 
 	// Create a Cosmos igniteClient instance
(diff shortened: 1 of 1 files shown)


tests/cl-go-client/main.go
```diff
@@ -82,26 +82,33 @@ func main() {
...
+				tokenDesired0 = sdk.NewCoin(denom0, sdk.NewInt(rand.Int63n(maxAmountDeposited)))
+				tokenDesired1 = sdk.NewCoin(denom1, sdk.NewInt(rand.Int63n(maxAmountDeposited)))
+			)
+
+			log.Println("creating position: pool id", expectedPoolId, "accountName", accountName, "lowerTick", lowerTick, "upperTick", upperTick, "token0Desired", tokenDesired0, "tokenDesired1", tokenDesired1, "defaultMinAmount", defaultMinAmount)
+			amt0, amt1, liquidity := createPosition(igniteClient, expectedPoolId, accountName, lowerTick, upperTick, tokenDesired0, tokenDesired1, defaultMinAmount, defaultMinAmount)
+			log.Println("created position: amt0", amt0, "amt1", amt1, "liquidity", liquidity)
+		}(i)
+
+		wg.Wait()
```
p0mvn, line 111: I think `wg.Wait` should be outside of the loop for the concurrency benefits to become apparent. Otherwise, we are blocking every iteration here

mattverse: Yeah actually it should be outside the loop, thanks for the review, fixed in the most recent commit

mattverse:
@p0mvn Fixed, r4r

p0mvn:
Have you gotten a chance to test this?

I'm asking because when I tried something similar on my end, it would give me an account sequence mismatch.

I tried it with 1 account though whereas here we use 8 chosen randomly. However, if we get 2 same accounts in a row, I think it's still possible to get an account sequence mismatch

mattverse:
> Have you gotten a chance to test this?
> 
> I'm asking because when I tried something similar on my end, it would give me an account sequence mismatch.
> 
> I tried it with 1 account though whereas here we use 8 chosen randomly. However, if we get 2 same accounts in a row, I think it's still possible to get an account sequence mismatch

@p0mvn  Yes, I did get a chance to test this, that is because now we are sending transactions too fast. I was going to push the fix I made in a separate PR, but I would just converge it and put it in this PR as well.

Basically, the way I used out of this was to catch error and retry upon getting error(this makes the program retry upon getting error instead of exitting right away). Tested this way, confirmed that this is working correctly

tests/cl-go-client/main.go
```diff
@@ -104,8 +105,16 @@ func main() {
...
-			log.Println("created position: amt0", amt0, "amt1", amt1, "liquidity", liquidity)
+
+			maxRetries := 100
+			for j := 0; j < maxRetries; j++ {
+				amt0, amt1, liquidity := createPosition(igniteClient, expectedPoolId, accountName, lowerTick, upperTick, tokenDesired0, tokenDesired1, defaultMinAmount, defaultMinAmount)
+				if err == nil {
+					log.Println("created position: amt0", amt0, "amt1", amt1, "liquidity", liquidity)
+					break
+				}
+				time.Sleep(8 * time.Second)
```
p0mvn, line 116: Why 8 seconds in particular? Would it potentially run faster if the sleep was for fewer seconds?

p0mvn (approved):
LGTM, nice work.

Only had a non-blocking question

tests/cl-go-client/main.go
mattverse: No specific reason, it's close to the estimated block time.
Most of the reason why the program would fail would be because we would be having too many transactions in a single block and get sequence mismatch. Thus the original intent of sleeping 8 second was to have the process sleep for 8 second and retry in the next block

Issue #4654, opened 2023-03-17
Change environment feature does not respect custom home directory via --home flag


DavidNix:
#### System information

Osmosis version: v15.0.0
OS & Version: Linux
Commit hash: ff18d82

#### Expected behavior

`osmosisd init --home custom` persists all files to the directory specified by the `--home` flag.

#### Actual behaviour

`osmosisd init --home custom` always creates a directory at `$HOME/.osmosisd` in addition to any custom directory via `--home` flag. This behavior cannot be customized or circumvented. 

The addition of [ChangeEnvironmentCmd()](https://github.dev/osmosis-labs/osmosis/blob/c9a496ab0e0d374c3021d6ab6e9e7393e414b4fb/cmd/osmosisd/cmd/change_environment.go#L21) related to https://github.com/osmosis-labs/osmosis/pull/3787 specifically the function [CreateEnvFile](https://github.dev/osmosis-labs/osmosis/blob/c9a496ab0e0d374c3021d6ab6e9e7393e414b4fb/cmd/osmosisd/cmd/init.go#L188-L223) in `InitCmd` appears to be the root cause. 

#### Steps to reproduce the behaviour

On your local machine using v15.0.0. 

1. `make build`
1.  `ls ~/.osmosisd` will error as expected because it does not exist yet.
1. `./build/osmosisd init --home custom moniker`
1. `ls -a ~/.osmosisd` was created and has `.env` file. This is unexpected.

## Context

As a node operator, saving data anywhere outside of `--home` is unexpected, inconsistent behavior. In other words, the current behavior breaks [the Principle of Least Surprise](http://principles-wiki.net/principles:principle_of_least_surprise). 

We always set `--home` to a consistent directory because of these advantages:
* We have a consistent place to look for files. Some chains do not follow a predictable pattern for the chain, binary, and home folder name. A consistent home folder allows us to more easily build automation for inspecting files, for example.
* We do not create a hidden (.dot) file so it's more easily discoverable. (i.e. Don't have to remember to always `ls -a`. )

This seeming innocent changed caused several hours of downtime for our Osmosis RPC deployment because:

```
Error: Failed to create environment file: mkdir /home/operator/.osmosisd: permission denied
```

We strive for strict security on all our deployments.

[...]

PR #4676, opened 2023-03-20
Flag for saving environent

hieuvubk:
Closes: #4654 

Refer from [cmt](https://github.com/osmosis-labs/osmosis/issues/4654#issue-1629502654).

One limitation of using environment variable is that we have to save to `.omsosisd`. This can cause trouble for users who don't want to use this feature.

So I added a bool flag `init-env` to `InitCmd` to determine if the user wants to use the current directory as environment variable. If not, `home` flag is used normally and no `.env` file is created. Then it is possible for users to set environment variables with the `set-env` command

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


cmd/osmosisd/cmd/init.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// FlagSeed defines a flag to initialize the private validator key from a specific seed.
 	FlagRecover = "recover"
+
+	// FlagInitEnv defines a flag to create environment file & save current home directory into it.
+	FlagInitEnv = "init-env"
```
mattverse, line 41:
```suggestion
	FlagSetEnv = "set-env"
```

cmd/osmosisd/cmd/init.go
```diff
@@ -181,6 +187,7 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 	cmd.Flags().BoolP(FlagOverwrite, "o", false, "overwrite the genesis.json file")
 	cmd.Flags().Bool(FlagRecover, false, "provide seed phrase to recover existing key instead of creating")
 	cmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
+	cmd.Flags().Bool(FlagInitEnv, false, "create environment file & save current home directory into it")
```
mattverse, line 190:
```suggestion
	cmd.Flags().Bool(FlagSetEnv, false, "set and save current directory as home directory")
```

mattverse (commented):
tACK. Confirmed that this is working as expected via using binary built from current branch.

Leaving one suggestion for the flag name, since the flag name had me confused for quite a while on what this feature does. Lmk if you think the suggested change would make sense!

mattverse (approved):
Nice Work! LGTM

p0mvn (approved):
Thanks

PR #4691, opened 2023-03-21
(CL): Add query for Liquidity in Ranges

mattverse:
Closes: #XXX

Adds Query for Liquidity in Ranges. 
Iterates over all ticks that are initialized starting from min tick, until we hit the upper tick given. We cumulate the liquidity net in the process of iterating it, using the final result in the last iteration as the value for the liquidity.

## Brief Changelog
- Add query for Liquidity in Ranges
## Testing and Verifying

Added new tests.


x/concentrated-liquidity/tick.go
```diff
@@ -183,6 +184,86 @@ func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTic
...
+	// if the second tick initialized is greater than the given upper tick, return first tick's liquidity net
+	if nextTick.Int64() > upperTick {
+		return tick.LiquidityNet, nil
+	}
+	// use the smallest tick initialized as the starting point for calculating liquidity.
+	currentLiquidity := tick.LiquidityNet
+	currentTick = nextTick.Int64()
+
+	totalLiquidityWithinRange := currentLiquidity
+	for currentTick <= upperTick {
```
p0mvn, line 233: I think that instead of looping here and calling `swapStrategy.NextInitializedTick`, we can iterate the tick prefix for that pool using this:
https://github.com/osmosis-labs/osmosis/blob/f447c87e11d024b342dc8cb3db1e51d3f757105d/osmoutils/store_helper.go#L101-L108

I haven't looked deep enough into it so please proceed with caution on this suggestion. However, if we get the iterator to work, it should be considerably more performant than looping

x/concentrated-liquidity/tick.go
```diff
@@ -183,6 +184,125 @@ func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTic
 	return math.GetMinAndMaxTicksFromExponentAtPriceOneInternal(exponentAtPriceOne)
 }
 
+// GetTickLiquidityForRangeInBatches returns an array of liquidity depth within the given range of lower tick and upper tick.
+func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64, lowerTick, upperTick int64) (sdk.Dec, error) {
```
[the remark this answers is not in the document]
mattverse: Sweet, sounds great, revised this PR so that it suits the first query.

x/concentrated-liquidity/tick.go
mattverse: Would you mind if we create an issue this for further improvements? I have took a look at it, seems like this implementation would take more time than the current implmenetation, while I do not want to block front end for this query time wise. 

Some of the implementation I was working on: https://github.com/osmosis-labs/osmosis/compare/mattverse/use-iterator?expand=1 
Mainly more time consuming because we need to utilize both key + value used, and we need to parse the key as well to receive the tick index for the key, while we do not have enough helper methods to do this right away

x/concentrated-liquidity/tick.go
```diff
@@ -183,6 +184,85 @@ func GetMinAndMaxTicksFromExponentAtPriceOne(exponentAtPriceOne sdk.Int) (minTic
...
+		// break and return the liquidity as is if
+		// - there are no more next tick that is initialized,
+		// - we hit upper limit
+		if !ok {
+			break
+		}
+
+		keyTick := types.KeyTick(poolId, nextTick.Int64())
+		tickStruct := model.TickInfo{}
+		found, err := osmoutils.Get(store, keyTick, &tickStruct)
```
p0mvn, line 240: Can we use the `GetTickByTickIndex` here?

x/concentrated-liquidity/tick.go
```diff
@@ -226,3 +306,17 @@ func (k Keeper) GetPerTickLiquidityDepthFromRange(ctx sdk.Context, poolId uint64
 
 	return liquidityDepths, nil
 }
+
+func (k Keeper) GetTickByTickIndex(ctx sdk.Context, poolId uint64, tickIndex sdk.Int) (model.TickInfo, error) {
```
p0mvn, line 310: nit: consider unexporting

x/concentrated-liquidity/tick_test.go
```diff
@@ -601,6 +608,132 @@ func (s *KeeperTestSuite) TestCrossTick() {
...
+func (s *KeeperTestSuite) TestGetTickLiquidityForRange() {
+	defaultTick := withPoolId(defaultTick, defaultPoolId)
+
+	tests := []struct {
+		name        string
+		presetTicks []genesis.FullTick
+
+		expectedLiquidityDepthForRange []query.LiquidityDepthWithRange
+	}{
+		{
```
p0mvn, line 620: Can we add a test case with 2 full range and 2 narrow range positions?

I don't think there is one where we have both full range and narrow range

p0mvn (approved):
LGTM once new comments are addressed. Nice work!

Nothing blocking and agreed to get v1 in for frontend and work on v2 in the next PR

x/concentrated-liquidity/tick.go
p0mvn: Agreed and approved, I trust you'll either make an issue or PR a follow up change 👍

proto/osmosis/concentrated-liquidity/pool-model/query.proto
```diff
@@ -110,6 +117,33 @@ message LiquidityDepth {
...
+  string lower_tick = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"tick\"",
+    (gogoproto.nullable) = false
+  ];
+  string upper_tick = 3 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"tick\"",
+    (gogoproto.nullable) = false
+  ];
```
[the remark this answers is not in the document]
mattverse: good catch, addressed

x/concentrated-liquidity/grpc_query.go
```diff
@@ -161,6 +161,25 @@ func (q Querier) LiquidityDepthsForRange(goCtx context.Context, req *clquery.Que
 	}, nil
 }
 
+// TickLiquidityInBatches returns array of liquidity depths in the given range of lower tick and upper tick.
+// Note that the space between the ticks in the returned array would always be guaranteed spacing greater than given batch unit.
+func (q Querier) TotalLiquidityForRange(goCtx context.Context, req *clquery.QueryTotalLiquidityForRangeRequest) (*clquery.QueryTotalLiquidityForRangeResponse, error) {
```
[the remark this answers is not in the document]
mattverse: Godoc revised

x/concentrated-liquidity/tick.go
mattverse: Changed :)

x/concentrated-liquidity/tick.go
mattverse: unexported

x/concentrated-liquidity/tick_test.go
mattverse: added :)

proto/osmosis/concentrated-liquidity/pool-model/query.proto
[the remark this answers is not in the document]
mattverse: Actually, this should not be removed. This is a proto struct being used to return values for the response

PR #4683, opened 2023-03-20
XCS + Registries integration: Prefix map

nicolaslara:
Closes: #4439 #4441 

## What is the purpose of the change

Prepare the registries for integration with XCS

## Brief Changelog

 * Split the registries struct into its own package
 * Added a reverse map for prefix->[]chain
 * Refactor XCS and Registries to use the dependencies from the registry package

## Testing and Verifying

There are a few new tests, but mostly old tests should continue to pass. Later on some may need to be adapted

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #4694, opened 2023-03-21
XCS + Registries integration: IBC forward generation using registries

nicolaslara:

This PR is contains #4683. While that one only replaced the prefix registry, this one replaces the whole IBC message creation to support forwarding. 

This is still a draft at the moment. Some things missing:

 * Pre-swap unwrapping
 * How to deal with validation and user specified memos
 * Tests specific to multiple hops
 * More/better tests
    * All chains have the osmo prefix, which makes it hard to test all potential validation paths. Changing this has proven harder than expected, so how we test may depend on our ability to change those prefixes in the testing app.
 * Add a registry to check if a chain allows forwarding. 
 * Cleanup

## Brief Changelog

 * Contracts now use the registries to build the forward package
 * Lots of refactoring

## Testing and Verifying

Tests have been adapted to the new changes.


cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -414,7 +445,14 @@ pub fn chain_to_prefix_operations(
...
                     &chain_name,
-                    &(map_entry.value, true).into(),
+                    &(map_entry.value.clone(), true).into(),
+                )?;
+                // Add to the reverse map of the enabled prefix
+                push_to_map_value(
+                    deps.storage,
+                    &CHAIN_TO_BECH32_PREFIX_REVERSE_MAP,
+                    &map_entry.value,
+                    chain_name.clone(),
```
czarcas7ic, line 455: I don't think this is correct for the enable / disable logic right? We need to set the boolean entry of the tuple to true here

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -429,8 +467,16 @@ pub fn chain_to_prefix_operations(
...
                     &chain_name,
-                    &(map_entry.value, false).into(),
+                    &(map_entry.value.clone(), false).into(),
+                )?;
+                // Remove from the reverse map of the disabled prefix
+                remove_from_map_value(
+                    deps.storage,
+                    &CHAIN_TO_BECH32_PREFIX_REVERSE_MAP,
+                    &map_entry.value,
+                    chain_name.clone(),
```
czarcas7ic, line 477: Same comment here, but adding false

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -1278,18 +1324,121 @@ mod tests {
...
+                prefix: "OSMO".to_string(),
+                new_prefix: None,
+            }],
+        };
+        contract::execute(deps.as_mut(), mock_env(), info.clone(), msg).unwrap();
+        assert_eq!(
+            CHAIN_TO_BECH32_PREFIX_MAP
+                .load(&deps.storage, "osmosis")
+                .unwrap(),
+            ("osmo", false).into()
```
czarcas7ic, line 1385: Why is this set to false here?

cosmwasm/contracts/crosschain-swaps/src/checks.rs
```diff
@@ -31,46 +31,45 @@ fn validate_explicit_receiver(receiver: &str) -> Result<(String, Addr), Contract
...
-            receiver: receiver.to_string(),
-        });
-    };
 
     let Ok(_) = bech32::decode(&address) else {
         return Err(ContractError::InvalidReceiver { receiver: receiver.to_string() })
     };
 
-    Ok((channel.to_string(), Addr::unchecked(address)))
+    let registry = Registry::default(deps);
```
czarcas7ic, line 39: Did we decide how we want to set the address for the default value yet?

tests/ibc-hooks/ibc_middleware_test.go
```diff
@@ -454,6 +456,11 @@ func (suite *HooksTestSuite) RelayPacket(packet channeltypes.Packet, direction D
 	ack, err := ibctesting.ParseAckFromEvents(receiveResult.GetEvents())
 	suite.Require().NoError(err)
 
+	if strings.Contains(string(ack), "error") {
+		errorCtx := gjson.Get(receiveResult.Log, "0.events.#(type==ibc-acknowledgement-error)#.attributes.#(key==error-context)#.value")
+		fmt.Println("ibc-ack-error:", errorCtx)
```
czarcas7ic, line 461: Not sure if you meant to leave these printlns here (line 461, 513, and 654)

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: I have no concept of enable/disable on the reverse map. It's basically a representation of the current state, so when we enable the value, we "add it" to the reverse map; when we disable it, we remove it. Does that make sense?

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: Because the operation above is disable: https://github.com/osmosis-labs/osmosis/blob/edad03916cfab32aecabc6dfa6f1b34924b64cb0/cosmwasm/contracts/crosschain-registry/src/execute.rs#L1374

cosmwasm/contracts/crosschain-swaps/src/checks.rs
nicolaslara: decided, no. But I added it as a compilation flag in this PR. So you can compile with the env variable `REGISTRY_CONTRACT` and that will set it. With the optimizer:

```
docker run --rm -v "$(pwd)":/code -e REGISTRY_CONTRACT=osmo14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sq2r9g9 \
  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
  cosmwasm/workspace-optimizer-arm64:0.12.11
  ```

tests/ibc-hooks/ibc_middleware_test.go
nicolaslara: I did! they only are displayed when there are errors, and I found myself having to put a breakpoint there just to find that string value when debugging. If they're annoying we can probably control it with a global var in that file `DEBUG_PRINTS = true` or something

cosmwasm/contracts/crosschain-registry/src/execute.rs
czarcas7ic: Makes sense, thank you!

czarcas7ic (approved):
All questions have been addressed, LGTM! Should be good to merge once merge conflicts are fixed

nicolaslara:
I still have the items in the TODO that need to be addressed:

* Pre-swap unwrapping
* How to deal with validation and user specified memos
* Tests specific to multiple hops
* More/better tests
* All chains have the osmo prefix, which makes it hard to test all potential validation paths. Changing this has proven harder than expected, so how we test may depend on our ability to change those prefixes in the testing app.
* Add a registry to check if a chain allows forwarding.

But maybe we should merge this and add the rest as individual issues

Issue #4697, opened 2023-03-22
Remove go.mod workflow


pysel:
## Background

go.mod auto update action was added in #4291. However, it turned out to be in many ways buggy. We need to remove it for now, because it brings more inconvenience than if we didn't have it.

I am working on solving all the bugs in #4667, and we will add this workflow back when all the bugs are resolved

PR #4698, opened 2023-03-22
Remove auto go.mod workflow

pysel:
Closes: #4697 

Removes go.mod auto upgrade workflow


p0mvn (approved):
Thanks. Discussed offline and looking forward for readd once token issues are resolved

PR #4710, opened 2023-03-23
GitHub Action to check for broken links in .md files

pysel:
Closes: #1850 

Runs daily, creates an issue if finds broken links


p0mvn:
What's the difference from #4712 ?

pysel:
none, actually

Issue #4704, opened 2023-03-23
[CL]  Improve Performance of Liquidity Query via using iterators


mattverse:
## Background
In PR https://github.com/osmosis-labs/osmosis/pull/4691 and https://github.com/osmosis-labs/osmosis/pull/4714, we introduced a query to obtain the total liquidity across the entire pool. Currently, we iterate to the next pool using the NextInitializedTick method, which presents an opportunity to enhance performance. We aim to achieve this performance improvement by employing iterators and store helpers.

Unfortunately, we cannot directly use store helpers, as they do not provide an iteration method with a stop function that returns both key-value pairs. Moreover, using store helpers might not be the most efficient approach, as we need to use the poolId to parse the key and extract the tick values.

To address these concerns and optimize performance, we can explore two potential methods: creating a new store helper in osmoutils or utilizing iterators directly. Regardless of the chosen approach, our goal is to improve the performance of this query by leveraging iterators more effectively.

## Suggested Design

- Remove using `nextInitializedTick` and changing it to use iterators

PR #4714, opened 2023-03-23
(CL): Add Liquidity Net In Direction Query

mattverse:
Closes: #XXX

A query to better support front-end. Given ZeroForOne and the tick designated to iterate until(optional), iterates over ticks that are initialized until we meet the designated tick, and returns the tick indexes along with the liquidity net in each tick.

## Brief Changelog
- Add liquidity Net in Direction Query 

## Testing and Verifying

- Tests are added


x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,68 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
 	return liquidityDepthsForRange, nil
 }
 
+// GetLiquidityNetInDirection returns an array of liquidity depth in the given zeroForOne direction.
+// Uses boundTick if given, and iterates until we hit bound Tick's index. If not, uses min tick || max tick depending on the direction of zeroForOne.
+func (k Keeper) GetLiquidityNetInDirection(ctx sdk.Context, poolId uint64, zeroForOne bool, boundTickPointer *sdk.Int) ([]query.LiquidityDepth, error) {
```
p0mvn, line 258: @jonator question for you.

What would be the preferred way for you to specify the direction of the swap?

The direction of the swap is determined by which token is being swapped in. Here, we are assuming that FE will provide a boolean with the knowledge of the direction.

However, maybe we don't want frontends to care about this and, instead, just provide the denom of the token being swapped in. WDYT?

p0mvn, line 258: Why do we have to have `*sdk.Int`? Can we have `sdk.Int` and call `boundTickPointer.IsNil()` instead for simplicity?

x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,68 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
...
+		if (zeroForOne && boundTick.LT(currentTick)) || (!zeroForOne && boundTick.GT(currentTick)) {
+			return []query.LiquidityDepth{}, types.InvalidDirectionError{ZeroForOne: zeroForOne, PoolTick: currentTick.Int64(), TargetTick: boundTick.Int64()}
+		}
+		boundTick = *boundTickPointer
+	}
+
+	liquidityDepths := []query.LiquidityDepth{}
+	swapStrategy := swapstrategy.New(zeroForOne, sdk.ZeroDec(), k.storeKey, sdk.ZeroDec())
+
+	for (currentTick.GT(boundTick) && zeroForOne) || (currentTick.LT(boundTick) && !zeroForOne) {
```
p0mvn, line 292: Can we expand https://github.com/osmosis-labs/osmosis/issues/4704 to also optimize this with an iterator?

I'm also guessing that it should be possible to share some of the iteration logic with "query for liquidity depths"?

x/concentrated-liquidity/tick_test.go
```diff
@@ -762,6 +763,182 @@ func (s *KeeperTestSuite) TestGetTickLiquidityForRange() {
...
+				withLiquidityNetandTickIndex(defaultTick, DefaultMaxTick, sdk.NewDec(-10)),
+				withLiquidityNetandTickIndex(defaultTick, -5, sdk.NewDec(20)),
+				withLiquidityNetandTickIndex(defaultTick, 5, sdk.NewDec(-20)),
+				withLiquidityNetandTickIndex(defaultTick, 2, sdk.NewDec(40)),
+				withLiquidityNetandTickIndex(defaultTick, 10, sdk.NewDec(-40)),
+			},
+
+			poolId:           defaultPoolId,
+			zeroForOne:       false,
+			boundTickPointer: nil,
```
p0mvn, line 900: Can you add test cases with non-nil tick pointer please?

p0mvn (commented):
Great work. A couple of nits but the general logic looks solid.

x/concentrated-liquidity/tick.go
jonator: I think denom of token swapped in could be a bit more robust for FE use. Passing in a boolean assumes the frontend has prior knowledge of the assets' in the pool being positioned as the 0 or 1 asset.

x/concentrated-liquidity/tick.go
mattverse: yes of course! I think in #4707, we should be addressing both queries. 
Addinng it to the issue description as well

x/concentrated-liquidity/tick.go
mattverse: We're taking in a pointer from the request in the grpc router so we need somewhere to handle the pointer logic. Agreed that having pointers in arguments can cause anti-simplicity, moved the logic of handling this to grpc_query.go

x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,68 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
...
+
+		if zeroForOne {
+			boundTick = sdk.NewInt(minTick)
+		} else {
+			boundTick = sdk.NewInt(maxTick)
+		}
+	} else { // if bound is not nil, sanity check that given bound is in the direction of swap
+		if (zeroForOne && boundTick.LT(currentTick)) || (!zeroForOne && boundTick.GT(currentTick)) {
+			return []query.LiquidityDepth{}, types.InvalidDirectionError{ZeroForOne: zeroForOne, PoolTick: currentTick.Int64(), TargetTick: boundTick.Int64()}
+		}
```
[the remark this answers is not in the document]
mattverse: AFAIR, we have a function in swap strategy of validating price limits. Given the bound tick, I'm not sure if it would be any more simpler to convert the given tick into sqrt price and then using the value to use the swapstrategy method vs doing checks directly here, but don't have strong opinions on this. What do you think?

x/concentrated-liquidity/tick.go
mattverse: Changed this so that it now takes in token in denom instead of taking in ZeroForOne

x/concentrated-liquidity/tick_test.go
mattverse: added!

x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,84 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
...
+	swapStrategy := swapstrategy.New(zeroForOne, sdk.ZeroDec(), k.storeKey, sdk.ZeroDec())
+
+	checkTickWithInBoundFn := func(zeroForOne bool, currentTick, boundTick sdk.Int) bool {
+		return (currentTick.GTE(boundTick) && zeroForOne) || (currentTick.LTE(boundTick) && !zeroForOne)
+	}
+
+	nextTick, ok := swapStrategy.NextInitializedTick(ctx, poolId, currentTick.Int64())
+	if !ok {
+		return []query.LiquidityDepth{}, nil
+	}
```
p0mvn, line 306: nit: I think we can have this at the top of the for loop. That way, there is no need to have `NextInitializedTick` in two places

x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,84 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
...
+		// - there are no more next tick that is initialized,
+		// - we hit upper limit
+
+		if !ok {
+			break
+		}
+		nextTick = nextInitTick
+	}
+
+	return liquidityDepths, nil
```
p0mvn, line 331: We also need to return the active tick liquidity as a separate field

jonator: IMO current liquidity at index 0 SGTM

p0mvn: On more thinking, we currently return the following structure:
```
query.LiquidityDepth{
	LiquidityNet: tick.LiquidityNet,
	TickIndex:    nextTick,
}
```

Current liquidity is semantically incompatible with it because it is not "liquidity net". So it might be better to return it separately so that we don't confuse current liquidity with "liquidity net".

x/concentrated-liquidity/tick.go
mattverse: It was originally it the top of the for loop, but moved here because of a specific reason. Notice how we start the iteration by checking the next tick prior to starting the iteration. If the check `checkTickWithInBoundFn` has not passed, the function would end without anything contained given that tick is greater than bound.
However, if we skip this check before the for loop happens, we would always have one item, even though the given tick is greater than the bound tick. Lmk if this makes sense

x/concentrated-liquidity/tick.go
```diff
@@ -253,6 +253,84 @@ func (k Keeper) GetTickLiquidityForRange(ctx sdk.Context, poolId uint64) ([]quer
...
+		exponentAtPriceOne := p.GetPrecisionFactorAtPriceOne()
+		minTick, maxTick := math.GetMinAndMaxTicksFromExponentAtPriceOneInternal(exponentAtPriceOne)
+
+		if zeroForOne {
+			boundTick = sdk.NewInt(minTick)
+		} else {
+			boundTick = sdk.NewInt(maxTick)
+		}
+	} else { // if bound is not nil, sanity check that given bound is in the direction of swap
+		if (zeroForOne && boundTick.GT(currentTick)) || (!zeroForOne && boundTick.LT(currentTick)) {
```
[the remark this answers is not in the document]
mattverse: good call, added the logic gate and added according test cases as well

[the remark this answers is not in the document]
mattverse: on the second thought, bound tick should be in between min tick and max tick in any given case

x/concentrated-liquidity/tick.go
mattverse: Added field for current pool liquidity

x/concentrated-liquidity/tick_test.go
```diff
@@ -762,6 +762,276 @@ func (s *KeeperTestSuite) TestGetTickLiquidityForRange() {
...
+			presetTicks: []genesis.FullTick{
+				withLiquidityNetandTickIndex(defaultTick, DefaultMinTick, sdk.NewDec(10)),
+				withLiquidityNetandTickIndex(defaultTick, DefaultMaxTick, sdk.NewDec(-10)),
+				withLiquidityNetandTickIndex(defaultTick, -10, sdk.NewDec(20)),
+				withLiquidityNetandTickIndex(defaultTick, 10, sdk.NewDec(-20)),
+			},
+
+			poolId:        defaultPoolId,
+			tokenIn:       USDC,
+			boundTick:     sdk.NewInt(-5),
```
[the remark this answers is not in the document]
mattverse: > Also, I would like to see a test where boundTick is greater than max causes an error and a bound tick smaller than min causes an error

Added in the most recent commit!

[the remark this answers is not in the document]
mattverse: > I think we should have a test with a different value of boundTick and make sure that both zero for one and one for zero are tested with the bound tick.

For this comment, I had two cases that each confirms bound ticks for zero for one and one for zero, https://github.com/osmosis-labs/osmosis/blob/9a3aca2c78/x/concentrated-liquidity/tick_test.go#L862-L910.

Is it that you would like to see more test cases for zero for one and one for zero using different values?

x/concentrated-liquidity/grpc_query.go
```diff
@@ -211,6 +211,39 @@ func (q Querier) TotalLiquidityForRange(goCtx context.Context, req *clquery.Quer
...
+	if err != nil {
+		return nil, err
+	}
+
+	pool, err := q.Keeper.getPoolById(ctx, req.PoolId)
+	if err != nil {
+		return nil, err
+	}
+
+	return &clquery.QueryLiquidityNetInDirectionResponse{LiquidityDepths: liquidityDepths, CurrentLiquidity: pool.GetLiquidity()}, nil
```
p0mvn, line 244: It seems to me that we include "current_tick" and it's liquidity net in the `liquidityDepths`. However, current tick's `liquidity_net` is irrelevant for FE swap estimate.

In terms of the current tick, we only need to know a) the current tick index b) current tick liquidity (which we already return)

So I suggest only including liquidity net amounts starting from the "next tick" from current in the swap direction

Instead, we can have a separate field called `CurrentTick` similar to `Current`Liquidity`

Let me know what you think

p0mvn, line 244: Consider renaming `liquidityDepths` -> `liquidityNetAmounts` or something like that because net amount is different from depths IMO

p0mvn: CC: @jonator for awareness on this suggestion

p0mvn (commented):
Overall LGTM

Some minor comments about handling the current tick separately from the rest. I resolved all comments that I think were addressed, leaving only the comments that still require replies/changes.

I'm ready to merge once the current tick request is addressed and the overall design is accepted by Jon. 

Will wait till Jon's acceptance before approving

proto/osmosis/concentrated-liquidity/pool-model/query.proto
```diff
@@ -154,7 +163,26 @@ message LiquidityDepthWithRange {
   ];
 }
 
-//=============================== TickLiquidityInBatches
+//=============================== LiquidityNetInDirection
+message QueryLiquidityNetInDirectionRequest {
+  uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string token_in = 2 [ (gogoproto.moretags) = "yaml:\"token_in\"" ];
+  string bound_tick = 3 [
```
jonator, line 170: How can we know we're requesting _all ticks_ (in a direction)? Should we just pass in the hardcoded next_tick value?

I'm thinking in the frontend, the estimates will go like:
* Try to reduce query size and _roughly_ estimate how many ticks we'll need based on size of trade and liquidity in pool.

Above is an edge case, as I want the above to work in most common cases (trading a moderate amount against a higher liquidity pool), this is an edge case scenario:
* If we run out of ticks in simulation iteration, we'll assume it's because we didn't request enough ticks. Then we'll try again with _all_ ticks in a given direction, by passing in either `maxTick` into `bound_tick` or using some special token/flag?? @mattverse @p0mvn 
* Only after we've tried getting the remaining tick net values, can we conclude the pool doesn't have enough liquidity to handle the user's current trade, where we will show an error.

p0mvn: To estimate the expected final tick very roughly, we could take the current tick liquidity, take some proportion of that (e.g. 95%) because we assume that as we move away from the current tick, the liquidity decreases. Note, that we can come up with an algorithm to estimate the proportion too but can be hard coded to start. Finally, compute sqrt price delta in the following way:

- swapping y for x:

$$\Delta \sqrt P = \Delta y / L$$

- swapping x for y:

$$\Delta \sqrt P = L / \Delta x$$

Once we have that, we compute the "next expected sqrt price". From sqrt price, we can compute the expected final tick.

If that fails, we could then do some form of binary search where we take the estimated tick delta $$\Delta T$$  and if that wasn't enough, we query the next:
- $$2 \Delta T$$
- $$4 \Delta T$$

exclusively from the ticks we already received during prior querying to avoid redundant processing.

Similar to how Go slices reallocate buffer as we append to them.

Do you think we can keep the current design of querying until max bound for v1 and then work out a design for bounds estimates separately? @jonator @mattverse 

I think merging this would allow @mattverse and I to parallelize and further iterate on improvements

Let me know what you all think

jonator: So we can run that calculation to get the next tick index then perhaps add a few extra ticks to avoid any rounding error?

p0mvn: Yeah, that sounds right

jonator (approved):

jonator:
Todo after this PR @p0mvn @mattverse 
* Support min ticks, since

Issue #4651, opened 2023-03-17
[CL][Key Refactor] Remove freeze duration as parameter


p0mvn:
## Background

We should not need freeze duration in `MsgCreatePosition` . We can remove freezing as a parameter and just putting all positions on all supported uptimes.

## Suggested Design

Remove it. The removal depends on the position key refactor

## Acceptance Criteria
- freeze duration is removed
- all incentives logic is updated

PR #4724, opened 2023-03-24
refactor[CL]: remove freezeDuration as a parameter

czarcas7ic:
Closes: #4651 

We no longer need the freezeDuration parameter, as we are now utilizing accumulators for all positions, and simply force users to forfeit accumulated rewards if they leave prior to the rewards minimum uptime charging period.

This PR removes the parameter and refactors various tests to achieve this. 

Most notably, we no longer determine forfeiting incentives as binary, and rather iterate a position ID over the various rewards accumulators to determine if the position has existed long enough to claim its share of the rewards, or forfeit them due to not being in the position long enough.

## Testing and Verifying

This PR refactors existing tests


AlpinYukseloglu:
Took a crack at refactoring incentives tests to make these changes work, but want to do at least one more iteration on it to make sure the tests don't end up extremely hacky/messy.

tests/e2e/configurer/chain/commands.go
```diff
@@ -101,7 +101,7 @@ func (n *NodeConfig) StoreWasmCode(wasmFile, from string) {
 
 func (n *NodeConfig) WithdrawPosition(from, liquidityOut string, positionId uint64) {
 	n.LogActionF("withdrawing liquidity from position")
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "withdraw-position", fmt.Sprint(positionId), liquidityOut, fmt.Sprintf("--from=%s", from)}
+	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "withdraw-position", fmt.Sprint(positionId), liquidityOut, fmt.Sprintf("--from=%s", from), "--gas=auto", "--gas-prices=0.1uosmo", "--gas-adjustment=1.3"}
```
czarcas7ic, line 104: TODO: Make an issue regarding this gas adjustment that had to be made for this transaction to go through

proto/osmosis/poolmanager/v1beta1/query.proto
```diff
@@ -45,7 +45,8 @@ service Query {
       EstimateSinglePoolSwapExactAmountOutRequest)
       returns (EstimateSwapExactAmountOutResponse) {
     option (google.api.http).get =
-        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate_out/single_pool_swap_exact_amount_out";
+        "/osmosis/poolmanager/v1beta1/{pool_id}/estimate_out/"
+        "single_pool_swap_exact_amount_out";
```
czarcas7ic, line 49: Checked this, apparently it happens automatically from the proto linter, weird.

x/concentrated-liquidity/store.go
```diff
@@ -48,18 +57,6 @@ func ParsePositionIdFromBz(bz []byte) (uint64, error) {
...
-// Returns an error if fails to unmarshal.
-func ParsePosition(value []byte) (model.Position, error) {
-	position := model.Position{}
-	err := proto.Unmarshal(value, &position)
-	if err != nil {
-		return model.Position{}, err
-	}
-	return position, nil
-}
-
```
p0mvn, line 62: nit: why not have it as a function and not callback in case we need to reuse it later?

x/concentrated-liquidity/msg_server_test.go
```diff
@@ -226,10 +227,17 @@ func (suite *KeeperTestSuite) TestCollectIncentives_Events() {
 			pool := suite.PrepareConcentratedPool()
 			suite.SetupDefaultPosition(pool.GetId())
 
+			position, err := suite.App.ConcentratedLiquidityKeeper.GetPosition(ctx, tc.positionId)
+			suite.Require().NoError(err)
+
+			ctx = ctx.WithBlockTime(position.JoinTime.Add(time.Hour * 24 * 7))
+
+			positionAge := ctx.BlockTime().Sub(position.JoinTime)
```
p0mvn, line 235: Can you please explain what's happening here? Also please consider adding comments from your reply

x/concentrated-liquidity/lp_test.go
```diff
@@ -339,18 +336,6 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 				amount1Expected: baseCase.amount1Expected.QuoRaw(2), // 2500 usdc
 			},
 		},
-		"position still unfreezing": {
```
p0mvn, line 342: Is there a test case left that covers early withdrawal and incentives being forfeited?

x/concentrated-liquidity/incentives.go
```diff
@@ -518,59 +510,71 @@ func prepareAccumAndClaimRewards(accum accum.AccumulatorObject, positionKey stri
...
+func (k Keeper) claimAllIncentivesForPosition(ctx sdk.Context, positionId uint64) (sdk.Coins, sdk.Coins, error) {
 	// Retrieve the position with the given ID.
 	position, err := k.GetPosition(ctx, positionId)
 	if err != nil {
-		return sdk.Coins{}, err
+		return sdk.Coins{}, sdk.Coins{}, err
 	}
 
+	// TODO: add validation to ensure this is never negative
+	positionAge := ctx.BlockTime().Sub(position.JoinTime)
```
p0mvn, line 526: nit: consider adding a quickie. Seems useful

x/concentrated-liquidity/position.go
```diff
@@ -56,12 +55,12 @@ func (k Keeper) initOrUpdatePosition(
 		return types.NegativeLiquidityError{Liquidity: liquidity}
 	}
 
-	err = k.initOrUpdatePositionUptime(ctx, poolId, liquidity, owner, lowerTick, upperTick, liquidityDelta, joinTime, freezeDuration, positionId)
+	err = k.initOrUpdatePositionUptime(ctx, poolId, liquidity, owner, lowerTick, upperTick, liquidityDelta, joinTime, positionId)
```
p0mvn, line 58: Copying from DMs:

From my understanding as it is in your PR - when update happens, the following occurs: 
- we update the position accumulator
- under the hood we claim rewards
    * however, I don't see logic for handling forfeited rewards there. Is the assumption that we are going to disable adding to a position in a separate PR?

If not, what is the right summary of the interaction during position update and incentive claiming/forfeit?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -2678,66 +2580,27 @@ func (s *KeeperTestSuite) TestCollectIncentives() {
...
+				upperTick:   2,
+				liquidity:   DefaultLiquidityAmt,
+				joinTime:    defaultJoinTime,
+				positionId:  DefaultPositionId,
+				collectTime: defaultJoinTime.Add(100),
+			},
+			timeInPosition: oneDay,
 			// Since there is no other existing liquidity, we expect all of the growth inside to accrue to be claimed for the
 			// uptimes the position qualifies for.
-			expectedIncentivesClaimed: expectedIncentivesFromUptimeGrowth(uptimeHelper.hundredTokensMultiDenom, DefaultLiquidityAmt, oneDayFreeze, defaultMultiplier),
```
p0mvn, line 2694: Why is this one removed?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -2798,14 +2659,12 @@ func (s *KeeperTestSuite) TestCollectIncentives() {
 			}
 
 			// System under test
-
-			s.Ctx = s.Ctx.WithBlockTime(tc.positionParams.collectTime)
-			actualIncentivesClaimed, err := clKeeper.CollectIncentives(ctx, tc.positionParams.owner, tc.positionParams.positionId)
+			actualIncentivesClaimed, err := clKeeper.CollectIncentives(ctx, ownerWithValidPosition, DefaultPositionId)
```
p0mvn, line 2662: I did not review test coverage of new logic in-detail but would like to confirm if you think that the new logic is sufficiently covered by the existing tests?

p0mvn (commented):
Would like to see the reply to how the position updates are currently handled.

Other than that, ready to approve. Great work!

x/concentrated-liquidity/store.go
czarcas7ic: Added it back here https://github.com/osmosis-labs/osmosis/pull/4724/commits/31eba27a87469b33a9c878cc31d92c8b51377055

x/concentrated-liquidity/msg_server_test.go
czarcas7ic: Yes, so in order for `CollectInventives` to emit collection events, the position must be old enough to qualify for that incentive. This is just setting up the position, and then increasing the block time 1 week in the future. I added in line comments here

https://github.com/osmosis-labs/osmosis/pull/4724/commits/6266ec5e3a186b4956c82bf0350bfb279308a40a

x/concentrated-liquidity/lp_test.go
czarcas7ic: Added a forfeit incentives test case!

https://github.com/osmosis-labs/osmosis/pull/4724/commits/47c582d32701a373e91e1325779773f15a81ef3a

x/concentrated-liquidity/incentives.go
czarcas7ic: Added negative duration validation and added a test

https://github.com/osmosis-labs/osmosis/pull/4724/commits/47c582d32701a373e91e1325779773f15a81ef3a

x/concentrated-liquidity/position.go
czarcas7ic: Discussed this further in DMs, TLDR:

Every createPosition call will be unique, since joinTime will differ and a uniqueID will be assigned to it.

x/concentrated-liquidity/incentives_test.go
czarcas7ic: So this is the only test case that tests multiple positions that have the same range. Previously, this was necessary to test because these positions would be combined as one under the hood when you claimed. We now treat every position independently (it could be one position or 100 position in the same range, a single position ID is needed to call claim). Therefore, this test felt obsolete. 

Please let me know what you think.

x/concentrated-liquidity/incentives_test.go
czarcas7ic: IMO it is sufficiently covered because we are going from broader to more specific. For instance, before we had to test that multiple positions get claimed together when they cover the same range and same pool. Now we rely on the unique ID which simplifies the logic.

With that being said, I think the tests are sufficient for now, but we may extend them after our first internal audit.

tests/e2e/configurer/chain/commands.go
czarcas7ic: Issue created https://github.com/osmosis-labs/osmosis/issues/4774

x/concentrated-liquidity/incentives.go
```diff
@@ -415,8 +415,16 @@ func (k Keeper) GetUptimeGrowthOutsideRange(ctx sdk.Context, poolId uint64, lowe
...
+// initOrUpdatePositionUptime either adds or updates records for all uptime accumulators `position` qualifies for
+func (k Keeper) initOrUpdatePositionUptime(ctx sdk.Context, poolId uint64, liquidity sdk.Dec, owner sdk.AccAddress, lowerTick, upperTick int64, liquidityDelta sdk.Dec, joinTime time.Time, positionId uint64) error {
+	// We update accumulators _prior_ to any position-related updates to ensure
+	// past rewards aren't distributed to new liquidity. We also update pool's
+	// LastLiquidityUpdate here.
+	err := k.updateUptimeAccumulatorsToNow(ctx, poolId)
+	if err != nil {
+		return err
+	}
+
```
czarcas7ic, line 427: NOTE: I had to revert the `initPositionUptime` logic that was conflicting with this PR. The update position uptime logic is needed (as far as I could tell), despite getting rid of freezeDuration.

p0mvn: In which cases do we now update position uptime? Is it if we withdraw early?

x/concentrated-liquidity/incentives_test.go
p0mvn: I think it might still be a useful test case depicting the exact behavior that you described - if you have multiple positions within a range, you claim only for one within that range by id

x/concentrated-liquidity/incentives.go
czarcas7ic: Differing from our recent conversation in DMs, I actually think we need both the init and update branch of this function. Any time we withdraw a partial amount of liquidity (or even an entire amount for that matter), the positionCustomAccum should be updated with the liquidity delta.

The previous point I mentioned in DMs still stands, with regards to how checking the positionAge (in place of freezeDuration) does not make sense, it it should be the way it currently is in this PR IMO.

x/concentrated-liquidity/incentives_test.go
czarcas7ic: Agreed, added it back here! 

https://github.com/osmosis-labs/osmosis/pull/4724/commits/c44b0fe018e9f662eb0d0bc97472db0ac5a1e91b

p0mvn (approved):
LGTM. Great work.

It will be helpful to update `architecture.md` with the state and updates in this PR but not-blocking for merge.

PR #4737, opened 2023-03-24
fix: e2e response parsing

czarcas7ic:
Closes: #XXX

In E2E, we parse the response of CreatePosition in order to determine the position_id that was created. The solution that was put in place proved to be flakey. This is an attempt at making the response parsing be more reliable by utilizing the json instead of plain text response.

An example of a failed run due to this parsing can be found here https://github.com/osmosis-labs/osmosis/actions/runs/4512657404/jobs/7946415111#step:9:483

## Brief Changelog

- CreatePosition returns json string
- Json string is parsed for position ID

## Testing and Verifying

This change is already covered by existing tests (E2E)

### DIFF
--- a/tests/e2e/configurer/chain/commands.go
+++ b/tests/e2e/configurer/chain/commands.go
@@ -73,14 +73,18 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, freezeDuration string, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
 
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), freezeDuration, fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId)}
-	outBuf, _, err := n.containerManager.ExecTxCmd(n.t, n.chainId, n.Name, cmd)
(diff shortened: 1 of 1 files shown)


tests/e2e/configurer/chain/commands.go
```diff
@@ -73,14 +73,18 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
...
-	outBuf, _, err := n.containerManager.ExecTxCmd(n.t, n.chainId, n.Name, cmd)
+	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), freezeDuration, fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "-o json"}
+	outJson, _, err := n.containerManager.ExecTxCmdWithSuccessString(n.t, n.chainId, n.Name, cmd, "code\":0")
 	require.NoError(n.t, err)
 
-	prefix := "position_id\",\"value\":\""
-	startIndex := strings.Index(outBuf.String(), prefix) + len(prefix)
-	endIndex := startIndex + strings.Index(outBuf.String()[startIndex:], "\"}")
-	positionID, err := strconv.ParseUint(outBuf.String()[startIndex:endIndex], 10, 64)
+	var txResponse map[string]interface{}
```
p0mvn, line 80: What if we define a struct with all json fields specified for parsing, like:

```go
type txResponse struct {
    events `json:"events,omitempty"`
   ...
}
```

this might help reduce the need for json parsing and let `json.Unmarshal` handle it

p0mvn (approved):

tests/e2e/configurer/chain/commands.go
czarcas7ic: I think we can do this in a follow on issue. I will merge this to stop CI from failing 50 percent of the time

Issue #4726, opened 2023-03-24
Do not check `localhost` links in weekly link breakage action


pysel:
## Background
In https://github.com/osmosis-labs/osmosis/pull/4712 a github action for weekly links check in .md files was added. The recent run of this workflow created this issue: https://github.com/osmosis-labs/osmosis/issues/4722. It included some links that contain `localhost`. I think, if there is a way not to report links containing `localhost`, we should not do it
## Suggested Design

Links containing `localhost` are not reported

## Acceptance Criteria

Not reported

pysel:
CC: @MasterPi-2124

PR #4712, opened 2023-03-23
daily check for broken links

MasterPi-2124:
Closes: #1850

## What is the purpose of the change

> Created a CI job that auto check for broken links in all MD files, and create an Issue of broken links. This can be done automatically every day at 00:00 UTC, or triggered manually

## Testing and Verifying

Tested at https://github.com/notional-labs/osmosis/actions/runs/4499627236 and created a new issue https://github.com/notional-labs/osmosis/issues/15

PR #4741, opened 2023-03-25
Master pi 2124/check broken links

MasterPi-2124:
Closes: #4726

As an improvement to #4712, we will exclude `localhost` and loopback IP.


pysel (approved):
utACK! Thanks

p0mvn (approved):
Awesome thanks

Issue #4745, opened 2023-03-26
Link Checker once per 2 weeks


pysel:
## Background
https://github.com/osmosis-labs/osmosis/issues/4740#issuecomment-1483922448

## Suggested Design

Link checking workflow runs every 2 weeks

## Acceptance Criteria

design implemented

pysel:
CC: @MasterPi-2124, hey, are you available to fix that?

PR #4746, opened 2023-03-26
GH Actions: link checker runs every two weeks

pysel:
Closes: #4745 

Runs link checking workflow on 1st and 15th of every month (cannot explicitly specify something like `every two weeks` in cron time)

## Testing and Verifying

minor

### DIFF
--- a/.github/workflows/check-broken-links.yml
+++ b/.github/workflows/check-broken-links.yml
@@ -2,7 +2,7 @@ name: Check Broken Links
 
 on:
   schedule:
-    - cron: '0 0 * * *' # Run daily at 00:00 UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/check-broken-links.yml
```diff
@@ -2,7 +2,7 @@ name: Check Broken Links
 
 on:
   schedule:
-    - cron: '0 0 * * *' # Run daily at 00:00 UTC
+    - cron: '0 10 1,15 * *' # Run on 1st and 15th days of every month at 10:00
```
pysel, line 5:
```suggestion
    - cron: '0 10 1,15 * *' # Run on 1st and 15th days of every month at 10:00 UTC
```

p0mvn (approved):
Sweet, thank you!

Issue #4748, opened 2023-03-26
CL: Rename `PrecisionFactorAtPriceOne` to `ExponentAtPriceOne` in `Pool` struct


pysel:
## Background

There is an inconsistency in the naming of a field in the Pool struct in the concentrated-liquidity module. The field `PrecisionFactorAtPriceOne` should be renamed to `ExponentAtPriceOne` to maintain consistency with the rest of the module and improve clarity for new developers.

## Suggested Design

* [Current pool struct](https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/model/pool.pb.go#L37)
* Expected pool struct: 
```go
type Pool struct {
    // ...
    ExponentAtPriceOne github_com_cosmos_cosmos_sdk_types.Int `protobuf:"bytes,9,opt,name=exponent_at_price_one,json=exponentAtPriceOne,proto3,customtype=github.com/cosmos/cosmos-sdk/types.Int" json:"exponent_at_price_one" yaml:"exponent_at_price_one"`
    // ...
}
```
## Acceptance Criteria

Design implemented

PR #4749, opened 2023-03-26
refactor[CL]: rename `PrecisionFactorAtPriceOne` to `ExponentAtPriceOne`

pysel:
Closes: #4748 

Better dev-UX

## Testing and Verifying

N/A

N/A


p0mvn (approved):

czarcas7ic:
Hey, I wasn't aware of this issue, but IMO the standard should be `PrecisionFactorAtPriceOne` instead of `exponentAtPriceOne`

mattverse:
I stand by this re-name, exponent at price one seems like a more intuitive naming for me compared to precision factor at price one.

> Hey, I wasn't aware of this issue, but IMO the standard should be `PrecisionFactorAtPriceOne` instead of `exponentAtPriceOne`

@czarcas7ic  Is there a reason you prefer `PrecisionFactorAtPriceOne`?

czarcas7ic:
My conviction on the name isn't that strong, happy to go with the exponent name since you all agree!

mattverse (approved):

pysel:
> My conviction on the name isn't that strong, happy to go with the exponent name since you all agree!

hey @czarcas7ic, my point is to use one name for the same thing throughout the repo to prevent confusions with the name, but I think I agree with you that `PrecisionFactorAtPriceOne` is a better name. 

ofc it is up to you guys to pick the final name, I only want to prevent naming confusions

CC: @mattverse @p0mvn

stackman27:
one simple nit that i noticed is that we haven't added [exponent-at-price-one] as a part of tx use field [here](https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/client/cli/tx.go#L30)

can we please add it create-concentrated-pool [denom-0] [denom-1] [tick-spacing] [exponent-at-price-one] [swap-fee]

p0mvn:
> one simple nit that i noticed is that we haven't added [exponent-at-price-one] as a part of tx use field [here](https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/client/cli/tx.go#L30)
> 
> can we please add it so that it looks like this;
> 
> ```
> create-concentrated-pool [denom-0] [denom-1] [tick-spacing] [exponent-at-price-one] [swap-fee]
> ```

@pysel can you please address this, and then we can merge this

czarcas7ic:
Added the comment, will merge once passes CI

pysel:
> Added the comment, will merge once passes CI

oh, apologies for a long response, thank you @czarcas7ic for adding!

Issue #4184, opened 2023-01-31
[Simulator] User bankKeeper.Keeper instead of simtypes.BankKeeper


stackman27:
## Background
Currently, if you do sim.BankKeeper()..... we can only access 3 functions that is registered in simtypes.BankKeeper. We can import other keeper methods here but we are limited by these keeper methods (https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/types/expected_keepers.go#L28). 

Because of this, methods that are outside of staking types expected keeper cannot be registered in simtypes.BankKeeper for instance `SetDenomMetaData(ctx sdk.Context, denomMetaData banktypes.Metadata)`. We want to fix this so that we can import any bankKeeper method 

## Suggested Design

We can go about this 2 ways; 
1. Abandon the entire simtypes.BanKeeper and start using bankKeeper.Keeper so that we have access to ALL the bankKeeper methods regardless of us registering in the simtypes.BankKeeper interface or not. We currently do this for stakingKeeper (https://github.com/osmosis-labs/osmosis/blob/main/simulation/simtypes/simctx.go#L91) 
2. Allow adding more bankKeeper methods in simtypes.BanKeeper interface which will break some other simulator like incentives and superfluid. We'd have to fix that as well

## Acceptance Criteria
1. Simulator should be able to add any bankKeeper method to the interface or access it directly 
2. All simulator test should work as expected

PR #4768, opened 2023-03-28
[Simulator] Use bankKeeper.Keeper instead of simtypes.BankKeeper

phamminh0811:
Closes: #4184 

As suggested, this pr follow the first way as abandon the entire simtypes.BanKeeper and start using bankKeeper.Keeper 

## Brief Changelog

- remove simtypes.BanKeeper and use bankKeeper.Keeper instead

## Testing and Verifying

This change is already covered by existing tests.

### DIFF
--- a/simulation/simtypes/app.go
+++ b/simulation/simtypes/app.go
@@ -6,6 +6,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
 )
 
@@ -15,7 +16,7 @@ type App interface {
 	GetBaseApp() *baseapp.BaseApp
 	AppCodec() codec.Codec
 	GetAccountKeeper() AccountKeeper
-	GetBankKeeper() BankKeeper
+	GetBankKeeper() bankkeeper.BaseKeeper
 	GetStakingKeeper() stakingkeeper.Keeper
 	ModuleManager() module.Manager
 }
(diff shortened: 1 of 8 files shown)


simulation/simtypes/app.go
```diff
@@ -24,10 +25,3 @@ type AccountKeeper interface {
 	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authtypes.AccountI
 	// GetAllAccounts(ctx sdk.Context) []authtypes.AccountI
 }
-
-type BankKeeper interface {
-	GetBalance(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin
-	GetAllBalances(ctx sdk.Context, addr sdk.AccAddress) sdk.Coins
-	// TODO: Revisit
-	SpendableCoins(ctx sdk.Context, addr sdk.AccAddress) sdk.Coins
-}
```
p0mvn, line 33: Why is it not possible to define the desired methods on this interface?

Allowing them all by default doesn't seem preferable to me. We should be limiting the interface only to the functionality that we need for the purposes of reducing complexity and the risk of accidentally misusing undesired functionality.

phamminh0811: hmm, so it is no need to handle this issue, right?

stackman27: @p0mvn so we can define the methods on the interface but there are restrictions to it. For instance if you try to add his method `SetDenomMetaData(ctx sdk.Context, denomMetaData banktypes.Metadata)`, the rest of the simulator breaks because if you check x/incentives/simulation/operations.go we use `bk stakingTypes.BankKeeper`  to get bankkeeper. 

this PR i think removes `stakingTypes.BankKeeper` and uses a unified bankKeeper interface, we donot necessarily have to use bankKeeper from cosmos-sdk instead we can keep the interface and use simtypes.BankKeeper

p0mvn (approved):

p0mvn:
@phamminh0811 hey can you please refresh the PR, and let's merge this?

phamminh0811:
Yes, I've created https://github.com/osmosis-labs/osmosis/pull/5028 for refreshing the PR as discussed, please check

PR #4675, opened 2023-03-20
feat/spike: CosmWasm Pool Type and Module

p0mvn:
Closes: #4389

## What is the purpose of the change

The CosmWasm Pool Module is an extension for the Osmosis pools, aiming to create a custom module that allows users to create and manage liquidity pools backed by CosmWasm smart contracts. The feature enables developers to build and deploy custom smart contracts that can be integrated with the rest of the pool types on the Osmosis chain.

The module is built on top of the CosmWasm smart contracting platform, which provides a secure and efficient way to develop and execute WebAssembly (Wasm) smart contracts on the Cosmos SDK.

Having pools in CosmWasm provides several benefits, one of which is avoiding the need for chain upgrades when introducing new functionalities or modifying existing ones related to liquidity pools. This advantage is particularly important in the context of speed of development and iteration.

An example of a CosmWasm pool type: https://github.com/osmosis-labs/transmuter

## Brief Changelog

- [x] created new SDK module - `cosmwasmpool`
- [x] wired it up to the app
- [x] made `cosmwasmpool` implement `poolmanagertypes.PoolModule`
- [x] created new Pool - `x/cosmwasm/model.Pool` of type `poolmanagertypes.CosmWasm`
- [x] made new pool extend `poolmanagertypes.PoolI`
- [x] implemented `MsgCreateCosmWasmPool`
- [x] implemented `cosmwasmpool.Keeper.InitializePool`
- [x] implemented pool interface extension `CosmWasmExtension`
- [x] implemented helpers for sending query and sudo message to a cosmwasm contract
- [x] implemented all `PoolI` and `CosmWasm` extension methods
- [x] implemented all  `poolmanagertypes.PoolModule`
- [ ] test
   - [x] pool.GetSwapFee()
   - [x] pool.SpotPrice()
   - [ ] cosmwasmpoolkeeper.InitializePool()
        - [x] full coverage with mocks
        - [ ] test with real `transmuter` pool
  - [x] cosmwasmpoolkeeper.CalcOutAmtGivenIn()
  - [ ] cosmwasmpoolkeeper.SwapOutAmtGivenIn()
       * ran into bug where contract balance is not updated upon joining the pool
- [ ] document state entries
- [ ] document messages
- [ ] document queries

## Testing and Verifying

TBD after getting concept-ACK from @iboss-ptk 

## Open Quesitons

- [x] What to do with the `creatorAddress` in `InitializePool`?

- [x] How to resolve codec issues around `PoolI`

Currently, getting the following:
[...]

PR #4792, opened 2023-03-30
feat/docs(CWPool): contract messages and basic readme

p0mvn:
Closes: #XXX

This PR merges https://github.com/osmosis-labs/osmosis/pull/4675 in smaller chunks.

The change here focuses on generating message boilerplate to interact between the chain and the contract.

Additionally, it adds basic READMEs. No state breaks in this PR.

**Review suggestion: only skim `proto` folder and README.md. This is under 300 LOC, the rest is generated boilerplate.**

## Testing and Verifying


proto/osmosis/cosmwasmpool/v1beta1/model/module_sudo_msg.proto
```diff
@@ -0,0 +1,81 @@
...
+  ];
+
+  // swap_fee is the swap fee for this swap estimate.
+  string swap_fee = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
+
+message SwapExactAmountOutSudoMsg {
```
nicolaslara, line 69: If we're gonna launch the contract first and pools chain side later, then the messages need to be execute, not sudo (as only the chain can send sudo messages).

x/cosmwasmpool/cosmwasm/msg/module_query_msg.go
```diff
@@ -0,0 +1,27 @@
+// This file defines helpers for querying
+// the cosmwasm pool contract from the cosmwasm pool module.
+package msg
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// CalcOutAmtGivenIn
+func NewCalcOutAmtGivenInRequest(tokenIn sdk.Coin, tokenOutDenom string, swapFee sdk.Dec) CalcOutAmtGivenInRequest {
+	return CalcOutAmtGivenInRequest{
```
nicolaslara, line 9: Just our of curiosity, these functions should always return the the same as the inputs for transmuter, right?

proto/osmosis/cosmwasmpool/v1beta1/model/module_sudo_msg.proto
p0mvn: Ideally, this should still be a sudo message, and the contract refactored and upgraded.

We should be customizing the contract to make the chain code as secure and general as possible instead of adjusting chain code to be transmuter-specific

p0mvn: https://github.com/osmosis-labs/osmosis/issues/4797

x/cosmwasmpool/cosmwasm/msg/module_query_msg.go
nicolaslara: yeah, that makes sense. My understanding of transmutter is that it just swaps tokens 1:1. So I guess these functions don't have anything to calculate: send 1 axlusdc, get 1 nobleusdc.

proto/osmosis/cosmwasmpool/v1beta1/model/module_sudo_msg.proto
nicolaslara: I still think it makes sense for this to be an exec. Is there a reason we don't want people using the pools directly?

p0mvn: Tagged you here with the reasoning: https://github.com/osmosis-labs/osmosis/pull/4810#discussion_r1154605426

Even if we decide to change this in the future, let's not block this PR on this request, please. The reason is that everything else in the spike #4675, including the test contract, depends on these messages being sudo. As a result, it would be blocked on Boss refactoring the test contract.

I will make sure to use #4797 as a tracker to address this based on the conclusions from discussions we are having here: https://github.com/osmosis-labs/osmosis/pull/4810#discussion_r1154605426

Let me know if that works

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,39 @@
+# CosmWasm Pool Module
```
p0mvn, line 1: Note: @iboss-ptk I'm down to eventually overwrite/merge this with yours https://github.com/osmosis-labs/osmosis/pull/4810, and can help with resolving conflicts

nicolaslara (approved):

p0mvn:
Thanks! Merging. Will refresh https://github.com/osmosis-labs/osmosis/pull/4675 and continue working there

Issue #4739, opened 2023-03-25
[CL] Implement function to fungify fully charged positions


AlpinYukseloglu:
## Background

One unfortunate byproduct of tracking granular uptimes is that it becomes very difficult to handle compounding rewards without creating a new position. While the current setup is technically safe, it seems unreasonable to push a growing load on frontends as users "compound" their positions by making an ever-increasing number of positions.

The first step to resolving this is giving users the option to batch together positions that _are_ fungible with each other (i.e. fully charged positions). It is possible that this might be sufficient to reach a local maximum on this issue on its own (e.g. by having frontends run this logic prior to letting users claim rewards), but there are future additions to consider as well.

After this change, we can consider running this fungification process automatically prior to each `claimRewards` call, but there are dependencies here that might pose potential risks (e.g. accidentally fungifying positions owned by contracts, which will always be a risk even if we give users the option to opt out of this fungification process). Either way, the first step is to implement `fungifyChargedPositions`, as this will be required regardless of the path we take.

## Suggested Design

- Implement `fungifyChargedPositions`, which takes in a list of position IDs and combines them into a new position in state. This function should also delete all the original positions from state
- Ensure that the caller can only fungify charged positions that 1. they own and 2. are fully charged 3. tick ranges match up. If any position ID in the passed  in list does not pass this validation, return an error.
- Ensure that there is some cap to the number of positions that can be fungified in one call
- Expose a message for this functionality

## Testing strategy
1. Specify `fullyChargedPositions` for each test case
2. Create the positions and let enough time elapse to charge all of them (i.e. `supportedUptimes(len(supportedUptimes) - 1)`)
3. Specify `invalidPositionToAdd` that creates a position after this and tacks the ID onto the list (use to specify pos with different tick range, uncharged pos, and pos owned by another account)
4. Specify `expectedFinalPosition` which should be the expected fungified position
5. Specify `expectedError` to ensure each error check returns a proper error type

## Acceptance Criteria

- All new and existing tests pass

PR #4800, opened 2023-03-30
feat[CL]: fungify charged position

czarcas7ic:
Closes: #4739 

Now that we have removed freeze duration and key by join time, every new position created is non fungible. We can, however, fungify positions that reach the fully charged duration. This PR introduces a `fungifyChargedPositions` that takes a list of position IDs, ensures the provided positions are fungible (fully charged, same owner, same pool, same liquidity range), and merges them into a single fully charged position.

## Brief Changelog

- Adds `fungifyChargedPositions` method explained above
- Tests this method as well as helper method

## Testing and Verifying

This change added gotest coverage for the newly introduced functions and logic.


x/concentrated-liquidity/position_test.go
```diff
@@ -496,3 +496,257 @@ func (s *KeeperTestSuite) TestCalculateUnderlyingAssetsFromPosition() {
...
+			}
+
+			// Increase block time by one more day
+			s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Hour * 24))
+
+			newPositionId, err := s.App.ConcentratedLiquidityKeeper.FungifyChargedPosition(s.Ctx, test.accountCallingMigration, test.positionIdsToMigrate)
+			if test.expectedErr != nil {
+				s.Require().Error(err)
+				s.Require().ErrorIs(err, test.expectedErr)
+				s.Require().Equal(test.expectedNewPositionId, newPositionId)
```
p0mvn, line 627: should this always be
```suggestion
				s.Require().Equal(0, newPositionId)
```

p0mvn, line 627: Suggest also checking that no state updates are applied

x/concentrated-liquidity/position_test.go
```diff
@@ -496,3 +496,257 @@ func (s *KeeperTestSuite) TestCalculateUnderlyingAssetsFromPosition() {
...
+			} else {
+				s.Require().NoError(err)
+				s.Require().Equal(test.expectedNewPositionId, newPositionId)
+
+				// Since the positionLiquidity of the old position should have been deleted, retrieving it should return an error.
+				for _, posId := range test.positionIdsToMigrate {
+					positionLiquidity, err := s.App.ConcentratedLiquidityKeeper.GetPositionLiquidity(s.Ctx, posId)
+					s.Require().Error(err)
+					s.Require().ErrorIs(err, types.PositionIdNotFoundError{PositionId: posId})
```
p0mvn, line 636: Nice - having concrete errors is proving to be handy here

x/concentrated-liquidity/lp.go
```diff
@@ -322,3 +322,58 @@ func emitLiquidityChangeEvent(ctx sdk.Context, eventType string, positionId uint
...
+	// Check if the provided tick range is valid according to the pool's tick spacing and module parameters.
+	if err := validateTickRangeIsValid(pool.GetTickSpacing(), pool.GetExponentAtPriceOne(), lowerTick, upperTick); err != nil {
+		return 0, err
+	}
+
+	// Create a cache context for the current transaction.
+	// This allows us to make changes to the context without persisting it until later.
+	// We only write the cache context (i.e. persist the changes) if the actual amounts returned
+	// are greater than the given minimum amounts.
+	cacheCtx, writeCacheCtx := ctx.CacheContext()
```
p0mvn, line 346: Why is this useful? This feels as overengineering because if an error is returned, it is expected to be propagated to the `runTx` logic in base app where it can be dropped.

I understand the goal of using `cacheCtx` where we want to create a non-mutative function (e.g. in "calc" swap estimate functions for queries) but not here

x/concentrated-liquidity/lp.go
```diff
@@ -322,3 +322,58 @@ func emitLiquidityChangeEvent(ctx sdk.Context, eventType string, positionId uint
 		sdk.NewAttribute(types.AttributeAmount1, actualAmount1.String()),
 	))
 }
+
+// migrateToSinglePosition is a function that creates a new liquidity position within a given pool.
+// It follows similar logic to createPosition, except it:
+// - takes a joinTime as a parameter
+// - does not transfer funds from the user to the pool since the funds are already in the pool
+// The underlying purpose of this function is to take the sum of liquidity from multiple positions and consolidate them into one.
+func (k Keeper) migrateToSinglePosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, liquidity sdk.Dec, lowerTick, upperTick int64, joinTime time.Time) (uint64, error) {
```
p0mvn, line 331: Why is it useful to have this second function where we do a similar operation of `validateTickRangeIsValid` and only marginally add new functionality?

From an architectural standpoint, modules should be deep. That is, instead of having multiple APIs with complex interfaces, it is preferable to have fewer with simpler interfaces and more functionality (deeper). ref: https://medium.com/@nathan.fooo/4-notes-modules-should-be-deep-ba5671c4288c

In simpler terms, I think having two APIs for fungification - `fungifyChargedPosition` and `migrateToSinglePosition` bloats our already complex API set without significant benefit. So my recommendation is to have `fungifyChargedPosition` only unless there is a specific reason I'm missing for having two.

p0mvn (commented):
Nice work.

There aren't any logic errors noticed. However, had some architectural suggestions that I would like to see replies to before approval.

x/concentrated-liquidity/lp.go
czarcas7ic: Sorry, I think I am misunderstanding your point. `migrateToSinglePosition` is needed in place of `createPosition` since we need to specify a joinTime as well as not transfer funds.

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,80 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	return newPositionId, nil
+}
+
+// validatePositionsAndGetTotalLiquidity checks that the positions are all in the same pool and tick range, and returns the total liquidity of the positions.
+func (k Keeper) validatePositionsAndGetTotalLiquidity(ctx sdk.Context, owner sdk.AccAddress, positionIds []uint64, basePosition model.Position) (sdk.Dec, error) {
```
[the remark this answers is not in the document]
czarcas7ic: Agreed on function input params, addressed here https://github.com/osmosis-labs/osmosis/pull/4800/commits/723125e9286c75d3e954024be0a0c796219b6b41

x/concentrated-liquidity/lp.go
czarcas7ic: That makes sense to me, addressed here https://github.com/osmosis-labs/osmosis/pull/4800/commits/8eefb3c7ec26c75846c117fd9194bc0de1c25e67

x/concentrated-liquidity/lp.go
czarcas7ic: Correct, removed here https://github.com/osmosis-labs/osmosis/pull/4800/commits/8eefb3c7ec26c75846c117fd9194bc0de1c25e67

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,101 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
+
+	// Check that all the positions are in the same pool, tick range, and are fully charged.
+	// Sum the liquidity of all the positions.
+	poolId, lowerTick, upperTick, liquidity, err := k.validatePositionsAndGetTotalLiquidity(ctx, owner, positionIds)
+	if err != nil {
+		return 0, err
+	}
+
+	// The new position's timestamp is the current block time minus the fully charged duration.
+	joinTime := ctx.BlockTime().Add(-types.FullyChargedDuration)
```
AlpinYukseloglu, line 249: Are we sure that this doesn't accidentally qualify the position for an additional two weeks of incentives?

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,101 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
+	}
+
+	return positionId, nil
+}
+
+// validatePositionsAndGetTotalLiquidity checks that the positions are all in the same pool and tick range, and returns the total liquidity of the positions.
+func (k Keeper) validatePositionsAndGetTotalLiquidity(ctx sdk.Context, owner sdk.AccAddress, positionIds []uint64) (uint64, int64, int64, sdk.Dec, error) {
+	totalLiquidity := sdk.ZeroDec()
+	// Note the first position's params to use as the base for comparison.
+	basePosition, err := k.GetPosition(ctx, positionIds[0])
```
AlpinYukseloglu, line 293: Just to make sure, we're guaranteed that `len(positionIds) > 2` since we check it in `ValidateBasic` right?

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,101 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
+			return 0, 0, 0, sdk.Dec{}, err
+		}
+		// Check that the caller owns all the positions.
+		if position.Address != owner.String() {
+			return 0, 0, 0, sdk.Dec{}, types.PositionOwnerMismatchError{PositionOwner: position.Address, Sender: owner.String()}
+		}
+
+		// Check that all the positions are fully charged.
+		fullyChargedMinTimestamp := position.JoinTime.Add(types.FullyChargedDuration)
+		if fullyChargedMinTimestamp.After(ctx.BlockTime()) {
```
AlpinYukseloglu, line 310: Nit: this should be `!fullyChargedMinTimestamp.Before(ctx.BlockTime())` since we would want positions that are charged for exactly two weeks to qualify (e.g. if someone wants to chain fungifications together). Up to you to make this change or leave as is

AlpinYukseloglu (approved):
Core logic LGTM, good work. Left a few questions/minor comments that I'd like addressed before merging, but overall looks good.

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,101 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
 	k.SetNextPositionId(ctx, nextPositionId+1)
 	return nextPositionId
 }
+
+// fungifyChargedPosition takes in a list of positionIds and combines them into a single position.
+// The previous positions are deleted from state and the new position ID is returned.
+// An error is returned if the caller does not own all the positions, if the positions are all not fully charged, or if the positions are not all in the same pool / tick range.
+func (k Keeper) fungifyChargedPosition(ctx sdk.Context, owner sdk.AccAddress, positionIds []uint64) (uint64, error) {
+	// Check we meet the minimum number of positions to combine.
+	if len(positionIds) <= MinNumPositionsToCombine {
```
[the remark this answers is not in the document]
czarcas7ic: Correct, thanks for catching

x/concentrated-liquidity/position.go
czarcas7ic: Yep! https://github.com/osmosis-labs/osmosis/blob/d918bccd3de6e3986c3b1f2cc24528140687cb32/x/concentrated-liquidity/types/msgs.go#L185-L187

x/concentrated-liquidity/position.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4800/commits/6568718072b72804aa07e59862799455398883c2

p0mvn (approved):
LGTM

x/concentrated-liquidity/position.go
```diff
@@ -226,3 +228,156 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
+			oldPositionName := string(types.KeyPositionId(positionId))
+			// Check if the accumulator contains the position.
+			hasPosition, err := uptimeAccum.HasPosition(oldPositionName)
+			if err != nil {
+				return 0, err
+			}
+			// If the accumulator contains the position, move the unclaimed rewards to the new position.
+			if hasPosition {
+				// Prepare the accumulator for the old position.
+				rewards, dust, err := prepareAccumAndClaimRewards(uptimeAccum, oldPositionName, uptimeGrowthOutside[uptimeIndex])
```
p0mvn, line 319: We should manually delete the accumulators upon claiming because `accum/ClaimRewards` only deletes the accumulators if we have zero num shares.

We cannot manually set the num shares to zero because that would break the claiming logic

czarcas7ic: Added manual delete position here. Also added tests to check that accum position AND cl positions are deleted post fungify call https://github.com/osmosis-labs/osmosis/pull/4800/commits/93aa0f949cfd7c93f86a5901fb5ea8c387ec8d6d

Issue #3427, opened 2022-11-17
Improve error message when trying to Exit 100% of a pool


joe-bowman:
#### System information

Osmosis version: `v13.0.0-rc2`

#### Expected behaviour

The last user in a pool should be able to completely exit the pool.

#### Actual behaviour

The last user in a pool must leave 1 gamm token in the pool

#### Steps to reproduce the behaviour

- Create a pool
- Join pool
- Exit pool with 100% of shares
- Receive error "100000000000000 resulted shares is larger than the max amount of 100000000000000"

ValarDragon:
This is expected behavior

joe-bowman:
Then presumably the error message wants changing? because this is somewhat confusing:
`100000000000000 resulted shares is larger than the max amount of 100000000000000`.

ValarDragon:
Ah yeah, lets change that error message! (Should just need to subtract one from the max amount displayed)

For context, it breaks some assumptions around other methods, and we are in undefined behavior if the last share exits.

The number of shares is sufficiently large, s.t. we believe that the last LP share is negligible

ValarDragon:
Want to submit or update the PR, to fix the error message?

PR #4806, opened 2023-03-31
Better error handling when exiting 100% of pool shares

pysel:
Closes: #3427


x/gamm/pool-models/balancer/pool.go
```diff
@@ -19,7 +19,7 @@ import (
 const (
 	nonPostiveSharesAmountErrFormat = "shares amount must be positive, was %s"
 	nonPostiveTokenAmountErrFormat  = "token amount must be positive, was %s"
-	sharesLargerThanMaxErrFormat    = "%s resulted shares is larger than the max amount of %s"
+	sharesLargerThanMaxErrFormat    = "resulted shares %s should be lower than the max amount of %s"
```
p0mvn, line 22: I don't think this is solving the problem presented in the issue. The problem seems that incorrect values are being formatted

p0mvn (changes requested):

x/gamm/pool-models/balancer/pool.go
pysel: So we could just subtract one from the max amount, how it was proposed [here](https://github.com/osmosis-labs/osmosis/issues/3427#issuecomment-1319369089)?

ValarDragon:
Closing for staleness

Issue #4728, opened 2023-03-24
[Routing] Pre-swap unwrapping


nicolaslara:
Epic: https://github.com/osmosis-labs/osmosis/issues/4344
Continuation of #4694

## Background

The contract is using the registries to unwrap tokens when sending. It should also do it before sending so any double-wrapped tokens are unwrapped first.
 
## Suggested Design

 * unwrap

## Acceptance Criteria

 * tests

PR #4809, opened 2023-03-31
Unwrap badly wrapped tokens before a XCS

nicolaslara:
Closes: #4728

With this change users can send IBC tokens from other chains directly to the XCS contract and the unwrapping will be handled for them. 

## Brief Changelog

 * Added pre-swap unwrapping.

## Testing and Verifying

 * Added a test verifying the success case. More tests need to be added for the potential failures


cosmwasm/contracts/crosschain-swaps/src/contract.rs
```diff
@@ -60,7 +60,7 @@ pub fn execute(
             on_failed_delivery,
         } => {
             let swap_coin = cw_utils::one_coin(&info)?;
-            execute::swap_and_forward(
+            execute::unwrap_or_swap_and_forward(
```
ValarDragon, line 63: I guess this is a gas question, but do we think its better for all swaps to do the registry check?

I'm guessing the idea here is that the counterparty just "always" sends their token to swap to osmosis, as normal, with no thoughts wrt unwinding?

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -1,24 +1,92 @@
...
+use crate::{state, ExecuteMsg};
+
+/// This function takes any token. If it's already something we can work with
+/// (either native to osmosis or native to a chain connected to osmosis via a
+/// valid channel), it will just proceed to swap and forward. If it's not, then
+/// it will send an IBC message to unwrap it first and provide a callback to
+/// ensure the right swap_and_forward gets called after the unwrap succeeds
+pub fn unwrap_or_swap_and_forward(
+    deps: DepsMut,
+    block_time: Timestamp,
```
ValarDragon, line 27: Maybe my ignorance, but why do we have block time as an argument here? Is it not included in deps?

ValarDragon: err should be included in env, right?

ValarDragon: I support a single `ctx` object!

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -1,24 +1,92 @@
...
+    contract_addr: Addr,
+    swap_coin: Coin,
+    output_denom: String,
+    slippage: swaprouter::Slippage,
+    receiver: &str,
+    next_memo: Option<SerializableJson>,
+    failed_delivery_action: FailedDeliveryAction,
+) -> Result<Response, ContractError> {
+    deps.api
+        .debug(&format!("executing unwrap or swap and forward"));
```
ValarDragon, line 37: Purely informational question, do these debugs get into the final cosmwasm binary? Or are they only in local development

nicolaslara: good question 🤔 . They're not printed, but not sure if they get removed. 

They must be in the binary, since the one we sue for testing is the same as the release one. My guess is that the call still exists (it's part of the Api trait) but they're a no-op. Still the string generation probably happens

ValarDragon: In this case, hopefully the format macro gets no-op'd, and the string content just appears in the static portion of the binary

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -1,24 +1,92 @@
...
+    let registry = Registry::default(deps.as_ref());
+    let path = registry.unwrap_denom_path(&swap_coin.denom)?;
+    if path.is_empty() {
+        return Err(RegistryError::InvalidDenomTracePath {
+            path: String::new(),
+            denom: swap_coin.denom,
+        }
+        .into());
+    }
```
ValarDragon, line 46: Is this right? What happens if you swap osmo here?

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -1,24 +1,92 @@
...
+
+    // If the path is larger than 2, we need to unwrap this token first
+    if path.len() > 2 {
```
ValarDragon, line 49: So what precisely do we mean by step on `path` here? Is this a step in the registry, or denom trace length?

If its denom trace length, then we may run into issues with tokens that are natively 2 hops right. I know that there is one asset like this (pstake), due to some legacy reason.

We don't really need to have the contract handle cleanly swapping those assets, but it should fail "safely' in such scenarios.

cosmwasm/contracts/crosschain-swaps/src/execute.rs
nicolaslara: the path is then of len() = 1 <chain=osmosis, via=None, denom=osmo>

cosmwasm/contracts/crosschain-swaps/src/execute.rs
nicolaslara: yeah, each element of the path is a hop (e.g. atom on juno sent to osmosis will have path length 3: osmosis, juno, hub). 

Didn't know 2-hop native assets was a thing. We could modify the registry so that we can tell which chain an asset is "native" to other than the default (native=not-ibc-wrapped), but for now, if we just don't include those in the registry, those calls will just fail with a path-not-found (or similar) error

cosmwasm/contracts/crosschain-swaps/src/execute.rs
nicolaslara: I'd guess the format is still executed unless the compiler does something smart there. But we can also remove those before releasing

ValarDragon: meh, doubt this is going to be make or break for anything. Was mostly just curious

cosmwasm/contracts/crosschain-swaps/src/execute.rs
ValarDragon: oh cool, path-not-found sounds like a safe failure! Can we write a go test to check that? If so, that sgtm

ValarDragon: (Can we add a comment here for definition of path length = IBC hops)

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -1,57 +1,141 @@
...
+        serde_json_wasm::to_string(&memo)?
+    } else {
+        String::new()
+    };
+
+    // Validate that the swapped token can be unwrapped. If it can't, abort
+    // early to avoid swapping unnecessarily
+    let registry = Registry::default(deps.as_ref());
+    registry.unwrap_coin_into(
+        Coin::new(1, output_denom.clone()),
```
ValarDragon, line 123: why do we use a coin representation here, instead of just pass the denom?

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -323,66 +344,121 @@ impl<'a> Registry<'a> {
...
+
+        // Define useful variables
+        let current_chain = path[0].on.clone();
+        let first_channel = path[0].via.clone();
+
+        let coin_type = if path.len() == 1 {
+            CoinType::Native
+        } else {
+            CoinType::Ibc
+        };
```
ValarDragon, line 381: Still not clear on what were doing for paths >= 3. Is that for a future PR to return an error here and rewind?

nicolaslara: We can build a memo for unwinding paths of any length. So a path >=3 will just do more hops before being sent to the destination. See this test, for example: https://github.com/osmosis-labs/osmosis/blob/main/tests/ibc-hooks/ibc_middleware_test.go#L933

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -323,66 +344,121 @@ impl<'a> Registry<'a> {
...
+        if receiver_addr[..receiver_prefix.len()] != receiver_prefix {
             return Err(RegistryError::InvalidReceiverPrefix {
-                receiver,
+                receiver: receiver_addr,
                 chain: receiver_chain.into(),
             });
         }
 
-        let ts = block_time.plus_seconds(PACKET_LIFETIME);
+        // Define the data to be used in the IBC MsgTransfer.
```
ValarDragon, line 419: I feel like we could split some of this logic for once we've determined its non-native, into another function.

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -323,66 +344,121 @@ impl<'a> Registry<'a> {
...
+            Some(channel) => Ok::<String, RegistryError>(channel.as_ref().to_string()),
+            None => {
+                // If there is no first channel (i.e. coin_type.is_native() ==
+                // True), we get the channel from the registry
+                assert!(coin_type.is_native());
+                assert!(first_transfer_chain == receiver_chain);
+                let channel = self.get_channel(current_chain.as_ref(), first_transfer_chain)?;
+                Ok(channel)
+            }
+        }?;
```
ValarDragon, line 441: Feels simpler to me to just not run this code and avoid `!native` checks throughout

cosmwasm/packages/registry/src/registry.rs
ValarDragon: Oh awesome!

nicolaslara:
I see what you mean about splitting into a separate function. That may mean we end up with some code repetition, but would probably be easier to follow. I'll merge this for now try to split it in the next PR

PR #4801, opened 2023-03-30
refactor: remove GetTotalShares, GetTotalLiquidity and GetExitFee from PoolI

p0mvn:
Closes: #4787

## What is the purpose of the change

This PR refactors `PoolI` to remove/move APIs that are either incompatible for all pools or must be defined on the pool module instead of a pool struct.

## Brief Changelog

- Move `poolmanager.PoolI.GetTotalPoolLiquidity` to
   * `poolmanager.PoolModuleI`
   * keep it on `gammtypes.CFMMPoolI` as well
- Remove/move  from `poolmanager.GetExitFee` to `gammtypes.CFMMPoolI`
   * ref: #4440 as no future pool will have an exit fee. We will continue having it on CFMMPools
- Remove/move `poolmanager.GetTotalShares` to `gammtypes.CFMMPoolI`
  * Don't have a concept of shares on CL pool.
- Implement `GetTotalPoolLiquidity` on gamm keeper
- Implement and test `GetTotalPoolLiquidity` on CL keeper
- Pass through helper function renames and exports where required.

## Testing and Verifying

Added tests and covered by existing

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? not applicable

PR #4810, opened 2023-03-31
Cosmwasmpool spec

iboss-ptk:

Adding specification to CosmWasm pool


x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+  Sender -- create poool --> x/cosmwasmpool
+  x/cosmwasmpool -- get next & set pool id --> x/poolmanager
+  x/cosmwasmpool -- instantiate contract --> x/wasm
+```
+
+The CosmWasm contract that is to be instanitiated needs to implement [CosmWasm Pool Contract Interface](#cosmwasm-pool-contract-interface) and store it on chain first. Then new pool can be created by sending `MsgCreateCosmWasmPool`.
+
+
+`MsgCreateCosmWasmPool` contains `InstantiateMsg`, which is a message that will be passed to the CosmWasm contract when it is instantiated. The structure of the message is defined by the contract developer, and can contain any information that the contract needs to be instantiated. JSON format is used for `InstantiateMsg`.
```
p0mvn, line 28: Should we decide on how code id should be handled?

I think we should be saving the code id on chain but having the flexibility to upgrade it via gov prop. WDYT?

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+## Swap
+
+One of the main reason why CosmWasm pool is implemented as a module + contract rather than a contract only is that it allows us to use the existing pool manager module to handle swap, which means things like swap routing, cross chain swap, and other functionality that depends on existing pool interface works out of the box.
+
+```mermaid
+graph TD;
+  Sender((Sender))
+  Sender -- swap --> x/poolmanager
+  x/poolmanager -- route msg to --> x/cosmwasmpool
+  x/cosmwasmpool -- sudo execute contract --> x/wasm
```
p0mvn, line 71: Interaction with the x/poolmanager happens prior to the interaction with the contract itself.

So we would get the id first, then instantiate the contract, and then update the address and save the pool struct to state on success

p0mvn: Love the diagrams

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+    Note over x/cosmwasmpool: Store CodeId, ContractAddress, and PoolId
+
+    x/cosmwasmpool -->>  Sender: MsgCreateCosmWasmPoolResponse {PoolId}
+```
+
+
+## Providing / Withdrawing Liquidity
+
+Currently, all existing pool types has their own way of providing liquidity and shares calculation. CosmWasm pool aims to be flexible that regards and let the contract define the way of providing liquidity. So there is no restriction here, and the contract developer can define the way of providing liquidity as they wish, potentially with execute endpoint since `MsgExecuteContract` triggers state mutating endpoint and can also attach funds to it.
```
p0mvn, line 55:
```suggestion
Currently, all existing pool types have their own way of providing liquidity and shares calculation. CosmWasm pool aims to be flexible that regards and let the contract define the way of providing liquidity. So there is no restriction here, and the contract developer can define the way of providing liquidity as they wish, potentially with execute endpoint since `MsgExecuteContract` triggers state mutating endpoint and can also attach funds to it.
```

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+    swap_fee: Decimal,
+},
+```
+
+The reason why this needs to be sudo endpoint, which can only be called by the chain itself, is that the chain can provide correct information about `swap_fee`, which can be deviated from contract defined `swap_fee` in multihop scenario.
+
+(TBD)
+Problem with this approach is that swap expects funds to be sent with the same amount as `token_in` but there is no way to validate that in contract, so we need to make sure of that in `x/cosmwasmpool` module.
+
+Alternative approach is to expose swap in execute endpoint instead and validate that msg sender is a specific module account so that we can leverage `MsgExecuteContract`'s `funds` field to send funds to the contract.
```
p0mvn, line 110: CC: @nicolaslara

p0mvn: I don't understand the benefit of the `MsgExecture` approach. The sudo endpoint seems to be better from a security standpoint. 

The only benefit of having it as execute is that users would be able to interact with the contract directly. However, that is blocked anyways because we provide swap fee as param and must validate that sender is a module account

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+```mermaid
+graph TD;
+  Sender((Sender))
+  Sender -- swap --> x/poolmanager
+  x/poolmanager -- route msg to --> x/cosmwasmpool
+  x/cosmwasmpool -- sudo execute contract --> x/wasm
+  x/wasm -- sudo --> wasm/pool
+
+  x/cosmwasmpool -- send token_in from sender to wasm/pool --> x/bank
+  wasm/pool -- send token_out to sender --> x/bank
```
p0mvn, line 75: From the diagram it's unclear to me where ths bank send happens. Could you please also elaborate on the latest decisions with respect to bank send? Do they happen on chain and why or why not?

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+On contract's sudo enpoint, `SetActive` can be called to deactivate the pool. This will prevent the pool from being used for swap, and also prevent users from providing liquidity to the pool. Contract needs to check if the pool is active before performing any state mutating operation except `SetActive`.
+
+```rs
+SetActive {
+    is_active: bool,
+}
+```
+
+(TBD) On how to handle the deactivation operationally.
```
p0mvn, line 122: Let's store this `isActive` flag on chain in the pool struct? It can be set to true or false by governance.

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+    #[returns(GetSwapFeeResponse)]
+    GetSwapFee {},
+
+    /// Returns whether the pool has swaps enabled at the moment
+    #[returns(IsActiveResponse)]
+    IsActive {},
+
+    /// GetTotalShares returns the total number of LP shares in the pool
+    #[returns(TotalSharesResponse)]
+    GetTotalShares {},
```
p0mvn, line 148: This is not on the `PoolI` anymore. Ref: https://github.com/osmosis-labs/osmosis/pull/4801

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+    /// GetSwapFee returns the pool's swap fee, based on the current state.
+    /// Pools may choose to make their swap fees dependent upon state
+    /// (prior TWAPs, network downtime, other pool states, etc.)
+    /// hence Context is provided as an argument.
+    #[returns(GetSwapFeeResponse)]
+    GetSwapFee {},
+
+    /// Returns whether the pool has swaps enabled at the moment
+    #[returns(IsActiveResponse)]
+    IsActive {},
```
p0mvn, line 144: From the previous comment, I think we can have this be on-chain only.

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Sudo
+
+```rs
+#[cw_serde]
+pub enum SudoMessage {
+    /// SetActive sets the active status of the pool.
+    SetActive {
+        is_active: bool,
+    },
```
p0mvn, line 226: Might have to remove this depending on our decision of where to store the flag

p0mvn (commented):
Great work and thank you @iboss-ptk .

Left some questions and comments. Let me know what you think

x/cosmwasmpool/README.md
nicolaslara: I wasn't aware of the swap fee issue. I thought that could be calculated or queried from the contract (but now I see why it can't).

The benefit was that the contract would be standalone and the module would be just a wrapper that doesn't need to do custom logic (send the tokens to the contract before calling execute); we could then let anyone execute the contract directly. But with the swap-fee issue in mind, then I agree that the sudo solution is better

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+The module is built on top of the CosmWasm smart contracting platform, which provides a secure and efficient way to develop and execute WebAssembly (Wasm) smart contracts on the Cosmos SDK.
+
+Having pools in CosmWasm provides several benefits, one of which is avoiding the need for chain upgrades when introducing new functionalities or modifying existing ones related to liquidity pools. This advantage is particularly important in the context of speed of development and iteration.
+
+An example of a CosmWasm pool type: [transmuter](https://github.com/osmosis-labs/transmuter)
+
+
+## Creating new CosmWasm Pool
+
+To create new CosmWasm Pool, there are 3 modules involved: `x/cosmwasmpool`, `x/wasm`, and `x/poolmanager`. Here is an overview of the process:
```
nicolaslara, line 14: The naming here seems odd. Shouldn't poolmanager be the entrypoint and then have to options cosmwasmpool or nativepool (or something like that)

nicolaslara: ahh, I get what you mean now! I guess what was confusing was the diagram. The one bellow clarified it for me.

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,284 @@
...
+    token_in_max_amount: Uint128,
+    token_out: Coin,
+    swap_fee: Decimal,
+},
+```
+
+The reason why this needs to be sudo endpoint, which can only be called by the chain itself, is that the chain can provide correct information about `swap_fee`, which can be deviated from contract defined `swap_fee` in multihop scenario.
+
+(TBD)
+Problem with this approach is that swap expects funds to be sent with the same amount as `token_in` but there is no way to validate that in contract, so we need to make sure of that in `x/cosmwasmpool` module.
```
nicolaslara, line 108: We should specify as part of the spec that the chain-side is required to set token_in as token_in_max_amount as *exactly* the amount of tokens it sends to the contract.

That's basically a rephrasing of this line so that it's part of the spec

nicolaslara (commented):
This looks great! I think using sylvia's interfaces on transmuter will be helpful. There can be a Pool trait that any struct can implement to become a pool

x/cosmwasmpool/README.md
iboss-ptk: Not sure if we need to save code id if we already save address, since we can query it from wasm module unless we have explicit need to store.

For gov prop, I think it would be nice if we can have prop type where it does
- store and init
- store and migrate

Else we need 2 separated gov prop every time new pool type is added / upgrade. It might not be very frequent but if it's not too hard to do I think it make sense.

x/cosmwasmpool/README.md
iboss-ptk: Seems that everyone is on the same page now, will remove TBD section then.

x/cosmwasmpool/README.md
iboss-ptk: if `isActive` is on chain, how should we stop join pool type action? (since that does not goes through chain)

iboss-ptk:
> This looks great! I think using sylvia's interfaces on transmuter will be helpful. There can be a Pool trait that any struct can implement to become a pool

Exactly! Planning on extracting what described in spec as sylvia pool interface! @nicolaslara

x/cosmwasmpool/README.md
iboss-ptk: I will add sequence diagram for clarifying this part then

x/cosmwasmpool/README.md
iboss-ptk: Not sure if you comment means there is anything wrong with the diagram or not? Or simply lack of detail. I will add sequence diagram to add more detail if so. And make sure that this diagram is just an overview on all interactions on this flow.

x/cosmwasmpool/README.md
iboss-ptk: Or maybe making it explicit that this is an overview, sort of system context diagram. And for detail, see sequence diagram below...

x/cosmwasmpool/README.md
iboss-ptk: Added!

p0mvn:
What is the plan with incentives for `x/cosmwasm` pools?

x/cosmwasmpool/README.md
nicolaslara: I think either of those works. My confusion was because I thought the Y part of the diagram meant you had to take one of the two paths

iboss-ptk:
> What is the plan with incentives for x/cosmwasm pools? Could you document that as well, please? @iboss-ptk
@p0mvn will look into it!

p0mvn:
Just thought of another consideration - we should probably mention how twap is handled. It probably does not make sense to have it integrated it for transmuter but might be worthwhile for other pools

iboss-ptk:
@p0mvn Add more info about incentives, I'm not really familiar with the module, let me know if I miss anything or it's unclear.

x/cosmwasmpool/README.md
p0mvn: I was thinking that it might be useful from a security standpoint.

However, on more thought, I agree that it is not needed.

We can make only admin (module account) be allowed to migrate a contract. The migration is controlled by governance.

> store code and migrate

I agree on having this.

> store code and init pool

Why would we also init pool? Shouldn't that be done by `MsgCreateCosmWasmPool` separately?

x/cosmwasmpool/README.md
p0mvn: Sweet

x/cosmwasmpool/README.md
p0mvn: Confirmed that this was added

x/cosmwasmpool/README.md
p0mvn: `PoolI` is a pool interface that is implemented by pool models. These pool models are also written to state.

Examples of pool models:
- balancer
- stableswap
- concentrated liquidity

we are going to add a cosmwasm pool model.

`PoolModuleI` is an interface implemented by the SDK module keepers encapsulating these pool models.

We have:
- x/gamm keeper
- x/concentrated-liquidity keeper

we are going to add:
- x/cosmwasm keeper

x/cosmwasmpool/README.md
p0mvn: Once we agree on this, could you please update the spec for migration logic and gov props?

p0mvn:
> 

I propose that we do not integrate this pool type into x/twap for now. No specific code changes should be required for this but can you please update the spec? Should be as simple as mentioning that x/twap is not implemented for cosmwasm pools but can be in the future if there is a need

p0mvn:
TODO for me: check if we can use tokenfactory for representing shares

x/cosmwasmpool/README.md
iboss-ptk: > Why would we also init pool? Shouldn't that be done by MsgCreateCosmWasmPool separately?

`MsgCreateCosmWasmPool` only create pool out of existing code stored on chain.

x/cosmwasmpool/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+On contract's sudo enpoint, `SetActive` can be called to deactivate the pool. This will prevent the pool from being used for swap, and also prevent users from providing liquidity to the pool. Contract needs to check if the pool is active before performing any state mutating operation except `SetActive`.
+
+```rs
+SetActive {
+    is_active: bool,
+}
+```
+
+(TBD) On how to handle the deactivation operationally.
```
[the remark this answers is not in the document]
iboss-ptk: totally

p0mvn (approved):
LGTM. Thank you @iboss-ptk 

I will be submitting further updates in future PRs

Issue #4815, opened 2023-04-02
`TestClaimAllIncentives` is not passing on `main`


pysel:
#### Expected behaviour
Test passes

#### Actual behaviour
Test Fails for tc named `error: negative duration` with:
```
expected: "duration cannot be negative (-576h0m0s)"
in chain: "duration cannot be negative (-552h0m0s)"
```

#### Steps to reproduce the behaviour
Run this test on `main` branch, or run `go test ./...` from the project's root

PR #4822, opened 2023-04-03
Fix: `TestClaimAllIncentives` is not passing on `main`

ThanhNhann:
Closes: #4815 

- Fix fail of the last test case of `TestClaimAllIncentives`

## Brief Changelog
- With the old version, when we update `joinTime` with add 1 month, the number of days in month will change so the `expectedError` with change too  => fail
- I updated `joinTime` with 28 days instead (minimum), so it's ok now

## Testing and Verifying

- Tested and passed

  - Does this pull request introduce a new feature or user-facing behavior changes? ( no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? ( no)


p0mvn (approved):
Nice catch, thank you!

PR #4824, opened 2023-04-03
Generic Callbacks for hooks

AdityaSripal:
Closes: #XXX

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


x/ibc-hooks/wasm_hook.go
```diff
@@ -300,18 +303,35 @@ func (h WasmHooks) OnAcknowledgementPacketOverride(im IBCMiddleware, ctx sdk.Con
 		return err
 	}
 
-	if !h.ProperlyConfigured() {
```
nicolaslara, line 303: we probably still want to call `!h.ProperlyConfigured()` to make sure the module isn't incorrectly wired (which is easy to do because the wasm keeper needs to be passed after initialization)

x/ibc-hooks/wasm_hook.go
```diff
@@ -300,18 +303,35 @@ func (h WasmHooks) OnAcknowledgementPacketOverride(im IBCMiddleware, ctx sdk.Con
...
-	contract := h.ibcHooksKeeper.GetPacketCallback(ctx, packet.GetSourceChannel(), packet.GetSequence())
-	if contract == "" {
-		// No callback configured
+	callbackPacketData, ok := concretePacketData.(ibcexported.CallbackPacketData)
+	if !ok {
 		return nil
 	}
 
-	contractAddr, err := sdk.AccAddressFromBech32(contract)
+	contractAddrStr := callbackPacketData.GetSourceCallbackAddress()
```
nicolaslara, line 321: we also need to implement this function, right?

PR #4826, opened 2023-04-03
Protorev: Backrun pool join and exit messages

NotJeremyLiu:

- After this change, ProtoRev will be able to backrun join and exit pool messages  that swap within their own internal logic. 
- Previously, this was not the case and searchers were successfully arbing these messages while ProtoRev was live.

## Brief Changelog

- Antehandler now looks for -- MsgJoinSwapExternAmountIn, MsgJoinSwapShareAmountOut, MsgExitSwapExternAmountOut, MsgExitSwapShareAmountIn -- when extracting swapped pools.
- Adds two additional extraction logic functions to extract pools given join and exit messages

## Testing and Verifying

- All current tests still pass (after a slight modification since ExtractSwappedPools now uses the keeper)
- New unit tests added to ExtractSwappedPools function for all new message types supported

- Update protorev.md documentation to account for new messages supported

### DIFF
--- a/x/protorev/keeper/posthandler.go
+++ b/x/protorev/keeper/posthandler.go
@@ -50,7 +50,7 @@ func (protoRevDec ProtoRevDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simu
 	}
 
 	// Extract all of the pools that were swapped in the tx
-	swappedPools := ExtractSwappedPools(tx)
+	swappedPools := protoRevDec.ProtoRevKeeper.ExtractSwappedPools(ctx, tx)
 	if len(swappedPools) == 0 {
 		return next(ctx, tx, simulate)
 	}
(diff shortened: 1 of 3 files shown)


x/protorev/keeper/posthandler.go
```diff
@@ -193,3 +201,49 @@ func extractSwapOutPools(routes []poolmanagertypes.SwapAmountOutRoute, tokenOutD
...
+			PoolId:        poolId,
+			TokenOutDenom: coin.Denom,
+			TokenInDenom:  tokenInDenom})
+	}
+
+	return swappedPools
+}
+
+// extractExitPools extracts the pools that were swapped on for a MsgExitSwapExternAmountOut or MsgExitSwapShareAmountIn
+func (k Keeper) extractExitPools(ctx sdk.Context, poolId uint64, tokenOutDenom string) []SwapToBackrun {
```
bpiv400, line 229: These two functions are basically identical. You could copy a lot less code by changing the abstraction, so the caller indicates whether tokenIn or tokenOut denom is being passed.

NotJeremyLiu: Changed in: 22248313b1f2664e02e64a48607c1d3e1f36e83e

PR #4827, opened 2023-04-03
Protorev: Update timing for pool updating and dev fee payout

NotJeremyLiu:

This PR updates the timing cadence of certain protorev logic:
1. Changes dev account payment from once per week to after every trade (this is because the accumulation into the module account and taking from it has bad UX for those tracking protorev profits)
2. Changes highest liquidity pools updating from once per week to daily
3. Sets up protorev module for v16 upgrade migration

## Brief Changelog

- Uses daily epoch instead of weekly epoch for highest liquidity updating
- Removes all developer fee logic from epoch triggering into ExecuteTrade triggering
- Adds an upgrade migration in ProtoRev to payout what is left in the store when the v16 upgrade happens. This store will no longer be used with the new logic
- Marks (via comments) no longer used functions in v16 as deprecated , and also denotes functions used only for the upgrade as removable in v17

## Testing and Verifying

- Removed tests for no longer used methods
- Added a test for the new SendDeveloperFee method
- All other tests pass after setting the developer account in the tests
- Added checks in the TestUpgrade() function in v16 upgrade_test.go to ensure protorev upgrades properly (transfers all leftover dev fee funds in the store the the dev account)

- Changelog added
- Documentation has been updated to reflect new changes


NotJeremyLiu:
This commit is a minimal code change version of sending developer profit after every trade: 5bacd9ecbace4f968367a5d4589bc5ffb02bb3a2

It's not optimized for this change since it still stores state to track the dev fee (which is now unnecessary). Will be changed in a future commit to be an optimal version

bpiv400:
Should we interpret this as unfinished until the unnecessary state tracking is removed?

davidterpay (approved):
Not too familiar with what migration of developer fees away from ProtoRev in v16 looks like, so I would double check with the team. However, core changes LGTM.

NotJeremyLiu:
R4R besides merge conflicts

NotJeremyLiu:
Note to reviewers: The part I'm most unsure about is the upgrade/migration logic. Please take a careful look at that, thank you! I got it to work in testing through trial and error

NotJeremyLiu:
merge conflicts resolved

NotJeremyLiu:
Another note: The e2e tests are failing, I imagine related to the way I handled upgrades? Will need some assistance from Osmosis team on this.

stackman27:
> Another note: The e2e tests are failing, I imagine related to the way I handled upgrades? Will need some assistance from Osmosis team on this.

yup looking into this rn

app/upgrades/v16/upgrades.go
```diff
@@ -128,6 +128,12 @@ func CreateUpgradeHandler(
 
 		updateTokenFactoryParams(ctx, keepers.TokenFactoryKeeper)
 
+		// Transfers out all the dev fees in kvstore to dev account during upgrade
+		// Expceted to error during e2e test due to dev account not being set
```
p0mvn, line 132:
```suggestion
		// Expected to error during e2e test due to dev account not being set
```

app/upgrades/v16/upgrades.go
```diff
@@ -128,6 +128,12 @@ func CreateUpgradeHandler(
...
+		if err := keepers.ProtoRevKeeper.SendDeveloperFeesToDeveloperAccount(ctx); err != nil {
+			ctx.Logger().Error("error sending developer fees to developer account during upgrade, expected in testing, not in prod", "error", err)
+		}
```
p0mvn, line 135: I would prefer it if this is handled properly. I know the process for updating e2e is cumbersome. However, handling this properly will save us time during the upgrade if any issues occur.

The process for setting this account in e2e is the following:

1. update e2e genesis for the protorev module in a separate PR: https://github.com/osmosis-labs/osmosis/blob/1e56cc620a17e0b5debb8228896aa44b6421a723/tests/e2e/initialization/config.go#L234-L237
2. backport that PR to `v15.x`
3. Grab the chain init docker image with v15 tag from here: https://hub.docker.com/r/osmolabs/osmosis-e2e-init-chain/tags
  * should be auto-uploaded
4. Copy this tag: https://github.com/osmosis-labs/osmosis/blob/1e56cc620a17e0b5debb8228896aa44b6421a723/tests/e2e/containers/config.go#L27

stackman27: hi @NotJeremyLiu is this PR r4r now. I think just needs rebasing right?

NotJeremyLiu: @stackman27 I just updated the tag and removed the failing test, letting it run through workflow to see if all is good

tests/e2e/containers/config.go
```diff
@@ -24,7 +24,7 @@ const (
 	// It should be uploaded to Docker Hub. OSMOSIS_E2E_SKIP_UPGRADE should be unset
 	// for this functionality to be used.
 	previousVersionOsmoRepository = "osmolabs/osmosis-dev"
-	previousVersionOsmoTag        = "v15.x-833705e4-1679340262"
+	previousVersionOsmoTag        = "v15.x-63e3f053-1685053483"
```
p0mvn, line 27: please also bump `previousVersionInitTag` below to v15.x-833705e4-1679340262

NotJeremyLiu:
All checks pass, finally r4r @stackman27

stackman27 (approved):
LGTM

x/protorev/keeper/developer_fees.go
```diff
@@ -32,6 +33,7 @@ func (k Keeper) SendDeveloperFeesToDeveloperAccount(ctx sdk.Context) error {
 	return nil
 }
 
+// Deprecated: Can be removed in v16
 // UpdateDeveloperFees updates the fees that developers can withdraw from the module account
 func (k Keeper) UpdateDeveloperFees(ctx sdk.Context, denom string, profit sdk.Int) error {
```
p0mvn, line 38: Let's remove now?

p0mvn (approved):
LGTM

x/protorev/keeper/developer_fees.go
p0mvn: This is unused now. Please submit separate PR if agreed. Not going to block merge on this

PR #4829, opened 2023-04-04
Protorev query highest liquidity pools

NotJeremyLiu:

- Add visibility into protorev for the highest liquidity pool stored given a denom pair via a query endpoint

## Brief Changelog

- Adds rpc / cli queries that return a pool id given a denom pair to look up in protorev's highest liquidity pool method

## Testing and Verifying

- All tests still pass
- Added test to verify new query
- Verified CLI works via localosmosis

- Added new queries to documentation


NotJeremyLiu:
Added label for non-state breaking, don't know if it should be backport or no-backport

proto/osmosis/protorev/v1beta1/query.proto
```diff
@@ -108,6 +108,12 @@ service Query {
       returns (QueryGetProtoRevEnabledResponse) {
     option (google.api.http).get = "/osmosis/v14/protorev/enabled";
   }
+
+  // GetProtoRevPool queries whether the module is enabled or not
```
davidterpay, line 112: Do we want this to also return the hot routes as well or only the highest liquidity pool?
```suggestion
  // GetProtoRevPool queries the pool id used for arbitrage route building given a pair of denominations
```

davidterpay: If we only care about the highest liquidity pool id, I would update the doc to say that.

x/protorev/client/cli/query.go
```diff
@@ -148,6 +149,14 @@ func NewQueryPoolWeightsCmd() (*osmocli.QueryDescriptor, *types.QueryGetProtoRev
 	}, &types.QueryGetProtoRevPoolWeightsRequest{}
 }
 
+// NewQueryPoolCmd returns the command to query the pool id for a given denom pair
+func NewQueryPoolCmd() (*osmocli.QueryDescriptor, *types.QueryGetProtoRevPoolRequest) {
+	return &osmocli.QueryDescriptor{
+		Use:   "pool [base_denom] [other_denom]",
+		Short: "Query the pool id for a given denom pair stored in ProtoRev",
```
davidterpay, line 156: Update this based on the comment above.

davidterpay (commented):
LGTM with small nits.

x/protorev/client/cli/query.go
NotJeremyLiu: yeah, just want highest liquidity pools. Specified here: https://github.com/osmosis-labs/osmosis/pull/4829/commits/821b247669016401caec222de0743718f194e4e8

x/protorev/keeper/grpc_query_test.go
```diff
@@ -355,3 +355,25 @@ func (suite *KeeperTestSuite) TestGetProtoRevEnabledQuery() {
 	suite.Require().NoError(err)
 	suite.Require().Equal(enabled, res.Enabled)
 }
+
+// TestGetProtoRevStatisticsByRoute tests the query for statistics by route
```
p0mvn, line 359:
```suggestion
// TestGetProtoRevPool tests the query for statistics by route
```

p0mvn (approved):
LGTM

x/protorev/keeper/grpc_query_test.go
NotJeremyLiu: fixed here: https://github.com/osmosis-labs/osmosis/pull/4829/commits/2dbb66767fd88b55d3713610d7cbb0953f0d859b

p0mvn:
Let's add a changelog entry for this and merge

NotJeremyLiu:
> Let's add a changelog entry for this and merge

Added in: https://github.com/osmosis-labs/osmosis/pull/4829/commits/7049f018ac75d82185bddc1819d0722e041c51a0 and rebased on main, so should be ready to merge.

Issue #4738, opened 2023-03-24
[Incentives Module][bugfix] Scale gas costs by denoms in gauge


AlpinYukseloglu:
## Background

Similarly to what we did for incentive records in CL, we should be scaling gas costs for adding to gauges to prevent spam, as distribution logic linearly iterates over all denoms in a gauge at epoch.

## Suggested Design

- Add scaling gas costs to `AddToGaugeRewards`, similar to how we did it for incentive records

## Acceptance Criteria

- New and existing tests pass

PR #4830, opened 2023-04-04
[Incentives Module][bugfix] Scale gas costs by denoms in gauge 

stackman27:
[Closes: #4738](https://github.com/osmosis-labs/osmosis/issues/4738)

Similarly to what we did for incentive records in CL, we should be scaling gas costs for adding to gauges to prevent spam, as distribution logic linearly iterates over all denoms in a gauge at epoch.

Add scaling gas costs to AddToGaugeRewards, similar to how we did it for incentive records

## Brief Changelog

- Add gas cost when we `AddToGaugeRewards`, linearly increase with coins to add

## Testing and Verifying
added new tests


x/incentives/keeper/gauge.go
```diff
@@ -152,6 +152,10 @@ func (k Keeper) AddToGaugeRewards(ctx sdk.Context, owner sdk.AccAddress, coins s
 	if gauge.IsFinishedGauge(ctx.BlockTime()) {
 		return errors.New("gauge is already completed")
 	}
+
+	// Fixed gas consumption adding reward to gauges based on the number of coins to add
+	ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForAddRewardToGauge*len(coins)), "add to gauge reward cost")
```
AlpinYukseloglu, line 157:
```suggestion
	ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForAddRewardToGauge*len(coins)), "scaling gas cost for adding to gauge rewards")
```

x/incentives/keeper/gauge_test.go
```diff
@@ -332,3 +332,80 @@ func (suite *KeeperTestSuite) TestChargeFeeIfSufficientFeeDenomBalance() {
...
+	}{
+		{
+			name:  "valid case: valid gauge",
+			owner: suite.TestAccs[0],
+			coinsToAdd: sdk.NewCoins(
+				sdk.NewCoin("uosmo", sdk.NewInt(100000)),
+				sdk.NewCoin("atom", sdk.NewInt(99999)),
+			),
+			gaugeId:            1,
+			minimumGasConsumed: uint64(2 * types.BaseGasFeeForAddRewardToGauge),
```
AlpinYukseloglu, line 354: Since the base gas cost for txns like these are usually in the 10k-20k gas range, we should add a test case with many denoms (>4) to ensure that the `minimumGasConsumed` check isn't passing trivially

AlpinYukseloglu (approved):
LGTM! Left a couple minor comments. Approving rn, but we should add the test case I suggested before merging.

x/incentives/keeper/gauge_test.go
stackman27: sounds good added! for 8 tokens it looks like this; 

```
gasConsumed: 161441  tc. minimumGasConsumed: 80000
```

x/incentives/keeper/gauge.go
```diff
@@ -152,6 +152,10 @@ func (k Keeper) AddToGaugeRewards(ctx sdk.Context, owner sdk.AccAddress, coins s
 	if gauge.IsFinishedGauge(ctx.BlockTime()) {
 		return errors.New("gauge is already completed")
 	}
+
+	// Fixed gas consumption adding reward to gauges based on the number of coins to add
+	ctx.GasMeter().ConsumeGas(uint64(types.BaseGasFeeForAddRewardToGauge*len(coins)), "scaling gas cost for adding to gauge rewards")
```
AlpinYukseloglu, line 157: Just realized that this should actually be charging `baseAmount * (len(gauge.Coins) + len(coins)` since we allow for adding multiple coins at once here. The current implementation only scales by the number of coins being added, not the ones that already exist.

stackman27: i see, i originally thought it was just for new denoms to add. I'll change it to accomodate existing denoms

p0mvn (approved):

p0mvn:
@stackman27 please add a CHANGELOG entry prior to merge

stackman27:
> @stackman27 please add a CHANGELOG entry prior to merge

added!

p0mvn (approved):
Nice

Issue #4836, opened 2023-04-05
[CL]: Treat balancer pool as a full range CL position for incentives


czarcas7ic:
## Background

After discussion, we do not want to completely remove incentives from balancer pools. That being said, we should still be ushering people towards the better CL pool type.

The first step in achieving this is treating the liquidity in the canonically linked balancer pool as a full range liquidity position, and distribute incentives accordingly to that balancer pool's gauge.

## Suggested Design

- Whenever we update CL incentive accumulators:
  - Calculate the balancer bonded and unbonding liquidity that qualifies 
  - "Add" to current accumulator
  - Create a record on the current accumulator
  - Run the usual logic
  - Claim the rewards for this record
  - AddToGauge these rewards
  - Delete the record

We must also ensure that the incentive record on the balancer pool is removed, and all incentives for that denom pair should now be going to the canonical CL pool

## Acceptance Criteria

- A full range CL pool position should be getting the same reward payout as its balancer pool equivalent

PR #4853, opened 2023-04-07
[CL][Incentives] Implement helper for clearing full range balancer shares from CL pool

AlpinYukseloglu:
Progress towards: #4836
> Note to reviewer: this PR just implements abstractions/helpers and does not actually call them in core CL code. This is to keep PRs more manageable in size.

This PR implements `claimAndResetFullRangeBalancerPool`, which is the second and final abstraction needed to implement reward splitting. This function handles claiming rewards for the "full range shares" of the canonical balancer pool for a CL pool, routes these rewards to the correct incentive gauge, and then clears the shares from the CL pool's uptime accumulators. The godoc explains the intended functionality of the abstraction in greater detail.

## Brief Changelog

- Implement `claimAndResetFullRangeBalancerPool`

## Testing and Verifying

- All changes are tested in `incentives_test.go`


x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+				normalizedEmissions := sdk.NormalizeCoins(decEmissions)
+				s.FundAcc(clPool.GetIncentivesAddress(), normalizedEmissions)
+			}
+			addToUptimeAccums(s.Ctx, clPool.GetId(), s.App.ConcentratedLiquidityKeeper, tc.uptimeGrowth)
+
+			// --- System under test ---
+
+			amountClaimed, err := s.App.ConcentratedLiquidityKeeper.ClaimAndResetFullRangeBalancerPool(s.Ctx, clPoolId, balancerPoolId)
+
+			// --- Assertions ---
```
czarcas7ic, line 3681: This is a great test so far, but wanted to note that I think we really need a functional test for this where we make assertions that not only the balancer gauge has the proper rewards, but also the cl accums themselves retain the expected rewards still and emit properly to cl LPs (can be future PR)

czarcas7ic (approved):
Nice work on this 👍

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,91 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+		balancerPositionName := string(types.KeyBalancerFullRange(clPoolId, balPoolId, uint64(uptimeIndex)))
+
+		// Ensure that the given balancer pool has a record on the given uptime accumulator
+		recordExists, err := uptimeAccum.HasPosition(balancerPositionName)
+		if err != nil {
+			return sdk.Coins{}, err
+		}
+		if !recordExists {
+			return sdk.Coins{}, types.BalancerRecordNotFoundError{ClPoolId: clPoolId, BalancerPoolId: balPoolId, UptimeIndex: uint64(uptimeIndex)}
+		}
```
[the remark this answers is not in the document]
AlpinYukseloglu: Good point, added comment

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,91 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+// Returns the number of coins that were claimed and distrbuted.
+// Returns error if either reward claiming, record deletion or adding to the gauge fails.
+func (k Keeper) claimAndResetFullRangeBalancerPool(ctx sdk.Context, clPoolId uint64, balPoolId uint64) (sdk.Coins, error) {
+	// Get CL pool from ID. This also serves as an early pool existence check.
+	clPool, err := k.getPoolById(ctx, clPoolId)
+	if err != nil {
+		return sdk.Coins{}, err
+	}
+
+	// Get longest lockup period for pool
```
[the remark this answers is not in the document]
AlpinYukseloglu: Expanded the godoc for the function to emphasize this more strongly

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,91 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+// pool.
+//
+// Returns the number of coins that were claimed and distrbuted.
+// Returns error if either reward claiming, record deletion or adding to the gauge fails.
+func (k Keeper) claimAndResetFullRangeBalancerPool(ctx sdk.Context, clPoolId uint64, balPoolId uint64) (sdk.Coins, error) {
+	// Get CL pool from ID. This also serves as an early pool existence check.
+	clPool, err := k.getPoolById(ctx, clPoolId)
+	if err != nil {
+		return sdk.Coins{}, err
+	}
```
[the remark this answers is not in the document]
AlpinYukseloglu: I figured since we use `prepareBalancerAsFullRange ` and `claimAndResetFullRangeBalancerPool` in tandem, it would save on state reads/writes (and computation) to have the latter require the caller to specify a valid CL/bal ID pair. The case where the pools are unlinked (or nonexistent) is still caught in this implementation, so from a safety standpoint it's the same (although generally more efficient).

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: I agree – this is a broader point on invariant testing for incentives, which we absolutely should do. Created issue here: #4863

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,94 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+
+	// Get longest lockup period for pool
+	lockableDurations := k.poolIncentivesKeeper.GetLockableDurations(ctx)
+	longestLockableDuration := lockableDurations[len(lockableDurations)-1]
```
p0mvn, line 213: What if the longest balancer duration becomes smaller than the largest uptime accumulator created? How is this case handled?

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,94 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+		recordExists, err := uptimeAccum.HasPosition(balancerPositionName)
+		if err != nil {
+			return sdk.Coins{}, err
+		}
+		if !recordExists {
+			return sdk.Coins{}, types.BalancerRecordNotFoundError{ClPoolId: clPoolId, BalancerPoolId: balPoolId, UptimeIndex: uint64(uptimeIndex)}
+		}
+
+		// Remove shares from record so it gets cleared when rewards are claimed
+		numShares, err := uptimeAccum.GetPositionSize(balancerPositionName)
```
p0mvn, line 247: I'm guessing the assumption is that the `numShares` is up to date because we called `prepareBalancerAsFullRange` prior to this method. Might be good to document that with a comment too

x/concentrated-liquidity/incentives.go
```diff
@@ -193,6 +193,94 @@ func (k Keeper) prepareBalancerPoolAsFullRange(ctx sdk.Context, clPoolId uint64)
...
+			return sdk.Coins{}, err
+		}
+		if recordExists {
+			return sdk.Coins{}, types.BalancerRecordNotClearedError{ClPoolId: clPoolId, BalancerPoolId: balPoolId, UptimeIndex: uint64(uptimeIndex)}
+		}
+	}
+
+	// After claiming accrued rewards from all uptime accumulators, add the total claimed amount to the
+	// Balancer pool's longest duration gauge.
+	err = k.incentivesKeeper.AddToGaugeRewards(ctx, clPool.GetIncentivesAddress(), totalRewards, gaugeId)
```
p0mvn, line 276: Following up on the previous comment, I think we need to make sure that the largest uptime safely maps to the largest lockable duration / gauge all the time.

p0mvn, line 276: Let's surround this with `if !totalRewards.IsZero()` please.

I'm seeing that there are hooks that run "after adding to gauge". Would be good to avoid triggering listeners redundantly when nothing actually gets added to gauge

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+
+				return
+			}
+
+			s.Require().NoError(err)
+
+			clPoolUptimeAccumulators, err := s.App.ConcentratedLiquidityKeeper.GetUptimeAccumulators(s.Ctx, clPool.GetId())
+			s.Require().NoError(err)
+
+			for _, uptimeAccum := range clPoolUptimeAccumulators {
```
p0mvn, line 3723: Let's add a check that `clPoolUptimeAccumulators` are not empty so that  assertions inside the loop aren't accidentally skipped

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+
+			// Calculate the number of tokens we expect to see in the balancer gauge
+			expectedTokensInGauge := expectedIncentivesFromUptimeGrowth(tc.uptimeGrowth, addedLiquidity, longestSupportedUptime, defaultMultiplier)
+
+			s.Require().Equal(expectedTokensInGauge, gauge.Coins)
+
+			// Pool liquidity should remain unchanged
+			updatedClPool, err := s.App.ConcentratedLiquidityKeeper.GetPoolById(s.Ctx, clPool.GetId())
+			s.Require().NoError(err)
+			s.Require().Equal(initialLiquidity, updatedClPool.GetLiquidity())
```
p0mvn, line 3748: Is there an assertion that uptime accumulators for full range balancer position are emptied? If not, please add it prior to merge

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+			balancerPoolAssets:  defaultBalancerAssets,
+			uptimeGrowth:        uptimeHelper.hundredTokensMultiDenom,
+		},
+		"valid pool IDs with no uptime growth": {
+			// 100 existing shares and 100 shares added from balancer
+			existingClLiquidity: defaultConcentratedAssets,
+			balancerPoolAssets:  defaultBalancerAssets,
+			uptimeGrowth:        uptimeHelper.emptyExpectedAccumValues,
+		},
+		"valid pool IDs with uneven uptime growth": {
```
p0mvn, line 3543: What makes it uneven?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
 	}
 }
+
+func (s *KeeperTestSuite) TestClaimAndResetFullRangeBalancerPool() {
+	lockableDurations := s.App.PoolIncentivesKeeper.GetLockableDurations(s.Ctx)
+	longestLockableDuration := lockableDurations[len(lockableDurations)-1]
+	uptimeHelper := getExpectedUptimes()
+
+	tests := map[string]struct {
+		existingClLiquidity sdk.Coins
```
p0mvn, line 3521: This is a general suggestion. Let's try to avoid referencing to CL as "CL" anywhere in the codebase please. Instead, expand the abbreviation. I bet anyone new to the codebase would have to spend cycles trying to figure out what various abbreviations mean in a couple of months

```suggestion
		existingConcentratedLiquidity sdk.Coins
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+			for _, growth := range tc.uptimeGrowth {
+				decEmissions := growth.MulDec(initialLiquidity.Add(addedLiquidity))
+				normalizedEmissions := sdk.NormalizeCoins(decEmissions)
+				s.FundAcc(clPool.GetIncentivesAddress(), normalizedEmissions)
+			}
+			addToUptimeAccums(s.Ctx, clPool.GetId(), s.App.ConcentratedLiquidityKeeper, tc.uptimeGrowth)
+
+			// --- System under test ---
+
+			amountClaimed, err := s.App.ConcentratedLiquidityKeeper.ClaimAndResetFullRangeBalancerPool(s.Ctx, clPoolId, balancerPoolId)
```
p0mvn, line 3679: I'm not seeing an assertion that `amountClaimed` is equal to expected when there is no error. IMO, this is blocking for this PR.

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3509,3 +3511,241 @@ func (s *KeeperTestSuite) TestPrepareBalancerPoolAsFullRange() {
...
+		"Balancer shares not yet added to CL pool accums": {
+			// 100 existing shares and 100 shares added from balancer
+			existingClLiquidity: defaultConcentratedAssets,
+			balancerPoolAssets:  defaultBalancerAssets,
+			uptimeGrowth:        uptimeHelper.hundredTokensMultiDenom,
+
+			balSharesNotAddedToAccums: true,
+			expectedError:             types.BalancerRecordNotFoundError{ClPoolId: 1, BalancerPoolId: 2, UptimeIndex: uint64(0)},
+		},
+	}
```
p0mvn, line 3623: Would be good to have a test where claimed rewards end up being greater than what's available in the incentives address

p0mvn (commented):
Great work.

Left some comments, please let me know what you think

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Right – I did expand on this in comments earlier in the function but it might be worthwhile to have more frequent reminders throughout

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: The longest balancer duration would still be getting the rewards, in the same way that if governance were incentivizing that pool it would choose the longest duration (even if it were, say, just one week).

Given how even a shorter bond is much, much stricter than a non-binding charging period, this doesn't seem problematic to me. Am I missing something?

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Why do we need to ensure this?

AlpinYukseloglu: Good point!

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: `varyingTokensMultiDenom` has different amounts for each denom

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: There's an interesting tradeoff here given how verbose "ConcentratedLiquidity" is, but I generally agree. Will adjust in this PR and keep in mind in the future

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Right – we currently check that `expectedAmount = amtAddedToGauge` when we should be checking `expectedAmount = amtAddedToGauge = amountClaimed`, nice catch

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: `clPoolUptimeAccumulators` can never be empty. Will add this check, but in general I think adding too many extraneous checks make tests much less readable/can be misleading towards the functionality of these components (e.g. someone might assume that it is possible for `clPoolUptimeAccumulators` to be empty since we check it)

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: I'm assuming you mean that the full range records have no more shares and are cleared from state. We check the former above in the following section:
```
// Ensure each accum has been cleared of the balancer full range shares
s.Require().Equal(initialLiquidity, currAccumShares)
```

Will add a check for the latter

x/concentrated-liquidity/incentives.go
p0mvn: No, this makes sense. Thanks

x/concentrated-liquidity/incentives.go
p0mvn: I was overthinking, another reply of yours helped me realize this is unnecessary

x/concentrated-liquidity/incentives_test.go
p0mvn: > clPoolUptimeAccumulators can never be empty. 

Thanks for adding it. I understand that this is the case today, and, currently, it does not make sense for them to be empty.

However, as the system evolves a lot of unexpected changes can happen. A few months from now,  I might not remember this guarantee. So a check like this would help me be confident that the assertions in the loop are running with no additional manual work.

I've also been noticing a lot of instances where we get false positive passes due to the lack of such checks so I definitely think it's useful.

x/concentrated-liquidity/incentives_test.go
p0mvn: I think having `osmoassert.ConditionalPanic` and an extra boolean `expectedPanic` in the test case struct is sufficient to cover this test case.

Another approach is to manually check the balancers in the function and return an error if the balance is insufficient

p0mvn (approved):
LGTM. Great work

There seem to be a few minor action items left in the discussions. Please review prior to merge

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: Turns out the panic is actually caught at a lower level of abstraction so it was actually even simpler than that to add!

PR #4856, opened 2023-04-07
remove redundant line

DongLieu:
adding this line is not necessary:
[`cmd.Short = fmt.Sprintf("Querying commands for the %s module", types.ModuleName)`](https://github.com/osmosis-labs/osmosis/blob/main/x/ibc-hooks/client/cli/query.go#L32)

because the above already has:
[`Short:                      fmt.Sprintf("Querying commands for the %s module", types.ModuleName),`](https://github.com/osmosis-labs/osmosis/blob/main/x/ibc-hooks/client/cli/query.go#L38)


pysel (approved):
LGTM! Thanks. Could you please apply backport labels to v15

PR #4861, opened 2023-04-07
fix(CL): invalid sqrt price initialization breaking liquidity net in direction query

p0mvn:
Closes: #XXX

Liquidity net in direction query was breaking in some cases due to a failing current square root price check.

Apparently, the pool's current square root price did not translate to the current tick correctly.

This is because we used an inaccurate `ApproxSqrt` function in initialization. Instead, this change initializes the current square root price by using our internal math functions to translate it from the current tick.

On quick test, was unable to reproduce the problem anymore.

Also, added some debug logs to the query for easier debugging if the error persists.

## Testing and Verifying

### DIFF
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -306,7 +306,7 @@ func (k Keeper) initializeInitialPositionForPool(ctx sdk.Context, pool types.Con
 		return types.InitialLiquidityZeroError{Amount0: amount0Desired, Amount1: amount1Desired}
 	}
 
-	initialSpotPrice := osmomath.NewBigDec(amount1Desired.Int64()).Quo(osmomath.NewBigDec(amount0Desired.Int64()))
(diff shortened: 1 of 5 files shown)


x/concentrated-liquidity/lp.go
```diff
@@ -305,15 +306,17 @@ func (k Keeper) initializeInitialPositionForPool(ctx sdk.Context, pool types.Con
...
+
+	// Calculate the initial tick from the initial spot price
+	initialTick, err := math.PriceToTick(initialSpotPrice.SDKDec(), pool.GetExponentAtPriceOne())
 	if err != nil {
 		return err
 	}
 
-	// Calculate the initial tick from the initial spot price
-	initialTick, err := math.PriceToTick(initialSpotPrice, pool.GetExponentAtPriceOne())
+	// Re-Calculate the spot price from initial tick so that it is aligned with out internal
```
czarcas7ic, line 317:
```suggestion
	// Re-Calculate the spot price from initial tick so that it is aligned with our internal
```

czarcas7ic (approved):
Good find on the sqrt function. My only question is if we need to use BigDec to determine the initialSpotPrice, but I suppose this shouldn't be a large overhead since its only run once on pool initialization.

Issue #4225, opened 2023-02-05
estimate-swap-exact-amount return Error: accepts 4 arg(s), received 3


EzrielS:
#### System information

Osmosis version: 14.0.0
OS & Version: Linux

#### Expected behaviour
`osmosisd q gamm estimate-swap-exact-amount-in 722 osmo1xwazl8ftks4gn00y5x3c47auquc62ssuds42sz 9746109223341728563ibc/6AE98883D4D5D5FF9E50D7130F1305DA2FFA0C652D1DD9C123657C6B4EB2DF8A --swap-route-pool-ids 722,678 --swap-route-denoms uosmo,ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858 --output json` should work.

#### Actual behaviour

It return `Error: accepts 4 arg(s), received 3`

mattverse:
Thanks for the issue! Can you confirm if you're binary is in the right version? (Latest Mainnet verison, which is V14)

I have tested this locally on the latest binary and would  return error if the given command is not 3 argument.

EzrielS:
Thank you for your reply, when I do `osmosisd version` I get 14.0.0 too. Maybe it's not the "latest 14.0.0" release ? I don'd know if there are such things.

mattverse:
Interesting, what do you get when you do the following query?
```
osmosisd q gamm estimate-swap-exact-amount-in 722 osmo1xwazl8ftks4gn00y5x3c47auquc62ssuds42sz 9746109223341728563ibc/6AE98883D4D5D5FF9E50D7130F1305DA2FFA0C652D1DD9C123657C6B4EB2DF8A --swap-route-pool-ids 722 --swap-route-denoms uosmo,ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858 --output json
```

mattverse:
It should not be omitting same error, could you double check if you do not have duplciated binary or if you have properly set GOPATHs?

EzrielS:
It's possible, I used `source <(curl -sL https://get.osmosis.zone/run)` but I had some problems at the installation, but the node seems to work well. How can I check if the binary is duplicated ? `whereis osmosisd` returns only one, and it seems to be the good one.

FruityFranck:
Hello, had the same error
I think the fix is on the top of v14.x branch but not on v14.0.0 tag.
[...]

PR #4877, opened 2023-04-10
rename QuerySwapExactAmountOut/InRequest and QuerySwapExactAmountOut/InResponse

DongLieu:

In osmocli Query there is func [ParseExpectedQueryFnName()](https://github.com/osmosis-labs/osmosis/blob/main/osmoutils/osmocli/parsers.go#L49) which can return QueryFnName, however it doesn't work for QuerySwapExactAmountInRequest and QuerySwapExactAmountOutRequest because its grpc is [EstimateSwapExactAmountIn](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/grpc_query.go#L426) and[ EstimateSwapExactAmountOut](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/grpc_query.go#L455).

The renaming will be optimal for general logic usage and future bug fixes and upgrades

refer #4225

## Brief Changelog

- Rename QuerySwapExactAmountInRequest to QueryEstimateSwapExactAmountInRequest
- Rename QuerySwapExactAmountOutRequest to QueryEstimateSwapExactAmountOutRequest

- Rename QuerySwapExactAmountInResponse to QueryEstimateSwapExactAmountInResponse
- Rename QuerySwapExactAmountOutResponse to QueryEstimateSwapExactAmountOutResponse

- remove the QueryFnName in [GetCmdEstimateSwapExactAmountIn()](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/client/cli/query.go#L216)
- remove the QueryFnName in [GetCmdEstimateSwapExactAmountOut()](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/client/cli/query.go#L230)

## Testing and Verifying

All current tests pass after renaming, and re-running Query estimate-swap-exact-amount-out and Query estimate-swap-exact-amount-in work properly

golangci-lint run --fix


DongLieu:
Ok, if that's the case, keep the name the same, but if there's a chance I think we should change the name accordingly

p0mvn:
> 

Agreed, thanks for trying to make this better. Sadly, we have to be locked in for backward-compatibility

p0mvn:
Per-agreement, closing for now unless a non-rename solution is found. Please reopen if so

PR #4878, opened 2023-04-10
Protorev: Backrun event emission

NotJeremyLiu:

- Adds protorev-specific event emission when the module successfully backruns to make it easier to index for tracking, research, and visualization purposes.

## Brief Changelog

- Emits a protorev_backrun event upon successful backrunning of a transaction. Emits details about 1) user denoms, 2) pool points remaining, 3) profit and denom, 4) amount in and out to execute trade, and 5) tx_hash

## Testing and Verifying

- All tests current tests pass after changing input params necessary for event
- Added unit test for the added event

- Added event section to protorev.md and added the protorev_backrun event type and attrs.

### DIFF
--- a/x/protorev/keeper/emit.go
+++ b/x/protorev/keeper/emit.go
@@ -1,0 +1,55 @@
+package keeper
+
+import (
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v15/x/protorev/types"
+)
+
+// EmitBackrunEvent updates and emits a backrunEvent
+func EmitBackrunEvent(ctx sdk.Context, backrunEvent sdk.Event, inputCoin sdk.Coin, profit, tokenOutAmount sdk.Int) {
+	// Update the backrun event and add it to the context
+	backrunEvent = backrunEvent.AppendAttributes(
+		sdk.NewAttribute(types.AttributeKeyProtorevProfit, profit.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevAmountIn, inputCoin.Amount.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevAmountOut, tokenOutAmount.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevArbDenom, inputCoin.Denom),
+	)
+	ctx.EventManager().EmitEvent(backrunEvent)
+}
+
+// CreateBackrunEvent creates a backrun event to be emitted if the trade is executed successfully
+func (k Keeper) CreateBackrunEvent(ctx sdk.Context, pool SwapToBackrun, remainingTxPoolPoints uint64) (sdk.Event, error) {
+	// Get pool points remaning in block
+	remainingBlockPoolPoints, err := k.remainingPoolPointsForBlock(ctx)
+	if err != nil {
+		return sdk.Event{}, err
+	}
+
+	// Create backrun event to be emitted if the trade is executed successfully
+	return sdk.NewEvent(
+		types.TypeEvtBackrun,
+		sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		sdk.NewAttribute(types.AttributeKeyUserDenomIn, pool.TokenInDenom),
+		sdk.NewAttribute(types.AttributeKeyUserDenomOut, pool.TokenOutDenom),
+		sdk.NewAttribute(types.AttributeKeyTxPoolPointsRemaining, strconv.FormatUint(remainingTxPoolPoints, 10)),
+		sdk.NewAttribute(types.AttributeKeyBlockPoolPointsRemaining, strconv.FormatUint(remainingBlockPoolPoints, 10)),
+	), nil
+}
+
+// RemainingPoolPointsForBlock calculates the number of pool points that can be consumed in the current block.
+func (k Keeper) remainingPoolPointsForBlock(ctx sdk.Context) (uint64, error) {
+	maxRoutesPerBlock, err := k.GetMaxPointsPerBlock(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	currentRouteCount, err := k.GetPointCountForBlock(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	return maxRoutesPerBlock - currentRouteCount, nil
+}
--- a/x/protorev/keeper/emit.go
+++ b/x/protorev/keeper/emit.go
@@ -13,28 +13,16 @@ import (
 )
 
 // EmitBackrunEvent updates and emits a backrunEvent
-func EmitBackrunEvent(ctx sdk.Context, backrunEvent sdk.Event, inputCoin sdk.Coin, profit, tokenOutAmount sdk.Int) {
-	// Update the backrun event and add it to the context
-	backrunEvent = backrunEvent.AppendAttributes(
-		sdk.NewAttribute(types.AttributeKeyProtorevProfit, profit.String()),
-		sdk.NewAttribute(types.AttributeKeyProtorevAmountIn, inputCoin.Amount.String()),
-		sdk.NewAttribute(types.AttributeKeyProtorevAmountOut, tokenOutAmount.String()),
-		sdk.NewAttribute(types.AttributeKeyProtorevArbDenom, inputCoin.Denom),
-	)
-	ctx.EventManager().EmitEvent(backrunEvent)
-}
-
-// CreateBackrunEvent creates a backrun event to be emitted if the trade is executed successfully
-func (k Keeper) CreateBackrunEvent(ctx sdk.Context, pool SwapToBackrun, remainingTxPoolPoints uint64) (sdk.Event, error) {
+func (k Keeper) EmitBackrunEvent(ctx sdk.Context, pool SwapToBackrun, inputCoin sdk.Coin, profit, tokenOutAmount sdk.Int, remainingTxPoolPoints uint64) error {
 	// Get pool points remaning in block
 	remainingBlockPoolPoints, err := k.remainingPoolPointsForBlock(ctx)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


x/protorev/types/events.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+const (
+	TypeEvtBackrun = "protorev_backrun"
+
+	AttributeValueCategory               = ModuleName
```
davidterpay, line 6: I think it would be a nice addition to also emit the hash of the transaction that is being backrun.
```suggestion
	AttributeValueCategory               = ModuleName
    AttributeKeyTransactionHash          
```suggestion
	AttributeValueCategory               = ModuleName
```= "tx_hash"
```

x/protorev/keeper/emit.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/osmosis-labs/osmosis/v15/x/protorev/types"
+)
+
+// EmitBackrunEvent updates and emits a backrunEvent
+func EmitBackrunEvent(ctx sdk.Context, backrunEvent sdk.Event, inputCoin sdk.Coin, profit, tokenOutAmount sdk.Int) {
```
davidterpay, line 12: Is there a reason we want to separate `EmitBackrunEvent` and `CreateBackrunEvent` instead of just wrapping it into a single function?

x/protorev/keeper/posthandler.go
```diff
@@ -128,7 +128,13 @@ func (k Keeper) ProtoRevTrade(ctx sdk.Context, swappedPools []SwapToBackrun) (er
 
 		// The error that returns here is particularly focused on the minting/burning of coins, and the execution of the MultiHopSwapExactAmountIn.
 		if maxProfitAmount.GT(sdk.ZeroInt()) {
-			if err := k.ExecuteTrade(ctx, optimalRoute, maxProfitInputCoin); err != nil {
+			// Create backrun event to be emitted if the trade is executed successfully
+			backrunEvent, err := k.CreateBackrunEvent(ctx, pool, remainingPoolPoints)
+			if err != nil {
+				return err
+			}
```
davidterpay, line 135: cc: the comment about merging the creation and emission of event into a single function.

Why are we building the event here and appending more attributes later instead of just building the entire event in one place?

davidterpay: imo I would just pass in the parameters `pool` and `remainingPoolPoints` into execute trade.

x/protorev/keeper/emit.go
```diff
@@ -0,0 +1,56 @@
...
+		sdk.NewAttribute(types.AttributeKeyUserDenomIn, pool.TokenInDenom),
+		sdk.NewAttribute(types.AttributeKeyUserDenomOut, pool.TokenOutDenom),
+		sdk.NewAttribute(types.AttributeKeyTxPoolPointsRemaining, strconv.FormatUint(remainingTxPoolPoints, 10)),
+		sdk.NewAttribute(types.AttributeKeyBlockPoolPointsRemaining, strconv.FormatUint(remainingBlockPoolPoints, 10)),
+	), nil
+}
+
+// RemainingPoolPointsForBlock calculates the number of pool points that can be consumed in the current block.
+func (k Keeper) remainingPoolPointsForBlock(ctx sdk.Context) (uint64, error) {
+	maxRoutesPerBlock, err := k.GetMaxPointsPerBlock(ctx)
```
davidterpay, line 45:
```suggestion
	maxPoolPointsPerBlock, err := k.GetMaxPointsPerBlock(ctx)
```

x/protorev/keeper/emit.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	currentRouteCount, err := k.GetPointCountForBlock(ctx)
```
davidterpay, line 50:
```suggestion
	currentPoolPointCount, err := k.GetPointCountForBlock(ctx)
```

davidterpay (commented):
Update LGTM, but would move the creation and event emission into the same place.

ValarDragon:
We discussed on call that we are concerned that the keeper call here may change gas. Resolution on call I believe was to:
* merge this PR
* on v15 backport be careful to ensure that we delete the getter and usage of that getter in the event

NotJeremyLiu:
> We discussed on call that we are concerned that the keeper call here may change gas. Resolution on call I believe was to:
> 
> * merge this PR
> * on v15 backport be careful to ensure that we delete the getter and usage of that getter in the event

I'm going to spend a bit of time today refactoring to not include an extra kvstore read, but to see if we can still output the block pool points. Tbh, this is the main reason why we want this event as it will allow us to know how much room we still have to per block to backrun more or less.

x/protorev/keeper/posthandler.go
NotJeremyLiu: Yeah I agree, the reason I did it this way was that I only had to pass in the event as a param to append to, but I like the idea of making one function that does it all in one place.

x/protorev/types/events.go
NotJeremyLiu: added in e89701b9b3c6e2638a341f3c41364e94447783a7

x/protorev/keeper/emit.go
NotJeremyLiu: changed in d2579c751e9767980863ae1b356eab7f4698a882

x/protorev/keeper/emit.go
NotJeremyLiu: changed in d2579c751e9767980863ae1b356eab7f4698a882

x/protorev/keeper/posthandler.go
NotJeremyLiu: combined here: 29f182574e40d3eb59d71047f59ad883835d8ed3

NotJeremyLiu:
Need to add tests to verify, but this commit intends to allow us to emit blockPoolPointsRemaining without doing an extra kvstore read: https://github.com/osmosis-labs/osmosis/pull/4878/commits/482cbf636602590252182e9a34b126c722877626

NotJeremyLiu:
Test added, ready for review

x/protorev/keeper/emit.go
```diff
@@ -0,0 +1,37 @@
...
+		sdk.NewAttribute(types.AttributeKeyBlockPoolPointsRemaining, strconv.FormatUint(remainingBlockPoolPoints, 10)),
+		sdk.NewAttribute(types.AttributeKeyProtorevProfit, profit.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevAmountIn, inputCoin.Amount.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevAmountOut, tokenOutAmount.String()),
+		sdk.NewAttribute(types.AttributeKeyProtorevArbDenom, inputCoin.Denom),
+	)
+	ctx.EventManager().EmitEvent(backrunEvent)
+
+	return nil
+}
```
p0mvn, line 37: There does not seem to be an error returned, can we remove `error` return?

x/protorev/keeper/rebalance.go
```diff
@@ -220,37 +222,43 @@ func (k Keeper) ExecuteTrade(ctx sdk.Context, route poolmanagertypes.SwapAmountI
 		return err
 	}
 
+	// Create and emit the backrun event and add it to the context
+	err = EmitBackrunEvent(ctx, pool, inputCoin, profit, tokenOutAmount, remainingTxPoolPoints, remainingBlockPoolPoints)
+	if err != nil {
+		return err
+	}
```
p0mvn, line 229: Note that adding error checks can be state-breaking. However, it does not seem that `EmitBackrunEvent` actually ever returns an error so we can probably remove it

x/protorev/keeper/rebalance.go
```diff
@@ -220,37 +222,43 @@ func (k Keeper) ExecuteTrade(ctx sdk.Context, route poolmanagertypes.SwapAmountI
...
-	numberOfIterableRoutes := maxRoutesPerBlock - currentRouteCount
-	if numberOfIterableRoutes > maxRoutesPerTx {
-		return maxRoutesPerTx, nil
+	numberOfAvailablePoolPoints := maxPoolPointsPerBlock - currentPoolPointsUsedForBlock
+	if numberOfAvailablePoolPoints > maxPoolPointsPerTx {
+		return maxPoolPointsPerTx, numberOfAvailablePoolPoints, nil
 	}
 
-	return numberOfIterableRoutes, nil
+	return numberOfAvailablePoolPoints, numberOfAvailablePoolPoints, nil
```
p0mvn, line 263: Could you please update godoc indicating what the return values are and when max vs available are returned?

x/protorev/keeper/emit.go
NotJeremyLiu: Resolved in: https://github.com/osmosis-labs/osmosis/pull/4878/commits/f632b421b63d63f15008b49d60a9f961094d8af7

x/protorev/keeper/rebalance.go
NotJeremyLiu: Noted, resolved in https://github.com/osmosis-labs/osmosis/pull/4878/commits/f632b421b63d63f15008b49d60a9f961094d8af7 by removing the unused error return

x/protorev/keeper/rebalance.go
NotJeremyLiu: Updated in: https://github.com/osmosis-labs/osmosis/pull/4878/commits/03f8e989217c3839ed158bed02182f53cfada502

NotJeremyLiu:
@p0mvn All requested changes have been done, re-review ready

p0mvn (approved):

NotJeremyLiu:
Rebased and added change log entry, should be ready to merge

p0mvn (approved):

PR #4885, opened 2023-04-10
[CL]: migrate balancer to concentrated (5/5)

czarcas7ic:
Closes: #XXX

This PR is one of five PRs in an effort to make #4854 more reviewable.

The scope of this PR is limited to adding the balancer to concentrated migration logic.


x/concentrated-liquidity/position.go
```diff
@@ -245,6 +244,12 @@ func (k Keeper) deletePosition(ctx sdk.Context,
 	// Remove the position ID to underlying lock ID mapping (if it exists)
 	key = types.KeyPositionIdForLock(positionId)
 	if store.Has(key) {
+		underlyingLockId, err := k.GetLockIdFromPositionId(ctx, positionId)
+		if err != nil {
+			return err
+		}
+		store.Delete(key)
+		key = types.KeyLockIdForPositionId(underlyingLockId)
 		store.Delete(key)
```
p0mvn, line 253: nit: Reusing the same identifier for various keys is confusing. Suggest changing other instances to be more specific as well

```suggestion
		lockIndexKey := types.KeyLockIdForPositionId(underlyingLockId)
		store.Delete(lockIndexKey)
```

x/concentrated-liquidity/position.go
```diff
@@ -547,11 +552,11 @@ func (k Keeper) validatePositionsAndGetTotalLiquidity(ctx sdk.Context, owner sdk
 	return basePosition.PoolId, basePosition.LowerTick, basePosition.UpperTick, totalLiquidity, nil
 }
 
-// GetPositionIdToLock returns the positionId to lock mapping in state.
-func (k Keeper) GetPositionIdToLock(ctx sdk.Context, positionId uint64) (uint64, error) {
+// GetLockIdFromPositionId returns the positionId to lock mapping in state.
```
p0mvn, line 555:
```suggestion
// GetLockIdFromPositionId returns the lock id associated with the given position id.
```

x/concentrated-liquidity/position.go
```diff
@@ -561,21 +566,46 @@ func (k Keeper) GetPositionIdToLock(ctx sdk.Context, positionId uint64) (uint64,
 	return sdk.BigEndianToUint64(value), nil
 }
 
-// SetPositionIdToLock sets the positionId to lock mapping in state.
+// GetPositionIdFromLockId returns the lockId to position mapping in state.
```
p0mvn, line 569:
```suggestion
// GetPositionIdFromLockId returns the position id associated with the given lock id.
```

x/concentrated-liquidity/position.go
```diff
@@ -561,21 +566,46 @@ func (k Keeper) GetPositionIdToLock(ctx sdk.Context, positionId uint64) (uint64,
...
 	store := ctx.KVStore(k.storeKey)
 
-	// Get the position ID to key mapping.
+	// Get the position ID to key mapping and set.
 	key := types.KeyPositionIdForLock(positionId)
 	store.Set(key, sdk.Uint64ToBigEndian(underlyingLockId))
+
+	// Get the lock ID to key mapping and set.
+	key = types.KeyLockIdForPositionId(underlyingLockId)
+	store.Set(key, sdk.Uint64ToBigEndian(positionId))
```
p0mvn, line 593: Checking that there is a test validating that this index is updated?

x/concentrated-liquidity/position.go
```diff
@@ -561,21 +566,46 @@ func (k Keeper) GetPositionIdToLock(ctx sdk.Context, positionId uint64) (uint64,
...
-	// Get the position ID to key mapping.
+	// Get the position ID to lock mapping.
 	key := types.KeyPositionIdForLock(positionId)
+	underlyingLockId := sdk.BigEndianToUint64(store.Get(key))
+
+	// Delete the mapping from state.
+	store.Delete(key)
+
+	// Get the lock ID to key mapping.
+	key = types.KeyLockIdForPositionId(underlyingLockId)
```
p0mvn, line 608: Please use different identifiers for keys so that it's clearer what exactly is being removed. Also, checking if the new logic is covered by test?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+	}
+	gammSharesInLock := lock.Coins[0]
+	preUnlockLock := *lock
+
+	// Before we break the lock, we must note the time remaining on the lock.
+	remainingLockTime := k.getExistingLockRemainingDuration(ctx, lock)
+
+	// Check if the lock has a corresponding synthetic lock.
+	synthLockBeforeMigration := k.lk.GetAllSyntheticLockupsByLockup(ctx, lockId)
```
p0mvn, line 48:
```suggestion
	// Check if the lock has a corresponding synthetic lock.
	// Synthetic lock existence implies that the lock is superfluid delegated or undelegating.
	synthLockBeforeMigration := k.lk.GetAllSyntheticLockupsByLockup(ctx, lockId)
```

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+
+	// If it does, check if it is superfluid delegated or undelegating.
+	wasSuperfluidUndelegatingBeforeMigration := len(synthLockBeforeMigration) > 0 && strings.Contains(synthLockBeforeMigration[0].SynthDenom, "superunbonding")
+	wasSuperfluidDelegatedBeforeMigration := len(synthLockBeforeMigration) > 0 && strings.Contains(synthLockBeforeMigration[0].SynthDenom, "superbonding")
```
p0mvn, line 52: Can we have a check that you can't be both at the same time?

Also, is there ever a case where synth lock can be neither?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+	// Finish unlocking directly for locked or unlocking locks
+	err = k.lk.ForceUnlock(ctx, *lock)
+	if err != nil {
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+	}
+
+	// If shares to migrate is not specified, we migrate all shares.
+	if sharesToMigrate.IsZero() {
+		sharesToMigrate = gammSharesInLock
+	}
```
p0mvn, line 81: Do you think it might make sense to force users to provide this value and error out if not given?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+		}
+	}
+
+	// If the lock was previously superfluid delegated, superfluid delegate the new concentrated lock to the same validator
+	if wasSuperfluidDelegatedBeforeMigration {
+		err := k.SuperfluidDelegate(ctx, sender.String(), concentratedLockId, intermediateAccount.ValAddr)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
+	}
```
p0mvn, line 119: This can be merged with the "if" branch avove

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+		clIntermediateAccount, err := k.GetOrCreateIntermediaryAccount(ctx, concentratedLockupDenom, valAddr)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
+
+		// Create a new synthetic lockup for the new intermediary account in an unlocking status
+		err = k.createSyntheticLockup(ctx, concentratedLockId, clIntermediateAccount, unlockingStatus)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
```
p0mvn, line 134: Can you elaborate on why we need to create this in the comment please?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,177 @@
...
+		if wasSuperfluidUndelegatingBeforeMigration {
+			gammIntermediateAccount, err := k.GetOrCreateIntermediaryAccount(ctx, remainingGammShares.Denom, valAddr)
+			if err != nil {
+				return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+			}
+			err = k.createSyntheticLockup(ctx, gammLockId, gammIntermediateAccount, unlockingStatus)
+			if err != nil {
+				return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+			}
+		}
```
p0mvn, line 165: We check `if wasSuperfluidDelegatedBeforeMigration` or `if wasSuperfluidUndelegatingBeforeMigration` multiple times throughout the execution flow.

Let's decide on the strategy ones at the top of the function and then complete the full flow specific for that strategy.

This especially makes sense because undelegating and delegated are mutually exclusive so there is no need to have the same flow for the

x/superfluid/keeper/migrate_test.go
```diff
@@ -1,253 +1,317 @@
...
+			sdk.MustNewDecFromStr("0.5"),
+		},
+		{
+			"lock that is superfluid undelegating, unlocking",
+			true,
+			true,
+			true,
+			sdk.MustNewDecFromStr("0.3"),
+		},
+	}
```
p0mvn, line 73: Any error cases worth adding?

x/concentrated-liquidity/position.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/2c6e24f680b0e1f48c2d06a423db51d50bafce17

x/concentrated-liquidity/position.go
czarcas7ic: Added the reverse check (lock to position) here https://github.com/osmosis-labs/osmosis/pull/4885/commits/a2f6c7a3f231c74196226ec7e9cfde10868b55b5

x/concentrated-liquidity/position.go
czarcas7ic: Changed keys here https://github.com/osmosis-labs/osmosis/pull/4885/commits/2c6e24f680b0e1f48c2d06a423db51d50bafce17

Added lock to position check here https://github.com/osmosis-labs/osmosis/pull/4885/commits/439b1799d4f0ea8fadde228f26946b86fe354d30

x/superfluid/keeper/migrate.go
czarcas7ic: These are the only two possibilities, as shown here

https://github.com/osmosis-labs/osmosis/blob/ed085524d91cc472ec20eb3795dcae833c3c5de1/x/superfluid/keeper/synthetic_lock_wrapper.go#L12-L17

I can add a defense in depth measure though to be sure

x/superfluid/keeper/migrate.go
czarcas7ic: The reason I made it this way was simplicity on the frontend, if a user wants to migrate everything then the frontend doesn't have to determine their current total liquidity. But I can change this if you think it should be the way you said in the comment

x/superfluid/keeper/migrate.go
czarcas7ic: expanded on comment here https://github.com/osmosis-labs/osmosis/pull/4885/commits/9a950666cc73e17f3f4b87024e83d130e944b306

x/superfluid/keeper/migrate.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/8df5fd0c9d0542dad2a9b426826f3bbfac62273f

x/superfluid/keeper/migrate.go
czarcas7ic: Agreed with the previous call of merging, but the remaining separation feels logical to me. Separated in three parts:

* Unlocking all previous locks
* Migrate what user desired
* Re-lock anything that user did not want to migrate

x/superfluid/keeper/migrate_test.go
czarcas7ic: Added three error cases here https://github.com/osmosis-labs/osmosis/pull/4885/commits/beb7ab9912a0ab7bb2cc55b5b8c5104d2fff1dd7

* Non existent lock
* Multi-asset lock
* CL lock

x/concentrated-liquidity/position.go
```diff
@@ -222,30 +221,36 @@ func (k Keeper) deletePosition(ctx sdk.Context,
...
-		store.Delete(key)
+	positionIdLockKey := types.KeyPositionIdForLock(positionId)
+	if store.Has(positionIdLockKey) {
+		underlyingLockId, err := k.GetLockIdFromPositionId(ctx, positionId)
+		if err != nil {
+			return err
+		}
+		store.Delete(positionIdLockKey)
+		lockIdPositionKey := types.KeyLockIdForPositionId(underlyingLockId)
+		store.Delete(lockIdPositionKey)
```
p0mvn, line 253: Do we have tests confirming the deletion /modification of both indexes? Applies to other CRUD methods

Also, as a sanity check, is import /export genesis handled correctly for both indexes?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
+	wasSuperfluidDelegatedBeforeMigration := len(synthLockBeforeMigration) > 0 && strings.Contains(synthLockBeforeMigration[0].SynthDenom, "superbonding")
+	if wasSuperfluidUndelegatingBeforeMigration && wasSuperfluidDelegatedBeforeMigration {
+		// This is a defense in depth measure and should never happen.
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, fmt.Errorf("lock %d is both superfluid delegated and undelegating", lockId)
+	}
+
+	// If it is superfluid delegated or undelegating, get the validator address from the synth denom.
+	valAddr := ""
+	if wasSuperfluidDelegatedBeforeMigration || wasSuperfluidUndelegatingBeforeMigration {
+		valAddr = strings.Split(synthLockBeforeMigration[0].SynthDenom, "/")[4]
```
p0mvn, line 62: Let's do a length check and address validation check please

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
+
+		// Create a new synthetic lockup for the new intermediary account in an unlocking status
+		err = k.createSyntheticLockup(ctx, concentratedLockId, clIntermediateAccount, unlockingStatus)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
+	}
```
p0mvn, line 138: Why do we not need to create a superfluid position and delegate it prior to running the undelegating logic in this case?

p0mvn (commented):
Would like to see replies to ` MigrateLockedPositionFromBalancerToConcentrated` and do another pass but the rest overall LGTM

x/concentrated-liquidity/position.go
```diff
@@ -547,44 +552,69 @@ func (k Keeper) validatePositionsAndGetTotalLiquidity(ctx sdk.Context, owner sdk
...
-	key := types.KeyPositionIdForLock(positionId)
+	// Get the position ID to lock mapping.
+	positionIdLockKey := types.KeyPositionIdForLock(positionId)
+	underlyingLockId := sdk.BigEndianToUint64(store.Get(positionIdLockKey))
+
+	// Delete the mapping from state.
+	store.Delete(positionIdLockKey)
+
+	// Get the lock ID to key mapping.
+	lockIdKey := types.KeyLockIdForPositionId(underlyingLockId)
```
mattverse, line 608: nit: consider grouping these two keys and working with them via single method. I noticed that these two keys would in normal cases, work as pairs. That is, if one is set the the other is set, if one is deleted, the other is deleted. Creating helper methods to deal with these two pair keys might help us with ensuring that we use them together

x/concentrated-liquidity/position.go
```diff
@@ -547,44 +552,69 @@ func (k Keeper) validatePositionsAndGetTotalLiquidity(ctx sdk.Context, owner sdk
...
 
-	// Get the position ID to key mapping.
-	key := types.KeyPositionIdForLock(positionId)
-	store.Set(key, sdk.Uint64ToBigEndian(underlyingLockId))
+	// Get the position ID to key mapping and set.
+	positionIdLockKey := types.KeyPositionIdForLock(positionId)
+	store.Set(positionIdLockKey, sdk.Uint64ToBigEndian(underlyingLockId))
+
+	// Get the lock ID to key mapping and set.
+	lockIdKey := types.KeyLockIdForPositionId(underlyingLockId)
```
mattverse, line 592: can we add checks for this setter in the tests as well?

x/superfluid/keeper/stake.go
```diff
@@ -249,47 +242,75 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
...
 	}
 	err = k.forceUndelegateAndBurnOsmoTokens(ctx, amount, intermediaryAcc)
 	if err != nil {
-		return err
+		return types.SuperfluidIntermediaryAccount{}, err
 	}
 
-	// Create a new synthetic lockup representing the unstaking side.
-	return k.createSyntheticLockup(ctx, lockID, intermediaryAcc, unlockingStatus)
+	if createSyntheticLockup {
```
mattverse, line 289: this logic gate seems to be newly added, can we add a test case for this?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
+
+	// Check if the lock has a corresponding synthetic lock.
+	// Synthetic lock existence implies that the lock is superfluid delegated or undelegating.
+	synthLockBeforeMigration := k.lk.GetAllSyntheticLockupsByLockup(ctx, lockId)
+
+	// If it does, check if it is superfluid delegated or undelegating.
+	wasSuperfluidUndelegatingBeforeMigration := len(synthLockBeforeMigration) > 0 && strings.Contains(synthLockBeforeMigration[0].SynthDenom, "superunbonding")
+	wasSuperfluidDelegatedBeforeMigration := len(synthLockBeforeMigration) > 0 && strings.Contains(synthLockBeforeMigration[0].SynthDenom, "superbonding")
+	if wasSuperfluidUndelegatingBeforeMigration && wasSuperfluidDelegatedBeforeMigration {
+		// This is a defense in depth measure and should never happen.
```
mattverse, line 55: Can we also add defense in depth measure for both not being true?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
+		clIntermediateAccount, err := k.GetOrCreateIntermediaryAccount(ctx, concentratedLockupDenom, valAddr)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
+
+		// Create a new synthetic lockup for the new intermediary account in an unlocking status
+		err = k.createSyntheticLockup(ctx, concentratedLockId, clIntermediateAccount, unlockingStatus)
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+		}
```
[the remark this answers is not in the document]
mattverse: I'm assuming `undelegateCommon`

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
-	// 	if preUnlockLock.IsUnlocking() {
-	// 		_, err := k.lk.BeginForceUnlock(ctx, newLock.ID, newLock.Coins)
-	// 		if err != nil {
-	// 			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, err
-	// 		}
-	// 	}
-	// }
-
-	return positionId, amount0, amount1, liquidity, joinTime, poolIdLeaving, poolIdEntering, newLockId, nil
+func (k Keeper) MigrateLockedPositionFromBalancerToConcentrated(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, poolIdLeaving, poolIdEntering, gammLockId, concentratedLockId uint64, err error) {
```
mattverse, line 20: nit: consider refactoring this method into a method that calls sub-methods depending on whether it is superfluid delegated / undelegating. The sub methods would each handle migrating superfluid delegated lock, migrating undelegating superfluid lock. I think this way we can have better code sanity instead of doing if statements throughout the method

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,180 @@
...
+	cltypes "github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/types"
+	gammtypes "github.com/osmosis-labs/osmosis/v15/x/gamm/types"
+	"github.com/osmosis-labs/osmosis/v15/x/superfluid/types"
 )
 
-// UnlockAndMigrate unlocks a balancer pool lock, exits the pool and migrates the LP position to a full range concentrated liquidity position.
-// If the lock is superfluid delegated, it will undelegate the superfluid position.
+// MigrateLockedPositionFromBalancerToConcentrated unlocks a balancer pool lock, exits the pool and migrates the LP position to a full range concentrated liquidity position.
+// If the lock is superfluid delegated, it will undelegate the superfluid position and redelegate it as the concentrated liquidity position.
+// If the lock is superfluid undelegating, it will undelegate the superfluid position and redelegate it as the concentrated liquidity position, but continue to unlock where it left off.
```
mattverse, line 17:
```suggestion
// If the lock is superfluid undelegating, it will instantly undelegate the superfluid position and redelegate it as the concentrated liquidity position, but continue to unlock where it left off.
```

x/superfluid/keeper/stake.go
```diff
@@ -249,47 +242,75 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
...
+
+	return intermediaryAcc, nil
+}
+
+// SuperfluidUndelegate starts undelegating superfluid delegated position for the given lock.
+// Undelegation is done instantly and the equivalent amount is sent to the module account
+// where it is burnt instantly. Note that this method does not include unbonding the lock
+// itself.
+func (k Keeper) SuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint64) error {
+	_, err := k.undelegateCommon(ctx, sender, lockID, true)
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/a61ce3bf13b90714bba6caebde7a9c4a70ef796f

x/superfluid/keeper/stake.go
```diff
@@ -249,47 +242,75 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
...
-func (k Keeper) SuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint64) error {
+// undelegateCommon is a helper function for SuperfluidUndelegate and SuperfluidUndelegateToConcentratedPosition.
+// It performs the following tasks:
+// - checks that the lock is valid for superfluid staking
+// - gets the intermediary account associated with the lock id
+// - deletes the connection between the lock id and the intermediary account
+// - deletes the synthetic lockup associated with the lock id
+// - undelegates the superfluid staking position associated with the lock id and burns the underlying osmo tokens
+// - returns the intermediary account
+// if createSyntheticLockup is true, then it creates a synthetic lockup in an unlocking status for the given lock id.
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/a61ce3bf13b90714bba6caebde7a9c4a70ef796f

x/superfluid/keeper/migrate.go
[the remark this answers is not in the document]
czarcas7ic: I think Roman is correct, I do not think we are deleting the old synth lock if the lock was superfluid unlocking prior to migration. Nice catch!

x/superfluid/keeper/migrate.go
czarcas7ic: Because a undelegating position has no funds delegated to a validator (when we initiate undelegate, we immediately are undelegated and the underlying osmo is burned). The way a undelegating lockup gets slashing is by iterating the unbonding synth locks. See below:
Screenshot 2023-04-19 at 12 21 09 PM

x/superfluid/keeper/migrate.go
czarcas7ic: Agreed, addressed this here https://github.com/osmosis-labs/osmosis/pull/4885/commits/2410b7e1b1fa07e10c31ea646edf6f84ba9b05c0

Made a single entry function that calls one of three separate functions depending on if it is sfDelegated, sfUndelegating, or just locked/unlocking

x/superfluid/keeper/migrate.go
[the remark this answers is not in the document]
czarcas7ic: Ah, no, Matt was right, it was in undelegateCommon

https://github.com/osmosis-labs/osmosis/blob/2410b7e1b1fa07e10c31ea646edf6f84ba9b05c0/x/superfluid/keeper/stake.go#L271-L276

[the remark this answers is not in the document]
czarcas7ic: The following commit should fix this, but I wont resolve this until I add a test that would have caught this / confirms this fixes it https://github.com/osmosis-labs/osmosis/pull/4885/commits/cfc886191c19981f3b6fee9bb0104a13233a9db0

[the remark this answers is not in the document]
czarcas7ic: Ah, so we actually do delete the old synth lock! This is done in ForceUnlock. I will add comments to make this more clear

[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/303dee645cd0e13cf5151dc5e809ee90c698ba3e

x/superfluid/keeper/stake.go
```diff
@@ -249,47 +242,75 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
...
+	}
+
+	return intermediaryAcc, nil
+}
+
+// SuperfluidUndelegate starts undelegating superfluid delegated position for the given lock.
+// Undelegation is done instantly and the equivalent amount is sent to the module account
+// where it is burnt instantly. Note that this method does not include unbonding the lock
+// itself.
+func (k Keeper) SuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint64) error {
```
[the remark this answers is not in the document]
czarcas7ic: A little confused by this comment. We expect a new synthetic lockup to be created for this method.

x/superfluid/keeper/msg_server_test.go
```diff
@@ -418,7 +430,18 @@ func (suite *KeeperTestSuite) TestUnlockAndMigrateSharesToFullRangeConcentratedP
 	suite.App.GAMMKeeper.SetMigrationInfo(suite.Ctx, migrationRecord)
 
 	// Superfluid delegate the balancer pool shares
-	_, _, locks := suite.setupSuperfluidDelegations(valAddrs, []superfluidDelegation{{0, 0, 0, 9000000000000000000}}, denoms)
+	_, _, locks := suite.setupSuperfluidDelegations(valAddrs, []superfluidDelegation{{0, 0, 0, 9000000000000000000}}, []string{poolDenom})
+
+	// Create full range concentrated position (needed to give the pool an initial spot price and liquidity value)
+	suite.CreateFullRangePosition(clPool, fullRangeCoins)
```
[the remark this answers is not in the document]
czarcas7ic: Yes, the problem arises when adding the CL LP share as a superfluid asset. This is because we need a spot price in order to determine the osmo equivalent multiplier. This means that we need to either create an initial position in the upgrade handler if we want to support superfluid / sf migration at upgrade time.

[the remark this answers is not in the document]
czarcas7ic: Added an addition error case here that demonstrates this https://github.com/osmosis-labs/osmosis/pull/4885/commits/839ea2c08c129388e1ced8d68e91ed1cfa7c33b9

x/concentrated-liquidity/position.go
czarcas7ic: Extended the gotest to include checks before and after position deletion for lockId to positionId record https://github.com/osmosis-labs/osmosis/pull/4885/commits/04545e72285f1d99188fd7222667e71c650d2af3

x/superfluid/keeper/migrate_test.go
```diff
@@ -1,253 +1,343 @@
...
+			suite.Require().Equal(clPoolId, poolIdEntering)
+
+			// exitPool has rounding difference.
+			// We test if correct amt has been exited and frozen by comparing with rounding tolerance.
+			defaultErrorTolerance := osmomath.ErrTolerance{
+				AdditiveTolerance: sdk.NewDec(2),
+				RoundingDir:       osmomath.RoundDown,
+			}
+			suite.Require().Equal(0, defaultErrorTolerance.Compare(joinPoolAmt.AmountOf(defaultPoolAssets[0].Token.Denom).ToDec().Mul(tc.percentOfSharesToMigrate).RoundInt(), amount0))
+			suite.Require().Equal(0, defaultErrorTolerance.Compare(joinPoolAmt.AmountOf(defaultPoolAssets[1].Token.Denom).ToDec().Mul(tc.percentOfSharesToMigrate).RoundInt(), amount1))
```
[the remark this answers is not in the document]
czarcas7ic: I don't think this is stemming from any new logic added, and is likely from existing rounding behaviors that need to get addressed in subsequent PRs.

x/superfluid/keeper/migrate.go
czarcas7ic: Both not being true is fine, as this implies we are migrating a non superfluid lock. This is more clear now since I refactored into three distinct methods.

x/superfluid/keeper/stake.go
czarcas7ic: Logic gate was removed as per Roman's suggestion, and `SuperfluidUndelegate` creates synthetic lockup as normal while `SuperfluidUndelegateToConcentratedPosition` does not.

x/concentrated-liquidity/position.go
czarcas7ic: This is tested in the CRUD test here https://github.com/osmosis-labs/osmosis/blob/04545e72285f1d99188fd7222667e71c650d2af3/x/concentrated-liquidity/position_test.go#L1146-L1157

x/concentrated-liquidity/position.go
czarcas7ic: Sorry I am not quite sure I understand. Right now, the helper method of `Remove` removes both lock to position and position to lock, and `Set` sets both lock to position and position to lock. Therefore, utilizing helper methods alone, you cannot remove one pair with out the other, and same for set. Is there a different way I should be grouping them?

mattverse: Apologies if my questions were unclear, I was referring to making a helper method, maybe something called `PositionIdToLockKeys` where it would return an array of key for `KeyPositionIdForLock` and `KeyLockIdForPositionId`. The reason why this might be a useful way is because as of right now, when we write code, we "must" have context or remember manually that these keys are grouped together thus have to be called together. Just a suggestion though, feel free to resolve this comment if you have different thoughts

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+		}
+	}
+
+	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, nil
+}
+
+// migrateSuperfluidUnbondingBalancerToConcentrated migrates a user's superfluid unbonding balancer position to a superfluid unbonding concentrated liquidity position.
+// The function force unlocks and exits the balancer pool, creates a full range concentrated liquidity position, and locks it. If there are any remaining gamm shares, they are re-locked and begin unlocking where they left off.
+// The function returns the concentrated liquidity position ID, amounts of tokens in the position, the liquidity amount, join time, and IDs of the involved pools and locks.
+func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context, sender sdk.AccAddress, poolIdLeaving, poolIdEntering uint64, preMigrationLock *lockuptypes.PeriodLock, gammSharesInLock sdk.Coin, lockId uint64, sharesToMigrate sdk.Coin, synthDenomBeforeMigration string, concentratedPool cltypes.ConcentratedPoolExtension, remainingLockTime time.Duration) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, gammLockId, concentratedLockId uint64, err error) {
```
mattverse, line 102: At brief glance, this method does not seem to support lock that was unbonding && undelegating, but only support lock that was locked but superfluid delegating. 
Can you confirm if this is the case or not?

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
-
-	return positionId, amount0, amount1, liquidity, joinTime, poolIdLeaving, poolIdEntering, newLockId, nil
+func (k Keeper) MigrateLockedPositionFromBalancerToConcentrated(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, poolIdLeaving, poolIdEntering, gammLockId, concentratedLockId uint64, err error) {
+	// Validate and retrieve pertinent data required for migration
+	poolIdLeaving, poolIdEntering, gammSharesInLock, concentratedPool, preMigrationLock, remainingLockTime, synthLockBeforeMigration, isSuperfluidBonded, isSuperfluidUnbonding, err := k.prepareMigration(ctx, sender, lockId, sharesToMigrate)
+	if err != nil {
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+	}
+
+	if isSuperfluidBonded {
```
mattverse, line 27: Would it be possible to have unit tests for these newly added three methods?
Please feel free to make an issue and handle it later instead if you feel this PR is getting to large

mattverse (commented):
Once question for the undelegating & unlocking lock answered, LGTM

x/superfluid/keeper/migrate.go
czarcas7ic: This is not the case. The following test case also tests this:

https://github.com/osmosis-labs/osmosis/blob/839ea2c08c129388e1ced8d68e91ed1cfa7c33b9/x/superfluid/keeper/migrate_test.go#L61-L66

x/superfluid/keeper/migrate.go
czarcas7ic: I added an individual comment to each test case to show which sub method it runs. The top level function is simply a router. Please let me know if you still feel these each need their own test and I will make a separate issue:

https://github.com/osmosis-labs/osmosis/blob/70aee3f3a22424fe43f152605f5c9fe63c0fface/x/superfluid/keeper/migrate_test.go#L41-L72

mattverse: @czarcas7ic I would want to have each individual tdd test cases with separate cases handling all source of failure as they can come in handy if we change one of the lines in one of the methods in the future. if you agree please do make a separate issue, if you do not, please feel free to resolve this

mattverse (approved):

x/concentrated-liquidity/position.go
czarcas7ic: Merged the set and delete methods (kept the getter separated, since there are times you only have one and want the other) https://github.com/osmosis-labs/osmosis/pull/4885/commits/89be60d0bfc3e8c250c1a607e10e83a329a37d4c

x/superfluid/keeper/migrate.go
czarcas7ic: Sounds good, will add this as a low pri issue (can be done while audit is happening)

czarcas7ic: Issue added here https://github.com/osmosis-labs/osmosis/issues/4977

x/superfluid/keeper/stake.go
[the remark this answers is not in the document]
p0mvn: Referenced the wrong method. Here, I think we should have a test verifying the fact of synth lock being created

x/superfluid/keeper/stake.go
```diff
@@ -154,11 +152,6 @@ func (k Keeper) validateLockForSFDelegate(ctx sdk.Context, lock *lockuptypes.Per
 	}
 
 	denom := lock.Coins[0].Denom
-	// If the denom is a concentrated liquidity token, we just strip the position data from the denom
-	if strings.HasPrefix(denom, cltypes.ClTokenPrefix) {
-		index := strings.LastIndex(denom, "/")
-		denom = denom[:index+1]
-	}
```
p0mvn, line 161: Why is this removed now?

x/superfluid/keeper/migrate.go
p0mvn: I don't think this is a low priority and should absolutely be tested pre-launch

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+// The function force unlocks and exits the balancer pool, creates a full range concentrated liquidity position, and locks it. If there are any remaining gamm shares, they are re-locked and begin unlocking where they left off.
+// The function returns the concentrated liquidity position ID, amounts of tokens in the position, the liquidity amount, join time, and IDs of the involved pools and locks.
+func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context, sender sdk.AccAddress, poolIdLeaving, poolIdEntering uint64, preMigrationLock *lockuptypes.PeriodLock, gammSharesInLock sdk.Coin, lockId uint64, sharesToMigrate sdk.Coin, synthDenomBeforeMigration string, concentratedPool cltypes.ConcentratedPoolExtension, remainingLockTime time.Duration) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, gammLockId, concentratedLockId uint64, err error) {
+	// Get the validator address from the synth denom and ensure it is a valid address.
+	valAddr := strings.Split(synthDenomBeforeMigration, "/")[4]
```
p0mvn, line 104: Please add a length check

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+		if err != nil {
+			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, err
+		}
+	}
+
+	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, nil
+}
+
+// migrateSuperfluidUnbondingBalancerToConcentrated migrates a user's superfluid unbonding balancer position to a superfluid unbonding concentrated liquidity position.
+// The function force unlocks and exits the balancer pool, creates a full range concentrated liquidity position, and locks it. If there are any remaining gamm shares, they are re-locked and begin unlocking where they left off.
```
p0mvn, line 100:
```suggestion
// The function force unlocks and exits the balancer pool, creates a full range concentrated liquidity position, and locks it. If there are any remaining gamm shares, they are re-locked and begin unlocking where they left off.
// The remaining duration of the new lock equals to the duration of the pre-existing lock.
```

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+// The function returns the concentrated liquidity position ID, amounts of tokens in the position, the liquidity amount, join time, and IDs of the involved pools and locks.
```
p0mvn, line 101: Please also comment about the creation of the synth lock, why the old one becomes invalid and the new one is needed

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, nil
+}
+
+// prepareMigration prepares for the migration of gamm LP tokens from the Balancer pool to the Concentrated pool. It performs the following steps:
+//
+// 1. Gets the pool ID of the Balancer pool from the gamm share denomination.
+// 2. Ensures a governance-sanctioned link exists between the Balancer pool and the Concentrated pool.
+// 3. Validates that the lock corresponds to the sender, contains the correct denomination of LP shares, and retrieves the gamm shares from the lock.
+// 4. Determines the remaining time on the lock.
+// 5. Checks if the lock has a corresponding synthetic lock, indicating it is superfluid delegated or undelegating.
```
p0mvn, line 223: Please list the return values, there are quite a few. This will make it browsing through code easier in the future

x/superfluid/keeper/migrate_test.go
```diff
@@ -1,253 +1,364 @@
...
+	superfluidtypes "github.com/osmosis-labs/osmosis/v15/x/superfluid/types"
+)
+
+// We test migrating in the following circumstances:
+// 1. Migrating lock that is not superfluid delegated, not unlocking.
+// 2. Migrating lock that is not superfluid delegated, unlocking.
+// 3. Migrating lock that is superfluid delegated, not unlocking.
+// 4. Migrating lock that is superfluid undelegating, not unlocking.
+// 5. Migrating lock that is superfluid undelegating, unlocking.
+func (suite *KeeperTestSuite) TestMigrateLockedPositionFromBalancerToConcentrated() {
```
p0mvn, line 27: I agree with Matt that the current testing strategy is not sufficient. 

It's also difficult to reason about the test case that has so many side effects.

Ideally, the test for the wrapper router should be basic, verifying that it routes to the right underlying method. And each method individually should be verified to exhibit the desired behavior. This will make it much easier to reason about each test case and confirm that the testing strategy used is sufficient.

Would like to bump the priority on this: https://github.com/osmosis-labs/osmosis/issues/4977 if we are punting it till later. Also, would appreciate if you could port the context from this comment to #4977

p0mvn (approved):
LGTM with the bump on #4977

Great work on this!

x/superfluid/keeper/msg_server_test.go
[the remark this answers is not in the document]
p0mvn: Could you document this, please?

x/superfluid/keeper/stake.go
czarcas7ic: CL denoms no longer contain position data

x/superfluid/keeper/migrate.go
czarcas7ic: Low priority since it can be done during audit

x/superfluid/keeper/migrate.go
czarcas7ic: Length check is not needed. It is done in ValAddressFromBech32

x/superfluid/keeper/migrate.go
```diff
@@ -1,99 +1,302 @@
...
+// The function first undelegates the superfluid delegated position, force unlocks and exits the balancer pool, creates a full range concentrated liquidity position, and locks it while superfluid delegating it.
+// If there are any remaining gamm shares, they are re-locked and superfluid delegated as normal. The function returns the concentrated liquidity position ID, amounts of
+// tokens in the position, the liquidity amount, join time, and IDs of the involved pools and locks.
+func (k Keeper) migrateSuperfluidBondedBalancerToConcentrated(ctx sdk.Context, sender sdk.AccAddress, poolIdLeaving uint64, preMigrationLock *lockuptypes.PeriodLock, gammSharesInLock sdk.Coin, lockId uint64, sharesToMigrate sdk.Coin, synthDenomBeforeMigration string, concentratedPool cltypes.ConcentratedPoolExtension, remainingLockTime time.Duration) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, gammLockId, concentratedLockId uint64, err error) {
+	// Get the validator address from the synth denom and ensure it is a valid address.
+	valAddr := strings.Split(synthDenomBeforeMigration, "/")[4]
```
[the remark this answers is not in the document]
czarcas7ic: Length check is not needed. It is done in ValAddressFromBech32

x/superfluid/keeper/migrate.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/2c40ba0afadd23adead9e14fe1fb161d81f4c658

x/superfluid/keeper/migrate.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/f2a52f14868a87ba3bbab6744141d75a65d36b04

x/superfluid/keeper/stake.go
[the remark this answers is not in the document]
czarcas7ic: For this specific comment, we are already checking that here https://github.com/osmosis-labs/osmosis/blob/12446f569a63487186a25bd1801f1cdbdd25cd5c/x/superfluid/keeper/stake_test.go#L368-L395

x/superfluid/keeper/stake.go
```diff
@@ -249,49 +242,71 @@ func (k Keeper) SuperfluidDelegate(ctx sdk.Context, sender string, lockID uint64
...
+	intermediaryAcc, err := k.undelegateCommon(ctx, sender, lockID)
+	if err != nil {
+		return err
+	}
 	// Create a new synthetic lockup representing the unstaking side.
 	return k.createSyntheticLockup(ctx, lockID, intermediaryAcc, unlockingStatus)
 }
 
+// SuperfluidUndelegateToConcentratedPosition starts undelegating superfluid delegated position for the given lock. It behaves similarly to SuperfluidUndelegate,
+// however it does not create a new synthetic lockup representing the unstaking side. This is because at the time this function is called, the new concentrated liquidity side
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/4885/commits/396c42fafa1968c1ca1c45f3f1924775c37d99aa

x/superfluid/keeper/msg_server_test.go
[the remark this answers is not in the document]
czarcas7ic: I created an issue, as I think this is a bug rather than something that should be worked around. https://github.com/osmosis-labs/osmosis/issues/4978

x/superfluid/keeper/migrate_test.go
czarcas7ic: Added this context to the issue itself

PR #4905, opened 2023-04-13
fix: Remove Denom metadata check upon CL Pool creation

mattverse:
Closes: #XXX

Currently we have a check in the CreatePool in the CL msg server, where we check if the given denom0 and denom1's denom meta data exists on chain. 

Unfortunately, we do not have all the denom metadata of ibc tokens registered on chain, thus this check needs to be removed. 

## Brief Changelog
- Remove Denom metadata check upon CL Pool Creation

## Testing and Verifying
Deletes existing tests 


czarcas7ic:
By any chance, do we know which denom meta data do not exist on chain? If not, do we have a ballpark estimate? I was under the impression the only one that didn't exist on chain was for native denoms of osmo and ion (could be wrong though)

mattverse:
@czarcas7icA Actually, none of the IBC denoms are, IIRC only uosmo and uion

p0mvn (approved):
ACK, I think this has only been causing problems for us in testing.

Balancer or stableswap pools do ot have this check either

Issue #4906, opened 2023-04-13
CL: Add CLI for msg `MigrateSharesToFullRangeConcentratedPosition`


ThanhNhann:
## Background

We are missing CLI for msg `MigrateSharesToFullRangeConcentratedPosition`, so we should implement it
https://github.com/osmosis-labs/osmosis/blob/17f4cd528f2acace5ebd015d758f04f0fa4f4adc/proto/osmosis/gamm/pool-models/balancer/tx/tx.proto#L14-L17

## Suggested Design

Implement CLI for msg `MigrateSharesToFullRangeConcentratedPosition`

ThanhNhann:
Yeah, @phamminh0811 is working on it

PR #4907, opened 2023-04-13
add migrate-shares-to-full-range-concentrated-position cli

phamminh0811:
Closes: #4906
## Brief Change Log
This PR add a migrate-shares-to-full-range-concentrated-position cmd on cli


x/gamm/client/cli/tx.go
```diff
@@ -290,6 +291,17 @@ Ex) 2,4,1,5 -> [(Balancer 2, CL 4), (Balancer 1, CL 5)]
 	return cmd
 }
 
+func NewMigrateSharesToFullRangeConcentratedPositionCmd() *cobra.Command {
+	cmd := osmocli.TxCliDesc{
+		Use:              "migrate-shares-to-full-range-concentrated-position [shares-to-migrate]",
+		Short:            "migrate shares to full range concentrated position",
```
ThanhNhann, line 297: hey @phamminh0811, can you refer to this and then update `Short` above: https://github.com/osmosis-labs/osmosis/blob/02afa8fac567502849e745abbb2040078ef287ff/x/gamm/pool-models/balancer/codec.go#L20

pysel (commented):
is it possible to make this command in an osmocli-style? even though it uses osmocli here, I think the final design that was chosen is [this (example)](https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/client/cli/tx.go#L28). Correct me if I am wrong. 

If that is true, I think we should be adding new cli commands that way.

x/gamm/client/cli/tx.go
```diff
@@ -610,6 +622,20 @@ func NewStableSwapAdjustScalingFactorsMsg(clientCtx client.Context, _args []stri
 	return msg, nil
 }
 
+func NewMigrateSharesToFullRangeConcentratedPositionMsg(clientCtx client.Context, args []string, fs *flag.FlagSet) (sdk.Msg, error) {
```
mattverse, line 625: unused argument
```suggestion
func NewMigrateSharesToFullRangeConcentratedPositionMsg(clientCtx client.Context, args []string) (sdk.Msg, error) {
```

x/gamm/client/cli/tx.go
```diff
@@ -290,6 +291,17 @@ Ex) 2,4,1,5 -> [(Balancer 2, CL 4), (Balancer 1, CL 5)]
 	return cmd
 }
 
+func NewMigrateSharesToFullRangeConcentratedPositionCmd() *cobra.Command {
```
mattverse, line 294: I think we can make it return `*osmocli.TxCliDesc` and use `osmocli.AddTxCmd` instead of using BuildCommandCustomFn since we are not using any custom flags here

mattverse (commented):
Thanks for the PR! Left some minor comments, please take a look!

x/gamm/client/cli/tx.go
czarcas7ic: Agreed on this

phamminh0811:
@mattverse  @czarcas7ic  I've updated the code so all comment is fixed. Please check

x/gamm/client/cli/tx.go
```diff
@@ -165,6 +166,16 @@ func NewExitSwapShareAmountIn() (*osmocli.TxCliDesc, *types.MsgExitSwapShareAmou
 	}, &types.MsgExitSwapShareAmountIn{}
 }
 
+func NewMigrateSharesToFullRangeConcentratedPosition() (*osmocli.TxCliDesc, *balancer.MsgMigrateSharesToFullRangeConcentratedPosition) {
+	cmd := &osmocli.TxCliDesc{
+		Use:              "migrate-position [shares-to-migrate]",
+		Short:            "migrate shares to full range concentrated position",
+		NumArgs:          1,
+		ParseAndBuildMsg: NewMigrateSharesToFullRangeConcentratedPositionMsg,
```
czarcas7ic, line 174: We shouldn't need a custom parse function for this. Osmocli should do it all automatically. See tx.go in concentratedliquidity for examples

x/gamm/client/cli/tx.go
```diff
@@ -165,6 +166,19 @@ func NewExitSwapShareAmountIn() (*osmocli.TxCliDesc, *types.MsgExitSwapShareAmou
 	}, &types.MsgExitSwapShareAmountIn{}
 }
 
+func NewMigrateSharesToFullRangeConcentratedPosition() (*osmocli.TxCliDesc, *balancer.MsgMigrateSharesToFullRangeConcentratedPosition) {
+	cmd := &osmocli.TxCliDesc{
+		Use:     "migrate-position",
+		Short:   "migrate shares to full range concentrated position",
+		Example: "migrate-position 1000stake --min-amounts-out=100stake,100uosmo --from=val --chain-id osmosis-1",
```
czarcas7ic, line 173:
```suggestion
		Example: "migrate-position 1000gamm/pool/1 --min-amounts-out=100stake,100uosmo --from=val --chain-id osmosis-1",
```

czarcas7ic (approved):
Please make above changes then we can merge, thanks!

Issue #4875, opened 2023-04-10
[CL]: Export state mappings to GenesisState


czarcas7ic:
## Background

We currently have five state mappings in CL:
* PositionID to position
* Address/Pool/PositionID to position
* PoolID to positionID
* PositionID to LockID
* Full range total liquidity

These all need an entry in genesisState in order to allow us to properly export/import these connections

Added a TODO in code here https://github.com/osmosis-labs/osmosis/pull/4882/commits/8436cd092f70047d37b9133211a06a087320ac28

## Suggested Design

1. Add entries into genesisState for CL
2. Edit genesis_test to ensure these connections are persisted

## Acceptance Criteria

All four connections persist after export / init genesis

PR #4912, opened 2023-04-14
[CL]: Export Position_lock_id mappings to GenesisState

stackman27:
Closes: (#4875)[https://github.com/osmosis-labs/osmosis/issues/4875]

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*


proto/osmosis/concentrated-liquidity/genesis.proto
```diff
@@ -57,7 +57,10 @@ message GenesisState {
 
   repeated Position positions = 3 [ (gogoproto.nullable) = false ];
 
-  uint64 next_position_id = 4
+  uint64 position_lock_id = 4
```
p0mvn, line 60: This doesn't make sense. The lock id must be defined on the proto-generated `PositionData` and that should be replace  `Position` in `GenesisState`.

Each position may have a lock id associated with it as opposed to some single `position_lock_id` in the entire genesis state

p0mvn (changes requested):
Suggested changes. Please take a look and update the PR description

proto/osmosis/concentrated-liquidity/genesis.proto
```diff
@@ -48,14 +48,19 @@ message PoolData {
...
+
 // GenesisState defines the concentrated liquidity module's genesis state.
 message GenesisState {
   // params are all the parameters of the module
   Params params = 1 [ (gogoproto.nullable) = false ];
   // pool data containining serialized pool struct and ticks.
   repeated PoolData pool_data = 2 [ (gogoproto.nullable) = false ];
 
-  repeated Position positions = 3 [ (gogoproto.nullable) = false ];
+  repeated PositionData position_data = 3 [ (gogoproto.nullable) = false ];
```
czarcas7ic, line 63: So, it seems like the only reason we need to do this is because lock_id is the only externally linked data in all of the KVStore pairs. Maybe it makes sense for us to include underlying lock ID in the position struct after all...

Thoughts @p0mvn ?

czarcas7ic (commented):
On glance the logic feels right, would like to hear what Roman has to say on moving lock to position struct before approving though. Nice work!

proto/osmosis/concentrated-liquidity/genesis.proto
p0mvn: Why did we remove it from the second index in the first place? Was it the fact that some position IDs don't have a lock id associated with them? If that's the case, I think the current approach is better.

p0mvn (approved):
Nice work

proto/osmosis/concentrated-liquidity/genesis.proto
czarcas7ic: I removed it from the denom because of the complexity it introduced when attempting to create intermediary accounts with the base denom rather than the full denom (it would try and gather all locks with the base denom rather the full one). Also, removing it reduced much of the custom lock logic that was added previously

stackman27: will proceed with current approach and we can change anything that comes up in the future

czarcas7ic (approved):

Issue #4850, opened 2023-04-06
refactor: charge gas for initialized ticks


p0mvn:
## Background

As spam protection, we need to charge gas for initialized ticks

## Acceptance Criteria
- gas charged
- only when tick is initialized
- tested

p0mvn:
addressed

PR #4917, opened 2023-04-14
refactor: charge gas for initialized ticks 

stackman27:
Closes: #[4850](https://github.com/osmosis-labs/osmosis/issues/4850) & https://github.com/osmosis-labs/osmosis/issues/4947

Add gas cost while initializing ticks

## Brief Changelog
Add gas cost of 10_000 while initializing ticks

## Testing and Verifying
- tested withe minGasConsumed

### DIFF
--- a/tests/e2e/configurer/chain/commands.go
+++ b/tests/e2e/configurer/chain/commands.go
@@ -86,7 +86,7 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
 
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "--gas=600000", "--fees=1500uosmo", "-o json"}
+	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "--gas=500000", "--fees=1250uosmo", "-o json"}
 	outJson, _, err := n.containerManager.ExecTxCmdWithSuccessString(n.t, n.chainId, n.Name, cmd, "code\":0")
 	require.NoError(n.t, err)
 
--- a/tests/e2e/configurer/chain/commands.go
+++ b/tests/e2e/configurer/chain/commands.go
@@ -86,7 +86,7 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
 
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "-o json", "--gas-prices=0.1uosmo", "--gas-adjustment=1.3"}
+	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "-o json", "--"}
 	outJson, _, err := n.containerManager.ExecTxCmdWithSuccessString(n.t, n.chainId, n.Name, cmd, "code\":0")
 	require.NoError(n.t, err)
 
--- a/tests/e2e/configurer/chain/commands.go
+++ b/tests/e2e/configurer/chain/commands.go
@@ -85,7 +85,8 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 // token{0,1} - liquidity to create position with
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
-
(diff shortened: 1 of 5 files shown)


x/concentrated-liquidity/tick_test.go
```diff
@@ -96,6 +96,7 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
 		tickExists             bool
 		expectedLiquidityNet   sdk.Dec
 		expectedLiquidityGross sdk.Dec
+		minimumGasConsumed     uint64
```
stackman27, line 99: set as testing param, just so if we ever want to change it from constant gas cost to linearly scaling gas cost, we can just change this value in test

x/concentrated-liquidity/tick.go
```diff
@@ -59,8 +59,10 @@ func (k Keeper) initOrUpdateTick(ctx sdk.Context, poolId uint64, currentTick int
 		tickInfo.LiquidityNet = tickInfo.LiquidityNet.Add(liquidityIn)
 	}
 
-	k.SetTickInfo(ctx, poolId, tickIndex, tickInfo)
+	// Fixed gas consumption to prevent spam
+	ctx.GasMeter().ConsumeGas(uint64(10000), "initialize tick gas fee")
```
p0mvn, line 63: Let's define `10000` as a `const` at the top of the file please

x/concentrated-liquidity/tick_test.go
```diff
@@ -109,6 +110,7 @@ func (s *KeeperTestSuite) TestInitOrUpdateTick() {
 			tickExists:             false,
 			expectedLiquidityNet:   DefaultLiquidityAmt.Neg(),
 			expectedLiquidityGross: DefaultLiquidityAmt,
+			minimumGasConsumed:     uint64(10000),
```
p0mvn, line 113: Building up on my previous comment, consider importing this from `tick.go`. That way, if we end up changing that value later, we only need to update one constant and zero tests

p0mvn (approved):
LGTM after comments are addressed. Nice work!

czarcas7ic:
It's likely running out of gas from making the position, which is worrisome

stackman27:
> It's likely running out of gas from making the position, which is worrisome

added tracking issue: https://github.com/osmosis-labs/osmosis/issues/4947

p0mvn:
> > It's likely running out of gas from making the position, which is worrisome
> 
> added tracking issue: #4947

Is it possible to address this in this PR, please? Personally, I'm not feeling great about merging something that we don't fully understand, might only cause unnecessary friction with concurrent PRs if this ends up being problematic or hard to debug

stackman27:
> > > It's likely running out of gas from making the position, which is worrisome
> > 
> > 
> > added tracking issue: #4947
> 
> Is it possible to address this in this PR, please? Personally, I'm not feeling great about merging something that we don't fully understand, might only cause unnecessary friction with concurrent PRs if this ends up being problematic or hard to debug

sounds good! resolved. I hardcoded gas and fees

tests/e2e/configurer/chain/commands.go
```diff
@@ -86,7 +86,7 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
 
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "-o json"}
+	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "--gas=600000", "--fees=5000uosmo", "-o json"}
```
p0mvn, line 89: Why are these values chosen specifically?

Is it possible to break them down to make it explicit where `600000` gas and `5000uosmo` are coming from, please?

czarcas7ic: If I'm not mistaken, this should only require the .0025uosmo min fee since we are using less than 1mil gas

stackman27: these are the values closest to the ones i found on local osmosis setup.sh

stackman27: @czarcas7ic so i looked it up and the fees i set to `Fees = txfeestypes.ConsensusMinFee.Mul(sdk.NewDec(GasLimit)).Ceil().TruncateInt64()` and in this case wouldnt' it be 60000 * 0.0025 = 1500uosmo

stackman27: okay so redid this a bit. 
New param; 
gas = 50,000 (e2e was default to 40,000, i added extra 10k because that's the hard coded gas to consume per tick initialization) 
fees = 1250 (because 50,000 * 0.0025 = 1250) 

added in comments too

czarcas7ic: Nice, this logic follows to me

czarcas7ic (approved):
I think we are good to merge this. The calculation that Sishir made in the comment feels correct to me, and at the very least we are not just setting this to auto and forgetting about it

tests/e2e/configurer/chain/commands.go
```diff
@@ -85,8 +85,9 @@ func (n *NodeConfig) CreateConcentratedPool(from, denom1, denom2 string, tickSpa
 // token{0,1} - liquidity to create position with
 func (n *NodeConfig) CreateConcentratedPosition(from, lowerTick, upperTick string, token0, token1 string, token0MinAmt, token1MinAmt int64, poolId uint64) uint64 {
 	n.LogActionF("creating concentrated position")
-
-	cmd := []string{"osmosisd", "tx", "concentratedliquidity", "create-position", lowerTick, upperTick, token0, token1, fmt.Sprintf("%d", token0MinAmt), fmt.Sprintf("%d", token1MinAmt), fmt.Sprintf("--from=%s", from), fmt.Sprintf("--pool-id=%d", poolId), "-o json"}
+	// gas = 50,000 because e2e  default to 40,000, we hardcoded extra 10k gas to initialize tick
+	// fees = 1250 (because 50,000 * 0.0025 = 1250)
```
p0mvn, line 89: Shouldn't we be initializing two ticks per position?

stackman27: i confirmed in `UpdatePosition` unit test to make sure we are using 20k gas while initializing test here's the output 
```
with consume gas 
=== RUN   TestKeeperTestSuite/TestUpdatePosition/update_existing_position_with_positive_amount
BEFORE  INITIALIZE TICK (Lower & Upper) :  345134
AFTER  INITIALIZE TICK (Lower & Upper) GAS:  396742

without consume gas 
=== RUN   TestKeeperTestSuite/TestUpdatePosition/update_existing_position_with_positive_amount
BEFORE  INITIALIZE TICK (Lower & Upper) GAS:  345134
AFTER INITIALIZE TICK (Lower & Upper) GAS:  376742
```

maybe in e2e the fees were too low to begin with?

p0mvn: Discussed offline to not block on this but keep this open question in mind. Thanks for looking into it @stackman27

p0mvn (approved):

Issue #4852, opened 2023-04-07
[CL] Get migration info by pool ID


AlpinYukseloglu:
## Background

In our migration logic (which is now also used for reward splitting between CL and Balancer pools), we currently get all links between Balancer and CL pools every time we run it. This leads to a significant amount of overhead that can be avoided by simply keying links by pool ID.

To make bi-directional link searching to work (finding balancer pool ID from CL pool ID and vice versa), it might even make sense to make two state entries for each link (unless there's a more space-efficient way to do this cc @p0mvn).

## Suggested Design

- Key migration links by pool IDs and change the following code to access the ID directly (instead of getting all and iterating):
https://github.com/osmosis-labs/osmosis/blob/93e992cb8465dc9f5bf7faa34c7ab99dea465a95/x/gamm/keeper/migrate.go#L221

## Acceptance Criteria

- All new and existing tests pass

PR #4940, opened 2023-04-17
refactor: bi-directional store for migration info

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/4852

Existing migration info store was not optimized for get set computation, as for single migration action, we would iterate over all migration link, search for the given pool id.

This problem is fixed in this PR via using keys to store the linked pool id bi-directionally directly. 

## Brief Changelog

Adds bi-directional store for migration info.

## Testing and Verifying

Passes existing tests, adds more tests


proto/osmosis/gamm/v1beta1/genesis.proto
```diff
@@ -42,7 +42,12 @@ message MigrationRecords {
...
-  uint64 cl_pool_id = 2;
+  PoolID balancerPool = 1 [ (gogoproto.nullable) = false ];
+  PoolID clPool = 2 [ (gogoproto.nullable) = false ];
 }
+
+// used for data serialization
+message PoolID {
+  option (gogoproto.equal) = true;
+  uint64 poolId = 1;
+}
```
p0mvn, line 53: Can we instead use `sdk.Uint64ToBigEndian` instead of generating this?

I think it will make code cleaner due to not having to initialize `PoolId` struct and being able to use `uint64` directly

p0mvn (approved):
LGTM once nit is addressed

proto/osmosis/gamm/v1beta1/genesis.proto
mattverse: Forgot for a moment that osmoutils does not support non proto sets, changed as requested

x/gamm/keeper/migrate.go
```diff
@@ -52,19 +52,89 @@ func (k Keeper) MigrateFromBalancerToConcentrated(ctx sdk.Context, sender sdk.Ac
...
+// GetLinkedConcentratedPoolID returns the Balancer pool Id linked for the given concentrated pool Id.
+// Returns error if link for the given pool id does not exist.
+func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint64) (uint64, error) {
+	store := ctx.KVStore(k.storeKey)
+	concentratedToBalancerPoolKey := types.GetKeyPrefixMigrationInfoPoolCLPool(concentratedPoolId)
+
+	balancerPoolIdBigEndian := store.Get(concentratedToBalancerPoolKey)
+	if balancerPoolIdBigEndian == nil {
+		return 0, types.ConcentratedPoolMigrationLinkNotFoundError{PoolIdLeaving: concentratedPoolId}
+	}
```
czarcas7ic, line 104: Wrong error, since this would be PoolIdEntering, not leaving

x/gamm/keeper/migrate.go
```diff
@@ -52,19 +52,89 @@ func (k Keeper) MigrateFromBalancerToConcentrated(ctx sdk.Context, sender sdk.Ac
...
+		prefixStore.Delete(iter.Key())
+	}
+
+	for _, balancerToCLPoolLink := range migrationInfo.BalancerToConcentratedPoolLinks {
+		balancerToClPoolKey := types.GetKeyPrefixMigrationInfoBalancerPool(balancerToCLPoolLink.BalancerPoolId)
+		store.Set(balancerToClPoolKey, sdk.Uint64ToBigEndian(balancerToCLPoolLink.ClPoolId))
+
+		clToBalancerPoolKey := types.GetKeyPrefixMigrationInfoPoolCLPool(balancerToCLPoolLink.ClPoolId)
+		store.Set(clToBalancerPoolKey, sdk.Uint64ToBigEndian(balancerToCLPoolLink.BalancerPoolId))
+	}
```
czarcas7ic, line 137: Do set functions usually delete all existing keys? Feel like these two should potentially be separate functions (if you disagree, please feel free to resolve. Not blocking)

x/gamm/keeper/migrate.go
```diff
@@ -213,29 +290,3 @@ func (k Keeper) UpdateMigrationRecords(ctx sdk.Context, records []types.Balancer
...
-// If a link exists, it returns the balancer pool ID.
-// If a link does not exist, it returns a 0 pool ID an error.
-func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, poolIdEntering uint64) (poolIdLeaving uint64, err error) {
-	migrationInfo := k.GetMigrationInfo(ctx)
-	for _, info := range migrationInfo.BalancerToConcentratedPoolLinks {
-		if info.ClPoolId == poolIdEntering {
-			return info.BalancerPoolId, nil
-		}
-	}
-	return 0, types.BalancerPoolMigrationLinkNotFoundError{PoolIdEntering: poolIdEntering}
```
czarcas7ic, line 240: This is the error you want that I mentioned before

czarcas7ic (approved):
Nice work :)

x/gamm/keeper/migrate.go
mattverse: Set function's purpose is to set the given object. Usually it overrides it automatically since set functions have a 1:1 relationship between the key and the object being stored.

This case is a special case, as we are having a 1:2 relationship. That is, within this set function, we are doing two different sets, causing the object being stored to not "automatically" over ride the object that was there before hand. That is why in this case, we manually iterate over the keys and delete all existing keys. 

I do think that what you are saying does make sense in terms of causing confusion, changing this to `OverrideAllMigrationInfo`

p0mvn (approved):

Issue #4939, opened 2023-04-17
fix: bug with liquidity net in direction query


p0mvn:
## Background

![image](https://user-images.githubusercontent.com/34196718/232377533-232090f3-ebd8-4e59-9b4a-673aeac3bac1.png)

https://osmosis-network.slack.com/files/U02UCH04W8J/F0531V32JJK/jon_osmosisd-local3.zip

cc: @jonator 

## Acceptance Criteria
- investigate and fix the bug

jonator:
* It seems the same [query](http://localhost:1317/osmosis/concentratedliquidity/v1beta1/query_liquidity_net_in_direction?pool_id=1&token_in=uosmo&use_cur_tick=true&use_no_bound=true) in the other direction with many more ticks has no issues. Perhaps it's the case with a single tick in the requested direction?

Issue #4923, opened 2023-04-15
(CL) Query `query_liquidity_net_in_direction` improvements


jonator:
#### System information

OS & Version: 
Commit hash: e35f5e6be71b812d0385187a71663476eaea2935

#### Expected behaviour

1. (minor) The name of the query: IMO the word `query` doesn't add any useful info. Query should be removed, and it should be something simpler: `liquidity_net_in_direction`

PR #4943, opened 2023-04-17
fix: liquidity net in direction query utilizing current sqrt price instead of current tick sqrt price

p0mvn:
Closes: #4939, #4923

See linked issue for context.

It seems that the query was failing because it was validating start tick against the current sqrt price. However, we should be validating against the "current tick sqrt price" instead.

Current sqrt price might be slightly to the left / right (depending on the query direction) of the current tick's sqrt price. However, we always set the start tick to the current tick when no start tick is provided.

## Brief Changelog

- fixed validation to be against the current tick sqrt price
- improved logs
- improved error message
- changed query per-request in https://github.com/osmosis-labs/osmosis/issues/4923
This is the query now:
```
curl "localhost:1317/osmosis/concentratedliquidity/v1beta1/liquidity_net_in_direction?pool_id=1&token_in=uosmo&use_cur_tick=true&use_no_bound=true"
```

## Testing and Verifying

- manually tested on localosmosis with @jonator 's state


x/concentrated-liquidity/tick.go
```diff
@@ -306,13 +306,13 @@ func (k Keeper) GetTickLiquidityNetInDirection(ctx sdk.Context, poolId uint64, t
 		return []query.TickLiquidityNet{}, err
 	}
 
-	ctx.Logger().Debug(fmt.Sprintf("userGivenStartTick %s, boundTick %s, currentTick %s\n", userGivenStartTick, boundTick, p.GetCurrentTick()))
+	ctx.Logger().Info(fmt.Sprintf("userGivenStartTick %s, boundTick %s, currentTick %s\n", userGivenStartTick, boundTick, p.GetCurrentTick()))
```
czarcas7ic, line 309: I personally don't think these should be info logs, as info logs are already pretty swamped. As a node operator, I would want info logs to give me pertinent information to the node's operation. If we added logs for every query, it would be pretty unusable for the avg node operator. This is just an opinion though and can be disregarded if you feel strongly :)

x/concentrated-liquidity/tick.go
```diff
@@ -343,30 +343,38 @@ func (k Keeper) GetTickLiquidityNetInDirection(ctx sdk.Context, poolId uint64, t
...
 	if err := validateTickIsInValidRange(boundTick); err != nil {
-		return []query.TickLiquidityNet{}, err
+		return []query.TickLiquidityNet{}, fmt.Errorf("failed validating bound tick (%s) with current sqrt price of (%s): %w", boundTick, currentTickSqrtPrice, err)
 	}
 
-	ctx.Logger().Debug("validating start tick")
+	ctx.Logger().Info("validating start tick")
 	if err := validateTickIsInValidRange(startTick); err != nil {
-		return []query.TickLiquidityNet{}, err
+		return []query.TickLiquidityNet{}, fmt.Errorf("failed validating start tick (%s) with current sqrt price of (%s): %w", startTick, currentTickSqrtPrice, err)
```
czarcas7ic, line 377: Can these errors be added to errors.go?

czarcas7ic (approved):
LGTM :miltank:

x/concentrated-liquidity/tick.go
czarcas7ic: (Info logs are the default setting)

jonator (approved):

x/concentrated-liquidity/tick.go
p0mvn: This is wrapping an existing error to provide more context. I don't think adding a specific type for the wrapper would be valuable here

x/concentrated-liquidity/tick.go
p0mvn: I left these in case I need to debug more if the fix doesn't work for Jon. This is to be reverted before the merge

Issue #4936, opened 2023-04-17
CL: implement pool creation enabled param


p0mvn:
## Background

This is to avoid having major pools created and unintentionally fragmenting the liquidity.

We want to be able to enable it via gov prop later

## Acceptance Criteria
- Flag added to params
- implement gov prop to flip the flag
- Disabled at launch
- e2e refactored so that all e2e tests run against the canonical pool from migrations for now

PR #4952, opened 2023-04-18
feat: CL permissionless pool creation param

p0mvn:
Closes: #4936

Adding a parameter to disable concentrated liquidity pool creation

## Brief Changelog

- implement param
- implement a check in pool manager
- test
- refactor e2e to validate the parameter
- drive by change: add canonical pool test

## Testing and Verifying

- tests in pool manager
- e2e test


tests/e2e/e2e_test.go
```diff
@@ -1460,3 +1480,38 @@ func (s *IntegrationTestSuite) TestGeometricTWAP() {
 	// quote assset supply / base asset supply = 1_000_000 / 2_000_000 = 0.5
 	osmoassert.DecApproxEq(s.T(), sdk.NewDecWithPrec(5, 1), afterSwapTwapBOverA, sdk.NewDecWithPrec(1, 2))
 }
+
+// Tests that v16 upgrade correctly creates the canonical OSMO-DAI pool in the upgrade.
+// Prefixed wit "A" to run before TestConcentratedLiquidity that resets the pool creation
```
czarcas7ic, line 1485:
```suggestion
// Prefixed with "A" to run before TestConcentratedLiquidity that resets the pool creation
```

x/concentrated-liquidity/README.md
```diff
@@ -1171,6 +1171,14 @@ to only a few denoms.
 Our list at launch is expected to consist of OSMO, DAI and USDC. These are set in the v16
 upgrade handler.
 
+- `IsPermisionlessPoolCreationEnabled` bool
+
+The flag indicating whether permissionless pool creation is enabled or not. For launch, we have
+decided to disable permissionless pool creation. It will still be enabled via governance.
+This is because we want to limit the number of pools for risk management and want to avoid
+fragmenting liquidity for major denom pairs with configurations of exponent at price one
+and tick spacing that are not ideal.
```
czarcas7ic, line 1180: The latter point will more than likely be moot since we are selecting a canonical exponent and tick spacing now

x/concentrated-liquidity/types/params.go
```diff
@@ -46,6 +48,7 @@ func DefaultParams() Params {
 			"ibc/0CD3A0285E1341859B5E86B6AB7682F023D03E97607CCC1DC95706411D866DF7", // DAI
 			"ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858", // USDC
 		},
+		IsPermissionlessPoolCreationEnabled: false,
```
czarcas7ic, line 51: Can we please make this true for the default? I can imagine with making custom testnets / state exports, this is just going to be one more thing we have to always change, whereas we know we are setting this to true in the upgrade handler which would just be a one time thing.

x/gamm/keeper/keeper.go
```diff
@@ -88,3 +88,9 @@ func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
 func (k Keeper) setParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
+
+// IsPermissionlessPoolCreationEnabled returns nil if permissionless pool creation in the module is enabled.
+// Error otherwise
+func (k Keeper) IsPermissionlessPoolCreationEnabled(ctx sdk.Context) error {
+	return nil
+}
```
czarcas7ic, line 96: This is kind of odd imo. Any time I see an "Is" method I always expect a bool rather than an error.

czarcas7ic (approved):
Nice work 👍

x/gamm/keeper/keeper.go
p0mvn: Renamed to `Validate`

x/concentrated-liquidity/types/params.go
p0mvn: I think this is actually preferable to keep this consistent with mainnet so that for out v16 testnets we are testing pool creation as disabled. This also helps to run e2e by default with CL pool creation disabled which is intended.

Once we change the param on mainnet, we can flip this flag here as well.

Let me know if that makes sense

x/concentrated-liquidity/README.md
p0mvn: Tick spacing is still configurable, isn't it?

czarcas7ic: I thought we wanted to do 100 all around? Even if that isn't the case, we should change it to say "want to avoid
fragmenting liquidity for major denom pairs with configurations of ~~exponent at price one
and~~ tick spacing that are not ideal.

x/concentrated-liquidity/types/params.go
czarcas7ic: Spoke offline on this. While I personally think the overhead of changing this for localosmosis and testnetify is greater than any risk this would expose, will not block on my opinion!

p0mvn: I will think more about this tomorrow. Will likely go with your suggestion

x/concentrated-liquidity/keeper.go
```diff
@@ -99,3 +99,12 @@ func (k *Keeper) SetListeners(listeners types.ConcentratedLiquidityListeners) *K
 
 	return k
 }
+
+// ValidatePermissionlessPoolCreationEnabled returns nil if permissionless pool creation in the module is enabled.
+// Otherwise, returns an error.
+func (k Keeper) ValidatePermissionlessPoolCreationEnabled(ctx sdk.Context) error {
```
mattverse, line 105: nit:consider changing to
```suggestion
func (k Keeper) IsPermissionlessPoolCreationEnabled(ctx sdk.Context) error {
```
For code style unification

x/concentrated-liquidity/types/params.go
mattverse: I actually agree with @p0mvn on keeping this flag as what is going to be in mainnet. For localosmosis, if necessary, we can  add a line to edit the genesis manually to have the param to true, for testnet, we can probably change it using gov proposals?

x/gamm/keeper/keeper.go
```diff
@@ -88,3 +88,9 @@ func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
 func (k Keeper) setParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
+
+// ValidatePermissionlessPoolCreationEnabled returns nil if permissionless pool creation in the module is enabled.
+// Error otherwise
```
mattverse, line 93:
```suggestion
// Pools in gamm module have permissionless pool creation enabled, thus always return nil.
```

mattverse (approved):
Nice work lgtm, do we have plans to integrate gov prop for changing this? If so, can we create an issue to track this?

p0mvn:
> Nice work lgtm, do we have plans to integrate gov prop for changing this? If so, can we create an issue to track this?

This is a param change proposal that does not need anything custom be implemented

x/concentrated-liquidity/keeper.go
p0mvn: ref: https://github.com/osmosis-labs/osmosis/pull/4952#discussion_r1170726335

x/concentrated-liquidity/README.md
p0mvn: done

x/concentrated-liquidity/types/params.go
p0mvn: Updated localosmosis config: [8205f7d](https://github.com/osmosis-labs/osmosis/pull/4952/commits/8205f7da85be0bd8f951ddba493662782c62c37f)

going to go with the flag as on mainnet. We can change this later if ends up being problematic

Issue #4520, opened 2023-03-06
[CL]: Auto Generate Queries


czarcas7ic:
## Background

See comment https://github.com/osmosis-labs/osmosis/pull/4501#discussion_r1125786593

We currently manually created CL queries when we should have utilized `make run-querygen`

## Suggested Design

- Implement structure needed to auto generate queries
- Auto generate the queries and replace what was manually created

## Acceptance Criteria

- Query tests still pass

PR #4956, opened 2023-04-18
refactor: use querygen in CL

p0mvn:
Closes: #4520

Minor refactor to integrate CL into the `make run-querygen` framework

## Brief Changelog

- regenerate protos
- `query.yml`
- perform necessary renames and exports

## Testing and Verifying


x/concentrated-liquidity/client/query_proto_wrap.go
```diff
@@ -0,0 +1,167 @@
...
+func (q Querier) Params(ctx sdk.Context, req clquery.ParamsRequest) (*clquery.ParamsResponse, error) {
+	return &clquery.ParamsResponse{Params: q.Keeper.GetParams(ctx)}, nil
+}
+
+// TotalLiquidityForRange returns an array of LiquidityDepthWithRange, which contains the range(lower tick and upper tick) and the liquidity amount in the range.
+func (q Querier) TotalLiquidityForRange(ctx sdk.Context, req clquery.TotalLiquidityForRangeRequest) (*clquery.TotalLiquidityForRangeResponse, error) {
+	liquidity, err := q.Keeper.GetTickLiquidityForFullRange(
+		ctx,
+		req.PoolId,
+	)
```
czarcas7ic, line 115: The description implies the user specifies a lower and upper tick to get the liquidity range for, but in the function itself we are calling tick liquidity for the full range only (I know this was just copied over, but still unsure of which logic this should be utilizing)

czarcas7ic (approved):
Nice work, approving but would like an ack or change on the one comment I made.

x/concentrated-liquidity/client/query_proto_wrap.go
czarcas7ic: I see the proto only expects a pool ID, but I'm confused on if the future intention is to allow users to provide a tick range?

p0mvn: Agree this doesn't make much sense. However, Jacky is integrating right now so it's best to check in with him first separately to avoid breaking him.

Made an issue to track: https://github.com/osmosis-labs/osmosis/issues/4958

p0mvn: Renamed: [7fada3b](https://github.com/osmosis-labs/osmosis/pull/4956/commits/7fada3b1adfeb58ee2c014b9859069e22db93fa0)

PR #4946, opened 2023-04-17
Refactor: integrate protorev into concentrated liquidity

stackman27:
[Closes: #4873](https://github.com/osmosis-labs/osmosis/issues/4873)

## What is the purpose of the change
Currently, protorev only supports gamm pools. We should add support for CL pools by routing all the logic via the x/poolmanager module

## Brief Changelog
n/a

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #4966, opened 2023-04-19
Jl/generalize protorev epoch hook for all pool modules

NotJeremyLiu:

To show what I view as the ideal generalized protorev interface for the epoch_hook to arbitrary pool modules that use the poolmanager interfaces.

## Brief Changelog

- Add a function in poolmanager that gets all the pool modules that can be used by other modules
- Add new function to protorev expected keepers
- Use this function to get all pools for individual pool modules within protorev epoch hook

### DIFF
--- a/x/poolmanager/router.go
+++ b/x/poolmanager/router.go
@@ -550,6 +550,11 @@ func (k Keeper) AllPools(
 	return sortedPools, nil
 }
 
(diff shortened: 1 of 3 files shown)


stackman27 (approved):
This LGTM will handle testing in #4946

Issue #4971, opened 2023-04-21
bug: collect fees - check that only position owner can collect fees


p0mvn:
## Background

The check currently does not exist

## Suggested Design

Add check

## Acceptance Criteria
- add check
- add test only owner can claim

p0mvn:
Confirm claim incentives has the check and test for this as well

PR #4972, opened 2023-04-21
refactor: position owner check at collection time and reduce boilerplate in fees

p0mvn:
Closes: #4971

I should not be able to claim fees or incentives if I'm not a position owner. This change adds a relevant changes and tests for it.

Additionally, it reduces boilerplate by introducing an internal `getClaimableFees` function.

## Brief Changelog

- reduce boilerplate
- add owner check
- add tests
- clean up tests

## Testing and Verifying

This change added tests


x/concentrated-liquidity/fees.go
```diff
@@ -196,39 +196,22 @@ func (k Keeper) getInitialFeeGrowthOutsideForTick(ctx sdk.Context, poolId uint64
...
 	}
 
-	// Compute the fee growth outside of the range between the position's lower and upper ticks.
-	feeGrowthOutside, err := k.getFeeGrowthOutside(ctx, position.PoolId, position.LowerTick, position.UpperTick)
-	if err != nil {
-		return sdk.Coins{}, err
-	}
-
-	// Prepare the position's accumulator for claiming rewards and claim the rewards.
-	feesClaimed, _, err := prepareAccumAndClaimRewards(feeAccumulator, positionKey, feeGrowthOutside)
```
czarcas7ic, line 231: I am personally highly against removing comments like these. It allows me to quickly skim the intent of each line, then re-read and check if the function is doing what each line intended.

Will not block on this as its clearly a preference thing, but I will forever write code in this way.

x/concentrated-liquidity/fees.go
```diff
@@ -264,15 +247,18 @@ func (k Keeper) collectFees(ctx sdk.Context, owner sdk.AccAddress, positionId ui
 func (k Keeper) GetClaimableFees(ctx sdk.Context, positionId uint64) (sdk.Coins, error) {
 	// Since this is a query, we don't want to modify the state and therefore use a cache context.
 	cacheCtx, _ := ctx.CacheContext()
+	return k.getClaimableFees(cacheCtx, positionId)
+}
 
+func (k Keeper) getClaimableFees(ctx sdk.Context, positionId uint64) (sdk.Coins, error) {
 	// Get the position with the given ID.
-	position, err := k.GetPosition(cacheCtx, positionId)
+	position, err := k.GetPosition(ctx, positionId)
```
czarcas7ic, line 255: I do not like the way this was separated. Any get function should not have the ability to change state, even if its private. In the future, if someone sees this "getClaimableFees" private methods and attempts to use it in a new function, it would actually claim the rewards. We should make it so that there is no way for the getter to modify state, even if misused.

x/concentrated-liquidity/position.go
```diff
@@ -89,6 +89,19 @@ func (k Keeper) hasAnyPositionForPool(ctx sdk.Context, poolId uint64) (bool, err
 	return osmoutils.HasAnyAtPrefix(store, key, parse)
 }
 
+// isPositionOwner returns true if the given positionId is owned by the given sender insider the given pool.
```
czarcas7ic, line 92:
```suggestion
// isPositionOwner returns true if the given positionId is owned by the given sender inside the given pool.
```

czarcas7ic (commented):
Good catch on this! Would like a discussion on the getter function not using cache ctx before approving.

x/concentrated-liquidity/fees.go
p0mvn: Modifying state is intentional since it is used by collectFees as well. I can change the name to reflect that this is not a getter

x/concentrated-liquidity/fees.go
p0mvn: Note that I did not remove any comments that still apply. I just did not add any comments for the new logic since it's trivial

I reduced code duplication by introducing an unexported getClaimableFees. So the old comments do not apply here anymore.

Prior to this change, GetClaimableFees and CollectFees were essentially the same with the exception of cache context

x/concentrated-liquidity/fees.go
czarcas7ic: If the name changes then I am good with that!

czarcas7ic (approved):
Approving since agreement on changing the getter name to something else

x/concentrated-liquidity/msg_server_test.go
```diff
@@ -231,6 +243,14 @@ func (suite *KeeperTestSuite) TestCollectIncentives_Events() {
 			expectedCollectIncentivesEvent:      3,
 			expectedMessageEvents:               4, // 1 for collect incentives, 3 for send messages
 		},
+		"error: three position IDs - not an owner": {
+			upperTick:                  DefaultUpperTick,
+			lowerTick:                  DefaultLowerTick,
+			positionIds:                []uint64{DefaultPositionId, DefaultPositionId + 1, DefaultPositionId + 2},
+			numPositionsToCreate:       2,
+			shouldSetupUnownedPosition: true,
+			expectedError:              cltypes.NotPositionOwnerError{},
```
AlpinYukseloglu, line 252: Shouldn't this error message contain the specific position ID and sender (using whichever of the two positions it errors on) as per the error type?

x/concentrated-liquidity/msg_server_test.go
```diff
@@ -138,51 +139,61 @@ func (suite *KeeperTestSuite) TestCollectFees_Events() {
...
+			expectedError:                 cltypes.NotPositionOwnerError{Address: suite.TestAccs[0].String(), PositionId: DefaultPositionId + 2},
 		},
 	}
 
 	for name, tc := range testcases {
 		suite.Run(name, func() {
 			suite.SetupTest()
-			ctx := suite.Ctx
+
+			msgServer := cl.NewMsgServerImpl(suite.App.ConcentratedLiquidityKeeper)
```
AlpinYukseloglu, line 164: Out of curiosity, what was the reasoning behind moving this line above the pool/position setup?

x/concentrated-liquidity/fees.go
AlpinYukseloglu: +1 on this change

AlpinYukseloglu (approved):
LGTM once the function name change is made

x/concentrated-liquidity/msg_server_test.go
p0mvn: I opted in for `ErrorAs` check which only checks the error type:
https://github.com/osmosis-labs/osmosis/blob/6270bc674f24f9352a5693efe3518b7d6f4542bc/x/concentrated-liquidity/msg_server_test.go#L196

The reason is that we run `SetupTest()` for every test case. As a result, `s.TestAccs[0]` gets reset on every iteration.

While checking that there are specific arguments formatted to the error is stricter than just the error type, I went with the faster approach to set it up since the marginal benefit of validating the formatting is insignificant IMO.

x/concentrated-liquidity/fees.go
p0mvn: Changed to `prepareClaimableFees` and added godoc

x/concentrated-liquidity/msg_server_test.go
p0mvn: No reason. This can be moved back if desired.

I think I might have suspected that ordering was causing some of the initial issues I was debugging and didn't eventually revert it.

p0mvn:
Merging since all comments are addressed

Issue #4357, opened 2023-02-17
test(CL): functional tests around LP, swaps and collect fees


p0mvn:
We can add this in a separate PR but I think that we also need a functional test where we update or delete a position. I can write something more detailed in the next few days

_Originally posted by @p0mvn in https://github.com/osmosis-labs/osmosis/pull/4282#discussion_r1107533256_

p0mvn:
TBD on context

PR #4973, opened 2023-04-21
 fix: fee bug around LP logic

p0mvn:
Closes: #4357

Currently, if the user tries to create a position at the tick range that previously had fees distributed, the creation will fail.

This is because the system would attempt to claim rewards on a newly initialized position. This stems from the fact that we always run the following for a new position in sequence:
- `initializeFeeAccumulatorPosition`
- `updateFeeAccumulatorPosition`

where `updateFeeAccumulatorPosition` claims rewards.

None of our tests covered completely draining a position and creating the same one on the tick range that had reward claimed previously.

As a result, this behavior went uncaught. Upon @AlpinYukseloglu in his work adding an additional `CreatePosition`  in the ` TestWithdrawPosition`, this bug was exposed.

The solution in this PR focuses on combining "initialization" and "update" logc in one function, checking whether this is a new position and running only one or the other flow.

The tests are updated to check that accumulator matches the desired values. A functional test is added to test various invariants around LP and fee claiming.

## TODO

- [x] functional test
- [x] clean up

## Brief Changelog

- add check to replicate bug
- consolidate "initialization" and "update" fee accumulator logic in one function `initOrUpdateFeeAccumulatorPosition`
- call it where "update" used to be called in `UpdatePosition`
- refactor and update tests.

## Testing and Verifying

This change added tests


.vscode/launch.json
```diff
@@ -6,25 +6,42 @@
...
+                "TestKeeperTestSuite/TestInverseRelation_CreatePosition_WithdrawPosition/lower_tick",
+                "-test.v"
+            ],
+        },
+        {
+            "name": "TestUpdatePosition",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${workspaceFolder}/x/concentrated-liquidity",
```
czarcas7ic, line 37: I assume these are left over from running the debugger?

x/concentrated-liquidity/fees_test.go
```diff
@@ -55,59 +56,83 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
-		expectedPass bool
+		expectedLiquidity sdk.Dec
+		expectedPass      bool
 	}
 	tests := []initFeeAccumTest{
 		{
-			name:           "first position",
-			positionFields: defaultPositionFields,
-			expectedPass:   true,
+			name:           "error: negative liqudity for the first position",
```
czarcas7ic, line 100:
```suggestion
			name:           "error: negative liquidity for the first position",
```

x/concentrated-liquidity/fees.go
```diff
@@ -65,72 +73,49 @@ func (k Keeper) initializeFeeAccumulatorPosition(ctx sdk.Context, poolId uint64,
...
+		// Initialize the position with the fee growth inside the tick range
+		if err := feeAccumulator.NewPositionCustomAcc(positionKey, liquidityDelta, feeGrowthInside, nil); err != nil {
+			return err
+		}
+	} else {
+		// Replace the position's accumulator in the fee accumulator with a new one
+		// that has the latest fee growth outside of the tick range.
+		// Assume the last time the position was created or modified was at time t.
+		// At time t, we track fee growth inside from 0 to t.
+		// Then, the update happens at time t + 1. The call below makes the positon's
```
czarcas7ic, line 102:
```suggestion
		// Then, the update happens at time t + 1. The call below makes the position's
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -847,39 +893,59 @@ func (s *KeeperTestSuite) TestQueryAndCollectFees() {
...
+			doesFeeGrowBeforeFirstCall: true,
+
+			// Growing fee before first position has no effect on the unclaimed rewards
+			// of either position because they are not initialized at that point.
+			expectedUnclaimedRewardsPositionOne: cl.EmptyCoins,
+			// For position two, growing fee has no effect on the unclaimed rewards
+			// because we never update it, only create it.
+			expectedUnclaimedRewardsPositionTwo: cl.EmptyCoins,
+		},
+		"2: fee charged between first and second call to InitOrUpdateFeeAccumulatorPosition, after positon one is created and before position two is created": {
```
czarcas7ic, line 922:
```suggestion
		"2: fee charged between first and second call to InitOrUpdateFeeAccumulatorPosition, after position one is created and before position two is created": {
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -888,39 +954,60 @@ func (s *KeeperTestSuite) TestUpdateFeeAccumulatorPosition() {
 			s.SetupTest()
 
 			// Setup two cl pools
-			poolOne := s.PrepareConcentratedPool()
+			pool := s.PrepareConcentratedPool()
```
czarcas7ic, line 957: This says setup two cl pools but I only see one

x/concentrated-liquidity/fees_test.go
```diff
@@ -1065,6 +1152,88 @@ func (s *KeeperTestSuite) TestFunctionalFees() {
 	s.CollectAndAssertFees(s.Ctx, clPool.GetId(), totalFeesExpected, positionIds, ticksActivatedAfterEachSwapTest, denomsExpected, positions)
 }
 
+// This test focuses on various functional testing around fees and LP logic.
+// It tests invariants such as the following:
+// - can create positions in the same range, swap between them and yet collect the correct fees.
+// - correct proportions of fees for overlapping positions are withrawn.
```
czarcas7ic, line 1158:
```suggestion
// - correct proportions of fees for overlapping positions are withdrawn.
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -1065,6 +1152,88 @@ func (s *KeeperTestSuite) TestFunctionalFees() {
...
+// - withdrawing full liquidity claims correctly under the hood.
+// - withdrawing partial liquidity does not withdraw but still lest to claims fees as desired.
```
czarcas7ic, line 1160:
```suggestion
// - withdrawing partial liquidity does not withdraw but still lets fee claim as desired.
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -1065,6 +1152,88 @@ func (s *KeeperTestSuite) TestFunctionalFees() {
...
+	amtDenom0, _, err := concentratedLiquidityKeeper.WithdrawPosition(ctx, owner, positionIdTwo, fullLiquidity)
+	s.Require().NoError(err)
+	balanceAfterWithdraw := s.App.BankKeeper.GetBalance(ctx, owner, ETH)
+
+	// Validate that the correct amount of ETH was collected in withdraw for position two.
+	// total fees * full liquidity / (full liquidity + half liquidity)
+	expectedPositionToWithdraw := totalFeesExpected.AmountOf(ETH).ToDec().Mul(fullLiquidity.Quo(fullLiquidity.Add(halfLiquidity))).TruncateInt()
+	s.Require().Equal(expectedPositionToWithdraw.String(), balanceAfterWithdraw.Sub(balanceBeforeWithdraw).Amount.Sub(amtDenom0).String())
+
+	// Validate cannot claim for withrawn position.
```
czarcas7ic, line 1218:
```suggestion
	// Validate cannot claim for withdrawn position.
```

x/concentrated-liquidity/fees_test.go
```diff
@@ -1065,6 +1152,88 @@ func (s *KeeperTestSuite) TestFunctionalFees() {
 	s.CollectAndAssertFees(s.Ctx, clPool.GetId(), totalFeesExpected, positionIds, ticksActivatedAfterEachSwapTest, denomsExpected, positions)
 }
 
+// This test focuses on various functional testing around fees and LP logic.
+// It tests invariants such as the following:
+// - can create positions in the same range, swap between them and yet collect the correct fees.
+// - correct proportions of fees for overlapping positions are withrawn.
+// - withdrawing full liquidity claims correctly under the hood.
+// - withdrawing partial liquidity does not withdraw but still lest to claims fees as desired.
+func (s *KeeperTestSuite) TestFunctional_Fees_LP() {
```
czarcas7ic, line 1161: Nice test

x/concentrated-liquidity/fees_test.go
```diff
@@ -1065,6 +1152,88 @@ func (s *KeeperTestSuite) TestFunctionalFees() {
...
+	_, err = s.App.ConcentratedLiquidityKeeper.CollectFees(ctx, owner, positionIdTwo)
+	s.Require().Error(err)
+
+	feesCollected = s.collectFeesAndCheckInvariance(ctx, 0, DefaultMinTick, DefaultMaxTick, positionIdOne, sdk.NewCoins(), []string{ETH}, [][]sdk.Int{ticksActivatedAfterEachSwap})
+
+	// total fees * half liquidity / (full liquidity + half liquidity)
+	expectesFeesCollected := totalFeesExpected.AmountOf(ETH).ToDec().Mul(halfLiquidity.Quo(fullLiquidity.Add(halfLiquidity))).TruncateInt()
+	s.Require().Equal(expectesFeesCollected.String(), feesCollected.AmountOf(ETH).String())
+
+	// Create position in the dafault range 3.
```
czarcas7ic, line 1228:
```suggestion
	// Create position in the default range 3.
```

x/concentrated-liquidity/keeper_test.go
```diff
@@ -313,3 +310,25 @@ func (s *KeeperTestSuite) validateListenerCallCount(
...
+
+// Crosses the tick and charges the fee on the global fee accumulator.
+// This mimics crossing an initialized tick during a swap and charging the fee on swap completion.
+func (s *KeeperTestSuite) crossTickAndChargeFee(poolId uint64, tickIndexToCross int64) {
+	// Cross the tick to update it.
+	_, err := s.App.ConcentratedLiquidityKeeper.CrossTick(s.Ctx, poolId, tickIndexToCross, DefaultFeeAccumCoins[0])
+	s.Require().NoError(err)
+	err = s.App.ConcentratedLiquidityKeeper.ChargeFee(s.Ctx, poolId, DefaultFeeAccumCoins[0])
+	s.Require().NoError(err)
+}
```
czarcas7ic, line 322: For my learning, does the order of these matter? In my mind, it feels like one should charge the fee before crossing the tick, but it could be the same.

x/concentrated-liquidity/keeper_test.go
```diff
@@ -313,3 +310,25 @@ func (s *KeeperTestSuite) validateListenerCallCount(
...
+	accum, err := s.App.ConcentratedLiquidityKeeper.GetFeeAccumulator(s.Ctx, poolId)
+	s.Require().NoError(err)
+	positionRecord, err := accum.GetPosition(cltypes.KeyFeePositionAccumulator(positionId))
+	s.Require().NoError(err)
+	if expectedUnclaimedRewards.IsZero() {
+		s.Require().Equal(expectedUnclaimedRewards, positionRecord.UnclaimedRewards)
+	} else {
+		s.Require().Equal(expectedUnclaimedRewards[0].Amount.Mul(DefaultLiquidityAmt), positionRecord.UnclaimedRewards.AmountOf(expectedUnclaimedRewards[0].Denom))
+	}
+}
```
czarcas7ic, line 334: Should we just take in a DecCoin here? Is it possible for fee growth to contain more than just a single denom?

x/concentrated-liquidity/lp_test.go
```diff
@@ -597,6 +598,13 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 				expectedAfterLastPoolPositionRemovedCallCount = 1
 			}
 			s.validateListenerCallCount(0, 0, expectedAfterLastPoolPositionRemovedCallCount, 0)
+
+			// Dumb sanity-check that creating a positon with the same liquidity amount after fully removing it does not error.
```
czarcas7ic, line 602:
```suggestion
			// Dumb sanity-check that creating a position with the same liquidity amount after fully removing it does not error.
```

x/concentrated-liquidity/lp_test.go
```diff
@@ -854,15 +862,17 @@ func (s *KeeperTestSuite) TestUpdatePosition() {
...
-		"try updating with ticks outside existing position's tick range - error because fee accumulator is uninitialized": {
-			poolId:         1,
-			ownerIndex:     0,
-			lowerTick:      DefaultUpperTick + 1,
-			upperTick:      DefaultUpperTick + 100,
-			joinTime:       DefaultJoinTime,
-			liquidityDelta: DefaultLiquidityAmt,
-			numPositions:   1,
-			expectedError:  true,
+		"try updating with ticks outside existing position's tick range - initializes the accummulator": {
```
czarcas7ic, line 865:
```suggestion
		"try updating with ticks outside existing position's tick range - initializes the accumulator": {
```

x/concentrated-liquidity/lp_test.go
```diff
@@ -875,14 +885,18 @@ func (s *KeeperTestSuite) TestUpdatePosition() {
...
+			lowerTick:                 DefaultLowerTick,
+			upperTick:                 DefaultUpperTick,
+			joinTime:                  DefaultJoinTime,
+			liquidityDelta:            DefaultLiquidityAmt,
+			numPositions:              1,
+			amount0Expected:           DefaultAmt0Expected,
+			amount1Expected:           DefaultAmt1Expected,
+			expectedPositionLiquidity: DefaultLiquidityAmt,
+			expectedTickLiquidity:     DefaultLiquidityAmt.Add(DefaultLiquidityAmt),
+			expectedPoolLiquidity:     DefaultLiquidityAmt.Add(DefaultLiquidityAmt),
```
czarcas7ic, line 899: This test case still says error but we no longer expect an error here.

x/concentrated-liquidity/lp_test.go
czarcas7ic: Also, "try" implies this test case should error

czarcas7ic (approved):
Nice work! I only have nits

x/concentrated-liquidity/fees_test.go
```diff
@@ -55,59 +56,83 @@ func (s *KeeperTestSuite) TestInitializeFeeAccumulatorPosition() {
...
 	}
 
 	clKeeper := s.App.ConcentratedLiquidityKeeper
 
 	type initFeeAccumTest struct {
-		name           string
-		positionFields positionFields
+		name             string
+		positionFields   positionFields
+		isPositionUpdate bool
```
AlpinYukseloglu, line 93: Might be missing it but where is this test field currently used?

x/concentrated-liquidity/fees_test.go
```diff
@@ -847,39 +893,59 @@ func (s *KeeperTestSuite) TestQueryAndCollectFees() {
...
 	tests := map[string]updateFeeAccumPositionTest{
-		"happy path": {
-			owner:            ownerOne,
-			positionIdSetup:  DefaultPositionId,
-			positionIdUpdate: DefaultPositionId,
-			liquidity:        DefaultLiquidityAmt,
-			updatedLiquidity: DefaultLiquidityAmt.Mul(sdk.NewDec(2)),
-			lowerTick:        DefaultLowerTick,
-			upperTick:        DefaultUpperTick,
+		"1: fee charged prior to first call to InitOrUpdateFeeAccumulatorPosition with position one": {
```
AlpinYukseloglu, line 912: Out of curiosity, is there a specific reason we're numbering these test cases?

x/concentrated-liquidity/lp_test.go
AlpinYukseloglu: Did a double take on this test case – I would have expected it to error since the ticks don't line up, but since `positionId` is not specified (and therefore set to 0), it ends up initializing a new position at ID 0. Is this the intended behavior @p0mvn?

x/concentrated-liquidity/keeper_test.go
AlpinYukseloglu: Since fee is always charged on the input token, I believe it can be either of the two denoms in the pool

AlpinYukseloglu (approved):
Nice work, LGTM once comments are addressed/ack'd

x/concentrated-liquidity/keeper_test.go
czarcas7ic: So then unclaimedRewards could potentially have two indices right? And it looks like we are only using the first index here.

.vscode/launch.json
p0mvn: Yeah, I'm keeping this until all comments are addressed to help with debugging if any major issues arise

x/concentrated-liquidity/fees_test.go
p0mvn: removed obsolete comment

x/concentrated-liquidity/keeper_test.go
p0mvn: The order matters because if we charged fee first, the tick's accumulators would be updated with the new value.

The order is consistent with our [swap logic](https://github.com/osmosis-labs/osmosis/blob/72762cc3c239dad8a1b5d64000cd0cf6111cecd9/x/concentrated-liquidity/swaps.go#L318-L393)

Note that we cross ticks as we swap, and then "charge fee" after the loop completes

x/concentrated-liquidity/lp_test.go
p0mvn: renamed

x/concentrated-liquidity/fees_test.go
p0mvn: yeah, it makes it easier to run a specific one with the debugger. I filter the desired test case by number since most of the names are the same, making it difficult to filter only one test case

x/concentrated-liquidity/keeper_test.go
p0mvn: Updated to validate both denoms if two are exptected: [e89cb65](https://github.com/osmosis-labs/osmosis/pull/4973/commits/e89cb6581c985df2de9f1c6562bfbaa4a5ecf8ef)

x/concentrated-liquidity/fees_test.go
p0mvn: No, removed this obsoleted flag

x/concentrated-liquidity/fees.go
```diff
@@ -65,72 +73,49 @@ func (k Keeper) initializeFeeAccumulatorPosition(ctx sdk.Context, poolId uint64,
...
 
-	// Replace the position's accumulator in the fee accumulator with a new one
-	// that has the latest fee growth outside of the tick range.
-	err = preparePositionAccumulator(feeAccumulator, positionKey, feeGrowthOutside)
-	if err != nil {
-		return err
-	}
+	if !hasPosition {
+		if !liquidityDelta.IsPositive() {
+			return fmt.Errorf("liquidityDelta (%s) must be positive for a new position with id (%d)", liquidityDelta, positionId)
```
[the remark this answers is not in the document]
p0mvn: Changed to custom

.vscode/launch.json
czarcas7ic: Its feels ridiculous that we need to add these for each individual test to use the debugger... surely there is some other way / authomation :(

x/concentrated-liquidity/keeper_test.go
czarcas7ic: Makes sense, resolving!

x/concentrated-liquidity/lp_test.go
p0mvn: I added a new method that does proper validation of the attempted update:
[106f64e](https://github.com/osmosis-labs/osmosis/pull/4973/commits/106f64ee4fd729bac8b338cf5d10e142d430b2a1)

Updated the test cases per requests

p0mvn:
Planning on merging this since all comments are now addressed

PR #4983, opened 2023-04-22
implement gas consume on denom creation

larry0x:

See on-chain proposal: https://www.mintscan.io/osmosis/proposals/489

This PR mirrors the [implementation by Juno](https://github.com/CosmosContracts/token-factory/pull/14) although implementation detail slightly differs (see `createdenom.go`). cc @Reecepbcups 

## Brief Changelog

- Add a `denom_creation_gas_consume` to tokenfactory params
- Consume a amount of gas corresponding to `denom_creation_gas_consume` when creating a new denom
- In v16 upgrade, set `denom_creation_fee` to nil and `denom_creation_gas_consume` to 10,000,000 gas 

## Testing and Verifying

- Added a test for the chain upgrade in v16/upgrades_test.go

  - Does this pull request introduce a new feature or user-facing behavior changes? it adds a new field to tokenfactory's params
  - How is the feature or change documented? updated x/tokenfactory/README.md


x/tokenfactory/keeper/createdenom.go
```diff
@@ -71,16 +71,25 @@ func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdeno
...
+			return err
+		}
+
+		if err := k.communityPoolKeeper.FundCommunityPool(ctx, params.DenomCreationFee, accAddr); err != nil {
 			return err
 		}
 	}
+
+	// if DenomCreationGasConsume is non-zero, consume the gas
+	if params.DenomCreationGasConsume != 0 {
```
Reecepbcups, line 90: What is the use case for requiring both DenomCreationFee and DenomCreationGasConsume?

The entire point of DenomCreationGasConsume is to remove the CreationFee token to make contract easier. What am I missing

larry0x: In your implementation the gas is only consumed if the fee is nil. However the variable name does not imply this behavior.

Imagine a developer setting gas consume to a non-zero value but the gas isn't consumed. How confusing would it be! The developer will need to read the comments or dig into the code to find out that they also need to set fee to nil.

In my opinion the logic in my implementation here is simpler and less likely to cause confusion.

Reecepbcups: Gotcha good point. Will update our implementation

app/upgrades/v16/upgrades.go
```diff
@@ -63,6 +66,12 @@ func CreateUpgradeHandler(
...
 		}
 
+		updateTokenFactoryParams(ctx, keepers.TokenFactoryKeeper)
+
 		return migrations, nil
 	}
 }
+
+func updateTokenFactoryParams(ctx sdk.Context, tokenFactoryKeeper *tokenfactorykeeper.Keeper) {
+	tokenFactoryKeeper.SetParams(ctx, tokenfactorytypes.NewParams(nil, NewDenomCreationGasConsume))
```
nicolaslara, line 76: I assume this is purposely setting the fee to nil, but commenting in case the intention was to add gas-consume without modifying the fee.

larry0x: @nicolaslara Proposal 489 (https://www.mintscan.io/osmosis/proposals/489) states that the fee is to be set "to zero in the UpgradeHandler of the next chain upgrade".

I'm going to add a test for the gas consumption then this PR should be ready.

larry0x:
One thing I want to get you guys' opinions on is the specific value for the gas consumption. Due to the very low gas price on Osmosis, the cost for creating a new denom will be significantly lower than 10 OSMO. If we set the gas consumption to 10M gas, then under `0.0025uosmo` gas price it'd only be 0.025 OSMO per denom. Is this ok?

What is Osmosis' block gas limit?

Reecepbcups:
> What is Osmosis' block gas limit?
120,000,000
```bash
osmosisd q params subspace baseapp BlockParams
key: BlockParams
subspace: baseapp
value: '{"max_bytes":"10485760","max_gas":"120000000"}'
```

larry0x:
I'm setting the gas consume amount to 40M, which corresponds to 0.1 OSMO per denom created.

x/tokenfactory/keeper/createdenom.go
```diff
@@ -71,16 +71,25 @@ func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdeno
...
-	if err != nil {
-		return err
-	}
-	if creationFee != nil {
-		if err := k.communityPoolKeeper.FundCommunityPool(ctx, creationFee, accAddr); err != nil {
+	params := k.GetParams(ctx)
+
+	// if DenomCreationFee is non-zero, transfer the tokens from the creator
+	// account to community pool
+	if params.DenomCreationFee != nil {
```
mattverse, line 78: would it be possible to add according unit tests to test this?

mattverse (commented):
logic LGTM, let's merge this after having test cases with the according changes

app/upgrades/v16/upgrades.go
```diff
@@ -81,6 +84,12 @@ func CreateUpgradeHandler(
...
```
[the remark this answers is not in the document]
larry0x: For backward compatibility mostly

x/tokenfactory/keeper/createdenom.go
larry0x: Will do

larry0x: @mattverse I've added the test, please see: https://github.com/osmosis-labs/osmosis/pull/4983/commits/e12c0680cadc348b6e9c66d0d336ff0db2b317b6

mattverse (approved):
Nice work! LGTM

Issue #4915, opened 2023-04-14
[CL][Incentives] Add supported uptimes as a module param


AlpinYukseloglu:
## Background

We want governance to be able to add/remove supported uptimes in the future while minimizing pool migration logic that needs to be run (or in general any overhead that comes from it). One approach for this is to have a preset number of options in the form of a available uptimes constant (what we currently have under SupportedUptimes) and then have a module param that is a subset of this list that determines which uptimes incentives can be created for.

## Suggested Design

- Rename `SupportedUptimes` to `PossibleUptimes`
- Add a `SupportedUptimes` module param with validation logic that ensures it's a subset of `PossibleUptimes`
- Update incentive record creation logic to ensure that only supported uptimes can be incentivized

## Acceptance Criteria

- All new and existing tests should pass

PR #4984, opened 2023-04-22
[CL][Incentives] Add authorized uptimes as a module param

AlpinYukseloglu:
Closes: #4915

This PR adds `AuthorizedUptimes` as a module param to CL. Note that all core logic will continue to accommodate the full list of `SupportedUptimes` to avoid requiring a state migration in case governance wants to enable charging/authorize more uptimes. As such, `AuthorizedUptimes` will always be a subset of `SupportedUptimes`, with a default value of just supporting 1 nanosecond (equivalent to status quo incentives.

Since accumulator updates are skipped when accums are empty, the performance overhead for continuing to accommodate these uptimes should be minimal.

## Brief Changelog

- Add `AuthorizedUptimes` as a module param
- Modify incentive creation logic to check against `AuthorizedUptimes` instead of `SupportedUptimes`
- Update incentive creation tests to test a range of authorized uptimes

## Testing and Verifying

- Changes are tested in `incentives_test.go`


x/concentrated-liquidity/incentives_test.go
```diff
@@ -2910,6 +2912,7 @@ func (s *KeeperTestSuite) TestCreateIncentive() {
 		recordToSet        types.IncentiveRecord
 		existingRecords    []types.IncentiveRecord
 		minimumGasConsumed uint64
+		authorizedUptimes  []time.Duration
```
AlpinYukseloglu, line 2915: Note to reviewer: we add this field to test a range of supported uptimes and, importantly, to ensure that existing test cases do not pass or fail trivially due to only authorizing 1 nanosecond uptime after disabling charging

x/incentives/keeper/distribute_test.go
```diff
@@ -181,6 +189,13 @@ func (suite *KeeperTestSuite) TestDistributeToConcentratedLiquidityPools() {
 			// We fix blocktime to ensure tests are deterministic
 			suite.Ctx = suite.Ctx.WithBlockTime(defaultGaugeStartTime)
 
+			// Set up authorized CL uptimes to robustly test distribution
+			if tc.authorizedUptimes != nil {
+				clParams := suite.App.ConcentratedLiquidityKeeper.GetParams(suite.Ctx)
+				clParams.AuthorizedUptimes = tc.authorizedUptimes
+				suite.App.ConcentratedLiquidityKeeper.SetParams(suite.Ctx, clParams)
+			}
+
```
czarcas7ic, line 198: Every test case here defines authorizedUptimes as the supportedUptimes. Do you want to add some other test cases that do not touch this branch? If not / you do not see any value in it, we can just remove the param from the test case entirely and set it on every test run here.

x/incentives/keeper/distribute_test.go
```diff
@@ -726,6 +746,14 @@ func (suite *KeeperTestSuite) TestDistributeConcentratedLiquidity() {
 		suite.Run(tc.name, func() {
 			suite.SetupTest()
 			suite.Ctx = suite.Ctx.WithBlockTime(defaultBlockTime)
+
+			// Set up authorized CL uptimes to robustly test distribution
+			if tc.authorizedUptimes != nil {
+				clParams := suite.App.ConcentratedLiquidityKeeper.GetParams(suite.Ctx)
+				clParams.AuthorizedUptimes = tc.authorizedUptimes
+				suite.App.ConcentratedLiquidityKeeper.SetParams(suite.Ctx, clParams)
+			}
```
czarcas7ic, line 755: Same comment here

x/concentrated-liquidity/incentives.go
```diff
@@ -851,15 +851,21 @@ func (k Keeper) CreateIncentive(ctx sdk.Context, poolId uint64, sender sdk.AccAd
...
+	// authorized by governance.
+	authorizedUptimes := k.GetParams(ctx).AuthorizedUptimes
+	osmoutils.SortSlice(authorizedUptimes)
+
 	validUptime := false
-	for _, supportedUptime := range types.SupportedUptimes {
-		if minUptime == supportedUptime {
+	for _, authorizedUptime := range authorizedUptimes {
+		if minUptime == authorizedUptime {
 			validUptime = true
```
czarcas7ic, line 864: I think we can break here on setting to true to save on iterations

x/concentrated-liquidity/types/params.go
```diff
@@ -149,3 +157,40 @@ func validateBalancerSharesDiscount(i interface{}) error {
...
+		return fmt.Errorf("authorized uptimes cannot be empty")
+	}
+
+	// Check if each passed in uptime is in the list of supported uptimes
+	for _, uptime := range authorizedUptimes {
+		supported := false
+		for _, supportedUptime := range SupportedUptimes {
+			if uptime == supportedUptime {
+				supported = true
+			}
```
czarcas7ic, line 187: I think we can also break here to save on unnecessary iterations.

czarcas7ic (approved):
Sorry my review got split up for some reason when I was reviewing via vscode. All looks good to me with minor nits!

x/incentives/keeper/distribute_test.go
AlpinYukseloglu: I had originally tabled this to avoid touching gauge logic per this issue #4985, but the change was small enough where I just tied it into this PR. Wider range of authorized uptimes (including default of 1ns) should be tested now.

x/incentives/keeper/distribute_test.go
AlpinYukseloglu: See c31100c75598a516e5b3106140f3b4c0654549f5 and ec471d648016de079af19f01b52771e6e5f98c39

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: Good point!

AlpinYukseloglu: 4060398207ce4038a9ab1307c30180f62b75aa17

x/concentrated-liquidity/types/params.go
AlpinYukseloglu: 4060398207ce4038a9ab1307c30180f62b75aa17

p0mvn:
Please make sure that when you run:
`make localnet-start-with-state`, the script does not break:
https://github.com/osmosis-labs/osmosis/tree/main/tests/cl-genesis-positions

The way to check is to confirm that there is a CL pool with id 1 and many positions created upon running `make localnet-start-with-state`

From what I can tell, there is one update that is necessary atm:
```bash
# from project root
cd tests/cl-genesis-positions
go get github.com/osmosis-labs/osmosis/v15@4060398
# observe go mod updating, commit it
```

Let me know if you need help with this but please do not merge, unless confirmed that the script still works to avoid breaking integrator testing

x/concentrated-liquidity/incentives.go
```diff
@@ -851,15 +851,24 @@ func (k Keeper) CreateIncentive(ctx sdk.Context, poolId uint64, sender sdk.AccAd
 		return types.IncentiveRecord{}, types.NonPositiveEmissionRateError{PoolId: poolId, EmissionRate: emissionRate}
 	}
 
-	// Ensure min uptime is one of the supported periods
+	// Ensure min uptime is one of the authorized uptimes.
+	// Note that this is distinct from the supported uptimes – while we set up pools and positions to
+	// accommodate all supported uptimes, we only allow incentives to be created for uptimes that are
+	// authorized by governance.
+	authorizedUptimes := k.GetParams(ctx).AuthorizedUptimes
+	osmoutils.SortSlice(authorizedUptimes)
```
p0mvn, line 859: What is the purpose of sorting?

p0mvn (approved):
LGTM upon testing localosmosis script not breaking

Also, please add a test for checking the param in the upgrade handler by adding a validation below these lines:
https://github.com/osmosis-labs/osmosis/blob/85a748062dce94efaacb488ffcca6c9d25c5e588/app/upgrades/v16/upgrades_test.go#L99-L101

x/concentrated-liquidity/incentives.go
AlpinYukseloglu: My assumption is that uptimes can technically be stored in non-increasing order, which I believe we might have some assumptions leaning on (even if just in tests). This was intended to be a low-cost guardrail to ensure that ordering is not an issue. Do you see an issue with this reasoning?

p0mvn:
@AlpinYukseloglu FYI starting to work on localosmosis, will merge once completed

p0mvn:
Tested localosmosis, it passes now. Should be good to merge as long as CI passes

Issue #4731, opened 2023-03-24
[Routing] Add a registry to check if a chain allows forwarding


nicolaslara:
Epic: https://github.com/osmosis-labs/osmosis/issues/4344
Continuation of https://github.com/osmosis-labs/osmosis/pull/4694

## Background

We need to check that the chains in the path allow forwarding before sending the transaction to prevent packets from getting stuck in the middle

## Suggested Design

 * Add a new registry

## Acceptance Criteria

 * The registry exists and is tested
 * XCS checks this before sending the packets

ValarDragon:
Is this easiest as a new column in the registry table, or as a gating condition for inclusion to the registry, and automatically checked by the contract. (Similar to how we were considering making contracts test that the counterparty is the same as Osmosis)

nicolaslara:
I was thinking as a column, but only having channels we can route through be accepted might be a better idea 🤔

PR #4993, opened 2023-04-24
Add PFM validation

nicolaslara:
Closes: #4731

Add a registry to check if a chain supports PFM

 * To propose validation for a chain, user can send a token from that chain.
 * Once it has properly been unwrapped, we store the chain as supporting PFM
 

## Brief Changelog

 * Add PFM validation
 * The registry checks that the chains support PFM before sending the tokens
 * [TODO] Add validation for channels as well? (maybe in a separate PR)

## Testing and Verifying

Added tests specific to setting forwarding

All tests should pass


cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +41,82 @@ pub struct ContractAliasInput {
...
+    ctx: (DepsMut, Env, MessageInfo),
+    chain: String,
+) -> Result<Response, ContractError> {
+    let (deps, env, info) = ctx;
+
+    let chain = chain.to_lowercase();
+
+    // Store the chain to validate. If validation fails this will be reverted
+    CHAIN_PFM_MAP.save(deps.storage, &chain, &ChainPFM::default())?;
+    // TODO: Check if the chain exists and only allow override on certain conditions
```
nicolaslara, line 54: Just noticed I forgot this todo. Who should be allowed to override this? I'd say anyone that can manage that chain.

This is different to the other registries because we're actually validating it, so overriding it would only unset it until the response is received.

cosmwasm/contracts/crosschain-registry/src/contract.rs
```diff
@@ -27,6 +30,15 @@ pub fn instantiate(
     let state = Config { owner };
     CONFIG.save(deps.storage, &state)?;
 
+    CHAIN_PFM_MAP.save(
+        deps.storage,
+        "osmosis",
```
ValarDragon, line 35:
```suggestion
        // TODO: Move this to a constant to be source-chain
        "osmosis",
```

(Also doesn't this cause issues in the E2E tests?)

nicolaslara: I can add it to the config and set it to "osmosis" when the contract is initialized.

We want the proper name on the chain so that it can be referenced (for example when querying `getChannel("juno", "osmosis")`)

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: Two cases:
 * Chain had PFM, but then removes it
 * Someone tried to add PFM to the chain but it didn't validate (we would have the old attempt in storage). I think the right call here is to always replace it if it isn't approved so we can try to validate again

If we assume chains will not remove PFM this becomes simpler. Otherwise the question is still: who can change this?

cosmwasm/contracts/crosschain-registry/src/contract.rs
nicolaslara: added these as consts

cosmwasm/packages/registry/src/error.rs
```diff
@@ -92,6 +98,9 @@ pub enum RegistryError {
 
     #[error("bech32 prefix does not exist for chain: {chain}")]
     Bech32PrefixDoesNotExist { chain: String },
+
+    #[error("Chain {chain} does not support forwarding")]
+    ForwardingUnsopported { chain: String },
```
p0mvn, line 103: nit: typo

cosmwasm/packages/registry/src/error.rs
```diff
@@ -30,6 +33,9 @@ pub enum RegistryError {
     #[error("serialization error: {error}")]
     SerialiaztionError { error: String },
 
+    #[error("registry inprolperly configured")]
+    InproperlyConfigured {},
```
p0mvn, line 37:
```suggestion
    #[error("registry improperly configured")]
    ImproperlyConfigured {},
```

cosmwasm/contracts/crosschain-registry/src/state.rs
```diff
@@ -47,6 +49,19 @@ impl<T: AsRef<str>> From<(T, bool)> for RegistryValue {
     }
 }
 
+#[cw_serde]
+#[derive(Default)]
+pub struct ChainPFM {
+    pub acknowledged: bool,
+    pub validated: bool,
```
p0mvn, line 56: What is the difference between the two booleans? Consider adding comments

cosmwasm/contracts/crosschain-registry/src/msg.rs
```diff
@@ -44,3 +50,32 @@ pub use registry::msg::{
...
+pub enum IBCLifecycleComplete {
+    #[serde(rename = "ibc_ack")]
+    IBCAck {
+        /// The source channel (osmosis side) of the IBC packet
+        channel: String,
+        /// The sequence number that the packet was sent with
+        sequence: u64,
+        /// String encoded version of the ack as seen by OnAcknowledgementPacket(..)
+        ack: String,
+        /// Weather an ack is a success of failure according to the transfer spec
```
p0mvn, line 64:
```suggestion
        /// Whether an ack is a success of failure according to the transfer spec
```

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +42,100 @@ pub struct ContractAliasInput {
...
+
+    // check if the chain is already registered or is in progress
+    if let Some(chain_pfm) = CHAIN_PFM_MAP.may_load(deps.storage, &chain)? {
+        if chain_pfm.is_validated() {
+            // Only authorized addresses can ask for a validated PFM to be re-checked
+            // If sender is the contract governor, then they are authorized to do do this to any chain
+            // Otherwise, they must be authorized to do manage the chain they are attempting to modify
+            let user_permission =
+                check_is_authorized(deps.as_ref(), info.sender, Some(chain.clone()))?;
+            check_action_permission(FullOperation::Change, user_permission)?;
```
p0mvn, line 74: Asking to learn: what is the reason it works this way?

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +42,100 @@ pub struct ContractAliasInput {
...
+        } else {
+            return Err(ContractError::PFMValidationAlreadyInProgress {
+                chain: chain.clone(),
+            });
+        }
```
p0mvn, line 79: In the else case we return `PFMValidationAlreadyInProgress`. Does not being validated guarantee that validation has started?

p0mvn: It seems to me that this case represents as "have been proposed but not necessarily started validation" as validation is a separate msg execute

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +42,100 @@ pub struct ContractAliasInput {
...
+        None,
+        own_addr.to_string(),
+        env.block.time,
+        format!(r#"{{"ibc_callback":"{}"}}"#, own_addr),
+        Some(Callback {
+            contract: own_addr,
+            msg: format!(r#"{{"validate_pfm": {{"chain": "{}"}} }}"#, chain).try_into()?,
+        }),
+        true,
+    )?;
```
p0mvn, line 101: Is my understanding correct that we send an ibc transfer and on successful completion, a wasm hook is called that runs `validate_pfm`?

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: I think it's just different semantics here. The way I was thinking about it is: when a PFM check is proposed, then validation is considered to be in progress (propose = a proposal to validate PFM on a chain).

nicolaslara: >  Does not being validated guarantee that validation has started?

Here it does, as we are calling `is_validated()` on a ChainPFM fetched from storage. If it is in storage, it means someone has proposed it. If validation fails, we remove it from storage

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: This is to be compatible with how the permissions for the other registries work. Here the action is always Change, but there the action is passed in the message.

The reason only the contract governor, or someone authorized to manage the chain can change this is that changing this value once packaged are already flowing through this contract would DOS them, so we can't let anyone execute this.

In this case, the denial of service would only be temporary (while the packet is in-flight) but someone could keep proposing PFM checks and keep the validation in progress. This ensures only authorized parties can do that.

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: yes! correct

cosmwasm/contracts/crosschain-registry/src/state.rs
nicolaslara: acknowledged = the IBC ack has been received.
validated = the validate_pfm wasm hook has been called (to check that what we received is what we expected) and the data has been updated.

nicolaslara:
Thanks for the feedback! I'll fix the nits/typos/comments

nicolaslara:
Removing the stale label

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -19,6 +19,41 @@ pub fn hash_denom_trace(unwrapped: &str) -> String {
...
+    let th = hasher0.finalize();
+
+    let mut hasher = Sha256::new();
+    hasher.reset();
+    hasher.update(th.as_slice());
+    hasher.update(sender.as_bytes());
+
+    let result = hasher.finalize();
+
+    let result_u5 = bech32::convert_bits(result.as_slice(), 8, 5, true)?;
```
iboss-ptk, line 42: comments on what's happening here/ some reference would be nice. Eg. 
- why hashing "ibc-wasm-hook-intermediary" string
- why convert 8 to 5 bits with padding. (and can't this be just `result.to_base32()` without the `filter_map` below?)

iboss-ptk: just a thought but I think `ibc-hook` deserves to have it's own crate for using with contract. Or maybe live in `osmosis-std`.

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +42,100 @@ pub struct ContractAliasInput {
     pub new_alias: Option<String>,
 }
 
+pub fn propose_pfm(
```
iboss-ptk, line 45: wondering if I'm understading this correctly:

- the proposer can be anyone and basically trying to send coin cross-chain with PFM memo
- waiting for ack ensure that the sent packet is ack-ed -> `acknowledged = true`
- `ibc-hook` execute this contract with `validate_pfm` -> `validated = true`
  - `validate_pfm` also checks if sender is actually from `ibc-hook`and first originated from this contract
- these info lives in registry contract ensuring that target chain is pfm enabled

cosmwasm/contracts/crosschain-registry/src/execute.rs
```diff
@@ -40,6 +42,100 @@ pub struct ContractAliasInput {
...
+) -> Result<Response, ContractError> {
+    let (deps, env, info) = ctx;
+
+    let chain = chain.to_lowercase();
+
+    let registry = Registry::default(deps.as_ref());
+    let channel = registry.get_channel(&chain, CONTRACT_CHAIN)?;
+    let own_addr = env.contract.address.as_str();
+    let original_sender = registry.encode_addr_for_chain(own_addr, &chain)?;
+    let expected_sender = registry::derive_wasmhooks_sender(&channel, &original_sender, "osmo")?;
```
iboss-ptk, line 118: as we have `CONTRACT_CHAIN` why don't we also have `CONTRACT_CHAIN_BECH32_PREFIX` for "osmo" as well?

x/ibc-rate-limit/contracts/rate-limiter/src/sudo.rs
```diff
@@ -21,6 +21,7 @@ pub fn process_packet(
     #[cfg(test)] channel_value_mock: Option<Uint256>,
 ) -> Result<Response, ContractError> {
     let (channel_id, denom) = packet.path_data(&direction);
+    #[allow(clippy::needless_borrow)]
```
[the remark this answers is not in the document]
nicolaslara: basically cause this is not the contract I'm working on, and just added that to get the linter to pass

cosmwasm/packages/registry/src/registry.rs
nicolaslara: yeah, this should probably live in osmosis-std. @iboss-ptk Can we work together in designing a good reusable api for it?

>  can't this be just result.to_base32()

No, but I can't remember why. I'll go through the code and add some comments for the future

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -19,6 +19,41 @@ pub fn hash_denom_trace(unwrapped: &str) -> String {
...
+    let result_u5: Vec<bech32::u5> = result_u5
+        .iter()
+        .filter_map(|&x| bech32::u5::try_from_u8(x).ok())
```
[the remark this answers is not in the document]
nicolaslara: same as above. I spent some time trying to simplify this, and there were some reasons related to the internals of the bech32 crate for why the simple way of doing this didn't work. Unfortunately I can't remember the details. I'll review it and add some comments, but if you have a better way I'd love to simplify this

cosmwasm/packages/registry/src/registry.rs
```diff
@@ -354,6 +418,7 @@ impl<'a> Registry<'a> {
         block_time: Timestamp,
         first_transfer_memo: String,
         receiver_callback: Option<Callback>,
+        skip_forwarding_check: bool,
```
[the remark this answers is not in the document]
nicolaslara: yes! we normally check if forwarding is allowed (i.e.: chain has pfm) and if not, we don't let users send the packet. When we want to validate PFM, we don't yet have information about the chain allowing forwarding or not (we *think* it does), so we have to skip the check and send the packet

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: correct! One caveat though... now that PFM enforces an intermediate sender... we will need to update this to validate that.

cosmwasm/contracts/crosschain-registry/src/execute.rs
nicolaslara: yeah, that makes sense. I was mostly being lazy there

cosmwasm/contracts/crosschain-registry/src/error.rs
```diff
@@ -16,9 +16,31 @@ pub enum ContractError {
     #[error("Unauthorized")]
     Unauthorized {},
 
+    #[error("Chain validation not started for {chain}")]
+    ValidationNotFound { chain: String },
+
+    #[error("coin from invalid chain. It belongs to {supplied_chain} and should be from {expected_chain}")]
+    CoinFronInvalidChain {
```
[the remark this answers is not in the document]
nicolaslara: fixed

x/ibc-rate-limit/contracts/rate-limiter/src/sudo.rs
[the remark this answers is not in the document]
nicolaslara: fixed in main and merged

cosmwasm/packages/registry/src/registry.rs
nicolaslara: added some comments. AFAICT there is no cleaner way to do this, but happy to be proven wrong

cosmwasm/packages/registry/src/registry.rs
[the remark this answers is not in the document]
nicolaslara: looked into this more. It seems like convert_bits still leaves the types as a `Vec<u8>` even if the values have been modified, so the filter map is just changing the type. If there's a cleaner solution here, I'd love to change it

iboss-ptk (approved):

Issue #4967, opened 2023-04-20
bug: osmomath.Power does not work for base values that are close to zero


p0mvn:
## Background

`osmomath.Power` computes power by 

$$exp2(exponent * log_2 base)$$

where `exp2` is defined only non-negative values

log is negative near zero. As a result, `Power` breaks.

## Suggested Design

TBD

## Acceptance Criteria
- add test cases for base close to zero
- investigate a solution

PR #5000, opened 2023-04-25
refactor(osmomath/cl): avoid using broken power function; panic on broken values

p0mvn:
Addresses: #4967

Addresses #4967 in the context of concentrated liquidity. Keeping #4967 open until completely resolved in the future

The `osmomath.Power` function is broken for base < 1. The reason is that the logarithm function is
negative between zero and 1, and the Exp2(k) is undefined for negative k.
As a result, this function panics if called for base < 1 as it is implemented by `Exp2(base.LogBase2().Mul(exponent))`

## Brief Changelog

- replace `Power` function with `PowerInteger` in `x/concentrated-liquidity module
- panic in `osmomath.Power` foe base < 1

## Testing and Verifying

- added test


czarcas7ic (approved):
This LGTM, but we still need to replace the osmomath hash in go.mod

Issue #4899, opened 2023-04-12
(CL): Investigate appropriate default swap fees, handle migrations of pool with different swap fees


mattverse:
## Background

https://github.com/osmosis-labs/osmosis/blob/91e61673b9/x/concentrated-liquidity/types/params.go#L33-L44

We have default swap fees chosen that is a param in CL Keeper. 
We want to have decision on what would be the appropriate number for these swap fees. 

Also, https://github.com/osmosis-labs/osmosis/blob/91e61673b9/app/upgrades/v16/concentrated_pool.go#L21 in the v16 upgrade handler attempts to migrate Balancer pool into a CL pool. We would also need to decide on how we want to handle edge case where existing swap fee of the pool is not one of the allowed swap fee in CL Pools. 

## Acceptance Criteria

- Figure out values that we want to use for default swap fees
- Figure out how we want to migrate pools with different swap fees

czarcas7ic:
@mattverse  Here is data of the first 1000 pools in Osmosis, sorted from most to least used:

```
[...]

PR #5003, opened 2023-04-25
[CL]: default swap fees

czarcas7ic:
Closes: #4899 

After speaking with Johnny, we decided on an initial set of authorized swap fees we can launch with. This can always we changed later via governance. The fees are as follows:

.01%
.05%
.1%
.2%
.3%
.5%

Additionally, there was a question on migrating future pools that do not fall into one of these above fee categories. There is no logic in the migration link function that requires the cl pool being linked to have the same swap fee as the balancer pool, so this so be a non issue.

## Brief Changelog

- Adds authorized swap fees as a const
- Sets DefaultParams to the authorized swap fees set in the const 


p0mvn (approved):
LGTM once tests pass

PR #5008, opened 2023-04-26
Simulator fix 

stackman27:
Closes: #XXX

sim fix

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/x/concentrated-liquidity/simulation/sim_msgs.go
+++ b/x/concentrated-liquidity/simulation/sim_msgs.go
@@ -55,8 +55,8 @@ func RandMsgCreatePosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
 
 	accountBalancePoolDenom0 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[0])
 	accountBalancePoolDenom1 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[1])
-	if accountBalancePoolDenom0.Amount.LT(sdk.Int(tokens[0].Amount)) || accountBalancePoolDenom1.Amount.LT(sdk.Int(tokens[1].Amount)) {
-		return nil, fmt.Errorf("insufficient funds")
+	if accountBalancePoolDenom0.Amount.LT(tokens[0].Amount) || accountBalancePoolDenom1.Amount.LT(tokens[1].Amount) {
+		return nil, fmt.Errorf("insufficient funds when creating a concentrated position")
 	}
 
 	return &cltypes.MsgCreatePosition{
@@ -101,7 +101,7 @@ func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sd
 	}
 
 	if withdrawAmount.GTE(position.Liquidity) {
-		return nil, fmt.Errorf("Insufficient funds")
+		return nil, fmt.Errorf("Insufficient funds from a  concentrated position")
 	}
 
 	return &cltypes.MsgWithdrawPosition{
--- a/x/concentrated-liquidity/simulation/sim_msgs.go
+++ b/x/concentrated-liquidity/simulation/sim_msgs.go
@@ -25,7 +25,6 @@ func RandomMsgCreateConcentratedPool(k clkeeper.Keeper, sim *osmosimtypes.SimCtx
 	}
 
 	// make sure the denoms are valid authorized quote denoms
-
 	defaultParams := cltypes.DefaultParams()
 	defaultParams.IsPermissionlessPoolCreationEnabled = true
 	defaultParams.AuthorizedQuoteDenoms = append(defaultParams.AuthorizedQuoteDenoms, coin1.Denom, coin0.Denom)
--- a/x/concentrated-liquidity/simulation/sim_msgs.go
+++ b/x/concentrated-liquidity/simulation/sim_msgs.go
@@ -172,7 +172,7 @@ func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Con
 		}
 
 		// perform swap from token0 to token1 until either token0 or token1 fund runs out
-		_, err = k.SwapExactAmountIn(ctx, swapOwner.Address, poolI, sdk.NewCoin(swapOwnerTokens[0].Denom, randToken0Amt), swapOwnerTokens[1].Denom, sdk.OneInt(), poolI.GetSwapFee(ctx))
+		_, err = k.SwapExactAmountIn(ctx, swapOwner.Address, poolI, sdk.NewCoin(swapOwnerTokens[0].Denom, randToken0Amt), swapOwnerTokens[1].Denom, sdk.OneInt(), sdk.ZeroDec())
 		if err != nil {
 			return nil, err
 		}
--- a/x/concentrated-liquidity/simulation/sim_msgs.go
+++ b/x/concentrated-liquidity/simulation/sim_msgs.go
@@ -55,7 +55,7 @@ func RandMsgCreatePosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
 
 	accountBalancePoolDenom0 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[0])
 	accountBalancePoolDenom1 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[1])
-	if accountBalancePoolDenom0.Amount.LT(tokens[0].Amount) || accountBalancePoolDenom1.Amount.LT(tokens[1].Amount) {
+	if accountBalancePoolDenom0.Amount.LT(sdk.Int(tokens[0].Amount)) || accountBalancePoolDenom1.Amount.LT(sdk.Int(tokens[1].Amount)) {
 		return nil, fmt.Errorf("insufficient funds")
 	}
 
@@ -72,7 +72,7 @@ func RandMsgCreatePosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
 }
 
 func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgWithdrawPosition, error) {
-	rand := sim.GetSeededRand("select random seed")
+	rand := sim.GetRand()
 	// get random pool
 	clPool, _, err := getRandCLPool(k, sim, ctx)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -201,7 +220,7 @@ func RandMsgCollectIncentives(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx s
 }
 
 func RandMsgCreateIncentives(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Context) (*cltypes.MsgCreateIncentive, error) {
-	rand := sim.GetRand()
+	rand := sim.GetSeededRand("select random seed")
```
p0mvn, line 223: I think each action gets domain separation so this shouldn't matter. ref: https://github.com/osmosis-labs/osmosis/blob/f436c298c67d59408c457228789530c7843a7cd2/simulation/executor/simulate.go#L241-L242

suggest reverting to reduce complexity

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -53,6 +53,12 @@ func RandMsgCreatePosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.
 		return nil, err
 	}
 
+	accountBalancePoolDenom0 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[0])
+	accountBalancePoolDenom1 := sim.BankKeeper().GetBalance(ctx, positionCreator, poolDenoms[1])
+	if accountBalancePoolDenom0.Amount.LT(tokens[0].Amount) || accountBalancePoolDenom1.Amount.LT(tokens[1].Amount) {
+		return nil, fmt.Errorf("insufficient funds")
```
p0mvn, line 59:
```suggestion
		return nil, fmt.Errorf("insufficient funds when creating a concentrated position")
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -94,6 +100,19 @@ func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sd
 		return nil, fmt.Errorf("Invalid withdraw Amount")
 	}
 
+	poolTotalLiquidity, err := k.GetTotalPoolLiquidity(ctx, clPool.GetId())
+	if err != nil {
+		return nil, err
+	}
+
+	if len(poolTotalLiquidity) < 2 {
+		return nil, fmt.Errorf("pool doesnot have 2 tokens")
```
p0mvn, line 109:
```suggestion
		return nil, fmt.Errorf("pool does not have 2 tokens")
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -94,6 +100,19 @@ func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sd
...
+	}
+
+	if poolTotalLiquidity[0].Amount.LTE(withdrawAmount.RoundInt()) || poolTotalLiquidity[1].Amount.LTE(withdrawAmount.RoundInt()) {
+		return nil, fmt.Errorf("Insufficient funds to withdraw")
```
p0mvn, line 113:
```suggestion
		return nil, fmt.Errorf("Insufficient funds to withdraw from a  concentrated position")
```

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -94,6 +100,19 @@ func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sd
...
+	poolTotalLiquidity, err := k.GetTotalPoolLiquidity(ctx, clPool.GetId())
+	if err != nil {
+		return nil, err
+	}
+
+	if len(poolTotalLiquidity) < 2 {
+		return nil, fmt.Errorf("pool doesnot have 2 tokens")
+	}
+
+	if poolTotalLiquidity[0].Amount.LTE(withdrawAmount.RoundInt()) || poolTotalLiquidity[1].Amount.LTE(withdrawAmount.RoundInt()) {
```
p0mvn, line 112: I think more importantly we need to check that a positon has enough liquidity to withdraw from. If a position has enough, then the pool will have enough for sure as well

stackman27:
resolved comments @p0mvn lmk what you think

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -94,6 +100,10 @@ func RandMsgWithdrawPosition(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sd
 		return nil, fmt.Errorf("Invalid withdraw Amount")
 	}
 
+	if withdrawAmount.GTE(position.Liquidity) {
```
p0mvn, line 103: Should we allow equality?

p0mvn (approved):

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: sounds good

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -143,22 +153,24 @@ func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Con
 		return nil, fmt.Errorf("user does not have pool tokens")
 	}
 
-	// perform swap until token 1 runs out
+	// perform 50 random swaps until token 0 runs out
 	remainingSwapOwnerToken0Amt := swapOwnerTokens[0].Amount
-	for remainingSwapOwnerToken0Amt.ToDec().TruncateInt().GT(sdk.ZeroInt()) {
+	swapsRemaining := 50
```
stackman27, line 158: @p0mvn also added this so that we only swap 50 time or until user's token0 runs out 

before we were swapping small amount for >100 times making simulator take a lot of time to run

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -158,7 +168,7 @@ func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Con
 			return nil, err
 		}
 
-		remainingSwapOwnerToken0Amt = remainingSwapOwnerToken0Amt.Sub(randToken0Amt)
+		remainingSwapOwnerToken0Amt = remainingSwapOwnerToken0Amt.Sub(randToken0Amt).ToDec().TruncateInt()
```
mattverse, line 171: what's the reason we do int -> dec -> int conversion?

alexanderbez (approved):

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: i couldn't just truncate sdk.Int. but i can truncate sdk.Dec to lowest int. So i had to convert it to dec first. please lmk if there's a better way of handling this

czarcas7ic: The result is already truncated, there is no need to turn it to a dec to re-truncate it

x/concentrated-liquidity/simulation/sim_msgs.go
```diff
@@ -146,9 +156,9 @@ func RandMsgCollectFees(k clkeeper.Keeper, sim *osmosimtypes.SimCtx, ctx sdk.Con
 	// perform swap until token 1 runs out
 	remainingSwapOwnerToken0Amt := swapOwnerTokens[0].Amount
 	for remainingSwapOwnerToken0Amt.ToDec().TruncateInt().GT(sdk.ZeroInt()) {
```
czarcas7ic, line 158: Ditto

x/concentrated-liquidity/simulation/sim_msgs.go
stackman27: oh yea that was dumb of me cause its an integer

x/incentives/keeper/gauge.go
```diff
@@ -108,9 +109,12 @@ func (k Keeper) CreateGauge(ctx sdk.Context, isPerpetual bool, owner sdk.AccAddr
 		}
 	}
 
-	// Ensure that the denom this gauge pays out to exists on-chain
-	if !k.bk.HasSupply(ctx, distrTo.Denom) && !strings.Contains(distrTo.Denom, "osmovaloper") {
-		return 0, fmt.Errorf("denom does not exist: %s", distrTo.Denom)
+	// check if distrTo.Denom is uosmo, if not ensure that the denom this gauge
+	// pays out to exists on-chain
+	if distrTo.Denom != appParams.BaseCoinUnit {
```
stackman27, line 114: @p0mvn want to make sure this a fair assumption to make? uosmo should ALWAYS exists on-chain right?

p0mvn: It is a fair assumption. However, I do not understand why we would want to have it? Why can't we check this even for uosmo

stackman27: its because in simulator we were randomly generating denoms and often times the sender wouldnot have uosmo, or not have enough supply of uosmo, and it would constantly fail here

x/incentives/keeper/gauge.go
```diff
@@ -108,9 +109,12 @@ func (k Keeper) CreateGauge(ctx sdk.Context, isPerpetual bool, owner sdk.AccAddr
...
+		if !k.bk.HasSupply(ctx, distrTo.Denom) && !strings.Contains(distrTo.Denom, "osmovaloper") {
```
czarcas7ic, line 115: Why would the denom ever contain osmovaloper?

stackman27: yea this looks like address prefix, not sure why this check was placed here. I didn't wanted to remove the coverage

x/incentives/keeper/gauge.go
p0mvn: Make sense, ty! I think we should then get in https://github.com/osmosis-labs/osmosis/pull/4768 and remove this line.

We should avoid hacking regular code to make convenience for tests

I followed up with @phamminh0811 to refresh #4768 and merge it. 

Please ping me if it's taking longer than Monday, and I can refresh and merge it

p0mvn: https://github.com/osmosis-labs/osmosis/pull/5028 wtih the relevant change is now merged

stackman27: fixed this https://github.com/osmosis-labs/osmosis/pull/5008/commits/5edb0345ba631d66d05b73f20eaa9a8a4f947b54

p0mvn (approved):

PR #5009, opened 2023-04-26
remove global rand seed from the simulator

p0mvn:
Closes: #XXX

In Go, the global rand variable is seeded by a default source, which is an implementation of a Source interface defined in the math/rand package. This source is based on the current time in nanoseconds and the process ID, and it is initialized the first time the math/rand package is imported.

In the simulator, we create a custom-seeded rand so that the simulator runs are deterministic.

This PR attempts to fix non-determinism by removing the use of global rand and using a simulator seeded rand.

## Testing and Verifying

### DIFF
--- a/x/gamm/simulation/sim_msgs.go
+++ b/x/gamm/simulation/sim_msgs.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	legacysimulationtype "github.com/cosmos/cosmos-sdk/types/simulation"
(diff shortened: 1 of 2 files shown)


stackman27:
We actually have used "math/rand" in other sim modules too. it's present in 
x/incentives/simulation/operations.go
x/superfluid/simulation/operations.go

p0mvn:
> x/incentives/simulation/operations.go

Note the way it is used there:
https://github.com/osmosis-labs/osmosis/blob/f436c298c67d59408c457228789530c7843a7cd2/x/incentives/simulation/operations.go#L66

It is used as an import for a parameter type. This is different from what is being fixed here which is using a global source of randomness

stackman27 (approved):

Issue #4962, opened 2023-04-19
CLI:  Add gas config to the client.toml config file 


nicolaslara:
## Background

Many CLI txs require custom gas configs. This is annoying to write every time and leads to long commands. We can add a config for this in client.toml specifying the default if they are not set

## Suggested Design

Base it on https://github.com/CosmosContracts/juno/pull/601

## Acceptance Criteria

Txs on the CLI don't require `--gas auto --gas-prices 0.1uosmo --gas-adjustment 1.3` as those can be directly picked up from the config

hieuvubk:
Is anyone working on this?

nicolaslara:
no, just added it yesterday when I was sick of copy pasting the gas params :p

PR #5020, opened 2023-04-28
Add gas config to the client.toml

hieuvubk:
Closes: #4962 

Save `gas`, `gas-prices`, `gas-adjustment` into `client.toml`. So user no more needs to pass gas flags in cmd, it will get data from config.

## Brief Changelog
- Overwrite `ConfigCmd`
- Use `OsmosisCustomClient ` instead of `ClientConfig`: the purpose is to integrate some of the fields mentioned above
- The config fields will be set & get directly on the client.toml file

## Flow
- You can configure `gas` or any field with the command
`osmosisd config gas 1000000`
- Run tx cmd, no need to pass `gas` flag more. Its will get data that saved in `client.toml`
- If user wanna change they config, use `osmosisd config [key] [value]`


faddat:
@hieuvubk can you please add a changelog to this pr?

faddat (approved):
looks good, once it's got a changelog!

heh, gotcha covered on it :)

czarcas7ic:
@hieuvubk has this been compiled and tested locally?

hieuvubk:
> @hieuvubk has this been compiled and tested locally?

yes, I tested in my locally, can write & cmd can read data from `client.toml`. 
One thing I wanna to ask, should we config `fees` cause I see all tx cmd need at least one fee denom.

czarcas7ic:
@hieuvubk yeah that would be great, please ping when that has been added and tested against mainnet and I will merge!

hieuvubk:
@czarcas7ic Fees added, this is my test tx against mainnet using `osmosisd tx bank send from to 100000uosmo --from key` without any flag. 
https://www.mintscan.io/osmosis/txs/87887D790741A45E495E3C5AB3CD89E49AC1BB906994B820B7D54C68624A86C3

czarcas7ic:
@hieuvubk will merge as soon as you get sim to pass!

hieuvubk:
@czarcas7ic  It passed after rerunning. Seem st wrong in the last run

czarcas7ic (approved):

czarcas7ic:
Nice work @hieuvubk , merging!

Issue #4646, opened 2023-03-17
fix: `CalcAmount0Delta` `Quo` -> `QuoTruncate`


p0mvn:
## Background

We should investigate if we need to truncate at precision end when `roundUp` in `math.CalcAmount0Delta` is false

## Suggested Design

Investigate this: https://github.com/osmosis-labs/osmosis/blob/7223b8b6ca5e5e3b6a17592935e72f82a910968c/x/concentrated-liquidity/internal/math/math.go#L72

## Acceptance Criteria
- if that should be a `QuoTruncate`, apply the change and fix tests

czarcas7ic:
Hey! I did some investigating into this and the conclusion I made was this was not needed for the following reason:

In CalcAmount0Delta, we determine the denom by multiplying sqrtPriceA and sqrtPriceB. 

https://github.com/osmosis-labs/osmosis/blob/3767fa15b2c3f44bec75a44877c6a0935e5f6df4/x/concentrated-liquidity/internal/math/math.go#L61

This multiplication will only ever round the denom up, therefore only ever making the final value lower (even without truncating)

This means that QuoTruncate is not required.

The way I came to the above conclusion, is I created two test cases.

```
		"rounding check, round up false": {
			liquidity:       sdk.MustNewDecFromStr("1517882923.648669463472052100"),
			sqrtPCurrent:    sdk.MustNewDecFromStr("70.710678118654752440"), // 5000
			sqrtPUpper:      sdk.MustNewDecFromStr("74.161984870956629487"), // 5500
			round:           false,
			amount0Expected: "998976.999999999999999999",
		},
		"rounding check, round up true": {
			liquidity:       sdk.MustNewDecFromStr("1517882923.648669463472052100"),
			sqrtPCurrent:    sdk.MustNewDecFromStr("70.710678118654752440"), // 5000
			sqrtPUpper:      sdk.MustNewDecFromStr("74.161984870956629487"), // 5500
			round:           true,
			amount0Expected: "998977.000000000000000000",
		},
```

The intent was to find a liquidity value that resulted in an amount0 that was on the edge of the next integer value. 

While the "rounding check, round up false" test case should have resulted in the amount0Expected shown in the test case, it actually resulted in a value of "998976.999999999999999989". 

When breaking down each part of the equation, I saw that the denom being calculated was "5244.044240850757734943" instead of "5244.044240850757734942", which will always result in a smaller than actual value in the non roundUp branch.

[...]

PR #5022, opened 2023-04-28
refactor/test: CalcAmount0Delta and CalcAmount1Delta rounding in pool's favor

p0mvn:
Closes: #4646

Our `CalcAmount0Delta` and `CalcAmount1Delta` were not necessarily rounding in pool's favor. This PR adds truncation at precision end in the desired direction depending on the value of the `roundUp` flag.

There are 2 primary use cases of these functions:
- adding / withdrawing liquidity
- swapping

When adding liquidity, we want to round up so that the user is requested to provide a greater  token amount for liquidity `L`
When withdrawing liquidity, we want to round down so that the user receives smaller token amount for liquidity L (in favor of the pool).

In swaps,
token in should be rounded up so that the user has to provide more.
token out should be rounded down so that the user receives fewer tokens (in favor of the pool)

I attempted to construct the test case that would have a high amplification when operating on large amounts.
It seems that we had an undiscovered bug that was rounding in user's favor for large amounts.

These test cases would break if we were to revert back to the old rounding logic:
https://github.com/osmosis-labs/osmosis/blob/e526991cdcdc6fa62df6bb7f286b00d189b603fb/x/concentrated-liquidity/math/math_test.go#L260-L295

## Brief Changelog

- changed rounding direction
- refactored tests
- added unit tests to attempt to amplify the rounding error

## Testing and Verifying

- Covered by existing and added new,

### DIFF
--- a/x/concentrated-liquidity/lp_test.go
+++ b/x/concentrated-liquidity/lp_test.go
@@ -368,7 +368,8 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 			setupConfig: baseCase,
 			sutConfigOverwrite: &lpTest{
 				amount0Expected: baseCase.amount0Expected, // 0.998976 eth
-				amount1Expected: baseCase.amount1Expected, // 5000 usdc
+				// Note: subtracing one due to truncations in favor of the pool when withdrawing.
+				amount1Expected: baseCase.amount1Expected.Sub(sdk.OneInt()), // 5000 usdc
 			},
 			timeElapsed: defaultTimeElapsed,
 		},
@@ -417,7 +418,8 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 			setupConfig: baseCase,
 			sutConfigOverwrite: &lpTest{
 				amount0Expected: baseCase.amount0Expected, // 0.998976 eth
-				amount1Expected: baseCase.amount1Expected, // 5000 usdc
+				// Note: subtracing one due to truncations in favor of the pool when withdrawing.
+				amount1Expected: baseCase.amount1Expected.Sub(sdk.OneInt()), // 5000 usdc
 			},
 			timeElapsed: 0,
 		},
@@ -856,16 +858,16 @@ func (s *KeeperTestSuite) TestUpdatePosition() {
 			expectedError:             false,
 		},
 		"update existing position with negative amount": {
-			poolId:         1,
-			ownerIndex:     0,
-			lowerTick:      DefaultLowerTick,
-			upperTick:      DefaultUpperTick,
-			joinTime:       DefaultJoinTime,
-			positionId:     DefaultPositionId,
-			liquidityDelta: DefaultLiquidityAmt.Neg(), // negative
+			poolId:          1,
+			ownerIndex:      0,
+			lowerTick:       DefaultLowerTick,
+			upperTick:       DefaultUpperTick,
+			joinTime:        DefaultJoinTime,
+			positionId:      DefaultPositionId,
+			liquidityDelta:  DefaultLiquidityAmt.Neg(), // negative
+			amount0Expected: DefaultAmt0Expected.Neg(),
 			// Note: rounds down in favor of the pool (compared to the positive case which rounds up).
-			amount0Expected:           DefaultAmt0Expected.Neg(),
-			amount1Expected:           DefaultAmt1Expected.Neg(),
+			amount1Expected:           DefaultAmt1Expected.Sub(sdk.OneInt()).Neg(),
 			expectedPositionLiquidity: sdk.ZeroDec(),
 			expectedTickLiquidity:     sdk.ZeroDec(),
 			expectedPoolLiquidity:     sdk.ZeroDec(),
@@ -1122,7 +1124,7 @@ func (s *KeeperTestSuite) TestInverseRelation_CreatePosition_WithdrawPosition() 
 			// 1. amount for denom0 and denom1 upon creating and withdraw position should be same
 			// Note: subtracting one because create position rounds in favor of the pool.
 			s.Require().Equal(amtDenom0CreatePosition.Sub(sdk.OneInt()).String(), amtDenom0WithdrawPosition.String())
-			s.Require().Equal(amtDenom1CreatePosition.String(), amtDenom1WithdrawPosition.String())
+			s.Require().Equal(amtDenom1CreatePosition.Sub(sdk.OneInt()).String(), amtDenom1WithdrawPosition.String())
 
 			// 2. user balance and pool balance after creating / withdrawing position should be same
 			userBalancePostPositionCreation := s.App.BankKeeper.GetAllBalances(s.Ctx, s.TestAccs[0])
@@ -1130,11 +1132,11 @@ func (s *KeeperTestSuite) TestInverseRelation_CreatePosition_WithdrawPosition() 
 
 			// Note: subtracing one since position creation rounds in favor of the pool.
 			s.Require().Equal(userBalancePrePositionCreation.AmountOf(ETH).Sub(sdk.OneInt()).String(), userBalancePostPositionCreation.AmountOf(ETH).String())
-			s.Require().Equal(userBalancePrePositionCreation.AmountOf(USDC).String(), userBalancePostPositionCreation.AmountOf(USDC).String())
+			s.Require().Equal(userBalancePrePositionCreation.AmountOf(USDC).Sub(sdk.OneInt()).String(), userBalancePostPositionCreation.AmountOf(USDC).String())
 
 			// Note: adding one since withdrawal rounds in favor of the pool.
 			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(ETH).Add(sdk.OneInt()).String(), poolBalancePostPositionCreation.AmountOf(ETH).String())
-			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(USDC).String(), poolBalancePostPositionCreation.AmountOf(USDC).String())
+			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(USDC).Add(sdk.OneInt()).String(), poolBalancePostPositionCreation.AmountOf(USDC).String())
 
 			// 3. Check that position's liquidity was deleted
 			positionLiquidity, err := clKeeper.GetPositionLiquidity(s.Ctx, tc.positionId)
@@ -1151,7 +1153,7 @@ func (s *KeeperTestSuite) TestInverseRelation_CreatePosition_WithdrawPosition() 
 
 			// Note: one ends up remaining due to rounding in favor of the pool.
 			s.Require().Equal(liquidityBefore.AmountOf(ETH).Add(sdk.OneInt()).String(), liquidityAfter.AmountOf(ETH).String())
-			s.Require().Equal(liquidityBefore.AmountOf(USDC).String(), liquidityAfter.AmountOf(USDC).String())
+			s.Require().Equal(liquidityBefore.AmountOf(USDC).Add(sdk.OneInt()).String(), liquidityAfter.AmountOf(USDC).String())
 		})
 	}
 }
--- a/x/concentrated-liquidity/lp_test.go
+++ b/x/concentrated-liquidity/lp_test.go
@@ -375,7 +375,8 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 		"withdraw full liquidity amount with underlying lock that has finished unlocking": {
 			setupConfig: baseCase,
 			sutConfigOverwrite: &lpTest{
-				amount0Expected:  DefaultAmt0,
+				// Note: subtracing one due to truncations in favor of the pool when withdrawing.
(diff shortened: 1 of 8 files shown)


x/concentrated-liquidity/math/math_test.go
```diff
@@ -150,25 +151,87 @@ func (suite *ConcentratedMathTestSuite) TestGetNextSqrtPriceFromAmount1RoundingD
...
+
+			roundingDir := osmomath.RoundUp
+			if !tc.roundUp {
+				roundingDir = osmomath.RoundDown
+			}
+
+			tolerance := osmomath.ErrTolerance{
+				MultiplicativeTolerance: sdk.SmallestDec(),
+				RoundingDir:             roundingDir,
+			}
```
p0mvn, line 230: Note: this invariant would not hold with the old rounding logic.

The additive error is in the order of magnitude of 6 with these numbers.

x/concentrated-liquidity/lp_test.go
```diff
@@ -855,16 +858,16 @@ func (s *KeeperTestSuite) TestUpdatePosition() {
...
-			positionId:     DefaultPositionId,
-			liquidityDelta: DefaultLiquidityAmt.Neg(), // negative
+			poolId:          1,
+			ownerIndex:      0,
+			lowerTick:       DefaultLowerTick,
+			upperTick:       DefaultUpperTick,
+			joinTime:        DefaultJoinTime,
+			positionId:      DefaultPositionId,
+			liquidityDelta:  DefaultLiquidityAmt.Neg(), // negative
+			amount0Expected: DefaultAmt0Expected.Neg(),
```
AlpinYukseloglu, line 868: It would be great if we can have a test case that covers asset 0 being rounded as well (e.g. one where we need to subtract 1 from asset 0 instead of asset 1)

x/concentrated-liquidity/lp_test.go
```diff
@@ -1121,19 +1124,19 @@ func (s *KeeperTestSuite) TestInverseRelation_CreatePosition_WithdrawPosition()
...
-			// Note: subtracing one since position creation rounds in favor of the pool.
+			// Note: subtracting one since position creation rounds in favor of the pool.
 			s.Require().Equal(userBalancePrePositionCreation.AmountOf(ETH).Sub(sdk.OneInt()).String(), userBalancePostPositionCreation.AmountOf(ETH).String())
-			s.Require().Equal(userBalancePrePositionCreation.AmountOf(USDC).String(), userBalancePostPositionCreation.AmountOf(USDC).String())
+			s.Require().Equal(userBalancePrePositionCreation.AmountOf(USDC).Sub(sdk.OneInt()).String(), userBalancePostPositionCreation.AmountOf(USDC).String())
 
 			// Note: adding one since withdrawal rounds in favor of the pool.
 			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(ETH).Add(sdk.OneInt()).String(), poolBalancePostPositionCreation.AmountOf(ETH).String())
-			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(USDC).String(), poolBalancePostPositionCreation.AmountOf(USDC).String())
+			s.Require().Equal(poolBalancePrePositionCreation.AmountOf(USDC).Add(sdk.OneInt()).String(), poolBalancePostPositionCreation.AmountOf(USDC).String())
```
AlpinYukseloglu, line 1139: Thoughts on having a variable called `roundingError` that's set to `OneInt()` for test readability?

x/concentrated-liquidity/math/math.go
```diff
@@ -58,19 +58,30 @@ func CalcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
 		sqrtPriceA, sqrtPriceB = sqrtPriceB, sqrtPriceA
 	}
 	diff := sqrtPriceB.Sub(sqrtPriceA)
-	denom := sqrtPriceA.Mul(sqrtPriceB)
 	// if calculating for amountIn, we round up
-	// if calculating for amountOut, we don't round at all
+	// if calculating for amountOut, we don't round but truncate at precision end
+	// in the rounding direction.
```
AlpinYukseloglu, line 63:
```suggestion
	// if calculating for amountOut, we round down
```

AlpinYukseloglu: These two statements are equivalent right?

x/concentrated-liquidity/math/math.go
```diff
@@ -58,19 +58,30 @@ func CalcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
 		return liq.Mul(diff).Quo(denom).Ceil()
 	}
-	// Investigate if this should be a QuoTruncate: https://github.com/osmosis-labs/osmosis/issues/4646
-	return liq.Mul(diff).Quo(denom)
+	// These are truncated at precision end to round in favor of the pool when:
+	// - calculating amount out during swap
+	// - withdrawing liquidity
+	// The denominator is rounded up to get a smaller final amount.
+	denom := sqrtPriceA.MulRoundUp(sqrtPriceB)
+	return liq.MulTruncate(diff).QuoTruncate(denom)
```
AlpinYukseloglu, line 84: Since we do fixed point multiplication, `Mul` shouldn't have any rounding error right? I would have expected this line to be:
```
return liq.Mul(diff).QuoTruncate(denom)
```

x/concentrated-liquidity/math/math.go
```diff
@@ -90,9 +101,18 @@ func CalcAmount1Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
+		// - calculating amountIn during swap
+		// - adding liquidity (request user to provide more tokens in in favor of the pool)
+		return liq.MulRoundUp(diff).Ceil()
 	}
-	return liq.Mul(diff)
+	// This is truncated at precision end to round in favor of the pool when:
+	// - calculating amount out during swap
+	// - withdrawing liquidity
+	// The denominator is rounded up to get a higher final amount.
+	return liq.MulTruncate(diff)
```
AlpinYukseloglu, line 115: Same point as above, but since this is the final output it might make sense to truncate. Still not 100% sure on what the source of the rounding error here would be, but I might have a misunderstanding of how `Mul` works.

x/concentrated-liquidity/swaps.go
```diff
@@ -394,7 +394,8 @@ func (k Keeper) calcOutAmtGivenIn(ctx sdk.Context,
 
 	// coin amounts require int values
 	// round amountIn up to avoid under charging
-	amt0 := tokenAmountInSpecified.Sub(swapState.amountSpecifiedRemaining).RoundInt()
+	// round amountOut down to avoid over refunding.
```
AlpinYukseloglu, line 397: nit: move this comment to right above `amt1` – did a double take because I thought it was referring to `amt0` (which rounds up)

AlpinYukseloglu (approved):
General approach LGTM but have a few comments I'd like addressed/ack'd before merging. Approving since I don't see any safety issues.

x/concentrated-liquidity/math/math_test.go
```diff
@@ -150,25 +151,87 @@ func (suite *ConcentratedMathTestSuite) TestGetNextSqrtPriceFromAmount1RoundingD
...
+			// min_sqrt_p = Decimal("0.000000152731791058")
+			// liq = Decimal("931361973132462178951297")
+			// liq * (max_sqrt_p - min_sqrt_p) / (max_sqrt_p * min_sqrt_p)
+			liquidity: sdk.MustNewDecFromStr("931361973132462178951297"),
+			// price: 0.000000000000023327
+			sqrtPA: sdk.MustNewDecFromStr("0.000000152731791058"),
+			// price: 952361284325389721913
+			sqrtPB:          sdk.MustNewDecFromStr("30860351331.852813530648276680"),
+			roundUp:         true,
+			amount0Expected: sdk.MustNewDecFromStr("6098022989717817431593106314408.888128101590393209").Ceil().String(), // truncated at precision end.
```
czarcas7ic, line 206: I think the comment here is wrong right? We round up at precision end here, not truncate

x/concentrated-liquidity/math/math_test.go
```diff
@@ -180,24 +243,64 @@ func (suite *ConcentratedMathTestSuite) TestCalcAmount0Delta() {
...
+			// min_sqrt_p = Decimal("0.000000152731791058")
+			// liq = Decimal("931361973132462178951297")
+			// liq * (max_sqrt_p - min_sqrt_p)
+			liquidity: sdk.MustNewDecFromStr("931361973132462178951297"),
+			// price: 0.000000000000023327
+			sqrtPA: sdk.MustNewDecFromStr("0.000000152731791058"),
+			// price: 952361284325389721913
+			sqrtPB:          sdk.MustNewDecFromStr("30860351331.852813530648276680"),
+			roundUp:         true,
+			amount1Expected: sdk.MustNewDecFromStr("28742157707995443393876876754535992.801567623738751734").Ceil().String(), // truncated at precision end.
```
czarcas7ic, line 294: Same question about truncated comment here.

x/concentrated-liquidity/swapstrategy/one_for_zero_test.go
```diff
@@ -95,8 +95,9 @@ func (suite *StrategyTestSuite) TestComputeSwapStepOutGivenIn_OneForZero() {
 			// sqrt_price_current + token_in / liquidity
 			expectedSqrtPriceNext:    sqrtPriceTargetNotReached,
 			expectedAmountInConsumed: defaultAmountOne.Sub(sdk.NewDec(100)).Ceil(),
-			expectedAmountOut:        amountZeroTargetNotReached,
-			expectedFeeChargeTotal:   sdk.ZeroDec(),
+			// subtracting 3 * smallest dec to account for truncations in favor of the pool.
+			expectedAmountOut:      amountZeroTargetNotReached.Sub(sdk.SmallestDec().MulInt64(3)),
```
czarcas7ic, line 99: How was 3 selected here?

czarcas7ic (approved):
Logic looks good to me! Only thing that has me confused is multiplying 3 by smallest dec (just trying to understand where that number came from, the others where we only subtract smallest dec makes sense to me)

x/concentrated-liquidity/math/math.go
p0mvn: What I wanted to say is that we round down at the precision end, not to the nearest integer. Let me modify your suggestion and commit the update

x/concentrated-liquidity/swapstrategy/one_for_zero_test.go
p0mvn: I just reran the tests, noticed that this is by how much it differed, and accepted it since it's insignificant

I think what matters that the rounding direction is desired and the difference is small

x/concentrated-liquidity/math/math.go
p0mvn: The way `sdk.Dec` is implemented is that it allows the multiplication and other math operations to go beyond the fixed point precision, then chop it and round at precision end.

So imagine we had a `Dec` multiplication `0.12 * 0. 12 = 0.0144` with the precision of 2.
After performing the operation, we would chop precision and truncate for it to become 0.01

ref: https://github.com/osmosis-labs/cosmos-sdk/blob/7a2cf2993434124b84075aafc7483b8a5ecfd9de/types/decimal.go#L303-L304

p0mvn: By default, it does banker's rounding so with the above example it would still be `0.01`

However, say we have:
```
>>> 0.42 * 0.22
0.0924
```

This would become `0.1` with regular `Mut` and `0.09` with `MulTruncate`

x/concentrated-liquidity/math/math.go
p0mvn: I'm pretty sure this matters at the precision end for the reasons in the earlier thread.

Let me resolve this one but keep the one above open

x/concentrated-liquidity/swaps.go
p0mvn: done

x/concentrated-liquidity/math/math_test.go
p0mvn: fixed

x/concentrated-liquidity/lp_test.go
p0mvn: done

x/concentrated-liquidity/lp_test.go
p0mvn: The problem with this is that I'd have to recalculate a lot of the values by hand, including ticks, and liquidity between them corresponding to the given amounts.

Let me make an issue to come back to this separately once we're in internal review mode

p0mvn: https://github.com/osmosis-labs/osmosis/issues/5023

p0mvn:
@AlpinYukseloglu @czarcas7ic I addressed all comments here. Keeping this open till tomorrow.

Please let me know if any follow-ups

Issue #4914, opened 2023-04-14
[CL] Implement `AddToPosition` for non-superfluid positions


AlpinYukseloglu:
## Background

Since we've decided to table charging for our MVP for the sake of position fungibility, we need to implement a separate `AddToPosition` function and message that captures the fungibility-related behavior we need without requiring us to do an API break if we re-introduce charging and, secondarily, not hurting contract UX.

## Suggested Design

- The simplest way to achieve this is likely to have `AddToPosition` delete the old position and then create a new one with the combined amounts. This approach also happens to be safe with the charging paradigm and thus minimizes risk of API breaks in the future. It also lets us avoid having to deal with uptime/charge of exactly zero, as 1 nanosecond is sufficient for these purposes

## Acceptance Criteria

- All new and existing tests pass

p0mvn:
My impression was that for removing charging we concluded  on a different direction. Followed up with @AlpinYukseloglu to update on the latest.

Going to treat this issue as a general "Removing charging from CL" for now

PR #5026, opened 2023-04-29
[CL] Implement `AddToPosition` for non-superfluid positions

AlpinYukseloglu:
Closes: #4914

This PR adds the `AddToPosition` keeper function and message for non-superfluid positions.

## Brief Changelog

- Implement `AddToPosition` (withdraws full position then creates new one with added amounts)
- Add tests that cover rounding-related boundaries in a readable way

## Testing and Verifying

- All new functionality should be tested in `lp_test.go`


x/concentrated-liquidity/lp.go
```diff
@@ -250,6 +250,57 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
...
+	if err != nil {
+		return 0, sdk.Int{}, sdk.Int{}, err
+	}
+
+	if !anyPositionsRemainingInPool {
+		return 0, sdk.Int{}, sdk.Int{}, types.AddToLastPositionInPoolError{PoolId: position.PoolId, PositionId: position.PositionId}
+	}
+
+	// Create new position with updated liquidity.
+	newPositionId, actualAmount0, actualAmount1, _, _, err := k.createPosition(ctx, position.PoolId, owner, amount0Withdrawn.Add(amount0Added), amount1Withdrawn.Add(amount1Added), sdk.ZeroInt(), sdk.ZeroInt(), position.LowerTick, position.UpperTick)
```
czarcas7ic, line 296: imo, the min for amount0 and amount1 can be set to the previous amount0 and amount1 withdrawn

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+			setupConfig: baseCase,
+
+			// system under test parameters
+			sutConfigOverwrite: &lpTest{
+				amount0Expected: amount0PerfectRatio,
+				amount1Expected: amount1PerfectRatio,
+			},
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: roundingError,
+			amount1ToAdd: sdk.ZeroInt(),
```
czarcas7ic, line 679: I appreciate this test case being added!!

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+				// thus we cannot use the full amount of asset1. We calculate the output here (rounding up):
+				// https://www.wolframalpha.com/input?i=3035764327.860030912175533748+*+%2870.710678118654752440+-+67.416615162732695594%29
+				amount1Expected: sdk.NewInt(9999998816),
+				expectedError:   types.PositionSuperfluidStakedError{PositionId: uint64(1)},
+			},
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: amount0PerfectRatio,
+			amount1ToAdd: amount1PerfectRatio,
+
+			createLockUnlocked: true,
```
czarcas7ic, line 762: If the underlying lock has finished unlocking, then shouldn't we update the underlyinglock field accordingly and therefore allow to add to position with no error?

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+			sutConfigOverwrite: &lpTest{
+				lowerTick:     -1, // valid tick at which no position exists
+				positionId:    DefaultPositionId + 3,
+				expectedError: types.PositionIdNotFoundError{PositionId: DefaultPositionId + 3},
+			},
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: amount0PerfectRatio,
+			amount1ToAdd: amount1PerfectRatio,
+		},
+		"error: attempt to add to last position in pool": {
```
czarcas7ic, line 821: Good foresight on this

x/concentrated-liquidity/msg_server_test.go
```diff
@@ -85,6 +85,73 @@ func (suite *KeeperTestSuite) TestCreateConcentratedPool_Events() {
...
+			// Create a cl pool with a default position
+			pool := suite.PrepareConcentratedPool()
+
+			// Position from current account.
+			posId := suite.SetupDefaultPositionAcc(pool.GetId(), suite.TestAccs[0])
+
+			if tc.shouldSetupUnownedPosition {
+				// Position from another account.
+				suite.SetupDefaultPositionAcc(pool.GetId(), suite.TestAccs[1])
+			}
```
czarcas7ic, line 124: While the "last position in pool" case uses this, I think you should just remove it as a test field (since its being used here differently than what it's original intent was)

czarcas7ic (commented):
Only have nits. Not going to explicitly approve since I did a quicker than normal pass before I sign off for the night. Going to check it again tomorrow. Nice work!

x/concentrated-liquidity/lp.go
```diff
@@ -250,6 +250,57 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
...
 }
 
+// nolint: unused
+// addToPosition attempts to add liquidityAmount to a position with the given pool id in the given tick range.
+// Returns error if
+// - Withdrawing full position fails
+// - Creating new position with added liquidity fails
+// - Position is superfluid staked
+// TODO: handle adding to SFS positions
+func (k Keeper) addToPosition(ctx sdk.Context, owner sdk.AccAddress, positionId uint64, amount0Added, amount1Added sdk.Int) (uint64, sdk.Int, sdk.Int, error) {
```
p0mvn, line 260: I don't see position ownership checked anywhere. I know that it is checked in `withdrawPosition` but I think it would be great to have it here as well to avoid depending on the internal logic downstream for such an important security check.

Also, please add a test where it fails for non-owner trying to add to a position

x/concentrated-liquidity/lp.go
```diff
@@ -250,6 +250,57 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
...
+		return 0, sdk.Int{}, sdk.Int{}, err
+	}
+
+	if amount0Added.IsNegative() || amount1Added.IsNegative() {
+		return 0, sdk.Int{}, sdk.Int{}, types.NegativeAmountAddedError{PositionId: position.PositionId, Asset0Amount: amount0Added, Asset1Amount: amount1Added}
+	}
+
+	// If the position is superfluid staked, return error.
+	// TODO: handle this case to allow LPs to add to SFS positions
+	positionHasUnderlyingLock, err := k.positionHasUnderlyingLockInState(ctx, positionId)
```
czarcas7ic, line 272: Note on this, I think I am going to change this to "positionHasActiveUnderlyingLockInState", so the method will not only check if there is a lock state entry, but also check that that lock still exists

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+				s.Require().ErrorContains(err, config.expectedError.Error())
+				return
+			}
+
+			s.Require().NoError(err)
+			s.Require().Equal(config.amount0Expected.String(), newAmt0.String())
+			s.Require().Equal(config.amount1Expected.String(), newAmt1.String())
+
+			// We expect the position ID to be 3 since we have two setup positions
+			s.Require().Equal(uint64(3), newPosId)
```
p0mvn, line 940: Please check the balances. That is, that `amount0Expected` and `amount1Expected` were withdrawn from the user and moved to the pool (not amounts provided e.g. `amount0ToAdd`)

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+		"add base amount to existing liquidity with perfect ratio": {
+			// setup parameters for creating a pool and position.
+			setupConfig: baseCase,
+
+			// system under test parameters
+			sutConfigOverwrite: &lpTest{
+				amount0Expected: amount0PerfectRatio.Add(amount0PerfectRatio).Sub(roundingError),
+				// Since we round on the other the asset when we withdraw, asset0 turns into the bottleneck and
+				// thus we cannot use the full amount of asset1. We calculate the output here (rounding up):
+				// https://www.wolframalpha.com/input?i=3035764327.860030912175533748+*+%2870.710678118654752440+-+67.416615162732695594%29
```
p0mvn, line 690: So `70.710678118654752440` is the default 5k, how do you get `4545.0`? Could you add a breakdown of that please?

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+				amount1Expected: sdk.NewInt(9999998816),
```
p0mvn, line 691: Is it correct to say that we lost `5000000000 * 2 - 9999998816 = 1184` uosmo due to a rounding error? How does it scale with greater amounts being used?

x/concentrated-liquidity/lp_test.go
```diff
@@ -622,6 +622,326 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
...
+			sutConfigOverwrite: &lpTest{
+				amount0Expected: amount0PerfectRatio.Sub(roundingError),
+				// Since we round on the other the asset when we withdraw, asset0 turns into the bottleneck and
+				// thus we cannot use the full amount of asset1. We calculate the output here (rounding up):
+				// https://www.wolframalpha.com/input?i=1517881404.211361448703481013+*+%2870.710678118654752440+-+67.416615162732695594%29
+				amount1Expected: sdk.NewInt(4999996906),
+			},
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: sdk.ZeroInt(),
+			amount1ToAdd: sdk.ZeroInt(),
```
p0mvn, line 664: Why don't we error in this case?

x/concentrated-liquidity/lp_test.go
AlpinYukseloglu: I figured all lock-related logic would fall under the superfluid `AddToPosition` PR – I did not want to step on your toes by implementing some of it piecewise. Happy to add this in if you'd like though

x/concentrated-liquidity/lp.go
```diff
@@ -250,6 +250,57 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
 	return actualAmount0.Neg(), actualAmount1.Neg(), nil
 }
 
+// nolint: unused
```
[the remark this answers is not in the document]
AlpinYukseloglu: This was from an earlier version of the PR prior to the message level implementation. Removed now.

x/concentrated-liquidity/lp.go
```diff
@@ -250,6 +250,57 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
...
+// addToPosition attempts to add liquidityAmount to a position with the given pool id in the given tick range.
```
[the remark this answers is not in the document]
AlpinYukseloglu: Fixed

x/concentrated-liquidity/msg_server_test.go
AlpinYukseloglu: Good point – I changed it the test field to `lastPositionInPool` since I think we should still have a msg-level error case

x/concentrated-liquidity/lp.go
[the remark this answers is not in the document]
AlpinYukseloglu: Added clarifying comments for these two points

x/concentrated-liquidity/lp.go
AlpinYukseloglu: Nice catch, added in

x/concentrated-liquidity/lp_test.go
AlpinYukseloglu: Added a comment to the godoc to make this dynamic a bit clearer

x/concentrated-liquidity/lp_test.go
AlpinYukseloglu: Added check here 710deaa

x/concentrated-liquidity/lp_test.go
AlpinYukseloglu: Just to clarify, where do you see `4545.0`? The wolfram link simply recalculates the asset1 amount using `asset1 = L * (sqrtPriceUpper - sqrtPriceLower)` (added clarifying comments for this). For context, I pulled the sqrtPrices from the actual positions that were created.

x/concentrated-liquidity/lp_test.go
czarcas7ic: Sorry for the late response. Just to be clear, zero logic from addToPosition for superfluid overlaps with this PR. So you can treat any attempt to addToPosition with an underlying lock as an error, while any position that has completed unlocking should be going through this flow if that makes sense.

p0mvn (approved):
Test vector updates LGTM, nice work

PR #5035, opened 2023-05-01
feat(osmoutils): helper to find ith smallest min or max

p0mvn:
Closes: #XXX

UPD: related to #5365 

This is a passion-project to review some divide-and-conquer algorithms.

This introduces a helper to efficiently find the ith min or max (determined by the `less` function) in O(n).

More info can be found here: 
- https://brilliant.org/wiki/median-finding-algorithm/
- https://www.youtube.com/watch?v=EzeYI7p9MjU

There is not use case for this at this time but might be a good helper to have in `osmoutils`.

## Testing and Verifying

This change added tests.

### DIFF
--- a/osmoutils/generic_helper.go
+++ b/osmoutils/generic_helper.go
@@ -21,14 +21,14 @@ func MakeNew[T any]() T {
 }
 
 func Min[T constraints.Ordered](a, b T) T {
-	if a < b {
+	if a <= b {
 		return a
 	}
 	return b
 }
 
 func Max[T constraints.Ordered](a, b T) T {
-	if a > b {
+	if a >= b {
 		return a
 	}
 	return b
--- a/osmoutils/generic_helper.go
+++ b/osmoutils/generic_helper.go
@@ -1,6 +1,10 @@
 package osmoutils
 
-import "reflect"
+import (
+	"reflect"
+
+	"golang.org/x/exp/constraints"
+)
 
 // MakeNew makes a new instance of generic T.
 // if T is a pointer, makes a new instance of the underlying struct via reflection,
@@ -14,4 +18,18 @@ func MakeNew[T any]() T {
 	} else {
 		return *new(T) // v is not ptr, alloc with new
 	}
+}
+
+func Min[T constraints.Ordered](a, b T) T {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func Max[T constraints.Ordered](a, b T) T {
+	if a > b {
+		return a
+	}
+	return b
 }
--- a/osmoutils/slice_helper.go
+++ b/osmoutils/slice_helper.go
@@ -1,6 +1,7 @@
 package osmoutils
 
 import (
+	"fmt"
 	"math/rand"
 	"reflect"
 	"sort"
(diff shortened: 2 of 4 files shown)


osmoutils/slice_helper.go
```diff
@@ -92,3 +93,103 @@ func MergeSlices[T any](slice1, slice2 []T, less LessFunc[T]) []T {
...
+			greaterThanPivot = append(greaterThanPivot, cur)
+		}
+	}
+	numberOfElementsSmallerThanPivot := len(smallerThanPivot)
+
+	if i < numberOfElementsSmallerThanPivot {
+		return iThSmallest(smallerThanPivot, i, less)
+	} else if i > numberOfElementsSmallerThanPivot {
+		// Handle edge case where there are duplicate pivots and it is the result.
+		if len(greaterThanPivot) == 0 && numEqualToPivot > 1 {
```
pysel, line 184:
```suggestion
                // we return pivot, if `i` lays between the sets of elements smaller than pivot and bigger than pivot
		if i < originalLength-len(greaterThanPivot) && i > numberOfElementsSmallerThanPivot && numEqualToPivot > 1 {
```

pysel: the reason I started digging into this, is because adding:
```
"failing tc": {
			s:           []int{1, 2, 3, 4, 5, 6, 7, 7, 7, 7, 8, 9, 10},
			i:           9,
			less:        less,
			expected:    7,
			expectPanic: false,
		},
```

will fail the test.

pysel: I think, the current implementation only accounts for a situation, if the pivot is ithSmallest, but only if it is located at the end of the array we are recursing on

pysel: In the `failing tc` case, we pick 7 as a pivot, but because there are multiple copies of it and the selected pivot 7 is not ithSmallest, we continue recursing on the array of greater elements (which in turn results in out of bounds error, because we left the answer behind)

pysel: I changed my local branch to the suggested design here, and it successfully passed with the `failing test` test case included

x/concentrated-liquidity/swapstrategy/one_for_zero_test.go
```diff
@@ -144,8 +144,9 @@ func (suite *StrategyTestSuite) TestComputeSwapStepOutGivenIn_OneForZero() {
 			expectedSqrtPriceNext: sqrt(100_000_100),
 
 			expectedAmountInConsumed: sdk.NewDec(1336900668450),
-			expectedAmountOut:        defaultAmountZero.TruncateDec(),
-			expectedFeeChargeTotal:   sdk.ZeroDec(),
+			// subtracting smallest dec as a rounding error in favor of the pool.
+			expectedAmountOut:      defaultAmountZero.TruncateDec().Sub(sdk.SmallestDec()),
+			expectedFeeChargeTotal: sdk.ZeroDec(),
```
pysel, line 149: I think this was added by mistake

osmoutils/slice_helper_test.go
```diff
@@ -91,3 +92,134 @@ func TestContainsDuplicateDeepEqual(t *testing.T) {
...
+			// 1, 3, 4, 10, 12, 13, 13, 14, 21, 23, 45, 48, 86, 98, 100
+			s: []int{
+				4, 21, 3, 1, 10,
+				98, 23, 13, 48, 12,
+				100, 45, 14, 86, 13,
+			},
+			i:        10,
+			less:     greater,
+			expected: 12,
+		},
```
pysel, line 192: in case you decide to add changes from my previous comment, here is the test case:
```suggestion
		},
		"edge case: copied pivot": {
			s:           []int{1, 2, 3, 4, 5, 6, 7, 7, 7, 7, 8, 9, 10},
			i:           9,
			less:        less,
			expected:    7,
			expectPanic: false,
		},
```

osmoutils/slice_helper_test.go
```diff
@@ -91,3 +92,134 @@ func TestContainsDuplicateDeepEqual(t *testing.T) {
 		require.Equal(t, tt.want, got)
 	}
 }
+
+func TestIThSmallest(t *testing.T) {
```
pysel, line 96: shall we add float and string test cases to cover one of each of `Integer | Float | ~string` in `constraints.Ordered`?

pysel (changes requested):
it was really interesting reviewing this PR, thanks!

osmoutils/slice_helper.go
pysel: also, I removed the `len(greaterThanPivot) == 0` condition, could you please clarify why we need it? maybe I am missing something

osmoutils/generic_helper.go
```diff
@@ -15,3 +19,17 @@ func MakeNew[T any]() T {
...
+
+func Min[T constraints.Ordered](a, b T) T {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func Max[T constraints.Ordered](a, b T) T {
+	if a > b {
```
AlpinYukseloglu, line 31:
```suggestion
	if a >= b {
```

If the two values are equal, we should preserve the input order

pysel: shouldn't it be like this in the Min case then?

ValarDragon:
Closing for staleness. @pysel please don't remove stale labels for something not actively being worked on!

If work is restarted, lets re-open PR at that time :)

Issue #5029, opened 2023-04-30
[CL] Add support for adding to superfluid positions


AlpinYukseloglu:
## Background

In our current implementation of `AddToPosition`, we return an error if there is an underlying lock for the position being added to. We need to update this to instead break the lock and create a new lock (with the same underlying state e.g. days remaining for unlocking) for the new position. This is because `AddToPosition` actually deletes the old position and creates a new one with the added amounts (to be backwards compatible if we re-introduce charging), so we cannot simply add to the old position.

## Suggested Design

- Follow a similar process to what is done for migration logic for breaking the old lock for a position and creating a new one for the new position

## Acceptance Criteria

- All new and existing tests should pass

PR #5042, opened 2023-05-02
feat: add to superfluid CL position

czarcas7ic:
Closes: #5029 

This PR adds the `AddToPosition` logic, but for CL positions that are currently superfluid staked. This PR also introduces the required wiring for the respective message / CLI.

## Brief Changelog

- `addToConcentratedLiquiditySuperfluidPosition` added
- Respective message event and logic tests added


x/superfluid/keeper/concentrated_liquidity.go
```diff
@@ -0,0 +1,106 @@
...
+// It returns the new position ID, actual added amounts of both tokens, new liquidity, new lock ID, and any potential errors.
+//
+// Returns:
+// newPositionId: ID of the newly created concentrated liquidity position.
+// actualAmount0: Actual amount of token 0 added.
+// actualAmount1: Actual amount of token 1 added.
+// newLiquidity: The new liquidity value.
+// newLockId: ID of the lock associated with the new position.
+// error: Any error that may occur during execution.
+func (k Keeper) addToConcentratedLiquiditySuperfluidPosition(ctx sdk.Context, owner sdk.AccAddress, positionId uint64, amount0Added, amount1Added sdk.Int) (uint64, sdk.Int, sdk.Int, sdk.Dec, uint64, error) {
```
mattverse, line 27: nit: consider changing to:
```suggestion
func (k Keeper) addToConcentratedLiquiditySuperfluidPosition(ctx sdk.Context, owner sdk.AccAddress, positionId uint64, amount0ToAdd, amount1ToAdd sdk.Int) (uint64, sdk.Int, sdk.Int, sdk.Dec, uint64, error) {
```

x/superfluid/keeper/concentrated_liquidity.go
```diff
@@ -0,0 +1,106 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	cltypes "github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/types"
+	"github.com/osmosis-labs/osmosis/v15/x/superfluid/types"
+)
+
+// addToConcentratedLiquiditySuperfluidPosition adds the specified amounts of tokens to an existing superfluid staked
+// concentrated liquidity position. It performs the following steps:
+// 1. Validates the input amounts and position state.
+// 2. Superfluid undelegates the position and performs a force unlock.
```
mattverse, line 13:
```suggestion
// 2. Instantly superfluid undelegate the position by performing a force unlock.
```

x/superfluid/keeper/concentrated_liquidity_test.go
```diff
@@ -0,0 +1,247 @@
...
+			suite.Require().NoError(err)
+
+			// Check if the old intermediary account has no delegation.
+			_, found := stakingKeeper.GetDelegation(ctx, oldIntermediaryAcc, valAddr)
+			suite.Require().False(found)
+
+			// Check if the new intermediary account has delegation.
+			_, found = stakingKeeper.GetDelegation(ctx, newIntermediaryAcc, valAddr)
+			suite.Require().True(found)
+
```
mattverse, line 172: Can we add check for staking denom supply check? (to ensure that appropriate amount of stake denom have been minted through this action)
Also can we add a check to check the increase of liquiidty in the pool?

x/superfluid/keeper/concentrated_liquidity_test.go
```diff
@@ -0,0 +1,247 @@
...
+			// Check if synthetic lockup for the new lock ID is created.
+			_, err = lockupKeeper.GetSyntheticLockup(ctx, newLockId, keeper.StakingSyntheticDenom(clPoolDenom, valAddr.String()))
+			suite.Require().NoError(err)
+
+			// Check if the old intermediary account has no delegation.
+			_, found := stakingKeeper.GetDelegation(ctx, oldIntermediaryAcc, valAddr)
+			suite.Require().False(found)
+
+			// Check if the new intermediary account has delegation.
+			_, found = stakingKeeper.GetDelegation(ctx, newIntermediaryAcc, valAddr)
```
mattverse, line 170: Can we add delegation amount check here?

x/superfluid/keeper/concentrated_liquidity_test.go
```diff
@@ -0,0 +1,247 @@
...
+			amount0Added:           sdk.NewInt(100000000),
+			amount1Added:           sdk.NewInt(100000000),
+			expectedError:          types.LockImproperStateError{LockId: 1, UnbondingDuration: suite.App.StakingKeeper.UnbondingTime(suite.Ctx).String()},
+		},
+		"error: non-existent position ID": {
+			overwritePositionId: true,
+			amount0Added:        sdk.NewInt(100000000),
+			amount1Added:        sdk.NewInt(100000000),
+			expectedError:       cltypes.PositionIdNotFoundError{PositionId: 5},
+		},
```
mattverse, line 80: Can we add test case for 
- Amount0Added negative
- Amount1Added negative?

x/superfluid/keeper/concentrated_liquidity.go
```diff
@@ -0,0 +1,106 @@
...
+// Returns:
+// newPositionId: ID of the newly created concentrated liquidity position.
+// actualAmount0: Actual amount of token 0 added.
+// actualAmount1: Actual amount of token 1 added.
+// newLiquidity: The new liquidity value.
+// newLockId: ID of the lock associated with the new position.
+// error: Any error that may occur during execution.
+func (k Keeper) addToConcentratedLiquiditySuperfluidPosition(ctx sdk.Context, owner sdk.AccAddress, positionId uint64, amount0Added, amount1Added sdk.Int) (uint64, sdk.Int, sdk.Int, sdk.Dec, uint64, error) {
+	position, err := k.clk.GetPosition(ctx, positionId)
+	if err != nil {
```
mattverse, line 29: Do we not need logic to verify that given owner is the owner of the position?

mattverse (commented):
main logic looking good! Left some comments, please take a look

x/superfluid/keeper/concentrated_liquidity.go
p0mvn, line 13: nit: listing what the function does is a leaky abstraction. A function or a method should be treated as a black box without leaking the implementation details. This only adds unnecessary complexity when browsing through the code. If developers need to understand the implementation, they can read the code and the comments inside the methods.

Listing out the specific steps in the spec tend to often go out of sync since not everyone is going to update them, only adding confusion and complexity in the future.

On the other hand, the spec should concisely summarize what a function/method output given the inputs without leaking the implementation details

x/superfluid/keeper/concentrated_liquidity.go
```diff
@@ -0,0 +1,106 @@
...
+// 6. Locks the new position and superfluid delegates it.
+// It returns the new position ID, actual added amounts of both tokens, new liquidity, new lock ID, and any potential errors.
+//
+// Returns:
+// newPositionId: ID of the newly created concentrated liquidity position.
+// actualAmount0: Actual amount of token 0 added.
+// actualAmount1: Actual amount of token 1 added.
+// newLiquidity: The new liquidity value.
+// newLockId: ID of the lock associated with the new position.
+// error: Any error that may occur during execution.
```
p0mvn, line 26: nit: consider specifying in which cases errors are returned

x/superfluid/keeper/concentrated_liquidity.go
p0mvn: This happens in `withdrawPositon` I'm guessing.

Agreed on making this explicit here.

x/superfluid/keeper/concentrated_liquidity_test.go
```diff
@@ -0,0 +1,247 @@
...
+		isLastPositionInPool   bool
+		expectedError          error
+	}
+	testCases := map[string]sendTest{
+		"add to position that is superfluid delegated, not unlocking": {
+			superfluidDelegated: true,
+			amount0Added:        sdk.NewInt(100000000),
+			amount1Added:        sdk.NewInt(100000000),
+		},
+		"add to position that is superfluid delegated, not unlocking, not enough funds to add": {
```
p0mvn, line 35: Please add a test cases where non-owner executes this

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,30 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
...
+	} else if err != nil {
+		return false, 0, err
+	}
+
+	// Check if the underlying lock is mature.
+	lockIsMature, err := k.isLockMature(ctx, lockId)
+	if err != nil {
+		return true, 0, err
+	}
+	if lockIsMature {
```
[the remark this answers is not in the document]
czarcas7ic: I do not see a reason why we would want to retain the lock to position ID state entry, as in an ideal world it would immediately be removed via hook once it was mature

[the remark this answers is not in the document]
czarcas7ic: (the only thing stopping me from implementing this via hook is import cycle between CL and lockup)

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,30 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
 	store.Delete(lockIdKey)
 }
 
-// positionHasUnderlyingLockInState checks if a given positionId has a corresponding lock in state.
-func (k Keeper) positionHasUnderlyingLockInState(ctx sdk.Context, positionId uint64) (bool, error) {
+// PositionHasActiveUnderlyingLockInState checks if a given positionId has a corresponding lock in state.
+// If it has a lock in state, checks if that lock is still active.
+// If lock is still active, returns true.
+// If lock is no longer active, removes the lock ID from the position ID to lock ID mapping and returns false.
+func (k Keeper) PositionHasActiveUnderlyingLockInState(ctx sdk.Context, positionId uint64) (bool, uint64, error) {
```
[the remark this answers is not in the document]
czarcas7ic: Renamed and added return value names here https://github.com/osmosis-labs/osmosis/pull/5042/commits/20543994b6e818f9a07f4c9207b1e562e0c9db17

x/superfluid/keeper/concentrated_liquidity.go
czarcas7ic: Changed here https://github.com/osmosis-labs/osmosis/pull/5042/commits/9bbe6f3d9da1295a301161250fb25118753146cc

x/superfluid/keeper/concentrated_liquidity_test.go
czarcas7ic: non owner test case added https://github.com/osmosis-labs/osmosis/pull/5042/commits/f2e4b9501ddebe8cfb314a01dddea47cbe8c513e

x/superfluid/keeper/concentrated_liquidity_test.go
czarcas7ic: Added both denom supply check and pool address funds increase check https://github.com/osmosis-labs/osmosis/pull/5042/commits/31cbf91f44678a9c4a90dcc2073aa3f2a24dfc1c

x/superfluid/keeper/concentrated_liquidity_test.go
czarcas7ic: added neg amt0 and amt1 tests here https://github.com/osmosis-labs/osmosis/pull/5042/commits/8540cf1b30394a11e5b9fab2f17609cda60b29a8

x/superfluid/keeper/concentrated_liquidity.go
czarcas7ic: Agreed, updated godoc https://github.com/osmosis-labs/osmosis/pull/5042/commits/b30706775d793f78f9472c520f5b277c9658a5c5

x/superfluid/keeper/concentrated_liquidity.go
czarcas7ic: Added specific error cases in godoc https://github.com/osmosis-labs/osmosis/pull/5042/commits/b30706775d793f78f9472c520f5b277c9658a5c5

x/superfluid/keeper/concentrated_liquidity_test.go
czarcas7ic: Added delegation amount check https://github.com/osmosis-labs/osmosis/pull/5042/commits/a9d4886bee0a5a8fa1ca8e790396055d55e6858b

x/superfluid/keeper/concentrated_liquidity.go
czarcas7ic: Yeah, this happens in withdrawPosition, but I think I am going to add a check higher in the stack to terminate earlier (as well as def in depth). Thanks for pointing this out

czarcas7ic: Added owner check higher in stack and changed respective gotest to check for this new error https://github.com/osmosis-labs/osmosis/pull/5042/commits/8be61905e567d5051e17ebf68fc8297066ce3ffd

x/concentrated-liquidity/position.go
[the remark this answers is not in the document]
czarcas7ic: IMO I think, since adding this to the lockup mature hook is not viable, this is the next best option (to call this every time we check a positions underlying lock). Please lmk what you think @mattverse @p0mvn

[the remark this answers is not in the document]
AlpinYukseloglu: The import cycle issue seems solvable with a minor package refactor, but it seems like triggering via hook automatically could expose a spam vector.

Lazily clearing these during checks (or some other regular user interaction) seems reasonable to me.

x/concentrated-liquidity/lp.go
```diff
@@ -141,28 +141,13 @@ func (k Keeper) withdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
...
-		if lockIsMature {
-			// Remove the link between the position and the underlying lock since the lock is mature.
-			k.RemovePositionIdToLock(ctx, positionId, lockId)
-		} else {
-			// Lock is not mature, return error.
-			return sdk.Int{}, sdk.Int{}, types.LockNotMatureError{PositionId: position.PositionId, LockId: lockId}
-		}
+	if positionHasActiveUnderlyingLock {
+		// Lock is not mature, return error.
+		return sdk.Int{}, sdk.Int{}, types.LockNotMatureError{PositionId: position.PositionId, LockId: lockId}
```
AlpinYukseloglu, line 150: Can you clarify what the purpose of this change was? It seems as though the new code does not call `RemovePositionIdToLock` for mature locks – I'm assuming this is intended?

czarcas7ic: This is because I refactored PositionHasActiveUnderlyingLock to lazily update whenever fetching the underlying lock

x/concentrated-liquidity/position.go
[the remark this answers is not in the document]
mattverse: > , but it seems like triggering via hook automatically could expose a spam vector.

I think I am missing something, how can using hooks expose use to spam vectors?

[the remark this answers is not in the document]
czarcas7ic: @mattverse the problem with this though, is `RemovePositionIdToLockIfMature` would / should always get called with `PositionHasActiveUnderlyingLock`, and if future calls did not remember to call the `remove` portion then we would have unexpected behavior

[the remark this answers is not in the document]
mattverse: @czarcas7ic in that case, can we not make it the opposite way, where we make a method that encapsulates the logic for `removePosition`, lets say `RemovePositionIdToLockIfMature` and then within that function we can call `PositionHasActiveUnderlyingLock`. 

Apologies if I am being picky about method names here, I just think that method names that performs state changes should be explicit in the method name or else would also expose us to unexpected behaviors or bugs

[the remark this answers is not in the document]
p0mvn: I'm supportive of having `RemovePositionIdToLockIfMature` as a separate method and either calling `RemovePositionIdToLockIfMature` and `PositionHasActiveUnderlyingLock` from another method that encapsulates them or simply calling them together.

The fact that for this use case we need to perform the 2 actions, doesn't justify encapsulating them under a method called `PositionHasActiveUnderlyingLock`. From the name, I would never be able to tell that it is a mutative method

p0mvn (commented):
I

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,54 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
...
+	} else if err != nil {
+		return false, 0, err
+	}
+
+	// Check if the underlying lock is mature.
+	lockIsMature, err := k.isLockMature(ctx, lockId)
+	if err != nil {
+		return true, 0, err
+	}
+	if lockIsMature {
```
p0mvn, line 655: Can we use ` PositionHasActiveUnderlyingLock` here?

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,54 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
 	store.Delete(lockIdKey)
 }
 
-// positionHasUnderlyingLockInState checks if a given positionId has a corresponding lock in state.
-func (k Keeper) positionHasUnderlyingLockInState(ctx sdk.Context, positionId uint64) (bool, error) {
+// PositionHasActiveUnderlyingLock is a non mutative method that checks if a given positionId has a corresponding lock in state.
+// If it has a lock in state, checks if that lock is still active.
+// If lock is still active, returns true.
+// If lock is no longer active, returns false.
+func (k Keeper) PositionHasActiveUnderlyingLock(ctx sdk.Context, positionId uint64) (hasActiveUnderlyingLock bool, lockId uint64, err error) {
```
p0mvn, line 617: nit: I think non-mutative version doesn't have to be exported

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,54 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
...
+	if errors.Is(err, types.PositionIdToLockNotFoundError{PositionId: positionId}) {
+		return false, 0, nil
+	} else if err != nil {
+		return false, 0, err
+	}
+
+	// Check if the underlying lock is mature.
+	lockIsMature, err := k.isLockMature(ctx, lockId)
+	if err != nil {
+		return true, 0, err
```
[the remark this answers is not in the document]
czarcas7ic: My thought process was, if we know the position at one point had a lock, and we go to query if the lock is mature and we get an error, its best to go with the more restrictive assumption (the lock is still active) as a defense in depth in case, for whatever reason, we dont stop at the error.

x/concentrated-liquidity/position.go
czarcas7ic: I think you meant the opposite, the mutative version does not need to be exported. Done here https://github.com/osmosis-labs/osmosis/pull/5042/commits/e3212b2d10f08587f1c9ea958875b641c4363427

x/concentrated-liquidity/position.go
czarcas7ic: As it currently is structured, no, because `PositionHasActiveUnderlyingLock` returns false, 0, nil when both the lockIsMature as well as if the PositionIdToLockNotFoundError is present. So there is no way to know what the previous lockId was.

x/concentrated-liquidity/position_test.go
```diff
@@ -1167,9 +1169,41 @@ func (s *KeeperTestSuite) TestPositionToLockCRUD() {
...
+	// Since the above method is mutative, we should no longer be able to retrieve the lockId from the positionId
+	retrievedLockId, err = s.App.ConcentratedLiquidityKeeper.GetLockIdFromPositionId(s.Ctx, positionId)
+	s.Require().Error(err)
+	s.Require().Equal(uint64(0), retrievedLockId)
+
+	// Set the position to have the lockId again
+	s.App.ConcentratedLiquidityKeeper.SetPositionIdToLock(s.Ctx, positionId, concentratedLockId)
+	retrievedLockId, err = s.App.ConcentratedLiquidityKeeper.GetLockIdFromPositionId(s.Ctx, positionId)
+	s.Require().NoError(err)
+	s.Require().Equal(concentratedLockId, retrievedLockId)
```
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/5042/commits/7f6fbe1592c4d280b99a2d51662ba82ace268459

x/concentrated-liquidity/position.go
p0mvn: I don't follow

When lockIsMature as well as if the PositionIdToLockNotFoundError, there is no lock exists that we should be removing because 0 signifies the lack of lock.

What lock id and why do we need to know? What am I missing?

czarcas7ic:
```
func (k Keeper) positionHasActiveUnderlyingLockAndUpdate(ctx sdk.Context, positionId uint64) (hasActiveUnderlyingLock bool, lockId uint64, err error) {
	// Get the lock ID for the position.
	hasActiveUnderlyingLock, lockId, err = k.PositionHasActiveUnderlyingLock(ctx, positionId)
	// There is currently no way to determine if lock `hasInactiveUnderlyingLock`
	// The case we want to handle is when hasActiveUnderlyingLock is false
	// We would ne to run:
	// if !hasActiveUnderlyingLock, re-retrieve the GetLockIdFromPositionId, and then remove the connection here
	if lockIsMature {
		k.RemovePositionIdToLock(ctx, positionId, lockId)
		return false, 0, nil
	}
	return true, lockId, nil
}
```

x/concentrated-liquidity/position.go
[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/5042/commits/df780d073fe10d10d7aab35923309b92617c6327

x/concentrated-liquidity/position.go
p0mvn: Yeah the approach you mentioned seems preferable to me because it reduces complex edge case handling, reducing code duplication and only requiring one additional call to get the lock id if `hasActiveUnderlyingLock`.

```
 if !hasActiveUnderlyingLock, re-retrieve the GetLockIdFromPositionId, and then remove the connection here"
```
Re-fetching seems fine. Please consider this change

x/concentrated-liquidity/position.go
```diff
@@ -621,17 +610,54 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
...
-		return false, nil
+		return false, 0, nil
+	} else if err != nil {
+		return false, 0, err
+	}
+
+	// Check if the underlying lock is mature.
+	lockIsMature, err := k.isLockMature(ctx, lockId)
+	if err != nil {
+		return true, 0, err
```
p0mvn, line 653: I still don't understand the `true` here. Especially given the fact that it is false in a similar branch of `PositionHasActiveUnderlyingLock`.

The complexity and similarity of both `positionHasActiveUnderlyingLockAndUpdate` and `PositionHasActiveUnderlyingLock` is the primary reason why I am also suggesting to call `PositionHasActiveUnderlyingLock` from  `positionHasActiveUnderlyingLockAndUpdate`

x/concentrated-liquidity/position_test.go
```diff
@@ -1113,82 +1113,175 @@ func (s *KeeperTestSuite) TestMintSharesLockAndUpdate() {
...
+		createPosition                     func(s *KeeperTestSuite) (uint64, uint64)
+		expectedHasActiveLock              bool
+		expectedLockError                  bool
+		expectedPositionLockID             uint64
+		expectedPositionLockIDErr          bool
+		expectedGetPositionLockIdErr       bool
+		expectedUpdateHasActiveLock        bool
+		expectedUpdateLockID               uint64
+		expectedUpdateLockIDErr            bool
+		expectedUpdateGetPositionLockIdErr bool
```
p0mvn, line 1132: I'm really unsure how to reason about all the edge cases when there are so many conditions and so many things being tested.

I gave this my best shot but there are too many things going on to be able to follow the logic IMO. I would like to request separating these as much as possible, please. Ideally, each method should have an individual unit test testing all of its edge cases.

p0mvn (commented):
What I'm mostly missing before approving this PR is being able to confirm that edge cases of each method are well-tested and be able to reason about them. Once my last comment about the test is addressed, this should be good to go

x/concentrated-liquidity/position.go
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/5042/commits/5821706c879fb2f2ad772a4b4da0c8333dabc45f

x/concentrated-liquidity/position.go
czarcas7ic: Changed https://github.com/osmosis-labs/osmosis/pull/5042/commits/5821706c879fb2f2ad772a4b4da0c8333dabc45f

x/concentrated-liquidity/position_test.go
czarcas7ic: Split up tests! https://github.com/osmosis-labs/osmosis/pull/5042/commits/6483d45cf7fab6975693c851ebe2af84f59b1a37

x/concentrated-liquidity/position.go
```diff
@@ -630,17 +619,46 @@ func (k Keeper) RemovePositionIdToLock(ctx sdk.Context, positionId, underlyingLo
...
+// positionHasActiveUnderlyingLockAndUpdate is a mutative method that checks if a given positionId has a corresponding lock in state.
+// If it has a lock in state, checks if that lock is still active.
+// If lock is still active, returns true.
+// If lock is no longer active, removes the lock ID from the position ID to lock ID mapping and returns false.
+func (k Keeper) positionHasActiveUnderlyingLockAndUpdate(ctx sdk.Context, positionId uint64) (hasActiveUnderlyingLock bool, lockId uint64, err error) {
+	hasActiveUnderlyingLock, lockId, err = k.PositionHasActiveUnderlyingLock(ctx, positionId)
+	if err != nil {
+		return false, 0, err
+	}
+	if !hasActiveUnderlyingLock && lockId != 0 {
```
p0mvn, line 655: Does it make sense to do validation and erorring for other combinations of these conditions?

E.g. `hasActiveUnderlyingLock` but `lockId == 0` should error?

p0mvn (approved):
Great work on refactoring and tests. Super clear now

x/concentrated-liquidity/position.go
czarcas7ic: Yeah, this should be impossible, but a defense in depth never hurts. Will add

PR #5045, opened 2023-05-02
Protorev: Use hooks instead of message parsing to trigger backruns

NotJeremyLiu:

- We've concluded a hook-based approach to protorev will be the best approach to keep protorev up to date given several new message types being introduced that allow for swaps.
- Previously, protorev would case on specific message types for a transaction, and extract the pools that were swapped against to check for backrun potential.
- This PR changes this approach to now plug protorev into the GAMM and Concentrated liquidity hooks, store the necessary swap information in the hook, and then read from that store in the protorev posthandler.
- We also add logic to support join/exit pool hooks for the GAMM module (treating single sided joins/exits as swaps), and pool creation (updating highest liquidity pools at creation of a new pool)

## Brief Changelog

- Add new kvstore for SwapsToBackrun and add setters and getters
- Add hooks.go file that plugs into GAMM and ConcratedLiquidity pool hooks and stores/updates necessary protorev kvstores as appropriate based on the hook
  - Add SwapsToBackrun when triggered by a swap hook or join/exit pool hook that also swaps
  - Update highest liquidity pools at pool creation for GAMM and at initial liquidity position for CL (triggered by hook)
- Add SwapsToBackrun delete logic in postahandler without using gas

## Testing and Verifying

- New tests added for all protorev hook logic in hooks_test.go
- Updated previous tests to use updated methods and no longer rely on messages as input
- Updated epoch_hook_test.go to account for new concentrated liquidity pool testing changes that sets default coins when creating a pool with a position range

- Added changelog entry


stackman27 (commented):
left some comments 🌮 

I noticed that there isn't much information about backrunning in protorev.md. Is it possible to add some more docs about it so that it helps in the review?

stackman27 (commented):
left some comments 🌮 

I noticed that there isn't much information about backrunning in protorev.md. Is it possible to add some more docs about it so that it helps in the review?

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+// storeSwap stores a swap to be checked by protorev when attempting backruns.
+func (k Keeper) storeSwap(ctx sdk.Context, poolId uint64, tokenIn, tokenOut string) {
+	swapToBackrun := types.Trade{
+		Pool:     poolId,
+		TokenIn:  tokenIn,
+		TokenOut: tokenOut,
+	}
+
+	if err := k.AddSwapsToSwapsToBackrun(ctx, []types.Trade{swapToBackrun}); err != nil {
+		ctx.Logger().Error("Protorev error adding swap to backrun from storeSwap", err)
```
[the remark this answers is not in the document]
NotJeremyLiu: All the hook implementation I saw for GAMM and CL pools (twap and pool-incentives) don't return anything, with twap using ctx.Logger for unexpected behavior: https://github.com/osmosis-labs/osmosis/blob/025e4c80b48635f9696e74be52494f2b767cb55d/x/twap/listeners.go#L27, so I just followed that pattern here as well.

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+		return
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			ctx.Logger().Error("Protorev error recovering from panic in AfterCFMMPoolCreated hook, likely an overflow error", r)
+		}
+	}()
+
+	liquidity := baseToken.Amount.Mul(otherToken.Amount)
```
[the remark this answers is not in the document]
NotJeremyLiu: GetTotalPoolLiquidity returns a vector of Coins, not a number. So some transformation is necessary to get it to a comparable number, which is the multiplication here.

See module implementations:
GAMM: https://github.com/osmosis-labs/osmosis/blob/025e4c80b48635f9696e74be52494f2b767cb55d/x/gamm/keeper/pool.go#L293
CL: https://github.com/osmosis-labs/osmosis/blob/025e4c80b48635f9696e74be52494f2b767cb55d/x/concentrated-liquidity/pool.go#L173

NotJeremyLiu:
> left some comments 🌮
> 
> I noticed that there isn't much information about backrunning in protorev.md. Is it possible to add some more docs about it so that it helps in the review?

Can def add more, don't think it should be a blocker on this PR though.

Is there anything specific on your hand that would be helpful, the section that covers what cyclic arb is is at the top, then the rest is just really tech details so I agree more can be done: https://github.com/osmosis-labs/osmosis/blob/main/x/protorev/protorev.md#cyclic-arbitrage

x/protorev/keeper/epoch_hook_test.go
```diff
@@ -162,7 +162,7 @@ func (s *KeeperTestSuite) TestUpdateHighestLiquidityPools() {
 			},
 			expectedBaseDenomPools: map[string]map[string]keeper.LiquidityPoolStruct{
 				"epochTwo": {
-					"uosmo": {Liquidity: sdk.NewInt(2000000), PoolId: 49},
+					"uosmo": {Liquidity: sdk.Int(sdk.NewUintFromString("999999999000000001000000000000000000")), PoolId: 49},
```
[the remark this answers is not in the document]
NotJeremyLiu: Yes, but the Uint implementation is actually a big.Int lol: https://github.com/osmosis-labs/cosmos-sdk/blob/fbf8ba8a41fe455ebea7785e9e4fbac750b8dc3b/types/uint.go#L12

The reason this change happened is because I use the test suite way to create a new CL pool based on another Osmosis team member recommendation, and then the liquidity levels created with that function changed to be large amounts, thus making the liquidity Mul this large: https://github.com/osmosis-labs/osmosis/blob/6c744ac96782041fdcc08f4993c53494d8a5fa38/x/protorev/keeper/keeper_test.go#L894

NotJeremyLiu:
After syncing up to main a test is failing, will deal with when I get into office.

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	if coins == nil {
+		if pool.GetType() == poolmanagertypes.Concentrated {
+			clPool, ok := pool.(cltypes.ConcentratedPoolExtension)
+			if !ok {
+				ctx.Logger().Error("Protorev error casting pool to concentrated pool in AfterCFMMPoolCreated hook")
+				return sdk.Coins{}
+			}
+			coins = sdk.Coins{sdk.NewCoin(clPool.GetToken0(), sdk.NewInt(0)), sdk.NewCoin(clPool.GetToken1(), sdk.NewInt(0))}
```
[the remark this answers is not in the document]
NotJeremyLiu: So in hindsight this is actually purposeful, since NewCoins() sanitizes coins with a 0 value, so ends up returning a blank coins object.

Reverted here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/6faeab31fe71d17a5963497542c0532f174db2fb

This change what caused the test to fail

NotJeremyLiu:
> After syncing up to main a test is failing, will deal with when I get into office.

See this comment for an explanation / resolution of failing test: https://github.com/osmosis-labs/osmosis/pull/5045#discussion_r1210514211

x/protorev/keeper/epoch_hook_test.go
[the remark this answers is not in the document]
stackman27: how about just sdk.NewIntFromString then? 

i believe it uses bigInt too

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
stackman27: i see gotchu

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
stackman27: oh nono i was just wondering if we can halt the function like this: https://github.com/osmosis-labs/osmosis/pull/5045/files#diff-3618ca814ac5b1546331b2f43868d3bc48d4b897033245b2bebd4891d610997bR113

x/protorev/keeper/epoch_hook_test.go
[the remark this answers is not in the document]
NotJeremyLiu: Just checked, NewIntFromString returns an sdk.Int and a bool (instead of just sdk.Uint value which can be converted inline), so wouldn't cleanly fit into the params for the test.

If no strong reason to convert, I think we should just keep as is.

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
stackman27: hmm, i see so this take into account that in CL liquidity = sum of the product of position liquidity?

x/protorev/keeper/epoch_hook_test.go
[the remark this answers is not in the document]
stackman27: sounds good

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
stackman27: i see, what do you think of adding an inline comment of why we dont return there?

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
NotJeremyLiu: The CL GetTotalPoolLiquidity for CL will return the balances of the CL pool, not just the current position: https://github.com/osmosis-labs/osmosis/blob/fdcae6138b924b03752671f48c3815f8da5d3de6/x/concentrated-liquidity/pool.go#L179

So this ends up comparing raw balance to raw balance for pools.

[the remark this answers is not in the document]
stackman27: yea i think that should be good

stackman27 (approved):
this LGTM!  thanks for addressing all the feedbacks

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
p0mvn: I'm worried about bypassing this constructor validation of zero coins. This might lead to panic elsewhere as the system evolves.

Can we instead special handle CL pools in the caller?
- if CL pool and is active => tokenA := clPool.GetToken0()  tokenB := clPool.GetToken1()

x/protorev/keeper/protorev.go
```diff
@@ -146,6 +146,57 @@ func (k Keeper) DeleteAllPoolsForBaseDenom(ctx sdk.Context, baseDenom string) {
...
+// SetSwapsToBackrun sets the swaps to backrun, updated via hooks
+func (k Keeper) SetSwapsToBackrun(ctx sdk.Context, swapsToBackrun types.Route) error {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefixSwapsToBackrun)
+
+	bz, err := swapsToBackrun.Marshal()
+	if err != nil {
+		return err
+	}
+
+	store.Set(types.KeyPrefixSwapsToBackrun, bz)
```
p0mvn, line 158: nit: we also have`osmoutils.Set` and `osmoutils.MustSet`. Feel free to resolve, non-blocking

p0mvn (commented):
Would like to do a deeper dive in the next 2 days. Leaving some initial nits

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+		tokenA := coins[0]
+		tokenB := coins[1]
+
+		if baseDenomMap[tokenA.Denom] {
+			k.compareAndStorePool(ctx, poolId, tokenA, tokenB)
+		}
+		if baseDenomMap[tokenB.Denom] {
+			k.compareAndStorePool(ctx, poolId, tokenB, tokenA)
+		}
+	}
```
czarcas7ic, line 129: This feels a little odd to me. We shouldn't need to make an empty coins struct just to compare denoms. It feels like we should just be directly getting the denoms from the pool here and using those in the map instead of determining this from the TotalPoolLiquidity

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			tc.param.executeSwap()
+
+			routes, err := s.App.ProtoRevKeeper.GetSwapsToBackrun(s.Ctx)
+			s.Require().NoError(err)
+			s.Require().Equal(tc.param.expectedTrades, routes.Trades)
+
+			s.App.ProtoRevKeeper.DeleteSwapsToBackrun(s.Ctx)
```
czarcas7ic, line 134: you just just run s.SetupTest at the beginning of this instead of running DeleteSwapsToBackrun at the end

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+
+// Create new ProtoRev hooks.
+func (k Keeper) Hooks() Hooks { return Hooks{k} }
+
+// ----------------------------------------------------------------------------
+// GAMM HOOKS
+// ----------------------------------------------------------------------------
+
+// AfterCFMMPoolCreated hook checks and potentially stores the pool via the highest liquidity method.
+func (h Hooks) AfterCFMMPoolCreated(ctx sdk.Context, sender sdk.AccAddress, poolId uint64) {
```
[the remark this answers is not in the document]
NotJeremyLiu: I'm external to context around cosmwasmpool (release time, usage, etc.), but hooks would be great so that ProtoRev can backrun swaps from these pools.

Maybe can take the convo to slack / a call, but also would be interested if they interface with poolmanager / we can use them throughout the entire protorev system.

x/protorev/keeper/hooks.go
NotJeremyLiu: The reason I grab coins instead of denoms (prior to CL was when the design decision was made) is because we compare liquidity by multiplying amounts of the coin when we call k.compareAndStorePool: https://github.com/osmosis-labs/osmosis/blob/6d7d411780b88f2da10382376061f67618ee8f1b/x/protorev/keeper/hooks.go#L185

Can switch the design to first call get denoms here, and then later call get total liquidity within compareAndStorePool, but I think I am currently leaning towards @p0mvn's suggestion: https://github.com/osmosis-labs/osmosis/pull/5045#discussion_r1212486103

x/protorev/keeper/protorev.go
NotJeremyLiu: Noting this down for a future PR to switch all the setters/getters in protorev using these. Resolving

x/protorev/keeper/hooks_test.go
NotJeremyLiu: dope, fixed: https://github.com/osmosis-labs/osmosis/pull/5045/commits/66c1e8cce646be8440da2e82dd51697f5dfef392

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
p0mvn: Cosmwasm pools are in v16 scope but let's cut our protorev out of scope for cw pools.

We don't have a specific date set for the launch of the first cw pool anyways so we can add this if needed in v17. Lmk if any concerns on the suggestion 

The x/cosmwasmpool is plugged into the x/poolmanager similar to gamm and CL modules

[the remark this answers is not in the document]
NotJeremyLiu: Got it, I'll add a note on my end to track it for v17.

x/protorev/keeper/hooks.go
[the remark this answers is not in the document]
NotJeremyLiu: My take on resolving: https://github.com/osmosis-labs/osmosis/pull/5045#issuecomment-1572845639

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,235 @@
...
+		return sdk.Coins{}
+	}
+
+	if coins == nil {
+		if pool.GetType() == poolmanagertypes.Concentrated {
+			clPool, ok := pool.(cltypes.ConcentratedPoolExtension)
+			if !ok {
+				ctx.Logger().Error("Protorev error casting pool to concentrated pool in AfterCFMMPoolCreated hook")
+				return sdk.Coins{}
+			}
```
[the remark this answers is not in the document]
NotJeremyLiu: No longer use this logic with the new way of updating pool after the initial position is created, so resolving

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// compareAndStorePool compares the liquidity of the new pool with the liquidity of the stored pool, and stores the new pool if it has more liquidity.
+func (k Keeper) compareAndStorePool(ctx sdk.Context, poolId uint64, baseDenom, otherDenom string) {
+	storedPoolId, err := k.GetPoolForDenomPair(ctx, baseDenom, otherDenom)
+	if err != nil {
+		// Error means no pool exists for this pair, so we set it
+		k.SetPoolForDenomPair(ctx, baseDenom, otherDenom, poolId)
+		return
+	}
```
p0mvn, line 151: Why do we have `storedPoolId` and `poold` here? What if for the same denom pair we have multiple pools stored - CL and balancer but the hoos is executed for one but not the other?

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,215 @@
...
+	newPoolliquidity := newPoolCoins[0].Amount.Mul(newPoolCoins[1].Amount)
+
+	// Get coins and calculate liquidity for the stored pool
+	storedPool, err := k.gammKeeper.GetPoolAndPoke(ctx, storedPoolId)
+	if err != nil {
+		ctx.Logger().Error("Protorev error getting storedPool in AfterCFMMPoolCreated hook", err)
+		return
+	}
```
p0mvn, line 172: This only gets balancer pools, what about CL?

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,215 @@
...
+		TokenOut: tokenOut,
+	}
+
+	if err := k.AddSwapsToSwapsToBackrun(ctx, []types.Trade{swapToBackrun}); err != nil {
+		ctx.Logger().Error("Protorev error adding swap to backrun from storeSwap", err) // Does not return since logging is last thing in the function
+	}
+}
+
+// compareAndStorePool compares the liquidity of the new pool with the liquidity of the stored pool, and stores the new pool if it has more liquidity.
+func (k Keeper) compareAndStorePool(ctx sdk.Context, poolId uint64, baseDenom, otherDenom string) {
```
p0mvn, line 145: I would like to see a direct unit test for every function please with CL and balancer pools

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,315 @@
...
+					return clPool.GetId()
+				},
+			},
+			expectPass: true,
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			poolId := tc.param.executePoolCreation()
```
p0mvn, line 303: I would like to see a test where CL and gamm pools created have the same denoms.

1. Balancer is created first and CL second
2. CL is created first and balancer second

x/protorev/keeper/hooks.go
NotJeremyLiu: Nice catch, this needs to use poolManager instead

x/protorev/keeper/hooks.go
NotJeremyLiu: The protorev module currently only stores a single pool for a given denom pair (irrespective of type).

So when we error here that means no pool is stored for the pair (irrespective of type), and then we proceed to store the poolId of the pool that was just created and triggered the hook

x/protorev/keeper/hooks.go
NotJeremyLiu: Addressed here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/91f16f0e0bc60638d002c21a7a2224e8fdd5bf72

x/protorev/keeper/hooks_test.go
NotJeremyLiu: Makes sense, added here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/c24c3b98f9c987f36a539713bc196d09cc829b52

davidterpay (approved):
this lgtm. excited to see how protorev improves with the new hook based approach!

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,222 @@
...
+
+	denoms, err := k.poolmanagerKeeper.RouteGetPoolDenoms(ctx, poolId)
+	if err != nil {
+		ctx.Logger().Error("Protorev error getting pool liquidity in afterPoolCreated", err)
+		return
+	}
+
+	// Pool must be active and the number of denoms must be 2
+	if pool.IsActive(ctx) && len(denoms) == 2 {
+		if baseDenomMap[denoms[0]] {
```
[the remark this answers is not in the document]
NotJeremyLiu: added: https://github.com/osmosis-labs/osmosis/pull/5045/commits/817fb285d8509cc1126af7a73635bd264b6ea69b

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,222 @@
...
+			k.compareAndStorePool(ctx, poolId, denoms[0], denoms[1])
+		}
+		if baseDenomMap[denoms[1]] {
```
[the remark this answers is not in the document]
NotJeremyLiu: added: https://github.com/osmosis-labs/osmosis/pull/5045/commits/817fb285d8509cc1126af7a73635bd264b6ea69b

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,315 @@
...
+func (s *KeeperTestSuite) TestSwapping() {
+	type param struct {
+		expectedTrades []types.Trade
+		executeSwap    func()
+	}
+
+	tests := []struct {
+		name       string
+		param      param
+		expectPass bool
```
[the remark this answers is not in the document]
NotJeremyLiu: Using it now: https://github.com/osmosis-labs/osmosis/pull/5045/commits/7a69efa58a70e15b29e5d4fecbbbcfc34ed4eda5 (keeping it in in case of use later)

x/protorev/keeper/hooks.go
NotJeremyLiu: Direct unit tests are added for every helper function used in hooks.go, except for AfterPoolCreatedWithCoins which is directly called from the respective hooks, and there are already tests for the used hooks themselves

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+		// Error means no pool exists for this pair, so we set it
+		k.SetPoolForDenomPair(ctx, baseDenom, otherDenom, poolId)
+		return
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			ctx.Logger().Error("Protorev error recovering from panic in AfterCFMMPoolCreated hook, likely an overflow error", r)
+		}
+	}()
```
p0mvn, line 191: Could you please elaborate on when and where can this panic? Not immediately clear to me

p0mvn: Would be great to have a test covering this as well to show how this can happen

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,775 @@
...
+			expectPass: true,
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			s.SetupTest()
+
+			// All pools are already created in the setup
+			s.App.ProtoRevKeeper.StoreJoinExitPoolSwaps(s.Ctx, s.TestAccs[0], tc.param.poolId, tc.param.denom, tc.param.isJoin)
```
p0mvn, line 599: can we have an error case for this where it fails because pool id is not a gamm pool?

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,775 @@
...
+		s.Run(tc.name, func() {
+			tc.param.executeLiquidityProviding()
+
+			routes, err := s.App.ProtoRevKeeper.GetSwapsToBackrun(s.Ctx)
+			if tc.expectPass {
+				s.Require().NoError(err)
+				s.Require().Equal(tc.param.expectedTrades, routes.Trades)
+			}
+
+			s.App.ProtoRevKeeper.DeleteSwapsToBackrun(s.Ctx)
```
p0mvn, line 243: nit: what is the reason for manually deleting this? We could also just call `s.Setup() at the start of the test case?

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,775 @@
...
+
+			routes, err := s.App.ProtoRevKeeper.GetSwapsToBackrun(s.Ctx)
+			s.Require().NoError(err)
+			s.Require().Equal(tc.param.expectedTrades, routes.Trades)
+		})
+	}
+}
+
+// Tests the hook implementation that is called after liquidity providing
+func (s *KeeperTestSuite) TestLiquidityChanging() {
```
p0mvn, line 141: q: we don't have CL liquidity changing infra because it does not affect price so nothing to backrun. Is my understanding correct?

NotJeremyLiu: Yes, we only support backrunning GAMM exit/join because there are helper functions to make single-sided joins/exits, which swap in the background (and what we are really after is backrunning the swap).

To my knowledge, CL does not have a single-sided join/exit equivalent (that swaps -> provides liquidity), and thus I did not add any backrunning logic here.

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// Pool must be active and the number of denoms must be 2
+	if pool.IsActive(ctx) && len(denoms) == 2 {
+		if _, ok := baseDenomMap[denoms[0]]; ok {
+			k.CompareAndStorePool(ctx, poolId, denoms[0], denoms[1])
+		}
+		if _, ok := baseDenomMap[denoms[1]]; ok {
+			k.CompareAndStorePool(ctx, poolId, denoms[1], denoms[0])
+		}
```
p0mvn, line 174: Please add a comment explaining the logic here in-detail

x/protorev/keeper/posthandler.go
```diff
@@ -63,6 +61,9 @@ func (protoRevDec ProtoRevDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simu
 		ctx.Logger().Error("ProtoRevTrade failed with error", err)
 	}
 
+	// Delete swaps to backrun for next transaction without consuming gas
+	protoRevDec.ProtoRevKeeper.DeleteSwapsToBackrun(ctx.WithGasMeter(upperGasLimitMeter))
```
p0mvn, line 65: The gas is still consumed, correct? Did you mean that the limit is refreshed?

p0mvn (commented):
Great work on expanding the tests! Looks great. I will do a short final pass once all comments are addressed prior to merging

x/protorev/keeper/hooks_test.go
```diff
@@ -0,0 +1,775 @@
...
+			name: "Nothing Stored, Store Balancer",
+			param: param{
+				baseDenom:  "uosmo",
+				matchDenom: "juno",
+				prepareStateAndGetPoolIdToCompare: func() (uint64, uint64) {
+					// Prepare a balancer pool with coins
+					poolId := s.PrepareBalancerPoolWithCoins(sdk.NewCoin("uosmo", sdk.NewInt(10)), sdk.NewCoin("juno", sdk.NewInt(10)))
+
+					// Delete all pools for the base denom uosmo so that all tests start with a clean slate
+					s.App.ProtoRevKeeper.DeleteAllPoolsForBaseDenom(s.Ctx, "uosmo")
```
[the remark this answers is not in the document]
NotJeremyLiu: Added: https://github.com/osmosis-labs/osmosis/pull/5045/commits/e5d17c636c2f8a3442cff3fc18ebe56ad0415180

x/protorev/keeper/hooks.go
NotJeremyLiu: Added comment specifying it catches an overflow panic and moved the panic catching into GetComparablePoolLiquidity (where it should belong since moving the multiplication into that function) : https://github.com/osmosis-labs/osmosis/pull/5045/commits/22d528e23f5ff3733c0d7b4c83f03e6914ecf5c4

NotJeremyLiu: Added test here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/ed66bce110964f30b32d9d2a2c88922c522d5556

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+		if r := recover(); r != nil {
+			ctx.Logger().Error("Protorev error recovering from panic in AfterCFMMPoolCreated hook, likely an overflow error", r)
+		}
+	}()
+
+	// Get comparable liquidity for the new pool
+	newPoolLiquidity, err := k.GetComparablePoolLiquidity(ctx, poolId)
+	if err != nil {
+		ctx.Logger().Error("Protorev error getting newPoolLiquidity in compareAndStorePool", err)
+		return
```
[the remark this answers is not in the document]
NotJeremyLiu: Added here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/c18ec190e1d60747e6b89a0578a111471cf6e601

x/protorev/keeper/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// Get comparable liquidity for the stored pool
+	storedPoolLiquidity, err := k.GetComparablePoolLiquidity(ctx, storedPoolId)
+	if err != nil {
+		ctx.Logger().Error("Protorev error getting storedPoolLiquidity in compareAndStorePool", err)
+		return
+	}
```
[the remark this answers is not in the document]
NotJeremyLiu: Added here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/c18ec190e1d60747e6b89a0578a111471cf6e601

x/protorev/keeper/hooks_test.go
NotJeremyLiu: Added here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/c2db6d772fad7e4fe77770e757e5561d05ac9c75

x/protorev/keeper/hooks_test.go
NotJeremyLiu: It was done originally because the last two tests called the ExitSwapShareAmountIn methods that required the test acc to have the pool shares in their balance, which a previous iteration of the test gave them.

Instead, I have now just given the testAccs the necessarily pool shares balance and just call s.SetupTest(): https://github.com/osmosis-labs/osmosis/pull/5045/commits/460da4a42d4c3cf341890d0537982419dd6aee51

x/protorev/keeper/posthandler.go
NotJeremyLiu: The gas is not consumed from the transaction's gas meter, but the new gas meter passed in (which is only used temporarily during the function call), added clarifying comment here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/61d3474a6a489d1247c556950e28aaa46209ebb0

This was made this way (as opposed to using the gas meter in cacheCtx) because no matter what happens with the backrun itself (whether or not it errors), we want to clear out the swaps to start fresh.

x/protorev/keeper/hooks.go
NotJeremyLiu: Added here: https://github.com/osmosis-labs/osmosis/pull/5045/commits/8966b6a0e3bf2cc11ba5c13f91f8429a7e014625

x/protorev/keeper/hooks.go
p0mvn: The simulator sometimes fails due to incorrect setup or false positives. Usually, we inspect the logs, and if it is something unrelated, simply rerunning it helps.

p0mvn (approved):
Excellent work @NotJeremyLiu

PR #5072, opened 2023-05-03
Add async support to wasm hooks

nicolaslara:

Packet Froward middleware introduced "async" acknowledgements in IBC: when onRecvPacket runs, the ack doesn't need to be written to state right away. Instead, that function can return `nil` and someone else (ideally the same module that received the packet) will write the acknowledgement later. If this doesn't happen the packet will timeout. 

We want to add support for this in ibc wasm hooks so that whitelisted contracts (specifically XCS) can do this and behave similarly to PFM. This will provide a better UX when interacting with contracts that forward tokens.

## Brief Changelog

 * Added mechanism for contracts called via wasm hooks to emit acks asynchronously

## Testing and Verifying

 * Added test


x/ibc-hooks/wasm_hook.go
```diff
@@ -117,7 +111,21 @@ func (h WasmHooks) OnRecvPacketOverride(im IBCMiddleware, ctx sdk.Context, packe
...
 	}
 
-	fullAck := ContractAck{ContractResult: response.Data, IbcAck: ack.Acknowledgement()}
+	// Check if the contract is requesting for the ack to be async.
+	var asyncAckRequest types.OnRecvPacketAsyncAckResponse
+	err = json.Unmarshal(response.Data, &asyncAckRequest)
+	if err == nil {
+		// If unmarshalling succeeds, the contract is requesting for the ack to be async.
+		if asyncAckRequest.IsAsyncAck { // in which case IsAsyncAck is expected to be set to true
+			// TODO: Do we want to check that only whitelisted contracts can do AsyncAcks?
```
nicolaslara, line 120: @ValarDragon also here. Any thoughts?

cosmwasm/contracts/crosschain-swaps/src/execute.rs
```diff
@@ -6,7 +6,7 @@ use swaprouter::msg::ExecuteMsg as SwapRouterExecute;
 
 use crate::checks::{check_is_contract_governor, ensure_key_missing, validate_receiver};
 use crate::consts::{MsgReplyID, CALLBACK_KEY};
-use crate::msg::{CrosschainSwapResponse, FailedDeliveryAction};
+use crate::msg::{AsyncResponse, CrosschainSwapResponse, FailedDeliveryAction};
```
p0mvn, line 9: Is this used anywhere?

tests/ibc-hooks/testutils/contracts/echo/src/ibc.rs
```diff
@@ -0,0 +1,63 @@
...
+    revision_height: Option<u64>,
+}
+
+// #[cw_serde]
+// pub struct FungibleTokenData {
+//     pub denom: String,
+//     amount: Uint256,
+//     sender: String,
+//     receiver: String,
+// }
```
p0mvn, line 20: nit: can probably be removed

x/ibc-hooks/types/msgs.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (m MsgEmitIBCAck) Route() string { return RouterKey }
+func (m MsgEmitIBCAck) Type() string  { return TypeMsgEmitIBCAck }
+func (m MsgEmitIBCAck) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Sender)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid sender address (%s)", err)
+	}
+	return nil
+}
```
p0mvn, line 23: Does it make sense to do additional validation of packet sequence and channel here?

x/ibc-hooks/wasm_hook.go
p0mvn: Asking to learn: why do we want to whitelist this functionality?

x/ibc-hooks/wasm_hook.go
```diff
@@ -117,7 +111,21 @@ func (h WasmHooks) OnRecvPacketOverride(im IBCMiddleware, ctx sdk.Context, packe
...
+			// Store the contract as the packet's ack actor and return nil
+			h.ibcHooksKeeper.StorePacketAckActor(ctx, packet.GetSourceChannel(), packet.GetSequence(), contractAddr.String())
+			return nil
+		}
+	}
```
p0mvn, line 125: Please consider having direct unit tests for this logic as quite a few edge cases are hapenning

x/ibc-hooks/wasm_hook.go
```diff
@@ -117,7 +111,21 @@ func (h WasmHooks) OnRecvPacketOverride(im IBCMiddleware, ctx sdk.Context, packe
 		return osmoutils.NewEmitErrorAcknowledgement(ctx, types.ErrWasmError, err.Error())
 	}
 
-	fullAck := ContractAck{ContractResult: response.Data, IbcAck: ack.Acknowledgement()}
+	// Check if the contract is requesting for the ack to be async.
+	var asyncAckRequest types.OnRecvPacketAsyncAckResponse
+	err = json.Unmarshal(response.Data, &asyncAckRequest)
+	if err == nil {
```
p0mvn, line 117: I'm guessing this design choice of returning nil was made because we want to maintain backward compatibility with other contracts that utilize a different response data structure. Is that correct?

x/ibc-hooks/README.md
```diff
@@ -180,6 +180,104 @@ pub enum SudoMsg {
 }
 ```
 
+### Async Acks
+
+IBC supports the ability to send an ack back to the sender of the packet asynchronously. This is useful for
+cases where the packet is received, but the ack is not immediately known. For example, if the packet is being
+forwarded to another chain, the ack may not be known until the packet is received on the other chain.
+
+#### Making contract Acks async
```
p0mvn, line 189: Could you please add a section similar to "Use case" under "Ack callbacks" and compare async vs non-sync? I think having that comparison on a real example would be useful

p0mvn (commented):
Only a high-level pass to acquire context with nits. Will do a more detailed pass in a subsequent review

x/ibc-hooks/wasm_hook.go
p0mvn: Discussed offline, whitelist makes sense to me

x/ibc-hooks/wasm_hook.go
p0mvn: Discussed offline that e2e test already sufficiently covers this. Leaving this for Nicolas to make final call on whether makes sense to add this

x/ibc-hooks/wasm_hook.go
p0mvn: Discussed offline - that seems to be correct and contract developers often return nil in their responses so we don't want to force them to return a specific response type

p0mvn (approved):
Great work! Was super interesting to learn about this.

I'm feeling pretty good about this change after stepping through the `TestWasmHooksAsyncAcks` test with a debugger.

However, I think it would be good to have another approval still

x/ibc-hooks/types/msgs.go
nicolaslara: I don't think we do. The packet sequence is an uint64, and the channel can be any string (even if they generally are `channel-<number>`

cosmwasm/contracts/crosschain-swaps/src/execute.rs
nicolaslara: it isn't removed

x/ibc-hooks/keeper/keeper.go
```diff
@@ -32,31 +42,130 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+	case "ack_response":
+		jsonAck, err := json.Marshal(ack.AckResponse)
+		if err != nil {
+			return nil, sdkerrors.Wrap(err, "could not marshal acknowledgement")
+		}
+		packet = ack.AckResponse.Packet
+		newAck = channeltypes.NewResultAcknowledgement(jsonAck)
+	case "ack_error":
+		packet = ack.AckError.Packet
+		newAck = osmoutils.NewEmitErrorAcknowledgement(ctx, types.ErrAckFromContract, ack.AckError.ContractError)
```
nicolaslara, line 156: Note to self: don't emit error acks but send result acks with information about being an error instead

This prevents the sender from releasing the escrowed tokens which could potentially cause a double spend

nicolaslara:
This is ready for final review and merge

ValarDragon (approved):
Great job! Feel terrible for how long this took me to review

Issue #5049, opened 2023-05-02
[CL]: incentives individual accum positions not exporting / importing


czarcas7ic:
## Background

[See the following commit](https://github.com/osmosis-labs/osmosis/pull/5024/commits/8e01a2634132b3105ad47b3d3188d82382d9a4c4). This fixed an issue where we did not import and export individual position's fee accum values. This exact fix will likely need to be done for the incentive accumulator positions as well.

CC @stackman27 since you are auditing that section of the code.

## Suggested Design

Look over above commit. Implement for incentives import export genesis logic.

## Acceptance Criteria

TBD

p0mvn:
Is this internal review related?

p0mvn:
Is this complete?

PR #5024, opened 2023-04-29
[CL Message Audit]: MsgCollectFees

czarcas7ic:
Clickup Checklist:
https://app.clickup.com/37420681/v/dc/13nzm9-19087/13nzm9-35247

## What is the purpose of the change

This PR includes changes made from the internal audit of MsgCollectFees.

## Brief Changelog

- exports `InitOrUpdatePosition` method in accum package (used in import genesis)
- adds `fee_accum_record` field to `PositionData`
  - added to fix bug where individual position's fee accums weren't getting imported or exported via genesis
  - expands init/export genesis tests to check for this new field
- drive by change that grabs largest position ID when using the localosmosis script
  - nextPositionId was not getting properly set without this change due to failed position messages skipping position Ids
- readme reformat
- readme additions specific to fee accum / collection
- adds `PrepareClaimableFees` test
- expands `TestQueryAndCollectFees`
- adds `IsPositionOwner` test

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  /  [Osmosis docs repo](https://github.com/osmosis-labs/docs)   /   not documented)

PR #5076, opened 2023-05-03
[CL][Internal Review][CollectIncentives] fix uptime Accumulator init genesis

stackman27:
Closes: [#5049](https://github.com/osmosis-labs/osmosis/issues/5049)

[See the following commit](https://github.com/osmosis-labs/osmosis/pull/5024/commits/8e01a2634132b3105ad47b3d3188d82382d9a4c4). This fixed an issue where we did not import and export individual position's fee accum values. This exact fix will likely need to be done for the incentive accumulator positions as well.

## Brief Changelog
n/a

## Testing and Verifying
added tests for initGenesis 


p0mvn:
Any updates here @stackman27 ? Is everything going alright? This change would be useful to have for some of the parallel work I'm doing in: https://github.com/osmosis-labs/osmosis/pull/5091

stackman27:
> Any updates here @stackman27 ? Is everything going alright? This change would be useful to have for some of the parallel work I'm doing in: #5091

hi @p0mvn working on test rn. Will try to have this r4r by tonight

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3464,6 +3463,18 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			initSenderBalances := s.App.BankKeeper.GetAllBalances(s.Ctx, defaultSender)
 			initPoolBalances := s.App.BankKeeper.GetAllBalances(s.Ctx, clPool.GetAddress())
 
+			uptimeAccums, err := clKeeper.GetUptimeAccumulators(s.Ctx, validPoolId)
+			s.Require().NoError(err)
+
+			var oldAccumRecord []accum.Record
+			for _, uptimeAccum := range uptimeAccums {
```
p0mvn, line 3470: Please check `uptimeAccums` are non-empty

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3534,6 +3545,25 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 				}
 				s.Require().Equal(expectedCoins, amountClaimed)
 				s.Require().Equal(sdk.Coins(nil), amountForfeited)
+
+				uptimeAccumsNew, err := clKeeper.GetUptimeAccumulators(s.Ctx, validPoolId)
+				s.Require().NoError(err)
+
+				var newAccumRecord []accum.Record
+				for _, uptimeAccum := range uptimeAccumsNew {
```
p0mvn, line 3553: ditto

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3534,6 +3545,25 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+					positionName := string(types.KeyPositionId(DefaultPositionId))
+					accRecord, err := uptimeAccum.GetPosition(positionName)
+					s.Require().NoError(err)
+
+					newAccumRecord = append(newAccumRecord, accRecord)
+				}
+				expectedUptimeAccValue := sdk.NewDecCoinsFromCoins(sdk.NewCoin("bar", sdk.NewInt(100)), sdk.NewCoin("foo", sdk.NewInt(100)))
```
p0mvn, line 3560: Where are these values coming from? Why is it 100 of each coin?

Let's avoid hardcoding, even the denoms

x/concentrated-liquidity/keeper_test.go
```diff
@@ -131,13 +131,13 @@ func (s *KeeperTestSuite) validatePositionUpdate(ctx sdk.Context, positionId uin
 
 // validateTickUpdates validates that ticks with the given parameters have expectedRemainingLiquidity left.
 func (s *KeeperTestSuite) validateTickUpdates(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, lowerTick int64, upperTick int64, expectedRemainingLiquidity sdk.Dec, expectedLowerFeeGrowthOutside, expectedUpperFeeGrowthOutside sdk.DecCoins) {
-	lowerTickInfo, err := s.App.ConcentratedLiquidityKeeper.GetTickInfo(s.Ctx, poolId, lowerTick)
+	lowerTickInfo, err := s.App.ConcentratedLiquidityKeeper.GetTickInfo(ctx, poolId, lowerTick)
```
p0mvn, line 134: Suggest removing context as a param and keeping `s.Ctx`. There is no reason to have a context param when there is one defined on the suite

p0mvn (commented):
Please consider splitting up changes into separate PRs for faster merge.

Also, updateAccumulatorToNow is an important change that deserves its own PR with a test to analyze it more in-depth

x/concentrated-liquidity/genesis.go
```diff
@@ -184,15 +196,35 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *genesis.GenesisState {
...
+		var uptimeAccumRecords []accum.Record
+
+		for _, uptimeAccum := range uptimeAccumObject {
+			uptimeAccumPositionRecord, err := uptimeAccum.GetPosition(uptimePositionKey)
+			if err != nil {
+				panic(err)
+			}
+
+			uptimeAccumRecords = append(uptimeAccumRecords, uptimeAccumPositionRecord)
+		}
```
p0mvn, line 221: Notice how here we are getting a position for every uptime.

Then in the import logic, for every exported position we initialize an uptime position which isn't correct.

I think the tests are passing because of being too simple where all records are the same. Please add test cases with different records, I think they might expose what I'm trying to point at

x/concentrated-liquidity/position.go
```diff
@@ -37,7 +37,6 @@ func (k Keeper) getOrInitPosition(
 // If a position is not present, it initializes the position with the provided liquidity delta.
 // If a position is present, it combines the existing liquidity in that position with the provided liquidity delta. It also
 // bumps up all uptime accumulators to current time, including the ones the new position isn't eligible for.
-// WARNING: this method may mutate the pool, make sure to refetch the pool after calling this method.
```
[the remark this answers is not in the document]
czarcas7ic: I personally like the warning there

czarcas7ic (commented):
I think Roman is correct with his Record matrix suggestion. I echo his suggestion of expanding the import/export tests to by slightly more complex to confirm. We can hop on a call tomorrow if you get stuck on this at all and we can work through it 👍

x/concentrated-liquidity/genesis.go
```diff
@@ -79,6 +79,18 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState genesis.GenesisState) {
...
+			panic(err)
+		}
+		uptimePositionKey := string(types.KeyPositionId(positionWrapper.Position.PositionId))
+
+		// for each position we initialize all uptime  accumulator
+		for _, uptimeRecord := range positionWrapper.UptimeAccumRecord {
+			for _, uptimeAccum := range uptimeAccumObject {
+				k.initOrUpdateAccumPosition(ctx, uptimeAccum, uptimeRecord.InitAccumValue, uptimePositionKey, uptimeRecord.NumShares, uptimeRecord.UnclaimedRewards, uptimeRecord.Options)
+			}
+		}
```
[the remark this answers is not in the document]
stackman27: > Each exported position record should be associated with one uptime only.

the way i thought of this was, each position can have multiple uptime records associated with it. So the outer loop checks for the position and the inner loop checks for all the uptime that could be associated with the position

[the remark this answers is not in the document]
p0mvn: Assume we have 1 position exported with 3 uptimes for simplicity (1ns, 1s and 1min)

Then, the `positionWrapper.UptimeAccumRecord` list here consists of 3 records, A (1ns), B(1s) and C (1min)

What ends up happening is that for all of A, B and C we initialize records with 1ns, 1s and 1min.

I believe that we end up having just 3 records C(1ns), C(1s) and C (1min) because the last outer loop iteration overwrites

Lmk if that makes sense

x/concentrated-liquidity/position.go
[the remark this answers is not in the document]
stackman27: this was a github automatically removing it while rebasing. I'll def add it back

x/concentrated-liquidity/genesis.go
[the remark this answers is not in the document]
stackman27: I thought we initialize 
A(1ns) A(1s) A(1min) B(1ns) B(1s) B(1min)  C(1ns) C(1s) C(1min) 

ooh so the inner loop should be outside? so that we initialize an accumPosition for every `uptimeRecord` for every `uptimeAccum`, am i correctly understanding this?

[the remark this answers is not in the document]
p0mvn: > I thought we initialize
A(1ns) A(1s) A(1min) B(1ns) B(1s) B(1min) C(1ns) C(1s) C(1min)

This would be the case if we had 3 positions.

In my example, I had 1 position with 3 records exported corresponding to every uptime A (1ns), B (1s) and C(1min)

> ooh so the inner loop should be outside?

Let's take a step back from the inner vs outer loop.

When exporting genesis, we can export a matrix `recordsForEachUptime` [][]Record where every element is a slice of all records for a supported uptime at index `i`.

That way, during import, when iterating over `recordsForEachUptime` we can attach all records to a specific uptime accumulator that we know those positions belong to.

In the current approach, we lose the context of which uptime accumulator a position belongs to by exporting it into a slice from the start

x/concentrated-liquidity/genesis.go
```diff
@@ -184,16 +195,36 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *genesis.GenesisState {
...
+
+			positionData = append(positionData, genesis.PositionData{
+				LockId:            lockId,
+				Position:          &position,
+				FeeAccumRecord:    feeAccumPositionRecord,
+				UptimeAccumRecord: uptimeAccumRecords[uptimeIndex],
+			})
+		}
+
 	}
```
p0mvn, line 228: This duplicates positions. We need to have one `PositionData` structure object associated with a position.

We don't need a `PositionData` for every uptime. This might introduce errors and waste genesis state for no reason.

Suggest following the matrix approach we discussed to preserve the relationship between an uptime accumulator and its positions while avoiding duplicating `PositionData`

stackman27: changed! lmk what you think

x/concentrated-liquidity/genesis.go
```diff
@@ -184,16 +194,36 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *genesis.GenesisState {
...
 		positionData = append(positionData, genesis.PositionData{
-			LockId:         lockId,
-			Position:       &position,
-			FeeAccumRecord: feeAccumPositionRecord,
+			LockId:            lockId,
+			Position:          &position,
+			FeeAccumRecord:    feeAccumPositionRecord,
+			UptimeAccumRecord: uptimeAccumObject,
 		})
+
```
stackman27, line 226:
```suggestion
```

proto/osmosis/concentrated-liquidity/genesis.proto
```diff
@@ -53,6 +53,8 @@ message PositionData {
   uint64 lock_id = 2 [ (gogoproto.moretags) = "yaml:\"lock_id\"" ];
   osmosis.accum.v1beta1.Record fee_accum_record = 3
       [ (gogoproto.nullable) = false ];
+  repeated osmosis.accum.v1beta1.Record uptime_accum_record = 4
```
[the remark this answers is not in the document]
czarcas7ic: Agreed with changing this

x/concentrated-liquidity/genesis.go
```diff
@@ -184,15 +194,34 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *genesis.GenesisState {
...
 		if err != nil {
 			panic(err)
 		}
 
+		// Retrieve uptime incentive accumulator state for position
+		positionName := string(types.KeyPositionId(position.PositionId))
+		uptimeAccumulators, err := k.GetUptimeAccumulators(ctx, position.PoolId)
+		if err != nil {
+			panic(err)
+		}
```
czarcas7ic, line 208: What we could do is, store this var outside the for loop as well as a var that says what the last pool ID uptime accumulator that was retrieved was. This way, if we have many positions in the same pool, we don't have to keep retrieving the same uptime accum.

x/concentrated-liquidity/genesis_test.go
```diff
@@ -188,14 +211,22 @@ func (s *KeeperTestSuite) TestInitGenesis() {
...
 							LockId:         0,
 							Position:       withPositionId(testPositionModel, 2),
 							FeeAccumRecord: testFeeAccumRecord,
+							UptimeAccumRecord: []accum.Record{
+								accumRecordWithDefinedValues(accumRecord, sdk.NewDec(10000), sdk.NewInt(100), sdk.NewInt(50)),
+								accumRecordWithDefinedValues(accumRecord, sdk.NewDec(1000), sdk.NewInt(100), sdk.NewInt(50)),
+								accumRecordWithDefinedValues(accumRecord, sdk.NewDec(100), sdk.NewInt(100), sdk.NewInt(50)),
+								accumRecordWithDefinedValues(accumRecord, sdk.NewDec(10), sdk.NewInt(100), sdk.NewInt(50)),
+								accumRecordWithDefinedValues(accumRecord, sdk.NewDec(1), sdk.NewInt(100), sdk.NewInt(50)),
+							},
```
czarcas7ic, line 229: I see you added UptimeAccumRecords to both the import and export test cases, but where are we checking that it was properly imported or exported?

x/concentrated-liquidity/genesis.go
stackman27: hmm, have we done something similar elsewhere in the code? just asking for reference

x/concentrated-liquidity/genesis_test.go
stackman27: yup 

export: https://github.com/osmosis-labs/osmosis/pull/5076/files#diff-98c98ae5496ec51279e773847b7deb617103edd50a19b6ffa78a7f4889f8307eR813

import: https://github.com/osmosis-labs/osmosis/pull/5076/files#diff-98c98ae5496ec51279e773847b7deb617103edd50a19b6ffa78a7f4889f8307eR564

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * [#4974](https://github.com/osmosis-labs/osmosis/pull/4974) Add lock id to `MsgSuperfluidUndelegateAndUnbondLockResponse`
   * [#2741](https://github.com/osmosis-labs/osmosis/pull/2741) Prevent updating the twap record if `ctx.BlockTime <= record.Time` or `ctx.BlockHeight <= record.Height`. Exception, can update the record created when creating the pool in the same block.
   * [#5129](https://github.com/osmosis-labs/osmosis/pull/5129) Relax twap record validation in init genesis to allow one of the spot prices to be non-zero when twap error is observed.
+  * [#5076](https://github.com/osmosis-labs/osmosis/pull/5076) Export uptime accumulator mappings to GenesisState
```
czarcas7ic, line 69: I personally don't think we should add specific changes to CL in the changelog, since CL has not been released yet so nothing has really "changed" if that makes sense

x/concentrated-liquidity/genesis.go
czarcas7ic: What I mean is, you can set a var outside this for loop that is called something like `currentPoolId` and set it to zero as well as `currentPoolUptimeAccumulators`, then, on each iteration, set the `currentPoolId` var to whatever the poolId is for that iteration as well as store the `currentPoolUptimeAccumulators` outside of the for loop. Then, if the pool ID matched twice in a row or more, we wouldn't fetch the uptime accumulator multiple times which I imagine would improve performance when doing a state export.

That being said its not at all blocking, as I don't really know how much better in terms of performance this would be.

x/concentrated-liquidity/genesis_test.go
```diff
@@ -70,8 +70,31 @@ var (
...
 	}
+
+	accumRecord = accum.Record{
+		NumShares:             sdk.OneDec(),
+		AccumValuePerShare:    sdk.NewDecCoins(sdk.NewDecCoin("foo", sdk.NewInt(50))),
+		UnclaimedRewardsTotal: sdk.NewDecCoins(sdk.NewDecCoin("foo", sdk.NewInt(25))),
+		Options:               nil,
+	}
+
+	// five records because we have 5 supportive uptimes
```
czarcas7ic, line 81:
```suggestion
	// five records because we have 5 supported uptimes
```

x/concentrated-liquidity/genesis_test.go
```diff
@@ -743,6 +809,15 @@ func (s *KeeperTestSuite) TestExportGenesis() {
 				}
 			}
 
+			// Validate uptime accumulators
+			for i, actualPositionData := range actualExported.PositionData {
+				expectedPositionData := expectedGenesis.PositionData[i]
+				// validate incentive accumulator
+				for i, incentiveAccumulator := range actualPositionData.UptimeAccumRecords {
+					s.Require().Equal(expectedPositionData.UptimeAccumRecords[i], incentiveAccumulator)
+				}
```
czarcas7ic, line 818: I don't think these vars should say inventiveAccumulator right?

czarcas7ic (approved):
I think the logic is good, but not confident enough for merge, would like a second approve

x/concentrated-liquidity/genesis.go
```diff
@@ -79,6 +79,16 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState genesis.GenesisState) {
...
 		k.initOrUpdateAccumPosition(ctx, feeAccumObject, positionWrapper.FeeAccumRecord.AccumValuePerShare, feePositionKey, positionWrapper.FeeAccumRecord.NumShares, positionWrapper.FeeAccumRecord.UnclaimedRewardsTotal, positionWrapper.FeeAccumRecord.Options)
+
+		positionName := string(types.KeyPositionId(positionWrapper.Position.PositionId))
+		uptimeAccumulators, err := k.GetUptimeAccumulators(ctx, positionWrapper.Position.PoolId)
+		if err != nil {
+			panic(err)
+		}
+
+		for uptimeIndex, uptimeRecord := range positionWrapper.UptimeAccumRecords {
+			k.initOrUpdateAccumPosition(ctx, uptimeAccumulators[uptimeIndex], uptimeRecord.AccumValuePerShare, positionName, uptimeRecord.NumShares, uptimeRecord.UnclaimedRewardsTotal, uptimeRecord.Options)
```
AlpinYukseloglu, line 90: Is there a reason we use this function instead of `initOrUpdatePositionUptimeAccumulators` (and `initOrUpdatePositionFeeAccumulator` for fees)?

Not sure what the ultimate use of these records are so it might not matter, but this abstraction sets the records directly in state and thus circumvents important internal variable updates (e.g. calling `GetTotalShares` on the accums would return 0 even after the records were created)

AlpinYukseloglu (commented):
Overall logic looks good but had an important question about the abstraction used for setting up accumulator records so will wait to approve until that's resolved

x/concentrated-liquidity/genesis.go
stackman27: the way i thought of this was this; 
When we `initGenesis` at the very beginning all the values are set to 0. If we `exportGenesis` and then run `initGenesis` we get the latest data that includes "internal variable updates" and we just set new genesis to be the exported value

thus no need to run complex init logic since if export handles it by getting latest values

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * [#4912](https://github.com/osmosis-labs/osmosis/pull/4912) Export Position_lock_id mappings to GenesisState
   * [#4974](https://github.com/osmosis-labs/osmosis/pull/4974) Add lock id to `MsgSuperfluidUndelegateAndUnbondLockResponse`
   * [#2741](https://github.com/osmosis-labs/osmosis/pull/2741) Prevent updating the twap record if `ctx.BlockTime <= record.Time` or `ctx.BlockHeight <= record.Height`. Exception, can update the record created when creating the pool in the same block.
+  * [#5129](https://github.com/osmosis-labs/osmosis/pull/5129) Relax twap record validation in init genesis to allow one of the spot prices to be non-zero when twap error is observed.
```
stackman27, line 70:
```suggestion
```

x/concentrated-liquidity/genesis.go
```diff
@@ -184,16 +195,36 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *genesis.GenesisState {
...
 		positionData = append(positionData, genesis.PositionData{
-			LockId:         lockId,
-			Position:       &position,
-			FeeAccumRecord: feeAccumPositionRecord,
+			LockId:             lockId,
+			Position:           &position,
+			FeeAccumRecord:     feeAccumPositionRecord,
+			UptimeAccumRecords: uptimeAccumObject,
 		})
+
```
stackman27, line 227:
```suggestion
```

x/concentrated-liquidity/genesis.go
AlpinYukseloglu: I see, thanks for the clarification

AlpinYukseloglu (approved):

p0mvn:
There are 2 approvals so I'm merging this

PR #5077, opened 2023-05-04
fix(CL/osmoutils): update fee accumulator in fungify; fix deletion

p0mvn:
Closes: #XXX

This PR fixes the bug around fee rewards not being moved to the new position when fungifying positions.

It uses similar logic to incentives. As a result, a new function `moveRewardsToNewPositionAndDeleteOld` is introduced that is shared between incentives and fees.

Additionally, a new method was added to the `osmoutils/accum` called `DeletePosition` that does incentive claiming and global accumulator share decrement. The problem with the current approach of `RemoveFromPosition` + `ClaimRewards` is that both of these perform claiming. As a result, we end up claiming rewards twice. If we only call `ClaimRewards`, then the old position accumulator is persisted.

Therefore, the new method is required that does all of - claim rewards, delete a position from state and decrement global accumulator share value and rewrite it to state

## Brief Changelog

- Add `DeletePosition` helper in `osmoutils` and tests
- Add  `moveRewardsToNewPositionAndDeleteOld`` helper and tests
- Use it for both uptime rewards and fees in the fungify logic
- Update tests, add fee claiming tests for fungify

## Testing and Verifying

This is covered by old and added new tests


mattverse (commented):
Most changes looks good to me, leaving some questions and comments

osmoutils/accum/accum_test.go
```diff
@@ -1711,3 +1711,83 @@ func (suite *AccumTestSuite) TestAddToUnclaimedRewards() {
...
+			// Update global accumulator.
+			accObject.AddToAccumulator(tc.globalGrowth)
+
+			// System under test.
+			unclaimedRewards, err := accObject.DeletePosition(tc.positionName)
+
+			// Assertions.
+			if tc.expectedError != nil {
+				suite.Require().Error(err)
+				suite.Require().ErrorIs(err, tc.expectedError)
```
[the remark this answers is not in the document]
p0mvn: I don't think this is guaranteed because if an error happens in `GetTotalShares`, the deletion might already happen

x/concentrated-liquidity/position_test.go
```diff
@@ -718,7 +719,7 @@ func (s *KeeperTestSuite) TestValidateAndFungifyChargedPositions() {
 				totalLiquidity = totalLiquidity.Add(liquidityCreated)
 			}
 
-			// Increase block time by the fully charged duration
+			// Increase block time by the fully charged duration to make sure previously added positions are charged.
 			s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(testFullChargeDuration))
```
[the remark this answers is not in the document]
p0mvn: The charge is determined by join time. So if I create a position at time t, and then execute operation X at time t + testFullChargeDuration, I am guaranteed to have full charge in operation X

x/concentrated-liquidity/position.go
```diff
@@ -463,49 +469,41 @@ func (k Keeper) fungifyChargedPosition(ctx sdk.Context, owner sdk.AccAddress, po
...
-				if err != nil {
+				if err := moveRewardsToNewPositionAndDeleteOld(ctx, uptimeAccum, oldPositionName, newPositionUptimeAccName, uptimeGrowthOutside[uptimeIndex]); err != nil {
 					return 0, err
 				}
 			}
 		}
+
+		// Move fees into the new fee accumulator.
+		oldPositionFeeName := types.KeyFeePositionAccumulator(oldPositionId)
+		if err := moveRewardsToNewPositionAndDeleteOld(ctx, feeAccumulator, oldPositionFeeName, newPositionFeeAccName, feeGrowthOutside); err != nil {
```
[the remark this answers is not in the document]
p0mvn: I don't understand the question.

For every pool, there is a global fee accumulator, and for every pool and uptime there are global incentive uptime accumulators

By global, I mean per-pool.

Each position in the pool has an accumulator tracker for every global accumulator.

What's being done here is for every old positon's uptime tracker, we move its rewards into the new position's uptime tracker. The same goes for the fee tracker of the old positions.

Let me know if anything doesn't make sense

x/concentrated-liquidity/position.go
```diff
@@ -463,49 +469,41 @@ func (k Keeper) fungifyChargedPosition(ctx sdk.Context, owner sdk.AccAddress, po
...
+			return 0, err
```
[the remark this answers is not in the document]
p0mvn: Hmm I don't think there is a need for testing internal methods at higher levels of abstraction. The unit test for `moveRewardsToNewPositionAndDeleteOld` already does this.

In addition, the desired behavior around this is checked and guaranteed by:
https://github.com/osmosis-labs/osmosis/blob/4c0b2a90d7c6ea24f3c6230c01235348e57750f6/x/concentrated-liquidity/position_test.go#L1126-L1131

The reason we cannot claim is because there are no accumulators

x/concentrated-liquidity/incentives.go
```diff
@@ -709,6 +709,34 @@ func prepareAccumAndClaimRewards(accum accum.AccumulatorObject, positionKey stri
...
+// The positions must be associated with the given accumulator.
+// The given growth outside the positions range is used for claim rewards accounting.
+// The rewards are moved as "unclaimed rewards" to the new position.
+// Returns nil on success. Error otherwise.
+func moveRewardsToNewPositionAndDeleteOld(ctx sdk.Context, accum accum.AccumulatorObject, oldPositionName, newPositionName string, growthOutside sdk.DecCoins) error {
+	if oldPositionName == newPositionName {
+		return types.ModifySamePositionAccumulatorError{PositionAccName: oldPositionName}
+	}
+
+	if err := preparePositionAccumulator(accum, oldPositionName, growthOutside); err != nil {
```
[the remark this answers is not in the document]
p0mvn: No, this seems correct because we want to prepare and claim from the old position accumulator.

Once claimed, the rewards are moved into the "unclaimed rewards field" of the new accumulator.

Thanks for checking this though

osmoutils/accum/accum_test.go
```diff
@@ -1711,3 +1711,83 @@ func (suite *AccumTestSuite) TestAddToUnclaimedRewards() {
...
+
+	for name, tc := range tests {
+		tc := tc
+		suite.Run(name, func() {
+			suite.SetupTest()
+
+			// Setup.
+			accObject := accumPackage.MakeTestAccumulator(suite.store, testNameOne, initialCoinsDenomOne, emptyDec)
+
+			// Setup
```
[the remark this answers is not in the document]
p0mvn: changed locally

osmoutils/accum/accum.go
```diff
@@ -312,6 +312,30 @@ func (accum *AccumulatorObject) SetPositionCustomAcc(name string, customAccumula
 	return nil
 }
 
+// DeletePosition claims rewards and deletes the position from the accumulator state.
```
czarcas7ic, line 315: Please add what the return value is in the godoc here

osmoutils/accum/accum_test.go
```diff
@@ -1711,3 +1711,82 @@ func (suite *AccumTestSuite) TestAddToUnclaimedRewards() {
 		})
 	}
 }
+
+func (suite *AccumTestSuite) TestDeletePosition() {
+	// We setup store and accum
+	// once at beginning.
+	suite.SetupTest()
```
czarcas7ic, line 1718: You should either run setup here or below in the for loop but not both

x/concentrated-liquidity/incentives_test.go
```diff
@@ -4222,3 +4222,105 @@ func (s *KeeperTestSuite) TestGetLargestAuthorizedUptime() {
...
+			numShares:                1,
+			expectedUnclaimedRewards: defaultGlobalGrowth,
+		},
+		"default global growth equals growth outside": {
+			oldPositionName:          oldName,
+			newPositionName:          newName,
+			growthOutside:            defaultGlobalGrowth,
+			numShares:                1,
+			expectedUnclaimedRewards: emptyCoins,
+		},
```
czarcas7ic, line 4259: Do we have any checks that ensure the growthOutside cant be greater than the global growth? It probably doesn't need to be tested here, just came to my mind as I was reviewing this.

x/concentrated-liquidity/incentives_test.go
```diff
@@ -4222,3 +4222,105 @@ func (s *KeeperTestSuite) TestGetLargestAuthorizedUptime() {
...
+			// It is created this way for setup convenience.
+			pool := s.PrepareConcentratedPool()
+			testAccumulator, err := k.GetFeeAccumulator(s.Ctx, pool.GetId())
+			s.Require().NoError(err)
+
+			err = testAccumulator.NewPosition(tc.oldPositionName, sdk.NewDec(tc.numShares), nil)
+			s.Require().NoError(err)
+
+			err = testAccumulator.NewPosition(tc.newPositionName, sdk.OneDec(), nil)
+			s.Require().NoError(err)
```
czarcas7ic, line 4297: wdyt about setting this two 2 instead of 1 shares? Just was thinking some bugs were not discovered previously due to expecting 1 share always.

x/concentrated-liquidity/position.go
```diff
@@ -463,49 +469,41 @@ func (k Keeper) fungifyChargedPosition(ctx sdk.Context, owner sdk.AccAddress, po
...
-				// Add the unclaimed rewards to the new position.
-				err = uptimeAccum.AddToUnclaimedRewards(newPositionName, unclaimedRewardsForPosition)
-				if err != nil {
+				if err := moveRewardsToNewPositionAndDeleteOldAcc(ctx, uptimeAccum, oldPositionName, newPositionUptimeAccName, uptimeGrowthOutside[uptimeIndex]); err != nil {
 					return 0, err
 				}
 			}
 		}
+
+		// Move fees into the new fee accumulator.
```
czarcas7ic, line 499:
```suggestion
		// Move fees into the new fee accumulator and delete the old accumulator.
```

x/concentrated-liquidity/position_test.go
```diff
@@ -1041,6 +1048,89 @@ func (s *KeeperTestSuite) TestHasAnyPosition() {
...
+	s.Ctx = s.Ctx.WithBlockTime(defaultBlockTime)
+	totalPositionsToCreate := sdk.NewInt(int64(numPositions))
+	requiredBalances := sdk.NewCoins(sdk.NewCoin(ETH, DefaultAmt0.Mul(totalPositionsToCreate)), sdk.NewCoin(USDC, DefaultAmt1.Mul(totalPositionsToCreate)))
+
+	// Set test authorized uptime params.
+	params := s.App.ConcentratedLiquidityKeeper.GetParams(s.Ctx)
+	params.AuthorizedUptimes = []time.Duration{time.Nanosecond, testFullChargeDuration}
+	s.App.ConcentratedLiquidityKeeper.SetParams(s.Ctx, params)
+
+	// Fund accounts
```
czarcas7ic, line 1084:
```suggestion
	// Fund account
```

x/concentrated-liquidity/position_test.go
```diff
@@ -1041,6 +1048,89 @@ func (s *KeeperTestSuite) TestHasAnyPosition() {
...
+	collected, err = s.App.ConcentratedLiquidityKeeper.CollectFees(s.Ctx, defaultAddress, newPositionId)
+	s.Require().NoError(err)
+	s.Require().Equal(sdk.Coins(nil), collected)
+
+	// Check that cannot claim old positions
+	for _, oldPositionId := range expectedPositionIds {
+		collected, err = s.App.ConcentratedLiquidityKeeper.CollectFees(s.Ctx, defaultAddress, oldPositionId)
+		s.Require().Error(err)
+		s.Require().Equal(sdk.Coins{}, collected)
+	}
```
czarcas7ic, line 1131: Can we also check that the position can't even be fetched since it should now be non-existent?

czarcas7ic (approved):
Nice work!

osmoutils/accum/accum_test.go
p0mvn: Copied over and didn't notice, removed now

x/concentrated-liquidity/incentives_test.go
p0mvn: If that happens, it would panic with a negative coin amount error during subtraction when claiming

x/concentrated-liquidity/incentives_test.go
p0mvn: I will set it but I don't think it matters for the new position.

I set the number of shares to a configurable value for the old position.

The reason why it does not matter for the new position is because the only mutation we expect is the move to the "unclaimed rewards" field which is independent from the number of shares.

For the old position, it does matter because the value we claim depends on the number of shares.

x/concentrated-liquidity/position_test.go
p0mvn: Done: https://github.com/osmosis-labs/osmosis/pull/5077/commits/cdd8ed9ed6798956965c5860b0126346ae473a05

PR #5078, opened 2023-05-04
fix(osmoutils/accum): mutate accumulator receiver

p0mvn:
Closes: #XXX

This PR fixes some subtle bugs in the accumulator package that were discovered during "MsgFungify" tests.

The core of the bug is that while we would update the accumulator in state, we would not mutate the receiver. As a result, subsequent operations would use out-of-date values from the accumulator fields.

This PR fixes the bug and expands every applicable test to make sure that both receiver and in-state global accumulators are updated.

For example,

```go
func (accum AccumulatorObject) NewPosition(name string, numShareUnits sdk.Dec, options *Options) error {
     ....
     accum.totalShares = accum.totalShares.Add(numShareUnits)
     return nil
}

func main() {
    accum := ... // create accum with zero shares
    accum.NewPosition("test", sdk.OneDec(), nil)
   // Note that at this point accum.totalShares is still zero
}
```

The reason is that we must have a pointer receiver on the `NewPosition` method for the change to take effect.

It looks like none of our CL test cases broke but I ran into this problem by running into a bug that wasn't covered by our test suite in "MsgFungify". The fix to that is coming separately.

It might be that there are more issues stemming from the lack of this change.

## Brief Changelog

- changed all mutative accumulator methods to have a pointer reciever
- refactored every applicable test to make sure that both state and the local var are updated.

## Testing and Verifying

Added assertions


osmoutils/accum/export_test.go
```diff
@@ -75,3 +77,15 @@ func (accum *AccumulatorObject) SetValue(value sdk.DecCoins) {
 func (o *Options) Validate() error {
 	return o.validate()
 }
+
+// WARNING: only used in tests to make sure that receiver is mutated.
+// Do not move out of export_tes.go and do not use in production code.
```
p0mvn, line 82:
```suggestion
// Do not move out of export_test.go and do not use in production code.
```

osmoutils/accum/export_test.go
```diff
@@ -75,3 +77,15 @@ func (accum *AccumulatorObject) SetValue(value sdk.DecCoins) {
...
+func (accum AccumulatorObject) GetTotalShareField() sdk.Dec {
+	return accum.totalShares
+}
+
+// WARNING: only used in tests to make sure that receiver is mutated.
+// Do not move out of export_tes.go and do not use in production code.
```
p0mvn, line 88:
```suggestion
// Do not move out of export_test.go and do not use in production code.
```

osmoutils/accum/accum.go
```diff
@@ -217,7 +219,7 @@ func (accum AccumulatorObject) AddToPositionCustomAcc(name string, newShares sdk
 // the unclaimed and newly accrued rewards and returns them alongside the redeemed shares. Then, it
 // overwrites the position record with the updated number of shares. Since it accrues rewards, it
 // also moves up the position's accumulator value to the current accum val.
-func (accum AccumulatorObject) RemoveFromPosition(name string, numSharesToRemove sdk.Dec) error {
+func (accum *AccumulatorObject) RemoveFromPosition(name string, numSharesToRemove sdk.Dec) error {
```
p0mvn, line 222: Note: this is the core of the change that makes the receiver updated correctly.

This is important so that if we use the local variable after calling a mutative method on it, we can be confident that the local variable is updated

osmoutils/accum/accum.go
```diff
@@ -98,7 +98,7 @@ func (accum AccumulatorObject) GetPosition(name string) (Record, error) {
 	return position, nil
 }
 
-func setAccumulator(accum AccumulatorObject, value sdk.DecCoins, shares sdk.Dec) {
+func setAccumulator(accum *AccumulatorObject, value sdk.DecCoins, shares sdk.Dec) {
```
mattverse, line 101: what's the reason we need to change the argument to pointer? Seems like here we might not need to change the argument

mattverse: no on the second thought, I think this change makes sense given that all the callers have accum as pointers anyways

mattverse (approved):
Feel like code readability has decreased with the mixture of usage of raw accum objects and pointer accum objects, but seems like it's unevitable. 

Everything in this PR makes sense to me, LGTM

osmoutils/accum/accum_test.go
```diff
@@ -284,10 +297,10 @@ func (suite *AccumTestSuite) TestNewPositionCustomAcc() {
 	suite.SetupTest()
 
 	// Setup.
-	accObject := accumPackage.MakeTestAccumulator(suite.store, testNameOne, initialCoinsDenomOne, emptyDec)
+	defaultAccObject := accumPackage.MakeTestAccumulator(suite.store, testNameOne, initialCoinsDenomOne, emptyDec)
```
czarcas7ic, line 300: nit: we called the defaultAccObject above an acc object with empty coins, but the defaultAccObject here has coins. Would be nice to stick to a single standard of what "default" is to make tests easier to follow.

osmoutils/accum/export_test.go
```diff
@@ -75,3 +77,15 @@ func (accum *AccumulatorObject) SetValue(value sdk.DecCoins) {
...
+// Do not move out of export_test.go and do not use in production code.
+func (accum AccumulatorObject) GetTotalShareField() sdk.Dec {
+	return accum.totalShares
+}
+
+// WARNING: only used in tests to make sure that receiver is mutated.
+// Do not move out of export_test.go and do not use in production code.
+func (accum AccumulatorObject) GetValueField() sdk.DecCoins {
+	return accum.value
+}
```
czarcas7ic, line 91: Why would getters ever be dangerous?

czarcas7ic (approved):
LGTM, nice find!

osmoutils/accum/export_test.go
p0mvn: We already have `GetValue` and `GetTotalShare` getters that read from the state instead of grabbing this value. Just left a note in case someone tries to duplicate it.

osmoutils/accum/accum_test.go
p0mvn: This is the default per-test. That's the reason why we have a default for every test

PR #5083, opened 2023-05-04
test: TestGetLiquidityFromAmounts

p0mvn:
Closes: #XXX

Came up in discussions with @czarcas7ic . Trying to find the relationship between sqrt price and amount proportions and why liquidity 0 is not equal to liquidity 1 when amount proportions are equal to price.

Current understanding:
The invariant must hold that both tokens are contributing to the liquidity equally. For non-full range, the distance between P_c and P_l or P_c and P_h is not proportional to the price because these are virtual reserves that might not be proportional to the price itself.

However, in the full range, having the amounts ration equal to price leads to same liquidities because all reserves are accounted for.

## Testing and Verifying

### DIFF
--- a/x/concentrated-liquidity/math/math_test.go
+++ b/x/concentrated-liquidity/math/math_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/osmosis-labs/osmosis/osmomath"
 	"github.com/osmosis-labs/osmosis/v15/app/apptesting"
 	"github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/math"
+	cltypes "github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity/types"
 )
 
 type ConcentratedMathTestSuite struct {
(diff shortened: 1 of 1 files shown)


x/concentrated-liquidity/math/math_test.go
```diff
@@ -323,6 +337,39 @@ func (suite *ConcentratedMathTestSuite) TestGetLiquidityFromAmounts() {
...
+			currentSqrtP:   sqrt(sdk.NewDec(2)),
+			sqrtPHigh:      sqrt(sdk.NewDec(3)),
+			sqrtPLow:       sqrt(sdk.NewDec(1)),
+			amount0Desired: sdk.NewInt(1),
+			amount1Desired: sdk.NewInt(2),
+
+			expectedLiquidity:  sdk.MustNewDecFromStr("4.828427124746190095").String(),
+			expectedLiquidity0: sdk.MustNewDecFromStr("7.706742302257039729"),
+			expectedLiquidity1: sdk.MustNewDecFromStr("4.828427124746190095"),
+		},
```
czarcas7ic, line 372: I really appreciate these added test cases to help better demonstrate this!

Do we know where the error in the 10^-9 place is coming from? I assume its because we divide in the liquidity1 calculation right?

czarcas7ic (approved):

x/concentrated-liquidity/math/math_test.go
p0mvn: It actually seems to be correct:

```
>>> 16 / (Decimal("4").sqrt() - Decimal("0.000000000000000001").sqrt())
Decimal('8.000000004000000002000000001')
```

It might be the closeness to 0 vs infinity

Issue #5087, opened 2023-05-04
bug: claims more incentives than expected after fungify


p0mvn:
## Background

Repro:
```bash
# pool
osmosisd tx concentratedliquidity create-pool uion uosmo 1 0.001 --from lo-test1 --keyring-backend test --chain-id localosmosis -b=block --fees 875uosmo

# create incentive (make sure to change the date)
osmosisd tx concentratedliquidity create-incentive uosmo 69082 0.1 "2023-05-04T20:14:10.419543805" 1ns --pool-id 1 --from lo-test1 --keyring-backend test --chain-id localosmosis -b=block --fees 875uosmo

# create some positions 
osmosisd tx concentratedliquidity create-position 0 100 1000000uosmo 1000000uion 1 0 --pool-id 1 --from lo-test1 --keyring-backend test -b=block --chain-id=localosmosis --fees 875uosmo

# check claimable incentives
osmosisd q concentratedliquidity claimable-incentives 1

# check claimable incentives
osmosisd q concentratedliquidity claimable-incentives 2

osmosisd tx concentratedliquidity create-position 0 100 2000000uosmo 2000000uion 1 0 --pool-id 1 --from lo-test1 --keyring-backend test -b=block --chain-id=localosmosis --fees 875uosmo

# check claimable incentives
osmosisd q concentratedliquidity claimable-incentives 3

# observe that can claim too much
```

## Acceptance Criteria
- understand and fix the bug

p0mvn:
Fixed

PR #5089, opened 2023-05-05
test(CL): fungify charged position incentive claiming test

p0mvn:
Closes: #XXX

This attempts to reproduce: https://github.com/osmosis-labs/osmosis/issues/5087

However, the numbers end up working out successfully with some negligible rounding error.

## Testing and Verifying


x/concentrated-liquidity/position_test.go
```diff
@@ -1141,6 +1142,115 @@ func (s *KeeperTestSuite) TestFungifyChargedPositions_SwapAndClaimFees() {
...
+	// Set test authorized uptime params.
+	params := s.App.ConcentratedLiquidityKeeper.GetParams(s.Ctx)
+	params.AuthorizedUptimes = []time.Duration{time.Nanosecond, testFullChargeDuration}
+	s.App.ConcentratedLiquidityKeeper.SetParams(s.Ctx, params)
+
+	// Fund accounts
+	s.FundAcc(defaultAddress, requiredBalances)
+
+	// Create CL pool
+	pool := s.PrepareCustomConcentratedPool(s.TestAccs[0], ETH, USDC, DefaultTickSpacing, swapFee)
```
czarcas7ic, line 1179: I think we used this exact logic in another test right? Do you think it would be helpful to make a helper function that abstracts this setup logic away?

x/concentrated-liquidity/position_test.go
```diff
@@ -1141,6 +1142,115 @@ func (s *KeeperTestSuite) TestFungifyChargedPositions_SwapAndClaimFees() {
...
+
+	roundingEror := int64(5)
```
czarcas7ic, line 1181:
```suggestion
	roundingError := int64(5)
```

czarcas7ic: Where exactly does 5 come from?

x/concentrated-liquidity/position_test.go
```diff
@@ -1141,6 +1142,115 @@ func (s *KeeperTestSuite) TestFungifyChargedPositions_SwapAndClaimFees() {
...
+
+	// sync accumulators
+	err := s.App.ConcentratedLiquidityKeeper.UpdateUptimeAccumulatorsToNow(s.Ctx, pool.GetId())
+	s.Require().NoError(err)
+
+	claimableIncentives := sdk.NewCoins()
+	for i := 0; i < numPositions; i++ {
+		positionIncentices, dust, err := s.App.ConcentratedLiquidityKeeper.GetClaimableIncentives(s.Ctx, uint64(i+1))
+		s.Require().NoError(err)
+		s.Require().Equal(sdk.Coins(nil), dust)
```
czarcas7ic, line 1222: Why is dust required to be nil?

x/concentrated-liquidity/position_test.go
```diff
@@ -1141,6 +1142,115 @@ func (s *KeeperTestSuite) TestFungifyChargedPositions_SwapAndClaimFees() {
...
+		claimableIncentives = claimableIncentives.Add(positionIncentices...)
+	}
+
+	actualClaimedAmount := claimableIncentives.AmountOf(USDC)
+	s.Require().Equal(0, roundingTolerance.Compare(expectedAmount, actualClaimedAmount), "expected: %s, got: %s", expectedAmount, actualClaimedAmount)
+
+	// First run non mutative validation and check results
+	newPositionId, err := s.App.ConcentratedLiquidityKeeper.FungifyChargedPosition(s.Ctx, defaultAddress, expectedPositionIds)
```
czarcas7ic, line 1230: This is "non-mutative" though right?

x/concentrated-liquidity/position_test.go
```diff
@@ -1141,6 +1142,115 @@ func (s *KeeperTestSuite) TestFungifyChargedPositions_SwapAndClaimFees() {
...
+	s.Require().NoError(err)
+
+	// Claim fees
+	collected, _, err := s.App.ConcentratedLiquidityKeeper.CollectIncentives(s.Ctx, defaultAddress, newPositionId)
+	s.Require().NoError(err)
+
+	// Validate that the correct incentives amount was collected.
+	actualClaimedAmount = collected.AmountOf(USDC)
```
czarcas7ic, line 1238: Can we check the entire collect coins object rather than just USDC so we know if the collected response is exactly what we are expecting?

czarcas7ic (commented):
Want to review again once I get some of these questions answered just so I can 100 percent understand.

x/concentrated-liquidity/position_test.go
p0mvn: The setup is not exactly the same between them so I think it's better to have some duplication for convenience

x/concentrated-liquidity/position_test.go
p0mvn: These are actually forfeited incentives, not dust. Renamed now.

We should not forfeit anything here

x/concentrated-liquidity/position_test.go
p0mvn: Done by checking length of the collected coins being one

x/concentrated-liquidity/position_test.go
p0mvn: I believe it's just truncating in favor of the pool during claiming. The strictest is actually 3, just changed it to 3

czarcas7ic (approved):
Approving, but the one thing I am not 100 percent on is where the three units are getting rounded away from. Does this number increase with more positions / longer emission time? I am just curious what the upper limit of this value is.

p0mvn:
> Approving, but the one thing I am not 100 percent on is where the three units are getting rounded away from. Does this number increase with more positions / longer emission time? I am just curious what the upper limit of this value is.

I confirmed that the rounding error increases linearly with the number of positions. 1 off for each increment of `numPositions`.

Going to add a comment and merge

PR #5091, opened 2023-05-05
[InternalReview][MsgFungifyChargedPositions]

p0mvn:
Closes: #XXX

https://app.clickup.com/37420681/v/dc/13nzm9-19302/13nzm9-35382

Review of `MsgFungifyChargedPositions`

This is part 1 only, will be continuing with more internal review.


.vscode/launch.json
```diff
@@ -6,25 +6,14 @@
             "type": "go",
             "request": "launch",
             "mode": "test",
-            "program": "${workspaceFolder}/tests/e2e",
```
p0mvn, line 9: Note to self: revert

x/concentrated-liquidity/msg_server.go
```diff
@@ -257,13 +255,6 @@ func (server msgServer) FungifyChargedPositions(goCtx context.Context, msg *type
 			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
 			sdk.NewAttribute(sdk.AttributeKeySender, msg.Sender),
 		),
-		sdk.NewEvent(
```
p0mvn, line 260: Note: moved to keeper

x/concentrated-liquidity/incentives.go
```diff
@@ -742,6 +764,20 @@ func moveRewardsToNewPositionAndDeleteOldAcc(ctx sdk.Context, accum accum.Accumu
 		return err
 	}
 
+	ctx.EventManager().EmitEvents(sdk.Events{
```
p0mvn, line 767: Note: I added these events to assist with debugging https://github.com/osmosis-labs/osmosis/issues/5087

x/concentrated-liquidity/lp_test.go
```diff
@@ -483,16 +482,16 @@ func (s *KeeperTestSuite) TestWithdrawPosition() {
 
 			// Create a position from the parameters in the test case.
 			if tc.createLockLocked {
-				_, _, _, liquidityCreated, _, _, err = concentratedLiquidityKeeper.CreateFullRangePositionLocked(ctx, pool.GetId(), owner, fundCoins, tc.timeElapsed)
+				_, _, _, liquidityCreated, _, _, err = concentratedLiquidityKeeper.CreateFullRangePositionLocked(s.Ctx, pool.GetId(), owner, fundCoins, tc.timeElapsed)
```
p0mvn, line 485: Note: these changes were made because some helpers would update `s.Ctx` from the suite, and that change would not be propagated to the local var `ctx`, causing some test failures

x/poolmanager/client/cli/query.go
```diff
@@ -68,6 +69,15 @@ func GetCmdNumPools() (*osmocli.QueryDescriptor, *queryproto.NumPoolsRequest) {
...
 
+// GetCmdAllPools return all pools available across Osmosis modules.
+func GetCmdAllPools() (*osmocli.QueryDescriptor, *queryproto.AllPoolsRequest) {
+	return &osmocli.QueryDescriptor{
+		Use:   "all-pools",
+		Short: "Query all pools on the Osmosis chain",
+		Long:  "{{.Short}}",
+	}, &queryproto.AllPoolsRequest{}
+}
+
```
czarcas7ic, line 80: Just a note here, under QueryDescriptor I think we need to add hasPagination to true. This will add the corresponding flags and such, but I don't think its properly paginates out of the box (there is no osmocli query that currently uses it, and any paginated query we actually rely on uses to old cli setup).

czarcas7ic (approved):
Nice work! Would be great if you could add the pagination to the AllPools query, but is not blocking (just don't want to forget about it until later when we need to use it)

x/poolmanager/client/cli/query.go
p0mvn: Pagination isn't implemented for this query. Since they are they only known client, spoke to Jon and decided not to added for now due to time constraints

x/concentrated-liquidity/position.go
```diff
@@ -533,14 +537,40 @@ func (k Keeper) fungifyChargedPosition(ctx sdk.Context, owner sdk.AccAddress, po
...
+			sdk.NewAttribute(types.AttributeClaimableIncentives, claimableIncentives.String()),
+		),
+	})
+
 	return newPositionId, nil
 }
 
 // validatePositionsAndGetTotalLiquidity validates a list of positions owned by the caller and returns their total liquidity.
 // It also returns the pool ID, lower tick, and upper tick that all the provided positions are confirmed to share.
+// - the caller does not own all the positions
```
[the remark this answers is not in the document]
p0mvn: added

PR #5116, opened 2023-05-07
faddat/forcetypeassert

faddat:
- fix forced type assertions
- revert change to linter config

Closes: #XXX

## Brief Changelog

*(for example:)*
 
## Testing and Verifying

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -192,11 +192,11 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 	s.Require().NoError(err)
 
 	var (
-		denom0      string  = "uion"
-		denom1      string  = "uosmo"
-		tickSpacing uint64  = 100
-		swapFee             = "0.001" // 0.1%
-		swapFeeDec  sdk.Dec = sdk.MustNewDecFromStr("0.001")
+		denom0             = "uion"
+		denom1             = "uosmo"
+		tickSpacing uint64 = 100
+		swapFee            = "0.001" // 0.1%
+		swapFeeDec         = sdk.MustNewDecFromStr("0.001")
 	)
 
 	// Get the permisionless pool creation parameter.
@@ -1056,12 +1056,34 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
 	var response map[string]interface{}
 	s.Require().Eventually(func() bool {
 		response, err = nodeA.QueryWasmSmartObject(contractAddr, fmt.Sprintf(`{"get_total_funds": {"addr": "%s"}}`, senderBech32))
-		totalFunds := response["total_funds"].([]interface{})[0]
-		amount := totalFunds.(map[string]interface{})["amount"].(string)
-		denom := totalFunds.(map[string]interface{})["denom"].(string)
+		if err != nil {
+			return false
+		}
+
+		totalFundsIface, ok := response["total_funds"].([]interface{})
+		if !ok || len(totalFundsIface) == 0 {
+			return false
+		}
+
+		totalFunds, ok := totalFundsIface[0].(map[string]interface{})
+		if !ok {
+			return false
+		}
+
+		amount, ok := totalFunds["amount"].(string)
+		if !ok {
+			return false
+		}
+
+		denom, ok := totalFunds["denom"].(string)
+		if !ok {
+			return false
+		}
+
 		// check if denom contains "uosmo"
-		return err == nil && amount == strconv.FormatInt(transferAmount, 10) && strings.Contains(denom, "ibc")
+		return amount == strconv.FormatInt(transferAmount, 10) && strings.Contains(denom, "ibc")
 	},
+
 		15*time.Second,
 		10*time.Millisecond,
 	)
@@ -1108,8 +1130,11 @@ func (s *IntegrationTestSuite) TestPacketForwarding() {
 		if err != nil {
 			return false
 		}
-		count := response["count"].(float64)
-		return err == nil && count == 0
+		countValue, ok := response["count"].(float64)
+		if !ok {
+			return false
+		}
+		return err == nil && countValue == 0
 	},
 		15*time.Second,
 		10*time.Millisecond,
--- a/x/concentrated-liquidity/pool_test.go
+++ b/x/concentrated-liquidity/pool_test.go
@@ -15,7 +15,10 @@ import (
 func (s *KeeperTestSuite) TestInitializePool() {
 	// Create a valid PoolI from a valid ConcentratedPoolExtension
 	validConcentratedPool := s.PrepareConcentratedPool()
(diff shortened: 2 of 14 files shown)


tests/e2e/e2e_test.go
```diff
@@ -1108,8 +1130,11 @@ func (s *IntegrationTestSuite) TestPacketForwarding() {
 		if err != nil {
 			return false
 		}
-		count := response["count"].(float64)
-		return err == nil && count == 0
+		countValue, ok := response["count"].(float64)
+		if !ok {
+			return false
+		}
+		return err == nil && countValue == 0
```
pysel, line 1137: I think there is no need to check for error being nil, because at the point of return it has already been checked (line 1130). 

```suggestion
		return countValue == 0
```

x/concentrated-liquidity/pool_test.go
```diff
@@ -209,7 +212,10 @@ func (s *KeeperTestSuite) TestPoolIToConcentratedPool() {
 
 	// Create default CL pool
 	concentratedPool := s.PrepareConcentratedPool()
-	poolI := concentratedPool.(poolmanagertypes.PoolI)
+	poolI, ok := concentratedPool.(poolmanagertypes.PoolI)
+	if !ok {
+		s.Fail("failed to cast pool to PoolI")
```
pysel, line 217: I think it makes more sense to use `FailNow` when a type conversion error happens
```suggestion
		s.FailNow("failed to cast pool to PoolI")
```

x/gamm/keeper/swap_test.go
```diff
@@ -210,20 +210,26 @@ func (suite *KeeperTestSuite) TestCalcOutAmtGivenIn() {
...
 			keeper := suite.App.GAMMKeeper
 			ctx := suite.Ctx
-
 			var pool poolmanagertypes.PoolI
 			if test.param.poolType == "balancer" {
 				poolId := suite.PrepareBalancerPool()
 				poolExt, err := suite.App.GAMMKeeper.GetPool(suite.Ctx, poolId)
 				suite.NoError(err)
-				pool = poolExt.(poolmanagertypes.PoolI)
+				poolI, ok := poolExt.(poolmanagertypes.PoolI)
```
pysel, line 218: why not use `pool` variable here directly?

pysel: same question for other similar `pool = poolI` cases

x/incentives/keeper/suite_test.go
```diff
@@ -1,8 +1,8 @@
 package keeper_test
 
 import (
+	"crypto/rand"
```
pysel, line 4: why change of package?

pysel (commented):
Thanks! Left a few comments:

x/incentives/keeper/suite_test.go
faddat: great question -- math/rand has been deprecated in favor of crypto/rand :)

x/gamm/keeper/swap_test.go
faddat: so, the "type assertion" -- which I had to spend a bunch of time researching today, allows you to get at the "underlying type" of the variable, and those are set up like this:

```
poolExt.(poolmanagertypes.PoolI)
```

but they can fail.  But... as for... why not just call it pool -- in fact, maybe we should :)

faddat:
Thanks for the review @pysel :)

x/concentrated-liquidity/pool_test.go
```diff
@@ -15,7 +15,10 @@ import (
 func (s *KeeperTestSuite) TestInitializePool() {
 	// Create a valid PoolI from a valid ConcentratedPoolExtension
 	validConcentratedPool := s.PrepareConcentratedPool()
-	validPoolI := validConcentratedPool.(poolmanagertypes.PoolI)
+	validPoolI, ok := validConcentratedPool.(poolmanagertypes.PoolI)
+	if !ok {
+		s.FailNow("validConcentratedPool is not a PoolI")
+	}
```
p0mvn, line 21: nit: less verbosity improves readability which is preferable for tests IMO
```suggestion
	s.Require().True(ok)
```

x/concentrated-liquidity/pool_test.go
```diff
@@ -209,7 +212,10 @@ func (s *KeeperTestSuite) TestPoolIToConcentratedPool() {
 
 	// Create default CL pool
 	concentratedPool := s.PrepareConcentratedPool()
-	poolI := concentratedPool.(poolmanagertypes.PoolI)
+	poolI, ok := concentratedPool.(poolmanagertypes.PoolI)
+	if !ok {
+		s.FailNow("failed to cast pool to PoolI")
+	}
```
p0mvn, line 218:
```suggestion
	s.Require().True(ok)
```

x/gamm/keeper/swap_test.go
```diff
@@ -273,27 +271,35 @@ func (suite *KeeperTestSuite) TestCalcInAmtGivenOut() {
...
 				suite.NoError(err)
-				pool = poolExt.(poolmanagertypes.PoolI)
+				pool, _ = poolExt.(poolmanagertypes.PoolI)
+			default:
+				suite.FailNow("unsupported pool type")
+			}
+
+			if pool == nil {
+				suite.FailNow("failed to cast pool to poolI")
 			}
```
p0mvn, line 298:
```suggestion
			suite.Require().NotNil(pool)
```

p0mvn (approved):
LGTM, nice work. Only nits

tests/e2e/e2e_test.go
```diff
@@ -1056,12 +1056,34 @@ func (s *IntegrationTestSuite) TestIBCWasmHooks() {
 	var response map[string]interface{}
 	s.Require().Eventually(func() bool {
 		response, err = nodeA.QueryWasmSmartObject(contractAddr, fmt.Sprintf(`{"get_total_funds": {"addr": "%s"}}`, senderBech32))
-		totalFunds := response["total_funds"].([]interface{})[0]
-		amount := totalFunds.(map[string]interface{})["amount"].(string)
-		denom := totalFunds.(map[string]interface{})["denom"].(string)
```
nicolaslara, line 1061: I find the old versions much more readable here (since it's just a test). I'm not strongly against the type assertion checks and get the point in chain code, but in tests I tend to favor making the test shorted since robustness doesn't matter.

For json in particular, we may want to use https://github.com/tidwall/gjson

nicolaslara (approved):
looks good to me. I have a stylistic preference that I left a comment for, but happy either way.

faddat:
@nicolaslara is it okay if I try to get your style preference into the final PR for #5061 ?

makes sense to me actually

nicolaslara:
sounds good! :)

czarcas7ic (approved):

PR #5119, opened 2023-05-08
fix[CL]: create position with underlying denoms only

czarcas7ic:
Closes: #XXX

We currently don't check the denoms when amounts are provided for the createPosition logic. 

Additionally, we take a single coin object for both amount0 and amount1. This becomes problematic when wanting to provide liquidity in a range that only requires a single asset, since a zero coin amount make a coin invalid.

This PR fixes the above two issues by accepting a single coins object and checks the underlying denoms before running the createPosition logic.

## Brief Changelog

- Checks amounts of pool token0 and pool token1 in the provided coins for createPosition
- Removed tokenDesired0 and tokenDesired1 from MsgCreatePosition in favor of tokensProvided
- Updates all tests that use createPosition to utilize Coins object instead of two individual Coin objects
- Requires CreateFullRangePosition to have a coins length of exactly 2
- Exapands validateBasic for MsgCreatePosition


mattverse:
rip merge conflicts with the msg create position PR 😭

x/concentrated-liquidity/lp_test.go
```diff
@@ -51,10 +51,10 @@ var (
 		lowerTick:                         DefaultLowerTick,
 		upperTick:                         DefaultUpperTick,
 		currentSqrtP:                      DefaultCurrSqrtPrice,
-		amount0Desired:                    DefaultAmt0,
+		tokensProvided:                    DefaultCoins,
+		customTokensProvided:              false,
```
czarcas7ic, line 55: This work around was the cleanest way to allow a nil sdk.Coins{} value to be provided in the test case and not get overridden.

czarcas7ic:
@mattverse we can get yours in first and I will correct the merge conflicts

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -27,20 +27,16 @@ message MsgCreatePosition {
...
-  cosmos.base.v1beta1.Coin token_desired0 = 5 [
-    (gogoproto.moretags) = "yaml:\"token_desired0\"",
-    (gogoproto.nullable) = false
-  ];
-  cosmos.base.v1beta1.Coin token_desired1 = 6 [
-    (gogoproto.moretags) = "yaml:\"token_desired1\"",
-    (gogoproto.nullable) = false
+  repeated cosmos.base.v1beta1.Coin tokens_provided = 5 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
```
p0mvn, line 32: Why do we have these as coins? Why can't we assume that if a user attempts to create a position in a pool where pool has denom A as token 0 and B as token 1, then the denoms are implied?

x/concentrated-liquidity/types/msgs.go
```diff
@@ -32,12 +32,22 @@ func (msg MsgCreatePosition) ValidateBasic() error {
...
 
-	if !msg.TokenDesired1.IsValid() || msg.TokenDesired1.IsZero() {
-		return fmt.Errorf("Invalid coins (%s)", msg.TokenDesired1.String())
+	if !msg.TokensProvided.IsValid() {
+		return fmt.Errorf("Invalid coins (%s)", msg.TokensProvided.String())
+	}
+
+	if len(msg.TokensProvided) > 2 {
+		return CoinLengthError{Length: len(msg.TokensProvided), MaxLength: 2}
+	}
```
p0mvn, line 45: Since there are always 2 tokens expected, I think it is better UX to have 2 separate fields - one for token0 and another one for token1.

This way, people would not be able to shoot themselves in the foot by providing more tokens and failing the check here

czarcas7ic: (this is what you were suggesting, I am just pointing out that just because the two fields make sense here does not mean we should also have it for createPosition as well)

proto/osmosis/concentrated-liquidity/tx.proto
p0mvn: discussed offline, sg

x/concentrated-liquidity/types/msgs.go
p0mvn: Ah it's because we cannot have a zero coin

p0mvn (approved):
LGTM

mattverse (approved):
Leaving some questions I would like to get an answer for before merge, otherwise lgtm

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -27,20 +27,16 @@ message MsgCreatePosition {
...
   int64 lower_tick = 3 [ (gogoproto.moretags) = "yaml:\"lower_tick\"" ];
   int64 upper_tick = 4 [ (gogoproto.moretags) = "yaml:\"upper_tick\"" ];
-  cosmos.base.v1beta1.Coin token_desired0 = 5 [
-    (gogoproto.moretags) = "yaml:\"token_desired0\"",
-    (gogoproto.nullable) = false
-  ];
-  cosmos.base.v1beta1.Coin token_desired1 = 6 [
-    (gogoproto.moretags) = "yaml:\"token_desired1\"",
-    (gogoproto.nullable) = false
+  repeated cosmos.base.v1beta1.Coin tokens_provided = 5 [
```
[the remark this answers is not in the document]
czarcas7ic: proto comment added https://github.com/osmosis-labs/osmosis/pull/5119/commits/7d47173e47f7320679d89f75b45fa45b4b6976fd

x/concentrated-liquidity/lp.go
```diff
@@ -43,6 +43,8 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 	if err := validateTickRangeIsValid(pool.GetTickSpacing(), lowerTick, upperTick); err != nil {
 		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
 	}
+	amount0Desired := tokensProvided.AmountOf(pool.GetToken0())
+	amount1Desired := tokensProvided.AmountOf(pool.GetToken1())
```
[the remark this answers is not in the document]
czarcas7ic: No, since if both are zero then the liquidityDelta that is calculated is zero. That being said, I think you are right if your point was we should catch this earlier, will add this now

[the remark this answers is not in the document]
czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/5119/commits/cbdc89814db9c94515e42522554b43346cdec3aa

Issue #5041, opened 2023-05-02
v16 blocker: twap records failing init genesis when state-exported v16 testnet is created from v15 genesis


p0mvn:
## Background

From @niccoloraspa 

Daily upgrade test failed again. This time before the upgrade.
On the initGenesis() on a state exported testnified v15 genesis.

```
2:23PM INF running initialization for module module=capability
2:23PM INF running initialization for module module=auth
2:24PM INF running initialization for module module=bank
2:24PM INF running initialization for module module=distribution
2:26PM INF running initialization for module module=downtimedetector
2:26PM INF running initialization for module module=staking
2:27PM INF running initialization for module module=slashing
2:27PM INF running initialization for module module=gov
2:27PM INF running initialization for module module=mint
2:27PM INF running initialization for module module=crisis
2:27PM INF running initialization for module module=ibc
2:27PM INF running initialization for module module=interchainaccounts
2:27PM INF running initialization for module module=gamm
2:27PM INF running initialization for module module=poolmanager
2:27PM INF running initialization for module module=protorev
2:27PM INF running initialization for module module=twap
panic: twap record p1 last spot price must be zero due to having an error, was (0.000000000009024277)

goroutine 1 [running]:
github.com/osmosis-labs/osmosis/v15/x/twap.Keeper.InitGenesis({{0x2ed92f8, 0xc0014e5c60}, 0xc0014e5d80, {{0x2f09580, 0xc00122c370}, 0xc000211a60, {0x2ed92f8, 0xc0014e5bb0}, {0x2ed9348, 0xc0014e5d70}, ...}, ...}, ...)
	github.com/osmosis-labs/osmosis/v15/x/twap/keeper.go:55 +0x2c5
github.com/osmosis-labs/osmosis/v15/x/twap/twapmodule.AppModule.InitGenesis({{}, {{0x2ed92f8, 0xc0014e5c60}, 0xc0014e5d80, {{0x2f09580, 0xc00122c370}, 0xc000211a60, {0x2ed92f8, 0xc0014e5bb0}, {0x2ed9348, ...}, ...}, ...}}, ...)
	github.com/osmosis-labs/osmosis/v15/x/twap/twapmodule/module.go:115 +0x105
[...]

PR #5129, opened 2023-05-08
fix(x/twap): genesis issue blocking edge net set up

p0mvn:
Closes: #5041

More context: https://osmosis-network.slack.com/archives/C027ELA290B/p1682692206195569

Currently, our validation for `TwapRecord`s in `InitGenesis()` is stricter than it should be.

If there is a twap error, it expects both "p0" spot price and "p1" spot price to be zero.

However, that is not the guarantee from the core logic:
https://github.com/osmosis-labs/osmosis/blob/b101d06dc867400e4309877ccf559f09c89edda3/x/twap/logic.go#L53-L64

In the core logic above, if "p0" gets a twap error but "p1" does not, "p0" is set to zero but  "p1" isn't.

Therefore, we should change our init genesis logic to match the core's constraints.

## Brief Changelog

- make the change
- add tests

## Testing and Verifying

- Refactored test cases


czarcas7ic (approved):
LGTM

PR #5137, opened 2023-05-09
chore: rename FeeGrowthOutside to FeeGrowthOppositeDirectionOfLastTraversal

czarcas7ic:
Closes: #XXX

When Dev was reviewing CL, he pointed out that FeeGrowthOutside is a wrong name to give the field that tickInfo tracks. I proceeded to diagram what is happening behind the scenes, and came to the conclusion with the team that would would go for the more verbose `FeeGrowthOppositeDirectionOfLastTraversal`, as this accurately conveys what this number represents. The below charts explain the behavior of tick traversal and fee growth outside calculations:

TickUpdates
FeeGrowthOutsideCalculations

## Brief Changelog

- Rename tickInfo's `feeGrowthOutside` to `FeeGrowthOppositeDirectionOfLastTraversal`
- Update readme (also added charts)
- Replace any previous comment / function names referencing growth outside (for the tick info value only, as the concept of growth outside is now strictly used when speaking about a pair of ticks, which makes sense)


x/concentrated-liquidity/README.md
```diff
@@ -1216,43 +1226,91 @@ feeChargeTotal = amountIn.Mul(swapFee)
 
 ## Overview
 
-Due to the nonfungibility of positions and ticks, incentives for concentrated liquidity requires a slightly different mechanism for distributing incentives compared to Balancer and Stableswap pools. In general, the design space of incentive mechanisms for concentrated liquidity DEXs is extremely underexplored, so our implementation takes this as an opportunity to break some new ground in the broader design space of order-book-style AMMs.
+Due to the nonfungibility of positions and ticks, incentives for concentrated liquidity requires a
```
czarcas7ic, line 1229: Drive by lint, this was the only section that was not formatted in this way.

proto/osmosis/concentrated-liquidity/tickInfo.proto
```diff
@@ -22,7 +22,12 @@ message TickInfo {
     (gogoproto.moretags) = "yaml:\"liquidity_net\"",
     (gogoproto.nullable) = false
   ];
-  repeated cosmos.base.v1beta1.DecCoin fee_growth_outside = 3 [
+  // Total fees accumulated in the opposite direction that the tick was last
+  // crossed. i.e. if the tick was last crossed from left to right, then this is
+  // the total fees accumulated below the tick. If the tick was last crossed
+  // from right to left, then this is the total fees accumulated above the tick.
```
ValarDragon, line 28:
```suggestion
  // crossed. i.e. if the current tick is to the right of this tick (meaning its currently a greater price),
  // then this is the total fees accumulated below the tick. 
  // If the current tick is to the left of this tick (meaning its currently at a lower price), 
  // then this is the total fees accumulated above the tick.
```

ValarDragon, line 28:
```suggestion
  // crossed. i.e. if the current tick is to the right of this tick (meaning its currently a greater price),
  // then this is the total fees accumulated below the tick. 
  // If the current tick is to the left of this tick (meaning its currently at a lower price), 
  // then this is the total fees accumulated above the tick.
```

ValarDragon: What is the behavior when this tick is the active tick? This variable should just not be used?

x/concentrated-liquidity/README.md
czarcas7ic: The space bar, enter key, and determination

proto/osmosis/concentrated-liquidity/tickInfo.proto
czarcas7ic: It depends on what we are calculating. If we are calculating fee growth below and the lower tick is the active tick we use `feeGrowthGlobal` - `FeeGrowthOppositeDirectionOfLastTraversal`. If we are calculating fee growth above and the upper tick is active, we utilize the tick's `FeeGrowthOppositeDirectionOfLastTraversal` directly as the fee growth above. I will add this in the comments as its a very important point, thanks for pointing out :)

PR #5139, opened 2023-05-10
chore: expose pprof in localosmosis

p0mvn:
Closes: #XXX

Making the pprof server be one step away from being exposed in localosmosis.

The one remaining step is exposing a 6060 port in the Dockerfile. I left it unexported since we want to keep the minimal number of ports exposed on a production image for security purposes.

## Testing and Verifying


tests/localosmosis/README.md
```diff
@@ -270,3 +270,9 @@ services:
 ```
 
 2. Checkout the Osmosis repository to a different `ref` that includes the new version, and then rebuild and restart LocalOsmosis using `make localnet-start`. Make sure to don't delete your `~/.osmosisd-local` folder.
+
+## FAQ
+
+Q: How do I enable pprof server in localosmosis?
+
+A: everything but the Dockerfile is already configured. Since we use a production Dockerfile in localosmosis, we don't want to expose the pprof server there by default. As a result, if you would like to use pprof, make sure to add `EXPOSE 6060` to the Dockerfile and rebuild the localosmosis image.
```
czarcas7ic, line 278: What do you think about addition this to the production file but commenting it out so its even simpler to activate? Just a suggestion, not at all a blocker.

czarcas7ic (approved):

tests/localosmosis/README.md
p0mvn: Done

Issue #3849, opened 2022-12-23
x/gamm unit test coverage could be improved


dalmirel:
This Issue is created during the Informal Systems audit, after analysis of the existing specification and code inspection for balancer code performed on commit hash:[ 42d73f1](https://github.com/osmosis-labs/osmosis/commit/42d73f1cc1c52e85561518be1014b730ef6b7a12) and for stableswap code performed on: [v13.0.0](https://github.com/osmosis-labs/osmosis/commit/7374795e0de22f3a291ca59c5faffa7851acf3bd).
Unit Tests were also analyzed on [v13.1.0](https://github.com/osmosis-labs/osmosis/commit/bb867c557c41cb8a9ada1a441f72a3c95d768f13).

**Summary:**
Currently unit test coverage for balancer and stableswap code is around 70-80%,  but focusing on covering corner cases is of higher  importance than aiming at higher coverage.

We would suggest:
- adding some corner cases, such as e.g.
    0 <= swapFee < 1
    0 <= exitFee < 1
    2 <= len(liquidity) <= 8	// liquidity == poolAssets
    0 <  scaledFactor <= 9223372036854775807
    1 <= scaledAmount <= 10^34	// = liquidity[denom] / scalingFactors[denom]

- Adding corner cases or expected failure test cases for values from atomkraft findings. There is a significant number of test cases in pool_suite_test.go and  randomized tests TestRandomizedJoinPoolExitPoolInvariants, but specific test cases could be added here:
   - ExitPool - for values greater than in the pool…
   - SwapExactAmountOut, SwapExactAmountIn: swapping the exact amount of tokens as in the pool, >50% of the amount in the pool and larger than in the pool…

- Panic and errors are usually not covered - possibly a large number of them are unreachable. With current code implementation atomkraft managed to hit several panics - try covering those panic branches.
- Not all functions covered in their .test files or through other levels, e.g: Cover error branches for pool_service.go: validateCreatedPool and calcPoolSharesInGivenSingleAssetOut() ...
- There are several tests:
    - that are commented due to existing issues - that have been resolved in the meantime. The tests should pass, but they are still failing: [TestJoinSwapExactAmountInConsistency](https://github.com/osmosis-labs/osmosis/blob/7374795e0de22f3a291ca59c5faffa7851acf3bd/x/gamm/keeper/pool_service_test.go#L771-L783)
[...]

PR #5140, opened 2023-05-10
improve test x/gamm

DongLieu:
Closes: #3849

add corner cases and test cases as described

## Testing and Verifying

run test and lint


faddat:
@DongLieu is there an issue with the description?  code looks good to me, but we should link to the issue if there is one :)

x/gamm/keeper/pool_service_test.go
```diff
@@ -840,29 +842,46 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
 		shareOutMinAmount sdk.Int
 		expectedSharesOut sdk.Int
 		tokenOutMinAmount sdk.Int
+		expectError       error
 	}{
 		{
-			name:              "single coin with zero swap and exit fees",
+			name:              "happy path: single coin with zero swap and exit fees",
 			poolSwapFee:       sdk.ZeroDec(),
 			poolExitFee:       sdk.ZeroDec(),
```
pysel, line 850: I might be wrong, but as far as I am concerned, the exit fee was removed from osmosis pools via governance recently. I do not know if the governance's decision was already reflected on the protocol, but if yes, I think we should remove the `poolExitFee` field at all here

x/gamm/keeper/pool_service_test.go
```diff
@@ -840,29 +842,46 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
...
-		// 	expectedSharesOut: sdk.NewInt(6226484702880621000),
-		// 	tokenOutMinAmount: sdk.ZeroInt(),
-		// },
+		{
+			name:              "corner case: single coin with positive swap fee and zero exit fee",
+			poolSwapFee:       sdk.NewDecWithPrec(1, 2),
+			poolExitFee:       sdk.ZeroDec(),
+			tokensIn:          sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(1_000_000))),
+			shareOutMinAmount: sdk.ZeroInt(),
+			expectedSharesOut: sdk.NewInt(6226484702880621000),
```
pysel, line 862: I think it would be nice if we had a link to the calculations, which led to this number. Reference: https://github.com/osmosis-labs/osmosis/blob/main/osmomath/decimal_test.go#L724

pysel: It might be out of scope of this PR, but I think it is not a big deal if we add it in it, if people agree with the importance of it

x/gamm/keeper/pool_service_test.go
```diff
@@ -840,29 +842,46 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
 		shareOutMinAmount sdk.Int
 		expectedSharesOut sdk.Int
 		tokenOutMinAmount sdk.Int
+		expectError       error
```
pysel, line 845:
```suggestion
		expectedError       error
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -840,29 +842,46 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
...
+			name:              "corner case: single coin with positive swap fee and zero exit fee",
+			poolSwapFee:       sdk.NewDecWithPrec(1, 2),
+			poolExitFee:       sdk.ZeroDec(),
+			tokensIn:          sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(1_000_000))),
+			shareOutMinAmount: sdk.ZeroInt(),
+			expectedSharesOut: sdk.NewInt(6226484702880621000),
+			tokenOutMinAmount: sdk.ZeroInt(),
+		},
+		{
+			name:              "error: calculated amount is lesser than min amount",
```
pysel, line 866: nit:
```suggestion
			name:              "error: calculated amount is less than min amount",
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -840,29 +842,46 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
...
+		},
+		{
+			name:              "error: non zero exit fee",
+			poolSwapFee:       sdk.NewDecWithPrec(1, 2),
+			poolExitFee:       sdk.NewDecWithPrec(1, 2),
+			tokensIn:          sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(1_000_000))),
+			shareOutMinAmount: sdk.ZeroInt(),
+			expectedSharesOut: sdk.NewInt(6265857020099440400),
+			tokenOutMinAmount: sdk.ZeroInt(),
+			expectError:       fmt.Errorf("can not create pool with non zero exit fee, got %v", sdk.NewDecWithPrec(1, 2)),
```
pysel, line 883: nit:
```suggestion
			expectError:       fmt.Errorf("cannot create pool with non zero exit fee, got %v", sdk.NewDecWithPrec(1, 2)),
```

x/gamm/keeper/pool_service_test.go
```diff
@@ -891,32 +910,34 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
 					ExitFee: tc.poolExitFee,
 				},
 			)
+			if err != nil {
+				suite.Require().Equal(err, tc.expectError)
```
pysel, line 914: I personally think that the expected error in unit tests should reflect an error that might happen inside a function we are testing, which is, in this case, `JoinSwapExactAmountIn`. I think if we want to test edge cases for balancer pool creation, it should be done in a separate test

pysel: if, for any reason, something breaks for pool creation, when the expected error is related to `JoinSwapExactAmountIn`, this require statement will compare an error happened during pool creation with some internal to `JoinSwapExactAmountIn` edge case

x/gamm/keeper/pool_service_test.go
```diff
@@ -891,32 +910,34 @@ func (suite *KeeperTestSuite) TestJoinSwapExactAmountInConsistency() {
...
+					suite.Require().True(tokenOutAmt.LTE(swapFeeAdjustedAmount))
+				}
+			}
 
-			// require swapTokenOutAmt + 10 > input
-			suite.Require().True(
-				swapFeeAdjustedAmount.Sub(tokenOutAmt).LTE(sdk.NewInt(10)),
-				"expected out amount %s, actual out amount %s",
-				swapFeeAdjustedAmount, tokenOutAmt,
-			)
```
pysel, line 919: why are we removing this?

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -1095,4 +1095,36 @@ func (suite *KeeperTestSuite) TestRandomizedJoinPoolExitPoolInvariants() {
 	for i := 0; i < 50000; i++ {
 		testPoolInvariants()
 	}
+	testCaseCorners := []testCase{
+		{ //join pool for values greater than in the pool
```
pysel, line 1099: I think this file is not `gofmt`'ed

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -1095,4 +1095,36 @@ func (suite *KeeperTestSuite) TestRandomizedJoinPoolExitPoolInvariants() {
...
+			initialTokensDenomIn:  1,
+			initialTokensDenomOut: 9_223_372_036_854_775_807,
+			percentRatio:          101,
+		},
+		{ //with percentRatio = 0
+			initialTokensDenomIn:  1,
+			initialTokensDenomOut: 9_223_372_036_854_775_807,
+			percentRatio:          0,
+		},
+		{ //
```
pysel, line 1109:
```suggestion
		{
```

x/gamm/pool-models/balancer/pool_suite_test.go
```diff
@@ -1095,4 +1095,36 @@ func (suite *KeeperTestSuite) TestRandomizedJoinPoolExitPoolInvariants() {
...
+			initialTokensDenomIn:  9_223_372_036_854_775_807,
+			initialTokensDenomOut: 1,
+			percentRatio:          101,
+		},
+		{
+			initialTokensDenomIn:  9_223_372_036_854_775_807,
```
pysel, line 1115: asking to learn: is there a specific reason for this number? or does it just need to be big enough? if latter, big enough compared to what?

pysel (commented):
Thanks for the great work! Left some comments:

x/gamm/keeper/pool_service_test.go
DongLieu: I tried changing it but it doesn't seem to work, I don't think it's necessary to change it yet

x/gamm/keeper/pool_service_test.go
DongLieu: Thank you for checking

x/gamm/pool-models/balancer/pool_suite_test.go
DongLieu: it is the maximum number of type int64, I think the edge cases are the ones that lie at border

faddat (approved):

faddat:
this looks good @DongLieu, thanks!  I'm just going to merge main and make sure it passes golangci-lint with:

```yaml
run:
  tests: true
```

faddat (changes requested):
I'm pushing a commit that changes the linter to lint the tests, and updates to main.  Can you please handle the linting @DongLieu ?

Thanks!

faddat (approved):

DongLieu:
Ok, thank sir!

faddat:
@DongLieu -- theres a large block of comments that I've removed from pull_test.go -- nice work here!

faddat:
@DongLieu -- I tried to merge this, it didn't work.  I'll revert my last commit.  Can you get main merged please?

faddat:
Sir, I made several flubs here, and am force pushing to a recent commit of yours...

x/gamm/keeper/keeper_test.go
```diff
@@ -39,16 +39,14 @@ func (suite *KeeperTestSuite) prepareCustomBalancerPool(
...
 	suite.fundAllAccountsWith(balances)
 
 	poolID, err := suite.App.PoolManagerKeeper.CreatePool(
 		suite.Ctx,
 		balancer.NewMsgCreateBalancerPool(suite.TestAccs[0], poolParams, poolAssets, ""),
 	)
-	suite.Require().NoError(err)
-
-	return poolID
+	return poolID, err
```
hieuvubk, line 49: Why do we need to bring `suite.Require().NoError(err)` to outside of this func for any purpose?

hieuvubk:
There are many conflicts here @DongLieu . Could u resolve them?

x/gamm/keeper/keeper_test.go
DongLieu: In issue #3849 there is suggest with corner cases 0 <= exitFee < 1
however cannot create pool with exitFee > 0. So i want to get that error outside

Above pysel says that "I could be wrong, but as far as I know, the exit fee was recently removed from the osmosis pool through the administration. I don't know if the admin's decision has been reflected. reflected in the protocol or not, but if so, I think we should delete this field poolExitFee here".
And I removed testcase poolExitFee > 0

faddat:
thanks a lot!

PR #5143, opened 2023-05-10
check github actions with dependabot

faddat:

Dependabot can ensure that github actions aren't out of date.  I noticed several that are, but instead of manually fixing the actions, dependabot will do it faster and more reliably over time. 

## Brief Changelog

* use dependabot to check for updates to `github-actions`

## Testing and Verifying

If change was successful, we should see dependabot make several PR's after this is merged. 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,10 @@
 version: 2
 updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
 - package-ecosystem: gomod
   directory: "/"
   schedule:


p0mvn (approved):
Awesome, thanks for adding this

.github/dependabot.yml
```diff
@@ -1,5 +1,10 @@
 version: 2
 updates:
+- package-ecosystem: github-actions
+  directory: "/"
```
p0mvn, line 4: Is the directory correct here? Asking because our workflows are located in: https://github.com/osmosis-labs/osmosis/tree/main/.github/workflows

p0mvn: Yes, based on https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot, this is correct

PR #5144, opened 2023-05-10
ci: update rust code with dependabot when there are security updates

faddat:

This PR changes gomod dependency checking to daily for consistency, and adds cargo to dependabot so that we keep rust code up to date automatically too. 

## Brief Changelog

* change gomod frequency to daily
* check cargo for updates daily

## Testing and Verifying

If PR is successful dependabot will check daily.  Might be an inital burst of cargo package upgrades, but also, that might not happen. 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,11 +3,16 @@ updates:
 - package-ecosystem: gomod
   directory: "/"
   schedule:
-    interval: weekly
+    interval: daily
   ignore:
     - dependency-name: "github.com/tendermint/tendermint"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
   ignore:
     - dependency-name: "github.com/cosmos/ibc-go"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
+  open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
   open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,5 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  allow:
+    - security
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,7 @@ updates:
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
   open-pull-requests-limit: 10
 - package-ecosystem: cargo
-  directory: "/"
+  directory: "/cosmwasm"
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,11 +3,16 @@ updates:
 - package-ecosystem: gomod
   directory: "/"
   schedule:
-    interval: weekly
+    interval: daily
```
p0mvn, line 6: Personally, I would prefer to keep this as weekly to limit the daily noise we have to deal with

.github/dependabot.yml
```diff
@@ -3,11 +3,16 @@ updates:
...
     - dependency-name: "github.com/tendermint/tendermint"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
   ignore:
     - dependency-name: "github.com/cosmos/ibc-go"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
   open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
```
p0mvn, line 17: ditto

p0mvn (commented):
Thanks for adding, would like to discuss the frequency more before approving

.github/dependabot.yml
faddat: sure!

.github/dependabot.yml
```diff
@@ -11,3 +11,8 @@ updates:
     - dependency-name: "github.com/cosmos/ibc-go"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
   open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "/"
```
p0mvn, line 15: Should this instead be pointing to https://github.com/osmosis-labs/osmosis/tree/main/cosmwasm

since we don't seem to have any cargo files in the root of the repo?

.github/dependabot.yml
p0mvn: thanks

.github/dependabot.yml
faddat: maybe?  I think it will find all of them in the repo, but might as well?  

Not 100% sure actually

ValarDragon:
I feel like dependabot should be a pre-release phase thing for cosmwasm contracts, not continually applied. We don't want random bytecode changes for no reason here

faddat:
hey @ValarDragon I missed this comment before, but I think that you're right.  For now I am going to close this, and will reopen if I know how to do the exact thing you've described. 

oop, stackoverflow....

https://stackoverflow.com/questions/64047526/how-to-get-dependabot-to-trigger-for-security-updates-only

moment please :)

faddat:
@ValarDragon - this seems like it should be set up in the repo configuration:

* https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates

ValarDragon:
thank you!

ValarDragon:
Tried enabling it, lets see what happens.

If were overwhelmed with useless JS stuff for swagger, will close and may need to reconsider something

faddat:
Sounds good, thank you.

PR #5157, opened 2023-05-10
[CL][InternalReview] MsgAddToConcentratedLiquiditySuperfluidPosition

p0mvn:
Closes: #XXX

Summary: https://app.clickup.com/37420681/v/dc/13nzm9-19722/13nzm9-36742

No major issues found. Only renames and comments.

Had some questions around truncations of liquidity when minting shares for an SF lock, left comments directly in code to further discuss.

## Testing and Verifying


x/concentrated-liquidity/position.go
```diff
@@ -378,6 +378,8 @@ func (k Keeper) mintSharesLockAndUpdate(ctx sdk.Context, concentratedPoolId, pos
 	}
 
 	// Create a coin object to represent the underlying liquidity for the cl position.
+	// Q: what is the impact of this on slashing or anything else. I think truncating liqudiity even slightly might cause the final token amount difference
+	// to end up being quite large. I think the risk of this is offset but superfluid risk factor but would like to confirm.
 	underlyingLiquidityTokenized = sdk.NewCoins(sdk.NewCoin(types.GetConcentratedLockupDenomFromPoolId(concentratedPoolId), position.Liquidity.TruncateInt()))
```
p0mvn, line 383: Note: would like to confirm with reviewers

x/superfluid/README.md
```diff
@@ -376,6 +376,43 @@ It then mints concentrated liquidity shares and locks them up for the
 staking duration. From there, the normal superfluid delegation logic
 is executed.
 
+## Add To Superfluid Concentrated Position
+
+This message allows a user to add liquidity to a concentrated liquidity position.
```
p0mvn, line 381:
```suggestion
This message allows a user to add liquidity to a concentrated liquidity superfluid position.
```

x/concentrated-liquidity/position.go
```diff
@@ -392,7 +394,9 @@ func (k Keeper) mintSharesLockAndUpdate(ctx sdk.Context, concentratedPoolId, pos
 
 	// Lock the position for the specified duration.
 	// Note, the end blocker for the lockup module contains an exception for this CL denom. When a lock with a denom of cl/pool/{poolId} is mature,
-	// it does not send the coins to the owner account and instead burns them. This is strictly to use well tested pre-existing methods rather than potentially introducing bugs with new logic and methods.
+	// it does not send the coins to the owner account but instead burns them.
+	// This is implemented in such a way to use well-tested pre-existing methods rather than
+	// completely re-implementing concentrated liquidity superfluid infrastructure that has a riks of introducing bugs with new logic and methods.
```
p0mvn, line 399:
```suggestion
	// completely re-implementing concentrated liquidity superfluid infrastructure that has a risk of introducing bugs with new logic and methods.
```

x/superfluid/keeper/slash.go
```diff
@@ -109,6 +109,9 @@ func (k Keeper) prepareConcentratedLockForSlash(ctx sdk.Context, lock *lockuptyp
 	if err != nil {
 		return sdk.AccAddress{}, sdk.Coins{}, err
 	}
+
+	// Q: this would make the amounts calculated in alculateUnderlyingAssetsFromPosition
+	// be rounded down. Is this fine? I think yes, should be minimal impact.
 	slashAmtNeg := slashAmt.Neg()
```
p0mvn, line 115: Note: would like to discuss this as well

x/concentrated-liquidity/position.go
czarcas7ic: The only potential issue I can see with this is if we had a 100 percent risk factor, which I am fairly certain we will never do. 

Also, even if we did, it would result in the position getting very slightly under slashed, which I also do not think is innately dangerous.

Good call out though. Maybe we add a note to risk factor pointing to this on why we wouldn't want 1:1 ratio for superfluid minting.

x/superfluid/keeper/concentrated_liquidity.go
```diff
@@ -46,6 +48,11 @@ func (k Keeper) addToConcentratedLiquiditySuperfluidPosition(ctx sdk.Context, ow
 		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, types.PositionNotSuperfluidStakedError{PositionId: position.PositionId}
 	}
 
+	// Defense in depth making sure that the position is full-range.
+	if position.LowerTick != cltypes.MinTick || position.UpperTick != cltypes.MaxTick {
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, 0, types.ConcentratedTickRangeNotFullError{ActualLowerTick: position.LowerTick, ActualUpperTick: position.UpperTick}
+	}
+
```
czarcas7ic, line 55: nit: maybe we can extract this validation with the validation below this function into a helper function to make this easier to parse.

x/superfluid/keeper/slash.go
czarcas7ic: To clarify, I think your concern is that, since CalcUnderlyingAssetsFromPositions rounds the assets up, the slashAmt does not take this into consideration right?

This does not feel problematic to me, but would be good to note in the code comments.

x/superfluid/keeper/stake.go
```diff
@@ -300,10 +300,11 @@ func (k Keeper) SuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint
...
-// however it does not create a new synthetic lockup representing the unstaking side. This is because at the time this function is called, the new concentrated liquidity side
-// lock has not yet been created. Once the new cl side lock is created, the synthetic lockup representing the unstaking side is created.
-func (k Keeper) SuperfluidUndelegateToConcentratedPosition(ctx sdk.Context, sender string, gammLockID uint64) (types.SuperfluidIntermediaryAccount, error) {
+// SuperfluidForceUndelegate starts undelegating superfluid delegated position for the given lock. It behaves similarly to SuperfluidUndelegate,
+// however it does not create a new synthetic lockup representing the unstaking side. This is because after the time this function is called, we might
+// want to perform more operations prior to creating a lock. Once the actual lock is created, the synthetic lockup representing the unstaking side
+// should eventually be created as well. Use thi function with caution to avoid accidentally missing synthetic lock creation.
+func (k Keeper) SuperfluidForceUndelegate(ctx sdk.Context, sender string, gammLockID uint64) (types.SuperfluidIntermediaryAccount, error) {
```
czarcas7ic, line 307: I do not agree with this new name, since it behaves identical to SuperfluidUndelegate but just does not create the unlocking synth lock at the end (there is no force aspect to it). Can we rename to something that better encapsulates this? Maybe SuperfluidUndelegateWithoutNewSynthLock or something along these lines?

czarcas7ic (approved):
Nice work!

czarcas7ic (approved):
Nice work!

x/superfluid/keeper/stake.go
p0mvn: Reverted

PR #5160, opened 2023-05-11
[CL Message Audit]: MsgUnlockAndMigrateSharesToFullRangeConcentratedPosition [2/2]

czarcas7ic:
Clickup Checklist:
https://app.clickup.com/37420681/v/dc/13nzm9-19570/13nzm9-36390

This PR includes changes made from the internal audit of `MsgUnlockAndMigrateSharesToFullRangeConcentratedPosition`.

This is the refactor portion of the audit.

## Brief Changelog

- Refactors the routing function:
  - We used to do validation and then route to the respective migration function
  - We now route first and then do validation within the migration function itself
    - This is to prevent issues in the future in case these migration functions are use directly and not through the router


x/superfluid/keeper/migrate.go
```diff
@@ -10,29 +10,30 @@ import (
...
 // If the lock is locked or unlocking but not superfluid delegated/undelegating, it will migrate the position and either start unlocking or continue unlocking where it left off.
 // Errors if the lock is not found, if the lock is not a balancer pool lock, or if the lock is not owned by the sender.
 func (k Keeper) RouteLockedBalancerToConcentratedMigration(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin, tokenOutMins sdk.Coins) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, poolIdLeaving, poolIdEntering, gammLockId, concentratedLockId uint64, err error) {
 	// Validate and retrieve pertinent data required for migration
-	poolIdLeaving, poolIdEntering, concentratedPool, preMigrationLock, remainingLockTime, synthLockBeforeMigration, isSuperfluidBonded, isSuperfluidUnbonding, err := k.prepareMigration(ctx, sender, lockId, sharesToMigrate)
+	synthLocksBeforeMigration, isSuperfluidBonded, isSuperfluidUnbonding, err := k.routeMigration(ctx, sender, lockId, sharesToMigrate)
```
p0mvn, line 23: Can we instead have this return an enum instead of multiple booleans?

p0mvn: High level idea:

```go
type Status int

const (
    Pending Status = iota
    Approved
    Rejected
    OnHold
)

func main() {
    var status Status = Approved

    switch status {
    case Pending:
        fmt.Println("Status: Pending")
    case Approved:
        fmt.Println("Status: Approved")
    case Rejected:
        fmt.Println("Status: Rejected")
    case OnHold:
        fmt.Println("Status: On Hold")
    }
}

```

p0mvn:
Would it be possible to separate the refactor from the rest of the nit / comment changes please?

I'm having a hard time telling what I should focus on reviewing in mode detail vs what can be a quick skim

czarcas7ic:
@p0mvn split it up so that this PR is the refactor and the following PR https://github.com/osmosis-labs/osmosis/pull/5178 is just the nits / test additions / etc

x/superfluid/keeper/migrate.go
czarcas7ic: Good suggestion on this, was way cleaner the way you suggested. Added here https://github.com/osmosis-labs/osmosis/pull/5160/commits/9fb6fcdeeae231b72bb4a597e07a4a58f931cbba

x/superfluid/keeper/migrate.go
```diff
@@ -318,14 +316,9 @@ func (k Keeper) prepareMigration(ctx sdk.Context, sender sdk.AccAddress, lockId
...
 // 5. Ensures that exactly two coins are returned.
-func (k Keeper) validateSharesToMigrateUnlockAndExitBalancerPool(ctx sdk.Context, sender sdk.AccAddress, poolIdLeaving uint64, lock *lockuptypes.PeriodLock, sharesToMigrate sdk.Coin, tokenOutMins sdk.Coins) (exitCoins sdk.Coins, err error) {
-	// Finish unlocking directly for locked or unlocking locks
-	// This also breaks and deletes associated synthetic locks.
-	err = k.lk.ForceUnlock(ctx, *lock)
-	if err != nil {
-		return sdk.Coins{}, err
-	}
-
+// 6. Any remaining shares that were not migrated are re-locked for the remaining time on the lock.
```
mattverse, line 319:
```suggestion
// 6. Any remaining shares that were not migrated are re-locked as a new lock for the remaining time on the lock.
```

x/superfluid/keeper/migrate.go
```diff
@@ -155,160 +156,157 @@ func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context
...
+
+	for _, synthLockBeforeMigration := range synthLocksBeforeMigration {
+		if strings.Contains(synthLockBeforeMigration.SynthDenom, "superbonding") {
+			migrationType = SuperfluidBonded
+		}
+		if strings.Contains(synthLockBeforeMigration.SynthDenom, "superunbonding") {
+			migrationType = SuperfluidUnbonding
+		}
+	}
+
```
mattverse, line 255: what happens if there are two synthlocks existing, one being superfluid bonded, one being unbonding?

Also curious if `synthLocksBeforeMigration` is sorted deterministically

x/superfluid/keeper/migrate.go
```diff
@@ -155,160 +156,157 @@ func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context
...
 	}
 
-	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, nil
+	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, poolIdLeaving, poolIdEntering, nil
 }
 
-// prepareMigration prepares for the migration of gamm LP tokens from the Balancer pool to the Concentrated pool. It performs the following steps:
+// routeMigration determines the status of the provided lock which is used to determine the method for migration.
+// It also returns the underlying synthetic locks of the provided lock, if any exist.
+func (k Keeper) routeMigration(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin) (synthLocksBeforeMigration []lockuptypes.SyntheticLock, migrationType MigrationType, err error) {
```
mattverse, line 243: Can we have unit test for this method?

x/superfluid/keeper/migrate.go
```diff
@@ -10,102 +10,102 @@ import (
...
-		newLock, err := k.lk.CreateLock(ctx, sender, sdk.NewCoins(remainingGammShares), remainingLockTime)
-		gammLockId = newLock.ID
-		if err != nil {
-			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, err
-		}
-
-		// If the gamm lock was previously superfluid bonded, superfluid delegate the gamm like normal
-		err = k.SuperfluidDelegate(ctx, sender.String(), gammLockId, valAddr)
+	// If there are any remaining gamm shares after the migration, we must re-superfluid delegate them as they were previously in the gamm pool.
+	if remainingSharesLock.ID != 0 {
```
mattverse, line 99: I'm wondering if we don't need to return the new lock id, considering the fact that we have had some integrators requesting messages to return lock id for the partial unlock msg

x/superfluid/keeper/migrate_test.go
```diff
@@ -452,21 +401,18 @@ func (suite *KeeperTestSuite) TestMigrateSuperfluidUnbondingBalancerToConcentrat
 			suite.Require().NoError(err)
 
 			// Run slashing logic and check if the new and old locks are slashed.
-			suite.SlashAndValidateResult(ctx, newGammLockId, concentratedLockId, poolIdEntering, tc.percentOfSharesToMigrate, valAddr, *balancerLock, true)
+			suite.SlashAndValidateResult(ctx, newGammLockId, concentratedLockId, clPoolId, tc.percentOfSharesToMigrate, valAddr, *balancerLock, true)
```
mattverse, line 404: ditto here and below, do you think it's possible to add more checks and validations for partial unbonding logic?

x/superfluid/keeper/migrate.go
czarcas7ic: Each path is tested in each individual method test:

https://github.com/osmosis-labs/osmosis/blob/058fcc43f9ae8872a97aa2350f9aa58d9c3861b9/x/superfluid/keeper/migrate_test.go#L244-L246

https://github.com/osmosis-labs/osmosis/blob/058fcc43f9ae8872a97aa2350f9aa58d9c3861b9/x/superfluid/keeper/migrate_test.go#L363-L366

https://github.com/osmosis-labs/osmosis/blob/058fcc43f9ae8872a97aa2350f9aa58d9c3861b9/x/superfluid/keeper/migrate_test.go#L454-L458

x/superfluid/keeper/migrate.go
czarcas7ic: Sorry I am not following, why would we not want to return the new lock id?

x/superfluid/keeper/migrate.go
czarcas7ic: In my audit, I have this listed as an unknown. In what instance does a single lock have both unbonding and bonded synth denoms? Right now, this just errors and is unsupported. But I want to know if this is a normal occurrence

mattverse: Is it the case that this errors right now? At a brief glance, within this function, it seems that while iterating `synthLocksBeforeMigration`, if the length of `synthLocksBeforeMigration` is more than 1, `migrationType` would be able to continuously change. 

I have checked the code to find out in which case a single lock could have both unbonding and bonded synth denoms, but could not find an answer for that. Regardless, wdyt of adding a check here just to make sure length of `GetAllSyntheticLockupsByLockup` is not more than 1?

x/superfluid/keeper/migrate.go
mattverse: Ah I see, thanks for the clarification!

x/superfluid/keeper/migrate_test.go
mattverse: since #5162  also touches same logic, will request this test case if needed over there

mattverse (approved):
LGTM once synth lock length issue has been  addressed || answered

x/superfluid/keeper/migrate.go
czarcas7ic: After DMs, agreed, my bad. Fixed here https://github.com/osmosis-labs/osmosis/pull/5160/commits/478aa100de7c3bea8d57a4514ef2b3a9744d909f

x/superfluid/keeper/migrate.go
```diff
@@ -10,102 +10,102 @@ import (
...
-	remainingGammShares := gammSharesInLock.Sub(sharesToMigrate)
-	if !remainingGammShares.IsZero() {
-		// Create a new lock with the remaining gamm shares for the remaining lock time.
-		newLock, err := k.lk.CreateLock(ctx, sender, sdk.NewCoins(remainingGammShares), remainingLockTime)
-		gammLockId = newLock.ID
+	// If there are any remaining gamm shares after the migration, we must re-superfluid delegate them as they were previously in the gamm pool.
+	if remainingSharesLock.ID != 0 {
+		gammLockId = remainingSharesLock.ID
+		// Superfluid delegate the gamm lock.
+		err = k.SuperfluidDelegate(ctx, sender.String(), remainingSharesLock.ID, valAddr)
```
p0mvn, line 102: nit: made me wonder why we do the assignment but still use `remainingSharesLock.ID`
```suggestion
		gammLockId = remainingSharesLock.ID
		// Superfluid delegate the gamm lock.
		err = k.SuperfluidDelegate(ctx, sender.String(), gammLockId, valAddr)
```

x/superfluid/keeper/migrate.go
```diff
@@ -155,157 +156,157 @@ func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context
...
+	// Ensure a governance sanctioned link exists between the balancer pool and a concentrated pool.
 	poolIdEntering, err = k.gk.GetLinkedConcentratedPoolID(ctx, poolIdLeaving)
 	if err != nil {
-		return 0, 0, nil, &lockuptypes.PeriodLock{}, 0, nil, false, false, err
+		return 0, 0, &lockuptypes.PeriodLock{}, 0, err
 	}
 
-	// Get the concentrated pool from the provided ID and type cast it to ConcentratedPoolExtension.
-	concentratedPool, err = k.clk.GetConcentratedPoolById(ctx, poolIdEntering)
+	// Further defense in depth, ensuring that the pool ID we are entering can be type cased to a concentrated pool extension.
```
p0mvn, line 294:
```suggestion
	// Further defense in depth, ensuring that the pool ID we are entering can be type casted to a concentrated pool extension.
```

p0mvn (approved):

x/superfluid/keeper/migrate.go
czarcas7ic: This no longer exists since we implemented partial superfluid undelegation :)

PR #5162, opened 2023-05-11
feat: partial superfluid undelegate

czarcas7ic:
Closes: #XXX

After running via localosmosis, I realized that the UnlockAndMigrate message was taking a very large amount of gas (~1.4million). When breaking the message apart, I realized that if we do a partial migration, we could save a lot of gas by retaining the old lock and removing assets from it rather than forceUnlocking the entire lock and migrating to a new lock (having to re-establish superfluid connections, etc).  This change bring the gas used to ~1million. We should investigate any further changes that can be made to reduce this overhead.

This is achieved by:
- Not sending the cl/pool/x funds to the user just to have them sent back to the lockup module
- Allow for a partial superfluid undelegate, so we dont have to delete the underlying lock and connection just to re-establish them (this happens when we do a partial migration)  

## Brief Changelog

*(for example:)*
 

x/lockup/keeper/lock.go
```diff
@@ -135,6 +135,30 @@ func (k Keeper) CreateLock(ctx sdk.Context, owner sdk.AccAddress, coins sdk.Coin
 	return lock, nil
 }
 
+// CreateLockNoSend behaves the same as CreateLock, but does not send the coins to the lockup module account.
+func (k Keeper) CreateLockNoSend(ctx sdk.Context, owner sdk.AccAddress, coins sdk.Coins, duration time.Duration) (types.PeriodLock, error) {
```
p0mvn, line 139: Would it be possible to move bank send to `CreateLock`, make everything no send by default and from `CreateLock` call `CreateLockNoSend` and then perform the bank send?

x/superfluid/keeper/migrate.go
```diff
@@ -329,37 +287,30 @@ func (k Keeper) validateSharesToMigrateUnlockAndExitBalancerPool(ctx sdk.Context
...
-		return sdk.Coins{}, lockuptypes.PeriodLock{}, err
+	if sharesToMigrate.Equal(gammSharesInLock) {
+		// If migrating the entire lock:
+
+		// This breaks and deletes associated synthetic locks.
+		err = k.lk.ForceUnlock(ctx, *lock)
+		if err != nil {
+			return sdk.Coins{}, err
+		}
 	}
```
p0mvn, line 302: checking: has the direct unit test been updated to account for this being conditional?

x/superfluid/keeper/migrate.go
```diff
@@ -329,37 +287,30 @@ func (k Keeper) validateSharesToMigrateUnlockAndExitBalancerPool(ctx sdk.Context
...
-
 	// Finish unlocking directly for locked or unlocking locks
-	// This also breaks and deletes associated synthetic locks.
-	err = k.lk.ForceUnlock(ctx, *lock)
-	if err != nil {
-		return sdk.Coins{}, lockuptypes.PeriodLock{}, err
+	if sharesToMigrate.Equal(gammSharesInLock) {
+		// If migrating the entire lock:
+
+		// This breaks and deletes associated synthetic locks.
```
p0mvn, line 297:
```suggestion
		// If migrating the entire lock, force unlock.
		// This breaks and deletes associated synthetic locks.
```

x/superfluid/keeper/migrate_test.go
```diff
@@ -153,34 +166,66 @@ func (suite *KeeperTestSuite) TestRouteLockedBalancerToConcentratedMigration() {
...
+					suite.Require().Equal(balancerIntermediaryAcc.GetAccAddress().String(), addr.String())
+
+					// Check if migration deleted synthetic lockup.
+					_, err = lockupKeeper.GetSyntheticLockup(ctx, originalGammLockId, keeper.StakingSyntheticDenom(balancerLock.Coins[0].Denom, valAddr.String()))
+					suite.Require().NoError(err)
+
+					// The delegation from the intermediary account holder should still exist.
+					_, found := stakingKeeper.GetDelegation(ctx, balancerIntermediaryAcc.GetAccAddress(), valAddr)
+					suite.Require().True(found, "expected delegation, found delegation no delegation")
+				}
```
mattverse, line 206: Can we do deeper checks here for partial unlocks?
Some of the checks that comes into my mind are 
- Check if superfluid delegated amount is existing amount - coinsToMigrate
- Check if what's remaining in the lock is existing amount - coinsToMigrate
- Check if the new superfluid staked amount is equal to coinsToMigrate

and so on

x/superfluid/keeper/migrate_test.go
```diff
@@ -377,7 +437,7 @@ func (suite *KeeperTestSuite) TestMigrateSuperfluidUnbondingBalancerToConcentrat
 			}
 
 			// System under test.
-			positionId, amount0, amount1, _, _, newGammLockId, concentratedLockId, poolIdLeaving, poolIdEntering, err := superfluidKeeper.MigrateSuperfluidUnbondingBalancerToConcentrated(ctx, poolJoinAcc, originalGammLockId, coinsToMigrate, synthLockBeforeMigration[0].SynthDenom, tc.tokenOutMins)
+			positionId, amount0, amount1, _, _, concentratedLockId, poolIdLeaving, poolIdEntering, err := superfluidKeeper.MigrateSuperfluidUnbondingBalancerToConcentrated(ctx, poolJoinAcc, originalGammLockId, coinsToMigrate, synthLockBeforeMigration[0].SynthDenom, tc.tokenOutMins)
 			if tc.expectedError != nil {
```
mattverse, line 441: ditto here for additional checks

x/superfluid/keeper/stake.go
```diff
@@ -287,6 +287,48 @@ func (k Keeper) undelegateCommon(ctx sdk.Context, sender string, lockID uint64)
 	return intermediaryAcc, nil
 }
 
+// partialSuperfluidUndelegate acts similarly to undelegateCommon, but undelegates a partial amount of the lock's delegation rather than the full amount. The amount
+// that is undelegated is placed in a new lock. This function returns the intermediary account associated with the original lock ID as well as the new lock that was created.
+// An error is returned if the amount to undelegate is greater than the locked amount.
+func (k Keeper) partialSuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint64, amountToUndelegate sdk.Coin) (types.SuperfluidIntermediaryAccount, *lockuptypes.PeriodLock, error) {
```
mattverse, line 293: consider returning named arguments to avoid confusion

x/superfluid/keeper/stake_test.go
```diff
@@ -1028,6 +1028,154 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+					suite.Require().Error(err)
+					continue
+				}
+				suite.Require().NoError(err)
+
+				// ensure post-superfluid delegations osmo supplywithoffset is the same while supply is not
+				postsupply := suite.App.BankKeeper.GetSupply(suite.Ctx, bondDenom)
+				postsupplyWithOffset := suite.App.BankKeeper.GetSupplyWithOffset(suite.Ctx, bondDenom)
+				suite.Require().False(postsupply.IsEqual(presupply), "presupply: %s   postsupply: %s", presupply, postsupply)
+				suite.Require().True(postsupplyWithOffset.IsEqual(presupplyWithOffset))
```
mattverse, line 1141: Can we do a check with the supply amount using the amount we unlocked instead of simply comparing if equal or not?

x/superfluid/keeper/stake_test.go
```diff
@@ -1028,6 +1028,154 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+				if err != nil {
+					lock = &lockuptypes.PeriodLock{}
+				}
+
+				// get pre-superfluid delgations osmo supply and supplyWithOffset
+				presupply := suite.App.BankKeeper.GetSupply(suite.Ctx, bondDenom)
+				presupplyWithOffset := suite.App.BankKeeper.GetSupplyWithOffset(suite.Ctx, bondDenom)
+
+				// superfluid undelegate
+				intermediaryAcc, _, err = suite.App.SuperfluidKeeper.PartialSuperfluidUndelegate(suite.Ctx, lock.Owner, lockId, tc.undelegateAmounts[index])
```
mattverse, line 1130: can we do additional checks (e.g amount) with the new lock that has been created and being returned?

mattverse (commented):
Main logic lgtm, requesting for some additions in the test cases!

x/lockup/keeper/lock.go
```diff
@@ -163,6 +187,28 @@ func (k Keeper) lock(ctx sdk.Context, lock types.PeriodLock, tokensToLock sdk.Co
 	return nil
 }
 
+// lockNoSend behaves the same as lock, but does not send the coins to the lockup module account.
+func (k Keeper) lockNoSend(ctx sdk.Context, lock types.PeriodLock, tokensToLock sdk.Coins) error {
```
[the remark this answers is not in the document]
czarcas7ic: I realized the best way was to make a common function that they both call. It wouldn't have been feasible to call lockNoSend directly from lock because we want to ensure the send happens before the hook is called https://github.com/osmosis-labs/osmosis/pull/5162/commits/749b82d857dd1149122ac2bc410d104a21da17fd

x/lockup/keeper/lock.go
czarcas7ic: This is a great suggestion, implementing now

x/lockup/keeper/lock.go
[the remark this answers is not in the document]
czarcas7ic: The following commit undoes what was done above in favor of your suggestion of making lockNoSend the default behavior https://github.com/osmosis-labs/osmosis/pull/5162/commits/d783cd5bcb795694c0dbbdf339b08604af12a8fb

x/lockup/keeper/lock.go
czarcas7ic: lockNoSend (with a warning about expectation from caller) done here https://github.com/osmosis-labs/osmosis/pull/5162/commits/d783cd5bcb795694c0dbbdf339b08604af12a8fb

Note: I think its important to use the lockNoSend nomenclature in place of lock since the expectation of the previous lock function involved sending the coins.

x/superfluid/keeper/migrate.go
czarcas7ic: This question led to uncovering a bug! We actually should have been doing a partialForceUnlock if the above branch is not hit. I also expanded the tests here to ensure this will get found in the future https://github.com/osmosis-labs/osmosis/pull/5162/commits/c7fd2bb93602bb9233c07a71123fd9b5031f473e

x/lockup/keeper/lock.go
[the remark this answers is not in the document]
czarcas7ic: unit test added for CreateLockNoSend (lock was replaced by lockNoSend along with its unit test as per Roman's suggestion) https://github.com/osmosis-labs/osmosis/pull/5162/commits/eb8b67d7db9b8195fe91a8f9a6e54e9af228519c

x/superfluid/keeper/stake_test.go
czarcas7ic: Done here https://github.com/osmosis-labs/osmosis/pull/5162/commits/0d1e2c59dc33aa6b49fb8f20d699fe28ce8d8148

x/superfluid/keeper/stake.go
czarcas7ic: Done here https://github.com/osmosis-labs/osmosis/pull/5162/commits/0d1e2c59dc33aa6b49fb8f20d699fe28ce8d8148

x/superfluid/keeper/migrate_test.go
czarcas7ic: Expanded these checks here, good call for this :) https://github.com/osmosis-labs/osmosis/pull/5162/commits/d71ad6938e87343a8d5ec65aefc88cbaa7044d4e

x/superfluid/keeper/migrate_test.go
czarcas7ic: I expanded checks for the route and delegating. For the undelegating checks though, what should be expanded? Since, with superfluid, once something is undelegating, it is instantly undelegated. So if something is in an undelegating status prior to migration, I am not sure what extra things we should check here.

czarcas7ic: Checked the exact amounts here https://github.com/osmosis-labs/osmosis/pull/5162/commits/c20ddedc31c5d790d0baee6d6df3fc8314f163a3

x/superfluid/keeper/stake_test.go
czarcas7ic: Checked the new lock here https://github.com/osmosis-labs/osmosis/pull/5162/commits/c20ddedc31c5d790d0baee6d6df3fc8314f163a3

x/superfluid/keeper/stake_test.go
czarcas7ic: Checked the exact amounts here https://github.com/osmosis-labs/osmosis/pull/5162/commits/c20ddedc31c5d790d0baee6d6df3fc8314f163a3

x/superfluid/keeper/migrate.go
```diff
@@ -57,55 +57,58 @@ func (k Keeper) migrateSuperfluidBondedBalancerToConcentrated(ctx sdk.Context,
...
-	// Superfluid undelegate the superfluid delegated position.
-	// This deletes the connection between the gamm lock and the intermediate account, deletes the synthetic lock, and burns the synthetic osmo.
-	intermediateAccount, err := k.superfluidUndelegateToConcentratedPosition(ctx, sender.String(), lockId)
-	if err != nil {
-		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
+	// Superfluid undelegate the portion of shares the user is migrating from the superfluid delegated position.
+	// If all shares are being migrated, this deletes the connection between the gamm lock and the intermediate account, deletes the synthetic lock, and burns the synthetic osmo.
+	intermediateAccount := types.SuperfluidIntermediaryAccount{}
+	if partialMigration {
+		intAccount, splitLock, err := k.partialSuperfluidUndelegate(ctx, sender.String(), lockId, sharesToMigrate)
```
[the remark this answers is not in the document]
czarcas7ic: lol I have no idea why it was like this, must have happened while I was troubleshooting something and never changed it back, reverted here https://github.com/osmosis-labs/osmosis/pull/5162/commits/91911c878ee38758adbeef4b9653fc87def2e6e6

x/superfluid/keeper/stake.go
```diff
@@ -287,6 +287,48 @@ func (k Keeper) undelegateCommon(ctx sdk.Context, sender string, lockID uint64)
...
+	}
+	err = k.forceUndelegateAndBurnOsmoTokens(ctx, amount, intermediaryAcc)
+	if err != nil {
+		return types.SuperfluidIntermediaryAccount{}, &lockuptypes.PeriodLock{}, err
+	}
+
+	// Move the funds from the old lock to a new lock with the remaining amount.
+	newLock, err := k.lk.SplitLock(ctx, *lock, sdk.NewCoins(amountToUndelegate), true)
+	if err != nil {
+		return types.SuperfluidIntermediaryAccount{}, &lockuptypes.PeriodLock{}, err
```
mattverse, line 326: coming to think about it, do we not need logic to handle the lock after calling `forceUndelegateAndBurnOsmoTokens`? Things that come to my mind are
- lock id and intermediary account connection
- handle  synthethic lock for the undelegating lock
- create synthlock for the unbonding amounts

x/superfluid/keeper/stake_test.go
```diff
@@ -1028,6 +1028,158 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+
+				// check bonding synthetic lockup is not deleted
+				_, err = suite.App.LockupKeeper.GetSyntheticLockup(suite.Ctx, lockId, keeper.StakingSyntheticDenom(lock.Coins[0].Denom, valAddr))
+				suite.Require().NoError(err)
+
+				// check unbonding synthetic lockup creation
+				// since this is the concentrated liquidity path, no new synthetic lockup should be created
+				synthLock, err := suite.App.LockupKeeper.GetSyntheticLockup(suite.Ctx, lockId, keeper.UnstakingSyntheticDenom(lock.Coins[0].Denom, valAddr))
+				suite.Require().Error(err)
+				suite.Require().Nil(synthLock)
```
mattverse, line 1159: How about the new lock? My assumption was that the new lock created should have unbonding synth lock, if this is the case, can you add test cases for this?

x/superfluid/keeper/stake.go
mattverse: Depending on the asnwer for this, could we have according checks added for each of ^?

x/superfluid/keeper/migrate.go
```diff
@@ -57,55 +57,55 @@ func (k Keeper) migrateSuperfluidBondedBalancerToConcentrated(ctx sdk.Context,
...
-		// Superfluid delegate the gamm lock.
-		err = k.SuperfluidDelegate(ctx, sender.String(), remainingSharesLock.ID, valAddr)
-		if err != nil {
-			return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, 0, err
-		}
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, err
 	}
 
-	return positionId, amount0, amount1, liquidity, joinTime, gammLockId, concentratedLockId, poolIdLeaving, poolIdEntering, nil
+	return positionId, amount0, amount1, liquidity, joinTime, concentratedLockId, poolIdLeaving, poolIdEntering, nil
```
mattverse, line 108: I was wondering if we should be returning the new lock (or the old one) if the lock has been splitted and a new one was created with a new id

x/superfluid/keeper/migrate_test.go
mattverse: If partial unlocking, is there a way we can check that the amount to undelegate is succesfully in the queue of undelegation?

mattverse: Can we also add checks for the new lock that has been created if this was a partial unlock? Apologies if I'm being picky here, but since we're touching core staking logic, I thought we might as well want to be as safe as possible

mattverse (commented):
Thanks for adding so much test cases, leaving second round of reivew

x/superfluid/keeper/migrate_test.go
czarcas7ic: If partial unlocking, there is no queue of undelegation. With superfluid, when we undelegate, the delegation amount is immediately removed from the validator and instantly unbonded

czarcas7ic: Added a check to bonded, unbonding, and normal lock tests that check the newly created lock. Please let me know if I should expand! https://github.com/osmosis-labs/osmosis/pull/5162/commits/1aa1f055621fe24ecfc0a91e6c7bd218a7236ac8

x/superfluid/keeper/stake_test.go
```diff
@@ -1028,6 +1028,158 @@ func (suite *KeeperTestSuite) TestRefreshIntermediaryDelegationAmounts() {
...
+			for index, lockId := range tc.superUnbondingLockIds {
+				// get intermediary account
+				accAddr := suite.App.SuperfluidKeeper.GetLockIdIntermediaryAccountConnection(suite.Ctx, lockId)
+				intermediaryAcc := suite.App.SuperfluidKeeper.GetIntermediaryAccount(suite.Ctx, accAddr)
+				valAddr := intermediaryAcc.ValAddr
+
+				lock, err := suite.App.LockupKeeper.GetLockByID(suite.Ctx, lockId)
+				if err != nil {
+					lock = &lockuptypes.PeriodLock{}
+				}
```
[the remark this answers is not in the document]
czarcas7ic: This is specifically for the "undelegating not available lock id" test case. We don't want it to fail early and we want to attempt to partial superfluid undelegate with a non existing lock rather than catching it early. I can add a logic branch specific for this test case though.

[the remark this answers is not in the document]
czarcas7ic: Added logic branch here https://github.com/osmosis-labs/osmosis/pull/5162/commits/ff187c6ac84650a315b8f6dd765c62298e4f020c

x/superfluid/keeper/stake.go
czarcas7ic: I am not sure which side of the action you are referring to.

If you are referring to the remaining amount in the lock that still exists:
- the lockID and intermediary account connection remains the same
- the synthetic lock for that lock also remains the same
- the underlying lock of the synthetic lock is reduced, so that handles the synth lock portion

If you are referring to the new amount the got put into a lock that got migrated:
- the lockID and intermediary account connection and synthetic lock is made in the migration function itself (see https://github.com/osmosis-labs/osmosis/blob/efb468ce574e910214a6cdad93d237b4ad786a77/x/superfluid/keeper/migrate.go#L150-L163)

czarcas7ic: This is a good catch, but I think the logic of this function needs to stay as it is for the following reason:

We don't to create all those connections just to remove those connections so we can migrate it. That means this function is to be used by CL migration only. That being said, two things need to be done:

- This function should be renamed and a warning given that connections are not made after the lock is split
- I can abstract this into a helper function to a real partialSuperfluidUndelegate function that DOES create the connections, but just not call that function and instead call the helper directly for this

Again, nice call on this!

czarcas7ic: So ended up making a shared function, I think you will agree with the implementation https://github.com/osmosis-labs/osmosis/pull/5162/commits/9a911853bb8b881df39a7577be321ff8d05a52d5

(still using the same logic as before, but if someone wants to make a partialSuperfluidUndelegate message in the future, they wont accidentally use the wrong method)

x/superfluid/keeper/stake_test.go
czarcas7ic: Now that I split the underlying partial unbond function apart this should make more sense now. We don't expect the new lock to have an unbonding synth lock associated with it. Expanded the test case here to show this https://github.com/osmosis-labs/osmosis/pull/5162/commits/83e672594f333ce231f2cf59ab24063813d24b4f

x/superfluid/keeper/migrate.go
czarcas7ic: No, its better to imagine it like this:

I have initial lock 1. I want to migrate a portion of it. A portion of lock 1 is removed from lock 1 and creates lock 2. lock 2 only exists so we can exit the pool with those tokens. Then, lock 3 is created that represents the concentratedLockID.

So if we start with lock 1 and we migrate a portion of it, we will still have lock 1. If we start with lock 1 and we migrate the whole thing, lock 1 just simply wont exist anymore.

Hopefully this clears it up!

czarcas7ic:
@mattverse addressed all comments, thanks for the thorough review!

x/superfluid/keeper/migrate.go
mattverse: I see, so the claim is that this is intended, as we call exit pool using the returned lock and then call superfluid delegate again anyway!

thanks for clearing this up, understood

mattverse (approved):
LGTM

x/lockup/keeper/lock_test.go
```diff
@@ -533,6 +533,60 @@ func (suite *KeeperTestSuite) TestCreateLock() {
 	suite.Require().Equal(sdk.NewInt(30), balance.Amount)
 }
 
+func (suite *KeeperTestSuite) TestCreateLockNoSend() {
```
AlpinYukseloglu, line 536: Is there a particular reason you made these non table driven?

x/superfluid/keeper/migrate_test.go
```diff
@@ -53,15 +53,28 @@ func (suite *KeeperTestSuite) TestRouteLockedBalancerToConcentratedMigration() {
 		"lock that is superfluid delegated, not unlocking (partial shares)": {
 			// migrateSuperfluidBondedBalancerToConcentrated
 			superfluidDelegated:      true,
-			percentOfSharesToMigrate: sdk.MustNewDecFromStr("0.5"),
+			percentOfSharesToMigrate: sdk.MustNewDecFromStr("0.4"),
```
AlpinYukseloglu, line 56: Out of curiosity, what motivated this change?

AlpinYukseloglu (commented):
The logic looks right, but still hesitant to approve personally since this is for part of the codebase I'm learning as I review. We should definitely revisit this flow during package-level review to ensure the holistic CL superfluid flow is airtight

x/superfluid/keeper/migrate_test.go
czarcas7ic: Migrating exactly half made a nice number, and I wanted to ensure it wasn't just happenstance that the expected numbers aligned

x/lockup/keeper/lock_test.go
czarcas7ic: Yes, the pre-existing `TestCreateLock` was not table driven, and I wanted to use the exact logic flow to ensure this method behaves just like `CreateLock`, just without the send

p0mvn (approved):
LGTM. Nice work. Going to address my comments that aren't likely to be disputed and will ping on anything remaining

x/superfluid/keeper/migrate.go
```diff
@@ -27,24 +27,24 @@ const (
...
 // If the lock is locked or unlocking but not superfluid delegated/undelegating, it will migrate the position and either start unlocking or continue unlocking where it left off.
 // Errors if the lock is not found, if the lock is not a balancer pool lock, or if the lock is not owned by the sender.
-func (k Keeper) RouteLockedBalancerToConcentratedMigration(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin, tokenOutMins sdk.Coins) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, poolIdLeaving, poolIdEntering, gammLockId, concentratedLockId uint64, err error) {
+func (k Keeper) RouteLockedBalancerToConcentratedMigration(ctx sdk.Context, sender sdk.AccAddress, lockId uint64, sharesToMigrate sdk.Coin, tokenOutMins sdk.Coins) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, poolIdLeaving, poolIdEntering, concentratedLockId uint64, err error) {
 	synthLocksBeforeMigration, migrationType, err := k.routeMigration(ctx, sender, lockId, sharesToMigrate)
 	if err != nil {
```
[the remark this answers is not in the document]
p0mvn: This will have to be done in the other PR that contains latest state

p0mvn:
Addressed all nits, leaving the rest @czarcas7ic

x/lockup/keeper/lock.go
```diff
@@ -116,11 +122,28 @@ func (k Keeper) AddTokensToLockByID(ctx sdk.Context, lockID uint64, owner sdk.Ac
...
+
+	// Run the createLock logic without the send since we sent the coins above.
+	lock, err := k.CreateLockNoSend(ctx, owner, coins, duration)
+	if err != nil {
+		return types.PeriodLock{}, err
+	}
+	return lock, nil
+}
+
+// CreateLockNoSend behaves the same as CreateLock, but does not send the coins to the lockup module account.
```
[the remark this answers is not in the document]
czarcas7ic: Comment expanded here https://github.com/osmosis-labs/osmosis/pull/5162/commits/c59c387781b5728c92ebad80cd166c3f736b9ba3

x/lockup/keeper/lock.go
```diff
@@ -135,18 +158,17 @@ func (k Keeper) CreateLock(ctx sdk.Context, owner sdk.AccAddress, coins sdk.Coin
 	return lock, nil
 }
 
-// lock is an internal utility to lock coins and set corresponding states.
+// lockNoSend is an internal utility to lock coins and set corresponding states.
 // This is only called by either of the two possible entry points to lock tokens.
 // 1. CreateLock
 // 2. AddTokensToLockByID
```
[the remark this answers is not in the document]
czarcas7ic: renamed from lockNoSend back to lock (with warning in godoc) https://github.com/osmosis-labs/osmosis/pull/5162/commits/d1c7b6f57ac87c15d4a887d92ea70507fe314c1b

x/superfluid/keeper/migrate.go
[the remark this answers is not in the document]
czarcas7ic: I still am not entirely comfortable with my understanding of this. Why does GetAllSyntheticLockupsByLockup return an array in the first place? Is it normal for a lockId to have multiple synthetic locks? Getting this answered I think is really important before this gets swept under the rug

x/superfluid/keeper/stake.go
```diff
@@ -303,11 +303,75 @@ func (k Keeper) SuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint
...
+
+	return intermediaryAcc, &newLock, nil
+}
+
+// partialSuperfluidUndelegate starts undelegating a portion of a superfluid delegated position for the given lock.
+// Undelegation is done instantly and the equivalent amount is sent to the module account
+// where it is burnt. Note that this method does not include unbonding the lock
+// itself.
+// nolint: unused
+func (k Keeper) partialSuperfluidUndelegate(ctx sdk.Context, sender string, lockID uint64, amountToUndelegate sdk.Coin) error {
```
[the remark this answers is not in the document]
czarcas7ic: I basically did the work to allow partial superfluid undelegating for regular superfluid positions (which has been a desired feature), but its out of scope for the release so I thought I would leave it here for future release / testing. I can comment the function out if that is desired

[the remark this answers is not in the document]
czarcas7ic: Same point above. Will action once we converge on what we want to do

x/superfluid/keeper/migrate.go
```diff
@@ -57,55 +57,55 @@ func (k Keeper) migrateSuperfluidBondedBalancerToConcentrated(ctx sdk.Context,
 	sharesToMigrate sdk.Coin,
 	synthDenomBeforeMigration string,
 	tokenOutMins sdk.Coins,
-) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, gammLockId, concentratedLockId, poolIdLeaving, poolIdEntering uint64, err error) {
+) (positionId uint64, amount0, amount1 sdk.Int, liquidity sdk.Dec, joinTime time.Time, concentratedLockId, poolIdLeaving, poolIdEntering uint64, err error) {
```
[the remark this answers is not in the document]
czarcas7ic: Comment expanded here https://github.com/osmosis-labs/osmosis/pull/5162/commits/315e8a024e3bfbf32072baed4f0a2e2fec596f0d

x/superfluid/keeper/migrate.go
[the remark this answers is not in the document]
p0mvn: I don't know with 100% confidence but I'm not aware of a specific use case where we have multiple lock-ups per lock up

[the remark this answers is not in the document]
czarcas7ic: Discussed with Roman offline, created an issue for this and will address once this PR and the PR it is merging into is merged to main

https://github.com/osmosis-labs/osmosis/issues/5246

PR #5167, opened 2023-05-12
[CL Message Audit]: AddToPositions 

mattverse:
Closes: #XXX

Not much notable things to audit since this message is simply a combination of withdrawPosition + createPosition.

The only notable things and changes I have made were:
- Noticed that the existing test for `AddToPositions` were invalid, were erroring out due to different reasons. Went back and revised all test cases
- Add min amount 0 and 1 for the message so that the user can specify the min amounts


AlpinYukseloglu (commented):
Logic changes seem okay but need to take another look at the changes made to test cases (which at a first glance seem drastic). Leaving current comments in the meantime

x/concentrated-liquidity/lp_test.go
```diff
@@ -851,53 +840,48 @@ func (s *KeeperTestSuite) TestAddToPosition() {
...
-			lastPositionInPool: true,
-			timeElapsed:        defaultTimeElapsed,
-			amount0ToAdd:       amount0PerfectRatio,
-			amount1ToAdd:       amount1PerfectRatio.Neg(),
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: amount0PerfectRatio,
+			amount1ToAdd: amount1PerfectRatio,
 		},
-		"error: attempt to add negative amounts for both assets to position": {
+		"error: minimum amount 1 is less than actual amount": {
```
AlpinYukseloglu, line 873: Should these min checks not be separate tests? I don't see why we would remove/replace the negative input error cases (lmk if I'm missing something here)

x/concentrated-liquidity/lp_test.go
```diff
@@ -744,21 +743,16 @@ func (s *KeeperTestSuite) TestAddToPosition() {
 			amount1ToAdd: amount1PerfectRatio.QuoRaw(2),
 		},
 
-		// Error catching
-
-		"error: attempt to add to a position with underlying lock that has finished unlocking": {
+		"Add to a position with underlying lock that has finished unlocking": {
```
AlpinYukseloglu, line 746: What was the change that made this no longer error?

x/concentrated-liquidity/types/msgs.go
```diff
@@ -87,12 +87,17 @@ func (msg MsgAddToPosition) ValidateBasic() error {
 		return fmt.Errorf("Invalid position id (%s)", strconv.FormatUint(msg.PositionId, 10))
 	}
 
-	if !msg.TokenDesired0.IsValid() || !msg.TokenDesired0.IsPositive() {
-		return fmt.Errorf("Invalid coins (%s)", msg.TokenDesired0.String())
+	if msg.Amount0.IsNegative() {
```
AlpinYukseloglu, line 90: Is there a reason you removed the validity check here too?

x/concentrated-liquidity/types/msgs.go
```diff
@@ -87,12 +87,17 @@ func (msg MsgAddToPosition) ValidateBasic() error {
...
+		return fmt.Errorf("Amount 0 cannot be negative, given amount: %s", msg.Amount0.String())
 	}
-
-	if !msg.TokenDesired1.IsValid() || !msg.TokenDesired1.IsPositive() {
-		return fmt.Errorf("Invalid coins (%s)", msg.TokenDesired1.String())
+	if msg.Amount1.IsNegative() {
```
AlpinYukseloglu, line 93: Same as above

x/concentrated-liquidity/types/msgs.go
```diff
@@ -87,12 +87,17 @@ func (msg MsgAddToPosition) ValidateBasic() error {
...
+		return fmt.Errorf("Amount 1 cannot be negative, given amount: %s", msg.Amount1.String())
+	}
+	if msg.TokenMinAmount0.IsNegative() {
+		return fmt.Errorf("Amount 0 cannot be negative, given token min amount: %s", msg.TokenMinAmount0.String())
+	}
+	if msg.TokenMinAmount1.IsNegative() {
+		return fmt.Errorf("Amount 0 cannot be negative, given amount: %s", msg.TokenMinAmount1.String())
```
AlpinYukseloglu, line 100:
```suggestion
		return fmt.Errorf("Amount 1 cannot be negative, given amount: %s", msg.TokenMinAmount1.String())
```

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -75,12 +80,34 @@ message MsgCreatePositionResponse {
 message MsgAddToPosition {
   uint64 position_id = 1 [ (gogoproto.moretags) = "yaml:\"position_id\"" ];
   string sender = 2 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
-  cosmos.base.v1beta1.Coin token_desired0 = 3 [
-    (gogoproto.moretags) = "yaml:\"token_desired0\"",
+  // amount0 represents the amount of token0 willing to put in.
+  string amount0 = 3 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"token_min_amount0\"",
```
AlpinYukseloglu, line 86:
```suggestion
    (gogoproto.moretags) = "yaml:\"amount_0\"",
```

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -75,12 +80,34 @@ message MsgCreatePositionResponse {
...
+    (gogoproto.nullable) = false
+  ];
+  // amount1 represents the amount of token1 willing to put in.
+  string amount1 = 4 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"token_min_amount1\"",
```
AlpinYukseloglu, line 92:
```suggestion
    (gogoproto.moretags) = "yaml:\"amount1\"",
```

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -75,12 +80,34 @@ message MsgCreatePositionResponse {
 message MsgAddToPosition {
   uint64 position_id = 1 [ (gogoproto.moretags) = "yaml:\"position_id\"" ];
   string sender = 2 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
-  cosmos.base.v1beta1.Coin token_desired0 = 3 [
-    (gogoproto.moretags) = "yaml:\"token_desired0\"",
+  // amount0 represents the amount of token0 willing to put in.
+  string amount0 = 3 [
```
AlpinYukseloglu, line 84: What was the motivation for changing this from coin to amount? (I believe createPosition does coins for desired input & int for min output)

x/concentrated-liquidity/lp_test.go
```diff
@@ -978,15 +967,23 @@ func (s *KeeperTestSuite) TestAddToPosition() {
...
+			// delta amount1 only exists if the actual amount from addToPosition is not equivilent to tokens provided.
+			// delta amount1 is calculated via (amount1 to create initial position) + (amount1 added to position) - (actual amount 1)
+			if fundCoins.AmountOf(pool.GetToken1()).Add(tc.amount1ToAdd).Sub(newAmt1).GT(sdk.ZeroInt()) {
+				expectedAmount1Delta = config.tokensProvided.AmountOf(pool.GetToken1()).Add(tc.amount1ToAdd).Sub(newAmt1)
+			}
+
+			postBalanceToken0 := s.App.BankKeeper.GetBalance(s.Ctx, sender, pool.GetToken0())
+			postBalanceToken1 := s.App.BankKeeper.GetBalance(s.Ctx, sender, pool.GetToken1())
+
+			var errTolerance osmomath.ErrTolerance
```
AlpinYukseloglu, line 981: Just to be sure: which rounding error is this meant to accommodate? I noticed in the test cases you added the withdraw amounts are rounded down in the expected values, so I'm assuming this is for the added values?

x/concentrated-liquidity/lp.go
```diff
@@ -298,7 +302,16 @@ func (k Keeper) addToPosition(ctx sdk.Context, owner sdk.AccAddress, positionId
 		return 0, sdk.Int{}, sdk.Int{}, err
 	}
 	tokensProvided := sdk.NewCoins(sdk.NewCoin(pool.GetToken0(), amount0Desired), sdk.NewCoin(pool.GetToken1(), amount1Desired))
-	newPositionId, actualAmount0, actualAmount1, _, _, err := k.createPosition(ctx, position.PoolId, owner, tokensProvided, amount0Withdrawn, amount1Withdrawn, position.LowerTick, position.UpperTick)
+	minimumAmount0 := amount0Withdrawn
+	minimumAmount1 := amount1Withdrawn
+
+	if !amount0MinGiven.IsZero() {
+		minimumAmount0 = amount0MinGiven
```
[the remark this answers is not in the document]
czarcas7ic: I was just typing this until I saw your comment. Agreed viewing this in terms of the entire amount rather than just the amount we are adding is not intuitive.

x/concentrated-liquidity/lp_test.go
```diff
@@ -744,21 +743,16 @@ func (s *KeeperTestSuite) TestAddToPosition() {
...
-				// Since we round on the other the asset when we withdraw, asset0 turns into the bottleneck and
-				// thus we cannot use the full amount of asset1. We calculate the below using the following formula and rounding up:
-				// amount1 = L * (sqrtPriceUpper - sqrtPriceLower)
-				// https://www.wolframalpha.com/input?i=3035764327.860030912175533748+*+%2870.710678118654752440+-+67.416615162732695594%29
-				amount1Expected: sdk.NewInt(9999998816),
-				expectedError:   types.PositionSuperfluidStakedError{PositionId: uint64(1)},
+				// 1998976eth (amount withdrawn with rounded down amounts) + 998977(token amount in)
+				amount0Expected: sdk.NewInt(2997953),
+				// tokens Provided for token1 is 9999999999 (amount withdrawn) + 5000000000 = 14999999999usdc, then liquidity needed is calculated internally
+				amount1Expected: sdk.NewInt(14997436189),
```
czarcas7ic, line 755: Can you show the math that results in this value?

x/concentrated-liquidity/lp_test.go
```diff
@@ -851,53 +840,48 @@ func (s *KeeperTestSuite) TestAddToPosition() {
...
 			},
-			lastPositionInPool: true,
-			timeElapsed:        defaultTimeElapsed,
-			amount0ToAdd:       amount0PerfectRatio,
-			amount1ToAdd:       amount1PerfectRatio.Neg(),
+			timeElapsed:  defaultTimeElapsed,
+			amount0ToAdd: amount0PerfectRatio,
+			amount1ToAdd: amount1PerfectRatio,
 		},
-		"error: attempt to add negative amounts for both assets to position": {
```
czarcas7ic, line 880: I think we still need tests for:
- negative amt0
- negative amt1
- zero for both amt0 and amt1

x/concentrated-liquidity/lp_test.go
```diff
@@ -960,8 +945,12 @@ func (s *KeeperTestSuite) TestAddToPosition() {
 				sender = invalidSender
 			}
 
+			// now we fund the sender account again for the amount0ToAdd and amount1ToAdd coins.
+			s.FundAcc(sender, sdk.NewCoins(sdk.NewCoin(ETH, tc.amount0ToAdd), sdk.NewCoin(USDC, tc.amount1ToAdd)))
+
 			// --- System under test ---
-			newPosId, newAmt0, newAmt1, err := concentratedLiquidityKeeper.AddToPosition(s.Ctx, sender, config.positionId, tc.amount0ToAdd, tc.amount1ToAdd)
+			newPosId, newAmt0, newAmt1, err := concentratedLiquidityKeeper.AddToPosition(s.Ctx, sender, config.positionId, tc.amount0ToAdd, tc.amount1ToAdd, config.amount0Minimum, config.amount1Minimum)
+			// config.amount0Minimum
```
czarcas7ic, line 953: I think we need to check that the old position is deleted when this is successful right?

x/concentrated-liquidity/types/msgs.go
czarcas7ic: Agreed on keeping validity check

x/concentrated-liquidity/types/msgs.go
czarcas7ic: Can we keep the same verbiage as well? "given token min amount" instead of "given amount"

proto/osmosis/concentrated-liquidity/tx.proto
mattverse: The motivation for this was that we do not use the denom given with the coin anyways, there is no need to take in coins

x/concentrated-liquidity/lp.go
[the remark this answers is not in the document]
mattverse: Sorry I dont quite follow, isn't @czarcas7ic 's claim and @AlpinYukseloglu 's claim contradicting? Maybe I'm misunderstanding..

So currently the code is viewing the given (amount min given) in terms of the entire amount, where we directly use what ever given from the user as the min amount for the new position being created.

x/concentrated-liquidity/types/msgs.go
mattverse: The reason validity check was removed was because we have changed this field from coins to int, thus there is nothing to call validity check anymore, lmk if this makes sense

x/concentrated-liquidity/types/msgs.go
mattverse: fixed

x/concentrated-liquidity/lp_test.go
mattverse: So this was the test case that made me revise all the test cases actually. 
My first approach was that I noticed that this case should not error, where user should be able to add to a position with underlying lock that has finished unlocking. Then I fmt.Println along the lines for this test code, and noticed that this was not erroring due to the error type we have specified, but was erroring due to the account not having enough funds. 

Funding the account prior to calling add to positions fixed this

x/concentrated-liquidity/lp_test.go
mattverse: Added wolfram alpha equation for this

x/concentrated-liquidity/lp_test.go
mattverse: yes this is for the rounding errors caused in the process of withdrawing

x/concentrated-liquidity/lp_test.go
mattverse: good catch, added this in

x/concentrated-liquidity/lp_test.go
mattverse: Good call on this, the original reason I took these out was because it would error when preparing environment for the testcase, but found a way out and added in

x/concentrated-liquidity/lp_test.go
mattverse: Same as above, added in

x/concentrated-liquidity/lp.go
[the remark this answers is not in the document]
p0mvn: This is breaking quite a few tests, going to make an issue for this to discuss / address separately

proto/osmosis/concentrated-liquidity/tx.proto
```diff
@@ -80,12 +85,34 @@ message MsgCreatePositionResponse {
...
+    (gogoproto.nullable) = false
+  ];
+  // token_min_amount1 represents the minimum amount of token1 desired from the
+  // new position being created. Note that this field indicates the min amount1
+  // corresponding to the total liquidity of the position, not just the
+  // liquidity that is being added.
+  string token_min_amount1 = 6 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.moretags) = "yaml:\"token_min_amount1\"",
     (gogoproto.nullable) = false
```
p0mvn, line 116: question: why is this useful in the context of add to position?

x/concentrated-liquidity/lp.go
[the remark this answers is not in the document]
p0mvn: https://github.com/osmosis-labs/osmosis/issues/5238

p0mvn (approved):
This LGTM

Previous reviews have been pretty thorough and comments are now addressed.

Made one follow-up issue here: https://github.com/osmosis-labs/osmosis/issues/5238

PR #5170, opened 2023-05-12
CL: benchmark for swaps

p0mvn:
Closes: #XXX

This PR adds a bench test for swap exact amount in.

This benchmark has already helped uncover some performance concerns and tick-to-price relationship issues when adding liquidity.

The benchmark works by creating a configurable number of randomized positions in the direction of the swap.
It also allows adding more full-range positions or positions concentrated around current square root price for deeper liquidity.

TBD preliminary results

  - [ ] Code comments?
  - [x] N/A


x/concentrated-liquidity/lp.go
```diff
@@ -65,6 +65,10 @@ func (k Keeper) createPosition(ctx sdk.Context, poolId uint64, owner sdk.AccAddr
 		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, err
 	}
 
+	if sqrtPriceLowerTick.Equal(sqrtPriceUpperTick) {
+		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, errors.New("sqrtPriceLowerTick and sqrtPriceUpperTick cannot be equal")
+	}
```
p0mvn, line 70: Note: I had to make this change to make the benchmark work

Keeping it with the understanding that @czarcas7ic 's change in a separate PR will fully address and test this

x/concentrated-liquidity/math/math.go
```diff
@@ -44,6 +46,10 @@ func Liquidity1(amount sdk.Int, sqrtPriceA, sqrtPriceB sdk.Dec) sdk.Dec {
 	sqrtPriceBBigDec := osmomath.BigDecFromSDKDec(sqrtPriceB)
 
 	diff := sqrtPriceBBigDec.Sub(sqrtPriceABigDec)
+	if diff.Equal(osmomath.ZeroDec()) {
+		panic(fmt.Sprintf("diff is zero: sqrtPriceA %s sqrtPriceB %s", sqrtPriceA, sqrtPriceB))
+	}
```
p0mvn, line 51: Note: makes the error more descriptive. Since this should never happen in-practice, I think we should keep this as is instead of returning error

ValarDragon (approved):
Thanks for getting the benchmark in!

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (s BenchTestSuite) createPosition(accountIndex int, poolId uint64, coin0, coin1 sdk.Coin, lowerTick, upperTick int64) {
+	tokensDesired := sdk.NewCoins(coin0, coin1)
+
+	_, _, _, _, _, err := s.App.ConcentratedLiquidityKeeper.CreatePosition(s.Ctx, poolId, s.TestAccs[accountIndex], tokensDesired, sdk.ZeroInt(), sdk.ZeroInt(), lowerTick, upperTick)
+	if err != nil {
+		// This can happen for ticks that map to the very small prices
+		// e.g 2 * 10^(-18) ends up mapping to the same sqrt price
+		fmt.Println("error creating position", err)
+	}
```
czarcas7ic, line 31: Can we branch this off of https://github.com/osmosis-labs/osmosis/pull/5173 so we can remove this?

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+		// Seed controlling determinism of the randomized positions.
+		seed = int64(1)
+	)
+	rand.Seed(seed)
+
+	for i := 0; i < b.N; i++ {
+		s := BenchTestSuite{}
+		s.Setup()
+
+		// Full all accounts with max amounts they would need to consume.
```
czarcas7ic, line 88:
```suggestion
		// Fund all accounts with max amounts they would need to consume.
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+		for _, acc := range s.TestAccs {
+			simapp.FundAccount(s.App.BankKeeper, s.Ctx, acc, sdk.NewCoins(sdk.NewCoin(denom0, maxAmountOfEachToken), sdk.NewCoin(denom1, maxAmountOfEachToken), sdk.NewCoin("uosmo", maxAmountOfEachToken)))
+		}
+
+		// Craete a pool
```
czarcas7ic, line 93:
```suggestion
		// Create a pool
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		// Setup numberOfPositions full range positions for deeper liquidity.
+		if shouldCreateFullRangePositions {
+			for i := 0; i < numberOfPositions; i++ {
+				lowerTick := types.MinTick
+				upperTick := types.MaxTick
+
+				tokenDesired0 := sdk.NewCoin(denom0, sdk.NewInt(maxAmountDeposited).MulRaw(5))
+				tokenDesired1 := sdk.NewCoin(denom1, sdk.NewInt(maxAmountDeposited).MulRaw(5))
```
czarcas7ic, line 156: I am confused on why we define a maxAmountDeposited as the maximum we can deposit per position, and then we create a single position with this var by multiplying it by five. Maybe we should redefine it?

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+				accountIndex := rand.Intn(len(s.TestAccs))
+
+				account := s.TestAccs[accountIndex]
+
+				simapp.FundAccount(s.App.BankKeeper, s.Ctx, account, tokensDesired)
+
+				s.createPosition(accountIndex, poolId, tokenDesired0, tokenDesired1, lowerTick, upperTick)
+			}
+		}
```
czarcas7ic, line 207: Imo we should make this expandable by taking into consideration the tick spacing. So for instance, tick spacing is 10 so the first for loop creates a concentrated position between x-10 and x +10 and the second for loop creates positions from x- 100 to x + 100 (as it currently does). But if the tick spacing is 100, the first for loop should concentrate at x- 100 and x + 100 and the second for loop should concentrate at x - 1000 and x + 1000 (or something like this)

x/concentrated-liquidity/lp.go
czarcas7ic: If possible, would be great to branch this off mine to just remove this entirely. But if we are time constrained this can be done later.

czarcas7ic (approved):
Nice work, comments are all suggestions

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+				// minTick <= lowerTick <= currentTick
+				lowerTick = rand.Int63n(currentTick.Int64()-types.MinTick+1) + types.MinTick
+				// Normalize lowerTick to be a multiple of tickSpacing
+				lowerTick = lowerTick + (tickSpacing - lowerTick%tickSpacing)
+				// lowerTick <= upperTick <= currentTick
+				upperTick = currentTick.Int64() - rand.Int63n(int64(math.Abs(float64(currentTick.Int64()-lowerTick))))
+			} else {
+				// Increasing price so want to be above current tick
+
+				// currentTick <= lowerTick <= upperTick
```
mattverse, line 131:
```suggestion
				// currentTick <= lowerTick <= maxTick
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+				upperTick int64
+			)
+
+			if denomIn == denom0 {
+				// Decreasing price so want to be below current tick
+
+				// minTick <= lowerTick <= currentTick
+				lowerTick = rand.Int63n(currentTick.Int64()-types.MinTick+1) + types.MinTick
+				// Normalize lowerTick to be a multiple of tickSpacing
+				lowerTick = lowerTick + (tickSpacing - lowerTick%tickSpacing)
```
mattverse, line 125:
```suggestion
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+				// lowerTick <= upperTick <= currentTick
+				upperTick = currentTick.Int64() - rand.Int63n(int64(math.Abs(float64(currentTick.Int64()-lowerTick))))
+			} else {
+				// Increasing price so want to be above current tick
+
+				// currentTick <= lowerTick <= upperTick
+				lowerTick := rand.Int63n(types.MaxTick-currentTick.Int64()+1) + currentTick.Int64()
+				// Normalize lowerTick to be a multiple of tickSpacing
+				lowerTick = lowerTick + (tickSpacing - lowerTick%tickSpacing)
```
mattverse, line 134:
```suggestion
```

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+				// lowerTick <= upperTick <= maxTick
+				upperTick = types.MaxTick - rand.Int63n(int64(math.Abs(float64(types.MaxTick-lowerTick))))
+			}
+			// Normalize upperTick to be a multiple of tickSpacing
```
mattverse, line 138:
```suggestion
			// Normalize lowerTick to be a multiple of tickSpacing
			lowerTick = lowerTick + (tickSpacing - lowerTick%tickSpacing)
			// Normalize upperTick to be a multiple of tickSpacing
```

mattverse: nit, but simply for code sanity

x/concentrated-liquidity/bench_test.go
mattverse: ^upvote for this comment

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+				tokenDesired1 := sdk.NewCoin(denom1, sdk.NewInt(maxAmountDeposited).MulRaw(5))
+				tokensDesired := sdk.NewCoins(tokenDesired0, tokenDesired1)
+
+				accountIndex := rand.Intn(len(s.TestAccs))
+
+				account := s.TestAccs[accountIndex]
+
+				simapp.FundAccount(s.App.BankKeeper, s.Ctx, account, tokensDesired)
+
+				s.createPosition(accountIndex, poolId, tokenDesired0, tokenDesired1, lowerTick, upperTick)
```
mattverse, line 205: Do we not need to normalize the lower tick and upper tick based on tick spacing here?

x/concentrated-liquidity/bench_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+		swapAmountIn := sdk.MustNewDecFromStr(amountIn).TruncateInt()
+		largeSwapInCoin := sdk.NewCoin(denomIn, swapAmountIn)
+
+		liquidityNet, err := clKeeper.GetTickLiquidityNetInDirection(s.Ctx, pool.GetId(), largeSwapInCoin.Denom, currentTick, sdk.Int{})
+		noError(err)
+
+		fmt.Println("num_ticks_traversed", len(liquidityNet))
+		fmt.Println("current_tick", currentTick)
+
```
mattverse, line 217: nit: consider printing amount of total positions created

x/concentrated-liquidity/math/math.go
mattverse: Agreed, can we address it in this PR or make an issue so that we make sure we come back to it?

mattverse (approved):
nice work! lgtm

mattverse (approved):
nice work! lgtm

x/concentrated-liquidity/bench_test.go
p0mvn: This is  incorrect

p0mvn: I'm wrong. This is correct. I thought we must normalize prior to initializing upper tick but it doesn't matter. Normalization can happen fter

x/concentrated-liquidity/bench_test.go
p0mvn: It's configured by the user so this is assumed to be known when running the test

x/concentrated-liquidity/bench_test.go
p0mvn: This can't be removed because this would still lead to an error

x/concentrated-liquidity/bench_test.go
p0mvn: changed to `maxAmountDepositedFullRange`

x/concentrated-liquidity/bench_test.go
p0mvn: I changed this to account for tick spacing

PR #5171, opened 2023-05-12
Speedup some of the swap logic by doing more pre-computation in tick math

ValarDragon:

Using Roman's CL benchmark as a guide, this PR provides a slight speedup to the CL math. (~8%)

I suggest we completely make `tickIndex` an `int64`, and `TickExponent` an int64, both of which will imo simplify the code, and together provide another 5% speedup.

## Testing and Verifying

  - [X] Code comments?

### DIFF
--- a/x/concentrated-liquidity/math/precompute.go
+++ b/x/concentrated-liquidity/math/precompute.go
@@ -1,0 +1,44 @@
+package math
+
+import (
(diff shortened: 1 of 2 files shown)


p0mvn (approved):
Awesome! LGTM

PR #5174, opened 2023-05-13
Speedup accumulator prefix keys

ValarDragon:

We were spending a lot of time in accumulator key formatting, for basically doing sprintf's with constants. This PR fixes that.

It also highlights a bug in the docs before (that ideally should get fixed prelaunch)

Also converts one `Add` to an `AddMut`

## Testing and Verifying

  - [ ] Code comments?
  - [x] N/A

### DIFF
--- a/osmoutils/accum/prefix.go
+++ b/osmoutils/accum/prefix.go
@@ -8,23 +8,16 @@ const (
 	positionPrefix    = "pos"
 )
 
-// formatAccumPrefix returns the key prefix used for any
-// accum module values to be stored in the KVStore.
-// Returns "accum/{key}" as bytes.
-func formatModulePrefixKey(key string) []byte {
-	return []byte(fmt.Sprintf("%s/%s", modulePrefix, key))
-}
-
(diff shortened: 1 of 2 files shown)


osmoutils/accum/prefix.go
```diff
@@ -8,23 +8,16 @@ const (
...
 // Returns "accum/acc/{name}" as bytes.
 func formatAccumPrefixKey(name string) []byte {
-	return formatModulePrefixKey(fmt.Sprintf("%s/%s", accumulatorPrefix, name))
+	return []byte(fmt.Sprintf(modulePrefix+"/"+accumulatorPrefix+"/%s", name))
 }
 
 // FormatPositionPrefixKey returns the key prefix used
 // specifically for position values in the KVStore.
-// Returns "accum/pos/{accumName}/{name}" as bytes.
+// Returns "accum/acc/pos/{accumName}/{name}" as bytes.
```
ValarDragon, line 20: notice that the old code actually would add `acc` here, making the old comment wrong :)

p0mvn (approved):

PR #5176, opened 2023-05-13
refactor: remove cache context from swap out given in

p0mvn:
Closes: #XXX

[The benchmark ](https://github.com/osmosis-labs/osmosis/pull/5170) helped as see that having an additional cache layer in swaps would cause performance overhead.

This PR refactors swap out given in to remove an additional cache layer during swaps. It would still be present in calc (estimates) but not the core swaps logic. Calcs are not used by frontend or any other important clients so it is not as a big of a concern for launch.

Some graphs to show the improvement:

Old:

image

New:
image

Notice how the proportion of time spend in the iterator is noticeably smaller.

```
benchstat old.bench new.bench
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
                     │  old.bench   │           new.bench           │
                     │    sec/op    │   sec/op     vs base          │
SwapExactAmountIn-16   777.9m ± 23%   750.3m ± 9%  ~ (p=0.436 n=10)
```

The p value here shows that the improvement isn't statistically significant but I believe that it is wrong since I see an improvement every time I run the test. Also, variance is decreased quite a bit.

Swap in given out will require the same changes in a separate PR.

## Testing and Verifying

  - [ ] Code comments?
  - [x] N/A


ValarDragon (approved):
Nice! Were making swaps mutative always, and just calc non-mutative via a cacheCtx

This looks good to me :)

x/concentrated-liquidity/swaps_test.go
```diff
@@ -2331,28 +2337,81 @@ func (s *KeeperTestSuite) TestCalcOutAmtGivenInWriteCtx() {
...
 
+			// check that fee accum has been correctly updated upon writeCtx
 			feeAccum, err := s.App.ConcentratedLiquidityKeeper.GetFeeAccumulator(s.Ctx, 1)
 			s.Require().NoError(err)
 
 			feeAccumValue := feeAccum.GetValue()
-			s.Require().Equal(0, feeAccumValue.Len())
-			s.Require().Equal(1,
+			s.Require().Equal(1, feeAccumValue.Len())
+			s.Require().Equal(0,
```
ValarDragon, line 2346: I don't really get why this changed

mattverse (approved):
Awesome! 

Took some time for me to understand what was going on here, but I like the approach that we have a third message compute and then call that message with ctx and cachectx depending on whether the caller is swap method or compute method. 

I am also down to apply this approach to gamm calc and swap methods, if you do agree please lmk, we can create an issue for this

x/concentrated-liquidity/swaps_test.go
p0mvn: Previously, this was a non-mutative method. As a result, we expected no accumulator updates. It would only mutate upon `writeCtx` being called. Now, it is mutative so the update is expected immediately after the call

x/concentrated-liquidity/swaps_test.go
```diff
@@ -1484,11 +1484,17 @@ func (s *KeeperTestSuite) TestCalcAndSwapOutAmtGivenIn() {
...
-			// perform calc
-			_, tokenIn, tokenOut, updatedTick, updatedLiquidity, sqrtPrice, err := s.App.ConcentratedLiquidityKeeper.CalcOutAmtGivenInInternal(
-				s.Ctx,
+			// Refetch the pool
+			pool, err = s.App.ConcentratedLiquidityKeeper.GetPoolById(s.Ctx, pool.GetId())
+			s.Require().NoError(err)
+
+			// perform compute
+			cacheCtx, _ := s.Ctx.CacheContext()
+			tokenIn, tokenOut, updatedTick, updatedLiquidity, sqrtPrice, err := s.App.ConcentratedLiquidityKeeper.ComputeOutAmtGivenIn(
```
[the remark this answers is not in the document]
p0mvn: The test name had to be changed, good catch. Done

Issue #5186, opened 2023-05-15
[CL][Incentives Query][Bug]: Implement support for CL gauges to incentivized pools query


AlpinYukseloglu:
## Background

Our current `IncentivizedPools` query only searches through `lockableDurations` and thus misses CL pools in our default genesis since CL gauges are created on epoch time. While this would technically not be an issue on mainnet (since epoch time overlaps with a lockable duration), we should expand the incentivized pools query to capture CL pools.

## Suggested Design

- Add epoch duration to list looped through if it does not already exist

## Acceptance Criteria

- All new and existing tests pass

PR #5187, opened 2023-05-15
[CL][Incentives Query][Bug]: Implement support for CL gauges to incentivized pools query

AlpinYukseloglu:
Closes: #5186

This PR expands our incentivized pools query to be inclusive of incentivized CL pools.

## Testing and Verifying

- Changes are tested in `x/pool-incentives/keeper/grpc_query_test.go`

  - [x] Code comments?
  - [ ] N/A


x/pool-incentives/keeper/grpc_query.go
```diff
@@ -87,6 +87,23 @@ func (q Querier) IncentivizedPools(ctx context.Context, _ *types.QueryIncentiviz
...
+		if lockableDuration == epochDuration {
+			epochAlreadyLockable = true
+			break
+		}
+	}
+
+	// Ensure that we only add epoch duration if it does not already exist as a lockable duration.
+	if !epochAlreadyLockable {
+		lockableDurations = append(lockableDurations, epochDuration)
+	}
```
p0mvn, line 105: On mainnet, epoch duration is 1 day, and we also have 1-day uptime. How are we going to differentiate between internal per-epoch gauges and 1 day uptime gauges?

p0mvn (dismissed):

x/pool-incentives/keeper/grpc_query.go
mattverse, line 105:
```suggestion
	epochDuration := q.incentivesKeeper.GetEpochInfo(sdkCtx).Duration
	for _, lockableDuration := range lockableDurations {
		if lockableDuration == epochDuration {
			lockableDurations = append(lockableDurations, epochDuration)	
			break
		}
	}
```

mattverse (approved):

x/pool-incentives/keeper/grpc_query.go
p0mvn: FYI, from @jonator 

```
Yeah we need to make sure it only refers to pools and gauges that are internally incentivized via mint incentives

For external incentives, we handle those in the opposite direction: by querying the gauges and linking those to pools

(Which, a separate topic, causes a lot of performance issues)

Like in hindsight, we could have reduced complexity by having the node return all gauges for a pool, internal and external
```

AlpinYukseloglu: This change seems to add epoch time to lockable durations only if it already exists in there - we want the exact opposite no?

p0mvn (approved):

Issue #5175, opened 2023-05-13
refactor: convert tick math to operate on int64 (move away from sdk.Int)


p0mvn:
## Background

Context: for performance reasons.

More context: https://github.com/osmosis-labs/osmosis/pull/5171

## Acceptance Criteria
- moved to int64
- confirmed no overflows

ValarDragon:
Also I feel like the math code will look nicer (getting to use standard operators)

PR #5189, opened 2023-05-16
perf: convert ticks to int64

p0mvn:
Closes: #5175

Converts ticks to int64 for performance and readability

  - [ ] Code comments?
  - [x] N/A


x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
...
 
 // CalculatePriceToTick takes in a price and returns the corresponding tick index.
 // This function does not take into consideration tick spacing.
-func CalculatePriceToTick(price sdk.Dec) (tickIndex sdk.Int) {
+func CalculatePriceToTick(price sdk.Dec) (tickIndex int64) {
 	// The formula is as follows: geometricExponentIncrementDistanceInTicks = 9 * 10**(-exponentAtPriceOne)
 	// Due to sdk.Power restrictions, if the resulting power is negative, we take 9 * (1/10**exponentAtPriceOne)
 	exponentAtPriceOne := types.ExponentAtPriceOne
-	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne.Neg()))
+	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne * -1)).TruncateInt64()
```
mattverse, line 171: Is truncation safe here? If so, why is that?

mattverse (commented):
All the logic introduced here seems good to me.

However, it does approach to me that it might even be possible to store int64 instead of sdk.Int for the ticks, which could potentially benefit us in a) less conversion between int64 <> sdk.Int b) less store required. Do you think this is a possible option?

x/concentrated-liquidity/math/tick.go
p0mvn: Yes, should be safe since taking power of an integer and multiplying by an integer would yield an integer

p0mvn:
> However, it does approach to me that it might even be possible to store int64 instead of sdk.Int for the ticks, which could potentially benefit us in a) less conversion between int64 <> sdk.Int b) less store required. Do you think this is a possible option?


Yessir, I'm going to submit changes over several PRs that maximize the usage of int64. The next one is here: https://github.com/osmosis-labs/osmosis/pull/5190

Will look into your suggestion after #5190 is in

tests/cl-genesis-positions/convert.go
```diff
@@ -186,8 +186,8 @@ func ConvertSubgraphToOsmosisGenesis(positionCreatorAddresses []sdk.AccAddress,
 		position, err := clMsgServer.CreatePosition(sdk.WrapSDKContext(osmosis.Ctx), &cltypes.MsgCreatePosition{
 			PoolId:          poolId,
 			Sender:          randomCreator.String(),
-			LowerTick:       lowerTickOsmosis.Int64(),
-			UpperTick:       upperTickOsmosis.Int64(),
```
ValarDragon, line 190: oh wow the positions were always in int64's

x/concentrated-liquidity/math/tick.go
```diff
@@ -51,44 +51,44 @@ func TickToSqrtPrice(tickIndex sdk.Int) (sdk.Dec, sdk.Dec, error) {
...
+func TickToPrice(tickIndex int64) (price sdk.Dec, err error) {
+	if tickIndex == 0 {
 		return sdk.OneDec(), nil
 	}
 
 	// The formula is as follows: geometricExponentIncrementDistanceInTicks = 9 * 10**(-exponentAtPriceOne)
 	// Due to sdk.Power restrictions, if the resulting power is negative, we take 9 * (1/10**exponentAtPriceOne)
 	exponentAtPriceOne := types.ExponentAtPriceOne
-	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne.Neg()))
+	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne * -1)).TruncateInt64()
```
ValarDragon, line 62:
```suggestion
	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(-exponentAtPriceOne)).TruncateInt64()
```
nit

x/concentrated-liquidity/math/tick.go
```diff
@@ -51,44 +51,44 @@ func TickToSqrtPrice(tickIndex sdk.Int) (sdk.Dec, sdk.Dec, error) {
...
 		return sdk.Dec{}, types.TickIndexMinimumError{MinTick: types.MinTick}
 	}
-	if tickIndex.GT(sdk.NewInt(types.MaxTick)) {
+	if tickIndex > types.MaxTick {
 		return sdk.Dec{}, types.TickIndexMaximumError{MaxTick: types.MaxTick}
 	}
 
 	// Use floor division to determine what the geometricExponent is now (the delta)
-	geometricExponentDelta := tickIndex.ToDec().QuoIntMut(geometricExponentIncrementDistanceInTicks.TruncateInt()).TruncateInt()
+	geometricExponentDelta := tickIndex / geometricExponentIncrementDistanceInTicks
```
ValarDragon, line 73: Should we add a comment here for this is floor division?

x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
 
 // PriceToTickRoundDown takes a price and returns the corresponding tick index.
 // If tickSpacing is provided, the tick index will be rounded down to the nearest multiple of tickSpacing.
-func PriceToTickRoundDown(price sdk.Dec, tickSpacing uint64) (sdk.Int, error) {
+func PriceToTickRoundDown(price sdk.Dec, tickSpacing uint64) (int64, error) {
```
ValarDragon, line 123: Should we add a comment or make an error if `tickSpacking >= (1 >> 63)`

x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
...
 
-func powTenBigDec(exponent sdk.Int) osmomath.BigDec {
-	if exponent.GTE(sdk.ZeroInt()) {
-		return osmomath.NewBigDec(10).PowerInteger(exponent.Uint64())
+func powTenBigDec(exponent int64) osmomath.BigDec {
+	if exponent > 0 {
+		return osmomath.NewBigDec(10).PowerInteger(uint64(exponent))
 	}
-	return osmomath.OneDec().Quo(osmomath.NewBigDec(10).PowerInteger(exponent.Abs().Uint64()))
+	return osmomath.OneDec().Quo(osmomath.NewBigDec(10).PowerInteger(uint64(exponent * -1)))
```
ValarDragon, line 162: Hrmm, I thought this was precomputed after https://github.com/osmosis-labs/osmosis/pull/5171

x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
...
-		return sdk.Int{}, err
+		return 0, err
 	}
 
 	// Round the tick index down to the nearest tick spacing if the tickIndex is in between authorized tick values
-	tickSpacingInt := sdk.NewIntFromUint64(tickSpacing)
-	tickIndexModulus := tickIndex.Mod(tickSpacingInt)
-	if !tickIndexModulus.Equal(sdk.ZeroInt()) {
-		tickIndex = tickIndex.Sub(tickIndexModulus)
+	// Note that this is Euclidean modulus.
```
ValarDragon, line 130: whats a euclidean modulus

ValarDragon: https://en.wikipedia.org/wiki/Euclidean_algorithm ?

is there a relevant detail about that algo here?

tests/e2e/e2e_test.go
```diff
@@ -365,15 +365,15 @@ func (s *IntegrationTestSuite) TestConcentratedLiquidity() {
 	// * Uncollected fees from multiple swaps are correctly summed up and collected
 
 	// tickOffset is a tick index after the next initialized tick to which this swap needs to move the current price
-	tickOffset := sdk.NewInt(300)
+	tickOffset := int64(300)
 	sqrtPriceBeforeSwap = concentratedPool.GetCurrentSqrtPrice()
 	liquidityBeforeSwap = concentratedPool.GetLiquidity()
-	nextInitTick := sdk.NewInt(40000) // address1 position1's upper tick
+	nextInitTick := int64(40000) // address1 position1's upper tick
```
czarcas7ic, line 371: Do we need to be explicit with int64 here? Below we are not and seems to be a non issue

x/concentrated-liquidity/math/tick.go
ValarDragon: If not lets delete this.

(As someone whose implemented modular reduction many times, I have no idea what meaning is being conveyed to me with this comment)

ValarDragon: If its that, when `tickIndex is negative, we get `tickspacing - (|tickInex| % tickspacing)`, then we can add a comment that thats what were dealing with here!

x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
...
 }
 
 // powTen treats negative exponents as 1/(10**|exponent|) instead of 10**-exponent
 // This is because the sdk.Dec.Power function does not support negative exponents
-func PowTenInternal(exponent sdk.Int) sdk.Dec {
-	if exponent.GTE(sdk.ZeroInt()) {
-		return sdkTenDec.Power(exponent.Uint64())
+func PowTenInternal(exponent int64) sdk.Dec {
+	if exponent > 0 {
+		return sdkTenDec.Power(uint64(exponent))
```
czarcas7ic, line 153: it was GTE before and now its just GT

x/concentrated-liquidity/math/tick.go
```diff
@@ -120,55 +120,59 @@ func PriceToTick(price sdk.Dec) (sdk.Int, error) {
...
 	// The formula is as follows: geometricExponentIncrementDistanceInTicks = 9 * 10**(-exponentAtPriceOne)
 	// Due to sdk.Power restrictions, if the resulting power is negative, we take 9 * (1/10**exponentAtPriceOne)
 	exponentAtPriceOne := types.ExponentAtPriceOne
-	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne.Neg()))
+	geometricExponentIncrementDistanceInTicks := sdkNineDec.Mul(PowTenInternal(exponentAtPriceOne * -1)).TruncateInt64()
 
 	// Initialize the current price to 1, the current precision to exponentAtPriceOne, and the number of ticks passed to 0
 	currentPrice := sdk.OneDec()
-	ticksPassed := sdk.ZeroInt()
+	ticksPassed := int64(0)
```
czarcas7ic, line 175: I think we can just use 0 here instead of explicit type

x/concentrated-liquidity/math/tick.go
ValarDragon: Err, the reason for this being safe is a bit different then that. (int64 to power of an int64 can be a 4000 bit number)

Its that exponentAtPriceOne is bounded s.t. `9 * 10**exponentAtPriceOne` is an int64.

But to be honest, this entire line should switch to operating only on ints. And then you don't even need a sub-routine, you could just compute via int64 math 10**(exponentAtPriceOne), and that'd likely be faster than a lookup into a table.

czarcas7ic (approved):
Nice! I'm curious what the actual difference is in terms of performance is that this change made

x/concentrated-liquidity/math/tick.go
p0mvn: I thought the concern was that I'm truncating a non-whole number to an integer.

Correct. The reason I wasn't concerned about overflows as well is that we have test cases at all bounds so this would have been caught.

Looking int the switching to ints / precomputing. Thanks

tests/cl-genesis-positions/convert.go
p0mvn: Yeah I think we've refactored this multiple times due to changes to tick spacing model from the original so it might not be very consistent everywhere

x/concentrated-liquidity/math/tick.go
p0mvn: @adam's comment is the reason why I did not add checks. I don't see a world where tick spacing would be this big. I will add a comment to the godoc for clarity in the future

p0mvn: [8a6a301](https://github.com/osmosis-labs/osmosis/pull/5189/commits/8a6a3015ee30af2e9fa0d77749cd6df5ab31f23d)

x/concentrated-liquidity/lp.go
```diff
@@ -438,7 +438,7 @@ func (k Keeper) initializeInitialPositionForPool(ctx sdk.Context, pool types.Con
 	// In such a case, we do not want to round the sqrt price to 100_000_000 X/Y, but rather
 	// let it float within the possible tick range.
 	pool.SetCurrentSqrtPrice(initialCurSqrtPrice)
-	pool.SetCurrentTick(initialTick)
+	pool.SetCurrentTick(sdk.NewInt(initialTick))
```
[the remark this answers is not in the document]
p0mvn: Correct, follow-up PR that should address this: https://github.com/osmosis-labs/osmosis/pull/5190

x/concentrated-liquidity/math/tick.go
p0mvn: So the Go's default modulus and Euclidean modules differ for negative divisors.

The Go's default would not work. As a result, we do:
```go
tickIndexModulus := tickIndex % int64(tickSpacing)
if tickIndexModulus < 0 {
  tickIndexModulus += int64(tickSpacing)
}
```

This is the algo:
```go
func euclideanModulus(dividend, divisor int) int {
    modulus := dividend % divisor
    if modulus < 0 {
        modulus += divisor
    }
    return modulus
}
```
Example rounding down logic
```
tickIndex = -17
tickSpacing = 10
tickIndexModulus = 17 % 10 = 3
tickIndex = -17 -3 = -20
```

I will break this down more in the comment. I think it didn't do a good job of explaining this

tests/e2e/e2e_test.go
p0mvn: yes, by default it is initiated to `int` so it would be impossible to add it up with another `int64`  and give as a param to `TickToSqrtPrice` without casting

x/concentrated-liquidity/math/tick.go
p0mvn: No, we cannot. Other functions operate on `int64` so we must have the types consistent

x/concentrated-liquidity/math/tick.go
p0mvn: I think there was a merge conflict that got #5171 reverted. Resolving it now

ValarDragon: oh its fine if its reverted

This PR makes it clear we can get rid of that as a problem entirely :)

PR #5196, opened 2023-05-16
fix: duplicate price for multiple ticks

czarcas7ic:
Closes: #XXX

Our CL tick logic works off of a precision of 10^-18. This, combined with the fact that our per-tick precision increases at every factor of 10, means that we can have multiple ticks that correlate to the same spot price. For instance. tick -161795100 (along with many other ticks around it) correlates to a spot price of 0.000000000000000002. We need a method to determine what the largest tick (rounding up) is that corresponds to a given spot price and only use that tick (kind of like setting a canonical tick) when referencing this price.

Changes:
- When creating a position, the provided ticks are checked to see if there is a larger tick (rounding up) that represents that price with the same level of precision. If so, that tick is changed to that value
  - Example: the lower tick -161795100 correlates to price 0.000000000000000002. The tick is checked, and a larger tick (-161000000) represents that same price. The lower tick is changed to -161000000.
- `TickToSqrtPrice` now also returns the price as well
  - This is to prevent losing precision when squaring the square root price 
- `CreatePosition` now returns the lowerTick and upperTick in the response 
- Update README

## Testing and Verifying

- Added `TestRoundTick`
- Expand `TestCreatePosition`and `TestInverseRelation_CreatePosition_WithdrawPosition`

  - [x] Specification (`x/{module}/README.md`)
  - [x] Code comments?
  - [ ] N/A


osmomath/decimal.go
```diff
@@ -887,16 +887,12 @@ func MaxDec(d1, d2 BigDec) BigDec {
 
 // DecEq returns true if two given decimals are equal.
 // Intended to be used with require/assert:  require.True(t, DecEq(...))
-//
-//nolint:thelper
```
p0mvn, line 891: Let's restore this please. It breaks local environment for some people

p0mvn (approved):

PR #5204, opened 2023-05-17
bug: add split route to message server in pool manager

p0mvn:
Closes: #XXX

the new split route swap message wasn't added to the message server. This PR is adding it.


x/poolmanager/msg_server.go
```diff
@@ -69,3 +69,53 @@ func (server msgServer) SwapExactAmountOut(goCtx context.Context, msg *types.Msg
...
+	if err != nil {
+		return nil, err
+	}
+
+	tokenInAmount, err := server.keeper.SplitRouteExactAmountOut(ctx, sender, msg.Routes, msg.TokenOutDenom, msg.TokenInMaxAmount)
+	if err != nil {
+		return nil, err
+	}
+
+	// Swap event is handled elsewhere
```
mattverse, line 111:
```suggestion
	// Swap event is handled in each pool module's SwapExactAmountOut
```

mattverse (approved):

PR #5211, opened 2023-05-17
perf: remove repeated reallocations in swap step iterations

p0mvn:
Closes: #XXX

This PR removes repeated reallocations of `sdk.Int` in swap iterations

- reallocations are removed by converting to mutative math
- precomputed `sdk.SmallestDec()`
- removed some obsolete method calls as a drive by

## Testing and Verifying

  - [ ] Code comments?
  - [x] N/A


x/concentrated-liquidity/swaps.go
```diff
@@ -309,8 +316,8 @@ func (k Keeper) computeOutAmtGivenIn(
 	// initialize swap state with the following parameters:
 	// as we iterate through the following for loop, this swap state will get updated after each required iteration
 	swapState := SwapState{
-		amountSpecifiedRemaining: tokenAmountInSpecified, // tokenIn
-		amountCalculated:         sdk.ZeroDec(),          // tokenOut
+		amountSpecifiedRemaining: tokenAmountInSpecified.Clone(), // tokenIn
```
czarcas7ic, line 319: I see you used clone here, but for computeIn in the same step you did not. Just pointing this out

czarcas7ic (approved):
Looks good 👍

x/concentrated-liquidity/swaps.go
mattverse: IMO we do not need to use clone here

x/concentrated-liquidity/swaps.go
```diff
@@ -538,29 +549,28 @@ func (k Keeper) computeInAmtGivenOut(
 
 		// update the swapState with the new sqrtPrice from the above swap
 		swapState.sqrtPrice = sqrtPrice
-		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(amountOut)
-		swapState.amountCalculated = swapState.amountCalculated.Add(amountIn.Add(feeChargeTotal))
+		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.SubMut(amountOut)
```
mattverse, line 552: what is the reason using mut here improves performance?

x/concentrated-liquidity/math/math.go
```diff
@@ -77,14 +75,14 @@ func CalcAmount0Delta(liq, sqrtPriceA, sqrtPriceB sdk.Dec, roundUp bool) sdk.Dec
...
 		denom := sqrtPriceA.MulTruncate(sqrtPriceB)
-		return liq.Mul(diff).Quo(denom).Ceil()
+		return liq.Mul(diff).QuoMut(denom).Ceil()
 	}
 	// These are truncated at precision end to round in favor of the pool when:
 	// - calculating amount out during swap
 	// - withdrawing liquidity
 	// The denominator is rounded up to get a smaller final amount.
 	denom := sqrtPriceA.MulRoundUp(sqrtPriceB)
-	return liq.MulTruncate(diff).QuoTruncate(denom)
```
ValarDragon, line 87: this is so sad that we have to manually do this

x/concentrated-liquidity/swaps.go
ValarDragon: if its remotely concerning, we can definitely keep cloning

(An extra heap allocation is better for confidence in 0 chance of bugs)

I'm supportive of doing clone to be more future proof here. The other changes are pretty trivial / isolated to convince yourself about, this one not as much

x/concentrated-liquidity/swaps.go
```diff
@@ -367,21 +374,21 @@ func (k Keeper) computeOutAmtGivenIn(
 		// Update the swapState with the new sqrtPrice from the above swap
 		swapState.sqrtPrice = sqrtPrice
 		// We deduct the amount of tokens we input in the computeSwapStep above from the user's defined tokenIn amount
-		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.Sub(amountIn.Add(feeCharge))
+		swapState.amountSpecifiedRemaining = swapState.amountSpecifiedRemaining.SubMut(amountIn.Add(feeCharge))
```
ValarDragon, line 377:
```suggestion
		swapState.amountSpecifiedRemaining.SubMut(amountIn.Add(feeCharge))
```

x/concentrated-liquidity/swaps.go
```diff
@@ -367,21 +374,21 @@ func (k Keeper) computeOutAmtGivenIn(
...
 			// Retrieve the liquidity held in the next closest initialized tick
-			liquidityNet, err := k.crossTick(ctx, p.GetId(), nextTick, sdk.NewDecCoinFromDec(tokenInMin.Denom, swapState.feeGrowthGlobal))
+			liquidityNet, err := k.crossTick(ctx, poolId, nextTick, sdk.NewDecCoinFromDec(tokenInMin.Denom, swapState.feeGrowthGlobal))
 			if err != nil {
 				return sdk.Coin{}, sdk.Coin{}, 0, sdk.Dec{}, sdk.Dec{}, err
 			}
 			liquidityNet = swapStrategy.SetLiquidityDeltaSign(liquidityNet)
 			// Update the swapState's liquidity with the new tick's liquidity
-			newLiquidity := math.AddLiquidity(swapState.liquidity, liquidityNet)
+			newLiquidity := swapState.liquidity.AddMut(liquidityNet)
```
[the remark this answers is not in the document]
ValarDragon: (note I am not checking this in PR review, trusting people with more context)

x/concentrated-liquidity/swaps.go
```diff
@@ -486,8 +497,8 @@ func (k Keeper) computeInAmtGivenOut(
 	// initialize swap state with the following parameters:
 	// as we iterate through the following for loop, this swap state will get updated after each required iteration
 	swapState := SwapState{
-		amountSpecifiedRemaining: desiredTokenOut.Amount.ToDec(), // tokenOut
-		amountCalculated:         sdk.ZeroDec(),                  // tokenIn
+		amountSpecifiedRemaining: tokenAmountOutSpecified, // tokenOut
```
ValarDragon, line 500: ah I now see Adam/Matt's point about inconsistency

x/concentrated-liquidity/swaps.go
ValarDragon: Basically, every Int/Decimal operation in the SDK allocates a new big.Int

Every new big.Int requires a heap allocation, which takes more time. (And then has to get garbage collected later)

This page talks a bit about perf differences stack/heap: https://stackoverflow.com/questions/36512509/why-is-stack-memory-cheaper-than-heap-memory-in-c  (and keep in mind that in C you manually free things. Garbage collection is far more expensive than a free call)

Lmk if your unfamiliar with stack/heap! (In which case the above post won't make any sense)

ValarDragon: So this line is going from `a = a - b` to instead `a -= b`. Now in every single compiled language, these will have no performance differences for normal code.

However in golang, we don't have interfaces for basic arithmetic that can get compiler optimized. So it can't infer that `a = a.Add(b)` is equivalent to `a += b`, and the internal methods are slower.

Rust doesn't have this problem

x/concentrated-liquidity/tick.go
```diff
@@ -55,7 +55,7 @@ func (k Keeper) initOrUpdateTick(ctx sdk.Context, poolId uint64, currentTick int
 
 	// note that liquidityIn can be either positive or negative.
 	// If negative, this would work as a subtraction from liquidityBefore
-	liquidityAfter := math.AddLiquidity(liquidityBefore, liquidityDelta)
+	liquidityAfter := liquidityBefore.Add(liquidityDelta)
```
ValarDragon, line 58: not checking this line in context, leaving to other reviewers

ValarDragon (approved):
LGTM, aside from things I noted as left to other reviewers

x/concentrated-liquidity/swaps.go
mattverse: Ah https://github.com/osmosis-labs/osmosis/pull/5211#discussion_r1197498713 this change was the question that I had in mind, everythings clear now thanks!

x/concentrated-liquidity/swaps.go
p0mvn: Added comment:
```
// N.B. We clone tokenAmountInSpecified because swapState.amountSpecifiedRemaining will get
// mutated during the compute swap step loop. However, we still need the original
// value of tokenAmountInSpecified to calculate the amount of tokenOut to return.
```

This does not apply to in given out. That's why we don't clone there

x/concentrated-liquidity/math/math.go
p0mvn: yeah lol

x/concentrated-liquidity/swaps.go
p0mvn: Replied in the thread above

p0mvn:
Thanks all, merging once CI passes

PR #5213, opened 2023-05-18
feat: v16 upgrade handler updates

czarcas7ic:
Closes: #XXX

A couple of things needed to be added to the upgrade handler:

- An initial position needed to be created in the new CL pool (otherwise there would be no spot price for superfluid code to run off of)
  - To help facilitate this, a new method that creates a position without sending the underlying assets was added 
- The new cl/pool/X denom needed to be added as a superfluid asset

## Testing and Verifying

The above items added are tested both in the e2e test as well as the upgrade handler gotest.

  - [ ] Code comments?
  - [ ] N/A


app/upgrades/v16/upgrades.go
```diff
@@ -80,7 +82,44 @@ func CreateUpgradeHandler(
...
+		// Retrieve the spot price of the OSMO/DAI pool at the time of the upgrade.
+		// This is used to initialize the balancerPool's spot price.
+		spotPrice, err := balancerPool.SpotPrice(ctx, DAIIBCDenom, DesiredDenom0)
+		if err != nil {
+			return nil, err
+		}
+		quoteAssetAmount := spotPrice.Mul(sdk.NewDec(1000000)).TruncateInt()
+
+		coins := sdk.NewCoins(sdk.NewCoin(DesiredDenom0, sdk.NewInt(1000000)), sdk.NewCoin(DAIIBCDenom, quoteAssetAmount))
+		_, _, _, _, _, err = createFullRangePositionNoSend(ctx, clPoolId, balancerPool.GetAddress(), coins, *keepers.ConcentratedLiquidityKeeper)
```
p0mvn, line 111: So we add liquidity with tokens that do not exist?

Is this appropriate? Can there be any unintended consequences of this?

E.g. how fee or incentive distributions is going to be affected?

tests/e2e/e2e_test.go
```diff
@@ -1601,4 +1601,22 @@ func (s *IntegrationTestSuite) TestAConcentratedLiquidity_CanonicalPool_And_Para
...
+		if superfluidAsset.Denom == cltypes.GetConcentratedLockupDenomFromPoolId(concentratedPoolId) {
+			found = true
+			break
+		}
+	}
+
+	s.Require().True(found, "concentrated liquidity pool denom not found in superfluid assets")
+
+	// This spot price is taken from the balancer pool that was initiated pre upgrade.
+	expectedSpotPrice, err := sdk.MustNewDecFromStr("0.73").ApproxSqrt()
```
p0mvn, line 1619: Can we get this spot price by querying the balancer pool? Otherwise, if something changes in the e2e setup in the future, this will have to be manually changed as well

x/concentrated-liquidity/lp.go
```diff
@@ -575,3 +575,94 @@ func (k Keeper) validatePositionUpdateById(ctx sdk.Context, positionId uint64, u
 
 	return nil
 }
+
+// CreatePositionNoSendUnsafe creates a new position without sending any coins to the pool. This is an unsafe method and should only be used in the upgrade handler.
+// The purpose for this method's existence is that the concentrated liquidity pool needs to have a spot price prior to us adding the cl denom as a superfluid asset.
+// We therefore have to create the position without sending any coins to the pool.
+func (k Keeper) CreatePositionNoSendUnsafe(ctx sdk.Context, poolId uint64, owner sdk.AccAddress, tokensProvided sdk.Coins, amount0Min, amount1Min sdk.Int, lowerTick, upperTick int64) (positionId uint64, actualAmount0 sdk.Int, actualAmount1 sdk.Int, liquidityDelta sdk.Dec, joinTime time.Time, lowerTickResult int64, upperTickResult int64, err error) {
```
p0mvn, line 582: Let's please have a common helper shared between `createPosition` and `CreatePositionNoSendUnsafe` to reduce code duplication below

app/upgrades/v16/upgrades.go
czarcas7ic: Yeah, I was torn on this, but after thinking about it I couldn't really think of any edge cases that this would effect (the pool acc would never actually redeem this position / collect fees / collect incentives). 

The only other alternative I could think of is sending funds directly from the balancer osmo/dai pool which also doesn't feel like a great solution. Happy to discuss further options though

czarcas7ic: Got on call, decided to potentially try and run a swap from the community pool to get some DAI and make a real position this way

x/concentrated-liquidity/lp_test.go
```diff
@@ -387,6 +387,216 @@ func (s *KeeperTestSuite) TestCreatePosition() {
 	}
 }
 
+func (s *KeeperTestSuite) TestCreatePositionNoSendUnsafe() {
+	tests := map[string]lpTest{
+		"error: non-existent pool": {
```
mattverse, line 392: I think we can also group these test cases into a list of struct called createPositionTestCases or so and de-duplicate them

app/upgrades/v16/upgrades.go
mattverse: Don't quite understand how this could work, would you be able to elaborate more on why creating full range positions without the actual coins would make sense?

czarcas7ic: It was initially just to make a spot price, but we decided to go with creating a real position with actual funds using the community pool instead

tests/e2e/e2e_test.go
czarcas7ic: Changed this, good call

x/concentrated-liquidity/lp.go
czarcas7ic: Function no longer exists

x/concentrated-liquidity/lp_test.go
czarcas7ic: Function no longer exists so resolving this

p0mvn (approved):

PR #5177, opened 2023-05-13
refactor: prototype reducing iterator overhead in swaps

p0mvn:
Closes: #XXX

## What is the purpose of the change

This change reduces the overhead stemming from re-initrializing the iterator while performing a swap.
Instead, this change reuses the sam iterator across swap steps.

It is currently in the prototype state and needs to be cleaned up for the PR to be ready. However, the result shows a significant improvement relative to what we had prior to starting benchmarking:
```
benchstat old.bench new2.bench
old.bench:5: missing iteration count
new2.bench:5: missing iteration count
goos: linux
goarch: amd64
pkg: github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
                     │  old.bench   │               new2.bench               │
                     │    sec/op    │    sec/op     vs base                  │
SwapExactAmountIn-16   777.9m ± 23%   343.6m ± 26%  -55.84% (p=0.000 n=10+6)
```

The benchmark above included the following improvements as well:
- https://github.com/osmosis-labs/osmosis/pull/5176
- https://github.com/osmosis-labs/osmosis/pull/5171
- https://github.com/osmosis-labs/osmosis/pull/5174

Notice that the overhead from iterators in pprof flame graphs is completely eliminated:

Old:
image

New:
image

## Testing and Verifying

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added unit test that validates ...*
  - *Added integration tests for end-to-end deployment with ...*
  - *Extended integration test for ...*
  - *Manually verified the change by ...*

## Documentation and Release Note

  - [ ] Does this pull request introduce a new feature or user-facing behavior changes?
  - [ ] Changelog entry added to `Unreleased` section of `CHANGELOG.md`?

Where is the change documented? 
  - [ ] Specification (`x/{module}/README.md`)
  - [ ] Osmosis documentation site
  - [ ] Code comments?
  - [ ] N/A

PR #5248, opened 2023-05-20
perf: iterator improvements for swap in given out and liquidity for full range query

p0mvn:
Closes: #XXX

This PR is a follow-up to: https://github.com/osmosis-labs/osmosis/pull/5177

Similarly to swap out given in, this PR optimizes in given out and liquidity for full range query.

Now, there is no need for some obsolete API on the compute swap strategy. As a result, the API and tests are removed.

## Testing and Verifying

  - [ ] Code comments?
  - [ ] N/A

### DIFF
--- a/x/concentrated-liquidity/swaps.go
+++ b/x/concentrated-liquidity/swaps.go
@@ -535,19 +535,30 @@ func (k Keeper) computeInAmtGivenOut(
 
 	totalFees = sdk.ZeroDec()
 
+	nextTickIter := swapStrategy.InitializeNextTickIterator(ctx, poolId, swapState.tick)
+	defer nextTickIter.Close()
+	if !nextTickIter.Valid() {
+		return sdk.Coin{}, sdk.Coin{}, 0, sdk.Dec{}, sdk.Dec{}, sdk.Dec{}, types.RanOutOfTicksForPoolError{PoolId: poolId}
+	}
+
 	// TODO: This should be GT 0 but some instances have very small remainder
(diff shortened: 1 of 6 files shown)


x/concentrated-liquidity/swaps.go
```diff
@@ -591,6 +602,13 @@ func (k Keeper) computeInAmtGivenOut(
 			if err != nil {
 				return sdk.Coin{}, sdk.Coin{}, 0, sdk.Dec{}, sdk.Dec{}, sdk.Dec{}, err
 			}
+
+			// Move next tick iterator to the next tick as the tick is crossed.
+			if !nextTickIter.Valid() {
+				return sdk.Coin{}, sdk.Coin{}, 0, sdk.Dec{}, sdk.Dec{}, sdk.Dec{}, types.RanOutOfTicksForPoolError{PoolId: poolId}
+			}
```
czarcas7ic, line 609: I don't think we need to check this again here right?

czarcas7ic (approved):
I think there is an extra nextTickIter validity check that isn't needed (but I could be wrong). All else looks good to me, nice work

x/concentrated-liquidity/swaps.go
p0mvn: I think you're correct. Updated: [74bd8ac](https://github.com/osmosis-labs/osmosis/pull/5248/commits/74bd8acce14f793d5984d1d1bf1e4e58959297d0)

Issue #5227, opened 2023-05-18
ci: auto add changelog and backport labels to dependabot


p0mvn:
## Background

I've been manually adding these to like 20 PRs per week. E.g. https://github.com/osmosis-labs/osmosis/pull/5219

Would love to see automation

## Acceptance Criteria
- labels are applied
- CI passes on dependabot PRs

PR #5250, opened 2023-05-20
Auto add changelog and backport labels to dependabot 

pysel:
Closes: #5227 

Adds labels to dependabot to avoid adding them manually 

## Testing and Verifying

Minor 

Changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,16 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  labels:
+    - "A:no-changelog"
+    - "V:state/compatible/no_backport"
 - package-ecosystem: gomod
   directory: "/"
   schedule:
     interval: weekly
+  labels:
+    - "A:no-changelog"
+    - "V:state/compatible/no_backport"  
   ignore:
     - dependency-name: "github.com/tendermint/tendermint"
       update-types: ["version-update:semver-major", "version-update:semver-minor"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,6 @@ updates:
     interval: weekly
   labels:
     - "A:no-changelog"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -5,10 +5,16 @@ updates:
...
+  labels:
+    - "A:no-changelog"
+    - "V:state/compatible/no_backport"
 - package-ecosystem: gomod
   directory: "/"
   schedule:
     interval: weekly
+  labels:
+    - "A:no-changelog"
+    - "V:state/compatible/no_backport"
```
pysel, line 17: as of my knowledge, it is not possible to add global labels, hence, here I have to duplicate labels from the previous `package-ecosystem`

ValarDragon: I don't think we should be doing this for the go mod updates, they can be state breaking

pysel: hm, I thought that the reviewer would change the label if the change is breaking, but I yeah your point makes sense, it is just better if we do not have labels for potentially breaking changes at all

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * [#5165](https://github.com/osmosis-labs/osmosis/pull/5165) Improve error message when fully exiting from a pool.
   * [#5187](https://github.com/osmosis-labs/osmosis/pull/5187) Expand `IncentivizedPools` query to include internally incentivized CL pools.
   * [#5239](https://github.com/osmosis-labs/osmosis/pull/5239) Implement `GetTotalPoolShares` public keeper function for GAMM.
+  * [#5250](https://github.com/osmosis-labs/osmosis/pull/5250) Add changelog and state compatibility labels to dependabot.
```
mattverse, line 73:
```suggestion
```

Issue #5238, opened 2023-05-19
CL: add to position and sf add to position token in amounts semantics change 


p0mvn:
Follow the discussion below for context:

_Originally posted by @AlpinYukseloglu in https://github.com/osmosis-labs/osmosis/pull/5167#discussion_r1194442647_
            
**A/C**
- discuss with the team if we want this
- implement if so

p0mvn:
We decided that we want to pursue this per Alpin's comment

PR #5267, opened 2023-05-23
fix: use min_amount field in AddToPosition Msg  correspond to amount being added

mattverse:
Closes: https://github.com/osmosis-labs/osmosis/issues/5238

Previously, the min amount fields in AddToPosition were corresponding to the min amount of the total liquidity of the position. 
This PR implements the change to use min amount field instead to use min amount field for the amount that is being added. 

## Testing and Verifying

Made fix to according tests

  - [ ] Code comments?
  - [ ] N/A

### DIFF
--- a/x/concentrated-liquidity/lp.go
+++ b/x/concentrated-liquidity/lp.go
@@ -252,7 +252,7 @@ func (k Keeper) WithdrawPosition(ctx sdk.Context, owner sdk.AccAddress, position
 // For the sake of backwards-compatibility with future implementations of charging, this function deletes the old position and creates
 // a new one with the resulting amount after addition. Note that due to truncation after `withdrawPosition`, there is some rounding error
 // that is upper bounded by 1 unit of the more valuable token.
-// Uses the amount0MinGiven,amount1MinGiven as the minimum token out for creating the new position.
+// Uses the amount0MinGiven + withdrawn amount0, amount1MinGiven + withdrawn amount1 as the minimum token out for creating the new position.
 // Note that these field indicates the min amount corresponding to the total liquidity of the position,
 // not only for the liquidity amount that is being added.
 // Uses amounts withdrawn from the original position if provided min amount is zero.
@@ -318,10 +318,10 @@ func (k Keeper) addToPosition(ctx sdk.Context, owner sdk.AccAddress, positionId 
 	minimumAmount1 := amount1Withdrawn
 
 	if !amount0MinGiven.IsZero() {
-		minimumAmount0 = amount0MinGiven
+		minimumAmount0 = amount0Withdrawn.Add(amount0MinGiven)
 	}
 	if !amount1MinGiven.IsZero() {
-		minimumAmount1 = amount1MinGiven
+		minimumAmount1 = amount1Withdrawn.Add(amount1MinGiven)
 	}
 	newPositionId, actualAmount0, actualAmount1, _, _, _, _, err := k.createPosition(ctx, position.PoolId, owner, tokensProvided, minimumAmount0, minimumAmount1, position.LowerTick, position.UpperTick)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


x/concentrated-liquidity/lp_test.go
```diff
@@ -981,7 +981,7 @@ func (s *KeeperTestSuite) TestAddToPosition() {
 				amount0Minimum: sdk.NewInt(1997960),
 				expectedError: types.InsufficientLiquidityCreatedError{
 					Actual:      sdk.NewInt(1997954),
-					Minimum:     sdk.NewInt(1997960),
+					Minimum:     sdk.NewInt(2996936),
```
p0mvn, line 984: Can you please define a success test case where if the minimum was one more unit, the test would fail? Also, please explain how the values are computed for that test cases.

x/concentrated-liquidity/lp_test.go
```diff
@@ -998,7 +998,7 @@ func (s *KeeperTestSuite) TestAddToPosition() {
 				amount1Minimum: sdk.NewInt(9999998916),
 				expectedError: types.InsufficientLiquidityCreatedError{
 					Actual:      sdk.NewInt(9999998816),
-					Minimum:     sdk.NewInt(9999998916),
+					Minimum:     sdk.NewInt(14999998915),
```
p0mvn, line 1001: Also, how are these values chosen? Please add a comment

mattverse: Added explanation and caculations per https://github.com/osmosis-labs/osmosis/pull/5267/commits/0c40801fa8c53145241ad7cf54c0711606081eea

x/concentrated-liquidity/lp.go
```diff
@@ -318,10 +318,10 @@ func (k Keeper) addToPosition(ctx sdk.Context, owner sdk.AccAddress, positionId
 	minimumAmount1 := amount1Withdrawn
 
 	if !amount0MinGiven.IsZero() {
-		minimumAmount0 = amount0MinGiven
+		minimumAmount0 = amount0Withdrawn.Add(amount0MinGiven)
```
AlpinYukseloglu, line 321: This should be the amount originally in the position prior to withdrawal, right? The actual withdrawn amount might vary due to rounding, so if e.g. `amount0MinGiven` is 1, then this would pass even if the final post-add position added 0

AlpinYukseloglu (approved):
Only had one comment that's blocking, but since it's a simple fix I'm approving to not block merge later

x/concentrated-liquidity/lp.go
czarcas7ic: Not sure if I follow here. Agreed that we could calculate based on the amount originally in the position, but what do you mean by "this would pass even if the final post-add position added 0"? @AlpinYukseloglu

AlpinYukseloglu: Exact case:
1. Starting position: 100 units of each asset, and we add 1 unit of each
2. `amount0MinGiven = 1`, meaning if we add less than 1 unit of asset0 we want to error. So any final result less than 101 units for asset0 should error.
3. Due to rounding/truncation, `amount0Withdrawn = 99`, meaning that `minimumAmount0 = 99 + 1 = 100`
4. This violates our expectation, since it sets the floor at `100` units, not `101`

Intuitively, if you have a 100 unit position and want to add at least 1 unit to it, your expected minimum outcome should be 101 units, not 100

mattverse: Not sure I follow here as well, I thought the consensus was to calculate based on the amount originally in the position?
If so, is it not opposed to statement number 2,

>  amount0MinGiven = 1, meaning if we add less than 1 unit of asset0 we want to error. So any final result less than 101 units for asset0 should error.

since we are expecting final result less than 1 unit of asset0 should error?

Maybe I did not understand the original issue correctly?

AlpinYukseloglu: Discussed offline & converged on replacing `amount0Withdrawn` here with the original amount in the position

mattverse: Re-discussed offline & converged to go as is for now

PR #5278, opened 2023-05-24
perf: remove repeated pool exists check

p0mvn:
Closes: #XXX

Removing repeated defense-in-depth sanity check. We should be checking pool existence much earlier in the execution flow.

Old:

image

New:

image

With 100k positions and 180k ticks traversed:
```
go test -bench . -run BenchmarkSwapExactAmountIn -timeout=30m -v -count=3
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz
BenchmarkSwapExactAmountIn
BenchmarkSwapExactAmountIn-4                           1        2412393389 ns/op
BenchmarkSwapExactAmountIn-4                           1        2406438654 ns/op
BenchmarkSwapExactAmountIn-4                           1        2307147354 ns/op
```

  - [ ] Code comments?
  - [x] N/A

### DIFF
--- a/x/concentrated-liquidity/bench_test.go
+++ b/x/concentrated-liquidity/bench_test.go
@@ -41,7 +41,7 @@ func runBenchmark(b *testing.B, testFunc func(b *testing.B, s *BenchTestSuite, p
 	b.StopTimer()
 
 	const (
-		numberOfPositions              = 100000
+		numberOfPositions              = 10000
 		maxAmountDeposited             = int64(1_000_000_000_000)
 		amountIn                       = "9999999999999999999"
 		shouldCreateFullRangePositions = true
--- a/x/concentrated-liquidity/bench_test.go
+++ b/x/concentrated-liquidity/bench_test.go
@@ -200,12 +200,8 @@ func runBenchmark(b *testing.B, testFunc func(b *testing.B, s *BenchTestSuite, p
 
 		swapAmountIn := sdk.MustNewDecFromStr(amountIn).TruncateInt()
 		largeSwapInCoin := sdk.NewCoin(denomIn, swapAmountIn)
-		s.Commit()
-
 		// Commit so that the changes are propagated to IAVL.
-		s.App.Commit()
-		// Reload latest store version
-		err = s.App.LoadLatestVersion()
+		s.Commit()
 		noError(b, err)
 
 		testFunc(b, &s, pool, largeSwapInCoin, currentTick)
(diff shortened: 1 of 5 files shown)


ValarDragon:
So this also wouldn't be expensive if we weren't unmarshalling the pool, and just checking that there is a value with non-zero length at that key. There should be helpers in `osmoutils` for this

ValarDragon:
This PR isn't including IAVL time AFAICT, as there is no commit happening

p0mvn:
> This PR isn't including IAVL time AFAICT, as there is no commit happening

Merged https://github.com/osmosis-labs/osmosis/pull/5282 here.

The original profiles and benchmarks include IAVL time but not level db compactions as they were done prior to the commit that removed `Commit()`: [1a94f36](https://github.com/osmosis-labs/osmosis/pull/5278/commits/1a94f3683e769363b5c1e32d9134c8f1a83927a3)

I will re-bench with level db for the next improvement. From a single run with level db, it ends up being approximately the same

p0mvn:
> So this also wouldn't be expensive if we weren't unmarshalling the pool, and just checking that there is a value with non-zero length at that key. There should be helpers in `osmoutils` for this

Agreedbut I don't think there is much value in this check. Pool existence should be checked at a higher level of abstraction.

I can implement the suggestion if people are feeling strongly but removing it feels cleaner to me.

x/concentrated-liquidity/bench_test.go
```diff
@@ -198,7 +200,9 @@ func runBenchmark(b *testing.B, testFunc func(b *testing.B, s *BenchTestSuite, p
 
 		swapAmountIn := sdk.MustNewDecFromStr(amountIn).TruncateInt()
 		largeSwapInCoin := sdk.NewCoin(denomIn, swapAmountIn)
+		// Commit so that the changes are propagated to IAVL.
 		s.Commit()
+		noError(b, err)
```
czarcas7ic, line 205: What error is this checking?

czarcas7ic (approved):
I don't think the extra noError call is needed, other than that LGTM

p0mvn:
> LGTM, I wonder if this function should be made private
> 
> (Lets comment that its up to the caller to ensure the poolID exists?)

Added comment. Looks like it has to be exported for some of the superfluid test that rely on this function.

I don't think there are any security concerns. The worst-case scenario if the pool does not exist is that the tick will not be found

x/concentrated-liquidity/bench_test.go
p0mvn: Removed

Issue #5044, opened 2023-05-02
Add option to designate reward recipient in Superfluid and Lockup 


mattverse:
## Background
In normal staking, users are given the option to specify reward reciever. (Reference `SetWithdrawAddress`).

We want to support the same action for distribution happening for Superfluid staking and bonding rewards, where through a secondary message, users are able to set the rewards receiver. 

This would also allow secondary cosmwasm contracts to have better UX throughout handling distribution rewards. 

## Acceptance Criteria

- Message to set reward receiver for have been added for each superfluid module and the lockup module. 
- According state entries have been added
- All unit tests have been added
- E2E Testing have been added

p0mvn:
More context: https://osmosis-network.slack.com/archives/C027ELA290B/p1682486556775829

PR #5281, opened 2023-05-24
feat: Add option to designate Reward Recipient to Lock and Incentives

mattverse:
Sub-component of https://github.com/osmosis-labs/osmosis/issues/5044

This PR introduces a new state entry `reward_receiver_address` to the period lock.  This field would be set to the lock owner by default at the time of lock's creation, and only would be changed by a separate message `SetLockRewardReceiverAddress`. 

Distribution at epoch time would be done to the reward receiver address set in state, instead of directly using the lock owner.

By utilizing this field, further integration with contracts or authz would be possible. 

## Testing and Verifying
According tests has been added.

  - [ ] Code comments?
  - [ ] N/A


czarcas7ic (commented):
Shouldn't we default rewardReceiver to a blank string, and only use this field if the field has an entry? This would save a bit of storage for all these locks. We also wouldn't need to do a state migration for all pre-existing locks.

czarcas7ic:
Also it appears e2e is failing at the create balancer pool step with this change

mattverse:
@czarcas7ic good call, will apply fix and reassign you for review

czarcas7ic:
Please ping on slack when its R4R, thank you @mattverse

x/incentives/keeper/distribute.go
```diff
@@ -230,6 +235,7 @@ func (k Keeper) doDistributionSends(ctx sdk.Context, distrs *distributionInfo) e
 func (k Keeper) distributeSyntheticInternal(
 	ctx sdk.Context, gauge types.Gauge, locks []lockuptypes.PeriodLock, distrInfo *distributionInfo,
 ) (sdk.Coins, error) {
+	ctx.Logger().Error("HERE")
```
czarcas7ic, line 238:
```suggestion
```

x/lockup/keeper/bench_test.go
```diff
@@ -69,7 +69,7 @@ func benchmarkResetLogic(b *testing.B, numLockups int) {
 		addr := addrs[r.Int()%numAccts]
 		simCoins := sdk.NewCoins(sdk.NewCoin(denom, sdk.NewInt(r.Int63n(100))))
 		duration := time.Duration(r.Intn(1*60*60*24*7)) * time.Second
-		lock := lockuptypes.NewPeriodLock(uint64(i+1), addr, duration, time.Time{}, simCoins)
+		lock := lockuptypes.NewPeriodLock(uint64(i+1), addr, addr.String(), duration, time.Time{}, simCoins)
```
czarcas7ic, line 72: We should actually be able to set this to a blank string and that would be analogous right?

EDIT: I see this gets converted to a blank string anyway, so this is fine.

x/lockup/keeper/lock.go
```diff
@@ -455,6 +456,38 @@ func (k Keeper) unlockMaturedLockInternalLogic(ctx sdk.Context, lock types.Perio
 	return nil
 }
 
+// SetLockReceiver changes the reward recipient address to the given address.
```
czarcas7ic, line 459:
```suggestion
// SetLockRewardReceiverAddress changes the reward recipient address to the given address.
```

x/lockup/keeper/lock.go
```diff
@@ -455,6 +456,38 @@ func (k Keeper) unlockMaturedLockInternalLogic(ctx sdk.Context, lock types.Perio
...
+
+	// check if the lock owner is the method caller.
+	if lock.GetOwner() != owner.String() {
+		return types.ErrNotLockOwner
+	}
+
+	// if the given receiver address is same as the lock owner, we store an empty string instead.
+	if lock.Owner == newReceiverAddress {
+		newReceiverAddress = ""
+	}
```
czarcas7ic, line 475: Nice

x/lockup/keeper/lock.go
```diff
@@ -455,6 +456,38 @@ func (k Keeper) unlockMaturedLockInternalLogic(ctx sdk.Context, lock types.Perio
...
+
+	if lock.RewardReceiverAddress == newReceiverAddress {
+		return types.ErrRewardReceiverIsSame
+	}
```
czarcas7ic, line 479: Double nice!

x/lockup/keeper/lock.go
```diff
@@ -801,7 +834,7 @@ func (k Keeper) SplitLock(ctx sdk.Context, lock types.PeriodLock, coins sdk.Coin
 	splitLockID := k.GetLastLockID(ctx) + 1
 	k.SetLastLockID(ctx, splitLockID)
 
-	splitLock := types.NewPeriodLock(splitLockID, lock.OwnerAddress(), lock.Duration, lock.EndTime, coins)
+	splitLock := types.NewPeriodLock(splitLockID, lock.OwnerAddress(), lock.OwnerAddress().String(), lock.Duration, lock.EndTime, coins)
```
czarcas7ic, line 837: This is wrong right? Shouldn't we be checking if the lock has a rewardReciever and setting it to that?

x/lockup/keeper/lock_test.go
```diff
@@ -236,7 +236,9 @@ func (s *KeeperTestSuite) TestUnlock() {
 		ctx := s.Ctx
 
 		addr1 := sdk.AccAddress([]byte("addr1---------------"))
-		_ = types.NewPeriodLock(1, addr1, time.Second, time.Time{}, tc.fundAcc)
+		lock := types.NewPeriodLock(1, addr1, addr1.String(), time.Second, time.Time{}, tc.fundAcc)
+		fmt.Println("----")
+		fmt.Println(lock.String())
```
czarcas7ic, line 241:
```suggestion
```

x/lockup/keeper/lock_test.go
```diff
@@ -534,6 +537,88 @@ func (s *KeeperTestSuite) TestCreateLock() {
...
+	for _, tc := range testCases {
+		s.Run(tc.name, func() {
+			s.SetupTest()
+
+			addr1 := s.TestAccs[0]
+			coins := sdk.Coins{sdk.NewInt64Coin("stake", 10)}
+
+			s.FundAcc(addr1, coins)
+
+			// create an account first
```
czarcas7ic, line 587:
```suggestion
```

x/lockup/keeper/migration.go
```diff
@@ -94,7 +94,7 @@ func MergeLockupsForSimilarDurations(
 				}
 				// create a normalized lock that will absorb the locks in the duration window
 				normalID = k.GetLastLockID(ctx) + 1
-				normalLock = types.NewPeriodLock(normalID, owner, normalizedDuration, time.Time{}, lock.Coins)
+				normalLock = types.NewPeriodLock(normalID, owner, lock.RewardReceiverAddress, normalizedDuration, time.Time{}, lock.Coins)
```
czarcas7ic, line 97: This also feels like unexpected behavior. What if we have 5 locks, all with different rewardReceiverAddresses? This would set the normalized lock to whatever the final lock's rewardRecieverAddress is. Should we require that all locks have the same rewardReceiverAddress in order to merge them?

x/lockup/keeper/msg_server.go
```diff
@@ -226,3 +226,24 @@ func (server msgServer) ForceUnlock(goCtx context.Context, msg *types.MsgForceUn
...
+
+	owner, err := sdk.AccAddressFromBech32(msg.Owner)
+	if err != nil {
+		return nil, err
+	}
+
+	newRewardRecepient, err := sdk.AccAddressFromBech32(msg.Owner)
+	if err != nil {
+		return nil, err
+	}
```
czarcas7ic, line 241: This needs to be changed from msg.Owner to msg.RewardReceiverAddress

Please also add a test case that catches this as well

x/lockup/types/msgs.go
```diff
@@ -205,3 +206,39 @@ func (m MsgForceUnlock) GetSigners() []sdk.AccAddress {
...
+func (m MsgSetRewardReceiverAddress) Route() string { return RouterKey }
+func (m MsgSetRewardReceiverAddress) Type() string  { return TypeMsgSetRewardReceiverAddress }
+func (m MsgSetRewardReceiverAddress) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Owner)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid owner address (%s)", err)
+	}
+	_, err = sdk.AccAddressFromBech32(m.RewardReceiver)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid owner address (%s)", err)
```
czarcas7ic, line 228:
```suggestion
		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid reward receiver address (%s)", err)
```

x/lockup/types/msgs.go
```diff
@@ -205,3 +206,39 @@ func (m MsgForceUnlock) GetSigners() []sdk.AccAddress {
...
+	}
+
+	if m.LockID <= 0 {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "lock id should be bigger than 1 (%s)", err)
```
czarcas7ic, line 232:
```suggestion
		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "lock id should be larger than zero, was (%s)", err)
```

x/superfluid/keeper/migrate_test.go
```diff
@@ -260,28 +260,28 @@ func (s *KeeperTestSuite) TestMigrateSuperfluidBondedBalancerToConcentrated() {
...
+		// "error: non-existent lock ID": {
+		// 	overwriteLockId:          true,
+		// 	percentOfSharesToMigrate: sdk.MustNewDecFromStr("1"),
+		// 	expectedError:            lockuptypes.ErrLockupNotFound,
+		// },
+		// "error: lock that is superfluid delegated, not unlocking (full shares), token out mins is more than exit coins": {
+		// 	percentOfSharesToMigrate: sdk.MustNewDecFromStr("1"),
+		// 	tokenOutMins:             sdk.NewCoins(sdk.NewCoin("foo", sdk.NewInt(100000))),
+		// 	expectedError:            gammtypes.ErrLimitMinAmount,
+		// },
```
czarcas7ic, line 284: This file seems incomplete

czarcas7ic (commented):
There are three things that need to be addressed prior to merge:
- Split lock issue
- Same reward receiver check in order to merge locks
- Msg server fix and test

x/lockup/keeper/lock.go
mattverse: you are correct, good catch!

x/lockup/keeper/migration.go
mattverse: IMO we should not be using this method anymore, since this was a method created use for migration in v6, and our codebase has changed alot since then. 

however, I do think that your case states a valid point thus added your suggestion above as a saftey measure just in case we do come across using this method

x/superfluid/keeper/migrate_test.go
mattverse: Forgot to uncomment this out, fixed in

x/lockup/keeper/msg_server.go
mattverse: Added test case in https://github.com/osmosis-labs/osmosis/pull/5281/commits/2d6aa3066c634f7dadc3feea7f9ce4c85dc14ae9

x/lockup/keeper/lock.go
mattverse: Also noticed that this method did not have unit tests, added in https://github.com/osmosis-labs/osmosis/pull/5281/commits/2d6aa3066c634f7dadc3feea7f9ce4c85dc14ae9

x/lockup/keeper/msg_server_test.go
```diff
@@ -455,3 +455,97 @@ func (s *KeeperTestSuite) TestMsgForceUnlock() {
...
+		s.Require().NoError(err)
+		s.Require().Equal(resp.Success, true)
+
+		// now check if the reward receiver address has been changed
+		newLock, err := s.App.LockupKeeper.GetLockByID(s.Ctx, lock.ID)
+		s.Require().NoError(err)
+		if test.param.isRewardReceiverAddressOwner {
+			s.Require().Equal(newLock.RewardReceiverAddress, "")
+		} else {
+			s.Require().Equal(newLock.RewardReceiverAddress, s.TestAccs[1].String())
```
czarcas7ic, line 547: nit: these are flipped. The expected value should be the first value and actual should be the second value.

czarcas7ic (approved):
Nice, looks solid to me!! Can you ping one more person to get a second ACK? Would feel more comfortable since this is core lockup logic.

x/incentives/keeper/distribute_test.go
```diff
@@ -78,13 +80,82 @@ func (s *KeeperTestSuite) TestDistribute() {
...
+					newReceiverAccIndex: 1,
+				},
+				{
+					lockId:              2,
+					newReceiverAccIndex: 0,
+				},
+				{
+					lockId:              3,
+					newReceiverAccIndex: 0,
+				},
```
p0mvn, line 119: Why do we change other lock ids to receiver acc index 0? The comment says that only first lock receiver is being changed

p0mvn: It seems like the last 2 change entries are redundant. If so, can we remove them?

x/lockup/keeper/lock.go
```diff
@@ -146,7 +146,8 @@ func (k Keeper) CreateLockNoSend(ctx sdk.Context, owner sdk.AccAddress, coins sd
 	ID := k.GetLastLockID(ctx) + 1
 	// unlock time is initially set without a value, gets set as unlock start time + duration
 	// when unlocking starts.
-	lock := types.NewPeriodLock(ID, owner, duration, time.Time{}, coins)
+	// the reward receiver is set as the owner by default when creating a lock, and we indicate this by using an empty string.
+	lock := types.NewPeriodLock(ID, owner, "", duration, time.Time{}, coins)
```
p0mvn, line 150: Should we add an option to have an optional receiver field at lock creation time?

x/lockup/keeper/lock.go
```diff
@@ -455,6 +456,38 @@ func (k Keeper) unlockMaturedLockInternalLogic(ctx sdk.Context, lock types.Perio
...
+	}
+
+	// check if the lock owner is the method caller.
+	if lock.GetOwner() != owner.String() {
+		return types.ErrNotLockOwner
+	}
+
+	// if the given receiver address is same as the lock owner, we store an empty string instead.
+	if lock.Owner == newReceiverAddress {
+		newReceiverAddress = ""
```
p0mvn, line 474: Can we define "" as a placeholder variable in `lock.types`?

```go
const defaultOwnerReceiverPlaceholder = ""
```

p0mvn: What is the reason we cannot just set it to the lock owner's address?

x/lockup/keeper/lock_test.go
```diff
@@ -534,6 +535,87 @@ func (s *KeeperTestSuite) TestCreateLock() {
 	s.Require().Equal(sdk.NewInt(30), balance.Amount)
 }
 
+func (s *KeeperTestSuite) TestSetLockRewardReceiverAddress() {
+	testCases := []struct {
+		name                  string
+		isnotOwner            bool
+		lockID                uint64
+		useNewReceiverAddress bool
+		expectedError         bool
```
p0mvn, line 544: nit: test errors

x/lockup/keeper/migration.go
```diff
@@ -65,6 +67,13 @@ func MergeLockupsForSimilarDurations(
 				continue
 			}
 			coin := lock.Coins[0]
+
+			// check if the reward receiver is the lock owner.
+			// if not, we do not normalize the lock.
+			if lock.RewardReceiverAddress != "" {
+				continue
+			}
```
p0mvn, line 75: ref: https://github.com/osmosis-labs/osmosis/pull/5281/files#r1208613288

I don't understand why we use an empty string placeholder when we could use the owner address directly. Please let me know

x/lockup/keeper/migration.go
p0mvn: If this method is unused, can we remove it?

p0mvn (approved):
This LGTM, nice work! Only had minor comments.

Please ping me for one more skim once the comments are addressed. I would like to see the reply to the "empty string placeholder" discussion

x/incentives/keeper/distribute_test.go
mattverse: Apologies for the confusion, the comment was outdated. 
The case is that lock Id 2 and 3 originally belongs to RecevierAccIndex 1 so we change those two to 1. 
Updated the comments

x/lockup/keeper/lock.go
mattverse: Agreed that this might be useful. Due to time constraints, I would like to implement this feature after getting this merged, developing same feature for superfluid staking and then come back if we have capacity. Created an issue here to track this https://github.com/osmosis-labs/osmosis/issues/5338

x/lockup/keeper/lock_test.go
mattverse: Addressed in https://github.com/osmosis-labs/osmosis/pull/5281/commits/80b11f5b33ad87b407ed876e98ece45f4bfa6191

x/lockup/keeper/lock.go
mattverse: Addressed in https://github.com/osmosis-labs/osmosis/pull/5281/commits/80b11f5b33ad87b407ed876e98ece45f4bfa6191, further convo about using lock owner address directly going on online

x/lockup/keeper/migration.go
mattverse: We still need the method though since we have the old upgrade handler :(

p0mvn (approved):

PR #5287, opened 2023-05-24
perf: prototype preventing redundant data refetch every swap step

p0mvn:
Closes: #XXX

According to latest swap profiles off of `main` at [this commit](https://github.com/osmosis-labs/osmosis/commit/7f0441ab1da0251a312606c50a851eb5ffebfa38), we have bottlenecks in refetching ticks, fee and incentive accumulators redundantly from state.

image

This PR prototypes removing all redundant data refetch. Actual changes will be submitted via separate PRs to isolate each change.

The core of the change is three-fold:
- avoid refetching ticks. They can be retrieved from the store iterator so there is no need to access the store with `GetTickInfo` again when crossing a tick
- fee accumulator is refetched every iteration
- incentives uptime accumulators are refetched every iteration

However, there is no need to refetch these accumulators as `accum` package guarantees that the variable is to be mutated together with state.

The resulting flafe graph from all 3 changes is attached below:
image

Notice 3 results visible on the flame graph:

1. `GetTickInfo` overhead is reduced and not visible
2. `GetUptimeAccumulators` overhead is reduced and not visible
3. `GetFeeAccumulators` overhead is reduced and not visible 

Benchmark with leveldb and 10k positions shows a 40% improvement:
```bash
benchstat current.bench latest.bench
current.bench:5: missing iteration count
latest.bench:5: missing iteration count
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz
                    │ current.bench │            latest.bench             │
                    │    sec/op     │    sec/op     vs base               │
SwapExactAmountIn-4    341.1m ± 14%   204.7m ± 11%  -39.97% (p=0.002 n=6)
```


p0mvn:
All f these improvements have been merged separately

Issue #5056, opened 2023-05-03
[CL][Internal Review][CollectIncentives] Update Accumulator while collecting incentives


stackman27:
## Background
currently collecting incentives are swaps dependent aka the ticks needs to be crossed to update accumulator. We should fix this

## Suggested Design
Call updateAccumulatorToNow in `claimAllIncentivesForPosition` test it out both via unit tests and localosmosis and we can continue discussing in the PR as long as nothing ends up breaking? On localosmosis test, we should also make sure to do a manual test confirming that the amounts claimed are correct

## Acceptance Criteria

all test and new test must pass

PR #5290, opened 2023-05-24
[CL: test] Update Accumulator while collecting incentives 

stackman27:
Closes: [#5056](https://github.com/osmosis-labs/osmosis/issues/5056)

- in the test setup we already set how much growthInside and growthOutside which updates the accumulator. in this PR we simply check if those were successfully implemented by comparing their sum with global pool uptime accumulator sum

## Testing and Verifying
- added more coverage and test cleanup 
- 
n/a

### DIFF
--- a/x/concentrated-liquidity/incentives_test.go
+++ b/x/concentrated-liquidity/incentives_test.go
@@ -2711,8 +2711,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 	uptimeHelper := getExpectedUptimes()
 	defaultSender := s.TestAccs[0]
 	tests := map[string]struct {
-		name              string
-		poolId            uint64
 		numShares         sdk.Dec
 		positionIdCreate  uint64
 		positionIdClaim   uint64
@@ -2723,7 +2721,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 		expectedError     error
 	}{
 		"happy path: claim rewards without forfeiting": {
-			poolId:           validPoolId,
 			positionIdCreate: DefaultPositionId,
 			positionIdClaim:  DefaultPositionId,
 			defaultJoinTime:  true,
@@ -2732,7 +2729,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			numShares:        sdk.OneDec(),
 		},
 		"claim and forfeit rewards (2 shares)": {
-			poolId:            validPoolId,
 			positionIdCreate:  DefaultPositionId,
 			positionIdClaim:   DefaultPositionId,
 			defaultJoinTime:   true,
@@ -2742,7 +2738,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			numShares:         sdk.NewDec(2),
 		},
 		"claim and forfeit rewards when no rewards have accrued": {
-			poolId:            validPoolId,
 			positionIdCreate:  DefaultPositionId,
 			positionIdClaim:   DefaultPositionId,
 			defaultJoinTime:   true,
@@ -2750,7 +2745,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			numShares:         sdk.OneDec(),
 		},
 		"claim and forfeit rewards with varying amounts and different denoms": {
-			poolId:            validPoolId,
 			positionIdCreate:  DefaultPositionId,
 			positionIdClaim:   DefaultPositionId,
 			defaultJoinTime:   true,
@@ -2763,8 +2757,7 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 		// error catching
 
 		"error: non existent position": {
-			poolId:           validPoolId + 1,
-			positionIdCreate: DefaultPositionId,
+			positionIdCreate: DefaultPositionId + 2,
 			positionIdClaim:  DefaultPositionId + 1, // non existent position
 			defaultJoinTime:  true,
 			growthInside:     uptimeHelper.hundredTokensMultiDenom,
@@ -2775,7 +2768,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 		},
 
 		"error: negative duration": {
-			poolId:           validPoolId,
 			positionIdCreate: DefaultPositionId,
 			positionIdClaim:  DefaultPositionId,
 			defaultJoinTime:  false,
@@ -2786,14 +2778,15 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			expectedError: types.NegativeDurationError{Duration: time.Hour * 504 * -1},
 		},
 	}
-	for _, tc := range tests {
+	for name, tc := range tests {
 		tc := tc
-		s.Run(tc.name, func() {
+		s.Run(name, func() {
 			// --- Setup test env ---
 
 			s.SetupTest()
 			clPool := s.PrepareConcentratedPool()
 			clKeeper := s.App.ConcentratedLiquidityKeeper
+			clPool.SetCurrentTick(DefaultCurrTick)
 
 			joinTime := s.Ctx.BlockTime()
 			if !tc.defaultJoinTime {
@@ -2804,7 +2797,6 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			err := clKeeper.InitOrUpdatePosition(s.Ctx, validPoolId, defaultSender, DefaultLowerTick, DefaultUpperTick, tc.numShares, joinTime, tc.positionIdCreate)
 			s.Require().NoError(err)
 
-			clPool.SetCurrentTick(DefaultCurrTick)
 			if tc.growthOutside != nil {
 				s.addUptimeGrowthOutsideRange(s.Ctx, validPoolId, defaultSender, DefaultCurrTick, DefaultLowerTick, DefaultUpperTick, tc.growthOutside)
 			}
@@ -2866,10 +2858,11 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 			}
 			s.Require().NoError(err)
 
+			newUptimeAccumValues, err := clKeeper.GetUptimeAccumulatorValues(s.Ctx, validPoolId)
+			s.Require().NoError(err)
+
 			// Ensure that forfeited incentives were properly added to their respective accumulators
 			if tc.forfeitIncentives {
-				newUptimeAccumValues, err := clKeeper.GetUptimeAccumulatorValues(s.Ctx, validPoolId)
-				s.Require().NoError(err)
 
 				// Subtract the initial accum values to get the delta
 				uptimeAccumDeltaValues, err := osmoutils.SubDecCoinArrays(newUptimeAccumValues, initUptimeAccumValues)
@@ -2892,6 +2885,29 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 				}
 				s.Require().Equal(expectedCoins, amountClaimed)
 				s.Require().Equal(sdk.Coins(nil), amountForfeited)
+
+				// Sum of growthInside
+				totalGrowthInside := sdk.NewDecCoins()
+				for _, coin := range tc.growthInside {
+					totalGrowthInside = totalGrowthInside.Add(coin...)
+				}
+
+				// sum of  growthOutside
+				totalGrowthOutside := sdk.NewDecCoins()
+				for _, coin := range tc.growthOutside {
+					totalGrowthOutside = totalGrowthOutside.Add(coin...)
+				}
+
+				// sum of uptimeAccumulatorGrowth
+				uptimeAccumulatorGrowth := sdk.NewDecCoins()
+				for _, coin := range newUptimeAccumValues {
+					uptimeAccumulatorGrowth = uptimeAccumulatorGrowth.Add(coin...)
+				}
+
+				// Calculate the totalSum, should be tc.growthInside + tc.growthOutside
+				expectedTotal := totalGrowthInside.Add(totalGrowthOutside...)
+
+				s.Require().Equal(expectedTotal, uptimeAccumulatorGrowth)
 			}
 
 			// Ensure balances have not been mutated
--- a/x/concentrated-liquidity/incentives_test.go
+++ b/x/concentrated-liquidity/incentives_test.go
@@ -2933,11 +2933,11 @@ func (s *KeeperTestSuite) TestClaimAllIncentivesForPosition() {
 	s.Require().NoError(err)
 
 	// attempting to claim with same blocktime
-	collectedInc, _, err := s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
-	s.Require().NoError(err)
+	// collectedInc, _, err := s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
+	// s.Require().NoError(err)
 
-	// expected collectedIncentives in same blockTime = 0
-	s.Require().Equal(sdk.NewCoins(), collectedInc)
+	// // expected collectedIncentives in same blockTime = 0
+	// s.Require().Equal(sdk.NewCoins(), collectedInc)
 
 	// add 1 hour to current blocktime
 	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Minute))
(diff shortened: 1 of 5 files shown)


x/concentrated-liquidity/incentives_test.go
```diff
@@ -3526,6 +3519,29 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+				// sum of uptimeAccumulatorGrowth
+				uptimeAccumulatorGrowth := sdk.NewDecCoins()
+				for _, coin := range newUptimeAccumValues {
+					uptimeAccumulatorGrowth = uptimeAccumulatorGrowth.Add(coin...)
+				}
+
+				// Calculate the totalSum, should be tc.growthInside + tc.growthOutside
+				expectedTotal := totalGrowthInside.Add(totalGrowthOutside...)
+
+				s.Require().Equal(expectedTotal, uptimeAccumulatorGrowth)
```
p0mvn, line 3544: Could you elaborate on what the goal of this is, please?

What I would expect for testing that uptime accumulators are updated is that the block time from position creation time to collection time is increased. As a result, when calling claim incentives, `updateUptimeAccumulatorsToNow` ends up distributing per every uptime that is satisfied

p0mvn: I think this test case can be used as a reference: https://github.com/osmosis-labs/osmosis/blob/af3a17400eef846fc0fafe8ce44b9d5029633b45/x/concentrated-liquidity/incentives_test.go#L2920-L3008

I think that porting some of the logic from there into a direct unit test for `claimAllIncentivesForPosition` would be sufficient for this issue.

The goal is to have strict coverage so that if I were to remove [this](https://github.com/osmosis-labs/osmosis/blob/af3a17400eef846fc0fafe8ce44b9d5029633b45/x/concentrated-liquidity/incentives.go#L838-L841), the direct unit tests would break.

Currently only `TestFunctional_ClaimIncentices_LiquidityChange_VaryingTime` breaks which is at another level of abstraction

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 	}
 }
 
+func (s *KeeperTestSuite) TestClaimAllIncentivesForPosition() {
```
stackman27, line 3531: reason why this is not tdd is because we aren't testing for the whole function but just one specific function call of the function `updatePoolUptimeAccumulatorsToNow` so this approach was a lot cleaner and simpler

x/concentrated-liquidity/incentives_test.go
stackman27: updated https://github.com/osmosis-labs/osmosis/pull/5290/commits/516cfbdbfaa41521cf9f39c51ffd1f5453dc0db6

thank you for the advice

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+	s.Require().NoError(err)
+
+	// expected collectedIncentives in same blockTime = 0
+	s.Require().Equal(sdk.NewCoins(), collectedInc)
+
+	// add 1 hour to current blocktime
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Minute))
+
+	// attempting to claim after 1min has passed
+	collectedInc, _, err = s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
```
mattverse, line 3577: can we also check the according total forfeited amounts here?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+	s.Require().NoError(err)
+
+	// expected collectedIncentives after 1 hour  = 59.999999999901820104usdc ~ 59usdc
```
mattverse, line 3580: Can we add comment or calculation on how we are getting 59.999999999901820104usdc?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+	s.Require().Equal(sdk.NewCoins(sdk.NewCoin(USDC, sdk.NewInt(59))), collectedInc)
+
+	// add 1 hour to current blocktime
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Hour))
+
+	// attempting to claim after 1hr has passed
+	collectedInc, _, err = s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
+	s.Require().NoError(err)
+
+	// expected collectedIncentives after 1 hour = 3599.999999998662853243usdc ~ 3599usdc
```
mattverse, line 3590: ditto here

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+		IncentiveRecordBody: types.IncentiveRecordBody{
+			RemainingAmount: sdk.NewDec(100000),
+			EmissionRate:    sdk.NewDec(1), // 1 per second
+			StartTime:       s.Ctx.BlockTime(),
+		},
+		MinUptime: time.Nanosecond,
+	}
+	err = s.App.ConcentratedLiquidityKeeper.SetMultipleIncentiveRecords(s.Ctx, []types.IncentiveRecord{testIncentiveRecord})
+	s.Require().NoError(err)
+
```
mattverse, line 3565: Can we add scenario to try claiming all incentives  for position with an invalid position id?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+	// add 1 hour to current blocktime
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Hour))
+
+	// attempting to claim after 1hr has passed
+	collectedInc, _, err = s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
+	s.Require().NoError(err)
+
+	// expected collectedIncentives after 1 hour = 3599.999999998662853243usdc ~ 3599usdc
+	s.Require().Equal(sdk.NewCoins(sdk.NewCoin(USDC, sdk.NewInt(3599))), collectedInc)
+}
```
mattverse, line 3592: Can we check if the uptime accum value has been correctly updated?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3528,69 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+
```
mattverse, line 3593: Can we also check the balance for the community pool?

mattverse (commented):
Overall, I feel like any of the branches that `claimAllIncentivesForPosition` has are untested, but only the basic happy test have been tested, would it be possible to add more checks and increase coverage of this test? (Does not have to be tdd if you prefer so)

stackman27:
converted to draft as i'm still debugging `claimAllIncentives`

czarcas7ic:
@stackman27 please ping on slack when this leaves draft status and is R4R

x/concentrated-liquidity/incentives_test.go
stackman27: do we send funds to the community pool here? i thought we just calculate incentives based on unclaimedTotal in accumulator

x/concentrated-liquidity/incentives_test.go
stackman27: yup added

x/concentrated-liquidity/incentives_test.go
stackman27: yup added

x/concentrated-liquidity/incentives_test.go
stackman27: yup added

x/concentrated-liquidity/incentives_test.go
stackman27: yup added

x/concentrated-liquidity/incentives_test.go
stackman27: i believe uptimeAccum value has been thoroughly tested in this unit test https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/incentives_test.go#L2718

not sure if i should have repetetive tests

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3395,8 +3389,7 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 		// error catching
 
 		"error: non existent position": {
-			poolId:           validPoolId + 1,
-			positionIdCreate: DefaultPositionId,
+			positionIdCreate: DefaultPositionId + 2,
```
czarcas7ic, line 3392: nit: this can just be DefaultPositionId right?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+
+	// add 1min to current blocktime
+	s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(time.Minute))
+
+	// attempting to claim after 1min has passed
+	collectedInc, forfeitInc, err = s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
+	s.Require().NoError(err)
+
+	// expected collectedIncentives after 1min = 59.999999999901820104usdc ~ 59usdc becasue 1usdc emitted every second
+	// foefeited incentives will be zero because 0.999999999901820104 truncated down is 0
```
czarcas7ic, line 3583:
```suggestion
	// expected collectedIncentives after 1min = 59.999999999901820104usdc ~ 59usdc because 1usdc emitted every second
	// forfeited incentives will be zero because 0.999999999901820104 truncated down is 0
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+	s.Require().Equal(sdk.NewCoins(sdk.NewCoin(USDC, sdk.NewInt(59))), collectedInc)
+	s.Require().Equal(0, len(forfeitInc))
+
+	//add 1 hour to current blocktime
```
czarcas7ic, line 3587:
```suggestion
	// add 1 hour to current blocktime
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+
+	// expected collectedIncentives after 1hour = 3599.999999998662853243usdc ~ 3599usdc because 1usdc emitted every second
+	// foefeited incentives will be zero because 0.999999998662853243 truncated down is 0
+	s.Require().Equal(sdk.NewCoins(sdk.NewCoin(USDC, sdk.NewInt(3599))), collectedInc)
+	s.Require().Equal(0, len(forfeitInc))
+
+	// claim with invalid position id
+	_, _, err = s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, 2)
+	s.Require().Error(err)
+
```
czarcas7ic, line 3602: I would like to add the following check:

Enough time elapses where all the remaining incentives in the incentive record are emitted, PLUS a little extra time. 

We should just sanity check that only the amount in the incentive record can be emitted and nothing else.

czarcas7ic (approved):
Once there is a reply/implementation to my question on the additional test vector I think this is good to go, thanks!

x/concentrated-liquidity/incentives_test.go
stackman27: addressed https://github.com/osmosis-labs/osmosis/pull/5290/files#diff-91569dd5b56ddbfc9450f57de47bc13061e1daceaba154cab15b635d98912d19R3557

also made test tdd as it was starting to get messy

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+					StartTime:       s.Ctx.BlockTime(),
+				},
+				MinUptime: time.Nanosecond,
+			}
+			err = s.App.ConcentratedLiquidityKeeper.SetMultipleIncentiveRecords(s.Ctx, []types.IncentiveRecord{testIncentiveRecord})
+			s.Require().NoError(err)
+
+			s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(tc.blockTimeElapsed))
+
+			collectedInc, forfeitInc, err := s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
```
AlpinYukseloglu, line 3597:
```suggestion
			collectedInc, forfeitInc, err := s.clk.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+				IncentiveDenom:       USDC,
+				IncentiveCreatorAddr: s.TestAccs[1].String(), // different address is going to create incentives
+				IncentiveRecordBody: types.IncentiveRecordBody{
+					RemainingAmount: sdk.NewDec(10_000), // 2hr 46m to emit all incentives
+					EmissionRate:    sdk.NewDec(1),      // 1 per second
+					StartTime:       s.Ctx.BlockTime(),
+				},
+				MinUptime: time.Nanosecond,
+			}
+			err = s.App.ConcentratedLiquidityKeeper.SetMultipleIncentiveRecords(s.Ctx, []types.IncentiveRecord{testIncentiveRecord})
```
AlpinYukseloglu, line 3592:
```suggestion
			err = s.clk.SetMultipleIncentiveRecords(s.Ctx, []types.IncentiveRecord{testIncentiveRecord})
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+
+			requiredBalances := sdk.NewCoins(sdk.NewCoin(ETH, sdk.NewInt(1_000_000)), sdk.NewCoin(USDC, sdk.NewInt(5_000_000_000)))
+			s.FundAcc(s.TestAccs[0], requiredBalances)
+			s.FundAcc(s.TestAccs[1], requiredBalances)
+
+			// Create CL pool
+			pool := s.PrepareConcentratedPool()
+
+			// Set up position
+			positionIdOne, _, _, _, _, _, _, err := s.App.ConcentratedLiquidityKeeper.CreatePosition(s.Ctx, pool.GetId(), s.TestAccs[0], requiredBalances, sdk.ZeroInt(), sdk.ZeroInt(), DefaultLowerTick, DefaultUpperTick)
```
AlpinYukseloglu, line 3577:
```suggestion
			positionIdOne, _, _, _, _, _, _, err := s.clk.CreatePosition(s.Ctx, pool.GetId(), s.TestAccs[0], requiredBalances, sdk.ZeroInt(), sdk.ZeroInt(), DefaultLowerTick, DefaultUpperTick)
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+			}
+			err = s.App.ConcentratedLiquidityKeeper.SetMultipleIncentiveRecords(s.Ctx, []types.IncentiveRecord{testIncentiveRecord})
+			s.Require().NoError(err)
+
+			s.Ctx = s.Ctx.WithBlockTime(s.Ctx.BlockTime().Add(tc.blockTimeElapsed))
+
+			collectedInc, forfeitInc, err := s.App.ConcentratedLiquidityKeeper.ClaimAllIncentivesForPosition(s.Ctx, positionIdOne)
+			s.Require().NoError(err)
+			s.Require().Equal(tc.expectedCoins, collectedInc)
+			s.Require().Equal(tc.expectedForfeit, len(forfeitInc))
```
AlpinYukseloglu, line 3600: Why do we do a length check here instead of the actual value?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+		blockTimeElapsed time.Duration
+		expectedCoins    sdk.Coins
+		expectedForfeit  int
+		expectedErr      bool
+	}{
+		{
+			name:             "Claim with same blocktime",
+			blockTimeElapsed: 0,
+			expectedCoins:    sdk.NewCoins(),
+			expectedForfeit:  0,
```
AlpinYukseloglu, line 3542: Would it be possible to check nonzero forfeited amounts as well?

AlpinYukseloglu (approved):
I believe this functionality should be well covered by existing tests, but it's always good to have more explicit test cases

Approving since no blocking issues but did leave a few comments I'd like acks on before merging

x/concentrated-liquidity/incentives_test.go
```diff
@@ -3535,6 +3527,81 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
...
+			blockTimeElapsed: time.Hour * 3,
+			expectedCoins:    sdk.NewCoins(sdk.NewCoin(USDC, sdk.NewInt(9999))), //  after 3hr > 2hr46min = 9999usdc.999999999901820104 ~ 9999usdc
+			expectedForfeit:  0,
+		},
+	}
+
+	for _, tc := range testCases {
+		s.Run(tc.name, func() {
+			// Init suite for the test.
+			s.SetupTest()
```
czarcas7ic, line 3567: This is much cleaner, thanks for this. This makes me think though, can we add a test case where we expect some amount to be forfeit? I would imagine this would involve setting one of the uptimes greater than 1ns and then having that test case claim before that time.

stackman27: this makes sense. the only thing is that forfeitedIncentives have been pretty thoroughly tested in `TestQueryAndClaimAllIncentives` https://github.com/osmosis-labs/osmosis/blob/main/x/concentrated-liquidity/incentives_test.go#L2877

this test was just created to check impact of `updatePoolUptimeAccumulatorsToNow` in `claimAllIncentivesForPosition`. Still happy to add it if you think its really important

x/concentrated-liquidity/incentives_test.go
stackman27: i wrote an explanation here about this: https://github.com/osmosis-labs/osmosis/pull/5290#discussion_r1209548349

also the main difference between this test and `TestQueryAndCollectIncentives` is that this test specifically only focuses on `updateUptimeAccumulatorsToNow` which was added after `TestQueryAndCollectIncentives`.  So if you remove `updateUptimeAccumulatorsToNow` and run `TestQueryAndCollectIncentives` it'll pass whereas this test will fail

x/concentrated-liquidity/incentives_test.go
stackman27: fixed

x/concentrated-liquidity/incentives_test.go
AlpinYukseloglu: I see, thanks for the clarification!

x/concentrated-liquidity/incentives_test.go
stackman27: aah i see, this was for forfeited incentives, which we technically donot handle because of 1ns uptime

PR #5293, opened 2023-05-24
[chore] add cli example in CL/PoolManager module

stackman27:
Closes: #XXX

- add cli example from clickup 
- 
## Testing and Verifying
all cli commands tested


czarcas7ic (approved):
LGTM, thanks!

PR #5309, opened 2023-05-25
[chore] Update CHANGELOG.md after v15.1.2

niccoloraspa:
## What is the purpose of the change

Update CHANGELOG.md after v15.1.2 release

## Testing and Verifying

This change is a trivial rework / code cleanup without any test coverage.

## Documentation and Release Note

  - [ ] Does this pull request introduce a new feature or user-facing behavior changes?
  - [x] Changelog entry added to `Unreleased` section of `CHANGELOG.md`?

Where is the change documented? 
  - [ ] Specification (`x/{module}/README.md`)
  - [ ] Osmosis documentation site
  - [ ] Code comments?
  - [x] N/A

PR #5297, opened 2023-05-25
feat: do not require changelog and state machine labels for markdown only changes

tungbq:
Closes: #5270 
## What is the purpose of the change

> Add a description of the overall background and high level changes that this PR introduces

- Update the .github/workflows/changelog-entry-reminder.yml and .github/workflows/required_labels.yml to not require changelog and state machine labels for markdown
- Using https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

## Testing and Verifying

*(Please pick one of the following options)*

This change was tested on my forked repository and be verified as follows:
 - To be updated

## Documentation and Release Note

  - [x] Does this pull request introduce a new feature or user-facing behavior changes?
  - [ ] Changelog entry added to `Unreleased` section of `CHANGELOG.md`?

Where is the change documented? 
  - [ ] Specification (`x/{module}/README.md`)
  - [ ] Osmosis documentation site
  - [ ] Code comments?
  - [ ] N/A

PR #5311, opened 2023-05-25
Revert "feat: do not require changelog and state machine labels for m…

p0mvn:
…arkdown only changes (#5297)"

This reverts commit c0cc5c969548b0b70d99f9637960be652d41aeb0.

Closes: #XXX

This PR: https://github.com/osmosis-labs/osmosis/pull/5297 seems to have broken the changelog check.

Another PR with purely changelog updates: https://github.com/osmosis-labs/osmosis/pull/5309 does not have the `state_compatability_labels` check activated to matter what I do.

I suspect is because it is being skipped while remaining to be required. As a result, the PR ends up being blocked.

I'm reverting the change until a working solution is found and tested on fork

### DIFF
--- a/.github/workflows/changelog-entry-reminder.yml
+++ b/.github/workflows/changelog-entry-reminder.yml
@@ -4,8 +4,6 @@ on:
     types: [assigned, opened, synchronize, reopened, labeled, unlabeled]
     branches:
       - main
-    paths-ignore:
-      - '**/*.md'
 jobs:
   build:
     name: Check Actions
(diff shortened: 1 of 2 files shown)


ValarDragon:
I see, if theres no diff it should return as passed?

p0mvn:
> I see, if theres no diff it should return as passed?

Yes, if there is no diff in code files and only .md files, it should pass. I suspect that in that case it ends up being skipped, as a result, blocking the CI

p0mvn:
Having the same problem here: https://github.com/osmosis-labs/osmosis/pull/5319

czarcas7ic (approved):

PR #5317, opened 2023-05-25
audit: fungify message

czarcas7ic:
Closes: #XXX

This is a second pass review of the fungify message flow. 

Primary changes:
- `validatePositionsAndGetTotalLiquidity` now takes `fullyChargedDuration` as a param. This is because this value is already retrieved prior to calling this validate method, so it prevents having to retrieve the value twice.
- check if position exists within the `moveRewardsToNewPositionAndDeleteOldAcc` method. We previously checked this prior to calling this method for the uptime accum but not the fee accum. It is now placed inside the method to ensure it gets called for both as well as prevent code duplication
- as a drive-by change, when creating a position, we now error if one of the tokens provided does not exist in the pool
- `validatePositionsAndGetTotalLiquidity` was previously only checking if one position Id was provided when it should be checking if 2 were provided. This check was fixed, and the duplicate check was removed prior to calling this method.

## Testing and Verifying

Only minor changes of pre-existing logic, all previous tests still pass.

  - [ ] Code comments?
  - [ ] N/A


x/concentrated-liquidity/position.go
```diff
@@ -468,12 +468,11 @@ func (k Keeper) getNextPositionIdAndIncrement(ctx sdk.Context) uint64 {
...
 
 // fungifyChargedPosition takes in a list of positionIds and combines them into a single position.
-// It validates that all positions belong to the same owner, are in the same ticks, in the same pool and fully charged. Fails if not.
+// It validates that all positions belong to the same owner, are in the same tick range, pool, and are fully charged. Fails if not.
 // Otherwise, it creates a completely new position P. P's liquidity equals to the sum of all
 // liquidities of positions given by positionIds. The uptime of the join time of the new position equals
 // to current block time - max authorized uptime duration (to signify that it is fully charged).
-// The previous positions are deleted from state. Prior to deleting, the rewards are claimed.
-// The old position's unclaimed rewards are transferred to the new position.
+// The previous positions are deleted from state. Prior to deleting, he old position's unclaimed rewards are transferred to the new position.
```
p0mvn, line 475:
```suggestion
// The previous positions are deleted from state. Prior to deleting, the old position's unclaimed rewards are transferred to the new position.
```

p0mvn (approved):

czarcas7ic:
Hmmm, I am able to reproduce this negative coin amount error with seed 8074979927502828623, going to look into it

czarcas7ic:
I determined the issue is unrelated to this PR, but still the issue is present even on main

czarcas7ic:
Merging since I am addressing the issue found in the simulator in a separate PR

Issue #5329, opened 2023-05-28
[CWPool]: Misc Tracking Issue


p0mvn:
## Background

- [ ] Re-enable pool model tests after merge is complete

## Acceptance Criteria
- all bullet points are complete

Issue #5328, opened 2023-05-28
[CWPool]: decide on the need between ContractAddress and PoolAddress


p0mvn:
## Background

Cosmwasm pools currently have both pool address and a contract address.

Pool address is only created because of the shared logic in the poolmanager module. However, it ends up being unused.

## Suggested Design

Determine whether both `ContractAddress` and `PoolAddress` are needed.

One option is to overwrite "address generator" in wasmd to generate predictable address from pool id (the same logic as we use in `x/poolmanager` for other pools. The downside of this approach is needing to make the modifications to `wasmd`. However, it is the cleanest

## Acceptance Criteria
- Decide on the need between `ContractAddress` and `PoolAddress`
- Implement the solution and remove the need for having both

p0mvn:
CC: @iboss-ptk @nicolaslara

nicolaslara:
I don't think it's a problem to have both identifiers as long as they both point to the same place; they are generated at creation and never modified.

However, if we want to have just the pool id, we could leverage create2 which allows us to create predictable addresses. The only caveat would be that we need to ensure that we don't allow anyone else to instantiate contracts in those addresses (maybe by reserving a label namespace? can we be sure it's not possible to find a collision in that case?). 

I'd also like to avoid modifying wasmd as we will probably want to unfork it once the sdk unfork is done, so maybe upstreaming changes that allow for us to add custom logic to address generation?

p0mvn:
Thanks for the feedback

> I don't think it's a problem to have both identifiers as long as they both point to the same place

Unsure what you mean exactly by pointing to the same place. `ContractAddress` is generated by `wasmd` via msg instantiate. At the same time, `PoolAddress` is generated[ by this from `pool id`](https://github.com/osmosis-labs/osmosis/blob/74dc21583284fe3605983fc87c042c8f4727a954/x/poolmanager/types/pool.go#L35-L38). So they both are associated with the same pool but they point to separate module accounts.

So I agree that this is not a problem but having two addresses is redundant. If we could utilize `ContractAddress` as the `PoolAddress`, I think that would be ideal.

However, there are 2 problems with using "ContractAddress" as the "PoolAddress":
1. x/poolmanager CreatePool logic would have to be extensively refactored. This is tractable, just a large change of critical code path
2. The "ContractAddress" is not derived in the same way as "PoolAddress"es of other pool models.
[...]

PR #5330, opened 2023-05-28
feat: cw pool model and helpers

p0mvn:
Closes: #XXX

This PR merges is the merge of the ongoing feature work here: https://github.com/osmosis-labs/osmosis/pull/5327

In particular, this PR focuses on merging the cosmwasm pool model and low-level helpers and the test contract

There are no state breaks and every TODO is tracked in: https://github.com/osmosis-labs/osmosis/issues/5329

Drive by change:
- downgrade Rust toolchain in CI to version 1.69.0 as the latest doesn't work for our contracts

## Testing and Verifying

This change added tests

  - [ ] Code comments?
  - [x] N/A


x/cosmwasmpool/model/codec.go
```diff
@@ -0,0 +1,54 @@
...
+	msgservice.RegisterMsgServiceDesc(registry, &_MsgCreator_serviceDesc)
+}
+
+var (
+	amino     = codec.NewLegacyAmino()
+	ModuleCdc = codec.NewAminoCodec(amino)
+)
+
+func init() {
+	// TODO: uncomment the lines below: https://github.com/osmosis-labs/osmosis/issues/5329
```
nicolaslara, line 45: I don't see this TODO in the tracking issue

x/cosmwasmpool/model/pool.go
```diff
@@ -0,0 +1,127 @@
...
+	_ poolmanagertypes.PoolI  = &Pool{}
+	_ types.CosmWasmExtension = &Pool{}
+)
+
+// NewCosmWasmPool creates a new CosmWasm pool with the specified parameters.
+func NewCosmWasmPool(poolId uint64, codeId uint64, instantiateMsg []byte) Pool {
+	pool := Pool{
+		CosmWasmPool: CosmWasmPool{
+			PoolAddress:     poolmanagertypes.NewPoolAddress(poolId).String(),
+			ContractAddress: "", // N.B. This is to be set in InitializePool()
```
nicolaslara, line 31: should this be validated somewhere? maybe only return active=true if the contract is set? (to be fair, this will probably change on https://github.com/osmosis-labs/osmosis/issues/5328)

nicolaslara (approved):

x/cosmwasmpool/model/codec.go
p0mvn: I believe it's the third one: "Uncomment lines in codec.go for cw pool model"

x/cosmwasmpool/model/pool.go
p0mvn: This is the way it is because in `x/poolmanager` the execution flow is as follows:
1. Create poool: https://github.com/osmosis-labs/osmosis/blob/3ea287152994d2d1d05d81b302def529cf9054e4/x/poolmanager/create_pool.go#L112
- Here, we do not have contract address set yet
2. `InitializePool`: https://github.com/osmosis-labs/osmosis/blob/3ea287152994d2d1d05d81b302def529cf9054e4/x/poolmanager/create_pool.go#L133
- This is where the contract address gets set at contract instantiation time

p0mvn: If any concerns, let's discuss in future PRs

nicolaslara: that makes sense. My thought was more adding a check that the pool has properly been on all methods, but I see that that makes the interface worse as they would have to return an error

PR #5346, opened 2023-05-29
docs(x/cosmwasmpool): code id whitelist via gov props

p0mvn:
Closes: #XXX

Adding documentation on how I'm planning to handle code id management via gov props.

We briefly discussed this approach via comments in: https://github.com/osmosis-labs/osmosis/pull/4810

Would like to converge on this before I spend time implementing.

## Testing and Verifying

  - [ ] Code comments?
  - [ ] N/A

### DIFF
--- a/x/cosmwasmpool/README.md
+++ b/x/cosmwasmpool/README.md
@@ -373,3 +373,99 @@ pub struct Coin {
     pub amount: Uint128,
 }
 ```
+
+### Governance and Code Id Management
+
+Despite code upload being permissioned by governance on Osmosis, it is allowed to be done by a certain
+set of addresses:
+
+```bash
+osmosisd q wasm params
+code_upload_access:
+  address: ""
+  addresses:
+  - osmo1cd4nn8yzdrrsfqsmmvaafq8r03xn38qgqt8fzh
+  - osmo1wl59k23zngj34l7d42y9yltask7rjlnxgccawc7ltrknp6n52fps94qsjd
+  - osmo19vxp8vq8qm368dr026qxh8v82satwaf79y235lfv6wmgpwxx8dtskedaku
+  - osmo1e0x2hnhhwyek7eq3kcxu2x6pt77wdnwz0lutz9fespdr9utq963qr0y5p5
+  - osmo14n3a65fnqz9jve85l23al6m3pjugf0atvrfqh5
+  - osmo15wna5dwylkuzvljsudyn6zfsd4zl0rkg5ge888mzk4vtnjpp0z5q4e9w58
+  - osmo1r02tlyyaqs6tmrfa4jf37t7ewuxr57qp8ghzly
+  permission: AnyOfAddresses
+instantiate_default_permission: Everybody
+```
+
+We would like to make sure that it is not possible to upload any pool code
+without governance approval. This is why we create two additional governance proposals:
+
+Note, that in both cases, x/cosmwasmpool module account will act as the admin and creator of the contract.
+
+#### 1. Store code and update code id whitelist.
+
+Proposal Name: `UploadCosmWasmPoolCodeAndWhiteListProposal`
+
+On successful passing of this proposal, the code id of the pool contract will be added to the whitelist.
+As a result, anyone would be able to instantiate a pool contract with this code id when creating a pol.
+No address will be able to maliciously upload a new code id and instantiate a pool contract with it without
+governance approval.
+
+Inputs
+ - `uploadByteCode` - `[]byte` - the raw wasm bytecode
+
+The created code id is emitted via `TypeEvtUploadedCosmwasmPoolCode` event.
+
+#### 2. Store code and migrate a specific code id to a new code id.
+
+Proposal Name: `MigratePoolContractsProposal`
+
+Similarly, if we want to migrate a contract, anyone can do so but they will need to go
+through a custom governance proposal.
+
+Migrates all given cw pool contracts specified by their IDs. It has two options to perform the migration.
+
+a. If the `codeID` is non-zero, it will migrate the pool contracts to a given `codeID` assuming that it has already been uploaded. uploadByteCode must be empty in such a case. Fails if `codeID` does not exist. Fails if uploadByteCode is not empty.
+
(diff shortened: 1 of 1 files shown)


x/cosmwasmpool/README.md
```diff
@@ -373,3 +373,42 @@ pub struct Coin {
...
+be able to maliciously upload a new code id and instantiate a pool contract with it without governance approval.
+
+2. Store code and migrate a specific code id to a new code id.
+
+Similarly, if we want to migrate a contract, anyone can do so but they will need to go
+through a custom governance proposal. This proposal will take the old code id and the new code id
+as well as other migration contract upload and migration parameters. In a single multi-message,
+it will upload the new contract code, migrate the contract, and update the code id whitelist.
+
+Note, that in both cases, x/cosmwasm module account will act as the admin and creator of the contract.
```
iboss-ptk, line 414: for `x/cosmwasm`, do you mean `x/cosmwasmpool`?

x/cosmwasmpool/README.md
```diff
@@ -373,3 +373,42 @@ pub struct Coin {
...
+We would like to make sure that it is not possible to upload any pool code
+without governance approval. This is why we create two additional governance proposals:
+
+1. Store code and update code id whitelist.
+
+On successful passing of this proposal, the code id of the pool contract will be added to the whitelist.
+As a result, anyone would be able to instantiate a pool contract with this code id. No address will
+be able to maliciously upload a new code id and instantiate a pool contract with it without governance approval.
+
+2. Store code and migrate a specific code id to a new code id.
```
nicolaslara, line 407:
```suggestion
2. Store code and migrate a contract to the newly uploaded code id.
```

Question: what if we want to migrate to another code id that has already been uploaded and whitelisted? If that is allowed, isn't this proposal redundant, as it's a combination of the one above and a migrate message?

p0mvn: I think we should require going through this message if the code id was uploaded separately.

The reason is that code upload is allowed to a certain set of addresses. I don't know who is the owner, and I don't think these addresses should be able to upload contract code that is meant to be used as a pool model without going through a governance proposal that is meant to add the code id to the whitelist.

So I think people can still upload the code separately and reuse another code id. However, they should go through a gov prop for that code id to be added to the whitelist.

If there is no such governance-approved whitelist for pool code ids, one of the addresses below could, in theory, upload a malicious contract and migrate a pool to it, causing a loss of funds.

```bash
osmosisd q wasm params
code_upload_access:
  address: ""
  addresses:
  - osmo1cd4nn8yzdrrsfqsmmvaafq8r03xn38qgqt8fzh
  - osmo1wl59k23zngj34l7d42y9yltask7rjlnxgccawc7ltrknp6n52fps94qsjd
  - osmo19vxp8vq8qm368dr026qxh8v82satwaf79y235lfv6wmgpwxx8dtskedaku
  - osmo1e0x2hnhhwyek7eq3kcxu2x6pt77wdnwz0lutz9fespdr9utq963qr0y5p5
  - osmo14n3a65fnqz9jve85l23al6m3pjugf0atvrfqh5
  - osmo15wna5dwylkuzvljsudyn6zfsd4zl0rkg5ge888mzk4vtnjpp0z5q4e9w58
  - osmo1r02tlyyaqs6tmrfa4jf37t7ewuxr57qp8ghzly
  permission: AnyOfAddresses
instantiate_default_permission: Everybody
```

p0mvn:
All comments are addressed. Looking for re-reviews please

x/cosmwasmpool/README.md
nicolaslara: I understand that we should require migrating to a code id that has been uploaded by that method. My reasoning is:

 * There's code_id 1 for a pool type (say, transmutter)
 * 10 transmutter pools get created
 * Someone creates an improved version of that code and uploads it and migrates their contract using this method. 
 * Now we have code_id 2 and one pool using that code. The other 9 use code_id 1
 * If they want to migrate to the new code, they shouldn't need to upload code ids 3..11, they should just be able to use code_id 2 as it has already been whitelisted for pools.

Does that make sense? or is the idea that uploading code_id 2 in the example will migrate *all* pools using code_id 1?

p0mvn: I understand. Good point. The way I've been thinking about it is that all pool instantiations of the same contract should be associated with the same code id..

Similar to how there is one kind of balancer, stable swap and CL pool. There aren't many different versions of these pool models. If we upgrade one, we upgrade all. 

For example, assuming transmuter, I would not want to have different iterations available. Pool models and contracts should be as stable as possible. So if there is an upgrade to transmuter, all pools should be upgraded and migrated at once. I wouldn't want some pools to be left behind on the old code id.

If someone needs a customized implementation of the transmuter, that's a different story. In that case, I think this customized version should be re-uploaded separately and follow a different migration path **for all** pool instances of that customized transmuter type.

However, putting in more thought into this now, I think migrating a bunch of pools at once has a risk of being resource intensive and not scalable. It is not a problem today since we have like 1000 pools in total. However, thinking long-term I'm not sure if this is ideal.

I was also thinking about a spam vector but I don't think that's a concern since pool creation fee is charged and this message is governance-gated.

I will think about this more and maybe start implementing to have a better idea of the trade-offs. If you have any thoughts in the meantime, please lmk

nicolaslara: > In both cases, if one of the pools is specified by the given poolID, the proposal fails.

Not sure what you mean by this, but generally the approach sounds good.

I don't see mention of checking that the codeID is in the whitelist for pool code ids, but I assime that's implied.

20 pools sounds low, but I don't think that matters as after unforking the sdk we can make proposals with several messages in them

p0mvn: Corrected and fixed to:
```
In both cases, if one of the pools specified by the given `poolID` does not exist, the proposal fails.
```

Thanks for the catch.

> 20 pools sounds low, but I don't think that matters as after unforking the sdk we can make proposals with several messages in them

I'm open to a higher number this is chosen since we have 1K pools in total right now and I just want some constant bound. Agreed that this problem also goes away soon.

nicolaslara (approved):

PR #5350, opened 2023-05-30
docs(x/cosmwasmpool): integrating into incentives via x/tokenfactory …

p0mvn:
…shares

Closes: #XXX

This PR updates docs with `x/tokenfactory` share logic for being plugged into incentives.

## Testing and Verifying

  - [ ] Code comments?
  - [x] N/A

### DIFF
--- a/x/cosmwasmpool/README.md
+++ b/x/cosmwasmpool/README.md
@@ -274,36 +274,15 @@ pub enum SudoMessage {
 
 ## Incentives and Shares
 
-In order to allow CosmWasm pool to work with incentives module (or being composable in general), the contract needs to be able to create shares token and have ability to mint/burn them.
+In order to allow CosmWasm pool to work with incentives module (or being composable in general),
+the contract needs to be able to create share tokens.
 
-Each pool have share denom with this pattern: `cosmwasmpool/address/{contract_address}`.
+We handle this by utilizing the `x/tokenfactory` module.
+Each pool have share denom with this pattern: `tokenfactory/{contract_address}/{pool_id}`.
+The  pool contract should have the knowledge of its pool id in state. Similarly, we store that in state in chain-code as well.
+Each contract is responsible for minting and burning its token factory shares. The chain does no interaction with tokenfactory.
 
-(Using `contract_address` instead of `pool_id` giving the advantage of not needing to query `pool_id` in contract to reconstuct denom. It's mapping is 1:1 and already stored on chain, so on chain logic can work that out as well if it needs to.
-
-Alternatively, we can use `cosmwasmpool/pool/{pool_id}` and have contract query for it to maintain consistency. (TBD))
-
-
-```go
-type MsgMintShares struct {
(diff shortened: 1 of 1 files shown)


x/cosmwasmpool/README.md
```diff
@@ -274,36 +274,15 @@ pub enum SudoMessage {
 
 ## Incentives and Shares
 
-In order to allow CosmWasm pool to work with incentives module (or being composable in general), the contract needs to be able to create shares token and have ability to mint/burn them.
+In order to allow CosmWasm pool to work with incentives module (or being composable in general),
+the contract needs to be able to create share tokens.
 
-Each pool have share denom with this pattern: `cosmwasmpool/address/{contract_address}`.
+We handle this by utilizing the `x/tokenfactory` module.
+Each pool have share denom with this pattern: `tokenfactory/{contract_address}/{pool_id}`.
```
iboss-ptk, line 281: - I think existing pattern for tokenfactory is `factory/{creator_address}/{subdenom}`, it can't start with `tokenfactory/...`
- Do we need to specify it at all? With this we can leave how the contract handle their own shares mechanism flexible unless there is a need to.

x/cosmwasmpool/README.md
```diff
@@ -274,36 +274,15 @@ pub enum SudoMessage {
 
 ## Incentives and Shares
 
-In order to allow CosmWasm pool to work with incentives module (or being composable in general), the contract needs to be able to create shares token and have ability to mint/burn them.
+In order to allow CosmWasm pool to work with incentives module (or being composable in general),
```
nicolaslara, line 277:
```suggestion
In order to allow CosmWasm pool to work with the incentives module (or being composable in general),
```

x/cosmwasmpool/README.md
nicolaslara: yeah, if the contract address already uniquely identifies a pool, then we should be able to let the contract use any name they want. 

What we could do is add a string prefix (say, `.../cosmwasmpool/`) so that the denoms are easily recognizable as shares.

p0mvn: > I think existing pattern for tokenfactory is factory/{creator_address}/{subdenom}, it can't start with tokenfactory/...

Copied from tokenfactory README but I will confirm and correct both

> Do we need to specify it at all? With this we can leave how the contract handle their own shares mechanism flexible unless you have reasons that I'm unaware of.

>yeah, if the contract address already uniquely identifies a pool, then we should be able to let the contract use any name they want.
> What we could do is add a string prefix (say, .../cosmwasmpool/) so that the denoms are easily recognizable as shares.

This is something I wasn't sure about and agreed with the proposal to use any name. Will update this

p0mvn: Addressed [8e97a12](https://github.com/osmosis-labs/osmosis/pull/5350/commits/8e97a12a3d069b07e630979638433383875d799f)

p0mvn:
Ready for re-review please

nicolaslara (approved):

Issue #5324, opened 2023-05-26
[CL] test changing authorized params


p0mvn:

p0mvn:
We should test and confirm that nothing breaks if we change authorized params

AlpinYukseloglu:
Just to clarify the proposal here – the intended behavior for most incentives operations is to operate on all supported uptime accumulators, regardless of which are authorized.

An example to illustrate is imagine if all uptimes were authorized and then, after incentives were created for them, they were unauthorized – `collectIncentives` should still collect from the emitting incentive records that were created when the uptimes were authorized.

So the only logic that should really depend on authorized uptimes should be incentive creation (which is already tested). The equivalent in any other scenario is functionally the same as testing the case where no records were added to unauthorized uptime accumulators.

AlpinYukseloglu:
Discussed offline and decided it makes sense to run incentives tests on multiple authorized uptime permutations for robustness.

PR #5359, opened 2023-05-30
[CL] Test multiple sets of authorized uptimes on all incentives tests

AlpinYukseloglu:
Closes: #5324

> Note to reviewer: the vast majority of the diff is indenting test cases to wrap them in the helper described below. While there is some test cleanup, the majority of the tests just worked out of the box.

This PR expands our incentives tests to run multiple times on different sets of underlying authorized uptimes. While this slightly increases test overhead, we agreed that the guarantee that the benefit of having our incentives tests aligned with different environments we might see onchain was worth it.

The specific approach is as follows:
1. Add test setup logic that checks to see if specific authorized uptimes were set prior to test setup (by default does nothing on all tests)
2. Add a test helper that takes in the execution part of a test (for loop through test cases) and loops through multiple preset authorized uptimes, triggering the change above during test setup
3. Wrap all incentives tests with the above helper

While steps 2 and 3 could technically have been avoided by tying all the logic into test setup, this would trigger multiple runs of _all_ CL tests, which we decided would be overkill. The effective change is still quite minimal

## Testing and Verifying

  - [ ] Code comments?
  - [ ] N/A


czarcas7ic (approved):
L for the diff rendering as it did lol. 

Nice job on this though, was actually very little change for a relatively large increase in test coverage.

AlpinYukseloglu:
Realized the patch for state leak I made for one of the tests was actually patching unintended behavior – fixed in #5361.

Will reopen PR when  that's merged.

Also sorry about the diff rendering :( Split view should help

x/concentrated-liquidity/incentives_test.go
```diff
@@ -290,40 +290,42 @@ func (s *KeeperTestSuite) TestCreateAndGetUptimeAccumulators() {
 			expectedPass:        false,
 		},
 	}
+	s.runMultipleAuthorizedUptimes(func() {
```
AlpinYukseloglu, line 293: Note to reviewer: sorry about the horrible diff rendering here – it should be manageable with split diff review, but it seems the cases that were indented + had cleanup changes got jumbled a bit.

Issue #5343, opened 2023-05-29
v16: unblock testnet by fixing a twap issue


p0mvn:
## Background

Reported by Niccolo. Some more context: wap issue: https://osmosis-network.slack.com/archives/C047MFF0U1X/p1685109827095009

## Acceptance Criteria
- We need to fix this to unblock the testnet

czarcas7ic:
current state of issue:

created branch https://github.com/osmosis-labs/osmosis/tree/adam/twap-record-upgrade-handler with what I believe to be the fix. If @niccoloraspa runs it and it fixes the issue I will open up a PR.

PR #5363, opened 2023-05-30
fix: twap record upgrade handler

czarcas7ic:
Closes: #5343 

We currently create a CL pool and create a position all within the same block in the upgrade handler. This causes the last error time to get set to now (this is default behavior when we create a pool) AND the spot prices are set when we create a position. This causes a genesis check to pop. 

This PR nils out the last error time in the upgrade handler for the CL pool's twap record

## Testing and Verifying

Nicco tested out this branch with mainnet state and ran the upgrade handler and no longer receives the twap error.

  - [ ] Code comments?
  - [ ] N/A

### DIFF
--- a/app/upgrades/v16/upgrades.go
+++ b/app/upgrades/v16/upgrades.go
@@ -126,6 +126,16 @@ func CreateUpgradeHandler(
 			return nil, err
 		}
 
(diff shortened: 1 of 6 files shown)


AlpinYukseloglu (approved):
Nice catch/fix!

p0mvn (approved):

Issue #5084, opened 2023-05-04
[CL][Internal Review] User positions query


czarcas7ic:
## Background

User positions query does not have pagination. This needs to be added

We should add a query position by ID query as well.

p0mvn:
Might be lot's of work due to osmocli and pagination. If taking too long, drop osmocli

PR #5377, opened 2023-05-31
chore: add pagination to user positions query

czarcas7ic:
Closes: #5084 

Our current implementation of get user positions does not include pagination. If a user has sufficiently many positions, it becomes difficult to query using existing methods.

This PR implements pagination to the GetUserPositions query.

## Testing and Verifying

Test added for the new method and also tested via localosmosis with the start-with-state command

  - [ ] Code comments?
  - [ ] N/A


x/concentrated-liquidity/bench_test.go
```diff
@@ -20,10 +22,15 @@ import (
...
+	suite.Suite
 	apptesting.KeeperTestHelper
+	mtx sync.Mutex
 }
 
-func (s BenchTestSuite) createPosition(accountIndex int, poolId uint64, coin0, coin1 sdk.Coin, lowerTick, upperTick int64) {
+func (s *BenchTestSuite) createPosition(accountIndex int, poolId uint64, coin0, coin1 sdk.Coin, lowerTick, upperTick int64) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
```
czarcas7ic, line 33: Drive by change to fix linter warning

x/concentrated-liquidity/keeper_test.go
```diff
@@ -84,10 +84,10 @@ func (s *KeeperTestSuite) SetupTest() {
 }
 
 func (s *KeeperTestSuite) SetupDefaultPosition(poolId uint64) {
-	s.SetupPosition(poolId, s.TestAccs[0], DefaultCoins, DefaultLowerTick, DefaultUpperTick, s.Ctx.BlockTime())
+	s.SetupPosition(poolId, s.TestAccs[0], DefaultCoins, DefaultLowerTick, DefaultUpperTick)
 }
 
-func (s *KeeperTestSuite) SetupPosition(poolId uint64, owner sdk.AccAddress, providedCoins sdk.Coins, lowerTick, upperTick int64, joinTime time.Time) (sdk.Dec, uint64) {
+func (s *KeeperTestSuite) SetupPosition(poolId uint64, owner sdk.AccAddress, providedCoins sdk.Coins, lowerTick, upperTick int64) (sdk.Dec, uint64) {
```
czarcas7ic, line 90: We were never actually using the joinTime, so I removed it as a param and had to update every instance it was used.

x/concentrated-liquidity/bench_test.go
```diff
@@ -20,10 +22,15 @@ import (
...
 type BenchTestSuite struct {
+	suite.Suite
 	apptesting.KeeperTestHelper
+	mtx sync.Mutex
 }
 
-func (s BenchTestSuite) createPosition(accountIndex int, poolId uint64, coin0, coin1 sdk.Coin, lowerTick, upperTick int64) {
+func (s *BenchTestSuite) createPosition(accountIndex int, poolId uint64, coin0, coin1 sdk.Coin, lowerTick, upperTick int64) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
```
p0mvn, line 32: Why does this need a mutex?

x/concentrated-liquidity/position.go
```diff
@@ -212,6 +214,60 @@ func (k Keeper) GetUserPositions(ctx sdk.Context, addr sdk.AccAddress, poolId ui
 	return positions, nil
 }
 
+// GetUserPositionsSerialized behaves similarly to GetUserPositions, but returns the positions in a way that can be paginated.
+func (k Keeper) GetUserPositionsSerialized(ctx sdk.Context, addr sdk.AccAddress, poolId uint64, pagination *query.PageRequest) ([]model.Position, *query.PageResponse, error) {
+	var prefix []byte
+	var expectedLen int
```
p0mvn, line 220: nit
```suggestion
	var expectedKeyPartCount int
```

x/concentrated-liquidity/position.go
```diff
@@ -212,6 +214,60 @@ func (k Keeper) GetUserPositions(ctx sdk.Context, addr sdk.AccAddress, poolId ui
...
+	})
+
+	// Retrieve each position from the store using its ID and add it to the result slice.
+	for _, positionId := range positionIds {
+		position, err := k.GetPosition(ctx, positionId)
+		if err != nil {
+			return nil, nil, err
+		}
+		positions = append(positions, position)
+	}
```
p0mvn, line 266: Why do we have this additional retrieval? We should be able to retrieve all data as we paginate.

x/concentrated-liquidity/bench_test.go
czarcas7ic: It was getting the following linter warning:

```
createPosition passes lock by value: github.com/osmosis-labs/osmosis/v15/x/concentrated-liquidity_test.BenchTestSuite contains github.com/osmosis-labs/osmosis/v15/app/apptesting.KeeperTestHelper contains github.com/stretchr/testify/suite.Suite contains sync.RWMutex
```

Github Copilot gave the following reason for this fix:

The createPosition function in the BenchTestSuite struct passes the lock variable by value. The lock variable is defined in the KeeperTestHelper struct, which is embedded in the BenchTestSuite struct. The KeeperTestHelper struct embeds the suite.Suite struct from the testify package, which in turn contains a sync.RWMutex variable named mu.

When a variable is passed by value, a copy of the variable is created and passed to the function. In this case, a copy of the lock variable is created and passed to the CreatePosition function. This means that any changes made to the lock variable inside the CreatePosition function will not affect the original lock variable.

However, since the lock variable is a sync.RWMutex, which is a reference type, the copy of the variable still points to the same underlying data as the original variable. This means that any changes made to the data protected by the lock variable inside the CreatePosition function will affect the data protected by the original lock variable.

In general, it is safe to pass a sync.RWMutex variable by value, since the underlying data is still protected by the same mutex. However, if you need to modify the mutex itself (e.g. by calling Lock or Unlock), you should pass it by reference instead.

x/concentrated-liquidity/position.go
czarcas7ic: renamed here https://github.com/osmosis-labs/osmosis/pull/5377/commits/164b07f85a2315bf027d092348106711c6b48c53

x/concentrated-liquidity/position.go
czarcas7ic: Good point, now retrieve the position in the pagination loop here https://github.com/osmosis-labs/osmosis/pull/5377/commits/c3ee593d8b340c233941c42ab1f16a58c7e042b7

x/concentrated-liquidity/bench_test.go
p0mvn: Is it possible to disable the complaint? That mutex is unused. Having it only adds confusion IMO

x/concentrated-liquidity/client/query_proto_wrap.go
```diff
@@ -33,7 +33,7 @@ func (q Querier) UserPositions(ctx sdk.Context, req clquery.UserPositionsRequest
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	userPositions, err := q.Keeper.GetUserPositions(ctx, sdkAddr, req.PoolId)
+	userPositions, pageRes, err := q.Keeper.GetUserPositionsSerialized(ctx, sdkAddr, req.PoolId, req.Pagination)
```
p0mvn, line 36: nit: right below we iterate over `userPositions` again which is redundant. We could instead achieve that in a single iteration inside `GetUserPositionsSerialized`

p0mvn (approved):
LGTM but I would prefer if the mutex logic was removed and linter complaints disabled if possible please

x/concentrated-liquidity/bench_test.go
czarcas7ic: according to chat gpt there is no way to mute this specific complaint without disabling the entire check which seems overkill

p0mvn: I don't understand. Can you elaborate?

The solution to this is a pointer receiver.

However, why do we define a mutex in the first place and why do we need to lock? Everything should be synchronous

I'm also not getting any complaints, what linter are you using. I'm on `golangci-lint has version v1.51.0`

ValarDragon: I don't understsand why theres a mutex either

ValarDragon: I don't understsand why theres a mutex either. I think pointer receiver fixes this

czarcas7ic: Shoot yall are right, fixed here https://github.com/osmosis-labs/osmosis/pull/5377/commits/535c38b621de7e59b0d87f6358b08fd1d6170362

x/concentrated-liquidity/client/query_proto_wrap.go
czarcas7ic: good call, changed here https://github.com/osmosis-labs/osmosis/pull/5377/commits/b79937e409043d0054a5cfe0a14d21122db3c536

p0mvn:
I'm resolving conflicts and merging this since this is a query

Issue #5052, opened 2023-05-02
[Internal Review][CL][deletePosition]: test edge cases where one of the indexes does not exist


p0mvn:
## Background

In `deletePosition`, we delete from multiple indexes. However, there is only one error case covered by tests - `PositionIdNotFoundError`

## Suggested Design

We should cover all remaining branches by manually deleting from store indexes during test setup

## Acceptance Criteria
- `types.AddressPoolPositionIdNotFoundError` error case covered by test
- `types.PoolPositionIdNotFoundError` error case covered by test

PR #5378, opened 2023-05-31
Internal Review][CL][deletePosition]: test edge cases where one of the indexes does not exist

stackman27:
Closes: [#5052](https://github.com/osmosis-labs/osmosis/issues/5052)

In deletePosition, we delete from multiple indexes. However, there is only one error case covered by tests - PositionIdNotFoundError

## Testing and Verifying
added test

### DIFF
--- a/x/concentrated-liquidity/position.go
+++ b/x/concentrated-liquidity/position.go
@@ -296,19 +296,19 @@ func (k Keeper) deletePosition(ctx sdk.Context,
 	}
 	store.Delete(positionIdKey)
 
-	// Remove the address-pool-position ID to position mapping.
-	addressPoolIdPositionIdKey := types.KeyAddressPoolIdPositionId(owner, poolId, positionId)
-	if !store.Has(addressPoolIdPositionIdKey) {
-		return types.AddressPoolPositionIdNotFoundError{Owner: owner.String(), PoolId: poolId, PositionId: positionId}
-	}
-	store.Delete(addressPoolIdPositionIdKey)
-
 	// Remove the pool-position ID mapping.
 	poolIdKey := types.KeyPoolPositionPositionId(poolId, positionId)
 	if !store.Has(poolIdKey) {
 		return types.PoolPositionIdNotFoundError{PoolId: poolId, PositionId: positionId}
 	}
 	store.Delete(poolIdKey)
+
+	// Remove the address-pool-position ID to position mapping.
+	addressPoolIdPositionIdKey := types.KeyAddressPoolIdPositionId(owner, poolId, positionId)
+	if !store.Has(addressPoolIdPositionIdKey) {
+		return types.AddressPoolPositionIdNotFoundError{Owner: owner.String(), PoolId: poolId, PositionId: positionId}
+	}
+	store.Delete(addressPoolIdPositionIdKey)
 
 	// Remove the position ID to underlying lock ID mapping (if it exists)
 	positionIdLockKey := types.KeyPositionIdForLock(positionId)
--- a/x/concentrated-liquidity/position_test.go
+++ b/x/concentrated-liquidity/position_test.go
@@ -557,28 +557,46 @@ func (s *KeeperTestSuite) TestGetAllUserPositions() {
 func (s *KeeperTestSuite) TestDeletePosition() {
 	defaultPoolId := uint64(1)
 	DefaultJoinTime := s.Ctx.BlockTime()
+	defaultCreator := s.TestAccs[0]
+	secondaryAccount := s.TestAccs[1]
 
 	tests := []struct {
 		name             string
 		positionId       uint64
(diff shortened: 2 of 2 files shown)


x/concentrated-liquidity/position.go
```diff
@@ -296,20 +296,20 @@ func (k Keeper) deletePosition(ctx sdk.Context,
...
-	store.Delete(addressPoolIdPositionIdKey)
-
 	// Remove the pool-position ID mapping.
 	poolIdKey := types.KeyPoolPositionPositionId(poolId, positionId)
 	if !store.Has(poolIdKey) {
 		return types.PoolPositionIdNotFoundError{PoolId: poolId, PositionId: positionId}
 	}
 	store.Delete(poolIdKey)
 
+	// Remove the address-pool-position ID to position mapping.
```
stackman27, line 306: note: wondering if the order of deletion matters?  because i had to move this down because, we would never hit `PoolPositionIdNotFoundError` case because the `AddressPoolPositionIdNotFoundError` would always error first. 

i can move it back up but it'll be a slightly more complex test.

p0mvn: It doesn't. If error occurs, any state updates are reverted in `runTx` in base app

x/concentrated-liquidity/position_test.go
```diff
@@ -642,7 +660,15 @@ func (s *KeeperTestSuite) TestDeletePosition() {
...
 			}
 
-			err = s.App.ConcentratedLiquidityKeeper.DeletePosition(s.Ctx, test.positionId, s.TestAccs[0], defaultPoolId)
+			if test.creator != nil {
+				defaultCreator = secondaryAccount
+			}
+
+			if test.poolId != 0 && test.poolId != defaultPoolId {
+				defaultPoolId = test.poolId
+			}
```
p0mvn, line 669: Consider setting `tc.poolId` to the desired value in the test case instead of modifying here

x/concentrated-liquidity/position_test.go
```diff
@@ -557,29 +557,47 @@ func (s *KeeperTestSuite) TestGetAllUserPositions() {
...
 		},
 		{
-			name:             "Delete a non existing position",
+			name:             "InValid case: Delete a non existing position",
 			positionId:       DefaultPositionId + 1,
 			underlyingLockId: 0,
 			expectedErr:      types.PositionIdNotFoundError{PositionId: DefaultPositionId + 1},
 		},
+		{
+			name:             "InValid case: Delete the address-pool-position ID to position mapping",
```
p0mvn, line 588:
```suggestion
			name:             "Invalid case: non-existent address-pool-position ID to position mapping",
```

x/concentrated-liquidity/position_test.go
```diff
@@ -557,29 +557,47 @@ func (s *KeeperTestSuite) TestGetAllUserPositions() {
...
+			positionId:       DefaultPositionId,
+			underlyingLockId: 0,
+			creator:          secondaryAccount,
+			expectedErr:      types.AddressPoolPositionIdNotFoundError{PoolId: DefaultPositionId, PositionId: DefaultPositionId, Owner: secondaryAccount.String()},
+		},
+		{
+			name:             "InValid case: Delete pool-position ID mapping",
```
p0mvn, line 595:
```suggestion
			name:             "Invalid case: non-existent pool-position ID mapping",
```

p0mvn (approved):
LGTM after comments are addressed

x/concentrated-liquidity/position_test.go
```diff
@@ -642,7 +664,15 @@ func (s *KeeperTestSuite) TestDeletePosition() {
...
-			err = s.App.ConcentratedLiquidityKeeper.DeletePosition(s.Ctx, test.positionId, s.TestAccs[0], defaultPoolId)
+			if test.creator != nil {
+				defaultCreator = secondaryAccount
+			}
+
+			if test.poolId != defaultPoolId {
+				defaultPoolId = test.poolId
+			}
+
+			err = s.App.ConcentratedLiquidityKeeper.DeletePosition(s.Ctx, test.positionId, defaultCreator, defaultPoolId)
```
p0mvn, line 675: What is the purpose of replacing defautls? We should be able to simply do:
```go
err = s.App.ConcentratedLiquidityKeeper.DeletePosition(s.Ctx, test.positionId, defaultCreator, defaultPoolId)
```

and se the desired `test.poolId` and `test.creator` for each test case

p0mvn (changes requested):
Would like to converge on my latest comment before merging

x/concentrated-liquidity/position_test.go
stackman27: yea that works too. changed

p0mvn (approved):
LGTM

Issue #4361, opened 2023-02-17
Randomize CL Functional Fee Test


czarcas7ic:
## Background

See https://github.com/osmosis-labs/osmosis/pull/4282#discussion_r1109412778

## Suggested Design

We should make a separate testing function that randomizes the following values:

https://github.com/osmosis-labs/osmosis/blob/9cd236692a27225f65cd8f66adbdf5a96e014fa4/x/concentrated-liquidity/fees_test.go#L1036-L1042

The only constraint I see is that numAccounts must be one larger than the largest numPosition (so we have enough accounts created to initialize these positions).

pysel:
hey @czarcas7ic. I do not really understand why numAccounts should be strictly one larger than max of `num<positionType>`? I understand why max of any position type cannot be greater than numAccounts, but are they not allowed to be equal?

PR #5382, opened 2023-06-01
CL: add function to generate random `Positions` instances

pysel:
Closes: #4361

This function is used to generate an instance of the Positions struct with randomized field values.

The function populates `numSwaps` and `numAccounts` fields with a random number in the range of [2, 100]. For the remaining fields: `numFullRange`, `numNarrowRange`, `numConsecutive`, and `numOverlapping`, it generates a random number in the range of [0, numAccounts].

# Documentation

Minor

### DIFF
--- a/x/concentrated-liquidity/spread_rewards_test.go
+++ b/x/concentrated-liquidity/spread_rewards_test.go
@@ -1359,11 +1359,16 @@ func (s *KeeperTestSuite) TestUpdatePosValueToInitValuePlusGrowthOutside() {
 	}
 }
 
-func NewRandomizedPositions() Positions {
-	rand.Seed(time.Now().UnixNano())
+// NewRandomizedPositions creates a randomized Positions struct for randomization in tests.
+// Constraint: numAccounts >= (numFullRange or numNarrowRange or numConsecutive or numOverlapping)
+// Reason: you need to have at least X accounts to create X positions
+// Parameter cap is used to set maximum possible value of any field of returning struct
+func NewRandomizedPositions(cap int) Positions {
+	// Preset seed to ensure deterministic test runs.
+	rand.Seed(2)
 
-	numSwaps := 2 + rand.Intn(99) // Random number between 2 and 100
-	numAccounts := 2 + rand.Intn(99)
+	numSwaps := 2 + rand.Intn(cap) // Random number between 2 and 100, modulo cap to cap the max possible value
+	numAccounts := 2 + rand.Intn(cap)
 
 	numFullRange := rand.Intn(numAccounts + 1) // Random number between 0 and numAccounts inclusive
 	numNarrowRange := rand.Intn(numAccounts + 1)
@@ -1390,14 +1395,7 @@ type Positions struct {
 }
 
 func (s *KeeperTestSuite) TestFunctional_SpreadRewards_Swaps() {
-	positions := Positions{
-		numSwaps:       7,
-		numAccounts:    5,
-		numFullRange:   4,
-		numNarrowRange: 3,
-		numConsecutive: 2,
-		numOverlapping: 1,
-	}
+	positions := NewRandomizedPositions(10)
 	// Init suite.
 	s.SetupTest()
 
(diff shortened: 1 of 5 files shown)


czarcas7ic (approved):

x/concentrated-liquidity/spread_rewards_test.go
```diff
@@ -1359,6 +1360,27 @@ func (s *KeeperTestSuite) TestUpdatePosValueToInitValuePlusGrowthOutside() {
 	}
 }
 
+func NewRandomizedPositions() Positions {
```
p0mvn, line 1363: So the function is made but why is it unused?

x/concentrated-liquidity/spread_rewards_test.go
```diff
@@ -1359,6 +1360,27 @@ func (s *KeeperTestSuite) TestUpdatePosValueToInitValuePlusGrowthOutside() {
...
+	rand.Seed(time.Now().UnixNano())
```
p0mvn, line 1364: Let's make the seed deterministic please

p0mvn (changes requested):

x/concentrated-liquidity/spread_rewards_test.go
pysel: I can change the concrete `Positions` struct initializations to use this function, the issue description only mentioned creating this function, hence, I left tests' code unchanged

x/concentrated-liquidity/spread_rewards_test.go
pysel: but if we use a deterministic seed, won't it mean, that for each test, which gets a positions struct with this function, the initialized object will be the same? afaiu, there will be no point in using randomness at all

x/concentrated-liquidity/spread_rewards_test.go
p0mvn: Can we please either make use of this or close this? I don't see the need for introducing unused code. The author probably implied making use of this

x/concentrated-liquidity/spread_rewards_test.go
p0mvn: Can we find a way to use a different but deterministic and known seed for every use of `NewRandomizedPositions`?

Having a random seed based on time that developers don't know isn't very useful because even if you get a test failure, you can't reproduce it

x/concentrated-liquidity/spread_rewards_test.go
pysel: makes sense

x/concentrated-liquidity/spread_rewards_test.go
```diff
@@ -1359,6 +1360,27 @@ func (s *KeeperTestSuite) TestUpdatePosValueToInitValuePlusGrowthOutside() {
...
+
+	numSwaps := 2 + rand.Intn(99) // Random number between 2 and 100
+	numAccounts := 2 + rand.Intn(99)
+
+	numFullRange := rand.Intn(numAccounts + 1) // Random number between 0 and numAccounts inclusive
```
AlpinYukseloglu, line 1369: It might be helpful to reference how we bound random inputs in the simulator. e.g. in this specific case, we probably want to have _some_ cap on the number of positions so tests don't run for 10+ minutes

Tldr is you can mod the output to bound the range of the final number, similar to what's done in #5476

czarcas7ic:
@pysel bumping work on this, please lmk if you intend to address issues.

pysel:
hey @czarcas7ic @AlpinYukseloglu. I made the seed deterministic and added a parameter to cap the max value of any field in randomized struct.

however, after trying to use the function in tests, it failed in one of them ([here](https://github.com/osmosis-labs/osmosis/pull/5382/files#diff-1a552fcc0a3d36556402345e840b658a7daccb101d7f73bca9ee2e3057ade2e6R2645)).

czarcas7ic:
@pysel this is because the current tick liquidity has changed based off the num positions:

That seed should be calculated with a liquidity of `liq = Decimal("7801543753.113516346960781435")`

pysel:
sure, I will finish that

ValarDragon:
Closing this, thank you for the work. We've moved onto other methods of randomly testing this part of the code, and should phase out of this. Thank you for the time put in!

PR #5383, opened 2023-06-01
Refactor createGauge to accomodate for CL gauges

hieuvubk:
Closes: #4526 

- Currently, `LockQueryType` enum has an index 0 as `ByDuration`. So if we pass `lockuptypes.QueryCondition{}`, `distrTo.LockQueryType` will default to `ByDuration` and enter the loop => I change the indexes for passing the loop.
- Add test for case `disTo.Denom = ""` when we pass `lockuptypes.QueryCondition{}`

## Testing and Verifying

*(or)*

*(or)*

*(example:)*

  - [ ] Code comments?
  - [ ] N/A


proto/osmosis/lockup/lock.proto
```diff
@@ -54,16 +54,16 @@ message PeriodLock {
 enum LockQueryType {
   option (gogoproto.goproto_enum_prefix) = false;
 
-  ByDuration = 0;
-  ByTime = 1;
+  ByTime = 0;
+  ByDuration = 1;
```
czarcas7ic, line 58: Doesn't this break every existing lock and would require a state migration in the upgrade handler?

stackman27, line 58: why was this changed again?

x/incentives/types/msgs.go
```diff
@@ -55,10 +55,6 @@ func (m MsgCreateGauge) ValidateBasic() error {
 		return errors.New("distribution period should be 1 epoch for perpetual gauge")
 	}
 
-	if lockuptypes.LockQueryType_name[int32(m.DistributeTo.LockQueryType)] != "ByDuration" {
```
stackman27, line 58: why was this removed as well?

stackman27 (commented):
thank you for the PR 

just have few questions before ACK :))

proto/osmosis/lockup/lock.proto
hieuvubk: Even if we pass `lockuptypes.QueryCondition{}` to the function, it will still get the default value of `distrTo.LockQueryType` which is the first value of `LockQueryType` (In this case `ByDuration` => thus causing an error in the loop. ) 
So I just change the index to bypass.

x/incentives/types/msgs.go
hieuvubk: In CreateGauge for CL pool we don't want to limit just using `ByDuration` right? Or lmk if im wrong

proto/osmosis/lockup/lock.proto
stackman27: for CL the we are not distributing via locktypes so, what i originally thought was that we could skip distrTo check  below; 
```
	if distrTo.LockQueryType == lockuptypes.ByDuration {
		durationOk := false
		for _, duration := range durations {
			if duration == distrTo.Duration {
				durationOk = true
				break
			}
		}
		if !durationOk {
			return 0, fmt.Errorf("invalid duration: %d", distrTo.Duration)
		}
	}

	// check if denom this gauge pays out to exists on-chain
	if distrTo.Denom != "" && !k.bk.HasSupply(ctx, distrTo.Denom) && !strings.Contains(distrTo.Denom, "osmovaloper") {
		return 0, fmt.Errorf("denom does not exist: %s", distrTo.Denom)
	}

```

by passing empty queryCondition. lmk if that makes sense

x/pool-incentives/keeper/keeper_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"github.com/stretchr/testify/suite"
 
 	"github.com/osmosis-labs/osmosis/v16/app/apptesting"
-	appParams "github.com/osmosis-labs/osmosis/v16/app/params"
```
AlpinYukseloglu, line 11: Assuming this was removed accidentally

proto/osmosis/lockup/lock.proto
hieuvubk: Yeah I understand ur idea. But the problem I face is even calling `CreateGauge` with empty queryCondition. It is still assigned a value in the function.
Thinking about adding new value to `LockQueryType` enum to bypass both ByDuration & ByTime. Lmk what u think about that

stackman27: i see what about changing the enumType to something like;
```
enum LockQueryType {
  option (gogoproto.goproto_enum_prefix) = false;

  UNSPECIFIED = 0;
  ByTime = 0;
  ByDuration = 1;
}

```
if this works, hopefully we donot have to make lots of core logic changes and it works with existing queryCondition

hieuvubk: Yeah we r same page 🙌

hieuvubk:
Hey @stackman27 I updated your last request & got new err about `cosmwasm`, even reverting to the previous commit its still happen. Duno what cause the err.

stackman27:
hi @hieuvubk thank you so much for all the work you've put into this PR but recently we caught a feature that was lacking with external gauges and decided to work on it in this PR (https://github.com/osmosis-labs/osmosis/pull/5459)

while doing so we also fixed this issue. So i believe we can hold this PR on for now and close it when we merge this [PR](https://github.com/osmosis-labs/osmosis/pull/5459)

hieuvubk:
> hi @hieuvubk thank you so much for all the work you've put into this PR but recently we caught a feature that was lacking with external gauges and decided to work on it in this PR (#5459)
> 
> while doing so we also fixed this issue. So i believe we can hold this PR on for now and close it when we merge this [PR](https://github.com/osmosis-labs/osmosis/pull/5459)
I got it thank u

AlpinYukseloglu:
Closing this as the other PR got merged. Thanks for working on this @hieuvubk

Issue #5270, opened 2023-05-23
CI: Don't require changelog & state machine labels for markdown only changes


ValarDragon:
## Background

We should stop requiring changelog & state machine labels for PR's that only touch markdown files. 

This will lower some of the fighting the CI / manual steps we & contributors have to do.

tungbq:
Hey @ValarDragon do you need any help? I'd like to work on the CI task like this. Can I pick up this task?

p0mvn:
> Hey @ValarDragon do you need any help? I'd like to work on the CI task like this. Can I pick up this task?

Hey @tungbq . We would appreciate help with this issue. Please feel free to PR a change

tungbq:
> state machine labels
> 
Thanks @p0mvn!
Hi @ValarDragon, regarding to the `state machine labels` that you mentioned, do they mean that they are in the required label list of https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/required_labels.yml workflow: `"V:state/breaking, V:state/compatible/no_backport, V:state/compatible/backport"`?

p0mvn:
That's correct. There is one more label that is applicable: `A:no-changelog`

p0mvn:
Please see: https://github.com/osmosis-labs/osmosis/pull/5311

tungbq:
> Please see: #5311

I'm on vacation for last several days so could not check it earlier. I'm looking into it to find another approach to resolve this.

tungbq:
New approach has been proposed in #5357, please have a look.

PR #5389, opened 2023-06-01
Excempt more PR-types from required labels

ValarDragon:
Cref #5270 

Along the same idea as that PR, no longer require state machine & changelog labels for test-only PRs, and CI change PR's.


AlpinYukseloglu:
Agreed – e.g. changelog label shouldn't be required if only unreleased modules (like CL) were touched

mattverse (approved):

PR #5400, opened 2023-06-02
feat[CL]: redirect distribution record on migrate link

czarcas7ic:
Closes: #XXX

When governance creates a migration link between a balancer pool and a supercharged pool, the intent is to also migrate incentives to the new pool as well. We currently had this logic in the upgrade handler, but most pools migration links will not be created via upgrade handler. 

This PR adds the logic used in the upgrade handler to modify the distribution records to redirect rewards to the new supercharged pool.

## Testing and Verifying

Added `TestRedirectDistributionRecord`

  - [ ] Code comments?
  - [ ] N/A


x/gamm/keeper/genesis.go
```diff
@@ -36,9 +36,9 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState, unpack
 	k.setTotalLiquidity(ctx, liquidity)
 
 	if genState.MigrationRecords == nil {
-		k.OverwriteMigrationRecords(ctx, types.MigrationRecords{})
+		k.OverwriteMigrationRecordsNoRedirect(ctx, types.MigrationRecords{})
 	} else {
-		k.OverwriteMigrationRecords(ctx, *genState.MigrationRecords)
+		k.OverwriteMigrationRecordsNoRedirect(ctx, *genState.MigrationRecords)
```
czarcas7ic, line 41: Note to reviewer: we need a separate NoRedirect method to strictly be used in the genesis initialization. This is because, when we run export genesis, we expect the distribution records to have already been properly handled. 

If we try to run distribution record logic within the gamm module in init genesis, this does not work, since the incentives module must get initialized post gamm module initialization (pools must exist prior to setting up the distribution records for them).

I am fairly confident with this logic, but just wanted to explain that in depth here since it wasn't straight forward initially when implementing it.

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,52 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
 	return sdk.BigEndianToUint64(balancerPoolIdBigEndian), nil
 }
 
-// OverwriteMigrationRecords sets the balancer to gamm pool migration info to the store
-// Deletes all existing records, migrationInfo in state is completely overwitten by the given migrationInfo.
-func (k Keeper) OverwriteMigrationRecords(ctx sdk.Context, migrationInfo types.MigrationRecords) {
+// deleteMigrationKeys deletes all keys with the given prefixKey.
```
p0mvn, line 112: Can you also elaborate what prefixKey can or cannot be?
```suggestion
// deleteMigrationKeys deletes all migration records with the given prefixKey.
```

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,52 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
...
 	}
+}
 
-	prefixStore = prefix.NewStore(store, types.KeyPrefixMigrationInfoCLPool)
-	iter = prefixStore.Iterator(nil, nil)
-	defer iter.Close()
+// OverwriteMigrationRecords sets the balancer to gamm pool migration info to the store and deletes all existing records
+// migrationInfo in state is completely overwitten by the given migrationInfo.
+// Additionally, the distribution record for the balancer pool is modified to redirect incentives to the new concentrated pool.
+func (k Keeper) OverwriteMigrationRecords(ctx sdk.Context, migrationInfo types.MigrationRecords) error {
```
p0mvn, line 127:
```suggestion
func (k Keeper) OverwriteMigrationRecordsAndRedirectDistrRecords(ctx sdk.Context, migrationInfo types.MigrationRecords) error {
```

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,52 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
...
+			return err
+		}
 	}
+	return nil
+}
+
+// OverwriteMigrationRecordsNoRedirect behaves the same as OverwriteMigrationRecords but does not modify the distribution record.
+// This is specifically used in init genesis, since we want to overwrite the migration records but not modify the distribution record,
+// since this should be handled at time of genesis export by the incentive module.
+func (k Keeper) OverwriteMigrationRecordsNoRedirect(ctx sdk.Context, migrationInfo types.MigrationRecords) {
```
p0mvn, line 152: Why does this have to be overwritten? If this is meant to be used in InitGenesis, let's call this simple `SetMigrationRecords` and remove the deletion logic

x/gamm/keeper/migrate.go
```diff
@@ -140,6 +165,49 @@ func (k Keeper) OverwriteMigrationRecords(ctx sdk.Context, migrationInfo types.M
...
+	// Get longest gauge duration from CFMM pool.
+	longestDurationGauge := cfmmGauges[0]
+	for i := 1; i < len(cfmmGauges); i++ {
+		if cfmmGauges[i].DistributeTo.Duration > longestDurationGauge.DistributeTo.Duration {
+			longestDurationGauge = cfmmGauges[i]
+		}
+	}
+
+	// Get concentrated liquidity gauge duration.
+	distributionEpochDuration := k.incentivesKeeper.GetEpochInfo(ctx).Duration
```
p0mvn, line 189: Can you confirm that this is always meant to be equal epoch duration?

x/gamm/keeper/migrate_test.go
```diff
@@ -876,3 +877,103 @@ func (s *KeeperTestSuite) TestGetAllMigrationInfo() {
...
+			suite.Require().NoError(err)
+			distrInfo := suite.App.PoolIncentivesKeeper.GetDistrInfo(suite.Ctx)
+			suite.Require().Equal(distrInfo.Records[0].GaugeId, concentratedPoolGaugeId)
+
+			// Validate that distribution record from another pool is not redirected.
+			suite.Require().Equal(distrInfo.Records[1].GaugeId, gaugeToNotRedirect)
+
+			// Validate that old gauge still exist
+			_, err = suite.App.IncentivesKeeper.GetGaugeByID(suite.Ctx, gaugeToRedirect)
+			suite.Require().NoError(err)
```
p0mvn, line 976: Excellent work on the test, super clear

p0mvn (approved):

p0mvn:
I'm going to refresh the PR and address comments

p0mvn:
I refreshed the PR and addressed nits, leaving the final comments @czarcas7ic and waiting for one more ACK on this

AlpinYukseloglu (commented):
Looks good on a first pass – not approving since I'm reviewing this quite late at night and would be good to do another pass later but feel free to merge if this is blocking something else

x/gamm/keeper/migrate.go
czarcas7ic: Renamed here https://github.com/osmosis-labs/osmosis/pull/5400/commits/52c27e483ac77f5189e4d9a2b50553f0813d8a06

x/gamm/keeper/migrate.go
czarcas7ic: Yeah its only meant to be called in initGenesis, renamed and removed deletion logic here https://github.com/osmosis-labs/osmosis/pull/5400/commits/6a449dce9f2a6ffd8768d3394c47a7d4f9a3bbbd

x/gamm/keeper/migrate.go
czarcas7ic: Good call-out. As far as I can tell, this logic is solid to fall back on. See the `Distribute` function in the incentives module:

https://github.com/osmosis-labs/osmosis/blob/6a449dce9f2a6ffd8768d3394c47a7d4f9a3bbbd/x/incentives/keeper/distribute.go#L380-L393

This always distributes to the duration returned from the `GetEpochInfo` EpochInfo.

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,46 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
 	return sdk.BigEndianToUint64(balancerPoolIdBigEndian), nil
 }
 
-// OverwriteMigrationRecords sets the balancer to gamm pool migration info to the store
-// Deletes all existing records, migrationInfo in state is completely overwitten by the given migrationInfo.
-func (k Keeper) OverwriteMigrationRecords(ctx sdk.Context, migrationInfo types.MigrationRecords) {
+// deleteMigrationKeys deletes all migration records with the given prefixKey.
+func (k Keeper) deleteMigrationKeys(ctx sdk.Context, prefixKey []byte) {
```
AlpinYukseloglu, line 113: I guess this is a general `deleteKeys` function right? Thoughts on moving this to osmoutils `store_helper.go` (probably in separate PR)?

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,46 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
...
+// OverwriteMigrationRecordsAndRedirectDistrRecords sets the balancer to gamm pool migration info to the store and deletes all existing records
+// migrationInfo in state is completely overwitten by the given migrationInfo.
+// Additionally, the distribution record for the balancer pool is modified to redirect incentives to the new concentrated pool.
+func (k Keeper) OverwriteMigrationRecordsAndRedirectDistrRecords(ctx sdk.Context, migrationInfo types.MigrationRecords) error {
+	store := ctx.KVStore(k.storeKey)
 
-	for ; iter.Valid(); iter.Next() {
-		prefixStore.Delete(iter.Key())
+	// delete all existing keys
+	k.deleteMigrationKeys(ctx, types.KeyPrefixMigrationInfoBalancerPool)
```
AlpinYukseloglu, line 131: Do we need to run these deletions? Does simply running `Set` below not overwrite them?

x/gamm/keeper/migrate.go
```diff
@@ -109,27 +109,46 @@ func (k Keeper) GetLinkedBalancerPoolID(ctx sdk.Context, concentratedPoolId uint
...
+	k.deleteMigrationKeys(ctx, types.KeyPrefixMigrationInfoCLPool)
+
+	for _, balancerToCLPoolLink := range migrationInfo.BalancerToConcentratedPoolLinks {
+		balancerToClPoolKey := types.GetKeyPrefixMigrationInfoBalancerPool(balancerToCLPoolLink.BalancerPoolId)
+		store.Set(balancerToClPoolKey, sdk.Uint64ToBigEndian(balancerToCLPoolLink.ClPoolId))
```
AlpinYukseloglu, line 136: We should be using `osmoutils.MustSet` here and below to ensure there are no marshaling-related errors right?

x/gamm/keeper/migrate.go
```diff
@@ -140,6 +159,49 @@ func (k Keeper) OverwriteMigrationRecords(ctx sdk.Context, migrationInfo types.M
...
+	concentratedGaugeId, err := k.poolIncentivesKeeper.GetPoolGaugeId(ctx, clPoolId, distributionEpochDuration)
+	if err != nil {
+		return err
+	}
+
+	// Iterate through all the distr records, and redirect the old balancer gauge to the new concentrated gauge.
+	distrInfo := k.poolIncentivesKeeper.GetDistrInfo(ctx)
+	for i, distrRecord := range distrInfo.Records {
+		if distrRecord.GaugeId == longestDurationGauge.Id {
+			distrInfo.Records[i].GaugeId = concentratedGaugeId
```
AlpinYukseloglu, line 195: Let's make sure to let Johnny know that we're mutating existing records so future gov props don't accidentally create duplicate incentives

x/gamm/keeper/migrate.go
```diff
@@ -245,7 +307,10 @@ func (k Keeper) ReplaceMigrationRecords(ctx sdk.Context, records []types.Balance
 
 	migrationInfo.BalancerToConcentratedPoolLinks = records
 
-	k.OverwriteMigrationRecords(ctx, migrationInfo)
+	err = k.OverwriteMigrationRecordsAndRedirectDistrRecords(ctx, migrationInfo)
```
AlpinYukseloglu, line 310: Just occurred to me – it doesn't seem like we check if `migrationInfo` refers to links that actually exist in state. If I'm not mistaken, this could trigger a panic when `deleteMigrationKeys` right?

AlpinYukseloglu (commented):
Left another round of comments – some were important enough for me to hesitate on approving, but will be quick to ACK once resolved

x/gamm/keeper/migrate.go
czarcas7ic: It seems appropriate for us to panic if gov attempts create a link to non-existent pools

AlpinYukseloglu: Hmm right but we should be catching this in validation logic so the proposal in invalidated without ever triggering the panic right?

Might be a nonissue so feel free to disregard if you're confident this is fine

czarcas7ic: Yeah I agree that this should be checked in validation logic (I thought it was, but I could be wrong)

AlpinYukseloglu: Should have clarified – what prompted the comment was the realization that we don't check this in validation logic (likely since it's general validation that also covers the case where records don't already exist in state):
https://github.com/osmosis-labs/osmosis/blob/eb54eaff14d84bfda1c6febbd93377e3e2c57e9d/x/gamm/keeper/migrate.go#L149

czarcas7ic: @AlpinYukseloglu do you think if we just remove deleteMigrationKeys from Overwrite that this should be good?

AlpinYukseloglu: This makes much more sense, thanks for the clarification

x/gamm/keeper/migrate.go
czarcas7ic: Discussed on call, deletion makes sense when using this method for both Replace and Update, where both methods provide the entire record to be set post deletion

x/gamm/keeper/migrate.go
czarcas7ic: used osmoutils for getting and setting here https://github.com/osmosis-labs/osmosis/pull/5400/commits/46e62fc6bd262659cd7f9ab5ecc310353d68025d

note this added more complexity, and I think when values are simple uints that marshalling related errors are probably not worth the added complexity when compared to using osmoutils MustSet for custom proto structs

x/gamm/keeper/migrate.go
czarcas7ic: added this as a helper method in osmoutils https://github.com/osmosis-labs/osmosis/pull/5400/commits/ad924fbdd9b0ef0251dd7bebb97b2ca537bc65bf

x/gamm/keeper/migrate.go
czarcas7ic: I actually reverted this change since it was adding a lot of complexity for unmarshalling a uint64 value

czarcas7ic: https://github.com/osmosis-labs/osmosis/pull/5400/commits/f9fef541ef3443b1bdfeca2462288afc1fa4ada0

AlpinYukseloglu (approved):

x/gamm/keeper/migrate.go
czarcas7ic: Let him know, thanks!

Issue #5391, opened 2023-06-01
Fix import paths github action 


pysel:
## Background

The import paths workflow seems broken:
https://github.com/osmosis-labs/osmosis/actions/runs/5115321530

## Acceptance

The workflow works when manually running it

pysel:
#5401 fixes the logic for getting the `version_to_replace` variable

there is another issue with this workflow:

the workflow will still fail, because after updating import paths to (for example) v16 from v15, [some](https://github.com/osmosis-labs/osmosis/blob/main/tests/cl-go-client/go.mod#L8) recently introduced modules will have an invalid version set. Ex:

How go.mod of `cl-go-client` looks after workflow run (from v15 to v16):
```
github.com/osmosis-labs/osmosis/v16 v15.0.0-20230530235412-9f091596cbcf
```

error it results in is:

> version "v15.0.0-20230530235412-9f091596cbcf" invalid: should be v16, not v15


[Here](https://github.com/osmosis-labs/osmosis/actions/runs/5150914405/jobs/9275587891) is how this error looks in gh actions.

Question to reviewers:
Are we running this workflow after the release is created or before? In former case, it would be easy to simply change the version of `github.com/osmosis-labs/osmosis/vX` in modules that import it

p0mvn:
Before

p0mvn:
@pysel thanks for the timely update of this script! Do you think you could also auto-apply state compatibility and changelog labels to the PR created by this CI job?

p0mvn:
Also, FYI, the proto-gen CI check is failing: https://github.com/osmosis-labs/osmosis/actions/runs/5159223921/jobs/9293878347?pr=5408

p0mvn:
The script doesn't seem to run `make proto-all` at the end, can we fix that?

pysel:
the logic for proto has not been changed, but it only runs `proto-gen`, w/o formatting them

pysel:
Closing as resolved

PR #5401, opened 2023-06-02
Fix update import paths workflow

pysel:
Part of: #5391 

- [x] Fix regex for getting `version_to_replace`
Fixes the logic for getting `version_to_replace` from the list of go modules (`go list -m`). It used to search for pattern `g.*v` and save a string after the `v` symbol. Before `epochs` became a standalone module, it worked, because the path with the version (aka `github.com/osmosis-labs/osmosis/vX`) was printing at the end of the list, so saving everything after `v` worked. 
After epochs have been added, the same logic returned `X github.com/osmosis-labs/osmosis/x/epochs` and saved it to `version_to_replace`. This issue can be seen in this run: https://github.com/osmosis-labs/osmosis/actions/runs/5115321530/jobs/9222845233

- [x] Do not automatically update paths in go.mods, which explicitly reference some version of previous version ([example](https://github.com/osmosis-labs/osmosis/blob/main/tests/cl-go-client/go.mod#L8))
As this is an urgent matter, I suggest we merge this PR, run the workflow, and then manually update `tests/cl-genesis-positions` and `tests/cl-go-client`, but I want to also add reminder into the body of created PR that some paths might need manual update

## Testing and Verifying

Manually tested that new regex results in correct `version_to_create`

N/A


pysel:
for reviewers: https://github.com/osmosis-labs/osmosis/issues/5391#issuecomment-1573037376

CC: @p0mvn @stackman27

pysel:
PR that was created by the run of the workflow: https://github.com/osmosis-labs/osmosis/pull/5403

CC: @stackman27 @p0mvn

p0mvn (approved):
Thanks for the quick PR sir! Any suggestions on how we can fix `cl-go-client` and avoid manual updates in the future?

pysel:
> Thanks for the quick PR sir! Any suggestions on how we can fix `cl-go-client` and avoid manual updates in the future?

I will create a separate issue to discuss this matter

PR #5409, opened 2023-06-02
[chore] update all gomod sdk version

stackman27:
Closes: #XXX

update sdk to the latest commit


p0mvn (approved):
LGTM as long as e2e passes, nice work on getting expedited proposals and SDK fork updates over the finish line

p0mvn:
We likely need to set the parameter in the upgrade handler and update e2e test. Also, changelog should be updated

PR #5434, opened 2023-06-06
fix: mutation testing script

czarcas7ic:
Closes: #XXX

There was an issue with the mutation-test.sh script, where it was attempting to search for modules in the same folder, but it needed to search one folder up.

Additionally, I added a feature where if no module names are given as an argument, it runs mutation testing on all modules.


AlpinYukseloglu (approved):
Thanks for catching/fixing!

PR #5441, opened 2023-06-06
fix(CL): user positions key parsing

p0mvn:
Closes: #XXX

Found a problem when looking into the UserPositions query issue. We parse the key and split by "key separator" which is a character "|".

However, in the key there are address bytes. These address bytes can contain this character. As a result, our parser is wrong.

There are 2 solutions:
1. Use bech32 address in the key. Its alphabet does not contain "|" character so we should be good when parsing
* Pros: safer and lets us avoid future problems
* Cons: state breaking
2. Search for the last index of the key separator instead of splitting by it. Since address bytes are in the beginning of the key, they will not break the parsing logic
* Pros: not state-breaking and FE can continue using their existing test env to integrate
* Cons: it's easy to make the same mistake in the future

I went with approach 2 to avoid breaking FE integrating.

## Testing and Verifying

  - [x] Code comments?
  - [ ] N/A


x/concentrated-liquidity/position.go
```diff
@@ -126,18 +126,27 @@ func (k Keeper) GetAllPositionIdsForPoolId(ctx sdk.Context, prefix []byte, poolI
 	for ; iterator.Valid(); iterator.Next() {
 		key := iterator.Key()
 
-		// Extract the components from the key
-		parts := bytes.Split(key, []byte(types.KeySeparator))
-		if len(parts) != 4 {
-			return nil, fmt.Errorf("invalid key format: %s", key)
```
czarcas7ic, line 132: How come this error wasn't popping? If the addr contained the key sep bytes shouldn't this have been 5 parts?

p0mvn: There weren't tests reproducing this. If there were, it would. I will add a new test case prior to merge

czarcas7ic (approved):
thanks, hoping we can get approach 1 in before locking in code but not the end of the world if we cant

mattverse (approved):
Did not know that this was even possible, I think we should come back to other parts in code where we use key separators as well and double check if we don't have any issues in the future

x/concentrated-liquidity/types/keys.go
```diff
@@ -160,18 +165,24 @@ func KeyPositionId(positionId uint64) []byte {
...
+// WARNING: do not parse this index by splitting by key separator as it contains
+// raw address bytes in its structure.
 func KeyAddressAndPoolId(addr sdk.AccAddress, poolId uint64) []byte {
 	return []byte(fmt.Sprintf("%s%s%x%s%d", PositionPrefix, KeySeparator, addr.Bytes(), KeySeparator, poolId))
 }
 
 // KeyUserPositions returns the prefix key used to create KeyAddressPoolIdPositionId, which only includes the addr.
 // This key can be used to iterate over all positions that a specific address has.
+// WARNING: do not parse this index by splitting by key separator as it contains
+// raw address bytes in its structure.
```
ValarDragon, line 185: Wait are you sure this is raw bytes??? If so then we have problems, I thought this was hex encoded when I was going through it in #5470

p0mvn: ref: https://github.com/osmosis-labs/osmosis/pull/5470#discussion_r1225693845

p0mvn: What problems do we have? I thought that as long as we don't use a key separator for parsing keys, it should be fine

p0mvn: You're right we encode hex: https://github.com/osmosis-labs/osmosis/pull/5470#discussion_r1225697435

p0mvn:
We encode hex so should not be needing this: https://github.com/osmosis-labs/osmosis/pull/5470#discussion_r1225697435

PR #5442, opened 2023-06-06
chore: mutation test (1/x) incentives updates (1/x)

czarcas7ic:
Closes: #XXX

One of the issues pointed out by the CL mutation tests is that we do not check in any of our tests that the amount added back into the pool's uptime accumulator is equal to the forfeited amount (with the dust added back as well). 

When implementing this testing vector, I also realized we should really be returning forefeited incentives as a DecCoin rather than a Coin, since any time this value is used we are also utilizing the dust value now that we are reinvesting forfeited dust back into the accum.

You will notice that we still normalize coins when checking the expected rewards claimed, since this value stays an Int.

## Testing and Verifying

- Tests that perviously checked a normalized Coin from DecCoin for forfeited incentives are now updated to check the exact decimal amount
- Test vector added to check that the amount being added back to the accum when forfeited is correct

  - [ ] Code comments?
  - [ ] N/A


proto/osmosis/concentrated-liquidity/query.proto
```diff
@@ -208,7 +208,8 @@ message ClaimableIncentivesResponse {
     (gogoproto.moretags) = "yaml:\"claimable_incentives\"",
     (gogoproto.nullable) = false
   ];
-  repeated cosmos.base.v1beta1.Coin forfeited_incentives = 2 [
+  repeated cosmos.base.v1beta1.DecCoin forfeited_incentives = 2 [
```
p0mvn, line 211: What is the benefit of returning `DecCoin`? I'm supportive of adding the missed tests but I think returning `Coin` as it is now is fine

czarcas7ic: The benefit or returning DecCoin is it simplifies test cases. Without adding this change, we need to determine the dust prior to claiming incentives, then add that dust to the amount that the method says that it forfeits. In reality, this is the actual value it forfeits, so it makes the most sense imo to return this value. Otherwise we are doing hacky work arounds to determine what the value we actually forfeited which feels more odd

czarcas7ic: Maybe we can modify the user query to return a coin object instead, but this feels the most natural since we need to know this value to determine if incentives were properly forfeited

p0mvn: Do we want this API-breaking change this late in the game? What do you think about waiting until `v16.x` branch is cut and then merge this to `main` so that it goes into v17?

My only concern is causing unnecessary friction with integrations for something that is mostly useful for test convenience

mattverse: I am for merging this PR, it feels like the API breakage here can't be dangerous + this is the correct value anyways that we should have been returning

czarcas7ic: I think it makes sense for the API to return the actual amount forfeited

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	cosmossdk.io/errors v1.0.0-beta.7
 	github.com/CosmWasm/wasmd v0.31.0
 	github.com/cosmos/cosmos-proto v1.0.0-beta.2
-	github.com/cosmos/cosmos-sdk v0.47.2
+	github.com/cosmos/cosmos-sdk v0.47.3
```
AlpinYukseloglu, line 9: Is this sdk version bump intentional to have in this PR?

czarcas7ic: Yeah this is automatic, and we should commit it since we replace at the end. Otherwise everyone's PR will keep showing this

x/concentrated-liquidity/incentives_test.go
```diff
@@ -1584,6 +1615,24 @@ func (s *KeeperTestSuite) TestInitOrUpdatePositionUptimeAccumulators() {
 			expectedInitAccumValue:   uptimeHelper.hundredTokensMultiDenom,
 			expectedUnclaimedRewards: uptimeHelper.emptyExpectedAccumValues,
 		},
+		"(lower < curr < upper) nonzero uptime trackers (position already existing)": {
```
p0mvn, line 1618:
```suggestion
		"(lower < curr < upper) non-zero uptime trackers (position already existing)": {
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -1600,6 +1649,24 @@ func (s *KeeperTestSuite) TestInitOrUpdatePositionUptimeAccumulators() {
 			expectedInitAccumValue:   uptimeHelper.twoHundredTokensMultiDenom,
 			expectedUnclaimedRewards: uptimeHelper.emptyExpectedAccumValues,
 		},
+		"(lower < upper < curr) nonzero uptime trackers (position already existing)": {
```
p0mvn, line 1652:
```suggestion
		"(lower < upper < curr) non-zero uptime trackers (position already existing)": {
```

x/concentrated-liquidity/incentives_test.go
```diff
@@ -1600,6 +1649,24 @@ func (s *KeeperTestSuite) TestInitOrUpdatePositionUptimeAccumulators() {
...
+				tickIndex:      50,
+				uptimeTrackers: wrapUptimeTrackers(uptimeHelper.threeHundredTokensMultiDenom),
+			},
+			existingPosition:         true,
+			addToGlobalAccums:        uptimeHelper.threeHundredTokensMultiDenom,
+			positionId:               DefaultPositionId,
+			currentTickIndex:         51,
+			globalUptimeAccumValues:  uptimeHelper.fourHundredTokensMultiDenom,
+			expectedInitAccumValue:   uptimeHelper.twoHundredTokensMultiDenom,
+			expectedUnclaimedRewards: uptimeHelper.emptyExpectedAccumValues,
```
p0mvn, line 1668: How is this calculated? Could you please add comments?

x/concentrated-liquidity/incentives_test.go
```diff
@@ -2884,6 +2991,11 @@ func (s *KeeperTestSuite) TestQueryAndClaimAllIncentives() {
 				}
 				s.Require().NoError(err)
 
+				communityPoolBalanceAfter := s.App.BankKeeper.GetAllBalances(s.Ctx, communityPoolAddress)
+				fmt.Println("communityPoolBalanceBefore", communityPoolBalanceBefore)
+				fmt.Println("communityPoolBalanceAfter", communityPoolBalanceAfter)
+				fmt.Println()
```
p0mvn, line 2997: Are these still useful?

p0mvn (approved):
LGTM

x/concentrated-liquidity/incentives_test.go
czarcas7ic: They are not, removed here https://github.com/osmosis-labs/osmosis/pull/5442/commits/0d3a36c6fce8d8c6a715abaf5c742a5143f4a581

x/concentrated-liquidity/incentives_test.go
czarcas7ic: Added more comments here for clarity! https://github.com/osmosis-labs/osmosis/pull/5442/commits/bb8298df493789d086491dae215ddc9ceadb84eb

PR #5444, opened 2023-06-06
feat: run mutation script on specific files

czarcas7ic:
Closes: #XXX

The mutation script allows users to specify specific modules to run mutation tests on. These tests still take a really long time, and many times a user might just want to run on a specific file. This modification allows for this. For example:

```
./mutation-test.sh "concentrated-liquidity" "../x/concentrated-liquidity/incentives.go"
```

This runs mutation tests on just incentives.go file and nothing else.

### DIFF
--- a/scripts/mutation-test.sh
+++ b/scripts/mutation-test.sh
@@ -1,39 +1,32 @@
 #!/usr/bin/env bash
 
 set -eo pipefail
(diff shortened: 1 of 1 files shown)


AlpinYukseloglu (approved):
Great addition, thanks

PR #5455, opened 2023-06-07
fix: sf unbonding migration patch

czarcas7ic:
Closes: #XXX

While creating E2E tests for CL, @ThanhNhann from Notional discovered the following bug:

In the `migrateSuperfluidUnbondingBalancerToConcentrated` method, we call createSyntheticLockup https://github.com/osmosis-labs/osmosis/blob/41af9f9708ad25053b9a4a5660affd0ff1bae7b0/x/superfluid/keeper/migrate.go#L169-L173 at the very end in an unlocking status. This is so that if, after their migration their validator gets slashed, they will also still get slashed approppriatly.

The bug is, in `createSyntheticLockup` https://github.com/osmosis-labs/osmosis/blob/41af9f9708ad25053b9a4a5660affd0ff1bae7b0/x/superfluid/keeper/synthetic_lock_wrapper.go#L42-L55 we use the unbondingTime parameter as the unbonding duration for the synth lock. So even if the user had 4 mins left, their synth lock would persist for the UnbondingTime parameter.

The fix is simple, adding a `createSytheticLockupWithDuration` method that takes in a duration rather than using the parameter directly

## Testing and Verifying

`TestMigrateNonSuperfluidLockBalancerToConcentrated` adds an hour wait block time increase to better simulate an unblonding / unlocking lock that would be calling this method. This test case properly fails with the old code, confirming the bug was valid.

  - [ ] Code comments?
  - [ ] N/A

### DIFF
--- a/x/superfluid/keeper/migrate.go
+++ b/x/superfluid/keeper/migrate.go
@@ -166,8 +166,8 @@ func (k Keeper) migrateSuperfluidUnbondingBalancerToConcentrated(ctx sdk.Context
 		return 0, sdk.Int{}, sdk.Int{}, sdk.Dec{}, time.Time{}, 0, 0, 0, err
 	}
 
-	// Create a new synthetic lockup for the new intermediary account in an unlocking status
-	err = k.createSyntheticLockup(ctx, concentratedLockId, clIntermediateAccount, unlockingStatus)
+	// Create a new synthetic lockup for the new intermediary account in an unlocking status for the remaining duration.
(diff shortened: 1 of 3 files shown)


ThanhNhann (commented):
The bug is solved with this fix, just need to pass CI and we can merge it

czarcas7ic:
I pushed the updated test but it is not showing yet because github PRs are degraded ATM

mattverse (approved):
Good catch, LGTM

PR #5458, opened 2023-06-07
feat/docs: CL external incentives tool

p0mvn:
Closes: #XXX

This PR modifies the script to add the ability to create external incentives via gauges and wait for an epoch so that it is converted into an incentive record.

## Testing and Verifying

  - [ ] Code comments?
  - [x] N/A

### DIFF
--- a/tests/cl-go-client/main.go
+++ b/tests/cl-go-client/main.go
@@ -267,7 +267,7 @@ func createExternalCLIncentive(igniteClient cosmosclient.Client, poolId uint64, 
 		log.Println("could not find information about previous epoch. If duration is too long, this test might be infeasible")
 	}
 
-	//.Create gauge
+	// Create gauge
 	runMessageWithRetries(func() error {
 		return createGauge(igniteClient, expectedPoolId, accountName, gaugeCoins)
 	})
--- a/tests/cl-go-client/main.go
+++ b/tests/cl-go-client/main.go
@@ -17,10 +17,14 @@ import (
 	"github.com/ignite/cli/ignite/pkg/cosmosaccount"
 	"github.com/ignite/cli/ignite/pkg/cosmosclient"
 
+	clqueryproto "github.com/osmosis-labs/osmosis/v16/x/concentrated-liquidity/client/queryproto"
 	"github.com/osmosis-labs/osmosis/v16/x/concentrated-liquidity/model"
 	cltypes "github.com/osmosis-labs/osmosis/v16/x/concentrated-liquidity/types"
+	incentivestypes "github.com/osmosis-labs/osmosis/v16/x/incentives/types"
+	lockuptypes "github.com/osmosis-labs/osmosis/v16/x/lockup/types"
 	poolmanagerqueryproto "github.com/osmosis-labs/osmosis/v16/x/poolmanager/client/queryproto"
 	poolmanagertypes "github.com/osmosis-labs/osmosis/v16/x/poolmanager/types"
+	epochstypes "github.com/osmosis-labs/osmosis/x/epochs/types"
 )
 
 // operation defines the desired operation to be run by this script.
@@ -38,6 +42,12 @@ const (
 	// and swaps it back while accounting for the spread factor. This is done to
 	// ensure that we cross ticks while minimizing the chance of running out of funds or liquidity.
 	makeManyInvertibleLargeSwaps
+
+	// createExternalCLIncentives creates external CL incentives.
+	createExternalCLIncentives
+
+	// createPoolOperation creates a pool with expectedPoolId.
+	createPoolOperation
 )
 
 const (
@@ -49,19 +59,22 @@ const (
 	denom1                          = "uusdc"
 	tickSpacing              int64  = 100
 	accountNamePrefix               = "lo-test"
-	numPositions                    = 100
-	numSwaps                        = 100
-	minAmountDeposited              = int64(1_000_000)
-	randSeed                        = 1
-	maxAmountDeposited              = 1_00_000_000
-	maxAmountSingleSwap             = 1_000_000
-	largeSwapAmount                 = 90_000_000_000
+	// Note, this is localosmosis-specific.
+	expectedEpochIdentifier = "hour"
+	numPositions            = 100
+	numSwaps                = 100
+	minAmountDeposited      = int64(1_000_000)
+	randSeed                = 1
+	maxAmountDeposited      = 1_00_000_000
+	maxAmountSingleSwap     = 1_000_000
+	largeSwapAmount         = 90_000_000_000
 )
 
 var (
 	defaultAccountName  = fmt.Sprintf("%s%d", accountNamePrefix, 1)
 	defaultMinAmount    = sdk.ZeroInt()
 	defaultSpreadFactor = sdk.MustNewDecFromStr("0.001")
+	externalGaugeCoins  = sdk.NewCoins(sdk.NewCoin("uosmo", sdk.NewInt(1000_000_000)))
 	accountMutex        sync.Mutex
 )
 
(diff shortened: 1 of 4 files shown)


tests/cl-go-client/main.go
```diff
@@ -231,6 +238,78 @@ func swapGivenLargeAmountsBothDirections(igniteClient cosmosclient.Client, poolI
...
+	clQueryClient := clqueryproto.NewQueryClient(igniteClient.Context())
+
+	incentiveRecordsResponse, err := clQueryClient.IncentiveRecords(context.Background(), &clqueryproto.IncentiveRecordsRequest{
+		PoolId: expectedPoolId,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Println("incentive records. If empty, something probably went wrong", incentiveRecordsResponse.IncentiveRecords)
```
stackman27, line 310: can we also check gauge Records after we distribute? FilledEpoch should be +1 and distribtued Coins should be appended to the gauge struct

wait also i'm surprised this didn't fail here(https://github.com/osmosis-labs/osmosis/blob/036b94d3e220959e1ecc6050674c06de60f548f6/x/incentives/keeper/distribute.go#L393)

stackman27: also should we consider all the created incentives regardless of poolId <> gaugeId association and distribute tokens to them as well?

p0mvn: The purpose of this is to make sure that incentive records are created. This print gets the job done.

I don't think what you're requesting is helpful for this tool. However, it would be a good assertion to make in tests

tests/cl-go-client/main.go
```diff
@@ -231,6 +238,78 @@ func swapGivenLargeAmountsBothDirections(igniteClient cosmosclient.Client, poolI
...
+	}
+
+	if len(epochInfosRespose.Epochs) > 0 {
+		lastEpochInfo := epochInfosRespose.Epochs[len(epochInfosRespose.Epochs)-1]
+		log.Println("epoch duration", lastEpochInfo, "next epoch start time", lastEpochInfo.StartTime.Add(lastEpochInfo.Duration))
+	} else {
+		log.Println("could not find information about previous epoch. If duration is too long, this test might be infeasible")
+	}
+
+	//.Create gauge
```
mattverse, line 270:
```suggestion
	//Create gauge
```

tests/cl-go-client/main.go
```diff
@@ -231,6 +238,78 @@ func swapGivenLargeAmountsBothDirections(igniteClient cosmosclient.Client, poolI
...
+	epochInfosRespose, err := epochsQueryClient.EpochInfos(context.Background(), &epochstypes.QueryEpochsInfoRequest{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if len(epochInfosRespose.Epochs) > 0 {
+		lastEpochInfo := epochInfosRespose.Epochs[len(epochInfosRespose.Epochs)-1]
+		log.Println("epoch duration", lastEpochInfo, "next epoch start time", lastEpochInfo.StartTime.Add(lastEpochInfo.Duration))
+	} else {
+		log.Println("could not find information about previous epoch. If duration is too long, this test might be infeasible")
```
mattverse, line 267: what do you mean by "if the duration is too long"? Are you referring to the epoch duration? This is the case where there aren't any epoch registered though right?

tests/cl-go-client/main.go
```diff
@@ -309,6 +388,57 @@ func makeSwap(client cosmosclient.Client, poolId uint64, senderKeyringAccountNam
 	return resp.TokenOutAmount, nil
 }
 
+func createGauge(client cosmosclient.Client, poolId uint64, senderKeyringAccountName string, gaugeCoins sdk.Coins) error {
+	accountMutex.Lock() // Lock access to getAccountAddressFromKeyring
+	senderAddress := getAccountAddressFromKeyring(client, senderKeyringAccountName)
+	accountMutex.Unlock() // Unlock access to getAccountAddressFromKeyring
```
mattverse, line 394: out of curiosity, what's the reason we need to do this?

tests/cl-go-client/main.go
```diff
@@ -309,6 +388,57 @@ func makeSwap(client cosmosclient.Client, poolId uint64, senderKeyringAccountNam
...
+	if err != nil {
+		if !strings.Contains(err.Error(), poolmanagertypes.FailedToFindRouteError{PoolId: expectedPoolId}.Error()) {
+			log.Fatal(err)
+		}
+		createdPoolId := createPool(igniteClient, defaultAccountName)
+		if createdPoolId != expectedPoolId {
+			log.Fatalf("created pool id (%d), expected pool id (%d)", createdPoolId, expectedPoolId)
+		}
+	} else {
+		log.Println("pool already exists. Tweak expectedPoolId variable if you want another pool, current expectedPoolId", expectedPoolId)
```
mattverse, line 438:
```suggestion
		log.Println("pool already exists. Tweak expectedPoolId variable if you want another pool, current expectedPoolId: ", expectedPoolId)
```

tests/cl-go-client/main.go
```diff
@@ -231,6 +238,78 @@ func swapGivenLargeAmountsBothDirections(igniteClient cosmosclient.Client, poolI
...
+		// Wait for 1 epoch to pass
+		currentEpochResponse, err = epochsQueryClient.CurrentEpoch(context.Background(), &epochstypes.QueryCurrentEpochRequest{
+			expectedEpochIdentifier,
+		})
+		if err != nil {
+			log.Fatal(err)
+		}
+		if epochAfterGaugeCreation == -1 {
+			log.Println("current epoch after gauge creation", currentEpochResponse.CurrentEpoch)
+			log.Println("waiting for next epoch...")
```
mattverse, line 286: does this get printed for an hour?

mattverse: maybe we can set a timer and retry every minute

mattverse (approved):
utACK.

A little concerned about the 

> log.Println("current epoch after gauge creation", currentEpochResponse.CurrentEpoch)
> 			log.Println("waiting for next epoch...")

getting logged for an hour, otherwise lgtm

tests/cl-go-client/main.go
p0mvn: Like if it is 1 full day.

The script is currently set up for localosmosis where epoch duration is 60 seconds.

tests/cl-go-client/main.go
p0mvn: On localosmosis, unintuitively, an "hour" epoch is actually 60 seconds so this shouldn't be a concern.

If this is used on testnet, there is no "hour" epoch so the script would fail early.

Thi script is mainly used for testing incentives on localosmosis. If there is an infinite loop somewhere, the user can just kill the script

mattverse: Ah I see, now I understand why everything was so confusing while reviewing this, now everything became more clearer

tests/cl-go-client/main.go
```diff
@@ -49,19 +59,22 @@ const (
...
 	accountNamePrefix               = "lo-test"
-	numPositions                    = 100
-	numSwaps                        = 100
-	minAmountDeposited              = int64(1_000_000)
-	randSeed                        = 1
-	maxAmountDeposited              = 1_00_000_000
-	maxAmountSingleSwap             = 1_000_000
-	largeSwapAmount                 = 90_000_000_000
+	// Note, this is localosmosis-specific.
+	expectedEpochIdentifier = "hour"
```
ValarDragon, line 63: Are we going to run into a problem trying to use this for cl-testnet?

p0mvn: The epoch identifier can manually be changed to "day" here.

A bigger problem is that we would have to wait for the epoch to pass for a while so that the incentive records are created

p0mvn: I will try running this against testnet before I merge and document all the changes that would be necessary

p0mvn:
OK, I'm going to merge this script to unblock the external incentives PR branched off of this. I asked Ibrar to test this out to make sure that all needed functionality is present

